PR #105, opened 2014-12-20
Create STPAPIClient

jflinter:
The objectives here are to:
- Move static methods on the `Stripe` class into instance methods on a new `STPAPIClient` class. This will allow us to centralize configuration there and reduce the insane number of `+[createTokenWithX:operationQueue:etc]` methods we currently have.
- Reduce the total number of classes/methods available in public headers, so that a) we can document things more easily and b) it's clearer, just from looking at the code, how to use the library.
- Do all of this in such a way that it won't break everyone's existing integration. Since I'm making classes like `STPAPIConnection` private, this will introduce *some* breaking changes, but I've taken steps to ensure most integrations should be fine (all the old static token-creation methods on `Stripe` still exist, but all fall through to `STPAPIClient` now).

Still to do:
- [x] Document the new classes
- [x] Add deprecation warnings to now-discouraged static methods
- [x] Write more tests for `STPAPIClient`
- [x] Update the example app to use `STPAPIClient`
- [x] Update the podspec
- [x] Update the README with updated examples, as well as notes on how to migrate
- [ ] Update the docs on the site

At this point, I think the code itself is mostly done now: the example app works with both old and new code. Mind taking a look @raycmorgan ?


Stripe/STPAPIClient.m
```diff
@@ -0,0 +1,495 @@
...
+#import "STPAPIClient.h"
+#import "STPBankAccount.h"
+#import "STPCard.h"
+#import "STPToken.h"
+#import "StripeError.h"
+
+static NSString *const apiURLBase = @"api.stripe.com";
+static NSString *const apiVersion = @"v1";
+static NSString *const tokenEndpoint = @"tokens";
+static NSString *const kStripeiOSVersion = @"2.2.2";
```
raycmorgan: Need bumping?

raycmorgan:
Overall big +1 on this change. You mention that this leaves some breaking changes, should this be bumped to version `3.0.0` for semver reasons?

README.md
```diff
@@ -66,6 +65,18 @@ These methods follow the validation method convention used by [key-value validat
...
 
+## Migration Guides
+
+### Migrating from versions < 2.3
+
+Before version 2.3, most token-creation methods were class methods on the `Stripe` class. These are now all instance methods on the `STPAPIClient` class. Where previously you might write
+    [Stripe createTokenFromCard:card publishableKey:myPublishableKey completion:completion];
+you would now instead write
+    STPAPIClient *client = [[STPAPIClient alloc] initWithPublishableKey:myPublishableKey];
+    [client createTokenFromCard:card completion:completion];
```
raycmorgan: MD formatting is not working here. Probably better to use triple-tick.

raycmorgan:
Looks good to me.

:+1:

PR #106, opened 2014-12-28
Implementation of the real amount (not $10)

lfarah:
The amount for the payment was being collected through self.amount, but on the payment flow, it was hardcoded with $10. This is solved on my pull request


jflinter:
Thanks!

PR #115, opened 2015-01-29
Fix success case

chriseidhof:
In case the callback returns success, we need to set `self.backendChargeSuccessful` to `YES`. If we don't do this, the `STPCheckoutEventFinish` will never call the finish delegate method with a success status.

### DIFF
--- a/Stripe/Checkout/STPCheckoutViewController.m
+++ b/Stripe/Checkout/STPCheckoutViewController.m
@@ -142,10 +142,11 @@
         [self.checkoutDelegate checkoutController:self
                                    didCreateToken:token
                                        completion:^(STPBackendChargeResult status, NSError *error) {
+                                           self.backendChargeSuccessful = (status == STPBackendChargeResultSuccess);
+
                                            if (status == STPBackendChargeResultSuccess) {
                                                [adapter evaluateJavaScript:payload[@"success"]];
                                            } else {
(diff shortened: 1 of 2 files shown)


jflinter:
Hey @chriseidhof - thanks for pointing this out. I ended up fixing this ever-so-slightly-differently in 5d1cd71b8029dd15383e4f52590a9ba11bd007d6 and 362652737357c536323becb62a1b9d7b8f6645c4, so closing this PR but only in spirit. Cheers mate!

chriseidhof:
Yay!

On Fri, Feb 20, 2015 at 2:16 AM, Jack Flintermann <notifications@github.com>
wrote:

> Hey @chriseidhof <https://github.com/chriseidhof> - thanks for pointing
> this out. I ended up fixing this ever-so-slightly-differently in 5d1cd71
> <https://github.com/stripe/stripe-ios/commit/5d1cd71b8029dd15383e4f52590a9ba11bd007d6>
> and 3626527
> <https://github.com/stripe/stripe-ios/commit/362652737357c536323becb62a1b9d7b8f6645c4>,
> so closing this PR but only in spirit. Cheers mate!
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/stripe/stripe-ios/pull/115#issuecomment-75173432>.
>



-- 
Chris Eidhof

PR #123, opened 2015-03-04
Add scroll-position-based checkout appearance methods

jflinter:


Stripe/Checkout/STPCheckoutInternalUIWebViewController.m
```diff
@@ -81,33 +88,6 @@ - (void)viewDidLoad {
...
-    if ([UIDevice currentDevice].userInterfaceIdiom == UIUserInterfaceIdiomPad) {
-        bottomMargin = 0;
-    }
-    [self.view addConstraint:[NSLayoutConstraint constraintWithItem:headerBackground
-                                                          attribute:NSLayoutAttributeHeight
-                                                          relatedBy:NSLayoutRelationEqual
-                                                             toItem:self.view
-                                                          attribute:NSLayoutAttributeHeight
-                                                         multiplier:1
-                                                           constant:bottomMargin]];
```
jamesreggio: Did this code ever work? What did it do?

jamesreggio:
Is it possible to just do something like this?

```objc
[self.navigationBar setBarTintColor: [UIColor something]];
self.navigationBar.translucent = NO;
```

I know it wouldn't hide the bar, but I'm not convinced that that's a superior user experience.

PR #126, opened 2015-03-22
Fixed UIActivityIndicatorViewStyle being left unused

schayes04:
Fixed an issue where UIActivityIndicatorViewStyle was set but the value never got used


jflinter:
Nice catch, thank you.

PR #128, opened 2015-04-02
Remove revoked certificate testing

jflinter:


berg:
LGTM! (Assuming that securiteam has signed off, etc., etc.!)

PR #130, opened 2015-04-03
Fixed custom example app "Unsupported content-type: text/html" Issue

Jonyzfu:
Setting RequestOperationManager Response Serializer to
HTTPResponseSerializer fixed the issue.


jflinter:
Yeah, this is actually a pretty decent fix here given that the example iOS backend returns text, not JSON. Thanks for the patch.

PR #131, opened 2015-04-07
iTunes submission fails when app includes Stripe iOS dynamic framework

samritchie:
The deployment target on the iOS framework is '6.0', which generates validation failures when submitted to iTunes. I’ve changed this to '8.0'.

### DIFF
--- a/Stripe.xcodeproj/project.pbxproj
+++ b/Stripe.xcodeproj/project.pbxproj
@@ -1280,7 +1280,7 @@
 				GCC_WARN_UNINITIALIZED_AUTOS = YES_AGGRESSIVE;
 				INFOPLIST_FILE = Stripe/Info.plist;
 				INSTALL_PATH = "$(LOCAL_LIBRARY_DIR)/Frameworks";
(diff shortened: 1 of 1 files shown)


jflinter:
Interesting. I wasn't aware Apple was doing that kind of submission-time validation. Sorry for the trouble you're seeing.

This is made a little more complicated by the fact that we'd ideally like to run tests against earlier iOS versions in the simulator to ensure we don't accidentally use an old API, etc. I'll play around a bit to try and come up with an elegant solution here.

jflinter:
So I took this opportunity to shave a yak that I've been putting off for a while, namely, moving configuration out of the `.xcodeproj` into a set of `.xcconfig`s in #140. Doing that makes it trivial to have a different deployment target for the `Debug` and `Release` configurations, which makes it easy to run tests against the `Debug` configuration and build the framework for, well, release in the `Release` configuration. I'll publish new copies of the frameworks when I cut a new release over the next few days.

Anyways, @samritchie, I really appreciate you bringing this to our attention, and sorry for not merging your PR as it currently stands.

PR #132, opened 2015-04-11
Fixed cancel loading bug

sebyddd:
When you cancel the initial payment view by pressing the cancel UIBarButtonItem, it dismisses the previous view as well. didFailLoadWithError gets called when you cancel the loading as well so without checking it messes up the interface's hierarchy.


jflinter:
Great catch, thanks for the patch.

PR #133, opened 2015-04-18
Call evaluate javascript on the main thread

SlayterDev:
When the checkout form completes, the completion handler evaluates some javascript on the webview which can cause it to crash giving the following error message: 

`Tried to obtain the web lock from a thread other than the main thread or the web thread. This may be a result of calling to UIKit from a secondary thread. Crashing now...`

We simply fix this by ensuring the app evaluates the javascript on the main thread.

### DIFF
--- a/Stripe/Checkout/STPIOSCheckoutWebViewAdapter.m
+++ b/Stripe/Checkout/STPIOSCheckoutWebViewAdapter.m
@@ -39,7 +39,9 @@
 }
 
 - (void)evaluateJavaScript:(NSString *)js {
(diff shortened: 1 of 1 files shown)


jflinter:
Unfortunately, as of #123 `evaluateJavaScript:` now has a return type of `NSString *` instead of `void`, so this approach won't work. We can, however, handle it by always calling `STPCheckoutDelegate` methods on the main thread (it's funny that they weren't already - this implies that sometimes `UIWebView` calls methods on its delegate outside the main thread). #137 makes this change. As such, I'm going to close this PR, but appreciate you opening it!

PR #136, opened 2015-04-22
Add name of billing address to request params

chanpei:
Add name of billingAddress in ApplePay to request parameters


jflinter:
Nice catch, thank you!

PR #145, opened 2015-05-05
added NULL-check to fix a crash in formEncodedDataForPayment 

vytk:
This is caused by `ABRecordCopyValue()` sometimes returning `NULL`.
Without this `NULL`-check, the crash is triggered a couple lines down in `CFRelease(addressValues)`.

vytk_formencodeddataforpayment_abrecordcopyvalue_null_crash


jflinter:
Ah, good catch. Thank you for opening this!

vytk:
:)

PR #146, opened 2015-05-05
Fix UIImage loading in checkout

jflinter:


Stripe/Checkout/STPCheckoutInternalUIWebViewController.m
```diff
@@ -63,10 +63,10 @@ - (void)viewDidLoad {
...
 
     if (self.options.logoImage && !self.options.logoURL) {
-        NSURL *url = [NSURL fileURLWithPath:[NSTemporaryDirectory() stringByAppendingPathComponent:[[NSUUID UUID] UUIDString]]];
-        BOOL success = [UIImagePNGRepresentation(self.options.logoImage) writeToURL:url options:0 error:nil];
-        if (success) {
-            self.logoURL = self.options.logoURL = url;
+        NSString *base64 = [UIImagePNGRepresentation(self.options.logoImage) base64Encoding];
+        if (base64) {
+            NSString *dataURLString = [NSString stringWithFormat:@"data:png;base64,%@", base64];
+            self.logoURL = self.options.logoURL = [NSURL URLWithString:dataURLString];
```
jamesreggio: The `self.logoURL` reference is used to delete the temporary file, which no longer exists. You can remove that cleanup code and the variable.

Stripe/Checkout/STPCheckoutInternalUIWebViewController.m
```diff
@@ -63,10 +63,10 @@ - (void)viewDidLoad {
     [super viewDidLoad];
 
     if (self.options.logoImage && !self.options.logoURL) {
-        NSURL *url = [NSURL fileURLWithPath:[NSTemporaryDirectory() stringByAppendingPathComponent:[[NSUUID UUID] UUIDString]]];
-        BOOL success = [UIImagePNGRepresentation(self.options.logoImage) writeToURL:url options:0 error:nil];
-        if (success) {
-            self.logoURL = self.options.logoURL = url;
+        NSString *base64 = [UIImagePNGRepresentation(self.options.logoImage) base64Encoding];
```
jamesreggio: How does this do for non-PNG inputs? Does it convert? No-op? Crash?

Stripe/Checkout/STPCheckoutInternalUIWebViewController.m
jflinter: Great catch, thanks James. Can't believe I missed that.

Stripe/Checkout/STPCheckoutInternalUIWebViewController.m
jflinter: It'll return `nil`, which subsequent code should handle correctly. See https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIKitFunctionReference/#//apple_ref/c/func/UIImagePNGRepresentation

Stripe/Checkout/STPCheckoutInternalUIWebViewController.m
jflinter: Yeah, thinking about this a bit more - this can be done a lot better. Opened https://github.com/stripe/stripe-ios/pull/147 to that effect.

Stripe/Checkout/STPCheckoutInternalUIWebViewController.m
jamesreggio: Cool, thanks! (I should've just looked.)

PR #147, opened 2015-05-06
Further improve UI/NSImage handling in checkout

jflinter:
#146 fixes UIImage handling in checkout, but now that we're doing it with a `data:` URI and there's no temporary-file state to clean up, we can move all of this logic to `STPCheckoutOptions` where it belongs. (It'll also be way easier to test, when we finally get around to it).

This also fixes the issue for OSX, which is much easier now that everything lives in `STPCheckoutOptions`.

r? @jamesreggio

### DIFF
--- a/Stripe/Checkout/STPCheckoutOptions.m
+++ b/Stripe/Checkout/STPCheckoutOptions.m
@@ -67,6 +67,23 @@
     return [[NSString alloc] initWithData:[NSJSONSerialization dataWithJSONObject:values options:0 error:nil] encoding:NSUTF8StringEncoding];
 }
 
+- (void)setLogoImage:(STP_IMAGE_CLASS * __stp_nullable)logoImage {
+    _logoImage = logoImage;
+#if TARGET_OS_IPHONE
+    NSString *base64 = [UIImagePNGRepresentation(logoImage) base64Encoding];
+#else
+    NSData *imageData = [logoImage TIFFRepresentation];
+    NSBitmapImageRep *imageRep = [NSBitmapImageRep imageRepWithData:imageData];
+    imageData = [imageRep representationUsingType:NSPNGFileType
+                                       properties:@{NSImageCompressionFactor: @1.0}];
+    NSString *base64 = [imageData base64EncodedStringWithOptions:0];
+#endif
+    if (base64) {
+        NSString *dataURLString = [NSString stringWithFormat:@"data:png;base64,%@", base64];
+        self.logoURL = [NSURL URLWithString:dataURLString];
+    }
+}
+
 #pragma mark - NSCopying
 
 - (id)copyWithZone:(__unused NSZone *)zone {
--- a/Stripe/Checkout/STPCheckoutViewController.m
+++ b/Stripe/Checkout/STPCheckoutViewController.m
@@ -86,11 +86,11 @@
 @implementation STPCheckoutViewController
 
 - (instancetype)initWithNibName:(__unused NSString *)nibNameOrNil bundle:(__unused NSBundle *)nibBundleOrNil {
-    return [self initWithOptions:nil];
+    return [self initWithOptions:[[STPCheckoutOptions alloc] init]];
 }
 
 - (instancetype)initWithCoder:(__unused NSCoder *)coder {
-    return [self initWithOptions:nil];
+    return [self initWithOptions:[[STPCheckoutOptions alloc] init]];
 }
 
 - (instancetype)initWithOptions:(STPCheckoutOptions *)options {
--- a/Stripe/PublicHeaders/Checkout/STPCheckoutOptions.h
+++ b/Stripe/PublicHeaders/Checkout/STPCheckoutOptions.h
@@ -8,8 +8,10 @@
 #import <Foundation/Foundation.h>
 #if TARGET_OS_IPHONE
 #import <UIKit/UIKit.h>
+#define STP_IMAGE_CLASS UIImage
 #else
 #import <AppKit/AppKit.h>
+#define STP_IMAGE_CLASS NSImage
 #endif
 
 #import "STPNullabilityMacros.h"
@@ -40,11 +42,7 @@
 /**
  *  You can also specify a local UIImage to be used as the Checkout logo header (see logoURL).
  */
-#if TARGET_OS_IPHONE
-@property (nonatomic, stp_nullable) UIImage *logoImage;
-#else
-@property (nonatomic, stp_nullable) NSImage *logoImage;
-#endif
+@property (nonatomic, stp_nullable) STP_IMAGE_CLASS *logoImage;
 
 /**
  *  This specifies the color of the header shown in Stripe Checkout. If you specify a logoURL (but not a logoImage) and leave this property nil, Checkout will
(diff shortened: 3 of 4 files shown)


Example/Stripe iOS Example (Custom)/Constants.m
```diff
@@ -14,7 +14,7 @@
  */
 
 // This can be found at https://dashboard.stripe.com/account/apikeys
-NSString *const StripePublishableKey = nil; // TODO: replace nil with your own value
+NSString *const StripePublishableKey = @"pk_test_4TDXAGLdZFGNbXYGajQlcstU"; // TODO: replace nil with your own value
```
jamesreggio: You probably don't want this change, right?

Example/Stripe iOS Example (Custom)/ViewController.m
```diff
@@ -142,7 +142,9 @@ - (IBAction)beginStripeCheckout:(id)sender {
     STPCheckoutOptions *options = [[STPCheckoutOptions alloc] initWithPublishableKey:[Stripe defaultPublishableKey]];
     options.purchaseDescription = @"Cool Shirt";
     options.purchaseAmount = 1000; // this is in cents
+    options.requireBillingAddress = @YES;
     options.logoColor = [UIColor purpleColor];
```
jamesreggio: I'm guessing that this color isn't the best complement to the Dribble logo. Would you mind substituting the color-detection color in? It's this:

```
hex: "#bd1f5b"
rgb: {r: 188, g: 31, b: 91}
```

Stripe/Checkout/STPCheckoutViewController.m
```diff
@@ -86,11 +86,11 @@ @interface STPCheckoutViewController () <STPCheckoutDelegate>
 @implementation STPCheckoutViewController
 
 - (instancetype)initWithNibName:(__unused NSString *)nibNameOrNil bundle:(__unused NSBundle *)nibBundleOrNil {
-    return [self initWithOptions:nil];
+    return [self initWithOptions:[[STPCheckoutOptions alloc] init]];
```
jamesreggio: For my own education, what does this change do?

Stripe/Checkout/STPCheckoutWebViewAdapter.h
```diff
@@ -22,8 +22,8 @@ static NSString * __stp_nonnull const checkoutRedirectPrefix = @"/-/";
 static NSString * __stp_nonnull const checkoutUserAgent = @"Stripe";
 static NSString * __stp_nonnull const checkoutRPCScheme = @"stripecheckout";
 
-static NSString * __stp_nonnull const checkoutHost = @"checkout.stripe.com";
-static NSString * __stp_nonnull const checkoutURLString = @"https://checkout.stripe.com/v3/ios/index.html";
+static NSString * __stp_nonnull const checkoutHost = @"localhost:5394";
+static NSString * __stp_nonnull const checkoutURLString = @"http://localhost:5394/v3/ios/index.html";
```
jamesreggio: Also probably don't want this.

Stripe/PublicHeaders/Checkout/STPCheckoutOptions.h
```diff
@@ -8,8 +8,10 @@
 #import <Foundation/Foundation.h>
 #if TARGET_OS_IPHONE
 #import <UIKit/UIKit.h>
+#define STP_IMAGE_CLASS UIImage
 #else
 #import <AppKit/AppKit.h>
+#define STP_IMAGE_CLASS NSImage
 #endif
```
jamesreggio: :+1:

Example/Stripe iOS Example (Custom)/ViewController.m
```diff
@@ -142,7 +142,9 @@ - (IBAction)beginStripeCheckout:(id)sender {
...
+    options.logoImage = [UIImage imageNamed:@"dribbble"];
```
jamesreggio: Do you need to add this image into the repo?

Stripe/Checkout/STPCheckoutOptions.m
```diff
@@ -67,6 +67,23 @@ - (NSString *)stringifiedJSONRepresentation {
     return [[NSString alloc] initWithData:[NSJSONSerialization dataWithJSONObject:values options:0 error:nil] encoding:NSUTF8StringEncoding];
 }
 
+- (void)setLogoImage:(STP_IMAGE_CLASS * __stp_nullable)logoImage {
```
jamesreggio: I don't see this method being called in any of the example code, but it seems necessary in order for the Dribbble logo to appear. Could you explain the usage of this method?

jamesreggio:
Feedback is in. Thanks for making these improvements.

Example/Stripe iOS Example (Custom)/ViewController.m
jflinter: Agh, sorry - I mistakenly committed a few changes that I was using for local testing. (crazyidea: I should have some kind of gitignored local overrides file for this). This shouldn't be here.

Stripe/Checkout/STPCheckoutOptions.m
jflinter: This has to do with how objective-c `@property`s work. Let's say in a header you declare that a class has, say, `@property NSString *title`. This will implicitly create an instance variable named `_title`, as well as setters and getters of the format `- (void)setTitle:(NSString *)title` and `- (NSString *)title`.

You can optionally override those setters and getters (as I've done here) to make them have additional side effects, or change how the information is stored, or all kinds of other weird things. These methods will be called under the hood when you use dot-syntax to access or modify that property. So if elsewhere in my code I call `checkoutOptions.logoImage = myImage`, it's equivalent to writing `[checkoutOptions setLogoImage:myImage]`.

Properties are capable of doing way more work for you - you can see some additional qualifiers (e.g. `strong`/`weak`, `nonatomic`/`atomic`, `readwrite`/`readonly`, and as of recently `nonnull`/`nullable`) elsewhere in the source.

If this wasn't already too much detail, you can learn more at https://developer.apple.com/library/mac/documentation/Cocoa/Conceptual/ProgrammingWithObjectiveC/EncapsulatingData/EncapsulatingData.html

Stripe/Checkout/STPCheckoutViewController.m
jflinter: I've changed the interface for `STPCheckoutViewController` just slightly to annotate its `checkoutOptions` as a `nonnull` property - in other words, that value should never be `nil`. So you should always create an `STPCheckoutViewController` with a valid instance of `STPCheckoutOptions`. While `initWithOptions:` is the designated initializer for `STPCheckoutViewController`, it might not always be initialized that way - for example, it might be loaded from a nib (in which case `initWithCoder:` would be called instead). In this case I think it's slightly preferable to load an empty `STPCheckoutOptions` (which will be preconfigured with your publishable key, if you've set it via `[Stripe setDefaultPublishableKey:]`, which we encourage) instead of a `nil` object. Either way, the proper thing to do in this case would be to explicitly set the `checkoutOptions` property before presenting an alternatively-instantiated `STPCheckoutViewController`.

Stripe/Checkout/STPCheckoutOptions.m
jamesreggio: Oh cool, so this is automagically invoked. Thanks for explaining!

Stripe/Checkout/STPCheckoutViewController.m
jamesreggio: Makes sense, thanks.

PR #157, opened 2015-05-13
Fixing a potential memory leak of CFStrings

loudnate:
..by transfering ownership to ARC


jflinter:
:facepalm:

Thanks, Nate, good catch.

loudnate:
These actually were spotted by the Xcode analyzer. I can't take any credit!

PR #167, opened 2015-06-04
Fix issue with Apple Pay in iOS simple example

robseward:
The iOS simple example was failing when it should have been succeeding and vise-versa with Apple Pay.


jflinter:
Cool, thank you @robseward!

PR #172, opened 2015-06-15
Fix/apple pay failure issue

robseward:
… completed

Because the backend charge is asynchronous, completion was called with failure before the backend could respond with success (or failure).


jflinter:
Ah, good catch, thank you.

PR #174, opened 2015-06-18
Fixed a documentation issue on "STPCard.h".

sebaven:
The comment on the expYear property was copied from expMonth and not updated.


jflinter:
Ah, thanks, nice catch! Thanks for the PR : )

PR #176, opened 2015-07-02
prepare dependencies in examples for iOS9

benzguo:
r? @jflinter
May be difficult to review since there's so much noise. Mostly straightforward dependency changes, but definitely check me on https://github.com/stripe/stripe-ios/commit/b6156159440901f1d10e39e539e5246ee3014061. I think it's the right thing to do.


jflinter:
LGTM

PR #178, opened 2015-07-06
Add support for simulated cards in iOS9

benzguo:
r? @jflinter 

Note that the data in the `PKPayment` is, by necessity, mostly fake. 
`ApplePayStubs~4242424242424242~2000~USD~[UUID]`

Our SDK doesn't have access to the user's original `PKPaymentRequest` (since we're no longer stubbing the entire flow), and I'm not sure whether it's worth making the simulated PKPayment look more like a real PKPayment. If we decide that we do want the payment to look more real, I'm sure we could come up with a lurky workaround (e.g. swizzling `PKPaymentAuthorizationViewController`'s initializer).

In addition to the tests in this PR, I've manually verified that the Apple Pay flow works as expected in the simulator.

cc @raycmorgan

### DIFF
--- a/Example/Stripe iOS Example (Custom)/ViewController.m
+++ b/Example/Stripe iOS Example (Custom)/ViewController.m
@@ -6,8 +6,8 @@
 //  Copyright (c) 2014 Stripe. All rights reserved.
 //
 
-#import <Stripe/Stripe.h>
 #import <AFNetworking/AFNetworking.h>
+#import <Stripe/Stripe.h>
 
 #import "ViewController.h"
 #import "PaymentViewController.h"
--- a/Example/Stripe iOS Example (Custom)/ViewController.m
+++ b/Example/Stripe iOS Example (Custom)/ViewController.m
@@ -7,6 +7,7 @@
 //
 
 #import <Stripe/Stripe.h>
+#import <AFNetworking/AFNetworking.h>
 
 #import "ViewController.h"
 #import "PaymentViewController.h"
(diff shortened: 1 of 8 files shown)


Stripe/ApplePay/PKPayment+Stripe.h
```diff
@@ -0,0 +1,18 @@
...
+
+@import PassKit;
+
+@interface PKPayment (Stripe)
+
+/// Returns true if the instance is a payment from the simulator.
+- (BOOL)isSimulated;
+
+/// Returns a fake transaction identifier with the expected ~-separated format.
++ (NSString *)testTransactionIdentifier;
```
jflinter: You should probably prefix these.

Tests/Tests/PKPayment+StripeTest.m
```diff
@@ -0,0 +1,45 @@
+//
+//  PKPayment+StripeTest.m
+//  Stripe
+//
+//  Created by Ben Guo on 7/6/15.
+//  Copyright © 2015 Stripe, Inc. All rights reserved.
+//
+
+#import <XCTest/XCTest.h>
```
jflinter: @import

Example/Stripe iOS Example (Custom)/ViewController.m
```diff
@@ -7,6 +7,7 @@
 //
 
 #import <Stripe/Stripe.h>
+#import <AFNetworking/AFNetworking.h>
```
jflinter: @ import

Stripe/ApplePay/PKPayment+Stripe.h
```diff
@@ -0,0 +1,18 @@
+//
```
jflinter: Maybe double-check that this is included in all of the correct targets.

jflinter:
LGTM!

PR #181, opened 2015-07-22
Leaked STAPIConnection

pm-dev:
Removing association made from `STAPIConnection` to `STAPIClient`. Referencing `connection` in `STAPIClient.m` line 119 creates a retain cycle on `connection`. This association is actually not necessary because self is already retained by the completion block on line 97 and 115.


jflinter:
This seems reasonable - actually, it shouldn't really matter if an API client gets prematurely deallocated anyway.

PR #182, opened 2015-07-30
Adding a minimum deployment target.

istx25:
I was trying to deploy a build to iTunes Connect, but I was receiving cryptic error after cryptic error. I couldn't figure out what was going on. I reached out to @codafi and on our third hour, he caught the issue and saved the day. 

one

For some reason Xcode would not infer what the minimum deployment target was and since you had nothing set, Xcode was like: 

two

"Glad I could help, but now I'm inconsolable about the state of our tools." — @codafi

So…I've set the minimum deployment target to **8.0**. 

That was painful! :sob:

### DIFF
--- a/Stripe.xcodeproj/project.pbxproj
+++ b/Stripe.xcodeproj/project.pbxproj
@@ -1201,6 +1201,7 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = 04F39F101AEF2AFE005B926E /* StripeiOS-Debug.xcconfig */;
 			buildSettings = {
+				IPHONEOS_DEPLOYMENT_TARGET = 8.0;
 			};
 			name = Debug;
 		};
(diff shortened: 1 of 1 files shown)


jflinter:
Hmm, I'm a bit confused here, as we do have a deployment target set in `StripeiOS-Shared.xcconfig`. Generally speaking, we prefer all of the configuration to live in xcconfigs and not the project file, for the purposes of readability and cleaner merges. Can you say a bit more about how you were integrating this library into your app? Were you building the framework yourself, or using a build tool like cocoapods or carthage? Also, can you include some of the "cryptic error messages" you were seeing from iTunes connect?

istx25:
I used Carthage to add the Stripe SDK into my project.

istx25:
And some of those errors include: 
untitled

jflinter:
Hi there! Thanks for pasting in those warnings, that was really helpful. It looks like the issue was not that we didn't have a deployment target set, it's that the value was too low. I've fixed this in #186, mind making sure the current master branch works for you?

istx25:
I was using the latest version, but okay. :+1:

PR #186, opened 2015-08-04
Modify deployment targets

jflinter:
Previously I was setting the deployment target of `StripeiOS` to 6.0, which is pretty obviously wrong, but was useful for linting the project with FauxPas (as it would then catch any accidental usage of unavailable APIs) and didn't seem to cause any issues. #182 has shown that that assumption was wrong, and that deployment target breaks app store submission. This sets iOS 8.0 as the minimum deployment target for `StripeiOS`, but changes `StripeiOSTests` to use 6.0 so that FauxPas will catch errors there (which kind of semantically makes more sense anyway).

r? @benzguo


benzguo:
:white_check_mark:

istx25:
:+1:

PR #187, opened 2015-08-04
Remove cocoapods subspecs

jflinter:


Tests/installation_tests/carthage/Cartfile
```diff
@@ -0,0 +1 @@
+git "/Users/jack/stripe/stripe-ios" "jack-remove-subspecs"
```
benzguo: not sure what this Cartfile is about, but the local folder reference is lurky

benzguo:
:green_apple: lgtm

PR #200, opened 2015-08-18
Add card computed prop to STPPaymentCardTextField

bradfol:
I usually end up creating an extension to add this property anyway.


jflinter:
Hmm, yeah, this seems reasonable. Thanks!

PR #201, opened 2015-08-18
Add images for STPPaymentCardTextField to podspec

pablogb:
When using Cocoapods with Frameworks enabled, the brand images used by STPPaymentCardTextField are missing. Adding the resources to the podspec fixes this issue.


jflinter:
I ended up using a slightly more permissive regex to (hopefully) avoid future issues like this, but thanks for bringing this up @pablogb.

PR #206, opened 2015-08-19
Resolve missing dependency in STPPaymentCardTextField.h

loudnate:


jflinter:
My bad, thanks @loudnate. Out of curiosity, how was this breaking for you in practice? It hadn't raised any errors on my end.

loudnate:
This fails if you import the header directly. `#import <Stripe/STPPaymentCardTextField.h>`

importing the stripe framework, or even the Stripe.h header should be fine.

jflinter:
Good to know, thanks!

PR #218, opened 2015-08-27
Allow for [UIColor clearColor] as a valid background color for STPPay…

sportlabsMike:
…mentCardTextField

I've embedded the payment card text fields within an inner view with clipsToBounds=YES to avoid need for the interstitialView. This way when backgroundColor == UIColor.clearColor() and the numberField is shrunk it doesn't appear beneath and to the left of the brandImageView.

So I get:
screen shot 2015-08-27 at 16 59 18

Rather than:
screen shot 2015-08-27 at 17 04 07


jflinter:
This seems totally reasonable to me. Thanks, @sportlabsmike. @benzguo, mind
sanity-checking this on your machine and handling the merge?

benzguo:
:+1:

PR #229, opened 2015-09-15
Remove dependencies from example apps

jflinter:
They currently build cleanly on Xcode 6.4, but not 7


benzguo:
nice. :skull: :+1:

PR #239, opened 2015-09-21
fix xcode 7 warnings

benzguo:
r? @jflinter 

Noticed a few warnings on Xcode 7

screenshot 2015-09-21 10 55 40
screenshot 2015-09-21 10 55 51


Stripe.xcodeproj/project.pbxproj
```diff
@@ -1320,6 +1320,8 @@
 			isa = XCBuildConfiguration;
```
jflinter: Mind moving these to xcconfigs?

jflinter:
Beyond that :+1:

Stripe.xcodeproj/project.pbxproj
benzguo: :+1:

PR #247, opened 2015-10-12
Introduce STPCardParams

jflinter:
This splits `STPCard`, which was previously used for both creating requests to the Stripe API and serializing responses from it, into `STPCardParams` and `STPCard`, respectively. (Same for `STPBankAccount`.) For compatibility reasons, `STPCard` inherits from `STPCardParams` - this way existing code won't break, although it will trigger deprecation warnings.
r? @benzguo

### DIFF
--- a/Stripe/PublicHeaders/STPAPIResponseDecodable.h
+++ b/Stripe/PublicHeaders/STPAPIResponseDecodable.h
@@ -1,0 +1,15 @@
+//
+//  STPAPIResponseDecodable.h
+//  Stripe
+//
+//  Created by Jack Flintermann on 10/14/15.
+//  Copyright © 2015 Stripe, Inc. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+@protocol STPAPIResponseDecodable <NSObject>
+
++ (nullable instancetype)decodedObjectFromAPIResponse:(nonnull NSDictionary *)response;
+
+@end
--- a/Stripe/PublicHeaders/STPBankAccount.h
+++ b/Stripe/PublicHeaders/STPBankAccount.h
@@ -8,11 +8,12 @@
 
 #import <Foundation/Foundation.h>
 #import "STPBankAccountParams.h"
+#import "STPAPIResponseDecodable.h"
 
 /**
  *  Representation of a user's bank account details that have been tokenized with the Stripe API. @see https://stripe.com/docs/api#cards
  */
-@interface STPBankAccount : STPBankAccountParams
+@interface STPBankAccount : STPBankAccountParams<STPAPIResponseDecodable>
 
 /**
  *  The last 4 digits of the bank account's account number.
@@ -69,12 +70,5 @@
 #define DEPRECATED_IN_FAVOR_OF_STPBANKACCOUNTPARAMS __attribute__((deprecated("For collecting your users' bank account details, you should use an STPBankAccountParams object instead of an STPBankAccount.")))
 
 - (void)setAccountNumber:(nullable NSString *)accountNumber DEPRECATED_IN_FAVOR_OF_STPBANKACCOUNTPARAMS;
-
-@end
-
-// This method is used internally by Stripe to deserialize API responses and exposed here for convenience and testing purposes only. You should not use it in your own code.
-@interface STPBankAccount (PrivateMethods)
-
-- (nonnull instancetype)initWithAttributeDictionary:(nonnull NSDictionary *)attributeDictionary;
 
 @end
--- a/Stripe/PublicHeaders/STPBankAccount.h
+++ b/Stripe/PublicHeaders/STPBankAccount.h
@@ -7,19 +7,17 @@
 //
 
 #import <Foundation/Foundation.h>
-
-
-
-/**
- *  Representation of a user's credit card details. You can assemble these with information that your user enters and
- *  then create Stripe tokens with them using an STPAPIClient. @see https://stripe.com/docs/api#create_bank_account_token
- */
-@interface STPBankAccount : NSObject
+#import "STPBankAccountParams.h"
 
 /**
- *  The account number for the bank account. Currently must be a checking account.
+ *  Representation of a user's bank account details that have been tokenized with the Stripe API. @see https://stripe.com/docs/api#cards
  */
-@property (nonatomic, copy, nullable) NSString *accountNumber;
+@interface STPBankAccount : STPBankAccountParams
+
+/**
+ *  The last 4 digits of the bank account's account number, if it's been set, otherwise nil.
+ */
+- (nonnull NSString *)last4;
 
 /**
  *  The routing number for the bank account. This should be the ACH routing number, not the wire routing number.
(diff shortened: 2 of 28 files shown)


Stripe/PublicHeaders/STPBankAccount.h
```diff
@@ -7,19 +7,17 @@
...
- *  Representation of a user's credit card details. You can assemble these with information that your user enters and
- *  then create Stripe tokens with them using an STPAPIClient. @see https://stripe.com/docs/api#create_bank_account_token
+ *  Representation of a user's bank account details that have been tokenized with the Stripe API. @see https://stripe.com/docs/api#cards
  */
-@interface STPBankAccount : NSObject
+@interface STPBankAccount : STPBankAccountParams
 
 /**
- *  The account number for the bank account. Currently must be a checking account.
+ *  The last 4 digits of the bank account's account number, if it's been set, otherwise nil.
```
benzguo: "otherwise nil" contradicts the `nonnull`

Stripe/PublicHeaders/STPBankAccountParams.h
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ *  The account number for the bank account. Currently must be a checking account.
+ */
+@property (nonatomic, copy, nullable) NSString *accountNumber;
+
+/**
+ *  The last 4 digits of the bank account's account number, if it's been set, otherwise nil.
+ */
+- (nullable NSString *)last4;
```
benzguo: do we need the `last4` getter, if this object just gets sent to the API?

Stripe/PublicHeaders/STPBankAccountParams.h
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ *  The routing number for the bank account. This should be the ACH routing number, not the wire routing number.
+ */
+@property (nonatomic, copy, nullable) NSString *routingNumber;
+
+/**
+ *  The country the bank account is in.
```
benzguo: maybe specify that this is a country code / ISO-X ?

Stripe/PublicHeaders/STPBankAccountParams.h
```diff
@@ -0,0 +1,42 @@
...
+ */
+@property (nonatomic, copy, nullable) NSString *country;
+
+/**
+ *  The default currency for the bank account.
+ */
+@property (nonatomic, copy, nullable) NSString *currency;
```
benzguo: maybe specify that this is a currency code / ISO-X ?

Stripe/PublicHeaders/STPCardParams.h
```diff
@@ -0,0 +1,90 @@
...
+ */
+@property (nonatomic) NSUInteger expMonth;
+
+/**
+ *  The card's expiration year.
+ */
+@property (nonatomic) NSUInteger expYear;
+
+/**
+ *  The card's security code, found on the back. This will be nil for cards retrieved from the Stripe API.
```
benzguo: remove "This will be nil for cards retrieved from the Stripe API." ?

Stripe/PublicHeaders/STPCardParams.h
```diff
@@ -0,0 +1,90 @@
...
+ */
+@property (nonatomic, copy, nullable) NSString *addressLine1;
+@property (nonatomic, copy, nullable) NSString *addressLine2;
+@property (nonatomic, copy, nullable) NSString *addressCity;
+@property (nonatomic, copy, nullable) NSString *addressState;
+@property (nonatomic, copy, nullable) NSString *addressZip;
+@property (nonatomic, copy, nullable) NSString *addressCountry;
+
+/**
+ *  This is only applicable when tokenizing debit cards to issue payouts to managed accounts. You should not set it otherwise. The card can then be used as a transfer destination for funds in this currency.
```
benzguo: specify that this is a currency code?

Stripe/PublicHeaders/UI/STPPaymentCardTextField.h
```diff
@@ -157,7 +157,7 @@
 /**
  *  Convenience method to create a STPCard from the currently entered information. Will return nil if not valid.
```
benzguo: nit: kinda confusing docstring, this isn't a method

jflinter:
Mind re-reviewing, @benzguo?
I added a lot more code to this PR, but the card -> cardparams stuff is unchanged. I added two new protocols, `STPFormEncodable` and `STPAPIResponseDecodable`, which basically are the ~same as `StripeAPITranslatable`.

Stripe/PublicHeaders/STPAPIResponseDecodable.h
```diff
@@ -0,0 +1,15 @@
...
+//
+//  Created by Jack Flintermann on 10/14/15.
+//  Copyright © 2015 Stripe, Inc. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+@protocol STPAPIResponseDecodable <NSObject>
+
++ (nullable instancetype)decodedObjectFromAPIResponse:(nonnull NSDictionary *)response;
```
benzguo: nice.

Stripe/PublicHeaders/STPBankAccountParams.h
benzguo: (still not sure why we need to expose this in the params object)

Stripe/PublicHeaders/STPBankAccountParams.h
```diff
@@ -0,0 +1,43 @@
...
+ */
+@property (nonatomic, copy, nullable) NSString *routingNumber;
+
+/**
+ *  Two-letter ISO code representing the country the bank account is located in.
+ */
+@property (nonatomic, copy, nullable) NSString *country;
+
+/**
+ *  The default currency for the bank account.
```
benzguo: nit: "Three-letter ISO currency code representing the default currency for the bank account."

Stripe/PublicHeaders/STPFormEncodable.h
```diff
@@ -0,0 +1,16 @@
...
+//  Created by Jack Flintermann on 10/14/15.
+//  Copyright © 2015 Stripe, Inc. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+@protocol STPFormEncodable <NSObject>
+
+- (NSString *)rootObjectName;
+- (NSDictionary *)propertyNamesToFormFieldNamesMapping;
```
benzguo: (no need to change) 
I've taken to just calling dictionaries `keyToValue`, e.g. `propertyNameToFormFieldName`

benzguo:
looks good! 
my only question may just be me missing something – why are we exposing `last4` in the params objects?

jflinter:
Backwards compatibility - that code existed with that behavior on
STPCard/STPBankAccount, and I don't want to break things (even if the
behavior is a little silly). The true breaking change will be to make
STPCard not inherit from STPCardParams, at which point we'll remove this.

benzguo:
ah, makes sense. :ship:

PR #248, opened 2015-10-13
NSURLConnection -> NSURLSession in STPAPIClient.m

jflinter:
r? @benzguo


Stripe/PublicHeaders/STPAPIClient.h
```diff
@@ -55,7 +55,7 @@ typedef void (^STPCompletionBlock)(STPToken * __nullable token, NSError * __null
 @property (nonatomic, copy, nullable) NSString *publishableKey;
 
 /**
- *  The operation queue on which to run the url connection and delegate methods. Cannot be nil. @see NSURLConnection
+ *  The operation queue on which to run completion blocks passed to the api client. Cannot be nil.
```
benzguo: nit: can probably remove "Cannot be nil" now that there's a `nonnull` annotation

benzguo: maybe note that this is the main queue by default?

Stripe/STPAPIClient.m
```diff
@@ -80,40 +87,43 @@ - (void)createTokenWithData:(NSData *)data completion:(STPCompletionBlock)comple
...
-                                 }
-                             }
-                         }];
+    [[self.urlSession dataTaskWithRequest:request completionHandler:^(NSData * _Nullable body, NSURLResponse * _Nullable response, NSError * _Nullable error) {
+        if (error) {
+            // If this is an error that Stripe returned, let's handle it as a StripeDomain error
+            if (body) {
+                NSDictionary *jsonDictionary = [NSJSONSerialization JSONObjectWithData:body options:0 error:NULL];
+                if ([jsonDictionary valueForKey:@"error"] != nil) {
+                    [self.operationQueue addOperationWithBlock:^{
```
benzguo: this seems odd – we initialize the session with `operationQueue` as its `delegateQueue`, so this is just putting the completion block on the same queue.

jflinter: I agree it's gross. It's because the `operationQueue` can be set after the fact, so if you initialize an API client and then change its operation queue, it won't update the underlying url session (since that property can only be set during initialization on NSURLSession). Really we should probably just deprecate this property.

PR #251, opened 2015-10-15
Add Isracards brand

kwysockibright:


jflinter:
Hey there, thanks for opening this up. I think the combination of Stripe not currently supporting Isracard, as well as the large amount of special-casing required here, means that we're going to pass on merging this for now.

PR #270, opened 2015-11-16
Update to recommended settings

benzguo:
r? @jflinter


jflinter:
This isn't a very... reviewable change?

benzguo:
true. I usually err on the side of making PRs just so changes are more visible, but I'll just push things like this in the future.

jflinter:
Yeah, I am pro-PR here! Just doesn't have to block on my reviewing it : )

benzguo:
ah, cool. :)

PR #271, opened 2015-11-19
Configure brand images

klaaspieter:
Useful for apps, like mine, that need to customize the brand images used by the card text field.

### DIFF
--- a/Stripe/PublicHeaders/UI/STPPaymentCardTextField.h
+++ b/Stripe/PublicHeaders/UI/STPPaymentCardTextField.h
@@ -128,6 +128,9 @@
  */
 - (void)clear;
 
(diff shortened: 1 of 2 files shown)


jflinter:
Ah, neat. Thanks Klaas!

bradfol:
@klaaspieter This implementation seems broken. The `cardBrand` parameter is just ignored in favor of the viewModel's state, which is computed based on the input in the text fields. 

Is this intended as a hook for override? If so, it seems the hook is implemented in the wrong place.

klaaspieter:
@bradfol The methods were indeed meant to be just hooks that subclasses can override. They return the viewModel state because that was the behavior before my PR, in other words it didn't break any existing implementations.

bradfol:
@klaaspieter I have submitted #284 that will allow for a similar hook (it is now a class method) while providing correct implementations of the new methods.

PR #272, opened 2015-11-19
Allow subclasses to override layout rectangles

klaaspieter:
This makes it easier for subclasses to alter how the text field lays out it's different subviews. Or to use the exposed information to add other subviews alongside the already existing views.

It's also a common pattern for `UIControl` subclasses. For example:

- [UIButton](https://developer.apple.com/library/ios/documentation/UIKit/Reference/UIButton_Class/#//apple_ref/occ/instm/UIButton/backgroundRectForBounds:)
- [UITextField](https://developer.apple.com/library/prerelease/tvos/documentation/UIKit/Reference/UITextField_Class/index.html#//apple_ref/occ/instm/UITextField/textRectForBounds:)

### DIFF
--- a/Stripe/PublicHeaders/UI/STPPaymentCardTextField.h
+++ b/Stripe/PublicHeaders/UI/STPPaymentCardTextField.h
@@ -108,6 +108,11 @@
 @property(nonatomic, strong, nullable) UIView *inputAccessoryView;
 
 /**
+ *  The curent brand image displayed in the receiver.
+ */
+@property (nonatomic, readonly, nullable) UIImage *brandImage;
+
+/**
  *  Causes the text field to begin editing. Presents the keyboard.
  *
  *  @return Whether or not the text field successfully began editing.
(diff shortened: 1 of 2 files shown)


Stripe/PublicHeaders/UI/STPPaymentCardTextField.h
```diff
@@ -108,6 +108,11 @@
 @property(nonatomic, strong, nullable) UIView *inputAccessoryView;
 
 /**
+ *  The curent brand image displayed in the receiver.
+ */
+@property (nonatomic, readonly, nullable) UIImage *brandImage;
```
klaaspieter: This is exposed so that subclassers can use the image for their custom `brandImageRectForBounds:` logic.

klaaspieter:
As an example of why this is useful. I was able to build this subclass using these methods:

image

bangaja7:
Ok

jflinter:
Thanks for the PR, @klaaspieter - I think this seems pretty reasonable (my fear is always that the public API for `STPPaymentCardTextField` will get too big and make it hard to change later on, but it's reasonable to assume that we'll always want to lay out those text fields somehow :smile: ). Would you mind rebasing?

ObjColumnist:
It would also be useful if the padding was exposed through the API e.g. `contentEdgeInsets` so that it can easily be lined up with other UI Elements. The same could be achieved with this API I'm sure, but it would require overriding all the layout methods to achieve it.

klaaspieter:
@jflinter Done.

Stripe/UI/STPPaymentCardTextField.m
```diff
@@ -620,6 +653,7 @@ - (void)updateImageForFieldType:(STPCardFieldType)fieldType {
         
         [self.brandImageView.layer addAnimation:transition forKey:nil];
     }
+    [self setNeedsLayout];
```
jflinter: This can probably be moved into the `if (image != self.brandImageView.image) {` check above to avoid the occasional unnecessary layout.

jflinter:
Thanks @klaaspieter! Hope all is well!

jflinter:
@ObjColumnist - a `contentEdgeInsets` property isn't crazy; I'll look into it.

PR #275, opened 2015-12-01
[ApplePay] Take name and address from shippingAddress if present

davidrothera:
At present if a PKPayment object contains a shippingAddress but no billingAddress then no name/address details are extracted and passed along to Stripe.

This means that when looking on the Stripe dashboard at ApplePay charges with no billingAddress that the name/address fields are empty which is still valid.

This diff changes the behaviour so that:
* If billingAddress and shippingAddress exist then we take the billingAddress
* If billingAddress is set and shippingAddress is not set then we take the billingAddress
* If shippingAddress is set and billingAddress is not set then we take the shippingAddress
* If neither are set then we don't do anything

I tested this and checked the generated URL (from the `formEncodedDataForPayment` method) and checked that only one copy of the name/address data was present when both shipping & billing were set.

### DIFF
--- a/Stripe/ApplePay/STPAPIClient+ApplePay.m
+++ b/Stripe/ApplePay/STPAPIClient+ApplePay.m
@@ -17,6 +17,52 @@
     [self createTokenWithData:[self.class formEncodedDataForPayment:payment] completion:completion];
 }
 
++ (NSString * _Nonnull)queryStringForAddress:(ABRecordRef _Nullable)addressRef characterSet:(NSCharacterSet * _Nonnull)set {
+    NSMutableDictionary *params = [NSMutableDictionary dictionary];
+    __block NSString *queryString = [NSString string];
+
+    NSString *firstName = (__bridge_transfer NSString*)ABRecordCopyValue(addressRef, kABPersonFirstNameProperty);
+    NSString *lastName = (__bridge_transfer NSString*)ABRecordCopyValue(addressRef, kABPersonLastNameProperty);
+    if (firstName.length && lastName.length) {
+        params[@"name"] = [NSString stringWithFormat:@"%@ %@", firstName, lastName];
+    }
+
+    ABMultiValueRef addressValues = ABRecordCopyValue(addressRef, kABPersonAddressProperty);
+    if (addressValues != NULL) {
+        if (ABMultiValueGetCount(addressValues) > 0) {
+            CFDictionaryRef dict = ABMultiValueCopyValueAtIndex(addressValues, 0);
+            NSString *line1 = CFDictionaryGetValue(dict, kABPersonAddressStreetKey);
+            if (line1) {
+                params[@"address_line1"] = line1;
+            }
+            NSString *city = CFDictionaryGetValue(dict, kABPersonAddressCityKey);
+            if (city) {
+                params[@"address_city"] = city;
+            }
+            NSString *state = CFDictionaryGetValue(dict, kABPersonAddressStateKey);
+            if (state) {
+                params[@"address_state"] = state;
+            }
+            NSString *zip = CFDictionaryGetValue(dict, kABPersonAddressZIPKey);
+            if (zip) {
+                params[@"address_zip"] = zip;
+            }
+            NSString *country = CFDictionaryGetValue(dict, kABPersonAddressCountryKey);
+            if (country) {
+                params[@"address_country"] = country;
+            }
+            CFRelease(dict);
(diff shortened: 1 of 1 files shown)


Stripe/ApplePay/STPAPIClient+ApplePay.m
```diff
@@ -68,6 +68,92 @@ + (NSData *)formEncodedDataForPayment:(PKPayment *)payment {
         }
     }
 
+    if (payment.shippingAddress && payment.billingAddress == nil) {
+        NSMutableDictionary *params = [NSMutableDictionary dictionary];
```
benzguo: There's a lot of repeated code here. It'd be better to extract this into a function:

```objc
- (NSString *)queryStringWithAddress:(ABRecord *)record {
...
}
```

davidrothera:
Agreed @benzguo, I have changed the code now to re-use a string generator method which trims down the code a little.

Stripe/PublicHeaders/ApplePay/STPAPIClient+ApplePay.h
```diff
@@ -21,6 +21,9 @@
  */
 - (void)createTokenWithPayment:(nonnull PKPayment *)payment completion:(nonnull STPTokenCompletionBlock)completion;
 
+// Encodes the data from an ABRecordRef (name/address) into a string. This method is used internally by STPAPIClient; you should not use it in your own code.
```
benzguo: No need to expose this in the header (I believe `formEncodedDataForPayment` was exposed for testing).

Stripe/ApplePay/STPAPIClient+ApplePay.m
```diff
@@ -17,6 +17,55 @@ - (void)createTokenWithPayment:(PKPayment *)payment completion:(STPTokenCompleti
     [self createTokenWithData:[self.class formEncodedDataForPayment:payment] completion:completion];
 }
 
++ (NSString * _Nonnull)queryStringForAddress:(ABRecordRef _Nullable)addressRef {
+    NSMutableDictionary *params = [NSMutableDictionary dictionary];
+    __block NSString *queryString = [NSString string];
+
+    NSMutableCharacterSet *set = [[NSCharacterSet URLQueryAllowedCharacterSet] mutableCopy];
```
benzguo: nitpick, but we're creating this character set twice. can we add an `allowedCharacters:(NSCharacterSet *)` param?

davidrothera:
* Updated to re-use existing NSCharacterSet
* Removed method declaration from public header

Stripe/ApplePay/STPAPIClient+ApplePay.m
```diff
@@ -25,47 +71,19 @@ + (NSData *)formEncodedDataForPayment:(PKPayment *)payment {
...
-                }
-                CFRelease(dict);
-                [params enumerateKeysAndObjectsUsingBlock:^(NSString *key, NSString *obj, __unused BOOL *stop) {
-                    NSString *param = [NSString stringWithFormat:@"&card[%@]=%@", key, [obj stringByAddingPercentEncodingWithAllowedCharacters:set]];
-                    payloadString = [payloadString stringByAppendingString:param];
-                }];
-            }
-            CFRelease(addressValues);
-        }
+    if (payment.billingAddress && payment.shippingAddress == nil) {
```
benzguo: sorry, one more nit :) 
you can consolidate this into two cases:
```objc
if (payment.shippingAddress) {
    NSString *addressString = [self queryStringForAddress:payment.shippingAddress characterSet:set];
    payloadString = [payloadString stringByAppendingString:addressString];
}
else if (payment.billingAddress) {
    NSString *addressString = [self queryStringForAddress:payment.billingAddress characterSet:set];
    payloadString = [payloadString stringByAppendingString:addressString];
}
```
thanks for catching and fixing this!

davidrothera: @benzguo my only reason for keeping all 3 options was so that if both the `shippingAddress` and the `billingAddress` were set then we would pick the `billingAddress`, doing this keeps with the same behaviour that is seen at present.

benzguo: whoops, just noticed that i'd swapped `billingAddress` and `shippingAddress` above (fixed now). 

the `if`-`else if` above is equivalent to your three `if`s. if both `billingAddress` and `shippingAddress` are non-nil, the first branch will be used – no need for a separate case.

davidrothera: Got it, amending now :+1:

benzguo:
:tada: thanks!

PR #276, opened 2015-12-03
Revert "[ApplePay] Take name and address from shippingAddress if present"

benzguo:
Reverts stripe/stripe-ios#275

@davidrothera – after further consideration, we've decided to revert this commit. `shippingAddress` != `billingAddress`, and falling back to `shippingAddress` could cause confusion.

Note that if you want to store the user's `shippingAddress`, you can pass the `/charges` endpoint an optional `shipping` hash. https://stripe.com/docs/api#create_charge-shipping

### DIFF
--- a/Stripe/ApplePay/STPAPIClient+ApplePay.m
+++ b/Stripe/ApplePay/STPAPIClient+ApplePay.m
@@ -17,52 +17,6 @@
     [self createTokenWithData:[self.class formEncodedDataForPayment:payment] completion:completion];
 }
 
-+ (NSString * _Nonnull)queryStringForAddress:(ABRecordRef _Nullable)addressRef characterSet:(NSCharacterSet * _Nonnull)set {
-    NSMutableDictionary *params = [NSMutableDictionary dictionary];
-    __block NSString *queryString = [NSString string];
-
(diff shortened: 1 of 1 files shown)


davidrothera:
@benzguo Does passing a `shipping` hash to the `/charges` endpoint then show the address in the portal, this change was mainly so that when I create an Apple Pay transaction (with only a shippingAddress and not a billingAddress) there is no name/address shown.

benzguo:
It won't appear in the dashboard, but it will appear in the charge object returned from the API.

davidrothera:
Ok, that is what this and the support ticket that I raised was for, in my normal flow (non Apple Pay) I ask for only the shipping address as we do not ship to addresses that are not the billing address.

Adding Apple Pay and with the way that the billingAddress and shippingAddress are handled means that either I have to submit no billingAddress and have the dashboard empty or ask the user for both in the payment view controller.

Do you think it would be acceptable to implement an option on the Stripe payment method to use the same functionality as in the reverted diff, something like a `shippingAddressIsBillingAddress` boolean?

benzguo:
We'd prefer that you ask for both the shipping address and billing address (even if you expect them to be the same), as it's more explicit.

davidrothera:
Agreed, thanks for the input @benzguo :+1:

PR #282, opened 2015-12-12
Implement didBeginEditing delegate method

bradfol:
Forwards the `didBeginEditing` delegate calls from the textField subviews.


benzguo:
@bradfol Sorry for the late response – [this](https://github.com/stripe/stripe-ios/pull/289) should provide similar functionality, and will be in the next release (coming soon!)

bradfol:
@benzguo thanks!

PR #288, opened 2016-01-04
Expose card icons through UIImage categories

benzguo:
This moves `brandImageForCardBrand`, `cvcImageForCardBrand`, and `safeImageNamed` to `UIImage+Stripe`.


bradfol:
Hi @benzguo would this still allow for a subclass of `STPPaymentCardTextField` to override the images returned for each brand? PR #271 initially introduced similar methods to be used as hooks to allow for custom brand images.

jflinter:
I think I did a bad job writing this up - all we really want here is an easier way for people to access the various card icons. In other words, the UIImage category would just expose methods like `stp_visaCardImage`, etc, so that people can avoid calling `imageNamed` directly and we can safely rename images, etc in the future without breaking implementations.

There is also the issue that the SDK already has a category named UIImage+Stripe.

benzguo:
ah, whoops, my bad – closing.

PR #290, opened 2016-01-05
Add setCard to STPPaymentCardTextField

### LINKED ISSUE #219: Allow setting STPPaymentCardTextField values externally
Setting card number, expiration, and cvc for STPPaymentCardTextField externally would be extremely useful. For example, when scanning a card with the camera there is no way to populate a STPPaymentCardTextField with the captured values.

benzguo:
Fixes #219 

8957a9b adds some testing UI to `PaymentViewController` – planning on removing the commit when this is good to go.

### DIFF
--- a/Stripe/PublicHeaders/UI/STPPaymentCardTextField.h
+++ b/Stripe/PublicHeaders/UI/STPPaymentCardTextField.h
@@ -231,14 +231,11 @@
 @property(nonatomic, readonly, nullable) NSString *cvc;
 
 /**
- *  Convenience method to create a STPCard from the currently entered information. Will return nil if not valid.
+ *  Convenience property for creating an STPCard from the currently entered information
+ *  or programmatically setting the field's contents.
+ *  Nil if the current information is not a valid card.
  */
-@property(nonatomic, readonly, nullable) STPCardParams *card;
-
-/**
- *  Configures the field with the given card details.
- */
-- (void)setCard:(nonnull STPCardParams *)card;
+@property(nonatomic, readwrite, nullable) STPCardParams *card;
 
 @end
 
@@ -264,7 +261,7 @@ __attribute__((deprecated("This protocol is provided only for backwards-compatib
 __attribute__((deprecated("This class is provided only for backwards-compatibility with PaymentKit. You shouldn't use it - use STPPaymentCardTextField instead.")))
 @interface PTKView : STPPaymentCardTextField
 @property(nonatomic, weak, nullable)id<PTKViewDelegate>delegate;
-@property(nonatomic, readonly, nonnull) PTKCard *card;
+@property(nonatomic, readwrite, nonnull) PTKCard *card;
 @end
 
 #pragma clang diagnostic pop
--- a/Stripe/PublicHeaders/UI/STPPaymentCardTextField.h
+++ b/Stripe/PublicHeaders/UI/STPPaymentCardTextField.h
@@ -231,11 +231,14 @@
 @property(nonatomic, readonly, nullable) NSString *cvc;
 
 /**
- *  Convenience property for creating an STPCard from the currently entered information
- *  or programmatically setting the field's contents.
- *  Nil if the current information is not a valid card.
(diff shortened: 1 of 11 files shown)


Stripe/PublicHeaders/UI/STPPaymentCardTextField.h
```diff
@@ -220,6 +220,11 @@
  */
 @property(nonatomic, readonly, nullable) STPCardParams *card;
```
jflinter: This should probably just update the property to be readwrite, rather than defining a separate method.

Stripe/UI/STPPaymentCardTextField.m
```diff
@@ -403,6 +403,55 @@ - (STPCardParams *)card {
     return c;
 }
 
+- (void)setCard:(nonnull STPCardParams *)card {
```
jflinter: This is kind of a lot of hard-to-reason-about code. Could you add some tests?

Stripe/PublicHeaders/UI/STPPaymentCardTextField.h
benzguo: ah, yup.

benzguo: actually, forgot there was a reason I did this – we probably don't want to change the `card` on the legacy `PTKView : STPPaymentCardTextField` interface to `readwrite`, and we can't update this property without also updating that subclass.

Stripe/UI/STPPaymentCardTextField.m
benzguo: :+1: added tests in https://github.com/stripe/stripe-ios/commit/003d702e001e99b8ccac0b0a7c501c81b8e97a9a

benzguo:
Note: I've removed the testing UI from this branch. It's still available on https://github.com/stripe/stripe-ios/tree/bg-setcardparams-testing if you want to play around.

Stripe/UI/STPPaymentCardTextField.m
benzguo: Note that these tests don't cover the cases when the field is first responder – we'd need to add an application test target. (Happy to do that if you think we should).

Stripe/PublicHeaders/UI/STPPaymentCardTextField.h
jflinter: I think it's fine to do that; that thing is deprecated anyway, and this would be an additive change.

Stripe/UI/STPPaymentCardTextField.m
```diff
@@ -403,6 +403,55 @@ - (STPCardParams *)card {
...
+    if (card.number.length) {
```
jflinter: Won't this break if you pass e.g. a card with an empty number to a field where the number is already set? (same with expiration and cvc below.)

jflinter: Like, ideally you could reset some of these values by passing an empty card object.

Stripe/UI/STPPaymentCardTextField.m
jflinter: What do you think is best?

benzguo: Moar tests! I'll add a new test target.

Stripe/PublicHeaders/UI/STPPaymentCardTextField.h
benzguo: :+1:

Stripe/UI/STPPaymentCardTextField.m
benzguo: Ah, good point – I'll update.

benzguo:
re-r? @jflinter 

updates:
- added an application test target (hosted by the Simple example), and added the new tests to CI.
- `card` is now a `readwrite` property
- setting a `card` with empty fields now behaves correctly

OhDavit:
why initially is implemented as a readonly?

PR #292, opened 2016-01-11
Expose card icons through UIImage category

benzguo:
r? @jflinter 

replaces https://github.com/stripe/stripe-ios/pull/288

It doesn't look like the SDK already has a UIImage category?

### DIFF
--- a/Stripe/UI/UIImage+Stripe.m
+++ b/Stripe/UI/UIImage+Stripe.m
@@ -1,0 +1,44 @@
+//
+//  UIImage+Stripe.m
+//  Stripe
+//
+//  Created by Ben Guo on 1/4/16.
+//  Copyright © 2016 Stripe, Inc. All rights reserved.
+//
+
+#import "UIImage+Stripe.h"
+#import "STPPaymentCardTextFieldViewModel.h"
+
+@implementation UIImage (Stripe)
+
++ (UIImage *)stp_amexCardImage {
+    return [STPPaymentCardTextFieldViewModel brandImageForCardBrand:STPCardBrandAmex];
+}
+
++ (UIImage *)stp_dinersClubCardImage {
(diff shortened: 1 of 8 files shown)


Stripe/UI/UIImage+Stripe.m
```diff
@@ -0,0 +1,44 @@
...
+//  Created by Ben Guo on 1/4/16.
+//  Copyright © 2016 Stripe, Inc. All rights reserved.
+//
+
+#import "UIImage+Stripe.h"
+#import "STPPaymentCardTextFieldViewModel.h"
+
+@implementation UIImage (Stripe)
+
++ (UIImage *)stp_amexCardImage {
```
jflinter: I think it'd be better to switch this around - in other words, these methods should just be calling `imageNamed:`, and then STPPaymentCardTextFieldViewModel would depend on this category, not the other way around.

jflinter: Then you can also move `safeImageNamed` to this category as well.

jflinter:
My mistake; there's no existing UIImage category.

jflinter:
It'd also be nice to have a really quick unit test case for this category that just asserts that the output of all of those methods is non-nil.

Stripe/UI/UIImage+Stripe.m
benzguo: ah, ok – I misread your "all we really want here" and reduced the scope of the changes. This is actually pretty close to what I had in https://github.com/stripe/stripe-ios/pull/288 .

benzguo:
> It'd also be nice to have a really quick unit test case for this category that just asserts that the output of all of those methods is non-nil.

:+1: , added tests

jflinter:
LGTM, sorry I missed this!

PR #293, opened 2016-01-11
Ignore iOS9 deprecation warnings

benzguo:
r? @jflinter 

I've also updated the deployment target of the cocoapods `without_frameworks` project to 9.2, and added a post_install hook to treat warnings as errors in the pods project.


jflinter:
\o/ lgtm

PR #295, opened 2016-01-15
Adding a zipcode field [do not merge]

seansu4you87:
Should function the same way the other fields do, automatically
selecting the next field / previous field based on editing


seansu4you87:
Oh whoops, I wasn't ready PR this into the original repo, my bad.

PR #300, opened 2016-01-31
Fixing document warning

glentregoning:
In Xcode 7.2 this incorrect parameter name was causing a warning.


benzguo:
:+1: thanks!

PR #302, opened 2016-02-01
Move UIImage+Stripe.h to public headers

### LINKED ISSUE #301: Unable to import "UIImage+Stripe.h"
Other categories work, but `UIImage+Stripe.h` does not.
Looks like the reason is it is not  in PublicHeaders folder.

benzguo:
r? @jflinter 

Fixes #301 :/

I've also added an `#import <Stripe/UIImage+Stripe.h>` to the cocoapods test project.


jflinter:
LGTM. Do you have any Feelings on how we could catch this automatically next time?

benzguo:
I've imported the header in the cocoapods test project – perhaps we should add imports for all public headers?

jflinter:
That seems like a reasonable first step.

benzguo:
en-phabbed: https://phab.stripe.com/T18617

PR #304, opened 2016-02-05
Add additionalFields to STPFormEncodable/STPAPIResponseDecodable

jflinter:
This adds the ability to specify custom fields in API requests and responses, to support beta fields that haven't been added to the SDK yet.


benzguo:
LGTM!

PR #327, opened 2016-04-21
Discover Support

andrewpthorp:
Stripe now supports Discover in the US. Enable it by default.

r? @jflinter

### DIFF
--- a/FauxPasConfig/main.fauxpas.json
+++ b/FauxPasConfig/main.fauxpas.json
@@ -1,11 +1,11 @@
 // Generated from the Faux Pas GUI
 {
     // Xcode target to check (String)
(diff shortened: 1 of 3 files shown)


jflinter:
I believe the `PKPaymentNetworkDiscover` constant is only available as of iOS 9.0, so you'll unfortunately need to only add it to the array of supported networks if that constant is available at runtime or this will crash on pre-iOS9 devices. I'm actually a little worried that this didn't cause the CI to break...

jflinter:
http://stackoverflow.com/questions/3122177/check-if-constant-is-defined-at-runtime-in-obj-c may be of interest here.

andrewpthorp:
Thanks!

PR #328, opened 2016-04-22
add STPAnalyticsClient

benzguo:
r? @jflinter 

At the moment, I'm leaving out the `resourceTiming` field completely. We could potentially track `responseStart` with `URLSession:dataTask:didReceiveResponse:`, but I don't think it'd be worth the added complexity.

😞 
https://github.com/WebKit/webkit/blob/master/Source/WebCore/platform/network/cocoa/ResourceLoadTiming.mm


Stripe/STPAnalyticsClient.m
```diff
@@ -0,0 +1,68 @@
...
+                      start:(NSDate *)startTime
+                        end:(NSDate *)endTime {
+    if (![[self class] shouldCollectAnalytics]) {
+        return;
+    }
+    if (![response isKindOfClass:[NSHTTPURLResponse class]]) {
+        return;
+    }
+    NSHTTPURLResponse *httpResponse = (NSHTTPURLResponse *)response;
+    NSNumber *start = @((NSInteger)[startTime timeIntervalSince1970]);
```
raybejjani: The browser-side version of this is reporting milliseconds. Any chance you could * 1000 before the integer cast?

raybejjani:
Apart from the 1 comment with the timestamps, +1!

Stripe/STPAnalyticsClient.m
benzguo: ah, good catch – thanks!

raybejjani:
lgtm

Stripe/STPAnalyticsClient.m
```diff
@@ -0,0 +1,102 @@
...
+
++ (NSNumber *)timestampWithDate:(NSDate *)date {
+    return @((NSInteger)([date timeIntervalSince1970]*1000));
+}
+
+- (instancetype)init {
+    self = [super init];
+    if (self) {
+        NSURLSessionConfiguration *config = [NSURLSessionConfiguration defaultSessionConfiguration];
+        config.HTTPAdditionalHeaders = @{
```
jflinter: I don't know if it makes sense to still lurk this into a "user agent that isn't actually a user agent" header - can you just put these in the payload instead?

Stripe/ApplePay/STPAPIClient+ApplePay.m
```diff
@@ -16,7 +16,9 @@
 @implementation STPAPIClient (ApplePay)
 
 - (void)createTokenWithPayment:(PKPayment *)payment completion:(STPTokenCompletionBlock)completion {
-    [self createTokenWithData:[self.class formEncodedDataForPayment:payment] completion:completion];
+    [self createTokenWithData:[self.class formEncodedDataForPayment:payment]
+                    tokenType:@"apple_pay"
```
jflinter: Can these tokenTypes be moved to an enum instead?

Stripe/NSMutableURLRequest+Stripe.h
```diff
@@ -0,0 +1,15 @@
+//
```
jflinter: Make sure to load any categories like this in STPCategoryLoader to avoid messing up static builds & fabric.

Stripe/STPAPIClient.m
```diff
@@ -36,6 +37,7 @@
 static NSString *const tokenEndpoint = @"tokens";
 static NSString *const stripeAPIVersion = @"2015-10-12";
 static NSString *STPDefaultPublishableKey;
+static BOOL STPShouldCollectAnalytics = YES;
```
jflinter: It feels like this should live in STPAnalyticsClient.

Stripe/STPAPIPostRequest.h
```diff
@@ -12,7 +12,7 @@
 
 @interface STPAPIPostRequest<__covariant ResponseType:id<STPAPIResponseDecodable>> : NSObject
 
-typedef void(^STPAPIPostResponseBlock)(ResponseType object, NSError *error);
+typedef void(^STPAPIPostResponseBlock)(ResponseType object, NSURLResponse *response, NSError *error);
```
jflinter: Can you just make this be an NSHTTPURLResponse, cast it internally, and avoid having to cast it elsewhere?

PR #339, opened 2016-05-24
Add STPBINRange to fix 13-digit visa error

jflinter:


benzguo:
Nice, this seems much more robust. lgtm!

Issue #349, opened 2016-06-22
Add more information and get card brand, swift 


taktikal17:
Hi, 
I follwed the swift documentation : 
I would like to add more information for the payment ( name, country, address) :
 
I tried simply like it :
 paymentTextField.cardParams.addressCity = "San Francisco"
 paymentTextField.cardParams.name = "Val"
 paymentTextField.cardParams.addressCountry = "US"

But the elements in the Response body of the Dashboard are still equal to null..

How can I do it?

Secondly, I would like to get the brand of the card when the information are sending.
How can I get it?

Thank you!!!

taktikal17:
Thanks @benzguo !!

PR #350, opened 2016-06-23
Fix missing cardParam address fields 

benzguo:
r? @jflinter 

Addresses #349 

We were previously discarding address fields when setting `STPPaymentCardTextField`'s `cardParams`.

Note that this doesn't fully support the usage pattern in #349 – we'd have to do some more work (i.e. KVO) to support chaining.

```swift
paymentTextField.cardParams.addressCity = "San Francisco"
paymentTextField.cardParams.name = "Val"
paymentTextField.cardParams.addressCountry = "US"
```

### DIFF
--- a/Stripe/STPPaymentCardTextField.m
+++ b/Stripe/STPPaymentCardTextField.m
@@ -36,6 +36,8 @@
 
 @property(nonatomic, assign)BOOL numberFieldShrunk;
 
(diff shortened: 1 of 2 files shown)


taktikal17:
Thank you a lot @benzguo and @jflinter !!

Should it work now?

Have a good one!

PR #358, opened 2016-07-01
Fix sms code entry using certain custom keyboards

benzguo:
r? @jflinter 

tl;dr, When the delete key is pressed, some custom keyboards will call `textDocumentProxy.insertText("")` if the field is empty, rather than `textDocumentProxy.deleteBackward()`. This seems entirely at the discretion of the developer, so we should support both possibilities.

I tested entering SMS codes with a few custom keyboards:
- Using the system number pad or Swype, when the delete key is pressed, `deleteBackward` is called.
- Using Gboard or SwiftKey, `deleteBackward` is not called – instead, `shouldChangeCharactersInRange:` is called with an empty string.

A more heavy-handed solution would be to use `UIKeyboardTypePhonePad`, but that doesn't seem ideal.

> Your custom keyboard is also ineligible to type into so-called phone pad objects, such as the phone number fields in Contacts. These input objects are exclusively for strings built from a small set of alphanumeric characters specified by telecommunications carriers and are identified by having one or another of the following two keyboard type traits: UIKeyboardTypePhonePad, UIKeyboardTypeNamePhonePad

https://developer.apple.com/library/prerelease/content/documentation/General/Conceptual/ExtensibilityPG/CustomKeyboard.html

### DIFF
--- a/Stripe/STPAddressFieldTableViewCell.m
+++ b/Stripe/STPAddressFieldTableViewCell.m
@@ -105,7 +105,7 @@
                 }
                 
                 self.textField.placeholder = NSLocalizedString(@"12345", nil);
-                self.textField.keyboardType = UIKeyboardTypeNumberPad;
+                self.textField.keyboardType = UIKeyboardTypePhonePad;
                 self.textField.preservesContentsOnPaste = NO;
                 self.textField.selectionEnabled = NO;
                 if (!lastInList) {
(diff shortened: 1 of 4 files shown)


jflinter:
Hey @benzguo sorry it took me so long to get to this.

Honestly, I think we should just use `UIKeyboardTypePhonePad` here (and on `STPPaymentCardTextField`, etc). I'm getting worried about the amount of magic keyboard logic we're maintaining. The only difference is the "+*#" button in the lower-left, and we already ignore those characters if the user taps them (which they won't). Thoughts?

benzguo:
That sounds reasonable to me – I've updated this PR to replace all occurrences of `UIKeyboardTypeNumberPad` with `UIKeyboardTypePhonePad`.

jflinter:
thanks, lgtm in that case!

PR #363, opened 2016-07-05
Example app improvements

benzguo:
r? @jflinter 

- In-memory card persistence now works
- I've updated `MyAPIClient` to return an error if `completeCharge` is called and `customerID` or `baseURL` aren't set, rather than fake-succeeding.

- Unrelated to the example app: I've tweaked the `STPAddCardViewControllerDelegate` documentation – we advise calling `completion(nil)` if no error occurs, but that actually isn't necessary. [0]

[0] https://github.com/stripe/stripe-ios/blob/master/Stripe/STPAddCardViewController.m#L328-L330

### DIFF
--- a/Example/Stripe iOS Example (Simple)/CheckoutViewController.swift
+++ b/Example/Stripe iOS Example (Simple)/CheckoutViewController.swift
@@ -71,8 +71,9 @@ class CheckoutViewController: UIViewController, STPPaymentContextDelegate {
         self.totalRow = CheckoutRowView(title: "Total", detail: "", tappable: false,
                                         theme: settings.theme)
         self.buyButton = BuyButton(enabled: true, theme: settings.theme)
-        MyAPIClient.sharedInit(baseURL: self.backendBaseURL, customerID: self.customerID)
-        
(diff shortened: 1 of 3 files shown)


jflinter:
Might it be easier to just make a `MyAPIClient` singleton?

benzguo:
We are using a `MyAPIClient` singleton, but `sharedInit` was previously always overwriting the singleton instance. 

```
static func sharedInit(baseURL baseURL: String?, customerID: String?) {
    sharedClient = MyAPIClient(baseURL: baseURL, customerID: customerID)
}
```

jflinter:
Can we just get rid of the `sharedInit` method altogether, I mean, and just have a regular singleton, with settable properties for `baseURL` and `customerID`? I don't think it's good to have a changing static variable like this. (Sorry to seem overly nitpicky, I'm just worried about people copy/pasting this code).

benzguo:
@jflinter ah, makes sense – done here https://github.com/stripe/stripe-ios/pull/363/commits/e9b25a460a4193713b162da14f01af0f76094fcb

jflinter:
Cool, LGTM now.

PR #370, opened 2016-07-12
Update export_builds to produce fat framework

benzguo:
r? @jflinter 

The dynamic frameworks we've been attaching to recent releases aren't fat frameworks, and can't be used on the simulator. Since `carthage` treats these attachments as prebuilt binaries, this was also affecting all carthage users (I've updated `Stripe.framework.zip` on the latest release to a fat framework, so this is no longer an issue).

It seems pretty reasonable to just use `carthage build` here, but we could also substitute the raw `xcodebuild`/`lipo` commands.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 build/*
+Carthage/*
 Tests/build/*
 Tests/installation_tests/cocoapods/**/Podfile.lock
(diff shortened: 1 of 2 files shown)


jflinter:
LGTM, and yeah, using Carthage is fine by me.

Issue #368, opened 2016-07-09
STPAddCardViewController Never stop spinning


MarcoCarnevali:
Hello all, I'm implementing stripe STPAddCardViewController to add cards in my app. This is my code: 

` 

    let addCardViewController = STPAddCardViewController()

    override func viewDidLoad() {
        super.viewDidLoad()
        
        
        
        addCardViewController.delegate = self
        
        let info = STPUserInformation()
        info.email = FIRAuth.auth()?.currentUser?.email
        addCardViewController.prefilledInformation = info

        
        let navigationController = UINavigationController(rootViewController: addCardViewController)
        
        self.presentViewController(navigationController, animated: false, completion: {
           
        })
       
        
      
    }
    
    func addCardViewControllerDidCancel(addCardViewController: STPAddCardViewController) {
        print("CANCEL")
    }
    

    
    func addCardViewController(addCardViewController: STPAddCardViewController, didCreateToken token: STPToken, completion: STPErrorBlock) {

        
    }
`
The problem is that when I tap on Done, the spinner at the top right never stop spinning!
Can anyone help me? Thank you in advance.

jflinter:
Hey there,I'd make sure you're using the latest version (v8.0.2) of our SDK. There was a bug in version 8.0.0 that could cause this issue. Let me know if you're still having problems after that!





On Sat, Jul 9, 2016 at 6:41 PM +0200, "Marco Carnevali" <notifications@github.com> wrote:












Hello all, I'm implementing stripe STPAddCardViewController to add cards in my app. This is my code: 



` let STPAddCard = STPAddCardViewController()

    let addCardViewController = STPAddCardViewController()

override func viewDidLoad() {
    super.viewDidLoad()



    addCardViewController.delegate = self

    let info = STPUserInformation()
    info.email = FIRAuth.auth()?.currentUser?.email
    addCardViewController.prefilledInformation = info


    let navigationController = UINavigationController(rootViewController: addCardViewController)

    self.presentViewController(navigationController, animated: false, completion: {

    })



}

func addCardViewControllerDidCancel(addCardViewController: STPAddCardViewController) {
    print("CANCEL")
}



func addCardViewController(addCardViewController: STPAddCardViewController, didCreateToken token: STPToken, completion: STPErrorBlock) {


}




`

The problem is that when I tap on Done, the spinner at the top right never stop spinning!

Can anyone help me? Thank you in advance.

MarcoCarnevali:
[...]

PR #371, opened 2016-07-13
Add more STPAddCardViewController tests

benzguo:
r? @jflinter 

I've also updated the behavior of `addCardViewController:didCreateToken:completion:` so that calling `completion(nil)` resets the view controller's loading state. (#368)


jflinter:
lgtm

PR #372, opened 2016-07-14
Fix @param declaration in comment for STPBackendAPIAdapter

JGiola:
Update the @param declaration for `selectDefaultCustomerSource:completion:` from the wrong name `card` to the correct one of `source`


jflinter:
Thanks again @JGiola!

PR #373, opened 2016-07-14
Fix @param declaration in comment for STPPaymentMethodsViewController

JGiola:
Update the @param declaration for `paymentMethodsViewController:didFailToLoadWithError:` from the wrong name `paymentMethod` to the correct one of `error`


jflinter:
Thanks for catching, @JGiola.

PR #381, opened 2016-07-27
Change status and scroll style from bg color

bdorfman:
## Summary

Automatically change the preferred status bar style and scroll view indicator style based on the colors of the set STPTheme

This sets the status bar style of any STPTheme'd UINavigationBars and UIScrollViews. Additionally it sets the preferredStatusBarStyle for the sdk's view controllers for use by user-controlled UINavigationBars

## Motivation

When using darker background colors, the status bar and scroll indicators could be hard to see or completely invisible.

## Testing

Tested using the different themes included in the example app.

### DIFF
--- a/Stripe/UINavigationBar+Stripe_Theme.m
+++ b/Stripe/UINavigationBar+Stripe_Theme.m
@@ -8,6 +8,7 @@
 
 #import "UINavigationBar+Stripe_Theme.h"
 #import "STPTheme.h"
+#import "STPColorUtils.h"
 
 static NSInteger const STPNavigationBarHairlineViewTag = 787473;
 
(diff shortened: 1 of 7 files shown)


benzguo:
awesome, lgtm!

Stripe/UINavigationBar+Stripe_Theme.m
```diff
@@ -18,6 +19,14 @@ - (void)stp_setTheme:(STPTheme *)theme {
     [self stp_artificialHairlineView].backgroundColor = theme.tertiaryBackgroundColor;
     self.barTintColor = theme.secondaryBackgroundColor;
     self.tintColor = theme.accentColor;
+
+    if ([STPColorUtils colorIsBright:theme.secondaryBackgroundColor]) {
```
jflinter: @bdorfman just out of curiosity, how does `UIBarStyleBlack` play with the `barTintColor` and `tintColor` properties? Will it just make the bar, uh, black?

bdorfman: The documentation is a little confusing, but as far as I can tell:
1. If you are a translucent bar, it only affects status bar rendering (this is our case).
2. If you are not translucent and you have a non-opaque background image, this also controls whether the background color behind that image is white or black. If you're using tintcolor instead of a bg image it appears to do nothing.

PR #383, opened 2016-07-29
Adds paste sms code from clipboard button.

bdorfman:
Adds button to SMSCodeVC that offers to paste a code from your clipboard if it is exactly 6 numeric characters (and clears clipboard after).
Adds weak/strong macro helpers (similar to libextc) for avoiding retain cycle in blocks.
Fixes and unifies keyboard avoiding logic for SMSCodeVC and AddCardVC

### DIFF
--- a/Stripe/STPSMSCodeTextField.m
+++ b/Stripe/STPSMSCodeTextField.m
@@ -229,11 +229,7 @@
         [nextField becomeFirstResponder];
     } else {
         [textField resignFirstResponder];
-        NSMutableString *code = [NSMutableString string];
-        for (UITextField *aTextField in self.textFields) {
-            [code appendString:aTextField.text];
-        }
-        [self.delegate codeTextField:self didEnterCode:code];
+        [self.delegate codeTextField:self didEnterCode:self.code];
     }
     return NO;
 }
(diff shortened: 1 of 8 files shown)


Stripe/STPSMSCodeTextField.m
```diff
@@ -254,4 +250,23 @@ - (UITextField *)fieldAfter:(UITextField *)field {
     return [self.textFields stp_boundSafeObjectAtIndex:index+1];
 }
 
+- (NSString *)code {
```
jflinter: I like this

jflinter:
I'd really like us to get into the habit as a team of splitting large changes like this into smaller PRs, to be able to reason about their contents individually and give better code review. (In this case, I'd probably make 1 for the pasteboard thing, 1 for the macros, 1 for the keyboard avoiding fixes). Would you mind doing that @bdorfman?

jflinter:
Also, this is failing because of 2 [FauxPas](fauxpasapp.com) rules (if you haven't downloaded it yet, strongly recommended, as it runs on our CI builds), both of which are ticky-tacky but changes worth making.

bdorfman:
Ok, I will close this and re-open with separate pull requests and fixed Faux Pas checks.

PR #384, opened 2016-07-29
Adds button to SMSCodeVC that offers to paste a code

bdorfman:
If you return from inactive on SMSCodeVC, we now check your pasteboard and offer to paste a code for you if it is exactly 6 numeric characters. We then clear the clipboard after pasting (so that stale OTP values are not left in there).

### DIFF
--- a/Stripe/STPSMSCodeViewController.m
+++ b/Stripe/STPSMSCodeViewController.m
@@ -31,6 +31,7 @@
 @property(nonatomic, weak)UIButton *cancelButton;
 @property(nonatomic, weak)UILabel *errorLabel;
 @property(nonatomic, weak)UILabel *smsSentLabel;
+@property(nonatomic, weak)UIButton *pasteFromClipboardButton;
 @property(nonatomic, weak)STPPaymentActivityIndicatorView *activityIndicator;
 @property(nonatomic)BOOL loading;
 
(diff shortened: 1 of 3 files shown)


Stripe/STPSMSCodeViewController.m
```diff
@@ -108,10 +109,38 @@ - (void)viewDidLoad {
...
+    
+    NSNotificationCenter *notificationCenter = [NSNotificationCenter defaultCenter];
+    [notificationCenter addObserver:self 
+                           selector:@selector(applicationDidBecomeActive) 
+                               name:UIApplicationDidBecomeActiveNotification 
+                             object:nil];
+}
+
+- (void)applicationDidBecomeActive {
+    if (self.view.superview != nil) {
```
benzguo: what's the rationale for this check? it seems reasonable, just not something i'm in the habit of doing

bdorfman: It checks to make sure the view is still actually on screen and visible. It's mostly in there because I copied this from another screen in the app that did the same check, but maybe not necessary.

benzguo: ah cool, i think it's fine to leave it in.

benzguo:
LGTM!

PR #385, opened 2016-07-29
Fix and unify keyboard avoiding code.

bdorfman:
Move UIScrollView keyboard avoiding logic into the category method.
Avoid clobbering scroll view contentInsets
Make sure frames are converted into correct windows
Enable scrolling and fix contentSize on SMSCodeVC


benzguo:
lgtm!

PR #391, opened 2016-08-03
Convert to using WEAK/STRONG macros

bdorfman:
These macros are based off libextc's @weakify and @strongify.
This change also fixes some incorrect self-capturing retain cycles and probably some potential errors with weakself being dealloc'd partway through block execution.

### DIFF
--- a/Stripe/STPCheckoutAPIClient.m
+++ b/Stripe/STPCheckoutAPIClient.m
@@ -13,6 +13,7 @@
 #import "STPCardValidator.h"
 #import "NSBundle+Stripe_AppName.h"
 #import "StripeError.h"
+#import "STPWeakStrongMacros.h"
 
 @interface STPCheckoutAPIClient()
 @property(nonatomic, copy)NSString *publishableKey;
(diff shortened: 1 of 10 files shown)


Stripe/STPCheckoutAPIClient.m
```diff
@@ -77,27 +78,28 @@ - (BOOL)readyForLookups {
 }
 
 - (STPPromise *)lookupEmail:(NSString *)email {
-    __weak typeof(self) weakself = self;
+    WEAK(self);
+    Class selfClass = self.class;
```
benzguo: curious – why is this necessary?

Stripe/STPPaymentContext.m
```diff
@@ -69,29 +70,32 @@ - (void)retryLoading {
...
-                    [weakself.delegate paymentContext:weakself didFailToLoadWithError:error];
+                    [self.delegate paymentContext:self didFailToLoadWithError:error];
                 }
             }];
         }
     }];
     [self.apiAdapter retrieveCustomer:^(STPCustomer * _Nullable customer, NSError * _Nullable error) {
-        if (!weakself) {
+        STRONG(self);
+        if (!self) {
```
benzguo: it's odd that we weren't strongifying self in any callbacks in this file...

benzguo:
LGTM

Stripe/STPCheckoutAPIClient.m
bdorfman: So if you pull out selfClass here, then you are capturing `selfClass`. If you reference `self.class` later, that is capturing `self` (and being converted to `[self class]` basically). Under the previous code, the weak self can never actually be nil because you are capturing it inside the `if (!weakself) {}` block. In the new method, the if statement works correctly, plus we still have a ref to the class type for use in the error message.

PR #392, opened 2016-08-03
Add ability to set PKPaymentSummaryItems

bdorfman:


benzguo:
It looks like you have a faux pas lint issue, I think it's a false positive that you can ignore using `FAUXPAUS_IGNORED_ON_LINE` [0]

https://github.com/stripe/stripe-ios/blob/master/Stripe/PublicHeaders/STPAddress.h#L94

bdorfman:
Yeah I think for iOS7 APIs in a header I have to just rely on the sdk integrator knowing to not use it. I'll add the Fauxpaus exception, and maybe some comment noting its 8+ only

Stripe/PublicHeaders/STPPaymentContext.h
```diff
@@ -104,6 +104,15 @@ NS_ASSUME_NONNULL_BEGIN
 @property(nonatomic, copy)NSString *paymentCurrency;
 
 /**
+ *  If you support Apple Pay, you can optionally set the PKPaymentSummaryItems you want to display here.
+ *  If nil, a single summary item will be automatically generated using paymentAmount and your companyName.
+ *  @see PKPaymentRequest for more information
+ 
+ *  @warning If you set this array, the last summary item's amount must match the value you set for paymentAmount or an assert will be thrown.
```
benzguo: Nice. It seems likely that a user may not realize the last item needs to be the total amount, so the assert is a nice guard. I was briefly considering the alternative where we use the last line item as the `paymentAmount`, but that's probably more surprising behavior.

benzguo:
other than the lint issue, lgtm!

Stripe/STPPaymentContext.m
```diff
@@ -167,6 +167,25 @@ - (void)setSelectedPaymentMethod:(id<STPPaymentMethod>)selectedPaymentMethod {
...
+    if (paymentSummaryItems.count > 0) {
+        PKPaymentSummaryItem *lastItem = paymentSummaryItems.lastObject;
+        NSDecimalNumber *amount = [NSDecimalNumber stp_decimalNumberWithAmount:self.paymentAmount
+                                                                      currency:self.paymentCurrency];
+        if ([amount isEqualToNumber:lastItem.amount]) {
+            _paymentSummaryItems = paymentSummaryItems;
+        }
+        else {
+            _paymentSummaryItems = nil;
+            NSAssert(0, @"Amount for last object in paymentSummaryItems does not equal paymentAmount (%@ vs %@, currency = %@)", lastItem.amount, amount, self.paymentCurrency);
```
jflinter: @bdorfman - I know I'm a little late to the party here. Do you think instead of `NSAssert`ing here you could just squash the current value of `paymentAmount` when `paymentSummaryItems` is set? (That way, developers who want to use this will only have to set one property instead of 2, and not worry about keeping them in sync.)

PR #400, opened 2016-08-08
Travis builds fail on analyzer issues

bdorfman:
Use scan-build to check the output of the xcode analyzer and fail if there are any issues.

I've intentionally reintroduced an analyzer issue into this commit to check that the build fails correctly when I open a PR.


benzguo:
nice! lgtm

jflinter:
@bdorfman - I'm obviously late to the party here, but this is great!

PR #402, opened 2016-08-09
Use tinted template images for card brands

bdorfman:
Also remove checks for if the current os supports templating since our minimum is now 7.0

NOTE: Don't merge until diner's club template image is in.


Stripe/STPImageLibrary.m
```diff
@@ -144,15 +92,64 @@ + (UIImage *)largeCardApplePayImage {
...
+            imageName = isTemplate ? @"stp_card_discover_template" : @"stp_card_discover";
+            break;
+            case STPCardBrandJCB:
+            imageName = isTemplate ? @"stp_card_jcb_template" : @"stp_card_jcb";
+            break;
+            case STPCardBrandMasterCard:
+            imageName = isTemplate ? @"stp_card_mastercard_template" : @"stp_card_mastercard";
+            break;
+            case STPCardBrandUnknown:
+            isTemplate = YES;
```
benzguo: FauxPas is unhappy with modifying the argument :/

Stripe/STPImageLibrary.m
```diff
@@ -144,15 +92,64 @@ + (UIImage *)largeCardApplePayImage {
     return [self safeImageNamed:@"stp_card_form_applepay" templateifAvailable:YES];
 }
 
++ (UIImage *)safeImageNamed:(NSString *)imageName
+        templateifAvailable:(BOOL)templateIfAvailable {
```
benzguo: typo: capitalize `if`

bdorfman: I didn't make this typo, I just moved this whole method from above into the private category. Should I still go through and fix all the refs to it?

Stripe/STPImageLibrary.m
bdorfman: Yeah, I'll fix when I push the missing images.

jflinter:
It looks like stp_card_applepay.png (and all the size variants) are the wrong size. They should match the other card icons (both in their dimensions, and the size of the transparent border they contain) for consistency (even though it's not the size Apple provides).

Stripe/STPPaymentMethodsInternalViewController.m
```diff
@@ -123,8 +123,9 @@ - (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(N
     cell.backgroundColor = self.theme.secondaryBackgroundColor;
     if (indexPath.section == STPPaymentMethodCardListSection) {
         id<STPPaymentMethod> paymentMethod = [self.paymentMethods stp_boundSafeObjectAtIndex:indexPath.row];
-        cell.imageView.image = paymentMethod.image;
+        cell.imageView.image = paymentMethod.templateImage;
         BOOL selected = [paymentMethod isEqual:self.selectedPaymentMethod];
+        cell.imageView.tintColor = selected ? self.theme.accentColor : [self.theme.primaryForegroundColor colorWithAlphaComponent:0.6f];
```
jflinter: Since this color is now used in 2 places in this VC (it's also used in `buildAttributedStringForCard:selected`) can you pull it into a helper method in this class?

Stripe/STPImageLibrary.m
```diff
@@ -52,107 +52,105 @@ + (UIImage *)unknownCardCardImage {
 }
 
 + (UIImage *)brandImageForCardBrand:(STPCardBrand)brand {
```
jflinter: I sorta think we should expose these icons publicly (a la `templatedBrandImageForCardBrand`); they'll be useful to some people.

jflinter:
Beyond my comments, lgtm

PR #404, opened 2016-08-12
Have fauxpas lint the dynamic framework

bdorfman:


jflinter:
LGTM

ci_scripts/check_fauxpas.sh
```diff
@@ -20,3 +20,4 @@ fi
 echo "Linting with Faux Pas..."
 
 fauxpas check Stripe.xcodeproj/ --target "StripeiOSStatic" --configFile "./FauxPasConfig/main.fauxpas.json" --minErrorStatusSeverity Concern
```
jflinter: Oh I also remembered! You need to add `set -e` here so that the script exits if the first check fails.

bdorfman: Ah right. I'll just `set -e` on both lines here so future copy pastes dont mess that up also.

ci_scripts/check_fauxpas.sh
```diff
@@ -19,4 +19,5 @@ fi
 
 echo "Linting with Faux Pas..."
 
-fauxpas check Stripe.xcodeproj/ --target "StripeiOSStatic" --configFile "./FauxPasConfig/main.fauxpas.json" --minErrorStatusSeverity Concern
+set -e fauxpas check Stripe.xcodeproj/ --target "StripeiOSStatic" --configFile "./FauxPasConfig/main.fauxpas.json" --minErrorStatusSeverity Concern
```
jflinter: You can just `set -e` once at the top of the file; you don't have to prepend it to every line.

bdorfman: Oh is that true? All of our other scripts redo the set every line so I assumed it only worked for the next command. I can change it.

jflinter: Yeah I think they are doing so needlessly. :shrug:

PR #405, opened 2016-08-16
Add modalPresentationStyle to STPPaymentContext

benzguo:
r? @jflinter 

A user wrote in requesting this feature.

### DIFF
--- a/Stripe/PublicHeaders/STPPaymentContext.h
+++ b/Stripe/PublicHeaders/STPPaymentContext.h
@@ -114,6 +114,14 @@ NS_ASSUME_NONNULL_BEGIN
 @property(nonatomic, copy, nullable)NSArray<PKPaymentSummaryItem *> *paymentSummaryItems; FAUXPAS_IGNORED_ON_LINE(APIAvailability);
 
 /**
+ *  The presentation style used for all view controllers presented modally by the context.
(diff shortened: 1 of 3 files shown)


jflinter:
Neat, this is cool. How does this work with the SMS code view controller?

benzguo:
At the moment, the SMS code view controller just inherits the add card view controller's modal presentation style [0]. The SMS screen looks kinda odd full screen on a horizontal iPad – we could consider always making it a PageSheet...





[0] https://github.com/stripe/stripe-ios/blob/bg-modal-presentation/Stripe/STPAddCardViewController.m#L463

jflinter:
Yeah, FullScreen SMS view controller looks bad. I'm open to anything that looks reasonable on both iPhone and iPad (it looks like the half-grey background stacks with FullScreen; it'd be nice to avoid that).

benzguo:
Cool. Let's go with FormSheet [0] – it's narrower than PageSheet. Both are a no-op on iPhone.

[0] https://github.com/stripe/stripe-ios/pull/405/commits/c3fa26376b059314d994098cc9d467deb2752527

simulator screen shot aug 16 2016 6 09 25 pm

simulator screen shot aug 16 2016 6 20 11 pm

jflinter:
Nice, that looks better to me!

PR #406, opened 2016-08-16
Change how summaryItems property works.

bdorfman:
After discussing with @jflinter, made changes to the new paymentSummaryItems property so you don't need to also set paymentAmount. Now on payment context, you only set either paymentAmount OR paymentSummaryItems and the other is generated for you based on that + your paymentCurrency.

Updated some tests to reflect this new logic.

### DIFF
--- a/Stripe/STPPaymentContext.m
+++ b/Stripe/STPPaymentContext.m
@@ -8,7 +8,6 @@
 
 #import <PassKit/PassKit.h>
 #import <objc/runtime.h>
-#import "NSDecimalNumber+Stripe_Currency.h"
 #import "PKPaymentAuthorizationViewController+Stripe_Blocks.h"
 #import "UIViewController+Stripe_ParentViewController.h"
 #import "STPPromise.h"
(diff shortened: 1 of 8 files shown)


Stripe/STPPaymentContext.m
```diff
@@ -21,6 +21,19 @@
 #import "STPPaymentConfiguration+Private.h"
 
 #define FAUXPAS_IGNORED_IN_METHOD(...)
+#define FAUXPAS_IGNORED_IN_CLASS(...)
```
jflinter: My weak preference would be to move this into a different file, but feel free to ignore.

jflinter:
Nicely done; I like pulling all of that logic into a dedicated model class. Beyond maybe extracting into its own file, LGTM

Stripe/STPPaymentContext.m
bdorfman: Yeah, it seems like people have just been pasting them around, but I might make a new PR to just add all the fauxpas headers in a single file.

PR #408, opened 2016-08-17
Have payment card confirm to UIKeyInput

theill:
## Summary

Allows sending custom keys to payment card view and apply same validation logic as if key was entered through a regular keyboard by updating payment card view to confirm to UIKeyInput protocol (hasText, insertText, deleteBackward methods)

## Motivation

I've made a custom keyboard but I'm using the great `STPPaymentCardTextField` for entering card, exp month/year and cvc. However since these text fields are private I can't (beautifully) send a key event to the `STPPaymentCardTextField` view and have it apply validation logic as if I entered a digit from the system keyboard.

With this change I can simply get my reference to the `STPPaymentCardTextField` and perform `insertText` and `deleteBackward` effectively cleaning up my code so I don't need to use `stp_findFirstResponder` in connection with subview iterations and `conformsToProtocol:@protocol(UIKeyInput)` checks :)

## Testing

I did standard boundary tests and ensured to implement protocol by calling already existing methods.


benzguo:
looks good to me @theill – thanks for contributing!

PR #409, opened 2016-08-17
Fix bad pbxproj merge from #402

bdorfman:
Fixes bad pbxproj merge from card templates PR #402
Also removes now-unused non-templated placeholder card images (these were only returned if `imageWithRenderingMode:` was not supported, but since our minimum is now iOS 7 this should never come up).


benzguo:
👍

PR #410, opened 2016-08-19
Add more complete integrations to installation tests

benzguo:
r? @jflinter 

I've fleshed out the integrations in our installation test projects, which should make it easier to troubleshoot installation issues. 

While running the manual installation project, I noticed some images were missing from the resources bundle – fixed in https://github.com/stripe/stripe-ios/commit/25469f8f12c1c28acafcbba45fbbb67a1fb5ac17 , with a CI check in https://github.com/stripe/stripe-ios/pull/410/commits/7981888c4ac4165ef84476fc7fe48a767abee809


jflinter:
It looks like you may have accidentally checked in Tests/installation_tests/manual_installation/ManualInstallationTest/Frameworks/StripeiOS-Static.zip

jflinter:
Can you move the missing resource fixes + resource bundle check into a separate PR?

jflinter:
There's a lot of duplicate code here. Could you maybe move all of the code that's actually used in the assorted test apps (AppDelegate, MyAPIClient, etc) into Tests/installation_tests/source/{objc, swift} and then just import those files into the various .xcodeproj's without copying them?

jflinter:
Also, I'm not entirely sure I see what's going on here. Can you give me an example of an issue we missed that this is designed to catch?

benzguo:
👍  – I've moved the resource bundle stuff to #414 , and unified the duplicate code.

> Also, I'm not entirely sure I see what's going on here. Can you give me an example of an issue we missed that this is designed to catch?

Running the manual installation example helped me spot the missing resources. But more generally, I feel like these installation tests should have pretty complete integrations that touch most, if not all, of the APIs. I was running into this issue with carthage [0] where running the installation test would check out a very old commit (pre-8.0.0). This wasn't apparent when the integration was just `Stripe.setDefaultPublishableKey('test')`. When we have UI or snapshot tests, we could run them on each installation method. It also just seems valuable to have these around for sanity-checking/investigating installation-method specific issues.

[0] https://github.com/Carthage/Carthage/issues/968

PR #415, opened 2016-08-23
Allows non-numeric zip/postal codes to be entered

zachwick:
## Summary

Allows non-numeric ZIP/Postal codes to be entered

## Motivation

Because we explicitly were not setting the keypad layout based on phone
number locale, it was ~impossible to enter non-numeric zip/postal codes
into the zip code type fields using the Stripe iOS SDK.

## Testing

I verified that this change still compiled, but I'm unsure how to fake locales in the iOS simulator

### DIFF
--- a/Stripe/STPAddressFieldTableViewCell.m
+++ b/Stripe/STPAddressFieldTableViewCell.m
@@ -100,12 +100,14 @@
             case STPAddressFieldTypeZip:
                 if ([STPPhoneNumberValidator isUSLocale]) {
                     self.captionLabel.text = NSLocalizedString(@"ZIP Code", nil);
+                    self.textField.placeholder = NSLocalizedString(@"12345", nil);
+                    self.textField.keyboardType = UIKeyboardTypePhonePad;
                 } else {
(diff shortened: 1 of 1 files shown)


bdorfman:
Looks good to me.

For future ref: To change in simulator, you have to actually go into the sim's Settings app and change it as you would on a real phone under General -> Language & Region

jflinter:
@bdorfman: http://nshipster.com/launch-arguments-and-environment-variables/

zachwick:
Thanks for the info guys! TIL

sadlerjw:
@zachwick shouldn't this be based on the selected country, not on the user's locale? Or at least be lenient in the implementation and always allow alphanumeric entry? In some situations this could put users unable to enter their correct billing information.

zachwick:
@sadlerjw You are probably correct. However, I made my patch consistent with the other locale checking in this method. I believe that there is currently a large development effort in cleaning up the l18n support in the SDK that I imagine will address this. (cc @bdorfman since I think he is spearheading that effort)

sadlerjw:
@zachwick thanks!

PR #416, opened 2016-08-24
Check for US country code not en_US

bdorfman:
For address entry, we just want to check that their region is US, regardless of the set language.


benzguo:
👍

PR #417, opened 2016-08-25
Initial Localized Strings Support

bdorfman:
This includes all my work on supporting localized string in the sdk, although the actual translations are not in yet.

* Replace uses of NSLocalizedString with STPLocalizedString, which goes to our bundle first for strings.
* Add translator comments for most strings
* Add localization checks to Fauxpas
* Change NSError constants to class methods instead so the localization will work correctly (technically a breaking api change?)
* Combine some string references that had been copy and pasted around to ensure correctness.


benzguo:
Nice! This looks good to me. Looks like travis was failing due to FauxPas falling down – just restarted the build.

PR #418, opened 2016-08-26
Changes "from" to "to"

zachwick:
## Summary

Changes the word "from" to the word "to" to make a sentence in the docs not sound awkward

## Motivation

I was reading the docs and this sentence struck me as very oddly worded, so I figured that I should fix it. This sentence originally read "They can also add and remove credit cards from the list." and at some point bg removed "and remove credit cards" which left the word "from" being used which sounds very awkward.

## Testing

I'm unsure how to regenerate the HTML files that live in the `docs/` directory to ensure that this commit only does what I expect it to, but I  have no reason to believe that this change would have any side effect.


benzguo:
thanks @zachwick ! I've updated the PR with newly generated docs. 
(these are the steps:
1. install jazzy (`gem install jazzy`)
2. run our script (`./ci_scripts/install_jazzy.sh)
3. undo deletion of `docs/index.html` (`gco -- docs/index.html`))

PR #421, opened 2016-08-30
Better handling of international postal codes

bdorfman:
Improves support on billing address entry screen for both alphanumeric postal code support, and countries which don't generally use postal codes at all.

The user's country is now filled in automatically from their phone's region, and postal information is shown/hidden based on that. It will also update live if full billing address is enabled and the user changes their country from the dropdown.


Stripe/STPAddressFieldTableViewCell.m
```diff
@@ -197,6 +205,19 @@ - (NSString *)captionForAddressField:(STPAddressFieldType)addressFieldType {
     }
 }
 
+- (void)delegateCountryCodeDidChange {
+    
+    NSString *countryCode = nil;
+    if ([self.delegate respondsToSelector:@selector(addressFieldTableViewCountryCode)]) {
```
benzguo: can we just pass in the country code here, instead of getting it from the delegate?

Stripe/STPAddressFieldTableViewCell.m
```diff
@@ -325,8 +351,14 @@ - (BOOL)potentiallyValidContents {
         case STPAddressFieldTypeCountry:
         case STPAddressFieldTypeLine2:
             return YES;
-        case STPAddressFieldTypeZip:
-            return [STPCardValidator stringIsNumeric:self.contents];
+        case STPAddressFieldTypeZip: {
+            if (self.postalCodeType == STPCountryPostalCodeTypeNumericOnly) {
+                return  [STPCardValidator stringIsNumeric:self.contents];
```
benzguo: nit: two spaces

benzguo:
This looks good to me, aside from the two comments!

bdorfman:
Updated if you want to take another look.

benzguo:
👍

PR #428, opened 2016-08-31
Change our podspec to use a resource bundle

bdorfman:
Changing to use a named Stripe.bundle like for manual/Fabric installation so that we don't clobber people's resources when they use non-framework Cocoapods installs. Since our code is already set up to look for a `Stripe.bundle` everywhere relevant this should require no other code changes.


benzguo:
👍

bdorfman:
This is actually broken because Cocoapods will also still make Stripe.bundle even when using dynamic frameworks, and we don't expect that. Will need to rewrite bundle locator code to check for this case as well before merging.

bdorfman:
Pushed more updates to this @bg if you want to take another look.

bdorfman:
Pushed more updates to this @benzguo if you want to take another look

benzguo:
ah word, good catch – lgtm!

PR #430, opened 2016-08-31
Remove some keys from resources plist

bdorfman:
Need to remove keys in plist that reference being built for the simulator as ITMS has started rejecting apps with those, even though it is a non-binary resource bundle.


benzguo:
👍

PR #431, opened 2016-08-31
Change STPLocalizationUtils to be project instead of private

bdorfman:
Fixes a warning with certain integration methods.


benzguo:
👍

Issue #425, opened 2016-08-30
paymentRequestWithMerchantIdentifier should have availability annotation


kballard:
`+[Stripe paymentRequestWithMerchantIdentifier:]` should be marked with the Apple availability annotations as available on iOS 8 and later, with a `nonnull` return value, instead of the current available everywhere with a `nullable` return value. This will play better with Swift, and for Obj-C it will be no different than someone creating a `PKPaymentRequest` directly (i.e. they need to check if it's available first, or if their project is iOS 8+, they know it's always available).

bdorfman:
I marked public methods that use PKPaymentRequest as iOS8+ only and changed the return value as suggested. It's in master now, thanks for the suggestion!

PR #435, opened 2016-09-06
Mark PKPayment methods as only available in iOS8+

bdorfman:
Should hopefully give better warnings to people supporting iOS7 and will play more nicely with Swift.

Solves https://github.com/stripe/stripe-ios/issues/425


benzguo:
nice, 👍

PR #438, opened 2016-09-08
Translated strings

bdorfman:
I'm planning to not merge this until we actually have translations in, but this PR has final versions of all the strings we want translated pending any feedback. 

@benzguo @jflinter if you want to give it a once over and see if there are any strings you think we should change before we send out for translation.


Stripe/STPStringUtils.h
```diff
@@ -0,0 +1,52 @@
...
+ *  
+ *  Completion is always called, location of range is NSNotFound with the unmodified
+ *  string if a match could not be found.
+ *
+ *  @param string     The string with tagged substrings.
+ *  @param tag        The tag to search for.
+ *  @param completion The string with the named tag removed and the range of the
+ *                    substring it covered.
+ */
++ (void)parseRangeFromString:(NSString *)string
```
jflinter: Could you add a few unit tests for this?

jflinter:
Strings look good to me!

Stripe/STPStringUtils.h
bdorfman: 👍

Stripe/STPStringUtils.h
```diff
@@ -0,0 +1,52 @@
...
+//  Created by Brian Dorfman on 9/7/16.
+//  Copyright © 2016 Stripe, Inc. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+@interface STPStringUtils : NSObject
+/**
+ *  Takes a string with the named html-style tags, removes the tags,
+ *  and then calls the completion modified string and the range in it that the
```
benzguo: `...calls the completion block with the...`

benzguo:
looks good to me!

bdorfman:
Updated from feedback and with fauxpas fixes.

I think I will probably merge this post 8.0.7 and open a new PR when translations are in since I don't want to block on that.

PR #441, opened 2016-09-09
set text content type for cc and phone numbers

zadr:
## Summary
iOS 10 adds a new [`textContentType`](https://developer.apple.com/reference/uikit/uitextinputtraits/1649656-textcontenttype) property on `UITextInput` that helps the system have 'semantic meaning expected for a text-entry area'.

In the best-case scenario, this can help the system show the right info to fill in as a predictive text input suggestion, and in the average-case scenario, nothing changes (although, the system can now automatically switch to the correct keyboard for us). 

## Motivation

One tap to fill in info is nicer than 11-28 taps (10 for phone number, 16 for cc, hitting return 2x) (US-centric numbers, but, I think the point stands regardless of your locale).

## Testing

I made the changes and ran it in an app that integrates the `stripe-ios` pod. It relies on UIKit behavior and per-user settings, and is rather hard to test otherwise.

### DIFF
--- a/Stripe/STPFormTextField.m
+++ b/Stripe/STPFormTextField.m
@@ -108,6 +108,9 @@ typedef NSAttributedString* (^STPFormTextTransformationBlock)(NSAttributedString
         case STPFormTextFieldAutoFormattingBehaviorNone:
         case STPFormTextFieldAutoFormattingBehaviorExpiration:
             self.textFormattingBlock = nil;
+            if ([self respondsToSelector:@selector(setTextContentType:)]) {
+                self.textContentType = nil;
+            }
             break;
         case STPFormTextFieldAutoFormattingBehaviorCardNumbers:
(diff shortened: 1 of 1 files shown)


zadr:
Sorry for just sending this in without an issue for discussion first; I wanted these changes for an app I'm working on, and, it's a small enough amount of work that if y'all have it duplicated somewhere, I'm not going to lose any sleep over it being turned down.

jflinter:
Hey @zadr - we're actually just in the process of adding more iOS 10 support, and will look this over as part of that. Thanks so much for the suggestion/PR!

bdorfman:
Thanks for submitting this. Tests were failing because this branch was set to run on Xcode7 travis image. so I'm closing this in favor of https://github.com/stripe/stripe-ios/pull/455 which is off post-Xcode8 support master. Also added some guards for running with pre-iOS10 sdks (so FauxPas will work).

PR #446, opened 2016-09-13
Shipping info

benzguo:
r? @jflinter  @bdorfman 

Note that the changes to the example app here are temporary, for testing. #451 updates the example to use `STPPaymentContext` to present the shipping form.


Stripe/PublicHeaders/STPShippingMethod.h
```diff
@@ -0,0 +1,66 @@
+//
```
jflinter: We're going to drop support for iOS 7 in version 9.0, so if this branch is going to come after that it's safe to not make classes like this and just use PKShippingMethod etc.

Stripe/STPPaymentConfiguration.m
```diff
@@ -43,6 +45,7 @@ - (id)copyWithZone:(__unused NSZone *)zone {
     copy.publishableKey = self.publishableKey;
     copy.additionalPaymentMethods = self.additionalPaymentMethods;
     copy.requiredBillingAddressFields = self.requiredBillingAddressFields;
+    copy.requiredShippingAddressFields = self.requiredShippingAddressFields;
     copy.companyName = self.companyName;
     copy.appleMerchantIdentifier = self.appleMerchantIdentifier;
     copy.smsAutofillDisabled = self.smsAutofillDisabled;
```
jflinter: missing copy.shippingType

Stripe/PublicHeaders/STPShippingMethod.h
benzguo: ah good point – I'll replace.

benzguo:
@jflinter I've updated this to replace `STPShippingMethod` with `PKShippingMethod`. I've left in `STPShippingMethod`, since `PKShippingMethod` was added in iOS 8.3. 

FauxPas is failing with API availability errors, but that'll go away after we release 8.0.0.

Stripe/PublicHeaders/STPShippingAddressViewController.h
```diff
@@ -0,0 +1,92 @@
...
+ */
+@interface STPShippingAddressViewController : UIViewController
+
+/**
+ *  A convenience initializer; equivalent to calling `initWithConfiguration:[STPPaymentConfiguration sharedConfiguration] theme:[STPTheme defaultTheme] currency:nil shippingAddress:nil selectedShippingMethod:nil`.
+ */
+- (instancetype)init;
+
+/**
+ *  Initializes a new `STPAddCardViewController` with the provided parameters.
```
benzguo: `STPAddCardViewController` -> `STPShippingAddressViewController`

Example/Stripe iOS Example (Simple)/CheckoutViewController.swift
```diff
@@ -23,7 +23,7 @@ class CheckoutViewController: UIViewController, STPPaymentContextDelegate {
 
     // 3) Optionally, to enable Apple Pay, follow the instructions at https://stripe.com/docs/mobile/apple-pay
     // to create an Apple Merchant ID. Replace nil on the line below with it (it looks like merchant.com.yourappname).
-    let appleMerchantID: String? = nil
+    let appleMerchantID: String? = "foo"
```
bdorfman: Probably didnt mean to commit this?

Stripe/STPAddressViewModel.m
```diff
@@ -48,71 +48,108 @@ - (instancetype)initWithRequiredBillingFields:(STPBillingAddressFields)requiredB
...
-                [self.delegate addressViewModel:self addedCellAtIndex:0];
+        if (self.isBillingAddress && self.requiredBillingAddressFields == STPBillingAddressFieldsZip) {
+            self.addressCells = @[
+                                  [[STPAddressFieldTableViewCell alloc] initWithType:STPAddressFieldTypeZip contents:@"" lastInList:YES delegate:self]
+                                  ];
+            [self.delegate addressViewModel:self addedCellAtIndex:0];
+            [self.delegate addressViewModelDidChange:self];
+        }
+        else if ((self.isBillingAddress && self.requiredBillingAddressFields == STPBillingAddressFieldsFull) ||
+                 (!self.isBillingAddress && (self.requiredShippingAddressFields & PKAddressFieldPostalAddress))) {
```
bdorfman: This check should be `((self.requiredShippingAddressFields & PKAddressFieldPostalAddress) == PKAddressFieldPostalAddress)` (See http://sam.dods.co/blog/2014/02/01/checking-for-a-value-in-a-bit-mask/ )

Also since these `if` statement checks are getting a bit long confusing, it might be worth separating out the logic into some other method or enum. Especially since they have to exactly match the logic in the opposite branch below.

Stripe/STPShippingAddressViewController.m
```diff
@@ -0,0 +1,349 @@
...
+                                                           style:UIBarButtonItemStylePlain
+                                                          target:self
+                                                          action:@selector(cancel:)];
+    self.cancelItem = [[UIBarButtonItem alloc] initWithBarButtonSystemItem:UIBarButtonSystemItemCancel
+                                                                    target:self
+                                                                    action:@selector(cancel:)];
+    UIBarButtonItem *nextItem;
+    switch (self.configuration.shippingType) {
+        case STPShippingTypeShipping:
+            nextItem = [[UIBarButtonItem alloc] initWithTitle:STPLocalizedString(@"Next", nil)
```
bdorfman: Don't add new `STPLocalizedString`s without developer comments (although FauxPas does not correctly catch them yet :\ )

Stripe/STPShippingAddressViewController.m
```diff
@@ -0,0 +1,349 @@
...
+}
+
+- (void)updateDoneButton {
+    self.stp_navigationItemProxy.rightBarButtonItem.enabled = self.addressViewModel.isValid;
+}
+
+- (void)handleShippingValidationError:(NSError *)error {
+    [[self firstEmptyField] becomeFirstResponder];
+    NSArray *tuples = @[
+                        [STPAlertTuple tupleWithTitle:STPLocalizedString(@"OK", nil) style:STPAlertStyleCancel action:nil],
```
bdorfman: Another localized string without dev comment here

Stripe/STPAddressViewModel.m
benzguo: ah, TIL – I'll update and move the conditional into a method.

Example/Stripe iOS Example (Simple)/CheckoutViewController.swift
benzguo: ah yup, fixed now

Stripe/STPShippingAddressViewController.m
benzguo: 👍

Stripe/STPShippingAddressViewController.m
benzguo: 👍

jflinter:
@benzguo - I talked to @blabus about this for a minute, and it sounds like he hasn't had time to really review this yet design-wise. I'd like to hold off on code-reviewing this until that's done.

benzguo:
@jflinter – bill's done design review, and said it looks good. ready for API / code review!

Stripe/PublicHeaders/STPShippingAddressViewController.h
```diff
@@ -0,0 +1,92 @@
...
+ *  The view controller's delegate. This must be set before showing the view controller in order for it to work properly. @see STPShippingAddressViewControllerDelegate
+ */
+@property(nonatomic, weak) id<STPShippingAddressViewControllerDelegate> delegate;
+
+/**
+ *  If you're pushing `STPShippingAddressViewController` onto an existing `UINavigationController`'s stack, you should use this method to dismiss it, since it may have pushed an additional shipping method view controller onto the navigation controller's stack.
+ *
+ *  @param hostViewController the view controller that the shipping address view controller was pushed onto.
+ */
+- (void)dismissWithHostViewController:(UIViewController *)hostViewController;
```
jflinter: Hmm, I like the idea here but I think the "hostViewController" parameter is confusing. I think you can just figure out what that VC is automatically, right? (If we're the root of the navigation VC, then it's the navigation VC's `presentingViewController`, otherwise it's the VC before us in the stack.) Then this method can just become `dismiss` (although I might prefer `dismissWithCompletion:` to give people a hook). (It would be good to add this method to STPPaymentMethodsViewController and STPAddCardViewController as well for consistency).

Stripe/STPAddressViewModel.m
```diff
@@ -48,77 +48,117 @@ - (instancetype)initWithRequiredBillingFields:(STPBillingAddressFields)requiredB
...
                 break;
         }
-        
-        [self updatePostalCodeCellIfNecessary];
+        [self commonInit];
+    }
+    return self;
+}
+
+- (instancetype)initWithRequiredShippingFields:(PKAddressField)requiredShippingAddressFields {
```
jflinter: This class is getting pretty big; could you add some tests for it?

Stripe/PublicHeaders/STPBlocks.h
```diff
@@ -47,3 +48,12 @@ typedef void (^STPErrorBlock)(NSError * __nullable error);
  *  @param error The error returned from the response, or nil in one occurs. @see StripeError.h for possible values.
  */
 typedef void (^STPTokenCompletionBlock)(STPToken * __nullable token, NSError * __nullable error);
+
+/**
+ *  A callback to be run with a validation result and shipping methods for a 
+ *  shipping address.
+ *
+ *  @param shippingValidationError If the shipping address is invalid, an error describing the issue with the address. Will be nil if the address is valid.
```
jflinter: I feel like we should make it easier for people to say "invalid shipping address", rather than having to initialize a (potentially localized) NSError object and call this callback with it. Two ideas:
- add a category on `NSError` to build one of these (might be a little weird with swift)
- add an additional `status` enum parameter to this method, a la `typedef void (^STPShippingMethodsCompletionBlock)(STPShippingStatus status, NSError * __nullable networkError, NSArray<PKShippingMethod *>* __nonnull shippingMethods);`

Stripe/PublicHeaders/STPBlocks.h
```diff
@@ -47,3 +48,12 @@ typedef void (^STPErrorBlock)(NSError * __nullable error);
...
+ *  @param shippingMethods The shipping methods available for the address.
+ */
+typedef void (^STPShippingMethodsCompletionBlock)(NSError * __nullable shippingValidationError, NSArray<PKShippingMethod *>* __nonnull shippingMethods);
```
jflinter: Also, can we add `selectedShippingMethod` as a parameter to this block as well?

Stripe/PublicHeaders/STPPaymentConfiguration.h
```diff
@@ -11,6 +11,21 @@
 #import "STPPaymentMethod.h"
 #import "STPTheme.h"
 
+/**
```
benzguo: yep! i've moved it to STPBlocks.h in fcf3c1c

Stripe/PublicHeaders/STPAddress.h
```diff
@@ -93,6 +93,7 @@ typedef NS_ENUM(NSUInteger, STPBillingAddressFields) {
 #pragma clang diagnostic pop
 
 - (BOOL)containsRequiredFields:(STPBillingAddressFields)requiredFields;
+- (BOOL)containsRequiredPKFields:(PKAddressField)requiredFields;
```
benzguo: 👍  fcf3c1c

Stripe/STPAddressViewModel.m
benzguo: 👍  added tests in 6559d03

Stripe/PublicHeaders/STPShippingAddressViewController.h
benzguo: Ah yeah, good call. I've changed this to `dismissWithCompletion` in 86144ae , and added the same method to `STPPaymentMethodsViewController` (but not `STPAddCardViewController`, since that VC doesn't push any view controllers on itself, and can just be dismissed normally).

Stripe/PublicHeaders/STPBlocks.h
benzguo: Good call – I've gone with the `STPShippingStatus` approach in c73b5d45

benzguo:
@jflinter ready for re-review!

jack-stripe:
lgtm

PR #448, opened 2016-09-13
International address updates

bdorfman:
Changes non-US address forms to not require the State/County field.
Updates phone number validation to use country codes other than from the device's locale.

### DIFF
--- a/Stripe/STPAddress.m
+++ b/Stripe/STPAddress.m
@@ -83,12 +83,12 @@
 }
 
 - (BOOL)hasValidPostalAddress {
-    return self.line1.length > 0 &&
-    self.city.length > 0 &&
-    self.state.length > 0 &&
-    self.country.length > 0 &&
-    [STPPostalCodeValidator stringIsValidPostalCode:self.postalCode 
-                                        countryCode:self.country];
+    return (self.line1.length > 0 
+            && self.city.length > 0 
+            && self.country.length > 0 
+            && (self.state.length > 0 || ![self.country isEqualToString:@"US"])  
+            && [STPPostalCodeValidator stringIsValidPostalCode:self.postalCode 
+                                                   countryCode:self.country]);
 }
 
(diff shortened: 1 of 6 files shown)


Stripe/STPPhoneNumberValidator.m
```diff
@@ -12,37 +12,81 @@
...
+    NSString *countryCode = nillableCode;
+    if (!countryCode) {
+        countryCode = [[NSLocale autoupdatingCurrentLocale] objectForKey:NSLocaleCountryCode];
+    }
+    return countryCode;
+}
+                                                           
 + (BOOL)stringIsValidPartialPhoneNumber:(NSString *)string {
-    if (![self isUSLocale]) {
+    return [self stringIsValidPartialPhoneNumber:string forCountryCode:nil];
```
benzguo: I think `isUSLocale` is now unused and can be deleted

bdorfman: 👍

Stripe/STPAddress.m
```diff
@@ -83,12 +83,12 @@ - (BOOL)containsRequiredFields:(STPBillingAddressFields)requiredFields {
 }
 
 - (BOOL)hasValidPostalAddress {
-    return self.line1.length > 0 &&
-    self.city.length > 0 &&
-    self.state.length > 0 &&
-    self.country.length > 0 &&
-    [STPPostalCodeValidator stringIsValidPostalCode:self.postalCode 
-                                        countryCode:self.country];
+    return (self.line1.length > 0
```
benzguo: can we add a few tests for `hasValidPostalAddress` to `STPAddressTests`?

bdorfman: I added a bunch of new checks to STPAddressTests. `containsRequiredFields:` with `Full` as the type calls through to hasValidPostalAddress so this should be covered well now.

PR #449, opened 2016-09-14
Sort countries by name instead of country code

benzguo:
r? @bdorfman 

We're currently sorting countries by country code, rather than display name, which can be confusing.

screenshot 2016-09-14 18 56 21

### DIFF
--- a/Stripe/STPAddressFieldTableViewCell.m
+++ b/Stripe/STPAddressFieldTableViewCell.m
@@ -63,7 +63,15 @@
         
         NSString *countryCode = [[NSLocale autoupdatingCurrentLocale] objectForKey:NSLocaleCountryCode];
         NSMutableArray *otherCountryCodes = [[NSLocale ISOCountryCodes] mutableCopy];
+        NSLocale *locale = [NSLocale currentLocale];
         [otherCountryCodes removeObject:countryCode];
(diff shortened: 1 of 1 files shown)


Stripe/STPAddressFieldTableViewCell.m
```diff
@@ -64,6 +64,14 @@ - (instancetype)initWithType:(STPAddressFieldType)type
         NSString *countryCode = [[NSLocale autoupdatingCurrentLocale] objectForKey:NSLocaleCountryCode];
         NSMutableArray *otherCountryCodes = [[NSLocale ISOCountryCodes] mutableCopy];
         [otherCountryCodes removeObject:countryCode];
+        [otherCountryCodes sortUsingComparator:^NSComparisonResult(NSString *code1, NSString *code2) {
+            NSString *localeID1 = [NSLocale localeIdentifierFromComponents:@{NSLocaleCountryCode: code1}];
+            NSString *localeID2 = [NSLocale localeIdentifierFromComponents:@{NSLocaleCountryCode: code2}];
+            NSLocale *locale = [NSLocale autoupdatingCurrentLocale];
```
bdorfman: Probably should pull out the locale outside the block for the weird race condition of it changing mid-sort?

benzguo: ah yup, good catch – fixed

PR #452, opened 2016-09-15
Xcode 8 / Swift 3 / App Extension

bdorfman:
Upgrades project to Xcode8 settings
Updates demo app to Swift 3
Enables "app extension api only" setting for dynamic framework.
Removes iOS7 support.


Example/Stripe iOS Example (Custom).xcodeproj/project.pbxproj
```diff
@@ -328,7 +337,7 @@
 				);
 				INFOPLIST_FILE = "Stripe iOS Example (Custom)/Info.plist";
 				LD_RUNPATH_SEARCH_PATHS = "$(inherited) @executable_path/Frameworks";
-				PRODUCT_BUNDLE_IDENTIFIER = "com.stripe.$(PRODUCT_NAME:rfc1034identifier)";
+				PRODUCT_BUNDLE_IDENTIFIER = com.stripe.CustomSDKExample;
```
jflinter: This might be unintended?

Stripe.podspec
```diff
@@ -6,11 +6,11 @@ Pod::Spec.new do |s|
   s.homepage                       = 'https://stripe.com/docs/mobile/ios'
   s.authors                        = { 'Jack Flintermann' => 'jack@stripe.com', 'Stripe' => 'support+github@stripe.com' }
   s.source                         = { :git => 'https://github.com/stripe/stripe-ios.git', :tag => "v#{s.version}" }
-  s.frameworks                     = 'Foundation', 'Security'
+  s.frameworks                     = 'Foundation', 'Security', 'WebKit'
```
jflinter: I've never fully understood what this setting does, but we should maybe add `AddressBook` as well?

jflinter: Should we add `AddressBook` here too?

Stripe/UIViewController+Stripe_Alerts.m
```diff
@@ -15,47 +15,6 @@
```
jflinter: I think this file can probably be deleted outright now, as it no longer really has any advantages over just using a bare `UIAlertController`.

Tests/installation_tests/cocoapods/with_frameworks/CocoapodsTestTests/CocoapodsTestTests.swift
```diff
@@ -28,7 +28,7 @@ class CocoapodsTestTests: XCTestCase {
     
     func testPerformanceExample() {
```
jflinter: You can just delete this function.

Stripe.podspec
bdorfman: Probably. It is unclear to me why PassKit and AddressBook are listed as weak frameworks instead of required. Any thoughts on that?

jflinter: Not sure, really. I think we should just make both of them required.

Tests/installation_tests/cocoapods/with_frameworks/CocoapodsTestTests/CocoapodsTestTests.swift
bdorfman: Yeah all of the integration test projects have these empty default test stuff in them (this one is just in the diff because the swift3 migrator made changes). I think we might want to go through and either delete them all or add tests to them all, but I think that is a separate PR?

bdorfman:
I removed the category stuff and updated the podspec. I'm going to delete/revamp integration test stuff in a later PR. Any other concerns before merging?

jflinter:
Nope! Merge away! Nice work!

jflinter:
lgtm

PR #457, opened 2016-09-21
Fix xcode autocompletion of escaping closures

jflinter:
r? @bdorfman


bdorfman:
I think having these extra compiler macros might be more confusing than just putting the block signature right into the method declaration, even with block syntax being so arcane.

jflinter:
I think given that this issue has been fixed in the latest Swift, and that a new Xcode is coming ~soon, that it's ok to just close this.

PR #459, opened 2016-09-21
fix nil window issue on iOS 8

benzguo:
r? @bdorfman 

While doing some testing, I noticed our example app doesn't respond to touches on an iOS 8 simulator [0]

[0] http://stackoverflow.com/questions/25963101/unexpected-nil-window-in-uiapplicationhandleeventfromqueueevent


bdorfman:
👍

PR #466, opened 2016-09-27
Status bar color improvements

benzguo:
r? @bdorfman 
cc @phred 

This PR fixes a few visual bugs:
- the "Add new card" cell wasn't using the theme's background color
- Navigation bars were using `secondaryBackgroundColor`, rather than `primary`
- Navigation bars were defaulting to `translucent=YES`

I've also added two new properties to `STPTheme`, `barStyle` and `translucentNavigationBar`.

### DIFF
--- a/Stripe/STPTheme.m
+++ b/Stripe/STPTheme.m
@@ -7,6 +7,7 @@
 //
 
 #import "STPTheme.h"
+#import "STPColorUtils.h"
 
 @interface STPTheme()
 @end
@@ -60,6 +61,10 @@ static UIFont  *STPThemeDefaultMediumFont;
         _errorColor = STPThemeDefaultErrorColor;
         _font = STPThemeDefaultFont;
         _emphasisFont = STPThemeDefaultMediumFont;
+        _translucentNavigationBar = NO;
+        // This is a sentinel value (the equivalent of nil).
+        // If unset, we return the default computed bar style
+        _barStyle = -1;
     }
     return self;
 }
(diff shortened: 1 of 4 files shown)


Stripe/STPTheme.m
```diff
@@ -60,6 +61,8 @@ - (instancetype)init {
         _errorColor = STPThemeDefaultErrorColor;
         _font = STPThemeDefaultFont;
         _emphasisFont = STPThemeDefaultMediumFont;
+        _translucentNavigationBar = NO;
+        _barStyle = -1;
```
bdorfman: Should probably initialize to Default instead of a non-enum value.

bdorfman:
Looks good otherwise.

Stripe/STPTheme.m
benzguo: I'm using -1 as a sentinel value: if the user hasn't set `barStyle`, we use the computed default value, otherwise we use their chosen value. I'm open to other suggestions for accomplishing the same behavior – maybe an internal boxed enum would be better?

```
- (UIBarStyle)barStyle {
    UIBarStyle defaultStyle = [self barStyleForColor:self.primaryBackgroundColor];
    if (_barStyle < 0) {
        return defaultStyle;
    }
    else {
        return _barStyle;
    }
}
```

bdorfman: Ah I see. Yeah I think this is fine, but maybe put some comments in where you assign it in init to make it clear.

benzguo: 👍 , added a comment

PR #468, opened 2016-09-28
add STPFraudSignals

benzguo:
r? @jflinter

### DIFF
--- a/Stripe.xcodeproj/project.pbxproj
+++ b/Stripe.xcodeproj/project.pbxproj
@@ -437,8 +437,6 @@
 		C1363BBA1D7633D800EB82B4 /* STPPaymentMethodTableViewCell.m in Sources */ = {isa = PBXBuildFile; fileRef = C1363BB61D7633D800EB82B4 /* STPPaymentMethodTableViewCell.m */; };
 		C13DA3BA1D5955BE00D36B95 /* STPOptimizationMetrics.h in Headers */ = {isa = PBXBuildFile; fileRef = C11CFE7F1D39384300E8CAE6 /* STPOptimizationMetrics.h */; };
 		C13DA3BB1D5958E700D36B95 /* STPOptimizationMetrics.m in Sources */ = {isa = PBXBuildFile; fileRef = C11CFE801D39384300E8CAE6 /* STPOptimizationMetrics.m */; };
-		C16A4CDD1D36B19B001F46D2 /* MockSTPCheckoutAPIClient.h in Headers */ = {isa = PBXBuildFile; fileRef = C16A4CDB1D36B19B001F46D2 /* MockSTPCheckoutAPIClient.h */; };
-		C16A4CDE1D36B19B001F46D2 /* MockSTPCheckoutAPIClient.m in Sources */ = {isa = PBXBuildFile; fileRef = C16A4CDC1D36B19B001F46D2 /* MockSTPCheckoutAPIClient.m */; };
 		C16F66AB1CA21BAC006A21B5 /* STPFormTextFieldTest.m in Sources */ = {isa = PBXBuildFile; fileRef = C16F66AA1CA21BAC006A21B5 /* STPFormTextFieldTest.m */; };
 		C1717DB11CC00ED60009CF4A /* STPAddress.h in Headers */ = {isa = PBXBuildFile; fileRef = C1080F471CBECF7B007B2D89 /* STPAddress.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		C17A030D1CBEE7A2006C819F /* STPAddressFieldTableViewCell.h in Headers */ = {isa = PBXBuildFile; fileRef = C17A030B1CBEE7A2006C819F /* STPAddressFieldTableViewCell.h */; };
--- a/Stripe.xcodeproj/project.pbxproj
+++ b/Stripe.xcodeproj/project.pbxproj
@@ -396,6 +396,8 @@
 		C11810C11CC7DA290022FB55 /* stp_card_form_front.png in Resources */ = {isa = PBXBuildFile; fileRef = C11810BB1CC7DA290022FB55 /* stp_card_form_front.png */; };
 		C11810C21CC7DA290022FB55 /* stp_card_form_front@2x.png in Resources */ = {isa = PBXBuildFile; fileRef = C11810BC1CC7DA290022FB55 /* stp_card_form_front@2x.png */; };
 		C11810C31CC7DA290022FB55 /* stp_card_form_front@3x.png in Resources */ = {isa = PBXBuildFile; fileRef = C11810BD1CC7DA290022FB55 /* stp_card_form_front@3x.png */; };
+		C11CFE811D39384300E8CAE6 /* STPFraudSignals.h in Headers */ = {isa = PBXBuildFile; fileRef = C11CFE7F1D39384300E8CAE6 /* STPFraudSignals.h */; };
+		C11CFE821D39384300E8CAE6 /* STPFraudSignals.m in Sources */ = {isa = PBXBuildFile; fileRef = C11CFE801D39384300E8CAE6 /* STPFraudSignals.m */; };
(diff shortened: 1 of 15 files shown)


Stripe.xcodeproj/project.pbxproj
```diff
@@ -433,6 +435,10 @@
 		C1363BB81D7633D800EB82B4 /* STPPaymentMethodTableViewCell.h in Headers */ = {isa = PBXBuildFile; fileRef = C1363BB51D7633D800EB82B4 /* STPPaymentMethodTableViewCell.h */; };
 		C1363BB91D7633D800EB82B4 /* STPPaymentMethodTableViewCell.m in Sources */ = {isa = PBXBuildFile; fileRef = C1363BB61D7633D800EB82B4 /* STPPaymentMethodTableViewCell.m */; };
 		C1363BBA1D7633D800EB82B4 /* STPPaymentMethodTableViewCell.m in Sources */ = {isa = PBXBuildFile; fileRef = C1363BB61D7633D800EB82B4 /* STPPaymentMethodTableViewCell.m */; };
+		C13DA3BA1D5955BE00D36B95 /* STPOptimizationMetrics.h in Headers */ = {isa = PBXBuildFile; fileRef = C11CFE7F1D39384300E8CAE6 /* STPOptimizationMetrics.h */; };
+		C13DA3BB1D5958E700D36B95 /* STPOptimizationMetrics.m in Sources */ = {isa = PBXBuildFile; fileRef = C11CFE801D39384300E8CAE6 /* STPOptimizationMetrics.m */; };
+		C16A4CDD1D36B19B001F46D2 /* MockSTPCheckoutAPIClient.h in Headers */ = {isa = PBXBuildFile; fileRef = C16A4CDB1D36B19B001F46D2 /* MockSTPCheckoutAPIClient.h */; };
```
jflinter: I think these were included by accident.

Stripe/STPOptimizationMetrics.h
```diff
@@ -0,0 +1,18 @@
...
+//  Copyright © 2016 Stripe, Inc. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+extern NSString *const STPUserDefaultsKeyFirstAppOpenTime;
+extern NSString *const STPUserDefaultsKeyTotalAppOpenCount;
+extern NSString *const STPUserDefaultsKeyTotalAppUsageDuration;
+
+@interface STPOptimizationMetrics : NSObject
```
jflinter: I think it would be better to just use `STPOptimizationMetrics` as a singleton, and then you can migrate all of the notification logic here instead. It'll make the interfaces cleaner too.

Stripe/STPAnalyticsClient.m
```diff
@@ -23,13 +23,16 @@
 #import "STPPaymentMethodsViewController.h"
 #import "STPPaymentMethodsViewController+Private.h"
 #import "STPAPIClient+ApplePay.h"
+#import "STPOptimizationMetrics.h"
+#import "STPSMSCodeViewController.h"
```
jflinter: Unneeded #import

Stripe/STPPaymentConfiguration.m
```diff
@@ -30,6 +26,7 @@ + (instancetype)sharedConfiguration {
 - (instancetype)init {
     self = [super init];
     if (self) {
+        __unused STPAnalyticsClient *client = [STPAnalyticsClient sharedClient];
```
jflinter: Can you add a comment explaining why this is needed?

Tests/Tests/STPAnalyticsClientTest.m
```diff
@@ -8,11 +8,15 @@
 
 #import <XCTest/XCTest.h>
 #import "STPAnalyticsClient.h"
+#import "STPPaymentConfiguration.h"
+#import "STPOptimizationMetrics.h"
+#import "STPPaymentMethodsViewController.h"
+#import "STPAddCardViewController.h"
```
jflinter: More unneeded #imports

Stripe/STPPaymentConfiguration.m
benzguo: yup! (it's so we can register for the didBecomeActive et al notifications)

benzguo:
@jflinter 👍 , I've made `STPOptimizationMetrics` a singleton, definitely cleaner now

benzguo:
ping @jflinter

benzguo:
@jflinter Can you take a look again? I've made this opt-in via `Stripe.enableFraudSignalsCollection()`.  I think it'll be easier to talk to merchants about enabling this once it's out in an SDK version.

Stripe/STPFraudSignals.m
```diff
@@ -30,12 +32,28 @@ + (instancetype)sharedInstance {
...
+#pragma clang diagnostic pop
+}
+
 - (instancetype)init {
     self = [super init];
     if (self) {
-        [UIDevice currentDevice].batteryMonitoringEnabled = YES;
-        [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(applicationDidBecomeActive) name:UIApplicationDidBecomeActiveNotification object:nil];
-        [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(applicationDidEnterBackground) name:UIApplicationDidEnterBackgroundNotification object:nil];
+        if ([STPFraudSignals shouldCollectFraudSignals]) {
```
jack-stripe: Shouldn't all this just get moved to `enable` now?

bg-stripe: oh whoops, yup – updated in 77f724

Stripe/STPFraudSignals.m
```diff
@@ -0,0 +1,151 @@
...
+    return sharedClient;
+}
+
++ (void)enable {
+    STPFraudSignalsCollectionEnabled = YES;
+    [[self sharedInstance] enable];
+}
+
+- (void)enable {
+    [UIDevice currentDevice].batteryMonitoringEnabled = YES;
```
jack-stripe: I would only execute this code if we're not currently enabled.

Stripe/STPAPIClient+ApplePay.m
```diff
@@ -11,25 +11,26 @@
...
+    NSDictionary *metrics = [[STPFraudSignals sharedInstance] serialize];
+    NSData *data = [self.class formEncodedDataForPayment:payment metrics: metrics];
+    [self createTokenWithData:data
                    completion:completion];
 }
 
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wdeprecated"
-+ (NSData *)formEncodedDataForPayment:(PKPayment *)payment {
++ (NSData *)formEncodedDataForPayment:(PKPayment *)payment metrics:(NSDictionary *)metrics {
```
jack-stripe: This makes me think that we've chosen the wrong abstraction in `STPAPIClient`. Ideally, all of these extensions would need no knowledge of `STPFraudMetrics`. They need this because `STPAPIClient` currently uses `NSData` to represent data that is ready to go over the wire. I think this is wrong. I think it should instead use an `NSDictionary` of key pairs that it then internally converts to an `NSData` and sends to the API. This would allow it to add things like additional metrics seamlessly under the hood (it could also do this now by deserializing the NSData, modifying it, and reserializing it, but that would be horrible.)

This would obviously need to be a separate change, and would need to be extremely well-tested, but I think would be the better way to approach changes like this in the future. I'd be bullish on making that change and then rebasing this PR off of it.

billinghamj:
Is this no longer happening? :s

bg-stripe:
@billinghamj – yeah, we've decided to postpone this project until we have a better value proposition. Was your team interested?

billinghamj:
Yes, definitely. We currently use Ravelin for this, as we can give them a lot more data than we can give to Stripe. If Stripe can start to collect more data, it can be effective enough by itself.

bg-stripe:
Cool, good to know – I'll make sure you're on our beta list for fraud related products.

PR #470, opened 2016-09-30
Update ObjC example to use /charge_card endpoint

benzguo:
r? @jflinter 

Companion to https://github.com/stripe/example-ios-backend/pull/7


jflinter:
LGTM

PR #474, opened 2016-10-10
Add new mastercard bin range

bdorfman:
Also converted code to use dispatch_once while I was in there.

r? @jflinter


jflinter:
lgtm

PR #483, opened 2016-10-20
Fix UIAppearance properties in STPPaymentCardTextField 

### LINKED ISSUE #482: STPPaymentCardTextField- UIAppearance proxies being overridden
## Summary

Setting the appearance proxies for `borderColor` and `borderWidth` on the `STPPaymentCardTextField` is being overridden by the default vales during initialization. 

## iOS version

9.2

## Installation method

Cocoapods

## SDK version

9.0.0

## Other information

```
    self.borderColor = [self.class placeholderGrayColor];
    self.cornerRadius = 5.0f;
    self.borderWidth = 1.0f;
```

These are ignoring any value set by the appearance proxy.

benzguo:
r? @bdorfman 

Fixes #482 

> UIAppearance swizzles all setters that have a default appearance, and tracks when they get changed, so that UIAppearance doesn’t override your customizations. This is a problem here, since we use setters in the initializer, and for UIAppearance it now looks as though we already customized the class ourselves.

http://petersteinberger.com/blog/2013/uiappearance-for-custom-views/

### DIFF
--- a/Stripe/STPPaymentCardTextField.m
+++ b/Stripe/STPPaymentCardTextField.m
@@ -45,6 +45,9 @@
 @synthesize textColor = _textColor;
 @synthesize textErrorColor = _textErrorColor;
 @synthesize placeholderColor = _placeholderColor;
+@synthesize borderColor = _borderColor;
+@synthesize borderWidth = _borderWidth;
+@synthesize cornerRadius = _cornerRadius;
 @dynamic enabled;
 
 CGFloat const STPPaymentCardTextFieldDefaultPadding = 13;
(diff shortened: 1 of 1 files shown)


Stripe/STPPaymentCardTextField.m
```diff
@@ -74,10 +77,12 @@ - (instancetype)initWithFrame:(CGRect)frame {
 }
 
 - (void)commonInit {
-    
-    self.borderColor = [self.class placeholderGrayColor];
-    self.cornerRadius = 5.0f;
-    self.borderWidth = 1.0f;
+    _borderColor = [self.class placeholderGrayColor];
+    _cornerRadius = 5.0f;
+    _borderWidth = 1.0f;
```
bdorfman: Since we're just passing through to the layer, can you just not have an instance variable and pass the setters and getters through, or does that not work with UIAppearance?

benzguo: We were doing that before, but UIAppearance tracks when setters are called, and won't override properties that have already been customized (hence the issue in #482).

bdorfman: I see. I think then probably you should add some comments explaining the issue in here so people dont delete these in a future refactor, but otherwise looks good.

Issue #477, opened 2016-10-16
STPPaymentMethodsViewController: This application is modifying the autolayout engine from a background thread


ermterm:
## Summary

Using STPPaymentContext, adding a card takes a **long time** before the STPPaymentMethodsViewController is dismissed (15-30 seconds with spinning loading indicator in top right)... Then after finally dismissing, autolayout engine reports error.

The basic steps to reproduce, are...

1)  Use a newly created Stripe customer.
2)  Call [self.paymentContext presentPaymentMethodsViewController];
3)  Enter in CC details and tap "Done"

**Expected Result:**  After a short load, the view controller should dismiss, and my UI should be updated via the "paymentContextDidChange" method.

**Actual:**  "paymentContextDidChange" is called _while_ the view controller is still present and loading.  Loading continues for 15-30 seconds. After finally dismissing, an error is logged:

```
2016-10-16 16:36:38.159 MyTestApp[57997:6514743] This application is modifying the autolayout engine from a background thread after the engine was accessed from the main thread. This can lead to engine corruption and weird crashes.
 Stack:(
	0   CoreFoundation                      0x000000011171a34b __exceptionPreprocess + 171
	1   libobjc.A.dylib                     0x000000011117b21e objc_exception_throw + 48
	2   CoreFoundation                      0x0000000111783265 +[NSException raise:format:] + 197
	3   Foundation                          0x000000010e3a4320 _AssertAutolayoutOnAllowedThreadsOnly + 180
	4   Foundation                          0x000000010e3a4068 -[NSISEngine _optimizeWithoutRebuilding] + 61
	5   Foundation                          0x000000010e1cbe8c -[NSISEngine optimize] + 91
	6   Foundation                          0x000000010e3a1f88 -[NSISEngine performPendingChangeNotifications] + 84
	7   UIKit                               0x000000010ee38517 -[UIView(CALayerDelegate) layoutSublayersOfLayer:] + 1704
	8   QuartzCore                          0x000000010e7cbcdc -[CALayer layoutSublayers] + 146
	9   QuartzCore                          0x000000010e7bf7a0 _ZN2CA5Layer16layout_if_neededEPNS_11TransactionE + 366
	10  QuartzCore                          0x000000010e7bf61e _ZN2CA5Layer28layout_and_display_if_neededEPNS_11TransactionE + 24
	11  QuartzCore                          0x000000010e74d62c _ZN2CA7Context18commit_transactionEPNS_11TransactionE + 280
[...]

PR #487, opened 2016-10-24
Guard all escaping completions with dispatch

bdorfman:
Any delegate method where we pass a completion block that might be called outside of the framework itself, we should make sure to dispatch back to the main thread ourselves inside the completion instead of relying on the user to make sure they are calling it on the correct thread.

I believe this fixes https://github.com/stripe/stripe-ios/issues/477

r? @benzguo

### DIFF
--- a/Stripe/STPDispatchFunctions.m
+++ b/Stripe/STPDispatchFunctions.m
@@ -1,0 +1,18 @@
+//
+//  STPDispatchFunctions.m
+//  Stripe
+//
+//  Created by Brian Dorfman on 10/24/16.
+//  Copyright © 2016 Stripe, Inc. All rights reserved.
+//
+
+#include "STPDispatchFunctions.h"
+
+void stpDispatchToMainThreadIfNecessary(dispatch_block_t block) {
+    if ([NSThread isMainThread]) {
+        block();
+    }
+    else {
+        dispatch_async(dispatch_get_main_queue(), block);
+    }
+}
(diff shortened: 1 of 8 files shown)


bdorfman:
Changed to only dispatch_async to main if we are not already on main. This maintains previous synchronous behavior for any code that is already firing on main thread (like some of our tests rely on) but prevents strange threading behavior or crashes for people not going back to main before firing our completion blocks.

Stripe/STPDispatchFunctions.m
```diff
@@ -0,0 +1,18 @@
...
+//  STPDispatchFunctions.m
+//  Stripe
+//
+//  Created by Brian Dorfman on 10/24/16.
+//  Copyright © 2016 Stripe, Inc. All rights reserved.
+//
+
+#include "STPDispatchFunctions.h"
+
+void stpDispatchToMainThreadIfNecessary(dispatch_block_t block) {
```
benzguo: can we update `STPPromise.m` to use this?

bdorfman:
Yeah totally. Wasn't sure if that was desired behavior or not, I wasn't 100% sure of the implications of changing STPPromise like this. It's probably fine though?? I can make the change.

bdorfman:
Changes pushed. I totally didn't realize STPPromise already did this main thread checking but just with a lot more copy-paste :( Should be better now hopefully.

PR #493, opened 2016-11-04
Move cartfile -> cartfile.private

bdorfman-stripe:


jack-stripe:
LGTM. Does this affect the contents of Cartfile.resolved at all? Maybe `rm Cartfile.resolved && carthage update` to be safe?

bdorfman-stripe:
I already did that and it regen'd an identical file so I think it's fine.

jack-stripe:
🆒

PR #505, opened 2016-11-17
Add [STPCard stringFromBrand]

### LINKED ISSUE #498: STPCardParams (or STPPaymentCardTextField) should return card brand if availaible
There is currently no way to get the Card Brand as a `NSString * __Nullable` from `STPCardParams` or `STPPaymentCardTextField`

`STPPaymentCardTextField` has a way to return the brand as an Image, but not as a string.

There are many applications where this would be useful, but my case is for Optimistic UI to display the card brand in a list before "technically" saving the card via the Stripe API.

jack-stripe:
Currently we don't publicly expose any way to actually convert an STPCardBrand to a string. This adds it to STPCard and deletes the private category on NSString that we were using to do it before. Fixes #498. r? @bdorfman-stripe

This reverts commit 709b543ec589a1601ff118648f67196357b26930.


bdorfman-stripe:
Looks good

PR #510, opened 2016-11-28
Check for nil self in loadingPromise completion

### LINKED ISSUE #500: BAD_ACCESS when user cancels payment before STPBackendAPIAdapter methods calls callback
## Summary

If User Tap Cancel button before my implementation of -(void)retrieveCustomer:(STPCustomerCompletionBlock)completion calls completion block, application crashes with BAD_ACCESS.

Exact place of crash is STPColorUtils.m line 15

## Code to reproduce

It's easy to reproduce this behaviour with small delay on callback call

```
dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(10 * NSEC_PER_SEC)), dispatch_get_main_queue(), ^{
    if (serializer.customer) {
        completion(serializer.customer, nil);
    } else {
        DDLogError(@"%@", serializer.error);
        completion(nil, serializer.error);
    }
});

```
## iOS version

9.3

## Installation method

CocoaPods

## SDK version

Stripe (9.1.0)

## Other information

I can't catch this in my code because there is no exception there just BAD_ACCESS. And I can't predict fault because there is no special state of payment context for this situation.

bg-stripe:
r? @bdorfman-stripe 
fixes #500 

I was able to reproduce this issue by pointing the example app to an example backend, adding a delay to `retrieveCustomer`, and hitting cancel before the view controller finishes loading. Either `STPPaymentMethodsVC` or `STPAddCardVC` can crash, depending on whether or not the customer has saved cards.

### DIFF
--- a/Stripe/STPPaymentMethodsViewController.m
+++ b/Stripe/STPPaymentMethodsViewController.m
@@ -113,6 +113,9 @@
     WEAK(self);
     [self.loadingPromise onSuccess:^(STPPaymentMethodTuple *tuple) {
         STRONG(self);
+        if (!self) {
+            return;
+        }
         UIViewController *internal;
         if (tuple.paymentMethods.count > 0) {
             internal = [[STPPaymentMethodsInternalViewController alloc] initWithConfiguration:self.configuration


bdorfman-stripe:
LGTM

PR #511, opened 2016-11-28
Fix country field bug in shipping form

### LINKED ISSUE #509: Country Picker Doesn't Update on First Selection
## Summary
This should be a relatively simple fix; however, on first selection the country code picker doesn't update. For example, when I click on the picker the option United States shows up but I actually have to scroll to Afghanistan and then scroll back up to United States for the value to update.

## Code to reproduce
Any project using Stripe's STPShippingAddressViewController has this issue. Just present the View Controller and try changing the country. When you select it the value isn't updated. You have to scroll it to Afghanistan and back to the United States in order to select the US.

## iOS version
iOS 10.1.1 (14B100)

## Installation method
Cocoapods

## SDK version
9.2.0

bg-stripe:
r? @bdorfman-stripe 

Fixes #509 

This bug occurs when `STPShippingAddressVC` is passed `prefilledInformation`:
  - We were creating a new address with a `nil` country
  - The address setter wasn't calling `delegateCountryCodeDidChange`

## Testing

I've added a regression assert to `STPAddressViewModelTest`

### DIFF
--- a/Stripe/STPShippingAddressViewController.m
+++ b/Stripe/STPShippingAddressViewController.m
@@ -75,6 +75,7 @@
         }
         else if (prefilledInformation != nil) {
             STPAddress *prefilledAddress = [STPAddress new];
+            prefilledAddress.country = _addressViewModel.address.country;
             if (self.configuration.requiredShippingAddressFields & PKAddressFieldEmail) {
                 prefilledAddress.email = prefilledInformation.email;
             }
--- a/Stripe/STPAddressFieldTableViewCell.m
+++ b/Stripe/STPAddressFieldTableViewCell.m
@@ -305,7 +305,11 @@
 
 - (void)setContents:(NSString *)contents {
     _contents = contents;
-    self.textField.text = contents;
+    if (self.type == STPAddressFieldTypeCountry) {
+        [self updateTextFieldsAndCaptions];
+    } else {
+        self.textField.text = contents;
+    }
     if ([self.textField isFirstResponder]) {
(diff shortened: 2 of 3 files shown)


Stripe/STPAddressViewModel.m
```diff
@@ -224,6 +224,7 @@ - (void)setAddress:(STPAddress *)address {
                 break;
             case STPAddressFieldTypeCountry:
                 cell.contents = address.country;
+                [cell delegateCountryCodeDidChange:address.country];
```
bdorfman-stripe: This should be called above on line 203? Is that not working properly somehow?

bg-stripe: ah yup, good point, I was a bit too quick to jump to this as a fix. Turns out the issue was actually in `setContents` – previously, we were setting the country field's text to the country code, instead of getting the name from the country picker.

Stripe/STPShippingAddressViewController.m
```diff
@@ -75,6 +75,7 @@ - (instancetype)initWithConfiguration:(STPPaymentConfiguration *)configuration
         }
         else if (prefilledInformation != nil) {
             STPAddress *prefilledAddress = [STPAddress new];
+            prefilledAddress.country = _addressViewModel.address.country;
```
bdorfman-stripe: Is this bug limited to the shipping screen or should we make the same fix for billing also?

bdorfman-stripe:
LGTM

PR #517, opened 2016-12-02
Force sdk to use same localization as main bundle.

bdorfman-stripe:
If the language supported by the main app's bundle is different than the language supported by our bundle, switch to pulling strings from the main bundle.
This lets users of the sdk fill in their own sdk localizations for languages we don't support in their main strings files instead of having to fork us.
It also prevents edge cases where the sdk could possibly use a different localization than the app (i think iOS actually prevents this, at least on iOS 10, but I don't like to rely on undocumented Foundation behavior).

### DIFF
--- a/Stripe/STPBundleLocator.m
+++ b/Stripe/STPBundleLocator.m
@@ -28,22 +28,27 @@
      4. main bundle (for people dragging all our files into their project)
      **/
     
-    NSBundle *ourBundle = [NSBundle bundleWithPath:@"Stripe.bundle"];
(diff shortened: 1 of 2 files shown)


jack-stripe:
👍 
Have you thought about how to document this behavior? Maybe a "localizations" section in the README or /docs/mobile/ios?

bdorfman-stripe:
Good idea, I will add a Localizations section to the readme and discuss with the docs team how best to add this to the stripe.com docs.

Issue #480, opened 2016-10-18
UINavigationBar Theme Styles Are Ignored By UIBarButtonItems


grahamburgsma:
## Summary

Setting navigation bar theme style colors are being ignored by the bar button items. The `primaryBackgroundColor` works but `accentColor` `primaryForegroundColor` and `secondaryForegroundColor` are ignored. Please see the screenshot and code below. I would like to be able to change the color of the back button and done button from the navigation bar theme.
## Code to reproduce

```
let theme = STPTheme()
theme.primaryBackgroundColor = UIColor.BBBlue() //Custom blue color
theme.accentColor = UIColor.white
theme.primaryForegroundColor = UIColor.white
theme.secondaryForegroundColor = UIColor.white

self.navigationController?.navigationBar.stp_setTheme(theme)

self.paymentContext.pushPaymentMethodsViewController()
```
## iOS version

iOS 9.0+
## Installation method

Cocoapods 1.0.1
## SDK version

9.0
## Screenshot

![navigationbar style](https://cloud.githubusercontent.com/assets/9092289/19478898/497f74ec-9512-11e6-8967-c4d1014ef144.png)

benzguo:
Hey @grahamburgsma –

It looks like `STPPaymentContext` is using bar button colors from the default theme.

In order to set the theme of the UI that `STPPaymentContext` pushes, you'll need to initialize `STPPaymentContext` with your theme:
```swift
let paymentContext = STPPaymentContext(apiAdapter: yourAPIAdapter,
                                       configuration: yourConfig,
                                       theme: yourTheme)
```

Let us know if you have any further issues!

Ben

grahamburgsma:
Hey @benzguo 

Thanks for your reply. I tried as you suggested but that will not fix the problem.

I propose that the navigation bar theme should affect bar button items as they are part of the same component. It is odd behaviour that the theme set for `STPPaymentContext` affects the bar button items. In the screenshot from above, I would like the navigation bar to match the rest of my app but the stripe view controllers to have the default theme.

To summarize: `UIBarButtonItems` should inherit the theme from the `UINavigationBar` instead of from the `ViewController` (`STPPaymentContext` in my case).

Let me know what you think of this proposal. Thanks!

Graham

benzguo:
Hey @grahamburgsma ,

I think I understand your issue now – you'd like to be able to control the theming of the navigation bar and stripe view controllers separately. 

[...]

PR #520, opened 2016-12-12
Bar button items inherit from bar theme if applicable

bdorfman-stripe:
If your UINavigationBar has an stp_theme, use that theme for bar button styling instead of the view controller theme.

Adds support for the use case described in https://github.com/stripe/stripe-ios/issues/480

r? @jack-stripe


jack-stripe:
lgtm!

Issue #507, opened 2016-11-25
Perform navigation bar config of STPPaymentMethodsViewController in viewDidLoad to allow for overriding


KieranLaffertySC:
To make it easier to diagnose your issue, please fill out the following:

## Summary

viewDidLoad is the most optimal place to put navigation bar configuration as it allows subclassers to override in the case that they want to customize the way the nav appears (i.e. put a hamburger nav button in top left instead of cancel). Currently to accomplish this you have to put navigation config in viewWillAppear: which isn't ideal.

## iOS version

10.2

## Installation method

Podfile

## SDK version

Latest
## Other information

Anything else you can include that'll make it easier for us to help you!

bdorfman-stripe:
Hi @KieranLaffertySC, 

We just made some tweaks to how back bar button items are configured in a recent commit. Some of the configuration still happens in viewWillAppear (since we want to provide automatic cancel vs back and there isn't really a good place before that where we know where we are in the nav stack), but when the view controller's are at the root of the nav controller, you should be able to change their left bar button items after init and not have them overridden. It's in master now and will be out officially in our next release.

Hope this helps your use case! Let us know if you're still having issues.

PR #521, opened 2016-12-12
Change back button overrideability

bdorfman-stripe:
Change left bar button item logic so its
1. Cancel by default.
2. Overidden to back in viewWillAppear for non-root VC's
3. Not overidden for root VC's

This allows people to change the left bar item after init for views that aren't in root (if it's not going to be presented as a modal and thus does not make sense to have a cancel).

Fixes the use case described in https://github.com/stripe/stripe-ios/issues/507 while still doing automatic back/cancel logic for existing usage.

r? @jack-stripe


jack-stripe:
lgtm! you'll probably need to make a new branch to fix these snapshot tests and rebase off that in order to get a green build.

bdorfman-stripe:
Sounds good. Gonna open a new pr to fix the unrelated snapshot test issues shortly.

bdorfman-stripe:
These snapshot failures are actually because previously the way they were captured, there was no left bar button item in the snapshot (due to how the test fakes a view controller for presentation at the correct size, and doesn't actually run in an app). After these changes, they now have a Cancel button since its set earlier in the lifecycle and not on view did load. I think this is actually what we want, and can just re-record the images? Nothing else seems to be different in them, let me know what you think.

jack-stripe:
cool, sounds good, re-record away

PR #522, opened 2016-12-13
If your key or backendurl is missing, fail in init

bdorfman-stripe:
As discussed in person the other week, added a nicer failure for people who run the example app without filling in their deets

r? @jack-stripe


jack-stripe:
lgtm

PR #523, opened 2016-12-13
Test cleanup

bdorfman-stripe:
* Lock ci_scripts tests to a specific ios version instead of latest.
* Show an error if you try to test in xcode without FBSnapshotTestCase installed
* Remove separate test scheme because it generates some confusion (you can just test through the normal StripeiOS framework scheme)

r? @bg-stripe


Stripe.xcodeproj/project.pbxproj
```diff
@@ -2057,6 +2058,19 @@
...
+			buildActionMask = 2147483647;
+			files = (
+			);
+			inputPaths = (
+			);
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
+			shellScript = "if [ ! -d \"$PROJECT_HOME/Carthage/Build/iOS/FBSnapshotTestCase.framework/\" ]; then\necho \"FBSnapshotTestCase not found. Run \\`carthage bootstrap --platform ios --configuration Release\\` and then run tests again.\"\nexit 0\nfi";
```
jack-stripe: Maybe `--configuration Debug`, so you get symbols and stuff for debugging?

bdorfman-stripe: I did release just because its what the travis ci test script does and I thought it was better to be consistent. Can change though if you want.

PR #525, opened 2016-12-15
Re-enable FauxPas, fix warnings

jack-stripe:
r? @bdorfman-stripe

### DIFF
--- a/FauxPasConfig/main.fauxpas.json
+++ b/FauxPasConfig/main.fauxpas.json
@@ -1,7 +1,7 @@
 // Generated from the Faux Pas GUI
 {
     // Xcode target to check (String)
-    "target": "StripeiOSStatic",
+    "target": "StripeiOS",
     // Xcode build configuration to check (String)
     "buildConfig": "Release",
     // Rules to apply (Array of strings)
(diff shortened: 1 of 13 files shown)


Stripe.xcodeproj/project.pbxproj
```diff
@@ -2353,23 +2353,13 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = 04F39F0D1AEF2AFE005B926E /* StripeiOS Tests-Debug.xcconfig */;
 			buildSettings = {
-				ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = YES;
```
bdorfman-stripe: Did you delete these lines on purpose? It seems like the tests would not compile without them

FauxPasConfig/main.fauxpas.json
```diff
@@ -1,7 +1,7 @@
 // Generated from the Faux Pas GUI
 {
     // Xcode target to check (String)
-    "target": "StripeiOSStatic",
+    "target": "StripeiOS",
```
bdorfman-stripe: Did you change this target intentionally? (from what i remember, you have to actually test both targets as they catch slightly different things? I think our ci script does that)

Stripe.xcodeproj/project.pbxproj
jack-stripe: Yep, these are already present in the .xcconfig, so the setting's unchanged.

FauxPasConfig/main.fauxpas.json
jack-stripe: Yeah - I think FP's checker is better for framework targets in general. The only reason we were formerly defaulting to StripeiOSStatic is that it had a minimum iOS target of 7.0, which was useful for APIAvailability stuff. Now that we only support 8.0, I think this makes more sense as the default thing to check. (And the CI script will still check both)

bdorfman-stripe:
LGTM

Issue #527, opened 2017-01-02
Set barTintColor on navBars for PaymentMethodsViewController and AddCardViewController


idris:
When using the ["simple" integration](https://stripe.com/docs/mobile/ios/standard#theming) the barTintColor of `PaymentMethodsViewController` and `AddCardViewController` is not set, so the navBar ends up being whatever `primaryBackgroundColor` is set to.

Instead, [according to the screenshot in the docs](https://stripe.com/docs/mobile/ios/standard#theming) it should be set to `secondaryBackgroundColor`.

As an alternative, we could add an additional theme value for `navBarBackgroundColor`.

## SDK version

v9.2.0

## Other information

I'd be happy to submit a pull request for this if that helps.

bdorfman-stripe:
Hi idris,

Are you seeing this happen when pushing the `PaymentMethodsViewController` or presenting it? We should be theming the nav bar if you use the present method on `STPPaymentContext` but if you are pushing it, we do not alter the nav bar, as in that case the navigation controller/bar are owned by you and are used elsewhere in the app (whereas for presenting we create our own new navigation controller that we can alter without worrying about unintended side effects elsewhere in your app).

If you are experiencing this issue when presenting, please re-open and we will investigate further, but if it is on push it is working as intended. I can add some clarification to the docs to make this clearer if it is currently confusing.

idris:
@bdorfman-stripe I am calling `paymentContext.presentPaymentMethodsViewController()`, not pushing it.

I don't have permission to re-open the issue.

bdorfman-stripe:
Hmm, can you provide some example code? In the example app you can change the default theming and it seems to be correctly changing the nav bar color, so I am not sure how to reproduce your bug.

idris:
@bdorfman-stripe thanks for your willingness to help.  Here's our code (the relevant parts)

```swift
    let config = STPPaymentConfiguration.shared()
    // ... config initialized here
    let theme = STPTheme.default()
    theme.primaryBackgroundColor = UIColor.blue
    theme.secondaryBackgroundColor = UIColor.green
    // no matter what we set in theme, the navbar is transparent (or colored with `primaryBackgroundColor`)
    let paymentContext = STPPaymentContext(apiAdapter: ConfigurableSTPBackendAPIAdapter.sharedClient, configuration: config, theme: theme)
    var topController = UIApplication.shared.keyWindow?.rootViewController
    paymentContext.hostViewController = topController
[...]

PR #528, opened 2017-01-04
use secondaryBackgroundColor for navigation bar

bg-stripe:
r? @bdorfman-stripe 

This fixes the navigation bar color issue reported in #527 (introduced in #466). 

I haven't been able to reproduce the status bar color issue, however – we're correctly defining `preferredStatusBarStyle` in STPPaymentMethodsViewController. Maybe their app isn't using `UIViewControllerBasedStatusBarAppearance`?

### DIFF
--- a/Stripe/STPAddCardViewController.m
+++ b/Stripe/STPAddCardViewController.m
@@ -224,7 +224,7 @@ typedef NS_ENUM(NSUInteger, STPPaymentCardSection) {
 }
 
 - (UIStatusBarStyle)preferredStatusBarStyle {
-    return ([STPColorUtils colorIsBright:self.theme.primaryBackgroundColor] 
+    return ([STPColorUtils colorIsBright:self.theme.secondaryBackgroundColor]
             ? UIStatusBarStyleDefault
             : UIStatusBarStyleLightContent);
 }
--- a/Example/Stripe iOS Example (Simple)/BrowseProductsViewController.swift
+++ b/Example/Stripe iOS Example (Simple)/BrowseProductsViewController.swift
@@ -43,7 +43,7 @@ class BrowseProductsViewController: UITableViewController {
         super.viewDidAppear(animated)
         let theme = self.settingsVC.settings.theme
         self.view.backgroundColor = theme.primaryBackgroundColor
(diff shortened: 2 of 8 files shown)


Stripe/STPAddCardViewController.m
```diff
@@ -224,7 +224,7 @@ - (void)updateAppearance {
 }
 
 - (UIStatusBarStyle)preferredStatusBarStyle {
-    return ([STPColorUtils colorIsBright:self.theme.primaryBackgroundColor] 
+    return ([STPColorUtils colorIsBright:self.theme.secondaryBackgroundColor]
```
bdorfman-stripe: I think these should actually be `self.navigationBar.stp_theme.secondaryBackgroundColor` maybe? But there might be a timing issue on when that is available/set.

bdorfman-stripe:
Yeah not sure about the status bar, but this looks good for now.

bdorfman-stripe:
Oh, do you need to update all the snapshot tests as well? I'm surprised the tests didnt fail.

Stripe/STPAddCardViewController.m
bg-stripe: ah yeah, we should probably be using
```
STPTheme *navBarTheme = self.navigationController.navigationBar.stp_theme ?: self.theme;
```
i'll update

bdorfman-stripe:
👍

idris:
On this branch, the navbar color works correctly, thanks for the quick fix!

screen shot 2017-01-04 at 10 58 21 am

I see two issues though:
1. status bar is white on white (I was seeing this before as well, so maybe this is something particular to my app?)
2. The "Cancel" button is gone!

bdorfman-stripe:
@idris I think both of those issues are a problem is your configuration. If you run the included example app you can see the status bar changes correctly with theme color and the cancel button is present. For status bar I would make sure you are using the `UIViewControllerBasedStatusBarAppearance` flag in your plist and/or not globally overriding the status bar color somewhere else in your app. For Cancel, I would make sure that you are not changing left bar button item anywhere (or changing its color to white).

If you are still having trouble, lets take it back to the issue thread and not here.

PR #529, opened 2017-01-04
Left bar button item not back bar button item

bdorfman-stripe:
Was using the wrong property in this vc


bg-stripe:
👍 nice catch!

bdorfman-stripe:
@bg-stripe Pushed another commit to this to fix tests if you want to take another quick peek.

bg-stripe:
cool, looks good

PR #530, opened 2017-01-05
Allow filling shipping from billing / billing from shipping

bg-stripe:
r? @bdorfman-stripe 

Note: Ready for review, but not ready to merge – still need to get translations for the new strings.

I've updated snapshot tests so you can see the wrapping behavior – I'll update them again once translations are in.


Stripe.xcodeproj/project.pbxproj
```diff
@@ -452,6 +452,12 @@
...
 		C17A030E1CBEE7A2006C819F /* STPAddressFieldTableViewCell.m in Sources */ = {isa = PBXBuildFile; fileRef = C17A030C1CBEE7A2006C819F /* STPAddressFieldTableViewCell.m */; };
+		C1A06F101E1D8A7F004DCA06 /* STPCard+Private.h in Headers */ = {isa = PBXBuildFile; fileRef = C1A06F0F1E1D8A6E004DCA06 /* STPCard+Private.h */; };
+		C1A06F111E1D8A7F004DCA06 /* STPCard+Private.h in Headers */ = {isa = PBXBuildFile; fileRef = C1A06F0F1E1D8A6E004DCA06 /* STPCard+Private.h */; };
+		C1B582AF1E1C3D750039400C /* STPSectionHeaderView.h in Headers */ = {isa = PBXBuildFile; fileRef = C1B582AD1E1C3D750039400C /* STPSectionHeaderView.h */; };
+		C1B582B01E1C3D750039400C /* STPSectionHeaderView.m in Sources */ = {isa = PBXBuildFile; fileRef = C1B582AE1E1C3D750039400C /* STPSectionHeaderView.m */; };
+		C1B582B11E1C3D8E0039400C /* STPSectionHeaderView.h in Headers */ = {isa = PBXBuildFile; fileRef = C1B582AD1E1C3D750039400C /* STPSectionHeaderView.h */; };
+		C1B582B21E1C3D910039400C /* STPSectionHeaderView.m in Sources */ = {isa = PBXBuildFile; fileRef = C1B582AE1E1C3D750039400C /* STPSectionHeaderView.m */; };
```
bdorfman-stripe: Seem to have duplicated some lines here, bad Xcode merge?

Stripe/PublicHeaders/STPShippingAddressViewController.h
```diff
@@ -38,13 +38,15 @@ NS_ASSUME_NONNULL_BEGIN
...
  *  @param shippingAddress           If set, the shipping address view controller will be pre-filled with this address. @see STPAddress
+ *  @param billingAddress            If set, the user will have the option to autofill with this address. @see STPAddress
  *  @param selectedShippingMethod    If set, the shipping methods view controller will use this method as the selected shipping method. If `selectedShippingMethod` is nil, the first shipping method in the array of methods returned by your delegate will be selected.
  *  @param prefilledInformation      If set, the shipping address view controller will be pre-filled with this information. @see STPUserInformation
  */
 - (instancetype)initWithConfiguration:(STPPaymentConfiguration *)configuration
                                 theme:(STPTheme *)theme
                              currency:(nullable NSString *)currency
                       shippingAddress:(nullable STPAddress *)shippingAddress
+                       billingAddress:(nullable STPAddress *)billingAddress
```
bdorfman-stripe: Probably want to add a new init instead of changing this one so its not a breaking api change (maybe deprecate the old one if you want)

bg-stripe: ah, right. What do you think about instead exposing billingAddress as a property?

Stripe/STPSectionHeaderView.h
```diff
@@ -0,0 +1,18 @@
...
+//  Stripe
+//
+//  Created by Ben Guo on 1/3/17.
+//  Copyright © 2017 Stripe, Inc. All rights reserved.
+//
+
+#import <UIKit/UIKit.h>
+#import "STPTheme.h"
+
+@interface STPSectionHeaderView : UIView
```
bdorfman-stripe: Should this be `UITableViewHeaderFooterView`?

Stripe/STPShippingAddressViewController.m
```diff
@@ -366,11 +390,29 @@ - (UIView *)tableView:(__unused UITableView *)tableView viewForFooterInSection:(
...
 }
 
 - (UIView *)tableView:(__unused UITableView *)tableView viewForHeaderInSection:(__unused NSInteger)section {
-    return [UIView new];
+    [self updateAddressHeaderView];
+    return self.addressHeaderView;
+}
+
+- (void)updateAddressHeaderView {
+    BOOL buttonVisible = (!self.addressViewModel.isValid && self.billingAddress != nil && !self.hasUsedBillingAddress);
```
bdorfman-stripe: Possibly changing the height of the header based on billing address usage and then not always triggering a recalc of the height of the table may lead to some layout weirdness I think. You probably need a `[tableView beginUpdates]; [tableView endUpdates];` after every time the height may change or something so the table actually has consistent heights, or some other height caching method.

Stripe/STPSectionHeaderView.h
bg-stripe: hm, is there any benefit to using `UITableViewHeaderFooterView` in this case? Seems like its main feature is reuse, which we wouldn't be using

bdorfman-stripe: I thought the non-reusable `viewForHeaderInSection` method was deprecated but I guess not.

Stripe/PublicHeaders/STPShippingAddressViewController.h
bdorfman-stripe: Maybe. Would it be better to put this under `STPPrefilledInformation` instead?

bg-stripe: ah yeah, that's a good call. I'll update to use `STPPrefilledInformation` here and `STPAddCardVC`.

Stripe.xcodeproj/project.pbxproj
bg-stripe: Ah, probably – I'll remove/re-add the files.

bg-stripe: fixed in 6fb0925

Stripe/PublicHeaders/STPShippingAddressViewController.h
bg-stripe: (see 611b777)

bg-stripe:
@bdorfman-stripe finally got all the translations reviewed and got around to rebasing this branch – re-r?

Stripe/STPAddCardViewController.m
```diff
@@ -644,4 +676,13 @@ - (UIView *)tableView:(__unused UITableView *)tableView viewForFooterInSection:(
     }
 }
 
+- (void)useShippingAddress:(UIButton *)sender {
+    self.addressViewModel.address = self.shippingAddress;
+    self.hasUsedShippingAddress = YES;
+    [[self firstEmptyField] becomeFirstResponder];
+    [UIView animateWithDuration:0.2f animations:^{
+        sender.alpha = 0;
+    }];
```
bdorfman-stripe: Since this code changes the sizeThatFits / heightForHeaderInSection value, you need to do `[tableView beginUpdates]; [tableView endUpdates]` so that the table view will immediately recalc the height and not get into a bad state later.

Also I think changing the alpha of the sender here and then relying on that value for calculations in the actual class is a little iffy. It would be better to have the hiding logic inside the header view class itself and a property to change whether it is showing the button or not.

bg-stripe: 👍  – updated in 5a8ba77

Stripe/STPSectionHeaderView.m
```diff
@@ -66,9 +66,14 @@ - (void)setTitle:(NSString *)title {
     [self setNeedsLayout];
 }
 
+- (void)setButtonHidden:(BOOL)buttonHidden {
+    _buttonHidden = buttonHidden;
+    self.button.hidden = buttonHidden;
```
bdorfman-stripe: Will this actually animate? I thought `hidden` wasn't animateable.

bg-stripe: ah yup, thanks for catching that – updated in 4bd20ba

bdorfman-stripe:
👍

PR #534, opened 2017-01-11
View Controller refactor

bdorfman-stripe:
Combines shared VC logic from our screens into base classes: STPCoreViewController, STPCoreScrollViewController, and STPTableViewController. Each inherits from the previous one and does shared logic for that type of VC that was previously copy and pasted around. There's probably even more condensing that can be done of shared logic later.


jack-stripe:
Small note, I think it's good to keep large style changes like header organization out of changes like this as it adds a bunch of churn to the PR and makes it harder to review.

Stripe/STPCoreScrollViewController+Private.h
```diff
@@ -0,0 +1,39 @@
...
+ */
+@property (nonatomic, nullable, readonly) UIScrollView *scrollView;
+
+/**
+ This method is used by the base implementation to create the object
+ backing the `scrollView` property. Subclasses can override to change the
+ type of the scroll view (eg UITableView or UICollectionView instead of
+ UIScrollView).
+ */
+- (UIScrollView *)createScrollView;
```
jack-stripe: This can probably be a class method

Stripe/STPPaymentMethodsInternalViewController.m
```diff
@@ -53,20 +52,14 @@ - (instancetype)initWithConfiguration:(STPPaymentConfiguration *)configuration
...
-    [tableView registerClass:[STPPaymentMethodTableViewCell class] forCellReuseIdentifier:STPPaymentMethodCellReuseIdentifier];
-    tableView.sectionHeaderHeight = 30;
-    tableView.separatorStyle = UITableViewCellSeparatorStyleNone; // use fake separator views
-    tableView.separatorInset = UIEdgeInsetsMake(0, 18, 0, 0);
-    self.tableView = tableView;
-    [self.view addSubview:tableView];
+- (void)createAndSetupViews {
+    [super createAndSetupViews];
+
+    self.tableView.allowsMultipleSelectionDuringEditing = NO;
```
jack-stripe: I think this and the `separatorInset` call can get moved to STPCoreTableVC?

Stripe/STPCoreScrollViewController+Private.h
bdorfman-stripe: Yeah I was leaving it as class for potential future usage that may need it to not be a class method (like init'ing a grouped vs plain style, or init'ing a collection view with a specific layout. Can swap if you want though.

bdorfman-stripe:
Ah yeah sorry, i was going to put the header sorting in a separate pr but i accidentally pushed to this branch and then  ¯\_(ツ)_/¯

Stripe/STPPaymentMethodsInternalViewController.m
bdorfman-stripe: Nowhere else makes these changes so I wasn't sure? Also `allowsMultipleSelectionDuringEditing` is NO by default anyway.

PR #536, opened 2017-01-17
Check month validity in validationStateForExpirationYear

### LINKED ISSUE #535: Query: Should validation of expiry month be able to exceed 12 in +validationStateForExpirationYear:inMonth:
Hi,

I have a query about the workings of the `STPCardValidator` when checking the card expiry date. In particular, this method:

```
/**
 *  Validates an expiration year, passed as a string representing the final 2 digits of the year. This considers the period between the current year until 2099 as valid times. An example valid value would be "16" (assuming the current year, as determined by [NSDate date], is 2015). Will return STPCardValidationStateInvalid for a month/year combination that is earlier than the current date (i.e. @"15" and @"04" in October 2015. Example invalid values are "00", "a", and "13". Any 1-digit string will return STPCardValidationStateIncomplete.
 *
 *  @param expirationYear A string representing a 2-digit expiration year for a payment card.
 *  @param expirationMonth A string representing a 2-digit expiration month for a payment card. See -validationStateForExpirationMonth for the desired formatting of this string.
 *
 *  @return STPCardValidationStateValid if the year is valid, STPCardValidationStateInvalid if the year is invalid, or STPCardValidationStateIncomplete if the year is a substring of a valid year (e.g. @"1" or @"2").
 */
+ (STPCardValidationState)validationStateForExpirationYear:(NSString *)expirationYear
                                                   inMonth:(NSString *)expirationMonth;
```

If you pass `@"13"` into the `expirationMonth` property along with a year greater than the current year, this method will return `STPCardValidationStateValid`. Is that desired functionality? 

The method implementation itself makes no attempt to check for this so that's either by design or a bug as the comment says to "See -validationStateForExpirationMonth for the desired formatting of this string."

`+validationStateForExpirationMonth:` then suggests from it's comment that a number greater than 12 is invalid and it's implementation checks for this.

In my opinion one of two things should happen (but i don't know the full story so please correct me if i'm wrong):

1. I raise a bug that `+validationStateForExpirationYear:inMonth:` doesn't follow the validation rules that `+validationStateForExpirationMonth:` enforces.
2. I raise a bug that the documentation for `+validationStateForExpirationYear:inMonth:` is incorrect and it's updated to reflect what is actually checked.

I look forward to hearing what other people think.

bg-stripe:
Fixes #535 

r? @bdorfman-stripe 

As noted in the issue, this function currently assumes the month string is valid. This is fine in our own usage, but validating the month too seems reasonable.


bdorfman-stripe:
Might want to add to the header documentation that it also validates the month for the return value now. Otherwise 👍

bg-stripe:
👍  updated the header doc

PR #539, opened 2017-01-24
[in progress] Add Source model and createSource

bg-stripe:
Note: the base branch of this PR `bg-sources`, not master

r? @bdorfman-stripe


Stripe/STPSourceParams.m
```diff
@@ -0,0 +1,43 @@
...
+
+#pragma mark - STPFormEncodable
+
++ (NSString *)rootObjectName {
+    return nil;
+}
+
++ (NSDictionary *)propertyNamesToFormFieldNamesMapping {
+    return @{
+             @"type": @"type",
```
bdorfman-stripe: Would be better to do NSStringFromSelector here so we get some sort of compile type checking on these.

bdorfman-stripe:
Looks pretty good. I think we might also want to have helper creators on STPSourceParams for the different APMs (eg `createSofortParams:`) so that it is more clear to people which params/settings are required for those payment types.

Stripe/STPSourceParams.m
bg-stripe: 👍

bg-stripe:
ah yup, good idea – I'll add some helpers for all the launched APMs

bg-stripe:
@bdorfman-stripe added APM constructors for `STPSourceParams`, along with functional tests for each one. re-r?

Stripe/PublicHeaders/STPSourceParams.h
```diff
@@ -0,0 +1,201 @@
...
+ *  The authentication flow of the source to create. `flow` may be "redirect",
+ *  "receiver", "verification", or "none". It is generally inferred unless a type
+ *  supports multiple flows.
+ */
+@property (nonatomic, copy, nullable) NSString *flow;
+
+/**
+ *  A set of key/value pairs that you can attach to a source object.
+ */
+@property (nonatomic, copy, nullable) NSDictionary *metadata;
```
bdorfman-stripe: Should we make these Dictionary properties into nonnull mutable dictionaries? Would that be easier for people trying to fill in the data / stop accidental clobbering of already filled in keys? Or is that over complicating the matter.

bg-stripe: I'd prefer keeping them nullable – if we made them nonnull/mutable dictionaries, I think our serialization logic (`STPFormEncoder`) would add empty dictionaries, rather than leaving out the key pairs.

bdorfman-stripe:
👍

PR #540, opened 2017-01-25
[sources] Add retrieveSource

bg-stripe:
Note: the base branch of this is `bg-sources`, not master

r? @bdorfman-stripe


bdorfman-stripe:
Is the only difference in the request classes the setting of POST vs GET? Would it be easier to make these one class or method with just a different setting or parameter? Also not 100% clear on why these are their own classes instead of just helper methods inside STPAPIClient or something.

bg-stripe:
It's the setting of POST vs GET, as well as the way they handle params (`STPAPIPostRequest` takes `postData:(NSData *)`, while `STPAPIGetRequest` takes `parameters:(NSDictionary *)`). I'll refactor this to be `STPAPIRequest` with get/post methods, though, that seems better.

I think `STPAPIPostRequest` is a separate class because you get the benefit of generics? (it's covariant with `ResponseType`)

bdorfman-stripe:
Ah that all makes sense. Thanks for the clarification!

bg-stripe:
@bdorfman-stripe refactored this so there's one request class – re-r?

bdorfman-stripe:
Looks good.

PR #544, opened 2017-01-27
[sources] Add source polling

bg-stripe:
r? @bdorfman-stripe 

I've tested some of the main edge cases manually – let me know if I've missed anything:
- polling stops when `source.status` changes
- polling stops when `STPAPIClient` is dealloced
- polling stops when `stopPolling` is called
- polling interval increases on 500, resets on 200
- polling stops with error when there's no network

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: objective-c
 osx_image: xcode8.1
-branches:
-  only: 
-    - master
+# branches:
+#   only:
+#     - master
 env:
   matrix:
   - TEST_TYPE=installation_manual
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: objective-c
 osx_image: xcode8.1
-# branches:
-#   only:
-#     - master
+branches:
+  only: 
+    - master
 env:
   matrix:
   - TEST_TYPE=installation_manual
--- a/Stripe/PublicHeaders/STPAPIClient.h
+++ b/Stripe/PublicHeaders/STPAPIClient.h
@@ -160,6 +160,33 @@ static NSString *const STPSDKVersion = @"9.4.0";
  */
 - (void)retrieveSourceWithId:(NSString *)identifier clientSecret:(NSString *)secret completion:(STPSourceCompletionBlock)completion;
 
+/**
+ *  Starts polling the Source object with the given ID. For payment methods that require
+ *  additional customer action (e.g. authorizing a payment with their bank), polling
+ *  allows you to determine if the action was successful. The provided callback will be
+ *  called as soon as the source is retrieved, and every time its status is updated.
+ *  Polling will end if the source's status is one of `chargeable`, `canceled`, or
+ *  `consumed`. Polling will also end if more than 5 minutes elapse, or if more than
+ *  5 consecutive errors occur. If polling ends due to an error, the callback will be
+ *  fired with the latest retrieved source and the error.
+ *
+ *  Note that if a poll is already running for a source, subsequent calls to `startPolling`
+ *  with the same source ID will do nothing.
+ *
+ *  @param identifier  The identifier of the source to be retrieved. Cannot be nil.
+ *  @param secret      The client secret of the source. Cannot be nil.
+ *  @param completion  The callback to run with the returned Source object, or an error.
+ */
+- (void)startPollingSourceWithId:(NSString *)identifier clientSecret:(NSString *)secret completion:(STPSourceCompletionBlock)completion NS_EXTENSION_UNAVAILABLE("Source polling is not available in extensions");;
+
+/**
+ *  Stops polling the Source object with the given ID. Note that the completion block passed to
+ *  `startPolling` will not be fired when `stopPolling` is called.
+ *
+ *  @param identifier  The identifier of the source to be retrieved. Cannot be nil.
+ */
+- (void)stopPollingSourceWithId:(NSString *)identifier NS_EXTENSION_UNAVAILABLE("Source polling is not available in extensions");;
+
 @end
 
 #pragma mark - Deprecated Methods
--- a/Stripe/STPAPIClient.m
+++ b/Stripe/STPAPIClient.m
@@ -19,6 +19,7 @@
 #import "STPPaymentConfiguration.h"
 #import "STPSource.h"
 #import "STPSourceParams.h"
+#import "STPSourcePoller.h"
 #import "STPToken.h"
 
 #if __has_include("Fabric.h")
@@ -61,6 +62,8 @@ static NSString *const stripeAPIVersion = @"2015-10-12";
 #endif
 @property (nonatomic, readwrite) NSURL *apiURL;
 @property (nonatomic, readwrite) NSURLSession *urlSession;
+@property (nonatomic, readwrite) NSMutableDictionary<NSString *,NSObject *>*sourcePollers;
+@property (nonatomic, readwrite) dispatch_queue_t sourcePollersQueue;
 @end
 
 @implementation STPAPIClient
@@ -103,7 +106,8 @@ static NSString *const stripeAPIVersion = @"2015-10-12";
                                                        @"Authorization": auth,
                                                        };
         _urlSession = [NSURLSession sessionWithConfiguration:sessionConfiguration];
-
+        _sourcePollers = [NSMutableDictionary dictionary];
+        _sourcePollersQueue = dispatch_queue_create("com.stripe.sourcepollers", DISPATCH_QUEUE_SERIAL);
     }
     return self;
 }
@@ -312,15 +316,37 @@ static NSString *const stripeAPIVersion = @"2015-10-12";
     NSCAssert(identifier != nil, @"'identifier' is required to create a source");
     NSCAssert(secret != nil, @"'secret' is required to create a source");
     NSCAssert(completion != nil, @"'completion' is required to use the source that is created");
+    [self retrieveSourceWithId:identifier clientSecret:secret responseCompletion:^(STPSource * object, __unused NSHTTPURLResponse *response, NSError *error) {
+        completion(object, error);
+    }];
+}
+
+- (NSURLSessionDataTask *)retrieveSourceWithId:(NSString *)identifier clientSecret:(NSString *)secret responseCompletion:(STPAPIResponseBlock)completion {
     NSString *endpoint = [NSString stringWithFormat:@"%@/%@", sourcesEndpoint, identifier];
     NSDictionary *parameters = @{@"client_secret": secret};
-    [STPAPIRequest<STPSource *> getWithAPIClient:self
-                                        endpoint:endpoint
-                                      parameters:parameters
-                                      serializer:[STPSource new]
-                                      completion:^(STPSource *object, __unused NSHTTPURLResponse *response, NSError *error) {
-                                          completion(object, error);
-                                      }];
+    return [STPAPIRequest<STPSource *> getWithAPIClient:self
+                                               endpoint:endpoint
+                                             parameters:parameters
+                                             serializer:[STPSource new]
+                                             completion:completion];
+}
+
+- (void)startPollingSourceWithId:(NSString *)identifier clientSecret:(NSString *)secret completion:(STPSourceCompletionBlock)completion {
+    [self stopPollingSourceWithId:identifier];
+    STPSourcePoller *poller = [[STPSourcePoller alloc] initWithAPIClient:self clientSecret:secret sourceID:identifier completion:completion];
+    dispatch_async(self.sourcePollersQueue, ^{
+        self.sourcePollers[identifier] = poller;
+    });
+}
+
+- (void)stopPollingSourceWithId:(NSString *)identifier {
+    dispatch_async(self.sourcePollersQueue, ^{
+        STPSourcePoller *poller = (STPSourcePoller *)self.sourcePollers[identifier];
+        if (poller) {
+            [poller stopPolling];
+            self.sourcePollers[identifier] = nil;
+        }
+    });
 }
 
 @end
--- a/Stripe/STPSourcePoller.m
+++ b/Stripe/STPSourcePoller.m
@@ -1,0 +1,209 @@
+//
+//  STPSourcePoller.m
+//  Stripe
+//
+//  Created by Ben Guo on 1/26/17.
+//  Copyright © 2017 Stripe, Inc. All rights reserved.
+//
+
+#import "STPAPIClient+Private.h"
+#import "STPAPIRequest.h"
+#import "STPSource.h"
+#import "STPSourcePoller.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+static NSTimeInterval const DefaultPollInterval = 1.5;
+static NSTimeInterval const MaxPollInterval = 24;
+// Stop polling after 5 minutes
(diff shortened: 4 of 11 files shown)


Stripe/STPSourcePoller.m
```diff
@@ -0,0 +1,108 @@
...
+    }
+    return self;
+}
+
+- (void)pollAfter:(NSTimeInterval)interval {
+    if (!self.apiClient) {
+        [self stopPolling];
+        return;
+    }
+    dispatch_time_t time = dispatch_time(DISPATCH_TIME_NOW, (int64_t)(interval * NSEC_PER_SEC));
```
bdorfman-stripe: Would it be better to use NSTimer here so we can stop immediately when a stopPolling is called? You can't cancel a queued dispatch.

Stripe/STPSourcePoller.m
```diff
@@ -0,0 +1,108 @@
...
+
+- (BOOL)shouldContinuePollingSource:(nullable STPSource *)source {
+    if (!source) {
+        return NO;
+    }
+    return source.status == STPSourceStatusPending;
+}
+
+- (void)stopPolling {
+    self.shouldStopPolling = YES;
```
bdorfman-stripe: This should probably also cancel currently in flight requests?

bdorfman-stripe:
I think you should probably also stop when backgrounding or become inactive, and restart on foreground/become active.

Stripe/STPSourcePoller.m
```diff
@@ -0,0 +1,108 @@
...
+}
+
+- (void)continueWithSource:(STPSource *)source
+                  response:(NSHTTPURLResponse *)response
+                     error:(NSError *)error {
+    if (self.shouldStopPolling) {
+        return;
+    }
+    NSUInteger status = response.statusCode;
+    if (!response || (status >= 400 && status < 500)) {
```
bdorfman-stripe: Where does this stop if there is a network error? Possibly i'm just misreading this.

Should instead this basically just stop/fallback on any non-200? Or is this behavior designed to match Stripe.js in some way?

Stripe/STPSourcePoller.m
bg-stripe: 👍  yeah, it'd definitely be better to use `NSTimer` and cancel, I'll update

Stripe/STPSourcePoller.m
bg-stripe: 👍  yeah, good point – I'll update so that we also cancel the `NSURLSessionTask`

Stripe/STPSourcePoller.m
bg-stripe: This method stops in branches that don't call `pollAfter`. If there's a network error, `response` will be nil, so this branch captures that case. 

The way we treat different error codes in this implementation is from stripe.js [0], and seems fairly reasonable to me, but open to tweaking it:
- retry with exponential backoff on 500 to reduce load
- retry with default poll interval on other non-200 status codes
- give up if there's a network error or a 4xx

[0] https://github.com/stripe-internal/stripe-js/blob/f3a4ac25ae9537c42492370a7d51033f844c2c61/src/index.coffee#L673

bdorfman-stripe: Hmm ok. I think putting this as a comment in the code would be helpful. I also think its very hard to read this branching if statement right now, especially trying to figure out which branches continue polling, which dont, and which fire completion blocks. Not sure if there is some way you can rewrite it to be clearer.

Stripe/STPSourcePoller.m
```diff
@@ -0,0 +1,108 @@
...
+        self.completion(nil, error);
```
bdorfman-stripe: Should this still return the source to the completion block, even if there was an error and we havent fetched a new source object?

bg-stripe: yeah, that's a good idea – I'll update

Stripe/STPSourcePoller.m
bg-stripe: totally – I'll add more comments, and see if I can make the branching clearer

bg-stripe:
@bdorfman-stripe re-r?
- improved canceling behavior with NSTimer/NSURLSessionDataTask
- added `MaxRetries` and `Timeout`
- stop/restart on app lifecycle events
- retry on connectivity errors

Stripe/STPAPIClient.m
```diff
@@ -312,15 +314,33 @@ - (void)retrieveSourceWithId:(NSString *)identifier clientSecret:(NSString *)sec
...
-                                      }];
+    return [STPAPIRequest<STPSource *> getWithAPIClient:self
+                                               endpoint:endpoint
+                                             parameters:parameters
+                                             serializer:[STPSource new]
+                                             completion:completion];
+}
+
+- (void)startPollingSourceWithId:(NSString *)identifier clientSecret:(NSString *)secret completion:(STPSourceCompletionBlock)completion {
+    [self stopPollingSourceWithId:identifier];
```
bdorfman-stripe: Header doc should make it clear that you can only have 1 poll per source at a time (or change this to be able to fire multiple completion blocks from one polling, although we might have some threading issues if we make it more complicated like this).

Stripe/PublicHeaders/STPAPIClient.h
```diff
@@ -160,6 +160,27 @@ static NSString *const STPSDKVersion = @"9.3.0";
...
+ *  @param completion  The callback to run with the returned Source object, or an error.
+ */
+- (void)startPollingSourceWithId:(NSString *)identifier clientSecret:(NSString *)secret completion:(STPSourceCompletionBlock)completion;
+
+/**
+ *  Stops polling the Source object with the given ID.
+ *
+ *  @param identifier  The identifier of the source to be retrieved. Cannot be nil.
+ */
+- (void)stopPollingSourceWithId:(NSString *)identifier;
```
bdorfman-stripe: Add a note that manually stopping will not fire completion block for that poll?

Stripe/STPAPIClient.m
```diff
@@ -312,15 +314,33 @@ - (void)retrieveSourceWithId:(NSString *)identifier clientSecret:(NSString *)sec
...
+    STPSourcePoller *poller = [[STPSourcePoller alloc] initWithAPIClient:self clientSecret:secret sourceID:identifier completion:completion];
+    self.sourcePollers[identifier] = poller;
```
bdorfman-stripe: I'm somewhat concerned about the thread safety issues inherent in mutating this dictionary. Maybe should be using a private synchronous queue for all polling actions?

Stripe/STPSourcePoller.m
```diff
@@ -0,0 +1,170 @@
...
+            if (!self.latestSource || source.status != self.latestSource.status) {
+                self.completion(source, nil);
+            }
+            self.latestSource = source;
+            if ([self shouldContinuePollingSource:source]) {
+                [self pollAfter:self.pollInterval];
+            } else {
+                [self stopPolling];
+            }
+            self.retryCount = 0;
```
bdorfman-stripe: A 200 should reset poll interval i think?

Stripe/STPSourcePoller.m
```diff
@@ -0,0 +1,170 @@
...
+        } else {
+            // Backoff on 500, otherwise reset poll interval
+            if (status == 500) {
```
bdorfman-stripe: Why only backoff on 500?

Stripe/STPSourcePoller.m
```diff
@@ -0,0 +1,170 @@
...
+                self.pollInterval = DefaultPollInterval;
+            }
+            [self pollAfter:self.pollInterval];
+            self.retryCount++;
+        }
+    } else {
+        // Retry if there's a connectivity error, otherwise stop polling
+        if (error && (error.code == kCFURLErrorNotConnectedToInternet ||
+                      error.code == kCFURLErrorNetworkConnectionLost)) {
+            [self pollAfter:self.pollInterval];
```
bdorfman-stripe: Retrycount needs to go up here. Perhaps should move the counting to inside `pollAfter:` to make sure it always gets incremented correctly.

Stripe/STPSourcePoller.m
```diff
@@ -0,0 +1,170 @@
...
+            [self stopPolling];
+        }
+    }
+}
+
+- (BOOL)shouldContinuePollingSource:(nullable STPSource *)source {
+    if (!source) {
+        return NO;
+    }
+    return source.status == STPSourceStatusPending;
```
bdorfman-stripe: Should we short circuit our polling logic earlier if this returns NO? i.e. don't even fire the first request.

bdorfman-stripe:
Comments in line. Also some other thoughts:

Should we take out a background task instead of stopping immediately on backgrounding? I'm slightly worried about the case where the user backgrounds immediately, and then doesn't return until after 5 minutes have expired and the poll fails without ever successfully making a request. We might want to put some guards in to make sure the timeout doesn't kick in until at least 1 network transaction has completed (whether or not it succeeded or failed).

Also: Probably should document that there is a retry/timeout more explicitly in the header comments.

Stripe/STPSourcePoller.m
bg-stripe: I think backoff is only necessary if there's a server error – 5xx would probably be better. Do you think we should backoff in other cases too?

Stripe/STPSourcePoller.m
bg-stripe: yep, it gets reset on line 116 above

Stripe/STPSourcePoller.m
bg-stripe: This was actually intentional – I didn't think it was necessary to throttle if there are connectivity issues, since the requests aren't going anywhere.

Stripe/STPSourcePoller.m
bg-stripe: I think this stops as soon as it can – we start polling with a source ID, so we have to make at least one request to get the status. If the source isn't `pending`, then the poller stops after one request.

Stripe/PublicHeaders/STPAPIClient.h
bg-stripe: 👍 . I think it's possible for the completion block to get fired with a cancelled error if there's an in-flight request – I should make sure we're filtering those out.

Stripe/STPAPIClient.m
bg-stripe: 👍

Stripe/STPAPIClient.m
bg-stripe: 👍  will do, good call

Stripe/STPSourcePoller.m
bg-stripe: Actually, thinking about this some more, it's probably simpler to just backoff always here – I'll update.

Stripe/STPSourcePoller.m
bg-stripe: (I'll move resetting retryCount up next to resetting the poll interval so it's clearer)

bdorfman-stripe: Ah sorry missed it, my bad

Stripe/STPSourcePoller.m
bdorfman-stripe: I think we should maybe still stop after 5 times or whatever even if there's a network issue. Otherwise we'll waste cpu just infinitely retrying if someone tries to start a poll while their internet doesn't work?

Actually thinking more on this, the "correct" way to do this is probably to include Reachability and then just stop and resume on network availability, the same way we do on background/foreground.

Stripe/STPSourcePoller.m
bdorfman-stripe: Ah ok, for some reason I thought we already had a full source object when we start, but that makes sense, my mistake.

Stripe/STPSourcePoller.m
bg-stripe: That's fair, it seems reasonable to throttle even if there's a network issue. 

It would be better to stop/resume on network availability, but not sure if it's worth including Reachability for now – it doesn't seem as necessary as handling background/foreground appropriately, and it's probably fine to just timeout?

bdorfman-stripe: Yeah I think it's fine. Presumably the integrator will be dealing with a no network condition anyway.

bg-stripe:
> Should we take out a background task instead of stopping immediately on backgrounding? I'm slightly worried about the case where the user backgrounds immediately, and then doesn't return until after 5 minutes have expired and the poll fails without ever successfully making a request. We might want to put some guards in to make sure the timeout doesn't kick in until at least 1 network transaction has completed (whether or not it succeeded or failed).

Yeah, that's a good suggestion – it'd be better to take out a background task for each request, and not cancel in-flight requests on backgrounding.  I'll also add a guard against the poller not firing any requests.

bg-stripe:
@bdorfman-stripe ready for another review!

.travis.yml
```diff
@@ -1,8 +1,8 @@
 language: objective-c
 osx_image: xcode8.1
-branches:
```
bdorfman-stripe: Need to revert this travis changes before merging to master probagly

Stripe/STPSourcePoller.m
```diff
@@ -0,0 +1,209 @@
...
+                               selector:@selector(restartPolling)
+                                   name:UIApplicationWillEnterForegroundNotification
+                                 object:nil];
+        [notificationCenter addObserver:self
+                               selector:@selector(pausePolling)
+                                   name:UIApplicationWillResignActiveNotification
+                                 object:nil];
+        [notificationCenter addObserver:self
+                               selector:@selector(pausePolling)
+                                   name:UIApplicationDidEnterBackgroundNotification
```
bdorfman-stripe: Probably need to remove/change this notification behavior now that we take out a background task.

bg-stripe: I've modified it since the last review, and I think it has the correct behavior? When the app enters the background, we pause polling – if there's a request in-flight, it'll finish in the background, but no more requests will be made until the app becomes active.

bdorfman-stripe: You're right sorry, I misinterpreted what `pausePolling` was doing.

Stripe/STPSourcePoller.m
```diff
@@ -0,0 +1,209 @@
...
+    return source.status == STPSourceStatusPending;
+}
+
+- (void)restartPolling {
+    if (self.pollingStopped) {
+        return;
+    }
+    self.pollingPaused = NO;
+    if (!self.timer && !self.dataTask) {
+        [self pollAfter:0 lastError:nil];
```
bdorfman-stripe: Should we take into account the pollInterval here for backoff purposes? Eg if you get a failure in the background. Not sure this edge case is worth too much extra logic, but maybe?

bdorfman-stripe:
I think this looks good. 👍  I'm ok to merge and/or send to Jack for a 2nd look.

Stripe/STPSourcePoller.m
bg-stripe: Good point – I think it's fine to keep it at 0 though, since backoff logic will kick in if the request continues erroring

bg-stripe:
cool. Once #551 gets into master I'll change the base branch of this and ask jack for another review.

bg-stripe:
@jack-stripe can you take a look at this to sanity check ?

PR #545, opened 2017-01-27
Use CVV as placeholder for Amex cards

bg-stripe:
r? @bdorfman-stripe 

- If you use set a custom placeholder, the field will always use that
- I've also updated the number placeholder to be "4242..."


Stripe/STPPaymentCardTextField.m
```diff
@@ -120,8 +120,8 @@ - (void)commonInit {
     cvcField.tag = STPCardFieldTypeCVC;
     cvcField.alpha = 0;
     self.cvcField = cvcField;
-    self.cvcPlaceholder = @"CVC";
-    self.cvcField.accessibilityLabel = self.cvcPlaceholder;
+    self.cvcPlaceholder = nil;
+    self.cvcField.accessibilityLabel = @"CVC";
```
bdorfman-stripe: Why hardcoding this here instead of changing it to match based on the card?

bg-stripe:
@bdorfman-stripe updated so that we set the accessibility label's text, and also updated snapshot tests. re-r?

Stripe/STPPaymentCardTextField.m
```diff
@@ -780,6 +783,23 @@ - (void)updateImageForFieldType:(STPCardFieldType)fieldType {
...
+    }
+}
+
+- (void)updateCVCPlaceholder {
+    if (self.cvcPlaceholder) {
+        self.cvcField.placeholder = self.cvcPlaceholder;
+    } else {
+        self.cvcPlaceholder = [self defaultCVCPlaceholder];
+    }
+    self.cvcField.accessibilityLabel = [self defaultCVCPlaceholder];
```
bdorfman-stripe: Why are these different? What is `self.cvcPlaceholder`?

Also do CVV and CVC need to be localized?

bg-stripe: `self.cvcPlaceholder` is a property that we expose that lets you set a custom CVC placeholder – if it's set, we use that, otherwise we use `CVC` or `CVV`.

We should probably localize `CVC` and `MM/YY`, but I think we should leave that out of the scope of this PR – I'll file a task to keep track of it.

bdorfman-stripe: Should we be setting the accessibility label to your custom placeholder as well? Or are you intentionally always setting it to the default?

bg-stripe: I initially thought it'd be fine to just use the default for the accessibility label, but I think you're right – updated to set the accessibility label with the custom placeholder too.

bdorfman-stripe:
👍

jack-stripe:
Why change the default placeholder? Most end users have no idea what 4242 means.

bdorfman-stripe:
We use 42 everywhere else as example/placeholder and were specifically asked to change it for consistency with other Stripe products.

jack-stripe:
👍

PR #548, opened 2017-02-01
Update remember me analytics

bdorfman-stripe:
Track more possible states for remember me usage so we can more accurately determine how people are using it.

### DIFF
--- a/Stripe/STPAddCardViewController.m
+++ b/Stripe/STPAddCardViewController.m
@@ -23,6 +23,7 @@
 #import "STPObscuredCardView.h"
 #import "STPPaymentActivityIndicatorView.h"
 #import "STPPaymentCardTextField.h"
+#import "STPPaymentConfiguration+Private.h"
 #import "STPPhoneNumberValidator.h"
 #import "STPRememberMeEmailCell.h"
 #import "STPRememberMePaymentCell.h"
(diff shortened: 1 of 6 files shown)


Stripe/STPAddCardViewController.m
```diff
@@ -328,9 +330,40 @@ - (void)nextPressed:(__unused id)sender {
...
+                 1. User saw the toggle, selected it, and entered valid email/phone details
+                 2. User saw the toggle, and did not select it or entered invalid details
+                 3. User did not see the toggle because the developer disabled it
+                 4. User did not see the toggle because the developer enabled it but they were otherwise ineligible
+                    (eg payment context flow disables if you already have >0 payment methods,
+                     or you had a cancelled SMS fill)
+                 5. User saw the email field and added card by SMS code from a previous Remember Me save
+                 */
+
+                STPAddCardRememberMeUsage rememberMeSelected;
```
bg-stripe: nit: might be less confusing to call this `rememberMeUsage`, since it's no longer a bool

bg-stripe:
👍

Stripe/STPAddCardViewController.m
bdorfman-stripe: 👍 Will change before merging.

PR #551, opened 2017-02-07
Add ability to create and retrieve Source objects

bg-stripe:
r? @jack-stripe 
cc @bdorfman-stripe 

Sorry this is such a large PR – it's mostly new model objects. 

- Updates `STPAPIPostRequest->STPAPIRequest` to support both GET and POST requests
- Adds two methods to `STPAPIClient`, `createSourceWithParams` and `retrieveSourceWithId`
- Adds `STPSourceParams` for creating sources, along with constructors for various payment methods
- Adds `STPSource` for fetched sources, along with classes for inner objects, e.g. `STPSourceRedirect`
- Renames the `<STPSource>` protocol to `STPSourceProtocol` 
   - The naming here is potentially confusing – I think it's better for the source object to be named `STPSource`, but not sure if `STPSourceProtocol` is the best name for the shim protocol.


bdorfman-stripe:
FWIW I have already reviewed this as a series of smaller PRs into the bg-sources branch, but it would be good to have another API review from you @jack-stripe

Stripe/PublicHeaders/STPSourceParams.h
```diff
@@ -0,0 +1,200 @@
...
+ *  will be chargeable once ready.
+ */
+@property (nonatomic, copy, nullable) NSString *currency;
+
+/**
+ *  The authentication flow of the source to create. `flow` may be "redirect",
+ *  "receiver", "verification", or "none". It is generally inferred unless a type
+ *  supports multiple flows.
+ */
+@property (nonatomic, copy, nullable) NSString *flow;
```
jack-stripe: Can this be an STPSourceFlow?

Stripe/PublicHeaders/STPSourceParams.h
```diff
@@ -0,0 +1,200 @@
...
+ *  source parameters.
+ */
+@property (nonatomic, copy, nullable) NSString *token;
+
+/**
+ *  Whether this source should be reusable or not. `usage` may be "reusable" or "single_use".
+ *  Some source types may or may not be reusable by construction, while other may leave the
+ *  option at creation.
+ */
+@property (nonatomic, copy, nullable) NSString *usage;
```
jack-stripe: Can this be an STPSourceUsage?

Stripe/PublicHeaders/STPSourceParams.h
```diff
@@ -0,0 +1,200 @@
...
+/**
+ *  An object representing parameters used to create a Source object. 
+ *  @see https://stripe.com/docs/api#create_source
+ */
+@interface STPSourceParams : NSObject<STPFormEncodable, NSCopying>
+
+/**
+ *  The type of the source to create. Required.
+ */
+@property (nonatomic, copy, nullable) NSString *type;
```
bg-stripe: 👍

bdorfman-stripe: So changing this limits the usage of this api to sources that are known about and manually built in to the SDK. Keeping the raw params as strings gives less type safety, but also allows people to use new sources as they are introduced to the API without us having to release a new SDK version. Do you think that is a fair tradeoff? I have some mixed feelings and could go either way.

bg-stripe: Yeah, I thought about this a bit more, and I think it's a reasonable tradeoff to make – I don't think it'll be necessary to add new source types so often that it's worth abandoning the type safety.

bdorfman-stripe: 👍

PR #473, opened 2016-10-06
Improve Stripe error bridging with Swift

Jake000:
## Summary

With the Swift evolution proposal [Improved NSError Bridging](https://github.com/apple/swift-evolution/blob/master/proposals/0112-nserror-bridging.md) having been implemented we can now directly cast NSErrors coming from Stripe to the `STPError` type that is created by Clang via the import process.
## Motivation

Most of the motivation is covered by the [Swift evolution proposal](https://github.com/apple/swift-evolution/blob/master/proposals/0112-nserror-bridging.md#motivation), particularly number 2. Working with NSError and fetching details out of the userInfo dictionary in Swift is less than ideal due to the amount of casting required.

With this change the Swift generated interface for StripeError.h turns from this:

``` Swift
public let StripeDomain: String

public enum STPErrorCode : Int {
    case STPConnectionError // Trouble connecting to Stripe.
    case STPInvalidRequestError // Your request had invalid parameters.
    case STPAPIError // General-purpose API error (should be rare).
    case STPCardError // Something was wrong with the given card (most common).
    case STPCancellationError // The operation was cancelled.
    case STPCheckoutUnknownError // Checkout failed
    case STPCheckoutTooManyAttemptsError // Too many incorrect code attempts
}

// (Usually determined locally:)
public let STPInvalidNumber: String
public let STPInvalidExpMonth: String
public let STPInvalidExpYear: String
public let STPInvalidCVC: String

// (Usually sent from the server:)
public let STPIncorrectNumber: String
public let STPExpiredCard: String
public let STPCardDeclined: String
public let STPProcessingError: String
public let STPIncorrectCVC: String
```

to this:

``` Swift
public let StripeDomain: String

public struct STPError {
    public init(_nsError: NSError)
    public static var _nsErrorDomain: String { get }

    public enum Code : Int {
        public typealias _ErrorType = STPError
        case STPConnectionError // Trouble connecting to Stripe.
        case STPInvalidRequestError // Your request had invalid parameters.
        case STPAPIError // General-purpose API error (should be rare).
        case STPCardError // Something was wrong with the given card (most common).
        case STPCancellationError // The operation was cancelled.
        case STPCheckoutUnknownError // Checkout failed
        case STPCheckoutTooManyAttemptsError // Too many incorrect code attempts
    }

    public static var STPConnectionError: STPError.Code { get }
    public static var STPInvalidRequestError: STPError.Code { get }
    public static var STPAPIError: STPError.Code { get }
[...]

PR #553, opened 2017-02-10
Improve Swift error bridging

bg-stripe:
r? @bdorfman-stripe 

Made a slight modification to #473 , adding an `STP_ERROR_ENUM` macro (the proposed `NS_ERROR_ENUM` macro still doesn't exist, but we can use our own for now).

Also to summarize the changes in practice:

```swift
// before
let errorCode: STPErrorCode = STPErrorCode.STPCardError // this is an Int
let cardErrorCode: String = STPCardDeclined // this is a String
```

```swift
// after
let errorCode: STPErrorCode = .STPCardError      // this is an enum(rawValue: Int)
let cardErrorCode: STPCardErrorCode = .declined  // this is an enum(rawValue: String)
```

### DIFF
--- a/MIGRATING.md
+++ b/MIGRATING.md
@@ -1,5 +1,9 @@
 ## Migration Guides
 
+### Migrating from versions < 10.0
+
+- `STPErrorCode` and `STPCardErrorCode` are now first class Swift enums (before, their types were `Int` and `String`, respectively)
+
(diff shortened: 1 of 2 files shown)


bdorfman-stripe:
Cool. Do you want to put a note in about the changes into the migrating doc so we don't forget later?

PR #554, opened 2017-02-10
Remove escaping note from docs

bg-stripe:
r? @bdorfman-stripe 

This bug has since been fixed (https://github.com/apple/swift/pull/4905)


bdorfman-stripe:
LGTM

PR #555, opened 2017-02-15
[sources-ui] STPSourceInfoViewController

bg-stripe:
@bdorfman-stripe This is pretty bare, but I think it'd be useful to review the basic structure and API and get this into a staging branch (`sources-ui`). 

I've created some more flexible cell abstractions, which will probably be useful for `STPAddSourceViewController` (e.g. we'll need a custom set of countries in the picker for the SEPA form).


Stripe/STPSourceInfoViewController.m
```diff
@@ -0,0 +1,307 @@
...
+    [self.view addGestureRecognizer:[[UITapGestureRecognizer alloc] initWithTarget:self action:@selector(endEditing)]];
+}
+
+- (void)createAndSetupCells {
+    STPTextFieldTableViewCell *nameCell = [[STPTextFieldTableViewCell alloc] init];
+    nameCell.placeholder = STPLocalizedString(@"Name", @"Caption for Name field on bank info form");
+    if (self.sourceParams.owner) {
+        nameCell.contents = self.sourceParams.owner[@"name"];
+    }
+    switch (self.sourceType) {
```
bdorfman-stripe: Should we try to pull out the source-specific logic from this class into some other delegate or data source?

bdorfman-stripe:
Looks pretty good. I think my only suggestion is to maybe keep the main view controller free of payment method specific logic, and instead make some sort of delegate/datasource protocol to fetch source-specific information. Just to keep all the custom logic each contained in their own files and make it easier to add new ones.

bg-stripe:
👍 – I'll pull out the logic into an `STPSourceInfoDataSource` protocol.

PR #556, opened 2017-02-17
Add source object parsing to STPCustomer

bdorfman-stripe:
All STPPaymentContext stuff that reads from Customer already checks to see if the source is an STPCard (and skips it if not) so I think this change should be ok.

Merge target is bg-sources because I need STPSource and it's not in master yet. Can change if it gets merged.


bg-stripe:
Can we also update `STPCustomerDeserializerTest`? (e.g. add a source object to the customer in `testInitWithJSONResponse_validJSON`)

bdorfman-stripe:
Good point, will add.

PR #558, opened 2017-02-22
Update STPSource enums

bg-stripe:
r? @bdorfman-stripe 

- Added comments to each enum value
- Changed "verification" to "code_verification" (I think the docs were wrong when I first implemented this)
- Added the "failed" status, which we were missing :/


bdorfman-stripe:
LGTM

PR #560, opened 2017-02-23
Update example app

bg-stripe:
r? @bdorfman-stripe 

I've updated the Obj-C example app so that it contains separate examples for four payment methods (Card, Card+3DS, Apple Pay, Sofort).

I've also updated the example backend (just the endpoint and parameter names): 
https://github.com/stripe/example-ios-backend/pull/10


Example/Stripe iOS Example (Custom)/BrowseExamplesViewController.m
```diff
@@ -0,0 +1,157 @@
...
+
+    // This passes the token off to our payment backend, which will then actually complete charging the card using your Stripe account's secret key
+    NSURLSessionConfiguration *config = [NSURLSessionConfiguration defaultSessionConfiguration];
+    NSURLSession *session = [NSURLSession sessionWithConfiguration:config];
+
+    NSString *urlString = [BackendBaseURL stringByAppendingPathComponent:@"create_charge"];
+    NSURL *url = [NSURL URLWithString:urlString];
+    NSMutableURLRequest *request = [[NSMutableURLRequest alloc] initWithURL:url];
+    request.HTTPMethod = @"POST";
+    NSString *postBody = [NSString stringWithFormat:@"source=%@&amount=%@", sourceID, @1099];
```
bdorfman-stripe: Not sure we should be passing up amounts like this, as I think it's considered bad security practice.

bg-stripe: I think it's fine for now – it would be more secure to create an order and pass up the order ID, but that would complicate the example, and I think that's out of the scope of this change (we'd have to update the swift example too).

bdorfman-stripe:
LGTM

And I think you can ignore the carthage failure on travis, it fails all the time for no reason.

PR #567, opened 2017-02-24
Add timeout to polling method

bg-stripe:
r? @bdorfman-stripe


bdorfman-stripe:
Looks fine. Although I also just realized that we need to dispatch poller completions back to main thread probably if you want to add that as well (or open a new pr for it).

bg-stripe:
👍 updated to dispatch the completion to the main thread in 1142a2a

PR #568, opened 2017-02-24
Add SEPA debit details object

bdorfman-stripe:
Missing some header documentation because I'm not actually clear on what all these properties are used for. Possibly having this object isn't even worthwhile at all, but the code is here in case we decide we want it for launch.


bg-stripe:
LGTM. I've filled in header documentation in 91bbafc

PR #569, opened 2017-02-24
Fix card details last4 nullability

bdorfman-stripe:


bg-stripe:
👍

billinghamj:
What are the circumstances in which `last4` could be null?

bg-stripe:
@billinghamj As Sources are still in preview, we've decided to be more conservative in setting nullability in our models. You can expect last4 to always be present, but you should provide a sensible fallback in case it is nil.

PR #572, opened 2017-03-01
Add muid to card token creation

bg-stripe:
r? @bdorfman-stripe 
cc @matt-stripe 

* I've replaced the data parameter in `createTokenWithData` and `postWithAPIClient` with a dictionary.
* For card tokens, I've added an `muid` parameter to the request.
* I've QA'd this pretty thoroughly, and our existing functional tests cover the change pretty well. I've also verified that the `muid` we send gets saved to the token's external metadata.


bdorfman-stripe:
LGTM

Also: At some point we should consider replacing some of the query param encoding logic with the built in iOS functions now that they exist (eg NSURLComponents, NSQueryItem, `stringByAddingPercentEncodingWithAllowedCharacters`, etc).

PR #574, opened 2017-03-03
[sources-ui] STPSourceInfoViewController

bg-stripe:
re-r? @bdorfman-stripe 

My previous PR (https://github.com/stripe/stripe-ios/pull/555) got corrupted somehow after rebasing, so I had to make a new PR. It's the same one you reviewed already, only I've moved source-specific logic into subclasses of `STPSourceInfoDataSource`.

### DIFF
--- a/Stripe/PublicHeaders/STPSourceInfoViewController.h
+++ b/Stripe/PublicHeaders/STPSourceInfoViewController.h
@@ -1,0 +1,86 @@
+//
+//  STPSourceInfoViewController.h
+//  Stripe
+//
+//  Created by Ben Guo on 2/9/17.
+//  Copyright © 2017 Stripe, Inc. All rights reserved.
+//
+
+#import <Stripe/Stripe.h>
+#import "STPCoreTableViewController.h"
+#import "STPSource.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@protocol STPSourceInfoViewControllerDelegate;
+@class STPSourceParams, STPPaymentConfiguration, STPUserInformation;
+
+/** 
+ *  You can use this view controller to collect information for payment sources
+ *  that require additional information upon creation (e.g. iDEAL, which requires 
+ *  your customer's name and bank.) Note that this view controller renders a right 
+ *  bar button item that submits the form, so it must be shown inside a
+ *  `UINavigationController`.
+ */
+@interface STPSourceInfoViewController : STPCoreTableViewController
+
+/**
+ *  Initializes a new `STPSourceInfoViewController` with the provided parameters. 
+ *  When the user submits the form, the view controller will return an STPSourceParams 
+ *  object its delegate, updated with the information entered by the user. If the 
+ *  given source type is unsupported, this initializer will return nil.
+ *
+ *  @param type                   The type of the source.
+ *  @param amount                 The amount of the source.
+ *  @param configuration          The configuration to use. This determines the source's returnURL.
+ *  @param prefilledInformation   Use this to provide any information you've already collected from your user.
+ *  @param theme                  The theme to use to inform the view controller's visual appearance. @see STPTheme
+ */
+- (nullable instancetype)initWithSourceType:(STPSourceType)type
+                                     amount:(NSInteger)amount
+                              configuration:(STPPaymentConfiguration *)configuration
+                       prefilledInformation:(STPUserInformation *)prefilledInformation
+                                      theme:(STPTheme *)theme;
+
+/**
+ *  You should check this property after initializing `STPSourceInfoViewController`.
+ *  If this property contains an STPSourceParams object, you already have sufficient
+ *  information about the source, and can immediately use this object to create
+ *  a source. Otherwise, if this property is nil, you must present the view controller
+ *  to gather additional info about the source from your user. When the user submits
+ *  the form, you can use the STPSourceParams object returned to your delegate to
+ *  create a source.
+ */
+@property(nullable, nonatomic, readonly) STPSourceParams *completeSourceParams;
+
+/**
+ *  The view controller's delegate. This must be set before showing the view 
+ *  controller in order for it to work properly. @see STPSourceInfoViewControllerDelegate
+ */
+@property(nonatomic, weak) id<STPSourceInfoViewControllerDelegate> delegate;
+
+@end
+
+/**
+ *  An `STPSourceInfoViewControllerDelegate` is notified when a user submits information or cancels in an `STPSourceInfoViewController`.
+ */
+@protocol STPSourceInfoViewControllerDelegate <NSObject>
+
+/**
+ *  This is called when the user cancels entering source information. You should dismiss (or pop) the view controller at this point.
+ *
+ *  @param sourceInfoViewController the view controller that has been cancelled
+ */
+- (void)sourceInfoViewControllerDidCancel:(STPSourceInfoViewController *)sourceInfoViewController;
+
+/**
+ *  This is called when the user finishes entering source information and submits the form. You should use the `STPSourceParams` object returned by this callback to create a source, and dismiss (or pop) the view controller.
+ *
+ *  @param sourceInfoViewController the view controller that has been cancelled
+ */
+- (void)sourceInfoViewController:(STPSourceInfoViewController *)sourceInfoViewController
+       didFinishWithSourceParams:(STPSourceParams *)sourceParams;
+
+@end
+
+NS_ASSUME_NONNULL_END
--- a/Stripe/STPBancontactSourceInfoDataSource.m
+++ b/Stripe/STPBancontactSourceInfoDataSource.m
@@ -1,0 +1,47 @@
+//
+//  STPBancontactSourceInfoDataSource.m
+//  Stripe
+//
+//  Created by Ben Guo on 3/2/17.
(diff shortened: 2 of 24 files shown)


Stripe/PublicHeaders/STPSourceInfoViewController.h
```diff
@@ -0,0 +1,59 @@
...
+ *  @param sourceParams     The source parameters to collect additional info for. @see STPSourceParams
+  *  @param theme           The theme to use to inform the view controller's visual appearance. @see STPTheme
+ */
+- (nullable instancetype)initWithSourceParams:(STPSourceParams *)sourceParams
+                                        theme:(STPTheme *)theme;
+
+/**
+ *  The view controller's delegate. This must be set before showing the view controller in order for it to work properly. @see STPSourceInfoViewControllerDelegate
+ */
+@property(nonatomic, weak) id<STPSourceInfoViewControllerDelegate> delegate;
```
bdorfman-stripe: I know we use delegates pretty much everywhere else, but what do you think about using completion blocks instead? Seems a bit easier for integrators to set up, and there are no callbacks here besides finished or cancelled.

Stripe/STPPickerTableViewCell.h
```diff
@@ -0,0 +1,30 @@
...
+
+@interface STPPickerTableViewCell : STPTextFieldTableViewCell
+
+@property(nonatomic) id<STPPickerDataSource> pickerDataSource;
+
+@end
+
+@protocol STPPickerDataSource <NSObject>
+
+- (NSInteger)numberOfRows;
```
bdorfman-stripe: Do we want to name these something more specific to picker data source to avoid potential future method name collisions?

Stripe/PublicHeaders/STPSourceInfoViewController.h
```diff
@@ -0,0 +1,59 @@
...
+ */
+@interface STPSourceInfoViewController : STPCoreTableViewController
+
+/**
+ *  Initializes a new `STPSourceInfoViewController` with the provided parameters. When the user submits the form, the view controller will return a copy of the provided `STPSourceParams` object to its delegate, updated with the information entered by the user.
+ *
+ *  @param sourceParams     The source parameters to collect additional info for. @see STPSourceParams
+  *  @param theme           The theme to use to inform the view controller's visual appearance. @see STPTheme
+ */
+- (nullable instancetype)initWithSourceParams:(STPSourceParams *)sourceParams
```
bdorfman-stripe: Do you want to use our existing "prefilled user information" class here? Or should those prefilled bits just be already set on the source params you are passing in?

Stripe/PublicHeaders/STPSourceInfoViewController.h
```diff
@@ -0,0 +1,59 @@
...
+
+/** You can use this view controller to collect information for payment sources that require additional information upon creation (e.g. iDEAL, which requires your customer's name and bank.) Note that this view controller renders a right bar button item that submits the form, so it must be shown inside a `UINavigationController`.
+ */
+@interface STPSourceInfoViewController : STPCoreTableViewController
+
+/**
+ *  Initializes a new `STPSourceInfoViewController` with the provided parameters. When the user submits the form, the view controller will return a copy of the provided `STPSourceParams` object to its delegate, updated with the information entered by the user.
+ *
+ *  @param sourceParams     The source parameters to collect additional info for. @see STPSourceParams
+  *  @param theme           The theme to use to inform the view controller's visual appearance. @see STPTheme
```
bdorfman-stripe: Should add a comment about returning nil if not a valid source type.

Tests/Tests/STPSourceInfoViewControllerTest.m
```diff
@@ -0,0 +1,109 @@
...
+@property(nonatomic)STPSourceInfoDataSource *dataSource;
+@end
+
+@interface STPSourceInfoViewControllerTest : XCTestCase
+
+@end
+
+@implementation STPSourceInfoViewControllerTest
+
+- (STPSourceInfoViewController *)sutWithParams:(STPSourceParams *)params {
```
bdorfman-stripe: What does `sut` stand for?

Tests/Tests/STPSourceInfoViewControllerTest.m
```diff
@@ -0,0 +1,109 @@
...
+#import <XCTest/XCTest.h>
+#import "STPSourceInfoViewController.h"
+#import "STPSourceInfoDataSource.h"
+#import "STPTextFieldTableViewCell.h"
+
+@interface STPSourceInfoViewController ()
+@property(nonatomic)STPSourceInfoDataSource *dataSource;
+@end
+
+@interface STPSourceInfoViewControllerTest : XCTestCase
```
bdorfman-stripe: Should this have snapshot tests? (can wait to add until designs are more final)

bdorfman-stripe:
Done with my initial review. Sorry for doing single comments instead of grouped github review.

Tests/Tests/STPSourceInfoViewControllerTest.m
bg-stripe: Yep – I was thinking we'd add snapshot tests when we do localization.

Tests/Tests/STPSourceInfoViewControllerTest.m
bg-stripe: ah sorry, i use that jargon out of habit:
https://en.wikipedia.org/wiki/System_under_test

Stripe/PublicHeaders/STPSourceInfoViewController.h
bg-stripe: 👍

Stripe/STPPickerTableViewCell.h
bg-stripe: good point – I'll update the names to things like`numberOfRowsInPicker`

Stripe/PublicHeaders/STPSourceInfoViewController.h
bg-stripe: Hm, I was imagining that we'd prefill the source params before passing them into STPSourceInfoVC, but it seems like it'd be better to accept `prefilledInformation` here – I'll update.

Stripe/PublicHeaders/STPSourceInfoViewController.h
bg-stripe: Hm – I'd prefer to leave this as a delegate, just because it's the pattern we've been using elsewhere. I think if we decide to move to completion blocks we should do it across STPAddCardVC, STPAddSourceVC, and STPSourceInfoVC, since they all just have finished and cancelled callbacks.

Stripe/PublicHeaders/STPSourceInfoViewController.h
bdorfman-stripe: Yeah, not sure what to do if the params have conflicting information to prefilledInformation? Would it be better to _just_ use prefilledInformation and return a new STPSourceParams? Or have a helper to fill source params from prefill?

Stripe/PublicHeaders/STPSourceInfoViewController.h
bdorfman-stripe: Ok. I think it might be worth doing that conversion at some point? But we can punt that for later.

Stripe/PublicHeaders/STPSourceInfoViewController.h
bg-stripe: I think it'd be best if you could immediately use the `STPSourceParams` object returned by this view controller, so it'd have to be prefilled with amount and returnURL, in addition to things in `prefilledInformation`. Maybe to avoid the conflicting information issue, we should just have an initializer that takes a source type, amount, returnURL, and prefilledInformation?

bdorfman-stripe: Sure that makes sense.

Other VC's we provide init with payment configuration also right? Might be better to use that to pull the returnURL off?

bg-stripe: ah yep, will use that instead of returnURL

bg-stripe:
@bdorfman-stripe addressed your comments in 3fe8572, and also added a new `completeSourceParams` property that you can check after initializing `STPSourceInfoVC` to determine if you need to present the view controller.

Stripe/STPBancontactSourceInfoDataSource.m
```diff
@@ -27,7 +27,7 @@ - (instancetype)initWithSourceParams:(STPSourceParams *)sourceParams {
     return self;
 }
 
-- (STPSourceParams *)completedSourceParams {
+- (STPSourceParams *)completeSourceParams {
```
bdorfman-stripe: Should we build this validation into STPSourceParams somehow? Or have a separate validator class as well? Instead of having it in this VC-specific data source type?

bg-stripe: Hm, so I think of these methods as being pretty tied to the VC – they construct STPSourceParams from the data source's contents, and then perform some very light validation that just checks whether the fields in the form have been filled.

I'm not sure there's a need for a more in-depth STPSourceParams validator at this point – STPSourceInfoVC will do its best to ensure the returned SourceParams object is filled, and I think we can account for edge cases like not setting your returnURL by printing an error when you initialize the view controller.

Issue #578, opened 2017-03-06
Add property attributes to non-GC Object


ogres:
Hello,

I am getting following compilation warning in few places: 
>"No 'assign', 'retain', or 'copy' attribute is specified - 'assign' is assumed".

Would be good to add proper attributes ( strong/weak/copy) to silence compiler.

Properties missing attributes:

- `STPUserInformation.billingAddress`
- `STPAddCardViewController.prefilledInformation`
- `STPPaymentContext.prefilledInformation`
- `STPPaymentMethodsViewController.prefilledInformation`

Stripe 9.4.0 via Cocoapods.

bdorfman-stripe:
Ah thanks, I'll add those.

By the way, any idea what project setting you have turned on to enable this warning? Not seeing it in our Xcode and would like to turn this on for us to prevent future regressions.

ogres:
I am not sure how to enable this by default in Xcode, but the warning only appears if a project contains mix of ARC enabled/disabled files.

bdorfman-stripe:
Ah ok. I couldn't find any specific setting to turn it on either, but fixed the issues you pointed out. In master now, and we should have a new official release with it imminently. Let me know if you find any other issues like this. Thanks again!

ogres:
Thank you!

PR #579, opened 2017-03-06
Fix several property declarations.

bdorfman-stripe:
Add some missing strong/nullable identifiers.

Fixes the ones in https://github.com/stripe/stripe-ios/issues/578 and some other issues I noticed.

### DIFF
--- a/Stripe/PublicHeaders/STPCard.h
+++ b/Stripe/PublicHeaders/STPCard.h
@@ -110,6 +110,11 @@ typedef NS_ENUM(NSInteger, STPCardFundingType) {
 /**
  *  The cardholder's address.
  */
+@property(nonatomic, copy, nullable) STPAddress *address;
+
+/**
+ *  The cardholder's address.
+ */
 @property (nonatomic, copy, nullable) NSString *addressLine1;
 @property (nonatomic, copy, nullable) NSString *addressLine2;
 @property (nonatomic, copy, nullable) NSString *addressCity;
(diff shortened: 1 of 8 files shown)


Stripe/PublicHeaders/STPCard.h
```diff
@@ -110,6 +110,11 @@ typedef NS_ENUM(NSInteger, STPCardFundingType) {
 /**
  *  The cardholder's address.
  */
+@property(nonatomic, copy, nullable) STPAddress *address;
```
bg-stripe: if this is inherited from `STPCardParams`, shouldn't it have the same nullability?

bdorfman-stripe: We were getting a warning from the analyzer because the implementation of this can return nil even though the super class says nonnull, so I added this to correct it. If you think the implementation should be changed instead we could do that.

Stripe/PublicHeaders/STPCardParams.h
```diff
@@ -51,9 +51,7 @@
 /**
  *  The cardholder's address.
  */
-#if TARGET_OS_IPHONE
```
bg-stripe: we should remove this check in the implementation too

Stripe/PublicHeaders/STPCard.h
bg-stripe: Ah i see, didn't realize there was a different implementation in `STPCard`. Looks fine then!

Stripe/PublicHeaders/STPCardParams.h
bdorfman-stripe: Ah good catch.

bg-stripe:
lgtm!

PR #581, opened 2017-03-10
Add name to card source params

bg-stripe:
r? @bdorfman-stripe 

Noticed we were missing this while working on sources UI – we should probably ship a patch version with the fix soon.


bdorfman-stripe:
Do you need a nil check in here for the dictionary inserts, or are they guaranteed non-nil? Looks good otherwise, I agree we should patch soon.

bg-stripe:
I don't think you need a nil check for this form of dictionary insert – `dict[@"key"] = nil;` clears the value if there is one, or is a no-op.

bdorfman-stripe:
Ah good point. Good to merge then 👍

PR #516, opened 2016-12-02
Adding support for the file uploads and personally identifiable information

scalessec:
## Summary

This PR adds file upload support for identity verification and evidence disputes. It also adds support for the tokenization of personally identifiable information (PII).

## Motivation

Hello again, @jack-stripe ! Over here at Zola we needed to add support for identity verification from our mobile client. We thought this would be a useful addition to the main repo. Here's a summary of the changes:

- `STPFile` - model object representing a file that has been upload to the Stripe API.
- `STPFileClient` - API client to upload a `UIImage` object for either identify verification or an evidence dispute.
- `STPMultipartFormDataPart` - Represents a single part of a `multipart/form-data` upload.
- `STPMultipartFormDataEncoder` - Transforms an array of `STPMultipartFormDataPart` objects into the HTTP body data required by a `multipart/form-data` upload.
- Added `createTokenWithPersonalIDNumber:completion:` to `STPAPIClient`.

## Example Usage

```swift
// Upload an identifying document
let client = STPFileClient(publishableKey: "pk_test_foo")
client.uploadImage(UIImage(named: "image")!, purpose: .identityDocument) { file, error in
     debugPrint(file, error)
}

// Tokenize a personal ID number
let client = STPAPIClient(publishableKey: "pk_test_foo")
client.createToken(withPersonalIDNumber: 000000000) { token, error in
     debugPrint(token, error)
}
```

## Testing

Your unit test target is currently broken on master (it looks like you're missing the `FBSnapshotTestCase` dependency). If you fix the target I'm happy to provide unit tests and functional tests for my additions.

PR #586, opened 2017-03-13
File uploads and PII

bdorfman-stripe:
Following on from https://github.com/stripe/stripe-ios/pull/516

Probably want to add client side file upload size limit detection (and maybe automatic resizing) before shipping.


Stripe/STPAPIClient.m
```diff
@@ -260,9 +260,42 @@ - (void)createTokenWithPersonalIDNumber:(NSString *)pii completion:(__nullable S
 
 @implementation STPAPIClient (Upload)
 
+- (NSData *)dataForUploadedImage:(UIImage *)image
```
bdorfman-stripe: @jack-stripe Do you think this is a reasonable auto image scaling algorithm? (just keep redoing the image at 0.8x the previous scale until its small enough). Also not sure how to test this, I guess I could just check in some really large photos into our repo?

jack-stripe: @bdorfman-stripe - any luck with doing this heuristically? (Based on max JPEG file sizes for certain resolutions?) Also, to test this, you could pass through to an internal `dataForUploadedImage:maxFileSize:` method, and then test that with a couple small test images.

bdorfman-stripe: JPEG file size is largely dependent on how complicated the image you are encoding is, which is hard to quantify exactly.

bdorfman-stripe:
@bg-stripe Can you do a quick final review of this? I added a few extra tests and fixed master merge conflicts. Want to get this merged in for our next release.

Tests/Tests/UIImage+StripeTests.m
```diff
@@ -0,0 +1,42 @@
...
+@implementation UIImage_StripeTests
+
+- (void)testJpegImageResizing {
+    // Strategy is to grab an image from our bundle and resize it to
+    // both smaller and bigger size than it already is and make sure we get
+    // what we expect
+
+    UIImage *testImage = [STPImageLibrary safeImageNamed:@"stp_shipping_form@3x.png"
+                                     templateIfAvailable:NO];
+
```
bg-stripe: to make sure we're exercising the resize logic, we should probably start with an assert on the original size:
```
NSData *data = UIImageJPEGRepresentation(testImage, 0.5);
XCTAssertTrue(data.length > 50000);
```

Tests/Tests/UIImage+StripeTests.m
```diff
@@ -0,0 +1,42 @@
...
+@interface UIImage_StripeTests : XCTestCase
+
+@end
+
+@implementation UIImage_StripeTests
+
+- (void)testJpegImageResizing {
+    // Strategy is to grab an image from our bundle and resize it to
+    // both smaller and bigger size than it already is and make sure we get
```
bg-stripe: nit: these tests don't resize to a bigger size

bdorfman-stripe: Right I meant to say putting a maxBytes larger than the initial size, not resize bigger. I will reword this comment.

Tests/Tests/UIImage+StripeTests.m
bdorfman-stripe: Ah good idea, I'll add that.

bdorfman-stripe:
@bg-stripe Updated

Stripe/UIImage+Stripe.m
```diff
@@ -13,17 +13,32 @@ @implementation UIImage (Stripe)
...
+    // Try something smarter first
+    if (imageData.length > maxBytes) {
+        // Assuming jpeg file size roughly scales linearly with area of the image
+        // which is ~correct (although breaks down at really small file sizes)
+        CGFloat percentSmallerNeeded = (CGFloat) maxBytes / imageData.length;
+
+        // Shrink to a little bit less than we need to try to ensure we're under
+        // (otherwise its likely our first pass will be over the limit due to
+        // compression variance and floating point rounding)
+        scale = scale * sqrt(percentSmallerNeeded - (percentSmallerNeeded * 0.05));
```
bg-stripe: ooc, what's the heuristic here?

bdorfman-stripe: It's just area scaling plus a little extra. Eg if you have something that is 100x100 (area == 10k) and you want to make it area == 5k, (make it 50% of the original size) you have to scale down the height and width by sqrt(0.5) ~= 0.71 (100*0.71 ~= 71, 71*71 == 5041). 

To try to ensure it actually goes under the max byte limit after rounding and re-jpeging (since the jpeg algo isn't actually 100% linear on area) I just scaled it down a little extra than needed (and extra 5% * whatever we were scaling it by). Not sure that's the best way to do "a little extra", could try some other things if you have suggestions.

bg-stripe: cool, that makes sense! was just wondering if the 5% came from anywhere in particular

bdorfman-stripe: It came from me testing a few values and being like "eh this is probably fine  ¯\_(ツ)_/¯" :)

PR #590, opened 2017-03-17
[sources-ui] STPAddSourceViewController

bg-stripe:
r? @bdorfman-stripe

### DIFF
--- a/Stripe/STPAddCardViewController.m
+++ b/Stripe/STPAddCardViewController.m
@@ -47,7 +47,6 @@
 @interface STPAddCardViewController ()<STPPaymentCardTextFieldDelegate, STPAddressViewModelDelegate, STPAddressFieldTableViewCellDelegate, STPSwitchTableViewCellDelegate, UITableViewDelegate, UITableViewDataSource, STPSMSCodeViewControllerDelegate, STPRememberMePaymentCellDelegate>
 @property(nonatomic)STPPaymentConfiguration *configuration;
 @property(nonatomic)STPAddress *shippingAddress;
-@property(nonatomic)BOOL hasUsedShippingAddress;
 @property(nonatomic)STPAPIClient *apiClient;
 @property(nonatomic, weak)UIImageView *cardImageView;
 @property(nonatomic)UIBarButtonItem *doneItem;
@@ -100,7 +99,6 @@ typedef NS_ENUM(NSUInteger, STPPaymentCardSection) {
 - (void)commonInitWithConfiguration:(STPPaymentConfiguration *)configuration {
     _configuration = configuration;
     _shippingAddress = nil;
-    _hasUsedShippingAddress = NO;
     _apiClient = [[STPAPIClient alloc] initWithConfiguration:configuration];
     _addressViewModel = [[STPAddressViewModel alloc] initWithRequiredBillingFields:configuration.requiredBillingAddressFields];
     _addressViewModel.delegate = self;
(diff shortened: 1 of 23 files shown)


Stripe/STPAddCardViewController.m
```diff
@@ -162,28 +160,14 @@ - (void)createAndSetupViews {
     self.tableView.dataSource = self;
     self.tableView.delegate = self;
 
-    STPSectionHeaderView *addressHeaderView = [STPSectionHeaderView new];
```
bg-stripe: Only touched this view controller to pull out some shared fill-address-button logic

Stripe/STPAddressFieldTableViewCell.m
```diff
@@ -55,24 +56,14 @@ - (instancetype)initWithType:(STPAddressFieldType)type
                                                                     action:@selector(nextTapped:)];
         toolbar.items = @[flexibleItem, nextItem];
         _inputAccessoryToolbar = toolbar;
-        
-        NSString *countryCode = [[NSLocale autoupdatingCurrentLocale] objectForKey:NSLocaleCountryCode];
```
bg-stripe: Refactored this to used the new data source I added in #574

Stripe/STPInfoFooterView.h
```diff
@@ -0,0 +1,23 @@
...
+//
+//  Created by Ben Guo on 3/15/17.
+//  Copyright © 2017 Stripe, Inc. All rights reserved.
+//
+
+#import <UIKit/UIKit.h>
+#import "STPTheme.h"
+
+NS_ASSUME_NONNULL_BEGIN
+@interface STPInfoFooterView : UIView<UITextViewDelegate>
```
bg-stripe: Pulled out this generic superclass from `STPRememberMeTermsView`

Stripe/STPAddressViewModel.m
```diff
@@ -93,16 +99,34 @@ - (instancetype)initWithRequiredShippingFields:(PKAddressField)requiredShippingA
...
 }
 
+- (instancetype)initWithSEPADebitFields {
+    self = [super init];
+    if (self) {
+        _addressType = STPAddressTypeSEPADebit;
+        _addressCells = @[
+                          [[STPAddressFieldTableViewCell alloc] initWithType:STPAddressFieldTypeLine1 contents:@"" lastInList:NO delegate:self],
+                          [[STPAddressFieldTableViewCell alloc] initWithType:STPAddressFieldTypeCity contents:@"" lastInList:NO delegate:self],
+                          [[STPAddressFieldTableViewCell alloc] initWithType:STPAddressFieldTypeZip contents:@"" lastInList:NO delegate:self],
```
bdorfman-stripe: Do we need to optionally hide postal code based on what country you have selected? Or do all SEPA countries have postal codes?

Stripe/STPAddress.m
```diff
@@ -202,6 +202,14 @@ - (BOOL)containsRequiredShippingAddressFields:(PKAddressField)requiredFields {
     return containsFields;
 }
 
+- (BOOL)containsRequiredSEPADebitFields {
+    return (self.line1.length > 0
+            && self.city.length > 0
+            && self.country.length > 0
+            && [STPPostalCodeValidator stringIsValidPostalCode:self.postalCode
```
bdorfman-stripe: Does SEPA api actually _require_ a postal code? Worried that this validator will not require some countries that SEPA thinks should have them?

Stripe/STPSectionHeaderView.m
```diff
@@ -8,13 +8,46 @@
...
 @end
 
 @implementation STPSectionHeaderView
 
++ (instancetype)addressHeaderWithConfiguration:(STPPaymentConfiguration *)config
+                                    sourceType:(STPSourceType)sourceType
+                              addressViewModel:(STPAddressViewModel *)addressViewModel
+                               shippingAddress:(STPAddress *)shippingAddress {
+    STPSectionHeaderView *headerView = [STPSectionHeaderView new];
+    headerView.title = STPLocalizedString(@"Billing Address", @"Title for billing address entry section");
```
bdorfman-stripe: Billing address is not necessarily the correct text to use on shipping screen if you're filling from SEPA.

Stripe/STPAddress.m
billinghamj: Many countries do not have postal codes - e.g. Ireland, but should still be able to use SEPA, so likely should not be required.

bg-stripe: The API currently requires postal_code, and will error if it's missing. I've made a task to look into this – we can update postal code behavior in a future PR.

Stripe/STPSectionHeaderView.m
bg-stripe: Yep – I think current designs say Billing Address, but I'll address this along with other design tweaks in a future PR.

Stripe/STPAddress.m
bg-stripe: Looked into this, and it turns out all SEPA countries do have postal codes (Ireland has postal codes as of 2014: https://en.wikipedia.org/wiki/Postal_addresses_in_the_Republic_of_Ireland). I'll make a PR to remove IE from `countriesWithNoPostalCodes`.

billinghamj: Huh interesting, thanks for the info

bdorfman-stripe: Commented on this on the other PR, but I am suspect that we should be requiring this for IE. We should very what the SEPA api is intended to be, and probably need to have separate postal code requirements for it vs. credit cards if they really require it for all SEPA accounts.

PR #591, opened 2017-03-19
Add property on STPPaymentCardTextField for card brand

RonKliffer:
## Summary

Add property on STPPaymentCardTextField for card brand.

## Motivation
There are cases that I want to know the user's card brand. For example, I want to identify when a user enters an Amex card, because its CVV length could be 4 digits, and I want to handle the UI differently.

## Testing
This property builds on the brand property in STPPaymentCardTextFieldViewModel, and returns the same value.

### DIFF
--- a/Stripe/PublicHeaders/STPPaymentCardTextField.h
+++ b/Stripe/PublicHeaders/STPPaymentCardTextField.h
@@ -264,6 +264,11 @@
  */
 @property(nonatomic, strong, readwrite, nonnull) STPCardParams *cardParams;
 
+/**
+ *  Convenience property for reading the enterred card's brand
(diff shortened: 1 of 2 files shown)


bdorfman-stripe:
Hey @RonKliffer,

In our official releases, you can can get this same information by doing `[STPCardValidator brandForNumber:field.cardParams.number]`. You likely want to do this in one of the delegate callbacks when the user edits the number so your own UI will update live to match.

We will take this as a feature request to have that data more directly/easily accessible off STPPaymentCardTextField, but are not sure we want to add it like this at the moment. Thanks for the suggestion though! Hope that helps.

PR #592, opened 2017-03-21
[sources-ui] Initial payment method support

bdorfman-stripe:
This adds in support for new APMs as payment methods, with some placeholder comments where the links to the other new VCs should go. It should function as it used to for apple pay and non-sources cards.

Changes STPPaymentMethodType to an actual classes and NSOrderedSet instead of c bitmask and adds values for the new APMs. Changes the way they are listed on payment methods VC to be separated into "new payment types" and "saved/previously used payment types" (pending final design).


Stripe/STPCard.m
```diff
@@ -191,17 +192,23 @@ - (NSString *)stripeID {
...
     return [STPImageLibrary brandImageForCardBrand:self.brand];
 }
 
-- (UIImage *)templateImage {
+- (UIImage *)paymentMethodTemplateImage {
     return [STPImageLibrary templatedBrandImageForCardBrand:self.brand];
 }
 
+- (STPPaymentMethodType *)paymentMethodType {
+    // TODO: Should this be applePay for apple pay created cards?
```
bg-stripe: +1 for checking `isApplePayCard` and making this `applePay` if it's an apple pay token

Stripe/STPCustomer+Stripe_PaymentMethods.m
```diff
@@ -0,0 +1,85 @@
...
+    NSArray<id<STPPaymentMethod>> *savedPaymentMethods = [self.class stp_sortedPaymentMethodsFromArray:filteredSavedPaymentMethods
+                                                                                       sortOrder:availableTypes];
+
+    NSMutableOrderedSet<STPPaymentMethodType *> *availablePaymentTypes = availableTypes.mutableCopy;
+    // If apple pay is misconfigured, remove it from the list
+    if (!configuration.applePayEnabled) {
+        [availablePaymentTypes removeObject:[STPPaymentMethodType applePay]];
+    }
+
+    // TODO: if other types are misconfigured, remove or warn? Also maybe need
```
bg-stripe: Hmm, so is this about guarding against e.g. using a redirect source without setting a return URL? Maybe we should add a `validatePaymentConfiguration` method?

bdorfman-stripe: Yeah I'm not sure exactly where we should do this validation? Can't just do it when you set the available types, because you might do the setters in a different order (eg set available types first, then set redirect). Doing it right before showing the available options to the customer seemed best?

Stripe/STPCard.m
bdorfman-stripe: 👍 

Do we still get all the card info from Apple Pay? eg last4 and such. Or does Apple hide that from us?

Stripe/STPPaymentMethodType.m
```diff
@@ -0,0 +1,305 @@
...
+- (UIImage *)paymentMethodImage {
+    return [STPImageLibrary addIcon];
+}
+
+- (UIImage *)paymentMethodTemplateImage {
+    return [STPImageLibrary addIcon];
+}
+
+- (NSString *)paymentMethodLabel {
+    return @"Credit Card";
```
bg-stripe: `STPLocalizedString`

Stripe/STPPaymentMethodType.m
```diff
@@ -0,0 +1,305 @@
...
+- (UIImage *)paymentMethodImage {
+    return [STPImageLibrary addIcon];
+}
+
+- (UIImage *)paymentMethodTemplateImage {
+    return [STPImageLibrary addIcon];
+}
+
+- (NSString *)paymentMethodLabel {
+    return @"SEPA Debit";
```
bg-stripe: `STPLocalizedString`

bg-stripe: `STPLocalizedString`, and note that we don't need to translate it?

Stripe/STPPaymentMethodType.m
```diff
@@ -0,0 +1,305 @@
...
+- (UIImage *)paymentMethodImage {
+    return nil;
+}
+
+- (UIImage *)paymentMethodTemplateImage {
+    return nil;
+}
+
+- (NSString *)paymentMethodLabel {
+    return @"Bancontact";
```
bg-stripe: `STPLocalizedString`

Stripe/STPPaymentMethodType.m
bdorfman-stripe: Ah yeah I'll add that everywhere.

Should we still mark all the brand names for localization, even though they likely wont get translated? (eg Bancontact)

Tests/Tests/STPAddCardViewControllerLocalizationTests.m
```diff
@@ -39,7 +39,8 @@ - (void)performSnapshotTestForLanguage:(NSString *)language delivery:(BOOL)deliv
     config.publishableKey = @"test";
     config.companyName = @"Test Company";
     config.requiredBillingAddressFields = STPBillingAddressFieldsFull;
-    config.additionalPaymentMethods = STPPaymentMethodTypeAll;
+    config.primaryPaymentMethodTypes = @[[STPPaymentMethodType applePay],
```
bg-stripe: `availablePaymentMethodTypes`?

Stripe/PublicHeaders/STPPaymentConfiguration.h
```diff
@@ -33,42 +34,75 @@ NS_ASSUME_NONNULL_BEGIN
...
+ * 
+ *  The methods in this list will be shown to users in STPPaymentMethodsViewController
+ *  for users to choose from in the order they are listed in this array.
+ *
+ *  Any methods types not in either this array will not be available to users 
+ *  of your app.
+ *
  */
-@property(nonatomic)STPBillingAddressFields requiredBillingAddressFields;
+@property (nonatomic, copy) NSOrderedSet<STPPaymentMethodType *> *availablePaymentMethodTypes;
```
bg-stripe: Can we make this an NSArray externally and an NSOrderedSet internally (for better swift compatibility)?

Stripe/STPPaymentMethodType.m
bg-stripe: Hm, not sure – maybe we can mark them, but note that they don't need to be translated?

Stripe/STPCard.m
bg-stripe: I'm pretty sure we do

Stripe/STPCustomer+Stripe_PaymentMethods.m
bg-stripe: Yeah, I'm not sure either. I think I'd prefer validating and printing warnings when you initialize STPPaymentContext, and afterwards we could just silently do things like removing invalid redirect payment methods.

bg-stripe:
Looks good! Had some documentation nits, but it's probably better to just take another pass on that stuff in a future PR.

Stripe/STPCustomer+Stripe_PaymentMethods.m
bdorfman-stripe: So like: set values in payment config, warn on init in payment context with those passed in values if invalid, continue silently removing stuff here? I can update to do that.

bdorfman-stripe:
Updated everything except returnURL validation. Will add that in a future pr when we hook up redirect source support for real.

Stripe/PublicHeaders/STPPaymentContext.h
```diff
@@ -91,11 +91,6 @@ NS_ASSUME_NONNULL_BEGIN
 @property(nonatomic, readonly, nullable)id<STPPaymentMethod> selectedPaymentMethod;
 
 /**
- *  The available payment methods the user can choose between. May be nil.
- */
-@property(nonatomic, readonly, nullable)NSArray<id<STPPaymentMethod>> *paymentMethods;
```
bg-stripe: we want to add this back, right?

bdorfman-stripe: Yeah, I'm gonna need to redo a bunch of payment method tuple/exposing stuff after designs are finalized anyway, so I think maybe just merge this for now and redo it later?

PR #594, opened 2017-03-22
Remove IE from countriesWithNoPostalCodes

bg-stripe:
r? @bdorfman-stripe 

Ireland has postal codes as of 2014:
https://en.wikipedia.org/wiki/Postal_addresses_in_the_Republic_of_Ireland


bdorfman-stripe:
IE has postal codes, but my understanding is that they are not commonly used and most people would not expect to have to fill them in on a payment form. I believe our existing behavior matches other Stripe products (like Checkout and Elements) so I'm not sure we should change this without consulting with someone.

bg-stripe:
ah, I see. I'll close this then, and follow up on whether we should require postal code for IE when creating SEPA sources.

PR #596, opened 2017-03-23
[sources-ui] Update STPSourceInfoVC

bg-stripe:
r? @bdorfman-stripe 

This updates STPSourceInfoVC to bring it closer to the current designs:

- Replaced the picker views for Sofort country / iDEAL bank with selectable cells
    - I've introduced a new data source type for this, `STPSelectorDataSource`. I initially tried extending `STPPickerDataSource` so that it'd support either a picker view or a selectable table view section, but ultimately decided that the requirements are different enough to merit two different types
   - `STPCountryPickerDataSource` is still useful (it backs the country field in `STPAddressFieldTableViewCell`), but I've removed `STPPickerTableViewCell`, as it's no longer used.
  - I've pulled out a superclass of `STPPaymentMethodTVC` for the selectable cells, `STPOptionTableViewCell`
- Added an `idealBank` field to `STPUserInformation`
- Various copy and behavior updates

We'll definitely have to take another pass at design updates after things solidify – in addition to tweaking copy and adding assets, it looks like the main remaining task will be adding a caption to the Name field.


Stripe/STPBancontactSourceInfoDataSource.m
```diff
@@ -10,13 +10,14 @@
...
 #import "STPLocalizationUtils.h"
+#import "STPTextFieldTableViewCell.h"
 
 @implementation STPBancontactSourceInfoDataSource
 
 - (instancetype)initWithSourceParams:(STPSourceParams *)sourceParams {
     self = [super initWithSourceParams:sourceParams];
     if (self) {
-        self.title = STPLocalizedString(@"Bancontact Info", @"Title for form to collect Bancontact account info");
+        self.paymentMethodLabel = STPLocalizedString(@"Bancontact", @"Text for Bancontact payment method");
```
bdorfman-stripe: Should we align with STPPaymentMethodType in these classes somehow? It has some of the UI image/text bits in it already, and i assume support for those types and these data sources will be 1:1

bg-stripe: ah, good point – I'll make this a `paymentMethodType` and get the label from there

bg-stripe:
(tests are passing except for the manual installation test flaking – it passes locally).

PR #597, opened 2017-03-23
[sources-ui] Update postal code handling in STPAddSourceVC

bg-stripe:
r? @bdorfman-stripe 

We're going to remove the postal code requirement when creating SEPA debit sources for Ireland. I've updated `STPAddressViewModel` to account for this, and also refactored some confusing/duplicative logic.

### DIFF
--- a/Stripe/STPAddressViewModel.m
+++ b/Stripe/STPAddressViewModel.m
@@ -124,6 +124,9 @@ typedef NS_ENUM(NSUInteger, STPAddressType) {
         return;
     }
     STPPostalCodeType postalCodeType = [STPPostalCodeValidator postalCodeTypeForCountryCode:_addressFieldTableViewCountryCode];
+    if (self.addressType == STPAddressTypeSEPADebit) {
+        postalCodeType = [STPPostalCodeValidator postalCodeTypeForSEPACountryCode:_addressFieldTableViewCountryCode];
+    }
     BOOL shouldBeShowingPostalCode = (postalCodeType != STPCountryPostalCodeTypeNotRequired);
     // Add postal code field
     if (shouldBeShowingPostalCode && !self.showingPostalCodeCell) {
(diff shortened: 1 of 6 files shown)


Stripe/STPAddressViewModel.m
```diff
@@ -116,72 +116,82 @@ - (instancetype)initWithSEPADebitFields {
...
-                [self.delegate addressViewModel:self addedCellAtIndex:zipFieldIndex];
-                [self.delegate addressViewModelDidChange:self];
-            }
+        STPAddressFieldTableViewCell *zipCell = [[STPAddressFieldTableViewCell alloc] initWithType:STPAddressFieldTypeZip contents:@"" lastInList:NO delegate:self];
+        NSUInteger zipFieldIndex;
+        if (previousFieldIndex != NSNotFound) {
+            zipFieldIndex = previousFieldIndex + 1;
+        } else {
+            zipFieldIndex = 0;
+            zipCell.lastInList = YES;
```
bdorfman-stripe: I dont think this logic is right? Just because its the first thing doesnt mean its also the last, and vice versa.

Also if being added as last in list and there are already other fields, need to alter the previous last in list's value to NO

bg-stripe: ah, good point. it happens to work in our use cases, but i'll update to make it more correct.

bg-stripe:
> Are we only removing the SEPA requirement for Ireland? Or removing more generally for SEPA? I am still somewhat concerned about using the same postal validator for both cards and SEPA if its not globally optional like cards are.

We're only removing the country requirement for Ireland. We have a separate validator for SEPA [0], and I think it's still correct, since `stringIsValidPostalCode:countryCode:` takes into account countries with no postal codes.

[0] https://github.com/stripe/stripe-ios/blob/sources-ui/Stripe/STPAddress.m#L205-L211

bdorfman-stripe:
My point is `stringIsValidPostalCode` has a lot more country information in it than just ireland, and allows omitting codes for all sorts of countries.. It seems like we should have a "string is valid SEPA postal code" method that requires all codes except for Ireland if that's what the actual API expects.

bg-stripe:
ah – IE is the only SEPA country in the list of `countriesWithNoPostalCodes`, so using the existing method still works. Do you still think we should use a separate method?

bdorfman-stripe:
I think we should still use separate logic as there is no guarantee that billing/shipping addres slogic and sepa logic will remain in sync in the future, and I am worried that someone will make changes later and not realize what else it affects.

bg-stripe:
That's reasonable – I'll update.

bg-stripe:
re-r? @bdorfman-stripe 
Updated in a350b61 , and also added some tests

bdorfman-stripe:
LGTM

PR #598, opened 2017-03-23
Connect add sources vc to payment methods vc

bdorfman-stripe:


Stripe/STPPaymentMethodsInternalViewController.h
```diff
@@ -15,7 +15,7 @@
 @protocol STPPaymentMethodsInternalViewControllerDelegate
 
 - (void)internalViewControllerDidSelectPaymentMethod:(id<STPPaymentMethod>)paymentMethod;
-- (void)internalViewControllerDidCreateToken:(STPToken *)token
+- (void)internalViewControllerDidCreateToken:(id<STPSourceProtocol>)tokenOrSource
```
bg-stripe: nit: can we rename this to `internalViewControllerDidCreateTokenOrSource:`?

Stripe/STPPaymentMethodsInternalViewController.m
```diff
@@ -134,8 +135,13 @@ - (void)tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath
             [self.navigationController pushViewController:paymentCardViewController animated:YES];
         }
         else {
-            // Go to Add Source VC
-            // TODO: Link to Add Source VC here
+            STPAddSourceViewController *addSourceViewController = [[STPAddSourceViewController alloc] initWithSourceType:paymentType.sourceType
```
bg-stripe: this is a nullable initializer, so we should probably check for nil before continuing.

Stripe/STPPaymentMethodsInternalViewController.h
bdorfman-stripe: 👍

Stripe/STPPaymentMethodsInternalViewController.m
bdorfman-stripe: Ah good point.

bdorfman-stripe:
Updated

PR #599, opened 2017-03-23
Handle payment request being concrete source

bdorfman-stripe:
If you hit request payment method with a concrete source object, either hit existing callback for flow = none or do redirect and poll if flow == redirect

Might need some new UI or flow to handle the "we need to show a spinner while polling" case? requestPayment is theoretically already async, I'm not sure how we handle that currently (or if we do at all).

### DIFF
--- a/Stripe/PublicHeaders/STPPaymentContext.h
+++ b/Stripe/PublicHeaders/STPPaymentContext.h
@@ -198,7 +198,7 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  *  Inside this method, you should make a call to your backend API to make a charge with that Customer + source, and invoke the `completion` block when that is done.
  *
- *  @note This method is not called for redirect-based sources. See `paymentContext:redirectDidReturnForSource:`
+ *  @note This method is not called for redirect-based sources.
  *
  *  @param paymentContext The context that succeeded
  *  @param paymentResult  Information associated with the payment that you can pass to your server. You should go to your backend API with this payment result and make a charge to complete the payment, passing `paymentResult.source.stripeID` as the `source` parameter to the create charge method and your customer's ID as the `customer` parameter (see stripe.com/docs/api#charge_create for more info). Once that's done call the `completion` block with any error that occurred (or none, if the charge succeeded). @see STPPaymentResult.h
@@ -240,6 +240,16 @@ didCreatePaymentResult:(STPPaymentResult *)paymentResult
 - (void)paymentContext:(STPPaymentContext *)paymentContext
 didUpdateShippingAddress:(STPAddress *)address
             completion:(STPShippingMethodsCompletionBlock)completion;
+
+/**
+ *  This method is called after the user returns to the app after completing
+ *  a redirect-based payment method.
+ *
+ *  You may want to show some sort of loading UI on screen when this happens,
+ *  while payment context polls for updated source information. When source polling
+ *  completes, the delegate will receive a call to `paymentContext:didFinishWithStatus:error:`
+ */
+- (void)paymentContextDidReturnFromRedirect:(STPPaymentContext *)paymentContext;
 
 @end
 
--- a/Stripe/PublicHeaders/STPPaymentContext.h
+++ b/Stripe/PublicHeaders/STPPaymentContext.h
@@ -19,7 +19,7 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-@class STPPaymentContext, STPAPIClient, STPTheme;
+@class STPPaymentContext, STPTheme;
 @protocol STPBackendAPIAdapter, STPPaymentMethod, STPPaymentContextDelegate;
 
 /**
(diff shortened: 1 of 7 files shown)


Stripe/PublicHeaders/STPPaymentContext.h
```diff
@@ -198,6 +198,8 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  *  Inside this method, you should make a call to your backend API to make a charge with that Customer + source, and invoke the `completion` block when that is done.
  *
+ *  @note This method is not called for redirect-based sources. See `paymentContext:redirectDidReturnForSource:`
```
bg-stripe: Do we still want to add `paymentContext:redirectDidReturnForSource:`? Not sure how necessary it is to expose that.

Stripe/STPPaymentContext.m
```diff
@@ -443,6 +443,69 @@ - (void)requestPayment {
...
+                                                                error:nil];
+                                        break;
+                                    case STPSourceStatusCanceled:
+                                    case STPSourceStatusFailed:
+                                    case STPSourceStatusUnknown:
+                                        // TODO: should this be user cancelled in the failed state?
+                                        // (which means the user did not do the required action during the redirect)
+                                        [self.delegate paymentContext:self
+                                                  didFinishWithStatus:STPPaymentStatusError
+                                                                error:nil];
```
bg-stripe: Users will be expecting a non-nil `error` here – can we add some new NSError objects for here and below?

Stripe/STPPaymentContext.m
```diff
@@ -443,6 +443,69 @@ - (void)requestPayment {
...
+                                        break;
+                                    case STPSourceStatusPending:
+                                        [self.delegate paymentContext:self
+                                                  didFinishWithStatus:STPPaymentStatusPending
+                                                                error:nil];
+                                        break;
+                                    case STPSourceStatusCanceled:
+                                    case STPSourceStatusFailed:
+                                    case STPSourceStatusUnknown:
+                                        // TODO: should this be user cancelled in the failed state?
```
bg-stripe: I think we should use `STPPaymentStatusUserCancellation` for the failed state

Stripe/STPPaymentContext.m
```diff
@@ -443,6 +443,69 @@ - (void)requestPayment {
             // Concrete source object, check flow to see if synchronous charge or webhook based
             // If a card source, we need to check if the user wants us to 3DS or not
 
+            STPSource *source = (STPSource *)self.selectedPaymentMethod;
```
bg-stripe: can we add a TODO: to update the block above where we present `STPAddCardVC` (L409)? Or just update it to present `STPAddSourceVC` depending on the config

bg-stripe:
Not sure about whether or not we need to display a spinner. Maybe we should just be more explicit in the docs that after you call `requestPayment`, you should block your own UI and indicate to your user that a payment is in progress, e.g. by displaying a spinner. 

I think the case in which we would want our own spinner is if we decide to handle the redirect in a container – in that case, we could keep the container up and show a spinner while polling.

Stripe/PublicHeaders/STPPaymentContext.h
bdorfman-stripe: Oh yeah I originally had added this method as a separate callback and then deleted it before I opened the PR and forgot to fix this comment. Will remove.

Stripe/STPPaymentContext.m
bdorfman-stripe: This will be automatically fixed by the other PR I have open on master already.

Stripe/STPPaymentContext.m
bdorfman-stripe: Ok, will change.

Stripe/STPPaymentContext.m
bdorfman-stripe: Yep, I'll add some new ones.

Stripe/STPPaymentContext.m
bdorfman-stripe: Here's the related PR to fix this https://github.com/stripe/stripe-ios/pull/600

bg-stripe: ah right, sorry I missed that

bdorfman-stripe:
The problem there is really no spot currently in the existing delegate callbacks where the user would know to put up a spinner. It's not immediately after requestPayment (as that could trigger some new view controllers pushed and user input). We might need to add some new callback for this or something.

bg-stripe:
ah, that's a good point. maybe it would just be better for us to create some kind of spinner view to display while polling.

bdorfman-stripe:
Yeah. If we do that we should also probably just do a spinner UI in some other spots also (like waiting for backend API adapter to complete). Alternatively I can add a "did return from redirect" callback. Whichever you think is better?

bdorfman-stripe:
Updated.

One other thing: We never actually give the user the final source object anywhere in this if it's not a reusable source? Which maybe is fine? Not sure if it should be added somehow.

bg-stripe:
lgtm!

> We never actually give the user the final source object anywhere in this if it's not a reusable source?

I think this may actually be a good thing, since it'll prevent users from trying to charge redirect sources. One thing we should remember to do is associate these single use sources with the customer (added a note to our todos)

bdorfman-stripe:
I don't think you can actually associate the single uses sources with a customer. You are supposed to just pass the customer id along with the source id when you charge it and it is connected in your dashboard, but not actually attached. We don't have a lot of control over that however, as it's done on the backend via a webhook.

PR #600, opened 2017-03-23
Fix requestPayment logic

bdorfman-stripe:
If there is no selectedPaymentMethod, we should just call requestPaymentMethodsVC instead. Besides de-duping code, current logic is wrong as it doesn't offer an option to use Apple Pay if its available.

Also seems like there was a missing else statement in this if/else flow. Not sure how this was working correctly tbh, seems like if you had no method and also no shipping info, it would try to push both? And/or set your `isMidShippingInRequestPayment` bit incorrectly.

### DIFF
--- a/Stripe/STPPaymentContext.m
+++ b/Stripe/STPPaymentContext.m
@@ -383,9 +383,6 @@ typedef NS_ENUM(NSUInteger, STPPaymentContextState) {
             STPShippingAddressViewController *addressViewController = [[STPShippingAddressViewController alloc] initWithPaymentContext:self];
             [navigationController pushViewController:addressViewController animated:YES];
         }
-        else {
-            self.state = STPPaymentContextStateNone;
-        }
     }];
 }
 
--- a/Stripe/STPPaymentContext.m
+++ b/Stripe/STPPaymentContext.m
@@ -406,16 +406,9 @@
             return;
         }
         if (!self.selectedPaymentMethod) {
-            STPAddCardViewController *addCardViewController = [[STPAddCardViewController alloc] initWithConfiguration:self.configuration theme:self.theme];
-            addCardViewController.delegate = self;
-            addCardViewController.prefilledInformation = self.prefilledInformation;
-            addCardViewController.shippingAddress = self.shippingAddress;
-            UINavigationController *navigationController = [[UINavigationController alloc] initWithRootViewController:addCardViewController];
-            navigationController.navigationBar.stp_theme = self.theme;
-            navigationController.modalPresentationStyle = self.modalPresentationStyle;
-            [self.hostViewController presentViewController:navigationController animated:YES completion:nil];
+            [self presentPaymentMethodsViewController];
         }
-        if (self.configuration.requiredShippingAddressFields != STPBillingAddressFieldsNone &&
+        else if (self.configuration.requiredShippingAddressFields != STPBillingAddressFieldsNone &&
             !self.shippingAddress)
         {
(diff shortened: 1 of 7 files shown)


Stripe/STPPaymentContext.m
```diff
@@ -406,16 +406,9 @@ - (void)requestPayment {
...
         if (!self.selectedPaymentMethod) {
-            STPAddCardViewController *addCardViewController = [[STPAddCardViewController alloc] initWithConfiguration:self.configuration theme:self.theme];
-            addCardViewController.delegate = self;
-            addCardViewController.prefilledInformation = self.prefilledInformation;
-            addCardViewController.shippingAddress = self.shippingAddress;
-            UINavigationController *navigationController = [[UINavigationController alloc] initWithRootViewController:addCardViewController];
-            navigationController.navigationBar.stp_theme = self.theme;
-            navigationController.modalPresentationStyle = self.modalPresentationStyle;
-            [self.hostViewController presentViewController:navigationController animated:YES completion:nil];
+            [self presentPaymentMethodsViewController];
```
bg-stripe: do we need to update the payment methods VC delegate callback to call `requestPayment` again? (I don't think we were ever presenting payment methods VC from the `requestPayment` flow before)

bg-stripe:
Good catch on those bugs. I think there's probably more involved in presenting payment methods VC, though, since we'll want to continue with the payment after the user selects a payment method. Maybe we can rename the `isMidShippingInRequestPayment` flag to `isMidRequestPayment` and use it here too.

bg-stripe:
I also think we should still handle the case where `availablePaymentMethods` only has one generic type, e.g. `CreditCard` or `SEPADebit`, and display the Add form rather than the payment methods screen. This saves taps, and I think that's pretty high priority, as it affects conversion.

bdorfman-stripe:
Handling the "theres only one available thing to add" is handled by payment methods VC so we don't need to add it here.

Yeah good point on needing to change the delegate callbacks to continue the flow, I'll add that. This is another reason to add block based completions btw, would be a lot easier to close over the state then having to have a different bool we check in the callback.

bg-stripe:
> Handling the "theres only one available thing to add" is handled by payment methods VC so we don't need to add it here.

Ah, right, forgot about that.

bdorfman-stripe:
Rewrote a bunch based on feedback an in person discussion. Now has an internal state to track whether its: doing nothing, showing a user-requested vc, showing a requestPayment-triggered VC. Used this to rework logic a bit and updated the paymentMethod callbacks accordingly.

Stripe/STPPaymentContext.m
```diff
@@ -325,18 +377,26 @@ - (void)pushShippingViewController {
...
         STRONG(self);
-        STPShippingAddressViewController *addressViewController = [[STPShippingAddressViewController alloc] initWithPaymentContext:self];
-        [navigationController pushViewController:addressViewController animated:YES];
+        if (self.state == STPPaymentContextStateNone) {
+            self.state = STPPaymentContextStateShowingRequestedViewController;
+
+            STPShippingAddressViewController *addressViewController = [[STPShippingAddressViewController alloc] initWithPaymentContext:self];
+            [navigationController pushViewController:addressViewController animated:YES];
+        }
+        else {
```
bg-stripe: is this else block correct? (seems like we should do nothing, rather than resetting state)

bdorfman-stripe: Ah yeah sorry. Will fix.

bdorfman-stripe:
Updated

bg-stripe:
LGTM

PR #601, opened 2017-03-23
Auto select payment method if only one

bdorfman-stripe:
Adds back public accessor for available payment methods on context.
Ensures that selected payment method cant be set to something not in the available list, and that if there is only one possible method it is auto-selected.

### DIFF
--- a/Stripe/STPPaymentMethodTuple.m
+++ b/Stripe/STPPaymentMethodTuple.m
@@ -8,6 +8,8 @@
 
 #import "STPPaymentMethodTuple.h"
 
+#import "STPPaymentMethodType.h"
+
 @implementation STPPaymentMethodTuple
 
 - (instancetype)initWithSavedPaymentMethods:(NSArray<id<STPPaymentMethod>> *)savedPaymentMethods
@@ -16,7 +18,18 @@
     if ((self = [super init])) {
         _savedPaymentMethods = savedPaymentMethods ? savedPaymentMethods.copy : @[];
         _availablePaymentTypes = availablePaymentTypes ? availablePaymentTypes.copy : @[];
-        _selectedPaymentMethod = selectedPaymentMethod;
+
+        _allPaymentMethods = [NSSet setWithArray:[_savedPaymentMethods arrayByAddingObjectsFromArray:_availablePaymentTypes]];
+
+        if (_allPaymentMethods.count == 1) {
+            _selectedPaymentMethod = _allPaymentMethods.anyObject;
+        }
+        else if ([_allPaymentMethods containsObject:selectedPaymentMethod]) {
+            _selectedPaymentMethod = selectedPaymentMethod;
+        }
+        else {
+            _selectedPaymentMethod = nil;
+        }
     }
     return self;
 }
--- a/Stripe/STPPaymentMethodTuple.m
+++ b/Stripe/STPPaymentMethodTuple.m
@@ -12,8 +12,8 @@
 
 @implementation STPPaymentMethodTuple
 
-- (instancetype)initWithSavedPaymentMethods:(NSArray<id<STPPaymentMethod>> *)savedPaymentMethods
-                      availablePaymentTypes:(NSArray<STPPaymentMethodType *> *)availablePaymentTypes
+- (instancetype)initWithSavedPaymentMethods:(nullable NSArray<id<STPPaymentMethod>> *)savedPaymentMethods
+                      availablePaymentTypes:(nullable NSArray<STPPaymentMethodType *> *)availablePaymentTypes
                       selectedPaymentMethod:(nullable id<STPPaymentMethod>)selectedPaymentMethod {
     if ((self = [super init])) {
         _savedPaymentMethods = savedPaymentMethods ? savedPaymentMethods.copy : @[];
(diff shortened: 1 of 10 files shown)


Stripe/STPPaymentMethodTuple.m
```diff
@@ -16,7 +18,34 @@ - (instancetype)initWithSavedPaymentMethods:(NSArray<id<STPPaymentMethod>> *)sav
...
-        _selectedPaymentMethod = selectedPaymentMethod;
+
+        _allPaymentMethods = [NSSet setWithArray:[_savedPaymentMethods arrayByAddingObjectsFromArray:_availablePaymentTypes]];
+
+        if ([_allPaymentMethods containsObject:selectedPaymentMethod]) {
+            _selectedPaymentMethod = selectedPaymentMethod;
+        }
+        else if (_allPaymentMethods.count == 1) {
+            id<STPPaymentMethod> method = _allPaymentMethods.anyObject;
+
```
bg-stripe: nit: blank lines

bdorfman-stripe:
Ah yeah, will do.

bdorfman-stripe:
Updated

Tests/Tests/STPPaymentMethodTupleTest.m
```diff
@@ -0,0 +1,96 @@
...
+    STPPaymentMethodTuple *tuple = [[STPPaymentMethodTuple alloc] initWithSavedPaymentMethods:nil
+                                                                        availablePaymentTypes:@[[STPPaymentMethodType creditCard]]
+                                                                        selectedPaymentMethod:nil];
+
+    XCTAssertNil(tuple.selectedPaymentMethod);
+}
+
+- (void)testOnlyOneAvailableNotAllowedPaymentMethod {
+    STPPaymentMethodTuple *tuple = [[STPPaymentMethodTuple alloc] initWithSavedPaymentMethods:nil
+                                                                        availablePaymentTypes:@[[STPPaymentMethodType creditCard]]
```
bg-stripe: this test case is the same as the one above – should we delete?

Tests/Tests/STPPaymentMethodTupleTest.m
```diff
@@ -0,0 +1,96 @@
...
+
+    XCTAssertNil(tuple.selectedPaymentMethod);
+}
+
+- (void)testOnlyOneAvailableAllowedPaymentMethod {
+    STPPaymentMethodTuple *tuple = [[STPPaymentMethodTuple alloc] initWithSavedPaymentMethods:nil
+                                                                        availablePaymentTypes:@[[STPPaymentMethodType bancontact]]
+                                                                        selectedPaymentMethod:nil];
+
+    XCTAssertEqualObjects(tuple.selectedPaymentMethod, [STPPaymentMethodType bancontact]);
```
bg-stripe: can we also add a `testOnlyOneAvailableSavedPaymentMethod`, e.g.

```
    STPPaymentMethodTuple *tuple = [[STPPaymentMethodTuple alloc] initWithSavedPaymentMethods:@[card]
                                                                        availablePaymentTypes:@[[STPPaymentMethodType creditCard]]
                                                                        selectedPaymentMethod:nil];

    XCTAssertEqualObjects(tuple.selectedPaymentMethod, card);
```

Tests/Tests/STPPaymentMethodTupleTest.m
```diff
@@ -0,0 +1,96 @@
...
+
+    STPPaymentMethodTuple *tuple = [[STPPaymentMethodTuple alloc] initWithSavedPaymentMethods:@[card, source]
+                                                                        availablePaymentTypes:@[[STPPaymentMethodType creditCard],
+                                                                                                [STPPaymentMethodType giropay]]
+                                                                        selectedPaymentMethod:[STPPaymentMethodType sofort]];
+
+    XCTAssertNil(tuple.selectedPaymentMethod);
+}
+
+- (void)testOnlyOneAvailableUnselectablePaymentMethod {
```
bg-stripe: Should we also handle the case where your customer has saved cards, but cards aren't in your available types? (I think we'd have to update the implementation a bit)

```
    STPPaymentMethodTuple *tuple = [[STPPaymentMethodTuple alloc] initWithSavedPaymentMethods:@[card]
                                                                        availablePaymentTypes:@[[STPPaymentMethodType giropay]]
                                                                        selectedPaymentMethod:nil];

    XCTAssertNil(tuple.selectedPaymentMethod);
```

bdorfman-stripe: The tuple does not actually do this check. I could move the logic in here? But as written that filtering of invalid saved cards is done elsewhere.

Tests/Tests/STPPaymentMethodTupleTest.m
bdorfman-stripe: Oh yeah, not sure what happened here, maybe a bad copy and paste or something. I'll delete one of them.

Tests/Tests/STPPaymentMethodTupleTest.m
bdorfman-stripe: That is not a valid case. You have multiple available payment methods there (new card or saved card) so nothing will be auto selected. I can add a test to make sure it stays unselected if you want.

bg-stripe: ah interesting. I guess this behavior makes sense, if the customer doesn't have a `defaultSource` we shouldn't auto-select a saved card.

bg-stripe:
lgtm!

Tests/Tests/STPPaymentMethodTupleTest.m
bdorfman-stripe: Yeah, defaultSource logic is done elsewhere. Like theoretically if you have one, that will be passed through as the selectedPaymentMethod from the part of the code that parses customers. Similarly if there are banned saved sources, the customer parsing logic removes them before coming here.

The logic in this class is mostly independent from that sort of STPCustomer parsing, instead just making sure that the selectedPaymentMethod you passed in is actually in the list of all possible things. And that if there is only one possible thing that can be chosen (and no further details are required) that it just gets auto-selected.

PR #603, opened 2017-03-24
[sources-ui] Update country field based on IBAN

bg-stripe:
r? @bdorfman-stripe 

This updates the country field in `STPAddSourceVC` based on the first two digits of the IBAN, which denote the ISO country code of the bank.

This also fixes a bug when programmatically setting the contents of a `SEPACountry` cell.


bdorfman-stripe:
LGTM

PR #604, opened 2017-03-24
[sources-ui] Auto-select Sofort country based on locale

bg-stripe:
r? @bdorfman-stripe 

- For the Sofort form, we now automatically select country based on locale, if it hasn't already been set via `prefilledInformation`. If we used locale to select a country, we'll still present the view controller for the user to verify (`completeSourceParams` will be nil).

- I've also done some refactoring, moving the `userInformation` logic into the individual data sources.

- Made some copy updates in c57f67e

- Fixed footer height in b3eabed


bdorfman-stripe:
LGTM

PR #605, opened 2017-03-27
[sources-ui] Add completion block to STPSourceInfoVC

bg-stripe:
r? @bdorfman-stripe 

After thinking about it a bit more, I don't think we really need to replace delegates with completion blocks in AddSource/AddCard – I think it'd be a bit confusing there, since the `didCreateToken` callback also has a completion parameter.


bdorfman-stripe:
Yeah I guess all the other view controller's have multiple different callbacks that aren't a simple completion, which is somewhat unfortunate. Could replace with multiple blocks (which I generally prefer personally in my own code) but it's probably less confusing to leave as-is.

Do you think we should actually make this change for source info, or just leave it as delegate to match everything else?

bg-stripe:
I think it's still worth making the change for SourceInfoVC, since the completion block is so much simpler, and it's less likely that people will be using SourceInfoVC on its own.

bdorfman-stripe:
k, LGTM then

PR #608, opened 2017-03-28
Use generic error if polling completes with no error and no source

bg-stripe:
r? @bdorfman-stripe 

As discussed yesterday


bdorfman-stripe:
LGTM

PR #611, opened 2017-03-29
Add tests for STPAddSourceViewController

bg-stripe:
r? @bdorfman-stripe 

- Added tests for AddSourceVC
- Improved AddCardVC tests (we're now testing filling address fields on the CardParams object)
- Added some more fixtures + json files
- Moved json files to their own folder/group


Stripe/STPAddressFieldTableViewCell.m
```diff
@@ -262,7 +262,7 @@ - (void)nextTapped:(__unused id)sender {
 #pragma mark - STPFormTextFieldDelegate
 
 - (void)formTextFieldTextDidChange:(STPFormTextField *)textField {
-    if (self.type != STPAddressFieldTypeCountry) {
+    if (self.type != STPAddressFieldTypeCountry && self.type != STPAddressFieldTypeSEPACountry) {
```
bg-stripe: This fixes a bug I caught while adding tests – for SEPA, we were using the country's display name (i.e. "Germany") in the SourceParams object rather than the ISO code ("DE").

bdorfman-stripe: Hmm, is this a bug for regular cards also? It looks like everywhere but here the code assumes `self.contents` is the country code.

Tests/Tests/STPAddCardViewControllerTest.m
```diff
@@ -31,31 +31,45 @@ @implementation STPAddCardViewControllerTest
...
     STPPaymentConfiguration *config = [STPFixtures paymentConfiguration];
+    config.requiredBillingAddressFields = STPBillingAddressFieldsFull;
     STPTheme *theme = [STPTheme defaultTheme];
     STPAddCardViewController *vc = [[STPAddCardViewController alloc] initWithConfiguration:config
                                                                                      theme:theme];
     XCTAssertNotNil(vc.view);
     return vc;
 }
 
+- (BOOL)cardParams:(STPCardParams *)params matchCardParams:(STPCardParams *)otherParams {
```
bdorfman-stripe: Should this be moved into an `isEqual:` method on STPCardParams?

Tests/Tests/STPSourceTest.m
```diff
@@ -111,12 +33,11 @@ - (void)testDecodingSource_ideal {
     XCTAssertEqual(source.status, STPSourceStatusPending);
     XCTAssertEqual(source.type, STPSourceTypeIDEAL);
     XCTAssertEqual(source.usage, STPSourceUsageSingleUse);
-    XCTAssertEqualObjects(source.details, response[@"ideal"]);
+    XCTAssertEqualObjects(source.details[@"bank"], @"ing");
```
bdorfman-stripe: Is it strange that the tests here are checking the decoding that is all happening inside STPFixtures? Should all the decoding tests be moved into like STPFixturesTest or something?

Stripe/STPAddressFieldTableViewCell.m
bg-stripe: The logic here prevents setting `self.contents` to the text field's text (which will be the country's display name, e.g. "Germany"). In `delegateCountryCodeDidChange:`, we set the cell's contents to the country code.

The bug was that we were doing this for cells of type `STPAddressFieldTypeCountry` (which we use in the card form), but not `STPAddressFieldTypeSEPACountry` (which we use in the SEPA form), so in the SEPA form, `cell.contents` was `Germany` rather than `DE`.

bdorfman-stripe: Ah I see, makes sense.

Tests/Tests/STPSourceTest.m
bg-stripe: good point – i'll instead test this against what's in `allResponseFields`, instead of the concrete values.

bg-stripe: I'll also add some fixtures tests to make sure `allResponseFields` matches the initial dictionary.

bg-stripe: (updated)

bdorfman-stripe:
👍

PR #612, opened 2017-03-29
[sources-ui] Add network activity notifications

bg-stripe:
r? @bdorfman-stripe 

This posts notifications whenever we call `backendAPIAdapter` or `apiClient` methods. I haven't modified `STPCheckoutAPIClient`, since we'll likely remove it. I've also modified the example app to initialize PaymentContext in `viewDidLoad`, so we can catch its initial loading notification.


bdorfman-stripe:
Hmm, are there any cases where we have nested or multiple simultaneous network connections where this could result in an early end being called? Should the example app do more of an increment/decrement counting thing instead of just turning the indicator on and off to represent this?

bg-stripe:
Ah yeah, good call – updated in 560b3ed

bdorfman-stripe:
👍 

Whats the best way to document this feature for people? Do you think the constants in the STPAPIClient header are enough?

bg-stripe:
I think the constants and usage in the example app are probably sufficient for now. It seems like more of a nice-additional-feature-you-can-use than something we'd need to document more explicitly.

PR #613, opened 2017-03-30
Redirect helper

bdorfman-stripe:
Adds a helper for doing source redirects. Pass in source and completion block, and it can do either SFSafariVC or openURL/bounce to Safari and call your completion block when done.

Rewrote existing handling to use this class instead.
Copied over some old classes to handle URL stuff that I wrote back in October and never shipped.

### DIFF
--- a/Stripe/NSURLComponents+Stripe.m
+++ b/Stripe/NSURLComponents+Stripe.m
@@ -1,0 +1,51 @@
+//
+//  NSURLComponents+Stripe.m
+//  Stripe
+//
+//  Created by Brian Dorfman on 1/26/17.
+//  Copyright © 2017 Stripe, Inc. All rights reserved.
+//
+
+#import "NSURLComponents+Stripe.h"
+
+@implementation NSURLComponents (Stripe)
+
+- (NSDictionary<NSString *, NSString *> *)stp_queryItemsDictionary {
+    NSMutableDictionary *queryItems = [NSMutableDictionary new];
+
+    for (NSURLQueryItem *queryItem in self.queryItems) {
+        queryItems[queryItem.name] = queryItem.value;
+    }
+
+    return queryItems.copy;
+}
+
+- (void)setStp_queryItemsDictionary:(NSDictionary<NSString *, NSString *> *)stp_queryItemsDictionary {
+    NSMutableArray *queryItems = [NSMutableArray new];
+    [stp_queryItemsDictionary enumerateKeysAndObjectsUsingBlock:^(NSString * _Nonnull key, NSString * _Nonnull obj, BOOL * __unused _Nonnull stop) {
+        [queryItems addObject:[NSURLQueryItem queryItemWithName:key value:obj]];
+    }];
+
+    self.queryItems = queryItems.copy;
+}
+
+- (BOOL)stp_matchesURLComponents:(NSURLComponents *)rhsComponents {
+    BOOL matches = ([self.scheme isEqualToString:rhsComponents.scheme]
+                    && [self.host isEqualToString:rhsComponents.host]
+                    && [self.path isEqualToString:rhsComponents.path]);
+
+    if (matches) {
+        NSDictionary<NSString *, NSString *> *rhsQueryItems = rhsComponents.stp_queryItemsDictionary;
+
+        for (NSURLQueryItem *queryItem in self.queryItems) {
+            if (![rhsQueryItems[queryItem.name] isEqualToString:queryItem.value]) {
+                matches = NO;
+                break;
+            }
+        }
+    }
+
+    return matches;
+}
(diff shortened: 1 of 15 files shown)


Stripe/NSURLComponents+Stripe.m
```diff
@@ -0,0 +1,51 @@
...
+            if (![rhsQueryItems[queryItem.name] isEqualToString:queryItem.value]) {
+                matches = NO;
+                break;
+            }
+        }
+    }
+
+    return matches;
+}
+
```
bg-stripe: We have to reference an empty method in this class so that it gets linked when using the static library:
https://github.com/stripe/stripe-ios/blob/master/Stripe/STPCategoryLoader.m

Stripe/NSURLComponents+Stripe.m
```diff
@@ -0,0 +1,51 @@
...
+//
+//  Created by Brian Dorfman on 1/26/17.
+//  Copyright © 2017 Stripe, Inc. All rights reserved.
+//
+
+#import "NSURLComponents+Stripe.h"
+
+@implementation NSURLComponents (Stripe)
+
+- (NSDictionary<NSString *, NSString *> *)stp_queryItemsDictionary {
```
bg-stripe: can we add tests for this category? i vaguely remember you having them before

Stripe/STPRedirectContext.h
```diff
@@ -0,0 +1,139 @@
...
+#import <Foundation/Foundation.h>
+#import "STPBlocks.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ Possible states for the redirect context to be in
+
+ - STPRedirectContextStateNotStarted: Initialized, but redirect not started
+ - STPRedirectContextStateInProgress: Redirect is in progress
```
bg-stripe: missing Cancelled

Stripe/STPRedirectContext.h
```diff
@@ -0,0 +1,139 @@
...
+ *  However, it is possible the when the redirect is "completed", the user may
+ *  have not actually completed the necessary actions to authorize the charge.
+ *
+ *  You can use `STPAPIClient` to listen for state changes on the source
+ *  object as a way to identify whether the user action succeeded or not.
+ *  @see `[STPAPIClient startPollingSourceWithId:clientSecret:timeout:completion:]`
+ *
+ *  You should not use either this class, nor `STPAPIClient`, as a way
+ *  to determine when you should charge the source. Use Stripe webhooks on your
+ *  backend server to listen for source state changes and to make the charge.
```
bg-stripe: nice docs!

Stripe/STPURLCallbackHandler.h
```diff
@@ -0,0 +1,31 @@
...
+
+@interface STPURLCallbackHandler : NSObject
+
++ (instancetype)shared;
+
+- (BOOL)handleURLCallback:(NSURL *)url;
+- (void)registerListener:(id<STPURLCallbackListener>)listener
+                  forURL:(NSURL *)url;
+
+- (void)unregisterListener:(id<STPURLCallbackListener>)listener
```
bg-stripe: do we need the URL variant? (looks like we're not using it)

Stripe/NSURLComponents+Stripe.m
bdorfman-stripe: Ah yep. Just forgot to resurrect those files when I copied this out of my old branch. I'll pull and update them.

Stripe/NSURLComponents+Stripe.m
bdorfman-stripe: Oh yes good catch. 

We should maybe invest in some sort of automated script that detects this? I'm worried now I've missed this in other places.

Stripe/STPRedirectContext.h
bdorfman-stripe: Ah yeah, will add thanks.

Stripe/STPRedirectContext.h
bdorfman-stripe: :)

Stripe/STPURLCallbackHandler.h
bdorfman-stripe: Ah yeah, can delete. I think a holdover from prev iteration.

Stripe/NSURLComponents+Stripe.m
bg-stripe: ah yeah, a test would be nice (added an issue to track it). did a check just now, and it looks like we are doing this in all our existing categories.

bdorfman-stripe:
Updated

bg-stripe:
lgtm!

PR #614, opened 2017-03-30
Fix bug when canceling in PaymentMethodsVC

bg-stripe:
r? @bdorfman-stripe 

To reproduce in the example app:
- tap the payment row
- add a payment method
- tap the payment row
- tap "Back"
- PaymentContext's state is stuck in `ShowingRequestedViewController`. Tapping the payment or shipping row does nothing

`STPPaymentMethodVC`'s navigation item gets forwarded to its internal view controller, so `handleBackOrCancelTapped` does nothing.


bdorfman-stripe:
Is this also a problem on master?

bg-stripe:
ah, forgot that we added the state machine stuff on master. i'll re-PR.

PR #615, opened 2017-03-30
Fix bug when canceling in PaymentMethodsVC

bg-stripe:
r? @bdorfman-stripe 

Same as #614 , but to master.


bdorfman-stripe:
I see, so the internal view controller was handling it and just dismissing itself without doing a callback, is that what happened? Good catch, LGTM

bg-stripe:
yep!

PR #616, opened 2017-03-30
Make SEPA line1 optional

bg-stripe:
r? @bdorfman-stripe

### DIFF
--- a/Stripe/STPSourceParams.m
+++ b/Stripe/STPSourceParams.m
@@ -150,17 +150,14 @@
     params.type = STPSourceTypeSEPADebit;
     params.currency = @"eur"; // SEPA Debit must always use eur
 
-    NSDictionary<NSString *,NSString *> *address =
-    @{
-      @"line1": addressLine1,
-      @"city": city,
-      @"postal_code": postalCode,
-      @"country": country
-      };
-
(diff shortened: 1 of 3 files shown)


Stripe/STPSourceParams.m
```diff
@@ -150,17 +150,16 @@ + (STPSourceParams *)sepaDebitParamsWithName:(NSString *)name
...
-      @"city": city,
-      @"postal_code": postalCode,
-      @"country": country
-      };
-
+    NSMutableDictionary<NSString *,NSString *> *address = [NSMutableDictionary new];
+    address[@"city"] = city;
+    address[@"postal_code"] = postalCode,
+    address[@"country"] = country;
+    if (addressLine1) {
```
bdorfman-stripe: Is this check even necessary? Setting a nil value with a mutable dictionary is valid iirc.

bdorfman-stripe:
LGTM.

Should we have some tests to verify optionality stuff is working correctly? I guess we are somewhat just enforcing it through nullability annotations so not sure if testable without compiler errors in our setup.

Stripe/STPSourceParams.m
bg-stripe: oh whoops, good point.

bg-stripe: (updated)

bdorfman-stripe:
👍

PR #617, opened 2017-03-31
Add metadata to STPUserInformation

bg-stripe:
r? @bdorfman-stripe

### DIFF
--- a/Stripe/PublicHeaders/STPSourceParams.h
+++ b/Stripe/PublicHeaders/STPSourceParams.h
@@ -49,25 +49,25 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  *  A set of key/value pairs that you can attach to a source object.
  */
-@property (nonatomic, copy, nullable) NSDictionary *metadata;
+@property (nonatomic, copy, nullable) NSDictionary<NSString *,NSString *>*metadata;
 
 /**
  *  Information about the owner of the payment instrument. May be used or required
  *  by particular source types.
  */
-@property (nonatomic, copy, nullable) NSDictionary *owner;
+@property (nonatomic, copy, nullable) NSDictionary<NSString *, id>*owner;
 
 /**
  *  Parameters required for the redirect flow. Required if the source is authenticated by 
  *  a redirect (`flow` is "redirect").
  */
-@property (nonatomic, copy, nullable) NSDictionary *redirect;
+@property (nonatomic, copy, nullable) NSDictionary<NSString *, id>*redirect;
 
 /**
  *  An optional token used to create the source. When passed, token properties will override 
  *  source parameters.
  */
-@property (nonatomic, copy, nullable) NSString *token;
+@property (nonatomic, copy, nullable) NSDictionary<NSString *, id>*token;
 
 /**
  *  Whether this source should be reusable or not. `usage` may be "reusable" or "single_use".
--- a/Stripe/PublicHeaders/STPSourceParams.h
+++ b/Stripe/PublicHeaders/STPSourceParams.h
@@ -67,7 +67,7 @@ NS_ASSUME_NONNULL_BEGIN
  *  An optional token used to create the source. When passed, token properties will override 
  *  source parameters.
  */
-@property (nonatomic, copy, nullable) NSDictionary<NSString *, id>*token;
+@property (nonatomic, copy, nullable) NSString *token;
 
 /**
  *  Whether this source should be reusable or not. `usage` may be "reusable" or "single_use".
--- a/Stripe/PublicHeaders/STPUserInformation.h
+++ b/Stripe/PublicHeaders/STPUserInformation.h
@@ -12,7 +12,10 @@
 NS_ASSUME_NONNULL_BEGIN
 
 /**
- *  You can use this class to specify information that you've already collected from your user. You can then set the `prefilledInformation` property on `STPPaymentContext`, `STPAddCardViewController`, etc and it will pre-fill this information whenever possible.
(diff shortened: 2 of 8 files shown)


Tests/Tests/STPAddSourceViewControllerTest.m
```diff
@@ -216,13 +238,13 @@ - (void)testCard_nextWithCreateSourceSuccessAndDidCreateSourceSuccess {
...
 
 - (void)testSEPA_nextWithCreateSourceError {
-    STPAddSourceViewController *sut = [self buildAddSourceViewControllerWithType:STPSourceTypeSEPADebit];
+    STPUserInformation *info = [self buildUserInfoForSEPATests];
+    STPAddSourceViewController *sut = [self buildAddSourceViewControllerWithType:STPSourceTypeSEPADebit
+                                                                        userInfo:info];
     id mockAPIClient = OCMClassMock([STPAPIClient class]);
     sut.apiClient = mockAPIClient;
-    sut.ibanCell.contents = @"GB82WEST12345698765432";
+    sut.ibanCell.contents = @"DE89370400440532013000";
```
bg-stripe: had to change these to IBANs to match the `sepaAddress` fixture, which is a German address (this line was changing the expected address's country to "GB")

Tests/Tests/STPAddSourceViewControllerTest.m
```diff
@@ -345,6 +370,18 @@ - (void)testSEPA_nextWithCreateSourceSuccessAndDidCreateSourceSuccess {
     [self waitForExpectationsWithTimeout:2 handler:nil];
 }
 
+- (void)testSettingIBANUpdatesCountry {
```
bg-stripe: added this new test after noticing the `ibanCell` behavior above

Stripe/PublicHeaders/STPUserInformation.h
```diff
@@ -38,6 +42,18 @@ NS_ASSUME_NONNULL_BEGIN
...
+ *  Metadata associated with your user. This will be attached to any Source
+ *  objects created by `STPPaymentContext`, `STPAddSourceViewController`, etc.
+ *  You should consider storing any order information (e.g., order number) here.
+ *  For payment methods that require additional user action, your backend will
+ *  need to listen to the `source.chargeable` webhook to create a charge request.
+ *  You can retrieve this metadata from the webhook event, and use it to fulfill
+ *  your customer's order.
+ *  https://stripe.com/docs/sources#best-practices
+ */
+@property(nonatomic, copy, nullable)NSDictionary *metadata;
```
bdorfman-stripe: Should add a comment about what types of objects are valid here. Maybe even do `NSDictionary<NSString *, id> *` at minimum, or enforce all strings only.

bg-stripe: 👍 updated

bdorfman-stripe: Should add to the STPUserInfo one also?

bg-stripe: oh whoops good catch

bg-stripe:
Updated with generic annotations – also realized we hadn't updated `copyWithZone` in SourceParams, and weren't copying `billingAddress`.

Stripe/PublicHeaders/STPSourceParams.h
```diff
@@ -49,25 +49,25 @@ NS_ASSUME_NONNULL_BEGIN
...
  */
-@property (nonatomic, copy, nullable) NSDictionary *redirect;
+@property (nonatomic, copy, nullable) NSDictionary<NSString *, id>*redirect;
 
 /**
  *  An optional token used to create the source. When passed, token properties will override 
  *  source parameters.
  */
-@property (nonatomic, copy, nullable) NSString *token;
+@property (nonatomic, copy, nullable) NSDictionary<NSString *, id>*token;
```
bdorfman-stripe: I don't think you meant to change this?

Stripe/PublicHeaders/STPSourceParams.h
```diff
@@ -49,25 +49,25 @@ NS_ASSUME_NONNULL_BEGIN
...
  */
-@property (nonatomic, copy, nullable) NSDictionary *owner;
+@property (nonatomic, copy, nullable) NSDictionary<NSString *, id>*owner;
 
 /**
  *  Parameters required for the redirect flow. Required if the source is authenticated by 
  *  a redirect (`flow` is "redirect").
  */
-@property (nonatomic, copy, nullable) NSDictionary *redirect;
+@property (nonatomic, copy, nullable) NSDictionary<NSString *, id>*redirect;
```
bdorfman-stripe: Why id instead of NSString for the values? Are there non-string types we need to support? I don't think our actual networking logic will work for any non-strings currently.

bg-stripe: These hashes can contain inner hashes (e.g. owner["address"]). I think that's the only example of an inner hash right now, but it seems better to be more permissive for everything except metadata

Stripe/PublicHeaders/STPSourceParams.h
bg-stripe: oh yikes – fixed

Stripe/PublicHeaders/STPSourceParams.h
bdorfman-stripe: Ah ok, makes sense.

bdorfman-stripe:
LGTM

PR #626, opened 2017-04-03
Update custom example to use redirect helper

bg-stripe:
r? @bdorfman-stripe 

I've also made STPRedirectHelper public, and updated the examples to display a label & disable elements instead of presenting an alert controller, as it looks a bit better.

### DIFF
--- a/Example/Stripe iOS Example (Custom)/SofortExampleViewController.m
+++ b/Example/Stripe iOS Example (Custom)/SofortExampleViewController.m
@@ -22,8 +22,8 @@
  */
 @interface SofortExampleViewController ()
 @property (nonatomic, weak) UIButton *payButton;
+@property (nonatomic, weak) UILabel *waitingLabel;
 @property (nonatomic, weak) UIActivityIndicatorView *activityIndicator;
-@property (nonatomic) STPSource *source;
 @end
 
 @implementation SofortExampleViewController
@@ -43,6 +43,14 @@
     self.payButton = button;
     [self.view addSubview:button];
 
+    UILabel *label = [UILabel new];
+    label.text = @"Waiting for payment authorization";
+    [label sizeToFit];
+    label.textColor = [UIColor grayColor];
+    label.alpha = 0;
+    [self.view addSubview:label];
+    self.waitingLabel = label;
+
     UIActivityIndicatorView *activityIndicator = [[UIActivityIndicatorView alloc] initWithActivityIndicatorStyle:UIActivityIndicatorViewStyleGray];
     activityIndicator.hidesWhenStopped = YES;
     self.activityIndicator = activityIndicator;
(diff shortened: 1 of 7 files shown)


Example/Stripe iOS Example (Custom)/SofortExampleViewController.m
```diff
@@ -79,31 +95,30 @@ - (void)pay {
...
+                                                                       [self.delegate exampleViewController:self didFinishWithMessage:@"Payment successfully created"];
+                                                                   } else if (source.status == STPSourceStatusFailed || source.status == STPSourceStatusCanceled) {
+                                                                       [self.delegate exampleViewController:self didFinishWithMessage:@"Payment failed"];
+                                                                   } else {
+                                                                       [self.delegate exampleViewController:self didFinishWithMessage:@"Order received"];
+                                                                   }
+                                                               }
+                                                           }];
+            }];
+            [redirect startSafariViewControllerRedirectFlowFromViewController:self];
```
bdorfman-stripe: Not sure we want to encourage people to not retain a reference to this context, although I think it might work currently because it retains itself for the length of the redirect.

Example/Stripe iOS Example (Custom)/SofortExampleViewController.m
```diff
@@ -79,31 +95,30 @@ - (void)pay {
...
+                [[STPAPIClient sharedClient] startPollingSourceWithId:sourceID
+                                                         clientSecret:clientSecret
+                                                              timeout:10
+                                                           completion:^(STPSource *source, NSError *error) {
+                                                               [self updateUIForPaymentInProgress:NO];
+                                                               if (error) {
+                                                                   [self.delegate exampleViewController:self didFinishWithError:error];
+                                                               } else {
+                                                                   if (source.status == STPSourceStatusConsumed ||
+                                                                       source.status == STPSourceStatusChargeable) {
```
bdorfman-stripe: Should use switch statement instead of ifs so we get errors if cases are missing.

Example/Stripe iOS Example (Custom)/ThreeDSExampleViewController.m
```diff
@@ -111,10 +128,30 @@ - (void)pay {
...
                     [self.delegate exampleViewController:self didFinishWithError:error];
                 } else {
-                    self.source = source;
-                    [self presentPollingUI];
-                    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(handleAppForeground) name:UIApplicationWillEnterForegroundNotification object:nil];
-                    [[UIApplication sharedApplication] openURL:source.redirect.url];
+                    // In order to use STPRedirectContext, you'll need to set up
+                    // your app delegate to forwards URLs to the Stripe SDK.
+                    // See `[Stripe handleStripeURLCallback:]`
+                    STPRedirectContext *redirect = [[STPRedirectContext alloc] initWithSource:source completion:^(NSString *sourceID, NSString *clientSecret, NSError *error) {
```
bdorfman-stripe: Same comments as above on redirect.

Example/Stripe iOS Example (Custom)/SofortExampleViewController.m
bg-stripe: ah yep, will change

Example/Stripe iOS Example (Custom)/SofortExampleViewController.m
bg-stripe: good point, better to retain the context in these examples – will update

bg-stripe:
> Also not sure if we should be calling cancel on the redirect anywhere if the user backs out of the screen? Or is the UI blocked in such a way that they can't do that?

We're blocking the navigation bar now when the payment is in progress, so it won't be possible to back out of the page.

bg-stripe:
Updated in afe776e

bdorfman-stripe:
LGTM

PR #627, opened 2017-04-04
Remove all references to credit cards.

bdorfman-stripe:
In favor of just "cards"


bdorfman-stripe:
Ah yeah. I always forget that test compile errors don't show up if you just hit build which is irritating. I'll fix.

PR #628, opened 2017-04-04
always require confirmation for iDEAL bank

bg-stripe:
r? @bdorfman-stripe 

As discussed with Stan yesterday. Also updated tests.


bdorfman-stripe:
👍

PR #629, opened 2017-04-04
Treat polling error as PaymentStatusPending

bg-stripe:
r? @bdorfman-stripe 

We now treat a polling error as `PaymentStatusPending`. Also updated docs with some additional notes about webhooks and the pending status.

### DIFF
--- a/Stripe/STPPaymentContext.m
+++ b/Stripe/STPPaymentContext.m
@@ -568,7 +568,7 @@ typedef NS_ENUM(NSUInteger, STPPaymentContextState) {
     STPSourceCompletionBlock onRedirectCompletion = ^(STPSource *finishedSource, NSError *error) {
         stpDispatchToMainThreadIfNecessary(^{
             if (error) {
-                [self didFinishWithStatus:STPPaymentStatusError error:error];
+                [self didFinishWithStatus:STPPaymentStatusPending error:nil];
             } else {
                 switch (finishedSource.status) {
                     case STPSourceStatusChargeable:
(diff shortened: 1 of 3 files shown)


Stripe/STPPaymentContext.m
```diff
@@ -568,7 +568,7 @@ - (void)requestSourceFlowRedirectPayment:(STPSource *)source {
     STPSourceCompletionBlock onRedirectCompletion = ^(STPSource *finishedSource, NSError *error) {
         stpDispatchToMainThreadIfNecessary(^{
             if (error) {
-                [self didFinishWithStatus:STPPaymentStatusError error:error];
+                [self didFinishWithStatus:STPPaymentStatusPending error:nil];
```
bdorfman-stripe: Are there any categories of error that we want to return as StatusError here? Is it possible to call into this and never actually present the user the dialog in a way we can know about? (eg if there is no network connection).

bg-stripe: Hm, I think we should treat all polling errors equally here – the source has been created at this point, and the user may have authorized the payment, but we don't know because there was an error. Even if there was a connectivity error while polling, it's still possible the user authorized the payment before they went offline.

bdorfman-stripe: I was thinking more of errors that occur before polling. Eg. Before we even attempt the redirect we could know there is not network. SFVC also tells you if the initial page load failed? I think it's fine to do pending here though. Maybe we should just think about building some better network offline/retry logic into the pipeline somewhere.

bg-stripe: ah, I see – I forgot that an SFVC error would end up here too. I think it's ok for now for an SFVC loading failure to result in a Pending status, but we should definitely try to fix that. Maybe we should have an STPRedirectContext error domain and check for that here – added to our todos.

bdorfman-stripe:
LGTM

PR #631, opened 2017-04-06
Add unit tests for requestPayment

bg-stripe:
r? @bdorfman-stripe 

This adds some unit tests for requestPayment, verifying that we do the right thing for different selected payment methods. I'm especially proud of the iDEAL test, which was tricky – it goes all to way to verifying that SFSafariVC gets presented after completing SourceInfoVC.

Next up, I'm going to add some unit tests for STPRedirectContext.

I'd also like to add some basic tests for PaymentMethodsVC, making sure it chooses the right `internalViewController` for a given configuration.

### DIFF
--- a/Stripe/STPPaymentContext.m
+++ b/Stripe/STPPaymentContext.m
@@ -29,19 +29,6 @@
 
 #define FAUXPAS_IGNORED_IN_METHOD(...)
 
-/**
- The current state of the payment context
-
- - STPPaymentContextStateNone: No view controllers are currently being shown. The payment may or may not have already been completed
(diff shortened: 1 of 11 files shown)


Stripe/STPPaymentContext.m
```diff
@@ -484,7 +471,7 @@ - (void)requestPayment {
         if (!self.selectedPaymentMethod) {
             [self presentPaymentMethodsViewControllerWithNewState:STPPaymentContextStateRequestingPayment];
         }
-        else if (self.configuration.requiredShippingAddressFields != STPBillingAddressFieldsNone &&
```
bg-stripe: yeah, i guess you can compare arbitrary enums if they're the same underlying type :(

Issue #632, opened 2017-04-06
STPCardValidator validationStateForNumber: is incorrect for "6"


chillpop:
## Summary

STPCardValidator validatationStateForNumber returns STPCardValidationStateInvalid for an input of "6" when validating card brand.  This should be STPCardValidationStateIncomplete, as it could be a Discover card.

## Code to reproduce

```swift
STPCardValidator.validationState(forNumber: "6", validatingCardBrand: true)
```
## iOS version

10.3

## Installation method

Cocoapods

## SDK version

10.0.1

bdorfman-stripe:
Thanks for the report! I agree that this is definitely wrong currently. We will work on adding a fix for this issue in our next release.

PR #633, opened 2017-04-07
Fix bin range testing.

bdorfman-stripe:
Previously partial numbers did not match to any ranges (eg 6 did not match to 622). This meant that the validator sometimes incorrectly said things were invalid when they were in fact incomplete and would also not correctly return `brandForNumber:` for these partial matches (in the case where they could only possibly complete to single brand). I changed this logic so partial numbers match bin ranges correctly and updated tests.

Also added a bunch of tests for hypthetical bin ranges which start with 0, which I expect is the most likely sort of future bug in this type of numeric comparison.

Fixes: https://github.com/stripe/stripe-ios/issues/632

### DIFF
--- a/Tests/Tests/STPBinRangeTest.m
+++ b/Tests/Tests/STPBinRangeTest.m
@@ -37,27 +37,64 @@
     binRange.qRangeHigh = @"167";
     
     XCTAssertFalse([binRange matchesNumber:@"0"]);
-    XCTAssertFalse([binRange matchesNumber:@"1"]);
+    XCTAssertTrue([binRange matchesNumber:@"1"]);
     XCTAssertFalse([binRange matchesNumber:@"2"]);
-    
+
+    XCTAssertFalse([binRange matchesNumber:@"00"]);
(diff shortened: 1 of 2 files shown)


Tests/Tests/STPBinRangeTest.m
```diff
@@ -95,11 +132,11 @@ - (void)testBinRangesForBrand {
     }
 }
 
-- (void)testMostSpecifiBinRangeForNumber {
+- (void)testMostSpecificBinRangeForNumber {
     STPBINRange *binRange;
     
     binRange = [STPBINRange mostSpecificBINRangeForNumber:@""];
-    XCTAssertEqual(binRange.brand, STPCardBrandUnknown);
+    XCTAssertNotEqual(binRange.brand, STPCardBrandUnknown);
```
bdorfman-stripe: FYI I am not even sure this test makes sense and may remove it? The most specific bin range for no string could be anything. Also the returned value for mostSpecific when more than one possible range has the same specific-ness is undefined.

jack-stripe:
lgtm, thanks!

PR #634, opened 2017-04-07
Add unit tests for STPRedirectContext

bg-stripe:
r? @bdorfman-stripe 

Learned some things while writing these. I initially wanted to verify that `dismiss` was called on the view controller presenting `SFSafariVC`, i.e.

```
OCMVerify([mockVC dismissViewControllerAnimated:YES completion:[OCMArg any]]);
```

but ran into some issues. Our unit tests aren't embedded in an application, which means we can't actually present a remote view controller (we get an error about the presenting view controller not being in the window hierarchy). To get around this, I started going down the route of mocking `SFSafariVC` and stubbing `presentingViewController`, but in the end decided that the approach here is cleaner. We partially mock `STPRedirectContext` and verify that an internal method, `unsubscribeFromNotifications...` is called. It's generally not great to leak the internals of the subject under test, but I think it's acceptable in this case.


bdorfman-stripe:
LGTM

PR #635, opened 2017-04-07
3DS payment context support

bdorfman-stripe:


Stripe/PublicHeaders/STPPaymentConfiguration.h
```diff
@@ -14,6 +14,21 @@
...
 NS_ASSUME_NONNULL_BEGIN
 
+
+/**
+ Possible values for 3DS source support
+
+ - STPThreeDSecureSupportTypeDisabled: 3DS sources will  never be created
+ - STPThreeDSecureSupportTypeStatic: An attempt to create 3DS sources will be done for all card sources that may possibly support 3ds
+   A failure to create a 3DS source for a card that does not require 3DS will result in the original card source being passed through 
+   to your backend API adapter for charging. If you want to forbid all non-3DS payments, you should not charge and throw an error
```
bg-stripe: the original card source gets returned in `didCreatePaymentResult:` – PaymentContext doesn't actually hook into your backend API adapter to tell it to charge.

bg-stripe:
(added a single comment by accident – deleted and batching in a review now)

Stripe/PublicHeaders/STPPaymentConfiguration.h
```diff
@@ -14,6 +14,21 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+
+/**
```
bg-stripe: I think we should go into a bit more detail here, and frame it from the perspective of the user's experience.

> Options for 3D Secure support.
> 
> STPThreeDSecureSupportTypeDisabled: Your customer will never be prompted for 3DS verification.
> 
> STPThreeDSecureSupportTypeStatic: Your customer will be prompted for 3DS verification if available. If 3DS is not available, the original card source will be passed to your app. Note that if you charge a card source, you will not be protected from fraud by 3DS. If you want to reject all card payments that haven't passed 3DS verification, you should update your logic accordingly in `paymentContext:didCreatePaymentResult:completion:`: if the payment result is a card source, instead of passing the source to your backend to be charged, you should call the completion block with an error.
>

Stripe/PublicHeaders/StripeError.h
```diff
@@ -83,6 +83,7 @@ FOUNDATION_EXPORT STPCardErrorCode __nonnull const STPIncorrectCVC;
 + (nonnull NSString *)stp_cardErrorExpiredCardUserMessage;
 + (nonnull NSString *)stp_cardErrorDeclinedUserMessage;
 + (nonnull NSString *)stp_cardErrorProcessingErrorUserMessage;
+    + (nonnull NSString *)stp_cardErrorPaymentMethodNotAvailableUserMessage;
```
bg-stripe: nit: extra indentation

Stripe/STPPaymentContext.m
```diff
@@ -554,6 +579,103 @@ - (void)requestSourceFlowNonePayment:(STPSource *)source {
...
+                case STPSourceStatusChargeable:
+                    // Success, we can just finish
+                    // (charge will happen on webhook)
+                    [self didFinishWithStatus:STPPaymentStatusSuccess
+                                        error:nil];
+                    break;
+                case STPSourceStatusUnknown:
+                case STPSourceStatusCanceled:
+                    [self didFinishWithStatus:STPPaymentStatusError
+                                        error:nil]; // TODO: Create STPPaymentContext error domain
```
bg-stripe: We've created a few of these STPPaymentContext error TODOs – mind adding one in this PR so we can start knocking them out?

bdorfman-stripe: I think creating a new error domain and all the associated codes is kind of a big diff. Was putting all these todos in so we can easily find where to go back and fill them in, but I'd like to do that in a separate PR instead of grouping it with this one.

bg-stripe: that's fair, fine with doing it later!

PR #636, opened 2017-04-10
Add unit tests for STPPaymentMethodsVC

bg-stripe:
r? @bdorfman-stripe 

Caught a bug while writing these (da8f06c) – we were calling the wrong delegate method when canceling AddSourceVC.

### DIFF
--- a/Stripe/STPPaymentMethodsViewController.m
+++ b/Stripe/STPPaymentMethodsViewController.m
@@ -228,8 +228,6 @@
 }
 
 - (void)addCardViewControllerDidCancel:(__unused STPAddCardViewController *)addCardViewController {
-    // Add card is only our direct delegate if there are no other payment methods possible
-    // and we skipped directly to this screen. In this case, a cancel from it is the same as a cancel to us.
     [self.delegate paymentMethodsViewControllerDidCancel:self];
 }
 
@@ -240,7 +238,7 @@
 }
 
 - (void)addSourceViewControllerDidCancel:(__unused STPAddSourceViewController *)addSourceViewController {
-    [self.delegate paymentMethodsViewControllerDidFinish:self];
+    [self.delegate paymentMethodsViewControllerDidCancel:self];
 }
 
 - (void)addSourceViewController:(__unused STPAddSourceViewController *)addSourceViewController
--- a/Stripe/STPPaymentMethodsViewController.m
+++ b/Stripe/STPPaymentMethodsViewController.m
@@ -228,6 +228,8 @@
 }
 
 - (void)addCardViewControllerDidCancel:(__unused STPAddCardViewController *)addCardViewController {
+    // AddCardVC is only our direct delegate if there are no other payment methods possible
+    // and we skipped directly to this screen. In this case, a cancel from it is the same as a cancel to us.
     [self.delegate paymentMethodsViewControllerDidCancel:self];
 }
 
@@ -238,6 +240,8 @@
 }
 
 - (void)addSourceViewControllerDidCancel:(__unused STPAddSourceViewController *)addSourceViewController {
+    // AddSourceVC is only our direct delegate if there are no other payment methods possible
+    // and we skipped directly to this screen. In this case, a cancel from it is the same as a cancel to us.
     [self.delegate paymentMethodsViewControllerDidCancel:self];
 }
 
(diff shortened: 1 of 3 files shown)


Stripe/STPPaymentMethodsViewController.m
```diff
@@ -228,8 +228,6 @@ - (void)internalViewControllerDidCancel {
 }
 
 - (void)addCardViewControllerDidCancel:(__unused STPAddCardViewController *)addCardViewController {
-    // Add card is only our direct delegate if there are no other payment methods possible
-    // and we skipped directly to this screen. In this case, a cancel from it is the same as a cancel to us.
```
bdorfman-stripe: Why delete this comment?

Tests/Tests/STPPaymentMethodsViewControllerTest.m
```diff
@@ -0,0 +1,237 @@
...
+    id<STPPaymentMethodsViewControllerDelegate>delegate = OCMProtocolMock(@protocol(STPPaymentMethodsViewControllerDelegate));
+    STPPaymentMethodsViewController *sut = [self buildViewControllerWithCustomer:customer
+                                                                   configuration:config
+                                                                        delegate:delegate];
+    XCTAssertTrue([sut.internalViewController isKindOfClass:[STPAddSourceViewController class]]);
+}
+
+/**
+ When the customer has no sources and the configuration uses sources,
+ STPPaymentMethodsInternalVC should be shown.
```
bdorfman-stripe: Description here is same as previous test?

Tests/Tests/STPPaymentMethodsViewControllerTest.m
```diff
@@ -0,0 +1,237 @@
...
+                                                                   configuration:config
+                                                                        delegate:delegate];
+    XCTAssertTrue([sut.internalViewController isKindOfClass:[STPAddCardViewController class]]);
+}
+
+/**
+ When the customer has no sources and the configuration uses sources,
+ STPAddSourceViewController should be shown.
+ */
+- (void)testInitWithNoSourcesAndUseSourcesOn {
```
bdorfman-stripe: Not sure these tests are exactly correct. Whether these screens show or not also depends on what your configuration available payment methods are. Eg. you don't go to these screens even if customer has no sources if there are multiple possible configs (like Apple Pay and Cards). Also even if useSouresForCards = NO you would still to add source if your only option was a new resuable source (like SEPA)

Tests/Tests/STPPaymentMethodsViewControllerTest.m
bg-stripe: ah whoops, good catch

Tests/Tests/STPPaymentMethodsViewControllerTest.m
bg-stripe: ah yeah, that's a good point – I'll explicitly set up availablePaymentMethods in the config, and add some additional cases.

Stripe/STPPaymentMethodsViewController.m
bg-stripe: ah, didn't mean to – I'll re-add here and in `addSource...DidCancel`

Tests/Tests/STPPaymentMethodsViewControllerTest.m
bg-stripe: actually slightly different :P – `STPPaymentMethodsInternalVC`, rather than `STPAddSourceViewController`

bdorfman-stripe: Right, but I mean the rest of the description is the same so something there is probably wrong? :)

bg-stripe:
@bdorfman-stripe  updated in 5d1708

Tests/Tests/STPPaymentMethodsViewControllerTest.m
bg-stripe: Oh my bad, totally missed that. I'll update

bdorfman-stripe:
Looks good. Although the Apple Pay test is gonna break as soon as I build in the filtering logic (which will remove it from the configured sources if Apple Pay is not allowed on device, which it won't be in the simulator :\ )

PR #638, opened 2017-04-12
Fix some requestPayment tests

bg-stripe:
r? @bdorfman-stripe 

There are two different approaches to mocking, both supported by OCMock:
1. `expect-run-verify`
- declare expected method calls on your mock with `OCMExpect([mock doSomething])`
- exercise the test
- verify that expected methods were called with `OCMVerifyAll(mock)`.

2. `verify-after-running`
- exercise the test
- verify expected method calls with `OCMVerify([mock doSomething])`

`verify-after-running` is simpler, and we should be consistent about using one approach in our tests, never using `OCMExpect`. I'll also make a note of this in the doc I'm writing.

I also realized the SourceInfo -> SafariVC test I wrote was actually a false positive. The counter approach doesn't actually work (the counter doesn't get incremented in subsequent calls – not sure why). I've updated the test to fulfill two expectations, rather than incrementing a counter.


bdorfman-stripe:
LGTM

PR #639, opened 2017-04-13
Remove remember me

bdorfman-stripe:


Stripe/STPAddCardViewController.m
```diff
@@ -120,12 +98,8 @@ - (void)createAndSetupViews {
     cardImageView.frame = CGRectMake(0, 0, self.view.bounds.size.width, cardImageView.bounds.size.height + (57 * 2));
     self.cardImageView = cardImageView;
     self.tableView.tableHeaderView = cardImageView;
-    self.emailCell = [[STPRememberMeEmailCell alloc] initWithDelegate:self];
-    if ([STPEmailAddressValidator stringIsValidEmailAddress:self.prefilledInformation.email]) {
-        self.emailCell.contents = self.prefilledInformation.email;
```
bg-stripe: We should also remove `phone` and `email` from `STPUserInformation`. 
They also get used in ShippingAddressVC, but I think we should just remove that behavior – the fields are there to support auto-filling remember me fields, and are only used in ShippingAddressVC for consistency.

bdorfman-stripe: 👍

bdorfman-stripe: Hmm, actually I dunno, it seems potentially useful for people to have the ability to prefill this stuff in shipping address form?

bg-stripe: That's fair – I'm fine with keeping them around.

PR #640, opened 2017-04-14
isValid -> KVO observable

fbernardo:
## Summary

Enabled STPPaymentCardTextField and STPPaymentCardTextFieldViewModel isValid properties to be KVO observable.

## Motivation

Before this PR if the user wanted to know when the isValid property changed the only way was to assign a delegate to STPPaymentCardTextField, implement the did change method and check isValid. In some situations KVO observing is faster and leads to better code than delegation.

## Testing

I just tested it by running (Xcode 8.3.1, iOS 10.3 Simulator) and it's working.

### DIFF
--- a/Stripe/STPPaymentCardTextField.m
+++ b/Stripe/STPPaymentCardTextField.m
@@ -48,7 +48,7 @@
 @synthesize borderColor = _borderColor;
 @synthesize borderWidth = _borderWidth;
 @synthesize cornerRadius = _cornerRadius;
-@dynamic enabled, isValid, valid;
+@dynamic enabled;
 
 CGFloat const STPPaymentCardTextFieldDefaultPadding = 13;
 
--- a/Stripe/STPPaymentCardTextField.m
+++ b/Stripe/STPPaymentCardTextField.m
@@ -48,7 +48,7 @@
 @synthesize borderColor = _borderColor;
 @synthesize borderWidth = _borderWidth;
 @synthesize cornerRadius = _cornerRadius;
-@dynamic enabled;
+@dynamic enabled, isValid, valid;
 
 CGFloat const STPPaymentCardTextFieldDefaultPadding = 13;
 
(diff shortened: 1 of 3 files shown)


Stripe/STPPaymentCardTextField.m
```diff
@@ -48,7 +48,7 @@ @implementation STPPaymentCardTextField
 @synthesize borderColor = _borderColor;
 @synthesize borderWidth = _borderWidth;
 @synthesize cornerRadius = _cornerRadius;
-@dynamic enabled;
+@dynamic enabled, isValid, valid;
```
bg-stripe: I don't think you need to declare `isValid` and `valid` as dynamic – the accessors are implemented here, not provided dynamically at runtime.

Stripe/STPPaymentCardTextFieldViewModel.m
```diff
@@ -12,6 +12,7 @@
 #define FAUXPAS_IGNORED_IN_METHOD(...)
 
 @implementation STPPaymentCardTextFieldViewModel
+@dynamic valid;
```
bg-stripe: same here – i don't think this @dynamic declaration is necessary, since we implement the getter

fbernardo:
@bg-stripe Thank you for your comments. Dynamic's removed. 👍

bdorfman-stripe:
It would be good to add some tests to make sure this stays working correctly. As well as adding comments to the implementations of `isValid` and `validationStateForField` to note that if you change the properties this field is computed from that you should update `keyPathsForValuesAffecting` as well, since it will be easy to forget.

Also the property on `STPPaymentCardTextFieldViewModel` should probably just be `isValid` directly (instead of `valid` with a customized getter) to match `STPPaymentCardTextField`.

bg-stripe:
Since writing a unit test for this is a bit non-trivial, I'm migrating this PR to #645 , where I've added a unit test and changed the view model's property to a readonly `isValid`.

fbernardo:
Thank _you_ @bg-stripe!

PR #643, opened 2017-04-14
Improve error parsing

bg-stripe:
r? @bdorfman-stripe 
cc @mrmcduff-stripe 

When creating card sources, validation errors have type `invalid_request`. This means we aren't currently localizing these errors, or returning an appropriate `STPCardErrorCode` value.

I've made some improvements to StripeError, and added tests:
- we now include the original error code and type from the Stripe API in `userInfo`
- we always attempt to localize the error message, as long as the type is not `api_error`. For `api_error`s, we provide a generic unexpected error message (I'm just preserving existing logic here – I think the rationale is that we don't want these error messages to be displayed to end users).
- we don't require that the error response has a message (it's optional in our docs)


bdorfman-stripe:
LGTM

PR #644, opened 2017-04-17
Add travis_retry

bg-stripe:
r? @bdorfman-stripe 

There's some discussion of people running into the same simulator warmup / XCTest timeout issue on Travis here: https://github.com/travis-ci/travis-ci/issues/6422

I think we can start with adding `travis_retry` before the flaking tests. If flakiness continues, we can try separating the build and test commands: http://stackoverflow.com/questions/37922146/xctests-failing-on-physical-device-canceling-tests-due-to-timeout/40790171#40790171

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,9 +25,9 @@ script:
 - "./ci_scripts/check_public_headers.rb"
 - "./ci_scripts/check_resource_bundle.rb"
 - '[ "$TEST_TYPE" != lint ] || ./ci_scripts/check_fauxpas.sh'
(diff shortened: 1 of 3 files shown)


jack-stripe:
Incidentally, I have used the strategy described in that StackOverflow
answer before and it has worked for me.

bg-stripe:
Cool – I've updated this so that we use the new `build-for-testing` / `test-without-building` commands, as well as `travis_retry`.

bdorfman-stripe:
Many 👍

PR #645, opened 2017-04-17
isValid KVO

bg-stripe:
r? @bdorfman-stripe 

Just helping get #640 through the door, as writing a KVO unit test is kinda hard.

Thanks for the contribution @fbernardo !


bdorfman-stripe:
👍

PR #646, opened 2017-04-17
Pass up app name for redirect screen in returnURL

bdorfman-stripe:
When using native urls you land on a hooks.stripe page that says "Return to Merchant" after completing redirect charges. This passes up the name of the app/company as set in pay config to style that as "Return to [App Name]" instead.

### DIFF
--- a/Stripe/STPSourceParams.m
+++ b/Stripe/STPSourceParams.m
@@ -7,6 +7,7 @@
 //
 
 #import "STPSourceParams.h"
+#import "STPSourceParams+Private.h"
 
 #import "STPCardParams.h"
 #import "STPFormEncoder.h"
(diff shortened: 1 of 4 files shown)


Stripe/STPSourceParams.m
```diff
@@ -202,6 +203,39 @@ + (STPSourceParams *)threeDSecureParamsWithAmount:(NSUInteger)amount
...
+        NSURLComponents *urlComponents = [NSURLComponents componentsWithURL:[NSURL URLWithString:self.redirect[@"return_url"]]
+                                                    resolvingAgainstBaseURL:NO];
+        NSMutableArray<NSURLQueryItem *> *queryItems = (urlComponents.queryItems ?: @[]).mutableCopy;
+        [queryItems addObject:[NSURLQueryItem queryItemWithName:@"redirect_merchant_name"
+                                                          value:self.redirectMerchantName]];
+        urlComponents.queryItems = queryItems;
+
+
+        NSMutableDictionary *redirectCopy = self.redirect.mutableCopy;
+        redirectCopy[@"return_url"] = urlComponents.URL;
```
bg-stripe: do we have to convert this to a string?

bdorfman-stripe: No, because everything gets converted to a string using `description` later, but I can manually convert if you want.

bdorfman-stripe:
It's not actually an error to not have a company name, so I'm not sure we should test to make sure it's in there. I can add a test to make sure it's getting forwarded through from payconfig if it should be though.

bg-stripe:
Ah yep, sorry my comment was confusing. I just meant that I'd expect the existing PaymentContext iDEAL request payment test to fail from adding these changes, but it didn't because the config we use in that test doesn't set a company name.
https://github.com/stripe/stripe-ios/blob/sources-ui/Tests/Tests/STPPaymentContextRequestPaymentTest.m#L120

After adding a company name there, the test should break because the `returnURL` doesn't match the expected SourceParams
https://github.com/stripe/stripe-ios/blob/sources-ui/Tests/Tests/STPPaymentContextRequestPaymentTest.m#L142

bdorfman-stripe:
Ah makes sense. I'll tweak the tests.

bdorfman-stripe:
Closing in favor of https://github.com/stripe/stripe-ios/pull/647

PR #647, opened 2017-04-18
Pass up app name for redirect screen in returnURL

bdorfman-stripe:


Tests/Tests/STPSourceParamsTest.m
```diff
@@ -46,4 +48,29 @@ - (void)testCardParamsWithCard {
...
+                                                        statementDescriptor:nil];
+
+    NSDictionary *params = [STPFormEncoder dictionaryForObject:sourceParams];
+    // Should be nil because we have no app name in tests
+    XCTAssertNil([self redirectMerchantNameQueryItemValueFromURLString:params[@"redirect"][@"return_url"]]);
+
+    sourceParams.redirectMerchantName = @"bar";
+    params = [STPFormEncoder dictionaryForObject:sourceParams];
+    XCTAssertEqualObjects([self redirectMerchantNameQueryItemValueFromURLString:params[@"redirect"][@"return_url"]], @"bar");
+}
```
bg-stripe: could we also add a test that makes sure we don't overwrite an existing `redirect_merchant_name` parameter?

bdorfman-stripe:
Updated

PR #649, opened 2017-04-20
Fix shipping methods state bug

bg-stripe:
r? @bdorfman-stripe 

To reproduce:
- Tap the shipping row
- Enter a shipping address / method
- PaymentContext's state hasn't been reset, so you get stuck

### DIFF
--- a/Stripe/STPPaymentContext.m
+++ b/Stripe/STPPaymentContext.m
@@ -426,6 +426,8 @@ typedef NS_ENUM(NSUInteger, STPPaymentContextState) {
         if (self.state == STPPaymentContextStateRequestingPayment) {
             self.state = STPPaymentContextStateNone;
             [self requestPayment];
+        } else {
+            self.state = STPPaymentContextStateNone;
         }
     }];
 }


bdorfman-stripe:
Hmm, this seems wrong because you are resetting the state before we are reading it later and so the requestPayment flow will never continue. I think you want to add an else block inside the dismiss completion to reset to None.

bg-stripe:
ah whoops, my bad – updated

Stripe/STPPaymentContext.m
```diff
@@ -424,9 +424,9 @@ - (void)shippingAddressViewController:(STPShippingAddressViewController *)addres
     [self.delegate paymentContextDidChange:self];
     [self appropriatelyDismissViewController:addressViewController completion:^{
         if (self.state == STPPaymentContextStateRequestingPayment) {
-            self.state = STPPaymentContextStateNone;
             [self requestPayment];
         }
+        self.state = STPPaymentContextStateNone;
```
bdorfman-stripe: Should be an `else` here, or otherwise you might clobber some state set up in request payment

bdorfman-stripe:
LGTM

PR #651, opened 2017-04-21
Add 3DS tests

bg-stripe:
r? @bdorfman-stripe 

I've also updated the `payment_method_not_available` error logic.

### DIFF
--- a/Stripe/STPPaymentContext.m
+++ b/Stripe/STPPaymentContext.m
@@ -598,6 +598,7 @@
                                                                                currency:self.paymentCurrency
                                                                               returnURL:self.configuration.returnURLBlock().absoluteString
                                                                                    card:cardSource.stripeID];
+    threeDSecureParams.metadata = self.sourceInformation.metadata;
 
     STPSourceCompletionBlock threeDSSourceCompletion = ^(STPSource * _Nullable threeDSSource, NSError * _Nullable error) {
         if (error
@@ -605,7 +606,6 @@
             || threeDSSource.flow != STPSourceFlowRedirect) {
             if (!threeDSecureRequired
                 && error.domain == StripeDomain
-                && error.code == STPCardError
                 && [error.userInfo[STPCardErrorCodeKey] isEqualToString:STPPaymentMethodNotAvailable]) {
                 // This is a card that doesn't actually support 3ds and the card source didn't say 3ds was required
                 // So we can just return the original card source to be charged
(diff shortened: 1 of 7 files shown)


Stripe/STPPaymentContext.m
```diff
@@ -598,15 +598,14 @@ three_d_secure types of required or optional (not_supported are filtered earlier
...
                                                                               returnURL:self.configuration.returnURLBlock().absoluteString
                                                                                    card:cardSource.stripeID];
+    threeDSecureParams.metadata = self.sourceInformation.metadata;
 
     STPSourceCompletionBlock threeDSSourceCompletion = ^(STPSource * _Nullable threeDSSource, NSError * _Nullable error) {
         if (error
             || threeDSSource == nil
             || threeDSSource.flow != STPSourceFlowRedirect) {
             if (!threeDSecureRequired
-                && error.domain == StripeDomain
```
bdorfman-stripe: I think you should still check that the error is a StripeDomain here?

Stripe/STPPaymentContext.m
```diff
@@ -598,15 +598,14 @@ three_d_secure types of required or optional (not_supported are filtered earlier
...
-                && error.code == STPCardError
-                && [error.userInfo[STPCardErrorCodeKey] isEqualToString:STPPaymentMethodNotAvailable]) {
+                && [error.userInfo[STPStripeErrorCodeKey] isEqualToString:@"payment_method_not_available"]) {
```
bdorfman-stripe: I think all the codes should still be converted to constants. I'm not a huge fan of just checking for the raw string values. Is this not still a card error? (given that 3ds sources are cards). At the least we should define string constants for the errors we check for I think.

Stripe/STPPaymentContext.m
bg-stripe: 👍

Stripe/STPPaymentContext.m
bg-stripe: That's fair. Source errors have type `invalid_request`, not `card_error`, so I'm a bit unhappy with putting this under `STPCardErrorCodeKey`, but I think it's probably easier to just do that.

bdorfman-stripe: I'm fine with not putting it under card_error I guess. If you want to just define an NSString const that is that expected error though? Not 100% sure of the Swift bridging implications of it being in that enumeration or not though.

bg-stripe:
updated!

bdorfman-stripe:
👍

PR #652, opened 2017-04-21
Reorganize Xcodeproj file groupings

bdorfman-stripe:
This is effectively impossible to diff textually. I recommend just opening this branch locally and looking at the new groupings and subgroupings to see if they all make sense to you.


bg-stripe:
👍

PR #653, opened 2017-04-24
Add STPTelemetryClient

bg-stripe:
r? @bdorfman-stripe 
cc @matt-stripe


Stripe/STPTelemetryClient.m
```diff
@@ -0,0 +1,190 @@
...
+                [fields addObject:@[[self screenSize]]];
+                break;
+            case STPMetricFieldTimeZoneOffset:
+                [fields addObject:@[[self timeZoneOffset]]];
+                break;
+            case STPMetricFieldUserAgent:
+                [fields addObject:@[[self userAgent]]];
+                break;
+            default:
+                [fields addObject:@[@""]];
```
jbalogh-stripe: Can you make this an empty array? That way we won't try to store the empty string.

Stripe/STPTelemetryClient.m
```diff
@@ -0,0 +1,190 @@
...
+                break;
+        }
+    }
+    payload[@"f"] = fields;
+    payload[@"d"] = @[
+                      @"",
+                      @{@"fr": @"",
```
jbalogh-stripe: Same here. The `fr` and `ft` params can be non-existent rather than being empty strings. The first param above does need to be an empty string though, just to keep the positions straight.

Stripe/STPTelemetryClient.m
bg-stripe: yep!

Stripe/STPTelemetryClient.m
bg-stripe: 👍

Stripe/STPAnalyticsClient.m
```diff
@@ -91,28 +89,20 @@ + (void)initializeIfNeeded {
...
-+ (void)disableAnalytics {
-    STPAnalyticsCollectionDisabled = YES;
-}
-
 + (BOOL)shouldCollectAnalytics {
 #if TARGET_OS_SIMULATOR
     return NO;
 #endif
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wunreachable-code"
```
bdorfman-stripe: Can just turn this into an elif to avoid the unreachable code warning?

Stripe/STPTelemetryClient.m
```diff
@@ -0,0 +1,188 @@
+//
+//  STPTelemetryClient.m
+//  Stripe
+//
+//  Created by Ben Guo on 4/18/17.
+//  Copyright © 2016 Stripe, Inc. All rights reserved.
+//
+
+@import WebKit;
```
bdorfman-stripe: `@import` and not `#import <>` ?

Stripe/STPTelemetryClient.m
```diff
@@ -0,0 +1,188 @@
...
+    });
+    return sharedClient;
+}
+
+- (instancetype)init {
+    self = [super init];
+    if (self) {
+        NSURLSessionConfiguration *config = [NSURLSessionConfiguration defaultSessionConfiguration];
+        _urlSession = [NSURLSession sessionWithConfiguration:config];
+        [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(applicationDidBecomeActive) name:UIApplicationDidBecomeActiveNotification object:nil];
```
bdorfman-stripe: You probably want willEnterForeground and not didBecomeActive

Stripe/STPTelemetryClient.m
```diff
@@ -0,0 +1,188 @@
...
+- (void)applicationDidBecomeActive {
+    self.appOpenTime = [NSDate date];
+}
+
+- (NSString *)muid {
+    NSString *muid = [[[UIDevice currentDevice] identifierForVendor] UUIDString];
+    return muid ?: @"";
+}
+
+- (NSNumber *)timeOnPage {
```
bdorfman-stripe: Is it correct to track appOpenTime here and not time since viewing a Stripe view?

Stripe/STPTelemetryClient.m
```diff
@@ -0,0 +1,188 @@
...
+    if (self) {
+        NSURLSessionConfiguration *config = [NSURLSessionConfiguration defaultSessionConfiguration];
+        _urlSession = [NSURLSession sessionWithConfiguration:config];
+        [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(applicationDidBecomeActive) name:UIApplicationDidBecomeActiveNotification object:nil];
+        WKWebView *webView = [WKWebView new];
+        [webView loadHTMLString:@"<html></html>" baseURL:nil];
+        [webView evaluateJavaScript:@"navigator.userAgent" completionHandler:^(id result, __unused NSError *error) {
+            self.userAgent = result;
+        }];
+        _webView = webView;
```
bdorfman-stripe: Not sure you need to retain this here?

Also I think we should consider moving the webview instantiation to a background thread for performance reasons.

Stripe/STPAnalyticsClient.m
bg-stripe: ah yep, good call

Stripe/STPTelemetryClient.m
bg-stripe: ah whoops, will change

Stripe/STPTelemetryClient.m
bg-stripe: I chose `didBecomeActive` because `willEnterForeground` doesn't get called on first launch. I know `didBecomeActive` can get called more often, but I think its behavior is ok for our purposes here.

Stripe/STPTelemetryClient.m
bg-stripe: Yeah – we want to have a time_on_page value for users who don't use our UI components too. On the web, this is "number of milliseconds between Stripe.js loading and the user submitting card information" – I think app open time should provide a pretty similar signal.

Stripe/STPTelemetryClient.m
bdorfman-stripe: Ah ok, makes sense.

Stripe/STPTelemetryClient.m
bdorfman-stripe: 👍

Stripe/STPTelemetryClient.m
```diff
@@ -111,8 +100,9 @@ - (NSString *)language {
 - (NSString *)platform {
     struct utsname systemInfo;
     uname(&systemInfo);
-    NSString *deviceType = @(systemInfo.machine);
-    return deviceType ?: @"";
+    NSString *deviceType = @(systemInfo.machine) ?: @"";
```
bdorfman-stripe: This seems like a strange type coercion? here? `@()` returns NSNumber but assigning to NSString?

bg-stripe:
@bdorfman-stripe re-r? updated with some additional ios-specific fields

Stripe/STPTelemetryClient.m
bg-stripe: I was surprised by this too – boxing can also used for C strings
https://clang.llvm.org/docs/ObjectiveCLiterals.html#boxed-c-strings

bdorfman-stripe: Oh woah! Mind = blown

bdorfman-stripe:
LGTM

bg-stripe:
re-r? @bdorfman-stripe 
updated with a new payload format, and removed some ios fields.

Stripe/STPAPIClient.m
```diff
@@ -336,8 +334,9 @@ @implementation STPAPIClient (CreditCards)
 
 - (void)createTokenWithCard:(STPCard *)card completion:(STPTokenCompletionBlock)completion {
     NSMutableDictionary *params = [[STPFormEncoder dictionaryForObject:card] mutableCopy];
-    params[@"muid"] = [STPAnalyticsClient muid];
+    [[STPTelemetryClient sharedInstance] addTelemetryFieldsToParams:params];
     [self createTokenWithParameters:params completion:completion];
```
bdorfman-stripe: Is there a reason we are not doing this inside createTokenWithParameters? Do we not want telemetry on non-card tokenizations?

bg-stripe: ah yeah, this is intentional – we're starting with only card tokens, but we'll be passing the same stuff to source creations in the future.

PR #656, opened 2017-04-24
[sources-ui] UI updates

bg-stripe:
r? @bdorfman-stripe 

I haven't added new snapshot tests yet – planning on doing that as part of localization. I've re-generated existing images to get tests to pass.


Stripe/STPIDEALBankSelectorDataSource.m
```diff
@@ -30,11 +30,11 @@ - (instancetype)init {
     self = [super init];
     if (self) {
         _bankNameToBankCode = @{
-                                @"ABN AMRO": @"abn_amro",
```
bdorfman-stripe: Do these need to all be localized or no?

Stripe/STPPaymentMethodType.m
```diff
@@ -65,15 +65,15 @@ @interface STPPaymentMethodTypeSEPADebit : STPPaymentMethodType
...
 }
 
 - (UIImage *)paymentMethodTemplateImage {
-    return [STPImageLibrary addIcon];
+    return [STPImageLibrary sepaIcon];
 }
 
 - (NSString *)paymentMethodLabel {
-    return STPLocalizedString(@"New SEPA Debit", @"Label for 'new sepa debit' payment method field") ;;
+    return STPLocalizedString(@"Direct Debit Account", @"Label for SEPA Direct Debit payment method field") ;;
```
bdorfman-stripe: Do we need to alter the voiceover text for this to make it clear it is SEPA for blind users? (as it might not be clear if you can't see the associated icon)

Stripe/STPSectionHeaderView.m
```diff
@@ -24,22 +24,25 @@ + (instancetype)addressHeaderWithConfiguration:(STPPaymentConfiguration *)config
...
     BOOL needsAddress = !addressViewModel.isValid;
     switch (sourceType) {
         case STPSourceTypeCard:
+            headerView.title = STPLocalizedString(@"BILLING ADDRESS", @"Title for billing address entry section");
             needsAddress = needsAddress && config.requiredBillingAddressFields != STPBillingAddressFieldsNone;
             break;
+        case STPSourceTypeSEPADebit:
+            headerView.title = STPLocalizedString(@"ACCOUNT HOLDER ADDRESS", @"Title for address entry section");
+            break;
         default:
```
bdorfman-stripe: Prob shouldn't use `default` here so that we'll get compiler warnings when more types are added that may be relevant.

Stripe/STPPaymentMethodTableViewCell.m
```diff
@@ -76,10 +86,10 @@ - (NSAttributedString *)buildAttributedStringForPaymentMethod:(id<STPPaymentMeth
...
 }
 
-- (NSAttributedString *)buildAttributedStringForCardBrand:(NSString *)brandString
-                                                    last4:(NSString *)last4
-                                                 selected:(BOOL)selected {
-    NSString *template = STPLocalizedString(@"%@ Ending In %@", @"{card brand} ending in {last4}");
+- (NSAttributedString *)buildAttributedStringWithName:(NSString *)brandString
+                                                last4:(NSString *)last4
+                                             selected:(BOOL)selected {
+    NSString *template = STPLocalizedString(@"%@ Ending In %@", @"{account name} ending in {last4}");
```
bdorfman-stripe: We should update the logic in STPSource/STPCard's STPPaymentMethod `paymentMethodLabel` method to return the same thing here probably.

Stripe/STPSectionHeaderView.m
bg-stripe: ah yep, good point

Stripe/STPPaymentMethodType.m
bg-stripe: ah yep, I can add a `paymentMethodAccessibilityLabel` property that maps to `paymentMethodLabel` by default

Stripe/STPIDEALBankSelectorDataSource.m
bg-stripe: I think we can leave brand names out of the localization pipeline. I think we should also not localize payment method names that are brand names like `SOFORT` and `Giropay` – I'll update.

Stripe/STPPaymentMethodTableViewCell.m
bg-stripe: ah yep, good catch

bg-stripe:
@bdorfman-stripe updated in 6bcbb03

bdorfman-stripe:
LGTM

PR #660, opened 2017-05-02
Update STPPaymentCardTextField icons

bg-stripe:
r? @bdorfman-stripe 

This updates the icons used in PaymentCardTextField to match Elements on the web. Behavior has also slightly changed – the field now displays an error icon when the card number is invalid.


bdorfman-stripe:
Hmm, the actual icons inside the png are aligned differently now. Does this not throw off the alignment in the payment field?

bg-stripe:
Ah, yep – the brand icons are now a off center so that their position matches the cvc / error icons, which have a bubble in the top right. I've tweaked the layout in bb2564d so that they're vertically centered.

screenshot 2017-05-02 14 55 28screenshot 2017-05-02 14 55 15

bdorfman-stripe:
LGTM

bdorfman-stripe:
Maybe actually leave a note in MIGRATING or CHANGELOG that the images are differently aligned?

bg-stripe:
👍 added a note in ea44cb6

ketsalot1:
Like the way I started this way back as the way it should be Free, 
Now I am sure that the whole thing is actually turning into something that might really do a task of free transactions 🤘💩🤘🤔

PR #661, opened 2017-05-03
Update analytics

bdorfman-stripe:
Bucket product usage client side
Add app name and version to analytics
Move PayTextField's commonInit to private header

### DIFF
--- a/Stripe/STPAnalyticsClient.m
+++ b/Stripe/STPAnalyticsClient.m
@@ -8,6 +8,7 @@
 
 #import "STPAnalyticsClient.h"
 
+#import "NSBundle+Stripe_AppName.h"
 #import "NSMutableURLRequest+Stripe.h"
 #import "STPAPIClient+ApplePay.h"
 #import "STPAPIClient.h"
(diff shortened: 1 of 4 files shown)


Stripe/STPAnalyticsClient.m
```diff
@@ -140,16 +149,38 @@ - (NSArray *)productUsage {
...
+}
+
 - (void)logTokenCreationAttemptWithConfiguration:(STPPaymentConfiguration *)configuration
                                        tokenType:(NSString *)tokenType {
     NSDictionary *configurationDictionary = [self.class serializeConfiguration:configuration];
     NSMutableDictionary *payload = [self.class commonPayload];
     [payload addEntriesFromDictionary:@{
                                         @"event": @"stripeios.token_creation",
                                         @"token_type": tokenType ?: @"unknown",
-                                        @"apple_pay_enabled": @([Stripe deviceSupportsApplePay]),
```
teich-stripe: why did we remove apple pay here?

bdorfman-stripe: I moved it to the common payload below, so it will be attached on all analytics, not just token creations. Seems to make more sense since that has all the device/app specific stuff.

PR #663, opened 2017-05-04
Update analytics for sources ui

bdorfman-stripe:
Rewrite hooking code to be less duplicative
Add add source vc hook
Remove code that was added back in a bad merge from master


Stripe/STPAnalyticsClient.m
```diff
@@ -50,52 +51,39 @@ + (void)initialize {
     [self initializeIfNeeded];
 }
 
++ (void)addHookForClass:(Class)hookedClass selector:(SEL)selector {
```
bg-stripe: nice, good call

PR #665, opened 2017-05-04
Update payconfig analytics serialization

bdorfman-stripe:


Stripe/STPAnalyticsClient.m
```diff
@@ -276,6 +277,21 @@ + (NSDictionary *)serializeConfiguration:(STPPaymentConfiguration *)configuratio
     }
     dictionary[@"company_name"] = configuration.companyName ?: @"unknown";
     dictionary[@"apple_merchant_identifier"] = configuration.appleMerchantIdentifier ?: @"unknown";
+    dictionary[@"use_card_sources"] = configuration.useSourcesForCards ? @"1" : @"0";
```
bg-stripe: should we use a boxed bool instead, to be consistent with `apple_pay_enabled`?

bdorfman-stripe: Ah, good call.

PR #671, opened 2017-05-10
[sources-ui] remove address fields from SEPA form

bg-stripe:
r? @bdorfman-stripe 

Since address params are now optional, we can remove the address fields from the SEPA form.


bg-stripe:
Synced with mitu & maddie and it sounds like this is indeed what we should do (remove address option entirely, without a configuration option).

bdorfman-stripe:
Cool. LGTM!

PR #673, opened 2017-05-12
Enable card scanning using card.io

bg-stripe:
r? @bdorfman-stripe 

Still need to get translations in and update snapshot tests before merging, but ready for review.


Example/Stripe iOS Example (Simple).xcworkspace/contents.xcworkspacedata
```diff
@@ -1,10 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
bg-stripe: this workspace in the Example directory was broken, and not useful, so I removed it.

Stripe/STPAnalyticsClient.m
```diff
@@ -192,7 +193,6 @@ - (void)logSourceCreationAttemptWithConfiguration:(STPPaymentConfiguration *)con
     [payload addEntriesFromDictionary:@{
                                         @"event": @"stripeios.source_creation",
                                         @"source_type": sourceType ?: @"unknown",
-                                        @"apple_pay_enabled": @([Stripe deviceSupportsApplePay]),
```
bg-stripe: deleted this field since it's dup'd in `commonPayload`

PR #676, opened 2017-05-16
[sources-ui] Dynamic 3DS

bdorfman-stripe:
Adds precheck bindings to API Client
Adds dynamic 3ds as an option for payment context
Adds/refactors 3ds paycontext tests


Stripe/STPPaymentContext.m
```diff
@@ -568,7 +544,116 @@ - (void)requestSynchronousSourcePayment:(id<STPSourceProtocol>)source {
...
+        return;
+    }
+
+    self.state = STPPaymentContextStateRequestingPayment;
+
+    STPSourcePrecheckParams *precheckParams = [STPSourcePrecheckParams new];
+    precheckParams.sourceID = source.stripeID;
+    precheckParams.paymentAmount = @(self.paymentAmount);
+    precheckParams.paymentCurrency = self.paymentCurrency;
+    // TODO: method for merchant to add metadata/additional fields
```
bg-stripe: We can get metadata from `self.sourceInformation.metadata`. Maybe we can add an `additionalParameters` property to `STPAdditionalSourceInfo`?

Stripe/STPSourceParams.m
```diff
@@ -13,6 +13,17 @@
 #import "STPFormEncoder.h"
 #import "STPSource+Private.h"
 
+static NSString *const STPSourceParamsKeyOwnerName = @"name";
```
bg-stripe: 👍 nice

Tests/Tests/STPPaymentContext3DSTest.m
```diff
@@ -90,6 +99,25 @@ - (BOOL)threeDSParams:(STPSourceParams *)sourceParams matchCardSource:(STPSource
...
+
+    id<STPSourceProtocol> source = nil;
+    if ([paymentContext.selectedPaymentMethod conformsToProtocol:@protocol(STPSourceProtocol)]) {
+        source = (id<STPSourceProtocol>)paymentContext.selectedPaymentMethod;
+    }
+
+    STPSourcePrecheckParams *expectedParams = [STPSourcePrecheckParams new];
+    expectedParams.sourceID = source.stripeID;
+    expectedParams.paymentAmount = @(paymentContext.paymentAmount);
+    expectedParams.paymentCurrency = paymentContext.paymentCurrency;
```
bg-stripe: We should also check that `metadata` matches `paymentContext.sourceInformation.metadata`.
Also should verify additional parameters once we add a way to set those.

Tests/Tests/STPPaymentContext3DSTest.m
```diff
@@ -100,31 +128,187 @@ - (NSError *)paymentMethodNotAvailableError {
...
+- (void)test3DSDynamicOptionalOnCardRequiredOnPrecheckAndStatusPending {
+    self.config.threeDSecureSupportType = STPThreeDSecureSupportTypeDynamic;
+    STPSource *cardSource = [self cardSourceWith3DSStatus:@"optional"];
+    self.sut.selectedPaymentMethod = cardSource;
+    OCMReject([self.mockDelegate paymentContext:[OCMArg any] didCreatePaymentResult:[OCMArg any] completion:[OCMArg any]]);
+    OCMReject([self.mockDelegate paymentContext:[OCMArg any] didFinishWithStatus:STPPaymentStatusSuccess error:[OCMArg any]]);
+
+    [self stubAndVerifyPrecheckCalledAndReturnResult:[self precheckResultWithRequiredActions:@[STPSourcePrecheckRequiredActionCreateThreeDSecureSource]]];
+    [self stubAndVerifyCreateSourceCalledAndReturn3DSSource:[self threeDSSourceWithStatus:@"pending"]];
+    [self stubAndVerifyHostViewControllerPresentsViewControllerOfClass:[SFSafariViewController class]];
```
bg-stripe: nice work, these helpers make these tests so much easier to parse ✨

Stripe/STPAPIClient.m
```diff
@@ -403,6 +406,47 @@ - (void)createSourceWithParams:(STPSourceParams *)sourceParams completion:(STPSo
...
+}
+
+- (void)precheckAndCreate3DSSourceIfNeededWithSourceParams:(STPSourceParams *)sourceParams
+                                                completion:(STP3DSSourcePrecheckCompletionBlock)completion {
+    NSCAssert(sourceParams != nil, @"'params' is required to create a source");
+    NSCAssert(completion != nil, @"'completion' is required to use the source that is created");
+    NSCAssert(sourceParams.type == STPSourceTypeThreeDSecure, @"Must use 3DS params");
+    NSCAssert(sourceParams.threeDSecureParamsCardId != nil, @"Must have a card id");
+
+    STPSourcePrecheckParams *precheckParams = [STPSourcePrecheckParams new];
```
bdorfman-stripe: Yeah, you can set additional parameters if you do the two requests separately, but the convenience method doesnt do it. I'm not sure the best way to resolve this, maybe actually just delete this convenience method.

Stripe/STPAPIClient.m
```diff
@@ -403,6 +406,47 @@ - (void)createSourceWithParams:(STPSourceParams *)sourceParams completion:(STPSo
...
+    precheckParams.sourceID = sourceParams.threeDSecureParamsCardId;
+    precheckParams.paymentAmount = sourceParams.amount;
+    precheckParams.paymentCurrency = sourceParams.currency;
```
bdorfman-stripe: No because the metadata isn't necessarily the same for charging vs source creation.

Stripe/STPPaymentContext.m
bdorfman-stripe: Meta data for precheck is not the same as source creation metadata (one is metadata you would attach to charge, the other is for attaching to source). Adding new properties for ithem onto additional source info could work maybe.

Tests/Tests/STPPaymentContext3DSTest.m
bdorfman-stripe: Same comments on metadata as before, but yep will add checks when we add a method to do so.

Stripe/STPAPIClient.m
bg-stripe: Removing the method seems reasonable to me, it's not providing _that_ much convenience

Stripe/STPAPIClient.m
bg-stripe: ah, good point

Stripe/STPAnalyticsClient.m
```diff
@@ -287,6 +287,8 @@ + (NSDictionary *)serializeConfiguration:(STPPaymentConfiguration *)configuratio
         case STPThreeDSecureSupportTypeStatic:
             threeDSecureType = @"static";
             break;
+        case STPThreeDSecureSupportTypeDynamic:
+            threeDSecureType = @"dynamic";
```
billinghamj: Missing break?

billinghamj:
I assume this is pre-release/private beta? Doesn't seem to be in the API docs atm

bdorfman-stripe:
@billinghamj Yes, you should only use released numbered versions of the API (https://github.com/stripe/stripe-ios/releases) . Other commits/branches may have work that is still under development.

bdorfman-stripe:
Ready for re-review @bg-stripe

PR #678, opened 2017-05-17
Fix payment method cell flicker

bg-stripe:
r? @bdorfman-stripe 

The selected payment method cell currently flickers with its highlight color. It's subtle when using the example app's default theming, which I think is why we didn't notice it earlier.

Not using `UITableViewCell`'s `selected` behavior fixes the issue, and seems like a reasonable workaround.

Before:
https://www.dropbox.com/s/18pvcvhtdw1rpur/flicker-before.mov?dl=0

After:
https://www.dropbox.com/s/r6m1mr1u9f7y708/flicker-after.mov?dl=0

### DIFF
--- a/Stripe/STPPaymentMethodTableViewCell.h
+++ b/Stripe/STPPaymentMethodTableViewCell.h
@@ -12,7 +12,7 @@
 
 @interface STPPaymentMethodTableViewCell : UITableViewCell
 
-- (void)configureWithPaymentMethod:(id<STPPaymentMethod>)paymentMethod theme:(STPTheme *)theme;
+- (void)configureWithPaymentMethod:(id<STPPaymentMethod>)paymentMethod selected:(BOOL)selected theme:(STPTheme *)theme;
 - (void)configureForNewCardRowWithTheme:(STPTheme *)theme;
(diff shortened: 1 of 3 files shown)


bdorfman-stripe:
I think this looks good. I'm a little worried there are other places UIkit calls the built in `setSelected` that we are now not capturing though?

bg-stripe:
I think it's ok – we don't use the `selected` property on this cell type elsewhere, and I'm pretty sure it wouldn't get set on its own. On row selection we reload the section, so it goes through `configureWithPaymentMethod`.

bdorfman-stripe:
Hmm ok. I thought UIKit sets the highlighted and selected bits on rows automatically when you tap on them, but I think this is fine to do manually like this.

bg-stripe:
The failing carthage test passes locally. Not sure how to work around this form of flakiness on travis. I suspect it might be caused by me squashing commits, causing a cached commit hash on Travis to become invalid? Re-running the test doesn't fix the issue.

bdorfman-stripe:
Yeah it happens all the time. I think its fine to just merge.

PR #679, opened 2017-05-17
use Alamofire in example app

bg-stripe:
r? @bdorfman-stripe 

Most of this diff is noise from adding Alamofire to the repo.
Probably best to just look at/comment on this commit: 7b8e078b


bg-stripe:
actually, hol

PR #680, opened 2017-05-18
use Alamofire in example app

bg-stripe:
r? @bdorfman-stripe 

- I've added Alamofire to the swift example app, replacing our custom networking code. 
- The example app no longer works out of the box – users will have to first run `./setup.rb`. I've added a note to the README. 
- I've made `STPCustomer` implement `STPAPIResponseDecodable`, since `STPCustomerDeserializer` doesn't make much sense unless you're rolling your own networking. (I'm deprecating `STPCustomerDeserializer` on my ephemeral keys branch).


bdorfman-stripe:
Looks pretty good. It would be nice if there was a way to have Xcode check for the existence of Alamofire and do the build instead of a separate script though.

bg-stripe:
@bdorfman-stripe updated as discussed – I've added a build phase that reminds you to run ./setup.sh if dependencies aren't installed. Also turned the setup script into a bash script.

README.md
```diff
@@ -26,7 +26,7 @@ There are 2 example apps included in the repository:
 - Stripe iOS Example (Simple) shows an integration using our prebuilt UI components [Swift]
 - Stripe iOS Example (Custom) shows how to use our low-level methods to accept payments using several different payment methods [Objective-C].
 
-To build and run the example apps, open `Stripe.xcworkspace` and choose the appropriate scheme.
+To build the example apps, you'll need to first run `./setup.rb`. Then, open `Stripe.xcworkspace` and choose the appropriate scheme.
```
bdorfman-stripe: Not a `.rb` any more

bdorfman-stripe:
Minor error in readme, but otherwise good to merge 👍

README.md
bg-stripe: ah good catch

PR #681, opened 2017-05-18
Add support for ephemeral keys

bg-stripe:
r? @bdorfman-stripe @joeydong-stripe


Stripe/NSError+STPCustomerContext.m
```diff
@@ -0,0 +1,22 @@
+//
```
jflinter: I'd prefer to move this (and all other stripe-specific errors) into StripeError so that we can see them all in one place.

bdorfman-stripe: The public facing error codes are in StripeError. This is just an internal-only helper to generate an NSError object (similar to the PaymentContext NSError category).

jflinter: What about adding a private header (StripeError+Private) for all those extra errors, then?

bg-stripe: 👍  seems reasonable to consolidate in one file – I'll add a private header.

Example/Stripe iOS Example (Simple)/CheckoutViewController.swift
```diff
@@ -13,13 +13,13 @@ class CheckoutViewController: UIViewController, STPPaymentContextDelegate {
 
     // 1) To get started with this demo, first head to https://dashboard.stripe.com/account/apikeys
     // and copy your "Test Publishable Key" (it looks like pk_test_abcdef) into the line below.
-    let stripePublishableKey = ""
+    let stripePublishableKey = "pk_test_4TDXAGLdZFGNbXYGajQlcstU"
```
bdorfman-stripe: Did you mean to commit this?

Example/Stripe iOS Example (Simple)/MyAPIClient.swift
```diff
@@ -38,62 +38,21 @@ class MyAPIClient: NSObject, STPBackendAPIAdapter {
...
-        }
-    }
-    
-    @objc func attachSource(toCustomer source: STPSourceProtocol, completion: @escaping STPErrorBlock) {
-        let url = self.baseURL.appendingPathComponent("customer/sources")
-        Alamofire.request(url, method: .post, parameters: [
-            "source": source.stripeID,
+
+    func createCustomerKey(withAPIVersion apiVersion: String, completion: @escaping STPJSONResponseCompletionBlock) {
+        let url = URL(string: "https://bg-stripe.runkit.io/591e05608ca4980012451c55/branches/master/ephemeral_keys")!
```
bdorfman-stripe: Again, probably not committing your specific url here? How is this going to work for actually shipping?

Stripe/PublicHeaders/STPAPIResponseDecodable.h
```diff
@@ -11,6 +11,11 @@
 @protocol STPAPIResponseDecodable <NSObject>
 
 /**
+ A string representing the object's type. May be nil.
+ */
++ (nullable NSString *)object;
```
bdorfman-stripe: I'm wary of adding protocol methods that are super generically named like this, for fear of conflict with other existing methods. Maybe something that makes it more stripe specific or specific to this protocol?

I'm also not sure the value of having this exposed anyway. It's helpful on web where everything is just json dicts, but on iOS if there's a json object <-> iOS class mismatch, it should just return nil from STPAPIResponseDecodeable. Not sure if we want competing instance.object and instance.class checking?

Stripe/PublicHeaders/STPCustomer.h
```diff
@@ -14,6 +14,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 /**
  *  An `STPCustomer` represents a deserialized Customer object from the Stripe API. You can use `STPCustomerDeserializer` to convert a JSON response from the Stripe API into an `STPCustomer`.
+ *  TODO: deprecate
```
bdorfman-stripe: Why deprecate STPCustomer? Will this not be accessible via CustomerContext?

Stripe/PublicHeaders/STPCustomerContext.h
```diff
@@ -0,0 +1,67 @@
...
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated"
+@interface STPCustomerContext : NSObject <STPBackendAPIAdapter>
+#pragma clang diagnostic pop
+
+/**
+ This is a convenience singleton CustomerContext. Before using the singleton
+ instance, you'll need to give it a key provider using `setKeyProvider`.
+ */
++ (instancetype)sharedInstance;
```
bdorfman-stripe: What is the reasoning behind providing this singleton  given they have to set the key provider anyway, instead of just making them use the standard init?

Stripe/STPCustomerContext.m
```diff
@@ -0,0 +1,151 @@
...
+                self.customer = customer;
+            }
+            if (completion) {
+                completion(customer, error);
+            }
+        }];
+    }];
+}
+
+- (void)attachSourceToCustomer:(id<STPSourceProtocol>)source completion:(STPErrorBlock)completion {
```
bdorfman-stripe: This changes the sources array on customer, but doesn't update the cached customer object or clear out the cache setting so your next fetch of the customer may be out of sync. Should force a re-fetch after this and/or add some way to force update customers ignoring the cache.

Stripe/PublicHeaders/STPCustomerContext.h
bg-stripe: I think it's useful to provide a singleton for the case where you're also presenting `PaymentMethodsVC` on its own in a settings page. In that case, you could e.g. initialize your singleton `CustomerContext` somewhere, and use the same instance for your checkout flow and your settings screen.

Stripe/STPCustomerContext.m
bg-stripe: ah, good point – I'll make this clear out the cached customer.

Stripe/PublicHeaders/STPCustomer.h
bg-stripe: Ah, this TODO is wrong – we should just deprecate the `customerWithStripeID:defaultSource:sources:` constructor. I'll update.

Example/Stripe iOS Example (Simple)/CheckoutViewController.swift
bg-stripe: ah yeah, just left these here while developing – i'll remove them before the demo

Example/Stripe iOS Example (Simple)/MyAPIClient.swift
bg-stripe: yep, this is temporary – we'll update bindings and the example backend before actually shipping

Stripe/PublicHeaders/STPAPIResponseDecodable.h
bg-stripe: Good call on renaming this to something more specific – maybe `stripeObject`?

I've added this because we now need to be able to parse a heterogenous response (adding a source to a customer can return either a token or source). The `parseResponse` method in `STPAPIRequest.m` now takes an array of serializers [0], and uses `object` to decide which one to use.

[0] This is actually misnamed, these are technically deserializers – I should rename the methods

bdorfman-stripe: Do we have to make it a public method? We already do heterogenous parsing like this in the current STPCustomerDeserializer implementation. I would rather keep it something internal if it's only used in this one spot and not really needed by our users.

bg-stripe: That's fair – I'll add a private superprotocol that adds an optional `stripeObject` method.

Stripe/STPCustomer.m
```diff
@@ -39,13 +39,20 @@ + (NSArray *)requiredFields {
...
     return deserializer.customer;
 }
+#pragma clang diagnostic pop
 
 @end
 
+/**
+ NOTE: STPCustomerDeserializer has been deprecated. When we remove 
+ STPBackendAPIAdapter, we should also remove STPCustomerDeserializer.
+ */
```
bdorfman-stripe: Should change this comment before shipping?

Stripe/STPAPIRequest.m
```diff
@@ -65,20 +79,34 @@ + (NSURLSessionDataTask *)getWithAPIClient:(STPAPIClient *)apiClient
...
         httpResponse = (NSHTTPURLResponse *)response;
     }
+
+    if (serializers.count == 0) {
+        returnedError = [NSError stp_genericFailedToParseResponseError];
+    } else {
+        NSDictionary *jsonDictionary = body ? [NSJSONSerialization JSONObjectWithData:body options:(NSJSONReadingOptions)kNilOptions error:NULL] : nil;
+        NSString *object = jsonDictionary[@"object"];
+        Class serializerClass = [serializers.firstObject class];
+        for (id<STPAPIResponseDecodable> serializer in serializers) {
```
bdorfman-stripe: This logic is a little weird. If none of the stripeObject keys match, it will end up with the firstObject's class. Is that intentional?

Stripe/STPAPIClient.m
```diff
@@ -449,3 +463,55 @@ - (void)stopPollingSourceWithId:(NSString *)identifier {
...
 
 @end
+
+#pragma mark - Customers
+
+@implementation STPAPIClient (Customers)
+
++ (STPAPIClient *)apiClientWithEphemeralKey:(STPEphemeralKey *)key {
+    STPAPIClient *client = [[self alloc] init];
+    client.apiKey = key.secret;
```
bdorfman-stripe: Should we be copying over other instance vars from ourself to this hidden sub-client (eg `client.configuration = self.configuration`)

Stripe/STPCustomerContext.m
```diff
@@ -0,0 +1,154 @@
...
+    if (!self.keyManager) {
+        completion([NSError stp_customerContextMissingKeyProviderError]);
+        return;
+    }
+    [self.keyManager getCustomerKey:^(STPEphemeralKey *ephemeralKey, NSError *retrieveKeyError) {
+        if (retrieveKeyError) {
+            if (completion) {
+                completion(retrieveKeyError);
```
bdorfman-stripe: Do these completions need to be dispatched back to main? (or to the original thread?)

Stripe/STPCustomer.m
bg-stripe: ah, yep – I'll remove.

Stripe/STPAPIRequest.m
bg-stripe: Ah, yeah, this is kind of weird – it's meant to support the case where there's one serializer that doesn't implement `stripeObject`. I'll update the logic.

Stripe/STPAPIClient.m
bg-stripe: Ah yeah, I can add a configuration parameter here – it could be useful in the future if we add telemetry to these requests.

Stripe/STPCustomerContext.m
bg-stripe: Ah yep, I think right now PaymentContext pretty consistently dispatches to main from `apiAdapter` calls, but it seems reasonable to make CustomerContext always dispatch to main.

Stripe/STPAPIClient.m
bg-stripe: Ah whoops, just realized these are class methods, so we wouldn't be able to copy over instance vars. `CustomerContext` doesn't reference a configuration either, so we wouldn't be able to pass one through. I think it's fine to not set one for now.

bdorfman-stripe: Hmm, should we make them not class methods? None of the other STPAPIClient methods are. This seems potentially limiting if we want to add config options for customer operations in the future.

bg-stripe: I decided to make these ephemeral key auth'd requests class methods because I think it's safer, and simpler to explain in the future when we make these methods public. Since each request requires an ephemeral key, and creates a one-off API client, we don't have to worry about guarding against a nil key, or what happens if you change the client's ephemeral key while a request is in progress.

I think we can still add config options in the future pretty easily, by adding a config parameter to each request.

bdorfman-stripe: Ok, sounds good

bg-stripe:
re-r? @bdorfman-stripe

Stripe/STPEphemeralKeyManager.h
```diff
@@ -0,0 +1,46 @@
...
+#import "STPEphemeralKeyProvider.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+typedef void (^STPEphemeralKeyCompletionBlock)(STPEphemeralKey * __nullable ephemeralKey, NSError * __nullable error);
+
+@interface STPEphemeralKeyManager : NSObject
+
+/**
+ If the current ephemeral key expires in less than this time interval, a call
```
bdorfman-stripe: Should we take the time to standardize on a comment format or does it not matter? (some of our stuff has preceding `*` on every line and some don't.

Happy to punt on this to a future PR or release just wanted to mention it as an inconsistency.

Stripe/PublicHeaders/STPEphemeralKeyProvider.h
```diff
@@ -0,0 +1,46 @@
...
+
+ @param apiVersion  The Stripe API version to use when creating a key.
+ You should pass this parameter to your backend, and use it to set the API version
+ in your key creation request. Passing this version parameter ensures that the
+ Stripe SDK can always parse the ephemeral key response from your server.
+ @param completion  Call this callback when you're done fetching a new ephemeral
+ key from your backend. For example, `completion(json, nil)` (if your call succeeds)
+ or `completion(nil, error)` if an error is returned.
+ */
+- (void)createCustomerKeyWithAPIVersion:(NSString *)apiVersion completion:(STPJSONResponseCompletionBlock)completion;
```
bdorfman-stripe: Are there no other parameters that might be necessary here? I'm thinking possibly the merchant account on which the customer is on (for Connect). Also does the api not have any other configuration options? (eg time interval for when the key expires?)

Stripe/StripeError.m
```diff
@@ -148,3 +148,15 @@ + (nonnull NSString *)stp_cardErrorProcessingErrorUserMessage {
 }
 
 @end
+
+@implementation NSError (STPCustomerContext)
+
++ (NSError *)stp_customerContextMissingKeyProviderError {
```
bdorfman-stripe: Should this be in a +Private.m instead of in the main implementation file?

Stripe/StripeError+Private.h
```diff
@@ -0,0 +1,15 @@
...
+//  StripeError+Private.h
+//  Stripe
+//
+//  Created by Ben Guo on 5/22/17.
+//  Copyright © 2017 Stripe, Inc. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+@interface NSError (STPCustomerContext)
```
bdorfman-stripe: Having `+Private.h` but the category actually being `(STPCustomerContext)` is confusing. Should just make a catch-all private (and maybe move any other related internal only error methods here) or rename the file.

Stripe/STPAPIClient.m
```diff
@@ -455,3 +469,55 @@ - (void)stopPollingSourceWithId:(NSString *)identifier {
...
++ (void)addSource:(NSString *)sourceID
+toCustomerUsingKey:(STPEphemeralKey *)ephemeralKey
+       completion:(STPSourceProtocolCompletionBlock)completion {
+    STPAPIClient *client = [self apiClientWithEphemeralKey:ephemeralKey];
+    NSString *endpoint = [NSString stringWithFormat:@"%@/%@/%@", customersEndpoint, ephemeralKey.customerID, sourcesEndpoint];
+    [STPAPIRequest<STPSourceProtocol> postWithAPIClient:client
+                                               endpoint:endpoint
+                                             parameters:@{@"source": sourceID}
+                                            serializers:@[[STPCard new], [STPSource new]]
+                                             completion:^(STPCard *object, __unused NSHTTPURLResponse *response, NSError *error) {
```
bdorfman-stripe: This completion is not typed correctly (object could be card or source).

bg-stripe: ah yep, good catch!

Stripe/STPEphemeralKeyManager.h
bg-stripe: Yeah, we should standardize and make sure all our comments are wrapped – let's punt for a future PR though

CHANGELOG
```diff
@@ -1,3 +1,8 @@
+== 11.0.0 2017-06-22
+* We've greatly simplified the integration for `STPPaymentContext`. See MIGRATING.md
+* We've made the `shippingAddress` property on `STPPaymentContext` readwrite, and you can now use it to programmatically pre-fill your user's shipping information.
+* `STPPaymentContext` will now save your user's shipping information to their Stripe customer object. Shipping information will automatically be pre-filled from the customer object for subsequent checkouts.
+
```
bg-stripe: 👍 will do

Stripe/StripeError+Private.h
bg-stripe: Ah yeah, I can rename the category to `(StripePrivate)`. There aren't really any other internal-only error methods to move over right now – there are a couple in `STPCheckoutAPIClient`, but it doesn't seem worth moving those if we're going to be removing the class.

Stripe/StripeError.m
bg-stripe: 👍 , I'll move it. Realizing we should probably also be doing the `linkCategoryForStaticLib` thing for this and `StripeError.m` – I'll add.

Stripe/PublicHeaders/STPEphemeralKeyProvider.h
bg-stripe: I think API version is the only parameter that the SDK needs to specify – the SDK can pass the Stripe API version it's using, but wouldn't know about a connected account ID. 

I think if you wanted PaymentContext to modify a customer on a connected account, you'd just need to use the connected account's publishable key in your PaymentConfiguration, and use its secret key to create an ephemeral key on your backend.

Stripe/PublicHeaders/STPCustomerContext.h
joeydong-stripe: I do find it a little weird that STPCustomerContext has a built in singleton feature but STPPaymentContext doesn't and also +1 to brian's point.

bdorfman-stripe: Yeah I think I am still against singleton. I'm fairly against singletons just in general, but also after more though this seems ripe for accidental logout issues (eg user logs out of the app, key is not cleared, non-logged in user or different user now has access to prev user's card data).

bg-stripe: Good points – I'll remove it.

bdorfman-stripe: I think we should also add some explicit documentation (or methods?) around clearing this data when appropriate on user logout.

Stripe/PublicHeaders/STPAPIResponseDecodable.h
joeydong-stripe: > [0] This is actually misnamed, these are technically deserializers – I should rename the methods

Haha I was thinking about this after our MLG yesterday

bg-stripe:
@bdorfman-stripe updated! I've removed the CustomerContext singleton, and updated the changelog/migrating to reflect this

Stripe/PublicHeaders/STPBackendAPIAdapter.h
```diff
@@ -37,7 +35,7 @@ typedef void (^STPCustomerCompletionBlock)(STPCustomer * __nullable customer, NS
  *  @see STPCard
  *  @param completion call this callback when you're done fetching and parsing the above information from your backend. For example, `completion(customer, nil)` (if your call succeeds) or `completion(nil, error)` if an error is returned.
  */
-- (void)retrieveCustomer:(STPCustomerCompletionBlock)completion;
+- (void)retrieveCustomer:(nullable STPCustomerCompletionBlock)completion;
```
joeydong-stripe: I wonder if there will be any unintended consequences of changing this API. If so, seems fairly straightforward for us to avoid it.

Stripe/PublicHeaders/STPCustomerContext.h
```diff
@@ -0,0 +1,48 @@
...
+#pragma clang diagnostic ignored "-Wdeprecated"
+@interface STPCustomerContext : NSObject <STPBackendAPIAdapter>
+#pragma clang diagnostic pop
+
+/**
+ Whenever the customer context retrieves a customer, it will return a cached
+ value if it was retrieved less than this number of seconds ago.
+ The default value is 60 seconds.
+ */
+@property (nonatomic, assign) NSTimeInterval cachedCustomerMaxAge;
```
joeydong-stripe: Was there a need for us to expose this `cachedCustomerMaxAge` feature? / In what situation would a developer need to change it? (+ document that)

joeydong-stripe: I guess in general how did we land on a 60 second expiration?

Stripe/STPAPIClient.m
```diff
@@ -75,6 +77,10 @@ @interface STPAPIClient()
 
 @implementation STPAPIClient
 
++ (NSString *)apiVersion {
+    return stripeAPIVersion;
```
joeydong-stripe: Are we safe to upgrade the version to something more recent?

Stripe/STPAPIClient.m
```diff
@@ -129,8 +131,20 @@ - (instancetype)initWithPublishableKey:(NSString *)publishableKey
...
 
+- (void)setApiKey:(NSString *)apiKey {
+    NSURLSessionConfiguration *sessionConfiguration = [NSURLSessionConfiguration defaultSessionConfiguration];
+    NSString *auth = [@"Bearer " stringByAppendingString:apiKey ?: @""];
+    sessionConfiguration.HTTPAdditionalHeaders = @{
+                                                   @"X-Stripe-User-Agent": [self.class stripeUserAgentDetails],
+                                                   @"Stripe-Version": stripeAPIVersion,
+                                                   @"Authorization": auth,
+                                                   };
+    self.urlSession = [NSURLSession sessionWithConfiguration:sessionConfiguration];
```
joeydong-stripe: Looks like we should work on dropping this `urlSession` property (later) to avoid this complexity.

Stripe/STPAPIRequest.h
```diff
@@ -20,10 +20,22 @@ typedef void(^STPAPIResponseBlock)(ResponseType object, NSHTTPURLResponse *respo
                                  serializer:(ResponseType)serializer
                                  completion:(STPAPIResponseBlock)completion;
 
++ (NSURLSessionDataTask *)postWithAPIClient:(STPAPIClient *)apiClient
+                                    endpoint:(NSString *)endpoint
+                                  parameters:(NSDictionary *)parameters
+                                 serializers:(NSArray<ResponseType>*)serializers
```
joeydong-stripe: should we go ahead and rename `serializers` to `deserializers` (or in a different PR)?

Stripe/PublicHeaders/STPCustomerContext.h
bg-stripe: 👍  I'll add some documentation that clears this up. I'm not sure about adding a method to clear a CustomerContext – it seems simpler / maybe less error prone to just advise that you create a new CustomerContext instance for each session, and warn against reusing it for different customers.

Stripe/STPAPIRequest.m
```diff
@@ -65,20 +79,42 @@ + (NSURLSessionDataTask *)getWithAPIClient:(STPAPIClient *)apiClient
...
+    if (serializers.count == 0) {
+        returnedError = [NSError stp_genericFailedToParseResponseError];
+    } else {
+        NSDictionary *jsonDictionary = body ? [NSJSONSerialization JSONObjectWithData:body options:(NSJSONReadingOptions)kNilOptions error:NULL] : nil;
+        NSString *object = jsonDictionary[@"object"];
+        Class serializerClass;
+        if (serializers.count == 1) {
+            serializerClass = [serializers.firstObject class];
+        } else {
+            for (id<STPAPIResponseDecodable> serializer in serializers) {
```
joeydong-stripe: I wonder if we can move `STPAPIResponseDecodable` to be a private protocol so that we can enforce the `stripeObject` method.

Stripe/PublicHeaders/STPCustomerContext.h
bdorfman-stripe: Yep if we're removing singleton I think that is the best advice 👍

Stripe/STPEphemeralKeyManager.m
```diff
@@ -0,0 +1,106 @@
...
+    return self.customerKey && self.customerKey.expires.timeIntervalSinceNow > self.expirationInterval;
+}
+
+// Eager key refreshes on app foreground are throttled to once per hour
+- (BOOL)shouldPerformEagerRefresh {
+    return !self.lastEagerKeyRefresh || self.lastEagerKeyRefresh.timeIntervalSinceNow > MinEagerRefreshInterval;
+}
+
+- (void)handleWillForegroundNotification {
+    if (!self.currentKeyIsUnexpired && self.shouldPerformEagerRefresh) {
```
joeydong-stripe: I was expecting this condition to be `!self.currentKeyIsUnexpired || self.shouldPerformEagerRefresh` but I'm guessing it's not that because we don't want to do foreground refreshes if no customer key was ever fetched in the current app lifecycle? If so, maybe just add a comment.

joeydong-stripe: aka don't refresh if we've never fetched one before

Stripe/STPAPIClient.m
bdorfman-stripe: Not sure what you mean by this?

Stripe/PublicHeaders/STPCustomerContext.h
bdorfman-stripe: Also: Would it be better/good to add a manual cache clearing method here?

Stripe/STPEphemeralKeyManager.m
```diff
@@ -0,0 +1,106 @@
...
+        self.lastEagerKeyRefresh = [NSDate date];
+        [self.keyProvider createCustomerKeyWithAPIVersion:self.apiVersion completion:^(NSDictionary *jsonResponse, __unused NSError *error) {
+            STPEphemeralKey *key = [STPEphemeralKey decodedObjectFromAPIResponse:jsonResponse];
```
bdorfman-stripe: Do we need some error handling here? Retry on errors, etc?

Stripe/STPAPIClient.m
joeydong-stripe: Nevermind, I thought `apiClient.urlSession` was only used for the file upload api but it's more complicated than just that.

Stripe/PublicHeaders/STPBackendAPIAdapter.h
bg-stripe: Hmm, I think it's unlikely for there to be consequences – I can't think of a reason for the developer to ever call this method themselves. I think in the next major version we should make all these `STPBackendAPIAdapter` methods private methods on `CustomerContext`.

Stripe/PublicHeaders/STPCustomerContext.h
bg-stripe: My thinking exposing it was that it makes it clearer that CustomerContext does caching, which might be surprising. After thinking about it some more, I think it's probably better to hide it until someone asks for the configuration option in the future. It's hard to come up with a good recommendation for how to come up with an appropriate value if you want to change it, since when CustomerContext retrieves a customer is opaque to the developer and out of their control.

bg-stripe: Also open to suggestions on the 60 seconds setting – I basically picked a low estimate for the time from app open -> purchase

Stripe/STPAPIClient.m
bg-stripe: (discussed in chat)

Stripe/STPEphemeralKeyManager.m
bg-stripe: Ah yeah, I can add a comment here. I think an `||` would be fine, but I feel a bit better being more conservative here – we don't want to end up in a situation where apps are hitting the `ephemeral_keys` endpoint on every foreground, so using an `&& self.shouldPerformEagerRefresh` is a failsafe against issues like the key always being nil due to some decoding issue.

Stripe/STPAPIRequest.m
bg-stripe: Interesting – we'd have to move `allResponseFields` to another public protocol if we did that, but I think that could be a reasonable move.

bdorfman-stripe: 👍 I think the only method in that protocol users should ever have to actually access is `allResponseFields` so I'm pro splitting some stuff up. Can do in a different PR maybe? Not sure if we want to make this change for 11.0 or not (maybe do it in the future version where we remove backend api adapter)

joeydong-stripe: 👍

Stripe/STPEphemeralKeyManager.m
bg-stripe: Hm, I think it's ok to not retry the eager refresh if it fails – it's not the end of the world if it fails, and it seems like we can leave retrying ephemeral key retrieval to the developer's implementation of `createCustomerKey`, e.g. if they have some retry logic in their API client.

Stripe/STPAPIRequest.m
bg-stripe: +1 on making this change in a separate PR / future version – added a task to track it

Stripe/PublicHeaders/STPCustomerContext.h
bg-stripe: Oh good idea adding a manual cache clearing method @bdorfman-stripe , I'll do that.

Stripe/STPAPIRequest.h
bg-stripe: ah yep, i'll rename here :)

bg-stripe:
@bdorfman-stripe made some more updates:
- removed `cachedCustomerMaxAge` from CustomerContext's public interface
- added a `clearCachedCustomer` method
- renamed `serializer` methods to `deserializer`

joeydong-stripe:
🎉 🎉 🎉 🎉

PR #684, opened 2017-05-19
add Carthage copy-frameworks build phase

bg-stripe:
r? @bdorfman-stripe 

Just realized I forgot to add this build phase in #680 – without it, the example app crashes when running on a device.


bdorfman-stripe:
Ah yeah good call. LGTM

PR #686, opened 2017-05-24
Move mocks to STPMocks

bg-stripe:
r? @bdorfman-stripe 

I also noticed the PII/file upload tests also weren't added to the project / test target (maybe due to a bad merge) – they're back now.


bdorfman-stripe:
👍 

Does this bring master back into line with sources-ui?

bg-stripe:
yep!

PR #688, opened 2017-05-24
Make didSelectPaymentMethod delegate method optional

bg-stripe:
r? @bdorfman-stripe 

During an IRL session with @irace today, we realized that we should make this method optional – if your settings page just has a "Change Payment Method" button, and you don't display the specific payment method, you won't need to know when a new method is selected.


bdorfman-stripe:
Makes sense. LGTM!

PR #691, opened 2017-05-30
Auto fill city/state from zip

bdorfman-stripe:
Works for anything using address view model (add card and shipping info).
Use CLGeocoder to get city and state from zip code.
Also moves zip code above city and state field so this makes sense, and requires zip codes to be 5 characters to be valid.


Stripe/STPAddressViewModel.m
```diff
@@ -167,6 +170,60 @@ - (void)addressFieldTableViewCellDidReturn:(STPAddressFieldTableViewCell *)cell
...
+        // Or if neither are non-nil
+        return;
+    }
+    else {
+
+        CLGeocoder *geocoder = [CLGeocoder new];
+        NSString *zipKey = (NSString *) kABPersonAddressZIPKey;
+        NSString *countryCodeKey = (NSString *) kABPersonAddressCountryCodeKey;
+
+        if (zipKey && countryCodeKey) {
```
bdorfman-stripe: This should be caught above i think? The first check in the method is that country code is US and zipCode is valid (although maybe nil zip codes are valid? I can check)

Stripe/STPAddressViewModel.m
```diff
@@ -167,6 +170,60 @@ - (void)addressFieldTableViewCellDidReturn:(STPAddressFieldTableViewCell *)cell
...
+            [geocoder geocodeAddressDictionary:@{zipKey : zipCode,
+                                                 countryCodeKey : _addressFieldTableViewCountryCode}
+                             completionHandler:^(NSArray<CLPlacemark *> * _Nullable placemarks, NSError * _Nullable error) {
+                                 if (placemarks.count > 0 && error == nil) {
+                                     CLPlacemark *placemark = placemarks.firstObject;
+                                     if (cityCell.contents.length == 0 && stateCell.contents.length == 0) {
+                                         // Check contents again to make sure they're still empty
+                                         cityCell.contents = placemark.locality;
```
bdorfman-stripe: 👍

bdorfman-stripe:
Updated

Stripe/STPAddressViewModel.m
bg-stripe: imo there's no harm in over-checking/being paranoid to prevent dict/array crashes – can we add just to be safe?
 could also nil-coalesce in the dictionary, e.g. `zipKey : zipCode ?: @""`.

bdorfman-stripe: Yep agreed. I do the check above at the start of the method now.

PR #693, opened 2017-06-01
Allow scanning cards via card.io

bg-stripe:
r? @bdorfman-stripe @joeydong-stripe 
cc @mrmcduff-stripe 

I experimented with the various configuration options [0], and settled on what I think is pretty reasonable default behavior:
* hiding the card.io logo
* disabling manual entry
* disabling presenting the scanned image for 0.1 sec (feels faster without this)

card.io allows changing the green guide color, and I considered setting that to the STPTheme's accent color, but decided against it – it's more important that the guides are bright. There's also an option to set the confirmation nav's barTintColor, but without other options (nav bar text color, background color) it doesn't make sense to use an STPTheme color for this.

[0] https://github.com/card-io/card.io-iOS-SDK/blob/master/CardIO/CardIOPaymentViewController.h


bg-stripe:
Looks like cloning card.io is too slow and breaks our tests. I'll add an install_cardio script to remedy this.

Stripe/STPAnalyticsClient.m
```diff
@@ -192,7 +210,7 @@ - (void)logSourceCreationAttemptWithConfiguration:(STPPaymentConfiguration *)con
     [payload addEntriesFromDictionary:@{
                                         @"event": @"stripeios.source_creation",
                                         @"source_type": sourceType ?: @"unknown",
-                                        @"apple_pay_enabled": @([Stripe deviceSupportsApplePay]),
```
bg-stripe: (this key is duplicated in `commonPayload`)

Example/Stripe iOS Example (Simple).xcworkspace/contents.xcworkspacedata
```diff
@@ -1,10 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
bg-stripe: this workspace was broken, and unnecessary, so I removed it

Example/Stripe iOS Example (Simple).xcodeproj/project.pbxproj
```diff
@@ -243,19 +243,19 @@
...
 			inputPaths = (
 			);
-			name = "Install Dependencies";
+			name = "Check Dependencies";
 			outputPaths = (
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "file=\"$PROJECT_DIR/Cartfile.resolved\";\nif [ ! -f \"$file\" ]; then\n    printf \"error: To build the example app, you'll need to first run the setup script.\\n\";\n    printf \"error: Please run ./setup.sh , and then try building again. \\n\";\n    exit 1;\nfi";
+			shellScript = "alamofire=\"$PROJECT_DIR/Carthage/Build/iOS/Alamofire.framework\";\nif [ ! -d \"$alamofire\" ]; then\n    printf \"error: To build the example app, you'll need to first run the setup script.\\n\";\n    printf \"error: Please run ./setup.sh , and then try building again. \\n\";\n    exit 1;\nfi";
```
joeydong-stripe: Should we just check in the dependency?

I guess this question is more complicated now that we have this soft dependency on `card.io`.

install_cardio.rb
```diff
@@ -0,0 +1,43 @@
...
+project_path = 'Example/Stripe iOS Example (Simple).xcodeproj'
+project = Xcodeproj::Project.open(project_path)
+
+# add framework
+lib_path = 'Carthage/Build/iOS/CardIO.framework'
+libRef = project['Frameworks'].new_file(lib_path);
+framework_buildphase = project.objects.select {
+  |x| x.class == Xcodeproj::Project::Object::PBXFrameworksBuildPhase
+}[0];
+framework_buildphase.add_file_reference(libRef);
```
joeydong-stripe: I wonder if we can shorten this script a little bit by defining it as an optional framework:

screen shot 2017-06-08 at 4 29 36 pm

Example/Stripe iOS Example (Simple).xcodeproj/project.pbxproj
bg-stripe: It would be nice to avoid the `setup` step to run the example app. I decided against checking in Alamofire because your swift version might not match the checked-in framework's version. Mayybe we could check in Alamofire, and add a build script that tells you to run `./setup.sh` if it detects a version mismatch.

install_cardio.rb
bg-stripe: Hmm, did some fiddling and reading, and I don't think this would work for our use case. As I understand it, weak linking is used for system frameworks – it conditionally links your app to the framework, so when your app runs on an OS version that doesn't have the framework, it won't crash. Even if a framework is weak linked, it still needs to be present on your machine in order to build the app.

Example/Stripe iOS Example (Simple).xcodeproj/project.pbxproj
bdorfman-stripe: I think it's fine to ask people to run something like this to set up the example app. It also has the benefit of making the example more "real" in some senses, in that if you were really making an app this is a way you could add in dependencies.

Stripe/STPAnalyticsClient.m
```diff
@@ -246,6 +264,7 @@ + (NSMutableDictionary *)commonPayload {
     payload[@"app_name"] = [NSBundle stp_applicationName];
     payload[@"app_version"] = [NSBundle stp_applicationVersion];
     payload[@"apple_pay_enabled"] = @([Stripe deviceSupportsApplePay]);
+    payload[@"card_io_enabled"] = @([STPCardIOProxy isCardIOAvailable]);
```
bdorfman-stripe: Do we want to make this less card.io specific in case we support other OCR in the future? Something like key=ocr_type value=none/card_io

Stripe/STPCardIOProxy.m
```diff
@@ -0,0 +1,129 @@
...
+//
+//  Created by Ben Guo on 5/12/17.
+//  Copyright © 2017 Stripe, Inc. All rights reserved.
+//
+
+#import "STPCardIOProxy.h"
+#import "STPCardParams.h"
+#import "STPAnalyticsClient.h"
+
+@protocol STPClassProxy
```
bdorfman-stripe: Would it be easier to copy a few of the CardIO headers into our project instead of doing the proxy this way? Then you can just do something like `if ([CardIOUtilities class]) { //foo }` and use the class directly if it exists (the same way we check for Apple class availability on older OS versions).

Stripe/STPAnalyticsClient.m
bg-stripe: Good call – I'll update.

Stripe/STPCardIOProxy.m
bg-stripe: Hmm – I think I still prefer this approach. It feels more lightweight not to add a bunch of external headers, and I like that these proxy classes explicitly only define the selectors that we use. I also like that this approach guards against our SDK trying to call into an incompatible version of card.io, by checking that each selector we call is present. We could still do that if we copied the headers, but I like that this way feels more explicit.

bg-stripe:
> What do you think about adding more support for people doing custom integrations? Is it worth adding in say, a category method on a CardIO object that returns you STPCardParams? (so you can pass directly in to STPPaymentCardTextField)

Hmm, I'm not sure this is worthwhile in the end – it's not that hard to translate between `STPCardParams` and `CardIOCreditCardInfo`. I also worry that if we provide a category method, we can't guarantee that it works on every version of their SDK, since their property names could change.

bdorfman-stripe:
All sounds good to me.

PR #690, opened 2017-05-30
update Apple Pay helpers with variants that allow specifying a country

bg-stripe:
r? @bdorfman-stripe 
cc @jack-stripe @mrmcduff-stripe 

This adds 3 new methods, and fleshes out our docs.
- `Stripe.supportedPaymentNetworksForCountry("GB")`
- `Stripe.deviceSupportsApplePayInCountry("GB")`
- `Stripe.paymentRequestWithMerchantIdentifier("merchant.com.foo", country: "GB")`

PR #694, opened 2017-06-01
Add paymentCountry and update Apple Pay helpers

bg-stripe:
r? @bdorfman-stripe 
Follow-up to #690 

- Deprecated `paymentRequestWithMerchantIdentifier:` in favor of a new variant that allows setting  country and currency.
- Added a `paymentCountry` property to `PaymentContext` that determines the `PKPaymentRequest`s countryCode.
- Per discussion, we're not going to constrain `supportedNetworks` based on country.


bdorfman-stripe:
LGTM

PR #695, opened 2017-06-08
Add temporary pull_translations script

bg-stripe:
r? @bdorfman-stripe 

Hopefully this will be useful – I manually added "Scan Card" translations for #693 and it was pretty tedious, so I wrote this script.

### DIFF
--- a/ci_scripts/temp_pull_translations.rb
+++ b/ci_scripts/temp_pull_translations.rb
@@ -1,0 +1,120 @@
+#!/usr/bin/env ruby
+# encoding: utf-8
+#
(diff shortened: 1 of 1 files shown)


bdorfman-stripe:
I wonder if it would be ok to let this update existing translations for current keys. Eg the script could specifically just do phrase fetching and updating for keys that exist locally (instead of pulling all strings from phrase).

bg-stripe:
Ah yep – I can tweak this so that it updates existing translations too.

bdorfman-stripe:
👍

PR #696, opened 2017-06-08
Use 10.3.1 simulator for tests

bg-stripe:
r? @bdorfman-stripe 

Tests are failing [0] because there's isn't a 10.3 simulator. It seems like this was caused by the Xcode 8.3.3 update [1]. I'll file an issue with travis, but for now we can just use the 10.3.1 simulator.

[0] https://travis-ci.org/stripe/stripe-ios/jobs/240876722
[1] https://forums.developer.apple.com/thread/79083


bdorfman-stripe:
👍

PR #697, opened 2017-06-09
Retrieve and update customer.shipping info via ephemeral key

bg-stripe:
r? @bdorfman-stripe 
cc @joeydong-stripe @mrmcduff-stripe 

(branching off `bg-ephemeral-keys`)

`PaymentContext` will now populate `shippingAddress` using the customer object's shipping info, and update the customer object when `shippingAddress` is updated.

`STPAddress` and `customer.shipping` have different shapes:
```
"shipping": {
    "address": {
        "city": "New York",
        "country": "US",
        "line1": "27 Smith St",
        "line2": "Apt 2",
        "postal_code": "10001",
        "state": "NY"
    },
    "name": "Jenny Rosen",
    "phone": "5555555555"
}

 STPAddress *address = [STPAddress new];
 address.name = @"Jenny Rosen";
 address.phone = @"5555555555";
 address.email = @"jrosen@example.com";
 address.line1 = @"27 Smith St";
 address.line2 = @"Apt 2";
 address.postalCode = @"10001";
 address.city = @"New York";
 address.state = @"NY";
 address.country = @"US";
```

Converting between them is pretty straightforward, with one exception: there's no `email` field in `customer.shipping` – email lives one level up on the customer object: `customer.email`. Requests auth'd with an ephemeral key currently don't have access to `customer.email`, but I'll follow up on whether we can add that.

### DIFF
--- a/Stripe/PublicHeaders/STPCustomer.h
+++ b/Stripe/PublicHeaders/STPCustomer.h
@@ -10,6 +10,8 @@
 #import "STPAPIResponseDecodable.h"
 #import "STPSourceProtocol.h"
 
+@class STPAddress;
+
 NS_ASSUME_NONNULL_BEGIN
 
 /**
@@ -49,6 +51,11 @@ NS_ASSUME_NONNULL_BEGIN
  *  The available payment sources the customer has (this may be an empty array).
  */
 @property(nonatomic, readonly) NSArray<id<STPSourceProtocol>> *sources;
+
+/**
+ *  The customer's shipping address.
+ */
+@property(nonatomic, readonly) STPAddress *shippingAddress;
 
 @end
 
--- a/Stripe/PublicHeaders/STPPaymentContext.h
+++ b/Stripe/PublicHeaders/STPPaymentContext.h
@@ -154,6 +154,13 @@ NS_ASSUME_NONNULL_BEGIN
  *  the current customer object. When PaymentContext loads, if you haven't
  *  manually set a prefilled value, any shipping information saved on the customer 
  *  will be used to prefill the shipping address form.
+ *
+ *  You should not rely on the shipping information stored on the Stripe customer 
+ *  for order fulfillment, as your user may change this information if they make 
+ *  multiple purchases. We recommend adding shipping information when you create
+ *  a chart (which can also help prevent fraud), or saving it to your own
+ *  database. https://stripe.com/docs/api#create_charge-shipping
+ *
  *  Note: by default, your user will still be prompted to verify a prefilled 
  *  shipping address. To change this behavior, you can set 
  *  `verifyPrefilledShippingAddress` to NO in your `STPPaymentConfiguration`.
--- a/Stripe/PublicHeaders/STPPaymentContext.h
+++ b/Stripe/PublicHeaders/STPPaymentContext.h
@@ -158,7 +158,7 @@ NS_ASSUME_NONNULL_BEGIN
  *  You should not rely on the shipping information stored on the Stripe customer 
  *  for order fulfillment, as your user may change this information if they make 
  *  multiple purchases. We recommend adding shipping information when you create
- *  a chart (which can also help prevent fraud), or saving it to your own
+ *  a charge (which can also help prevent fraud), or saving it to your own
  *  database. https://stripe.com/docs/api#create_charge-shipping
  *
  *  Note: by default, your user will still be prompted to verify a prefilled 
--- a/Stripe/PublicHeaders/STPPaymentContext.h
+++ b/Stripe/PublicHeaders/STPPaymentContext.h
@@ -148,6 +148,15 @@ NS_ASSUME_NONNULL_BEGIN
 
 /**
  *  The user's shipping address. May be nil.
+ *  If you've already collected a shipping address from your user, you may
+ *  prefill it by setting a shippingAddress in PaymentContext's prefilledInformation.
+ *  When your user enters a new shipping address, PaymentContext will save it to 
+ *  the current customer object. When PaymentContext loads, if you haven't
+ *  manually set a prefilled value, any shipping information saved on the customer 
+ *  will be used to prefill the shipping address form.
+ *  Note: by default, your user will still be prompted to verify a prefilled 
+ *  shipping address. To change this behavior, you can set 
+ *  `verifyPrefilledShippingAddress` to NO in your `STPPaymentConfiguration`.
  */
 @property(nonatomic, readonly, nullable)STPAddress *shippingAddress;
 
--- a/Stripe/PublicHeaders/STPPaymentContext.h
+++ b/Stripe/PublicHeaders/STPPaymentContext.h
@@ -153,7 +153,10 @@ NS_ASSUME_NONNULL_BEGIN
  *  When your user enters a new shipping address, PaymentContext will save it to 
  *  the current customer object. When PaymentContext loads, if you haven't
  *  manually set a prefilled value, any shipping information saved on the customer 
- *  will be used to prefill the shipping address form.
+ *  will be used to prefill the shipping address form. Note that because your
+ *  customer's email may not be the same as the email provided with their shipping
+ *  info, PaymentContext will not prefill the shipping form's email using your 
+ *  customer's email.
  *
  *  You should not rely on the shipping information stored on the Stripe customer 
  *  for order fulfillment, as your user may change this information if they make 
--- a/Stripe/STPCustomer.m
+++ b/Stripe/STPCustomer.m
@@ -83,6 +83,9 @@
         shipping.state = shippingDict[@"address"][@"state"];
         shipping.postalCode = shippingDict[@"address"][@"postal_code"];
         shipping.country = shippingDict[@"address"][@"country"];
+        // Note: email is currently one level above in the customer API,
+        // but the SDK stores it in customer.shippingAddress
+        shipping.email = dict[@"email"];
         customer.shippingAddress = shipping;
     }
     NSMutableArray *sources = [NSMutableArray array];
--- a/Stripe/STPCustomer.m
+++ b/Stripe/STPCustomer.m
@@ -8,20 +8,18 @@
 
 #import "STPCustomer.h"
 
+#import "STPAddress.h"
 #import "STPCard.h"
 #import "STPSource.h"
 #import "StripeError.h"
+#import "NSDictionary+Stripe.h"
 
-/**
- NOTE: STPCustomer has been deprecated. When we remove STPBackendAPIAdapter,
- we should rename STPCustomer with STPLimitedCustomer, to avoid confusion –
- customers retrieved using an ephemeral key have a limited set of fields.
- */
 @interface STPCustomer()
 
 @property(nonatomic, copy)NSString *stripeID;
 @property(nonatomic) id<STPSourceProtocol> defaultSource;
 @property(nonatomic) NSArray<id<STPSourceProtocol>> *sources;
+@property(nonatomic) STPAddress *shippingAddress;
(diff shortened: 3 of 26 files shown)


Stripe/STPPaymentContext.m
```diff
@@ -145,6 +146,7 @@ - (void)retryLoading {
                 [self.loadingPromise fail:error];
                 return;
             }
+            self.shippingAddress = customer.shipping;
```
bdorfman-stripe: I think we need to add something in here to distinguish between addresses that are auto-filled in from the server and addresses which have been definitively viewed and approved by the user. I think also maybe payment context should not skip the "show shipping VC" step of request payment if the user has not verified this autofilled address (or maybe that behavior should be configurable).

bg-stripe: Ah, yep – making some additional updates here also to allow prefilling `shippingAddress` programmatically. I think the behavior should probably be: don't skip the "show shipping VC" step of request payment if:
  - the address is incomplete
  - the user hasn't verified the address
  - the delegate implements `paymentContext:didUpdateShippingAddress:` and `selectedShippingMethod` is nil

bg-stripe:
Ready for re-review – I've made some updates:
  - `shippingAddress`  is now readwrite. If you prefill `shippingAddress` manually, or it gets filled from the Customer object, or your user hasn't selected a shipping method, `requestPayment` will still present the shipping VC.
  - we now pass shipping address and method in the example app's `completeCharge` method. 
there's a small example backend change that goes along with this – I'll include it with the ephemeral keys example backend PR.

bg-stripe:
re-r? @bdorfman-stripe

Stripe/PublicHeaders/STPCustomer.h
```diff
@@ -50,6 +52,11 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property(nonatomic, readonly) NSArray<id<STPSourceProtocol>> *sources;
 
+/**
+ *  The customer's shipping address.
+ */
+@property(nonatomic, readonly) STPAddress *shipping;
```
bdorfman-stripe: just `shipping` is a bit confusing. I think we call this `shippingAddress` in other places.

Stripe/PublicHeaders/STPPaymentContext.h
```diff
@@ -148,8 +148,16 @@ NS_ASSUME_NONNULL_BEGIN
...
+ *  If you've already collected a shipping address from your user, you may
+ *  prefill it by setting this property. When your user enters a new address,
+ *  PaymentContext will save it to the current customer object. When PaymentContext
+ *  loads, if you haven't manually set a prefilled value, any shipping information
+ *  saved on the customer will be used to prefill shippingAddress.
+ *  Note: by default, your user will still be prompted to verify a prefilled 
+ *  shipping address. To change this behavior, you can set 
+ *  `verifyPrefilledShippingAddress` to NO in your `STPPaymentConfiguration`.
+ */
+@property(nonatomic, nullable)STPAddress *shippingAddress;
```
bdorfman-stripe: I think it's somewhat confusing that the shipping prefill is on payment context, but all the other prefills are in the prefilled user info block. I think we should reconsider this structure. I think this might be a good time to also clear up how properties that are listed in multiple places are used (like phone/email)

Stripe/STPCustomer.m
```diff
@@ -8,15 +8,23 @@
...
+#import "STPAddress.h"
 #import "STPCard.h"
 #import "STPSource.h"
 #import "StripeError.h"
+#import "NSDictionary+Stripe.h"
 
+/**
+ NOTE: STPCustomer has been deprecated. When we remove STPBackendAPIAdapter,
+ we should rename STPCustomer with STPLimitedCustomer, to avoid confusion –
+ customers retrieved using an ephemeral key have a limited set of fields.
```
bdorfman-stripe: Not sure this rename is really needed, especially given that most people probably aren't using it except through payment context?

Stripe/STPCustomer.m
```diff
@@ -8,15 +8,23 @@
...
+ */
 @interface STPCustomer()
 
 @property(nonatomic, copy)NSString *stripeID;
 @property(nonatomic) id<STPSourceProtocol> defaultSource;
 @property(nonatomic) NSArray<id<STPSourceProtocol>> *sources;
+@property(nonatomic) STPAddress *shipping;
```
bdorfman-stripe: Again, I think `shippingAddress`.

Stripe/STPCustomerContext+Private.h
```diff
@@ -0,0 +1,23 @@
...
+
+#import "STPBlocks.h"
+
+@class STPAddress;
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface STPCustomerContext (Private)
+
+- (void)updateCustomerWithShipping:(STPAddress *)shipping completion:(nullable STPErrorBlock)completion;
```
bdorfman-stripe: `ShippingAddress:` ?

Also should this be public?

Stripe/PublicHeaders/STPCustomer.h
bg-stripe: 👍 – I called it shipping to match the API object, but I think you're right that it's better to use shippingAddress

Stripe/STPCustomer.m
bg-stripe: ah yeah, meant to delete this comment, I think it got carried over from an old version of `sources-ui`. I agree, i think it's fine to stick with STPCustomer.

Stripe/STPCustomerContext+Private.h
bg-stripe: 👍  I'll rename.

I think we should keep `CustomerContext`'s public API surface really small for now – ideally the `BackendAPIAdapter` methods would be private too, but that's not possible right now. I think if we decide to expose customer methods publicly for standalone use without PaymentContext, we should think more about the use case and maybe expose a different API.

bdorfman-stripe: Sounds good.

Stripe/PublicHeaders/STPPaymentContext.h
bg-stripe: Ah, that's a good point – I can move `shippingAddress` to STPUserInformation. 

I'll try to unify/document our behavior for phone and email too.

bg-stripe:
@bdorfman-stripe updated!

Stripe/STPPaymentContext.m
```diff
@@ -205,6 +201,13 @@ - (void)setDelegate:(id<STPPaymentContextDelegate>)delegate {
     }];
 }
 
+- (void)setPrefilledInformation:(STPUserInformation *)prefilledInformation {
+    _prefilledInformation = prefilledInformation;
+    if (prefilledInformation.shippingAddress) {
+        self.shippingAddress = prefilledInformation.shippingAddress;
```
bdorfman-stripe: I'm not sure you want to clobber `self.shippingAddress` here (it probably should only be set from a successful view of the VC?). At minimum setting the prefil shouldn't overwrite if we already have a shipping address set.

Suggested flow: A nil `_shippingAddress` until it is confirmed by the user (you can use this to track verification state). Possibly override the getter so that if the instance var is nil, it returns `self.prefilledInformation.shippingAddress`

Stripe/STPShippingAddressViewController.m
```diff
@@ -89,16 +89,8 @@ - (instancetype)initWithConfiguration:(STPPaymentConfiguration *)configuration
...
-            STPAddress *prefilledAddress = [STPAddress new];
-            prefilledAddress.country = _addressViewModel.address.country;
-            if (self.configuration.requiredShippingAddressFields & PKAddressFieldEmail) {
-                prefilledAddress.email = prefilledInformation.email;
-            }
-            if (self.configuration.requiredShippingAddressFields & PKAddressFieldPhone) {
-                prefilledAddress.phone = prefilledInformation.phone;
-            }
-            _addressViewModel.address = prefilledAddress;
+        else if (prefilledInformation.shippingAddress != nil) {
```
bdorfman-stripe: Is this prefill completely redundant now since you pass in shippingAddress directly as a param?

bg-stripe: Yeah, this part is pretty redundant. We need to pass in `prefilledInformation` to get `billingAddress` though, so we still need to handle `prefilledInformation.shippingAddress` appropriately.

Stripe/STPPaymentContext.m
bg-stripe: ah yeah, good point – I'll check if we already have a shipping address set, also set `shippingAddressNeedsVerification = YES;`.

I think I still prefer using a `shippingAddressNeedsVerification` flag here rather than using a nil `_shippingAddress`, since things get complicated with the `shippingAddressNeedsVerification` configuration option and `customer.shipping`, and I think it's better to be more explicit.

bg-stripe:
> Need to also update MIGRATING with whatever prefill changes you end up making.
Yep! Updated MIGRATING, and also fixed the `prefilledInformation.shippingAddress` logic. 
One more 👀 ?

bdorfman-stripe:
Does this actually update Customer email correctly? In our api, email is at the top level under customer, not part of shipping address.

bg-stripe:
> Does this actually update Customer email correctly? In our api, email is at the top level under customer, not part of shipping address.

It doesn't, due to a current restriction with ephemeral keys – noted this in my comment at the top. Requests auth'd with an ephemeral key currently don't have access to `customer.email`. I think this is ok behavior for v1, and it shouldn't be too hard to add in the future. wdyt? I can add a task to track  this.

bdorfman-stripe:
I think this is not great, given that we claim we update shipping info but email is just totally dropped on the floor undocumented. I think we should really try to get this fixed on the server end before shipping (we've already had a lot of confusion/complaints in the past about how the email for Remember Me doesn't actually make it to the customer at all).

bg-stripe:
That's reasonable. Just to avoid delaying things more, what do you think about leaving this PR out of 11.0, and saving it for an 11.1 release? I can update the docs and `ephemeral-keys` branch so we no longer reference this change.

bdorfman-stripe:
I think that's ok if it's going to be a big delay. Maybe sync with the appropriate people to see how quickly email access via ephemeral key can be added?

bg-stripe:
Cool. It sounds like this can actually be added pretty quickly, so no need to punt on this.

joeydong-stripe:
👀

bg-stripe:
@bdorfman-stripe ephemeral keys can now update customer.email, and I've updated the PR in 3fa6e45 to save/retrieve email too

bdorfman-stripe:
I'm now second-guessing my request to add email. How concerned are we that `customer.email` might not actually be the same as `shipping.email`? (eg for shipping an item to someone that is not the customer)

Stripe/PublicHeaders/STPPaymentContext.h
```diff
@@ -154,6 +154,13 @@ NS_ASSUME_NONNULL_BEGIN
  *  the current customer object. When PaymentContext loads, if you haven't
  *  manually set a prefilled value, any shipping information saved on the customer 
  *  will be used to prefill the shipping address form.
+ *
+ *  You should not rely on the shipping information stored on the Stripe customer 
+ *  for order fulfillment, as your user may change this information if they make 
+ *  multiple purchases. We recommend adding shipping information when you create
+ *  a chart (which can also help prevent fraud), or saving it to your own
```
bdorfman-stripe: chart should be charge

bg-stripe: oh d'oh, thanks

PR #701, opened 2017-06-13
Create source from pkpayment

bdorfman-stripe:


Stripe/PublicHeaders/STPPaymentContext.h
```diff
@@ -133,7 +133,7 @@ NS_ASSUME_NONNULL_BEGIN
  *
  *  @warning `PKPaymentSummaryItem` is only available in iOS8+. If you support iOS 7 you should do a runtime availability check before accessing or setting this property. 
  */
-@property(nonatomic, copy)NSArray<PKPaymentSummaryItem *> *paymentSummaryItems NS_AVAILABLE_IOS(8_0);
+@property(nonatomic, copy)NSArray<PKPaymentSummaryItem *> *paymentSummaryItems;
```
bg-stripe: should update the iOS7 comment above too

bdorfman-stripe: Ah good catch

PR #703, opened 2017-06-14
Update to be able to build against iOS 11 SDK

Daniel1of1:
## Summary

- Add new iOS 11 APIs to conform to new `PKPaymentAuthorizationViewControllerDelegate` protocol.
- This should have no change on current supported versions.
- Currently none of the additional functionality available has been leveraged.
- Happy for this to be an open discussion 😄.

## Motivation

iOS 11 introduced some breaking API changes in ApplePay, which are used in `Stripe/PKPaymentAuthorizationViewController+Stripe_Blocks.m` specifically `PKPaymentAuthorizationViewControllerDelegate`. see [API diffs](https://developer.apple.com/documentation/passkit/pkpaymentauthorizationviewcontrollerdelegate?changes=latest_minor)

This PR allows anyone to use `stripe-ios` as a dependancy while building against iOS 11 (beta) SDK.
It will have no effect when deploying to versions already supported, and on iOS 11 it makes the best attempt to have the same behaviour as previous versions (which is to say _not_ taking advantage of a couple of new features around better error displaying.)

see [What's new in Apple Pay & Wallet (2017)](https://developer.apple.com/videos/play/wwdc2017/714/) for further reference.

## Testing

None I'm afraid - but not breaking anything existing at least 😅

### DIFF
--- a/Stripe/PKPaymentAuthorizationViewController+Stripe_Blocks.m
+++ b/Stripe/PKPaymentAuthorizationViewController+Stripe_Blocks.m
@@ -33,8 +33,30 @@ typedef void (^STPPaymentAuthorizationStatusCallback)(PKPaymentAuthorizationStat
 
 @implementation STPBlockBasedApplePayDelegate
 
+#if __IPHONE_OS_VERSION_MAX_ALLOWED >= 110000 // __IPHONE_11_0
+-(PKPaymentAuthorizationResult * _Nonnull)resultFromStatus:(PKPaymentAuthorizationStatus)status {
+    //TODO: can improve errors from PKPaymentAuthorizationStatus
+    return [[PKPaymentAuthorizationResult alloc] initWithStatus:status errors:nil];
+}
+
+// required iOS 11 API - call through to old method to maintain similar functionality
+// Does not take advantage of providing more granular information
 - (void)paymentAuthorizationViewController:(__unused PKPaymentAuthorizationViewController *)controller
-                       didAuthorizePayment:(PKPayment *)payment completion:(STPPaymentAuthorizationStatusCallback)completion {
+                       didAuthorizePayment:(__unused PKPayment *)payment
+                                   handler:(__unused void (^)(PKPaymentAuthorizationResult * _Nonnull))completion {
+    
+    [self paymentAuthorizationViewController:controller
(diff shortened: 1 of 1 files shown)


Daniel1of1:
FWIW the failed travis test is an installation test (carthage) which seems to fail semi-consistently e.g. https://travis-ci.org/stripe/stripe-ios/builds/242586409?utm_source=github_status&utm_medium=notification

bdorfman-stripe:
Thanks for your submission! 

It looks from the API diffs that the old methods are not actually removed, just deprecated, and so the current iOS 10 version should continue to work? Is this not the case? We will evaluate support for iOS 11, however I think we might want to wait for it to be closer to GM before comitting changes to our SDK as often Apple SDKs can change from beta to beta.

Daniel1of1:
@bdorfman-stripe hey thanks for the response - so:

> It looks from the API diffs that the old methods are not actually removed...

Correct however there is still a build error. This is caused by the new method:

```objective-c 
paymentAuthorizationViewController:didAuthorizePayment:handler:
```

It's `required` so `STPBlockBasedApplePayDelegate`  no longer conforms to `PKPaymentAuthorizationViewControllerDelegate` in iOS 11. The other two methods added are indeed only optional so were just there for feature completion.

> We will evaluate support for iOS 11 ...

Sure totally understand + agree on all the beta bits 😄.

Along those lines that was my intention - to not go too far into incorporating the new APIs since this is a decision I'm sure your team would like to make. 

So I have no strong feelings... since for my particular use case (dependancy managed with Carthage), I can simply point it at a fork (thanks for open sourcing it) - as could anyone else with the same issue, but happy to move forward / modify if you think it will be of use.

Daniel1of1:
Oh one last thing - I realised I'm PR'ing into master. This should instead be into a working `iOS-11` branch, but obviously I can't create one.

bg-stripe:
Hey @Daniel1of1 , thanks for getting started on this! We've created an iOS 11 branch [here](https://github.com/stripe/stripe-ios/tree/ios11), which essentially has the same changes in your PR. We'll be evaluating the iOS 11 API changes further closer to the GM.

Daniel1of1:
Awesome! Thanks @bg-stripe.

PR #707, opened 2017-06-16
Add description implementations

joeydong-stripe:
Wanted to get some early feedback on this before I dive into adding description implementations to a bunch more classes.

- What do you think about the formatting?
  - I think there's a little too much spacing but it's nice that it looks like stock iOS descriptions
- Should we make `[STPCard stringFromFunding:]` public or just keep it private?
- How does the description string building pattern / ordering feel?
- I decided to redact the cardholder details. Are there others we should consider?

Here's what the output would look like:

```
<__NSArrayI 0x6000002454f0>(
<STPCard: 0x600000151930; cardId = card_1AUJxDEOD54MuFwSyKCDbNuU; brand = Visa; last4 = 1881; expMonth = 11; expYear = 2029; funding = Credit; country = CA; currency = (null); dynamicLast4 = (null); isApplePayCard = NO; name = (null); address = (null)>,
<STPCard: 0x600000151b40; cardId = card_1AUJunEOD54MuFwSnU9HEPUO; brand = Visa; last4 = 4242; expMonth = 4; expYear = 2024; funding = Credit; country = US; currency = (null); dynamicLast4 = (null); isApplePayCard = NO; name = (null); address = (null)>,
<STPCard: 0x600000151670; cardId = card_1AVRojEOD54MuFwSxr93QJSx; brand = Visa; last4 = 5556; expMonth = 12; expYear = 2034; funding = Debit; country = US; currency = (null); dynamicLast4 = (null); isApplePayCard = NO; name = <redacted>; address = <redacted>>
)
```

cc @bg-stripe @bdorfman-stripe

### DIFF
--- a/Stripe/PublicHeaders/STPCard.h
+++ b/Stripe/PublicHeaders/STPCard.h
@@ -78,6 +78,15 @@ typedef NS_ENUM(NSInteger, STPCardFundingType) {
 + (STPCardFundingType)fundingFromString:(NSString *)string;
 
 /**
+ * Returns a string representation for the provided card funding
+ *
+ * @param funding The enum value you want to covert to a string
+ *
+ * @return A string representation of the funding, suitable for displaying to a user
+ */
++ (NSString *)stringFromFunding:(STPCardFundingType)funding;
+
+/**
  *  The last 4 digits of the card.
  */
 @property (nonatomic, readonly) NSString *last4;
--- a/Stripe/PublicHeaders/STPCard.h
+++ b/Stripe/PublicHeaders/STPCard.h
@@ -78,15 +78,6 @@ typedef NS_ENUM(NSInteger, STPCardFundingType) {
 + (STPCardFundingType)fundingFromString:(NSString *)string;
 
 /**
- * Returns a string representation for the provided card funding
- *
- * @param funding The enum value you want to covert to a string
- *
- * @return A string representation of the funding, suitable for displaying to a user
- */
-+ (NSString *)stringFromFunding:(STPCardFundingType)funding;
-
-/**
  *  The last 4 digits of the card.
  */
 @property (nonatomic, readonly) NSString *last4;
--- a/Stripe/STPCard.m
+++ b/Stripe/STPCard.m
@@ -97,19 +97,6 @@
     }
 }
 
-+ (NSString *)displayStringFromFunding:(STPCardFundingType)funding {
-    switch (funding) {
-        case STPCardFundingTypeCredit:
-            return @"Credit";
-        case STPCardFundingTypeDebit:
-            return @"Debit";
-        case STPCardFundingTypePrepaid:
-            return @"Prepaid";
-        case STPCardFundingTypeOther:
-            return @"Other";
-    }
-}
-
 - (instancetype)init {
     self = [super init];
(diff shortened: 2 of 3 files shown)


Stripe/PublicHeaders/STPCard.h
```diff
@@ -78,6 +78,15 @@ typedef NS_ENUM(NSInteger, STPCardFundingType) {
 + (STPCardFundingType)fundingFromString:(NSString *)string;
 
 /**
+ * Returns a string representation for the provided card funding
+ *
+ * @param funding The enum value you want to covert to a string
+ *
+ * @return A string representation of the funding, suitable for displaying to a user
+ */
++ (NSString *)stringFromFunding:(STPCardFundingType)funding;
```
bdorfman-stripe: Can this be private?

Stripe/STPCard.m
```diff
@@ -135,6 +148,35 @@ - (BOOL)isEqualToCard:(STPCard *)other {
     return [self.cardId isEqualToString:other.cardId];
 }
 
+- (NSString *)description {
```
bdorfman-stripe: Should this be `debugDescription` instead? I think `description` is supposed to be potentially user-facing text theoretically.

Stripe/STPCard.m
```diff
@@ -135,6 +148,35 @@ - (BOOL)isEqualToCard:(STPCard *)other {
...
+    NSMutableArray *descriptionParts = [[NSMutableArray alloc] init];
+
+    // NSObject
+    [descriptionParts addObject:[NSString stringWithFormat:@"%@: %p", NSStringFromClass([self class]), self]];
```
bdorfman-stripe: These lines are all very similar, is it possible to do this more automatically somehow to make it easier to add new properties? Eg. something like `propertyNamesToFormFieldNamesMapping`

Stripe/PublicHeaders/STPCard.h
bg-stripe: nit: would prefer calling this `displayStringFromFunding`, to make clear that it's not the inverse of `fundingFromString`

joeydong-stripe: oh nice good points 👍

Stripe/STPCard.m
joeydong-stripe: I just dug into this a little bit. Calling `NSLog(card)` will use the `description` and not `debugDescription`. I feel like that's a common use case that we want to support.

Stripe/STPCard.m
joeydong-stripe: Yeah let me at least make it a big `NSArray` literal.

Stripe/STPCard.m
bdorfman-stripe: Ah ok. I thought the default NSObject implementation of `description` would just forward to `debugDescription` or something, but doing whatever gets proper NSLog out I think is good.

joeydong-stripe:
I'm starting to backpedal on the "string for X" functions and just keeping them inside the `description` implementations for now. The reusability of some of them is not clear yet.

joeydong-stripe:
I'll open a new PR when this is closer to being ready.

PR #711, opened 2017-06-20
Add description implementations for public models

joeydong-stripe:
## Summary

Adding a bunch of `description` implementations that will make logging and debugging objects produced by the SDK a little easier. Decided to start with the public models because I think that would be the most impactful.

## Motivation

Before:

```
<STPCard: 0x600000151930>
```

After:

```
<STPCard: 0x600000151930; cardId = card_1AUJxDEOD54MuFwSyKCDbNuU; brand = Visa; last4 = 1881; expMonth = 11; expYear = 2029; funding = Credit; country = CA; currency = (null); dynamicLast4 = (null); isApplePayCard = NO; name = (null); address = (null)>
```

## Testing

Updated test coverage but could still use some more improvement in a future PR.

r? @bdorfman-stripe 
cc @bg-stripe

### DIFF
--- a/Stripe.xcodeproj/project.pbxproj
+++ b/Stripe.xcodeproj/project.pbxproj
@@ -374,7 +374,22 @@
 		04F94DD31D22A23F004FC826 /* NSBundle+Stripe_AppName.h in Headers */ = {isa = PBXBuildFile; fileRef = 049A3F971CC76A2400F57DE7 /* NSBundle+Stripe_AppName.h */; };
 		04F94DD41D22A242004FC826 /* NSBundle+Stripe_AppName.m in Sources */ = {isa = PBXBuildFile; fileRef = 049A3F981CC76A2400F57DE7 /* NSBundle+Stripe_AppName.m */; };
 		04FCFA191BD59A8C00297732 /* STPCategoryLoader.h in Headers */ = {isa = PBXBuildFile; fileRef = 04FCFA171BD59A8C00297732 /* STPCategoryLoader.h */; };
+		8B429AD81EF9D4B400F95F34 /* STPBankAccountParams+Private.h in Headers */ = {isa = PBXBuildFile; fileRef = 8B429AD71EF9D4A300F95F34 /* STPBankAccountParams+Private.h */; };
+		8B429AD91EF9D4B500F95F34 /* STPBankAccountParams+Private.h in Headers */ = {isa = PBXBuildFile; fileRef = 8B429AD71EF9D4A300F95F34 /* STPBankAccountParams+Private.h */; };
+		8B429ADB1EF9E18400F95F34 /* STPBankAccount+Private.h in Headers */ = {isa = PBXBuildFile; fileRef = 8B429ADA1EF9E15900F95F34 /* STPBankAccount+Private.h */; };
+		8B429ADC1EF9E18400F95F34 /* STPBankAccount+Private.h in Headers */ = {isa = PBXBuildFile; fileRef = 8B429ADA1EF9E15900F95F34 /* STPBankAccount+Private.h */; };
+		8B429ADE1EF9EFF900F95F34 /* STPFile+Private.h in Headers */ = {isa = PBXBuildFile; fileRef = 8B429ADD1EF9EFF600F95F34 /* STPFile+Private.h */; };
+		8B429ADF1EF9EFFA00F95F34 /* STPFile+Private.h in Headers */ = {isa = PBXBuildFile; fileRef = 8B429ADD1EF9EFF600F95F34 /* STPFile+Private.h */; };
 		8B8DDBB31EF887A4004B141F /* STPBankAccountParamsTest.m in Sources */ = {isa = PBXBuildFile; fileRef = 8B8DDBB21EF887A4004B141F /* STPBankAccountParamsTest.m */; };
+		8BD87B881EFB131700269C2B /* STPSourceCardDetails+Private.h in Headers */ = {isa = PBXBuildFile; fileRef = 8BD87B871EFB131400269C2B /* STPSourceCardDetails+Private.h */; };
+		8BD87B891EFB131800269C2B /* STPSourceCardDetails+Private.h in Headers */ = {isa = PBXBuildFile; fileRef = 8BD87B871EFB131400269C2B /* STPSourceCardDetails+Private.h */; };
+		8BD87B8B1EFB136F00269C2B /* STPSourceCardDetailsTest.m in Sources */ = {isa = PBXBuildFile; fileRef = 8BD87B8A1EFB136F00269C2B /* STPSourceCardDetailsTest.m */; };
+		8BD87B8D1EFB152B00269C2B /* STPSourceRedirect+Private.h in Headers */ = {isa = PBXBuildFile; fileRef = 8BD87B8C1EFB152800269C2B /* STPSourceRedirect+Private.h */; };
+		8BD87B8E1EFB152B00269C2B /* STPSourceRedirect+Private.h in Headers */ = {isa = PBXBuildFile; fileRef = 8BD87B8C1EFB152800269C2B /* STPSourceRedirect+Private.h */; };
+		8BD87B901EFB17AA00269C2B /* STPSourceRedirectTest.m in Sources */ = {isa = PBXBuildFile; fileRef = 8BD87B8F1EFB17AA00269C2B /* STPSourceRedirectTest.m */; };
+		8BD87B921EFB1C1E00269C2B /* STPSourceVerification+Private.h in Headers */ = {isa = PBXBuildFile; fileRef = 8BD87B911EFB1C1E00269C2B /* STPSourceVerification+Private.h */; };
+		8BD87B931EFB1C1E00269C2B /* STPSourceVerification+Private.h in Headers */ = {isa = PBXBuildFile; fileRef = 8BD87B911EFB1C1E00269C2B /* STPSourceVerification+Private.h */; };
+		8BD87B951EFB1CB100269C2B /* STPSourceVerificationTest.m in Sources */ = {isa = PBXBuildFile; fileRef = 8BD87B941EFB1CB100269C2B /* STPSourceVerificationTest.m */; };
 		8BE5AE8B1EF8905B0081A33C /* STPCardParamsTest.m in Sources */ = {isa = PBXBuildFile; fileRef = 8BE5AE8A1EF8905B0081A33C /* STPCardParamsTest.m */; };
 		C1080F491CBECF7B007B2D89 /* STPAddress.h in Headers */ = {isa = PBXBuildFile; fileRef = C1080F471CBECF7B007B2D89 /* STPAddress.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		C1080F4A1CBECF7B007B2D89 /* STPAddress.m in Sources */ = {isa = PBXBuildFile; fileRef = C1080F481CBECF7B007B2D89 /* STPAddress.m */; };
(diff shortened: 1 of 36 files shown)


Stripe/STPBankAccount.m
```diff
@@ -54,7 +54,64 @@ - (BOOL)isEqualToBankAccount:(STPBankAccount *)bankAccount {
...
+            accountHolderTypeDescription = @"company";
+            break;
+    }
+
+    NSArray *props = @[
+                       // Object
+                       [NSString stringWithFormat:@"%@: %p", NSStringFromClass([self class]), self],
+
+                       // Identifier
+                       [NSString stringWithFormat:@"bankAccountId = %@", self.bankAccountId],
```
bg-stripe: maybe we should use `NSStringFromSelector` for these to guard against drifting if we rename properties?

Stripe/STPCard.m
```diff
@@ -135,6 +137,54 @@ - (BOOL)isEqualToCard:(STPCard *)other {
     return [self.cardId isEqualToString:other.cardId];
 }
 
+#pragma mark - Description
+
+- (NSString *)description {
+    NSString *fundingDescription;
+
+    switch (self.funding) {
```
bg-stripe: nit: can we pull this out into a `stringFromFunding` method, and refactor `fundingFromString` so that we derive the outputs of both the function and its inverse from the same dictionary? [0]

I know the conversion is kinda tedious and we don't use the pattern super consistently, but I generally like coupling inverse functions like these so they don't drift apart

https://github.com/stripe/stripe-ios/blob/master/Stripe/STPSource.m#L42-L67

Stripe/STPSource.m
```diff
@@ -149,7 +149,60 @@ - (BOOL)isEqualToSource:(STPSource *)source {
     return [self.stripeID isEqualToString:source.stripeID];
 }
 
-#pragma mark STPAPIResponseDecodable
+#pragma mark - Description
+
+- (NSString *)description {
+    NSString *statusDescription;
+
+    switch (self.status) {
```
bg-stripe: same comment as above, can we pull this out into `stringFromStatus` and couple it with `statusFromString`?

Stripe/STPCardParams.m
```diff
@@ -53,6 +53,30 @@ - (void)setAddress:(STPAddress *)address {
...
+- (NSString *)description {
+    NSArray *props = @[
+                       // Object
+                       [NSString stringWithFormat:@"%@: %p", NSStringFromClass([self class]), self],
+
+                       // Basic card details
+                       [NSString stringWithFormat:@"last4 = %@", self.last4],
+                       [NSString stringWithFormat:@"expMonth = %lu", (unsigned long)self.expMonth],
+                       [NSString stringWithFormat:@"expYear = %lu", (unsigned long)self.expYear],
+                       [NSString stringWithFormat:@"cvc = %@", (self.cvc) ? @"<redacted>" : nil],
```
bg-stripe: Seems like it might be worth pulling this into a helper, maybe a private `NSObject` category with `stp_redactedDescription`?

Stripe/STPSourceCardDetails.m
```diff
@@ -61,4 +61,60 @@ + (STPSourceCard3DSecureStatus)threeDSecureStatusFromString:(NSString *)string {
...
+            fundingDescription = @"prepaid";
+            break;
+        case STPCardFundingTypeOther:
+            fundingDescription = @"other";
+            break;
+    }
+
+    NSString *threeDSecureStatusDescription;
+
+    switch (self.threeDSecure) {
```
bg-stripe: (same comment as above here about pulling this out)

Stripe/STPSourceRedirect.m
```diff
@@ -33,7 +33,40 @@ + (STPSourceRedirectStatus)statusFromString:(NSString *)string {
     }
 }
 
-#pragma mark STPAPIResponseDecodable
+#pragma mark - Description
+
+- (NSString *)description {
+    NSString *statusDescription;
+
+    switch (self.status) {
```
bg-stripe: (same comment as above here)

Stripe/STPSourceVerification.m
```diff
@@ -32,7 +32,39 @@ + (STPSourceVerificationStatus)statusFromString:(NSString *)string {
     }
 }
 
-#pragma mark STPAPIResponseDecodable
+#pragma mark - Description
+
+- (NSString *)description {
+    NSString *statusDescription;
+
+    switch (self.status) {
```
bg-stripe: (same comment as above about pulling this out)

Tests/Tests/STPBankAccountParamsTest.m
```diff
@@ -0,0 +1,100 @@
...
+    XCTAssert(bankAccountParams.description);
+}
+
+#pragma mark - STPFormEncodable Tests
+
+- (void)testRootObjectName {
+    XCTAssertEqualObjects([STPBankAccountParams rootObjectName], @"bank_account");
+}
+
+- (void)testPropertyNamesToFormFieldNamesMapping {
```
bg-stripe: nice, good addition :)

Tests/Tests/STPBankAccountParamsTest.m
```diff
@@ -0,0 +1,100 @@
...
+
+- (void)testDescriptionWorks {
+    STPBankAccountParams *bankAccountParams = [[STPBankAccountParams alloc] init];
+    bankAccountParams.accountNumber = @"000123456789";
+    bankAccountParams.routingNumber = @"123456789";
+    bankAccountParams.country = @"US";
+    bankAccountParams.currency = @"usd";
+    bankAccountParams.accountHolderName = @"John Doe";
+    bankAccountParams.accountHolderType = STPBankAccountHolderTypeCompany;
+    XCTAssert(bankAccountParams.description);
```
bg-stripe: nit: should we use a more strict assertion here? I think this would still pass if we didn't implement a custom description. Maybe assert that the description contains the class name?

Tests/Tests/STPBankAccountTest.m
```diff
@@ -78,4 +92,12 @@ - (void)testBankAccountEquals {
     XCTAssertEqualObjects(bankAccount1, bankAccount2, @"bank account with equal data should be equal");
 }
 
+#pragma mark - Description Tests
+
+- (void)testDescriptionWorks {
+    STPBankAccount *bankAccount = [STPBankAccount decodedObjectFromAPIResponse:[self completeAttributeDictionary]];
+    bankAccount.routingNumber = @"123456789";
+    XCTAssert(bankAccount.description);
```
bg-stripe: same comment here, maybe use a stronger assertion?

Tests/Tests/STPCardParamsTest.m
```diff
@@ -0,0 +1,133 @@
...
+    cardParams.cvc = @"123";
+    cardParams.name = @"John Smith";
+    cardParams.addressLine1 = @"55 John St";
+    cardParams.addressLine2 = @"#3B";
+    cardParams.addressCity = @"New York";
+    cardParams.addressState = @"NY";
+    cardParams.addressZip = @"10002";
+    cardParams.addressCountry = @"US";
+    cardParams.currency = @"usd";
+    XCTAssert(cardParams.description);
```
bg-stripe: same comment here, maybe use a stronger assertion?

Tests/Tests/STPCardTest.m
```diff
@@ -93,6 +105,15 @@ - (void)testCardEquals {
     XCTAssertEqualObjects(card1, card2, @"cards with equal data should be equal");
 }
 
+#pragma mark - Description Tests
+
+- (void)testDescriptionWorks {
+    STPCard *card = [STPCard decodedObjectFromAPIResponse:[self completeAttributeDictionary]];
+    XCTAssert(card.description);
```
bg-stripe: same comment as above, maybe use a stronger assertion?

Stripe/STPBankAccount.m
joeydong-stripe: Do you mean something like this?

```
[NSString stringWithFormat:@"%@ = %@", NSStringFromSelector(bankAccountId), self.bankAccountId],
```

I think I like the clarity with what I have here because it breaks down when we have something generated like:

```
[NSString stringWithFormat:@"status = %@", statusDescription]
```

Stripe/STPCard.m
joeydong-stripe: Yeah I agree we can improve here. A few questions about the current pattern:

- Should we be friendly and `lowercaseString` or be more strict?
  - I would lean towards being more strict
- Should every enum have an "Unknown" case? We would need to add one in this case for `funding`.
  - Feels like adding "Unknown" makes downstream code more complicated
- More generally, I feel like we can come up with something that looks more elegant. Maybe leverage the fact that tests will always be run?

Stripe/STPBankAccount.m
bg-stripe: that's fair, I don't feel super strongly about it!

Stripe/STPCard.m
joeydong-stripe: I'll make the changes to align with the current pattern and we can think about it more later.

bg-stripe: - I agree about being more strict
- Ah yeah, good question. I think we should do it on a case-by-case basis. For source type, it definitely makes sense to have an unknown case, since source types are added pretty frequently. In this case, we should probably not add a new unknown case publicly, since we don't expect the enum to change. Maybe we can make a private sentinel value to represent an unknown enum value across the SDK? 
- Definitely agree that the current pattern isn't that elegant, and we can probably come up with something better, open to alternatives!

Stripe/STPCardParams.m
joeydong-stripe: so something like this?:

```
- (NSString *)stp_redacted {
  return @"<redacted>";
}
```

Maybe too clever 😄?

joeydong-stripe: I'm starting to like it but it looks a little extra extraneous out of context.

bg-stripe: heh yeah, it probably does look kinda silly out of context. i don't feel that strongly either way, up to you :)

bg-stripe: i think calling it `stp_redactedDescription` and giving it a lil comment would probably help make it look less extraneous

Stripe/STPBankAccount.m
joeydong-stripe: 👍

Stripe/STPCard.m
joeydong-stripe: 👍

Stripe/STPSource.m
joeydong-stripe: 👍

Stripe/STPCardParams.m
joeydong-stripe: 👍

Stripe/STPSourceCardDetails.m
joeydong-stripe: 👍

Stripe/STPSourceRedirect.m
joeydong-stripe: 👍

Tests/Tests/STPBankAccountParamsTest.m
joeydong-stripe: I might come back to this one in a separate PR since this one is getting a little out of hand. I agree we can use a stricter test but just making sure the custom/stock implementation doesn't crash is a decent start.

Tests/Tests/STPBankAccountParamsTest.m
joeydong-stripe: 👍

Stripe.xcodeproj/project.pbxproj
```diff
@@ -374,6 +374,23 @@
 		04F94DD31D22A23F004FC826 /* NSBundle+Stripe_AppName.h in Headers */ = {isa = PBXBuildFile; fileRef = 049A3F971CC76A2400F57DE7 /* NSBundle+Stripe_AppName.h */; };
```
joeydong-stripe: Changes in `project.pbxproj`:

- Added a bunch of test files

- Added a few private header files with project target membership
screen shot 2017-06-21 at 2 57 31 pm

- Moved `STPFile.m` into the `Models` group next to `STPFile.h`

Tests/Tests/STPBankAccountParamsTest.m
bg-stripe: sounds good!

Stripe/PublicHeaders/STPFile.h
```diff
@@ -48,7 +48,7 @@ typedef NS_ENUM(NSInteger, STPFilePurpose) {
 /**
  * Returns the string value for a purpose.
  */
-+ (NSString *)stringFromPurpose:(STPFilePurpose)purpose;
++ (nullable NSString *)stringFromPurpose:(STPFilePurpose)purpose;
```
joeydong-stripe: This is a public API change 😮

Stripe/STPCard.m
```diff
@@ -84,19 +87,34 @@ + (NSString *)stringFromBrand:(STPCardBrand)brand {
...
 }
 
+#pragma mark - STPCardFundingType
+
++ (NSDictionary <NSString *, NSNumber *> *)stringToFundingMapping {
+    return @{
+             @"credit": @(STPCardFundingTypeCredit),
+             @"debit": @(STPCardFundingTypeDebit),
+             @"prepaid": @(STPCardFundingTypePrepaid),
+             @"other": @(STPCardFundingTypeOther),
```
joeydong-stripe: I might need to do some cleanup here to align with what our API does:

screen shot 2017-06-21 at 3 03 11 pm

https://stripe.com/docs/api#card_object-funding

Stripe/STPSource+Private.h
```diff
@@ -8,11 +8,22 @@
...
++ (nullable NSString *)stringFromType:(STPSourceType)type;
+
++ (STPSourceFlow)flowFromString:(NSString *)string;
++ (nullable NSString *)stringFromFlow:(STPSourceFlow)flow;
 
++ (STPSourceStatus)statusFromString:(NSString *)string;
++ (nullable NSString *)stringFromStatus:(STPSourceStatus)status;
+
++ (STPSourceUsage)usageFromString:(NSString *)string;
++ (nullable NSString *)stringFromUsage:(STPSourceUsage)usage;
```
joeydong-stripe: Changed a bunch of these `stringFrom*` to be `nullable` here to account for returning `nil` when the enum value is `*Unknown`

Stripe/PublicHeaders/STPFile.h
bdorfman-stripe: If it's already possible for it to be nil and it's just not annotated correctly, I think it's ok to change this without it being a major bump or api change (we should treat it as a bug fix and not an api change).

Alternatively, we could just change this to return the empty string or "unknown" or something. 

Or more alternatively, we could just leave it as nonnull since theoretically it can't be if you actually pass in an `STPFilePurpose`. It's kinda only nullable in ObjC where you can pass in ints (and most ObjC won't care about nullability that much) whereas in Swift you can't really get a wrong value?

Stripe/STPBankAccount.m
```diff
@@ -24,6 +27,33 @@ @implementation STPBankAccount
 
 @synthesize routingNumber, country, currency, accountHolderName, accountHolderType;
 
+#pragma mark - STPBankAccountStatus
+
++ (NSDictionary<NSString *, NSNumber *> *)stringToStatusMapping {
+    return @{
+             @"new": @(STPBankAccountStatusNew),
```
bdorfman-stripe: I generally like this dictionary style of doing string <-> enum mappings and we use it a lot of places, but it does somewhat worry me as this method does not create a compiler error if we add a new enum value (whereas doing a switch statement without a default case _does_). 

Switch statements also avoid the "unknown" problem. An exhaustive switch case where we return a string from each possible enum should compile fine I believe? Whereas in this method the compiler will complain that theoretically the map lookup will fail and you could get a nil string returned.

Any feelings on this @bg-stripe ?

Stripe/STPBankAccount.m
```diff
@@ -24,6 +27,33 @@ @implementation STPBankAccount
...
+}
+
++ (STPBankAccountStatus)statusFromString:(NSString *)string {
+    NSNumber *statusNumber = [self stringToStatusMapping][string];
+
+    if (statusNumber) {
+        return (STPBankAccountStatus)[statusNumber integerValue];
+    }
+
+    return STPBankAccountStatusNew;
```
bdorfman-stripe: Why return new for unknown values?

Stripe/PublicHeaders/STPFile.h
joeydong-stripe: > If it's already possible for it to be nil and it's just not annotated correctly, I think it's ok to change this without it being a major bump or api change (we should treat it as a bug fix and not an api change).

👍 

> Alternatively, we could just change this to return the empty string or "unknown" or something.

Oh yeah, I dug into this option a little bit. I think returning "unknown" will change our form encoding behavior because the `STPFormEncoder` only encodes non-nil values. Not in the `STPFile` case but definitely in the `STPSourceParams`.

> Or more alternatively, we could just leave it as nonnull since theoretically it can't be if you actually pass in an STPFilePurpose. It's kinda only nullable in ObjC where you can pass in ints (and most ObjC won't care about nullability that much) whereas in Swift you can't really get a wrong value?

You can pass in a `STPFilePurposeUnknown` and get `nil` back.

Stripe/STPBankAccount.m
joeydong-stripe: > I generally like this dictionary style of doing string <-> enum mappings and we use it a lot of places, but it does somewhat worry me as this method does not create a compiler error if we add a new enum value (whereas doing a switch statement without a default case does).

Yeah I agree there's room for improvement here but at least a test case will fail now if you incompletely add/remove an enum case.

Stripe/STPBankAccount.m
joeydong-stripe: We don't actually have an unknown case for `STPBankAccountStatus`. I think the current pattern is to add an "unknown" case only if we think the API is going to add more cases in the near future.

Stripe/STPCard.m
joeydong-stripe: Cleaned up in https://github.com/stripe/stripe-ios/pull/711/commits/2093f279c4b9f47ce1b84ce86d585362bf7ba0cd

Stripe/PublicHeaders/STPFile.h
bdorfman-stripe: Ah I didn't realize about the unknown thing. So yeah I think this should be considered a nullabilty bugfix and not a breaking api change (although maybe still mention it in the migrating doc).

Stripe/STPBankAccount.m
bg-stripe: Yeah, it is kind of unfortunate that we lose the compiler validating the incomplete switch. Since joey's written thorough tests that include switch statements, I think we'll get the same validation now via tests :)

I think the unknown thing would be an issue whether or not we were using switch statements, since it's a question of what value we return in `...fromString` when the string is invalid. I think the way we're handling this is fine though, just picking an arbitrary fallback unless we really expect to see unknown types from the API.

Tests/Tests/STPBankAccountTest.m
```diff
@@ -10,17 +10,76 @@
...
+- (void)tearDown {
+    // Put teardown code here. This method is called after the invocation of each test method in the class.
+    [super tearDown];
+}
+
+#pragma mark - STPBankAccountStatus Tests
+
+- (void)testStatusFromString {
+    XCTAssertEqual([STPBankAccount statusFromString:@"new"], STPBankAccountStatusNew);
+    XCTAssertEqual([STPBankAccount statusFromString:@"NEW"], STPBankAccountStatusNew);
```
joeydong-stripe: yeah at first I was hesitant on adding flexibility to the `*fromString` methods but if there is a casing mismatch its probably non-intentional and we should make a best effort return value

PR #712, opened 2017-06-22
Fix `ci_script/run_analyzer.sh`

joeydong-stripe:
## Summary

Fixed the script to start exiting with a non-zero status code when static analyzer warnings are found. This should then trigger a build failure.

Added a `STPNonLocalizedString` (open to naming suggestions) helper to explicitly ignore non-localized string warnings in appropriate cases.

## Motivation

Make sure we deal with static analyzer warnings as soon as they appear.

## Testing

Verified `STPPaymentCardTextField` continues to function as expected and that `STPNonLocalizedString` correctly returns the input parameter unchanged.

r? @bdorfman-stripe

### DIFF
--- a/Stripe/STPLocalizationUtils.h
+++ b/Stripe/STPLocalizationUtils.h
@@ -21,3 +21,9 @@
 static inline NSString * _Nonnull STPLocalizedString(NSString* _Nonnull key, NSString * _Nullable __unused comment) {
     return [STPLocalizationUtils localizedStripeStringForKey:key];
 }
+
+/// Use to explicitly ignore static analyzer warning: "User-facing text should use localized string macro"
+__attribute__((annotate("returns_localized_nsstring")))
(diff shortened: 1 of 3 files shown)


Stripe/STPLocalizationUtils.h
```diff
@@ -21,3 +21,9 @@
 static inline NSString * _Nonnull STPLocalizedString(NSString* _Nonnull key, NSString * _Nullable __unused comment) {
     return [STPLocalizationUtils localizedStripeStringForKey:key];
 }
+
+/// Use to explicitly ignore static analyzer warning: "User-facing text should use localized string macro"
+__attribute__((annotate("returns_localized_nsstring")))
+static inline NSString * _Nonnull STPNonLocalizedString(NSString * _Nonnull string) {
```
bdorfman-stripe: 👍 

This is good also for other developers to know that strings are intentionally not translated (and not that they were accidentally missed)

PR #714, opened 2017-06-26
Improve test coverage for public models

joeydong-stripe:
## Summary

Improves test coverage for our public models and improves some implementations of `STPFormEncodable`

## Motivation

Better test coverage should help us iterate with more confidence.

## Testing

Verified unit tests still pass.

r? @bdorfman-stripe 
cc @bg-stripe


Tests/Tests/STPBankAccountTest.m
```diff
@@ -100,53 +117,82 @@ - (void)testBankAccountEquals {
...
+
+    XCTAssertEqualObjects(bankAccount.bankAccountId, @"ba_1AXvnKEOD54MuFwSotKc6xq0");
+    XCTAssertEqualObjects(bankAccount.accountHolderName, @"Jane Austen");
+    XCTAssertEqual(bankAccount.accountHolderType, STPBankAccountHolderTypeIndividual);
+    XCTAssertEqualObjects(bankAccount.bankName, @"STRIPE TEST BANK");
+    XCTAssertEqualObjects(bankAccount.country, @"US");
+    XCTAssertEqualObjects(bankAccount.currency, @"usd");
+    XCTAssertEqualObjects(bankAccount.fingerprint, @"C5fW7AwE3of8bHvV");
+    XCTAssertEqualObjects(bankAccount.last4, @"6789");
+    XCTAssertNil(bankAccount.routingNumber);
```
joeydong-stripe: This one's interesting. Should we map the routingNumber to the bankAccount?

joeydong-stripe:
Sorry for the commit churn. Also, I can break this PR up if this is too much at once.

Stripe/STPBankAccountParams+Private.h
```diff
@@ -15,6 +15,7 @@ NS_ASSUME_NONNULL_BEGIN
 + (STPBankAccountHolderType)accountHolderTypeFromString:(NSString *)string;
 + (NSString *)stringFromAccountHolderType:(STPBankAccountHolderType)accountHolderType;
 
+- (instancetype)init;
```
bdorfman-stripe: Does this need to be here? Don't all NSObject's inherit init anyway?

Tests/Tests/STPBankAccountTest.m
```diff
@@ -8,21 +8,20 @@
...
+@interface STPBankAccountTest : XCTestCase
 
 @end
 
 @implementation STPBankAccountTest
 
 - (void)setUp {
     [super setUp];
-    _bankAccount = [[STPBankAccount alloc] init];
+    // Put setup code here. This method is called before the invocation of each test method in the class.
```
bdorfman-stripe: If there is no setup needed, you should just delete the method.

Tests/Tests/STPBankAccountTest.m
```diff
@@ -100,53 +117,82 @@ - (void)testBankAccountEquals {
...
-- (void)testDescriptionWorks {
+- (void)testDescription {
     STPBankAccount *bankAccount = [STPBankAccount decodedObjectFromAPIResponse:[self completeAttributeDictionary]];
-    bankAccount.routingNumber = @"123456789";
     XCTAssert(bankAccount.description);
 }
 
 #pragma mark - STPAPIResponseDecodable Tests
 
 - (NSDictionary *)completeAttributeDictionary {
```
bdorfman-stripe: Test json like this should probably be moved to STPFixtures? And/or an actual json file.

Tests/Tests/STPBankAccountTest.m
bdorfman-stripe: Yeah it seems like an oversight that this isn't parsed currently? Not sure if there was some intentional reason we don't parse this (although I assume not because we have a property for it). Also `routingNumber` is declared as nonnull, which is just entirely wrong. Should change that annotation and/or parse it and add it to required keys I guess.

Somewhat relatedly: It's suuuuper weird to me that STPBankAccount and STPCard are subclasses of their own Params objects. Glad we didn't do that for Sources.

What do you think @bg-stripe?

Tests/Tests/STPCardTest.m
```diff
@@ -13,15 +13,13 @@
...
-@property (nonatomic) STPCard *card;
-
 @end
 
 @implementation STPCardTest
 
 - (void)setUp {
     [super setUp];
-    _card = [[STPCard alloc] init];
+    // Put setup code here. This method is called before the invocation of each test method in the class.
```
bdorfman-stripe: Same as other setup comment.

Tests/Tests/STPCardTest.m
```diff
@@ -151,38 +149,86 @@ - (void)testStringFromFunding {
...
+- (void)testInit {
+    STPCard *card = [[STPCard alloc] init];
+    XCTAssertEqual(card.brand, STPCardBrandUnknown);
+    XCTAssertEqual(card.funding, STPCardFundingTypeOther);
+}
+
 - (void)testLast4ReturnsCardNumberLast4WhenNotSet {
-    self.card.number = @"4242424242424242";
-    XCTAssertEqualObjects(self.card.last4, @"4242", @"last4 correctly returns the last 4 digits of the card number");
+    STPCard *card = [[STPCard alloc] init];
```
bdorfman-stripe: can just do `[STPCard new]` instead of alloc/init? Or is that a style guide holy war? :)

Tests/Tests/STPCardTest.m
```diff
@@ -191,66 +237,118 @@ - (void)testCardEquals {
...
-             @"address_line1": @"221A Baker Street",
-             @"address_city": @"New York",
-             @"address_state": @"NY",
-             @"address_zip": @"12345",
-             @"address_country": @"USA",
-             @"last4": @"1234",
-             @"dynamic_last4": @"5678",
-             @"brand": @"MasterCard",
-             @"country": @"Japan",
+             @"id": @"card_1AVRojEOD54MuFwSxr93QJSx",
```
bdorfman-stripe: Where do these ids actually come from? I have different ones on my screen. Are we sure these aren't "real" tokens that are populated from the logged in Stripe account or something?

Also again I think all fake json stuff should be through STPFixtures now.

Tests/Tests/STPBankAccountTest.m
bg-stripe: Yeah it is super weird – I _think_ the history there is that previously `STPCard` was used as both the params object and the returned object, and we wanted to introduce `STPCardParams` with a deprecation, rather than a breaking change.

I think the best fix here would be to make `STPBankAccount` not a subclass of `STPBankAccountParams` and do the same with `CardParams` – maybe we should save that for a separate PR though.

Tests/Tests/STPCardTest.m
bg-stripe: :P +1 for `new` in the style guide

Tests/Tests/STPCardTest.m
bg-stripe: (I usually use dummy IDs like `card_123` in fixtures)

Tests/Tests/STPBankAccountTest.m
joeydong-stripe: Cool, I'll at least map the `routingNumber`. I think overall `STPBankAccount` has a few nullability bugs that we can fix (cause public api changes) but I'm guessing the object isn't used very often so we don't hear much about it.

Are we ever using the fact that `STPBankAccount` is a subclass of `STPBankAccountParams` to allow people to retrieve/modify/resubmit a bank account easily? Maybe that's too magical.

Stripe/STPBankAccountParams+Private.h
joeydong-stripe: I think I was getting at making sure we have an explicit designated initializer for everything and seeing if that provides any compiler safety but didn't finish it. I'll remove this for now and restart the investigation later.

Tests/Tests/STPCardTest.m
joeydong-stripe: I added this as one of our style guide discussion points 😉

Tests/Tests/STPCardTest.m
joeydong-stripe: haha yeah looks like it's one of my dummy test cards. i'll figure something out.

Tests/Tests/STPCardTest.m
```diff
@@ -151,106 +142,176 @@ - (void)testStringFromFunding {
...
+    XCTAssertEqualObjects(card.currency, @"usd");
+    XCTAssertEqualObjects(card.dynamicLast4, @"5678");
+    XCTAssertEqual(card.expMonth, (NSUInteger)5);
+    XCTAssertEqual(card.expYear, (NSUInteger)2017);
+    XCTAssertEqual(card.funding, STPCardFundingTypeCredit);
+    XCTAssertEqualObjects(card.last4, @"4242");
+    XCTAssertEqualObjects(card.name, @"Jane Austen");
+
+    XCTAssertNotEqual(card.allResponseFields, response);
+    XCTAssertEqualObjects(card.allResponseFields, [response stp_dictionaryByRemovingNullsValidatingRequiredFields:@[]]);
```
joeydong-stripe: I'll come back to fixing this `stp_dictionaryByRemovingNullsValidatingRequiredFields` usage in another PR. The category is missing test coverage and we should probably deeply remove `[NSNull null]`s.

joeydong-stripe:
wdyt @bdorfman-stripe ?

Tests/Tests/STPBankAccountTest.m
joeydong-stripe: 👍

Tests/Tests/STPBankAccountTest.m
joeydong-stripe: 👍

Tests/Tests/STPCardTest.m
joeydong-stripe: 👍

joeydong-stripe:
Just moved all of the private header definitions I added that were specific to tests in 0a89c95 because it looks like that's where our style is headed.

cc @bg-stripe

Issue #717, opened 2017-06-28
3D Verification dismissing issue


gravera:
## Summary

If a redirection flow is started from an already presented view controller, pressing the 'Done' button on the SFSafariViewController causes the presented controller to dismiss also. The reason is pressing the 'Done' dismisses the SFSafariViewController by default and then in the delegate you are dismissing it again through -handleRedirectCompletionWithError method, which calls:
STPRedirectContext line 147
[self.safariVC.presentingViewController dismissViewControllerAnimated:YES
                                                                   completion:^{
                                                                       stpDispatchToMainThreadIfNecessary(^{
                                                                           [self handleRedirectCompletionWithError:nil];
                                                                       });
                                                                   }];

and since the SFSafariViewController is auto dismissed, this code makes the presentingViewController to dismiss also.

## Code to reproduce

Initiate redirection flow from already presented view controller.

## iOS version

9, 10

## Installation method

Cocoa pods

## SDK version

10.2.0

## Other information

A temporary fix for us would be to dismiss only self.safariVC in STPRedirectContext line 147, rather than self.safariVC.presentingViewController.

bdorfman-stripe:
Thanks for the report! I'll look into it and get a fix for this out as soon as possible.

bdorfman-stripe:
This should be fixed in version 11.0.0, released just now! Thanks again for the bug report.

gravera:
Awesome. Thank you!

PR #718, opened 2017-06-28
Fix double-view dismissal

bdorfman-stripe:
Hitting the done button in SFSafariVC makes it dismiss itself, so we don't want to manually dismiss it in that case because that will cause two dismissals to be fired and that can cascade to dismiss other view controllers. Instead, only dismiss in the other completion cases.

Also remove an uncessary completion block in the dismiss logic.

Fixes https://github.com/stripe/stripe-ios/issues/717

### DIFF
--- a/Stripe/STPRedirectContext.m
+++ b/Stripe/STPRedirectContext.m
@@ -133,7 +133,7 @@ NS_ASSUME_NONNULL_BEGIN
     [self unsubscribeFromNotifications];
 
     if (shouldDismissViewController) {
-        [self dimissPresentedViewController];
+        [self dismissPresentedViewController];
     }
 
     self.completion(self.source.stripeID, self.source.clientSecret, error);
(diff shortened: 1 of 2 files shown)


bdorfman-stripe:
Not sure if I should actually be merging this in to the 11.0 branch? That might be easiest.

Stripe/STPRedirectContext.m
```diff
@@ -138,18 +149,21 @@ - (void)subscribeToUrlAndForegroundNotifications {
...
 
+- (void)dimissPresentedViewController {
     if (self.safariVC) {
         [self.safariVC.presentingViewController dismissViewControllerAnimated:YES
-                                                                   completion:^{
-                                                                       stpDispatchToMainThreadIfNecessary(^{
-                                                                           [self handleRedirectCompletionWithError:nil];
-                                                                       });
-                                                                   }];
+                                                                   completion:nil];
```
joeydong-stripe: so we don't need the `completion` anymore because we're *sure* that one of the delegates will be called?

maybe leave a comment so we don't regress on this.

bdorfman-stripe: We never needed it afaict, it was just an oversight. The only ways this can get called is: 
1. Via through the `handleRedirectCompletionWithError` which calls completion after.
2. You are getting dealloc'd (in which case I'm not sure calling completion makes sense?)
3. You manually called `cancel` on the redirect context (again, in which case I don't think it makes sense to fire the completion)

bg-stripe:
(fixed up tests and a typo in 34a1d95)

PR #719, opened 2017-06-29
Remove Remember Me

bdorfman-stripe:
Not sure if you want to re-cut the 11.0.0 branch from master with this and the other PR, or merge both into that branch and then merge into master later.


bg-stripe:
(lint was complaining about an unused image – removed it in 4de9256)

PR #723, opened 2017-07-05
Improve `STPAPIRequest` class

joeydong-stripe:
## Summary & Motivation & Testing

I was in the process of adding support for `DELETE` requests in `STPAPIRequest` and decided to take a few steps back and improve the surrounding implementation first.

In particular, I started by [ensuring 100% test coverage for `[STPAPIRequest parseResponse:...]`](https://github.com/stripe/stripe-ios/commit/15a4bab5e69d380e1e4ef661795fb5cc8bd94c1f) before [untangling it](https://github.com/stripe/stripe-ios/commit/84b9794b31148f92859a7917d8dc5fcacb1863ce).

Then I [added support for making `DELETE` requests](https://github.com/stripe/stripe-ios/commit/44b3250e71f9689dc85e0b352c626691197508df) along with [corresponding test coverage](https://github.com/stripe/stripe-ios/commit/5c026855c9b27cbc2fde32ba06b12f9064071fb2). These tests turned out pretty complicated because of all the side effects. What do you think?

r? @bg-stripe 
cc @bdorfman-stripe

### DIFF
--- a/Stripe/STPAPIRequest.m
+++ b/Stripe/STPAPIRequest.m
@@ -9,27 +9,22 @@
 #import "STPAPIRequest.h"
 
 #import "NSMutableURLRequest+Stripe.h"
-#import "STPInternalAPIResponseDecodable.h"
-#import "STPAPIClient+Private.h"
 #import "STPAPIClient.h"
-#import "STPCard+Private.h"
+#import "STPAPIClient+Private.h"
 #import "STPDispatchFunctions.h"
-#import "STPFormEncoder.h"
-#import "STPSource+Private.h"
+#import "STPInternalAPIResponseDecodable.h"
 #import "StripeError.h"
 
 @implementation STPAPIRequest
+
+#pragma mark - POST
 
 + (NSURLSessionDataTask *)postWithAPIClient:(STPAPIClient *)apiClient
                                    endpoint:(NSString *)endpoint
                                  parameters:(NSDictionary *)parameters
                                deserializer:(id<STPAPIResponseDecodable>)deserializer
                                  completion:(STPAPIResponseBlock)completion {
-    return [self postWithAPIClient:apiClient
-                          endpoint:endpoint
-                        parameters:parameters
-                     deserializers:@[deserializer]
-                        completion:completion];
+    return [self postWithAPIClient:apiClient endpoint:endpoint parameters:parameters deserializers:@[deserializer] completion:completion];
 }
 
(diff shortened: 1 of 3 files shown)


Stripe/STPAPIRequest.m
```diff
@@ -9,120 +9,183 @@
...
+#import "STPInternalAPIResponseDecodable.h"
 #import "StripeError.h"
 
 @implementation STPAPIRequest
 
+static NSString * const HTTPMethodPOST = @"POST";
+static NSString * const HTTPMethodGET = @"GET";
+static NSString * const HTTPMethodDELETE = @"DELETE";
+
+static NSString * const JSONKeyObject = @"object";
```
joeydong-stripe: Not sure about these constant names. Also I'm assuming we won't have any name spacing issues because it's in the implementation?

Stripe/STPAPIRequest.m
```diff
@@ -9,120 +9,183 @@
...
-    request.HTTPBody = [query dataUsingEncoding:NSUTF8StringEncoding];
-
-    NSURLSessionDataTask *task = [apiClient.urlSession dataTaskWithRequest:request completionHandler:^(NSData * _Nullable body, NSURLResponse * _Nullable response, NSError * _Nullable error) {
-        [[self class] parseResponse:response
-                               body:body
-                              error:error
-                      deserializers:deserializers
-                         completion:completion];
+    request.HTTPMethod = HTTPMethodPOST;
+    [request stp_setFormPayload:parameters];
```
joeydong-stripe: this line is a bit of a change in behavior but I think it's for the better

Stripe/STPAPIRequest.m
bg-stripe: This seems good to me! And yeah, I don't think we have to worry about namespacing these

Stripe/STPAPIRequest.m
bg-stripe: ah yeah, we should've been using `stp_setFormPayload` before

joeydong-stripe:
@bg-stripe I realized this morning my test case was broken (it was exercising the "no deserializers" path) and fixed that in https://github.com/stripe/stripe-ios/pull/723/commits/4d67f4e2a0adf097ec6efb18a74972441aa85d0b. The case is a little unrealistic (both `body` and `error` is defined) but now we have an expected behavior for it.

bg-stripe:
ah, good catch, and lgtm!

PR #725, opened 2017-07-06
Add postal code to payment card text field

bdorfman-stripe:
Adds zip/postal code optionally to payment card text field.
Also improves postal code validation/formatting and improves layout logic of card text field.

Need some new translations for the short "zip" and "postal" strings (we currently only have "zip code" and "postal code") but otherwise good to review I think.


Stripe/STPPaymentCardTextField.m
```diff
@@ -476,24 +615,138 @@ - (void)setText:(NSString *)text inField:(STPCardFieldType)field {
...
             textField = self.cvcField;
             break;
+        case STPCardFieldTypePostalCode:
+            textField = self.postalCodeField;
+            break;
     }
     textField.text = nonNilText;
 }
 
+// TODO: cache these values
```
bdorfman-stripe: Ah good call :)

Stripe/STPPaymentCardTextField.m
```diff
@@ -686,36 +1183,80 @@ - (void)formTextFieldTextDidChange:(STPFormTextField *)formTextField {
...
+        paymentTextFieldDidBeginEditing = YES;
+    }
+    self.fieldEditingTransitionState = STPEditingTransitionStateFalse;
+
+    [self layoutViewsToFocusField:@(textField.tag)
+                         animated:YES
+                       completion:nil];
+
+    if (paymentTextFieldDidBeginEditing) {
+        NSLog(@"Did begin editing payment field");
```
bdorfman-stripe: Whoops!

bdorfman-stripe:
> There's a little jump when leaving a field for the first time
This is due to some weird internals of UILabel as far as I can tell (I think it's doing some internal view swapping or resizing that it expects won' be animated, but in this instance our animation cascade down to it and they didn't put it inside a `withoutAnimation` block I guess). I'll poke around and try to see if I can do something here.

> There's a pretty noticeable delay between entering the last character in a field and moving to the next field. The delay goes away after the first time – maybe this has to do with string sizing?
I think this is also the same weird internal UILabel thing as above I think.

> When you tap into a field that's already filled, typing a number does nothing (doesn't advance to the next field)
I think this is correct? The previous version worked the same iirc.

> It isn't possible to enter a ZIP+4 in the US postal field
Design specifically requested I not allow ZIP+4 since only the first five is used for billing address purposes anyway.

> We should probably change the keyboard type for the non-US postal field
Good call.

> Not sure if we should also change the keyboard type for the US postal field after 5 digits. Can a US zip really have non-digits? I've just never seen one myself, and haven't been able to find anything on the internet.
US zips are numeric-only (unless you count the separator between the first 5 and +4, which we auto fill in for you anyway).

> We should probably update the Obj-C example app and site docs to make it more apparent that collecting postal is an option
👍

Stripe/STPPaymentCardTextField.m
```diff
@@ -329,12 +430,32 @@ - (BOOL)becomeFirstResponder {
...
-    } else if ([self.viewModel validationStateForField:STPCardFieldTypeExpiration] != STPCardValidationStateValid) {
-        return self.expirationField;
-    } else {
-        return self.cvcField;
+    STPFormTextField *currentSubResponder = self.currentFirstResponderField;
+    if (currentSubResponder) {
+        NSUInteger index = [self.allFields indexOfObject:currentSubResponder];
+        if (index != NSNotFound) {
+            index += 1;
+            if (self.allFields.count > index) {
```
bdorfman-stripe: Ah I'll swap to that!

Stripe/STPPaymentCardTextField.m
```diff
@@ -8,35 +8,74 @@
...
+ as a whole (instead of on a per-subview basis).
+ 
+ The state transitons in the should/did begin/end editing callbacks for all 
+ our subfields. If we get a shouldEnd AND a shouldBegin before getting either's
+ matching didEnd/didBegin, then we are transitioning focus between our subviews
+ (and so we ourselves should not consider us to have begun or ended editing).
+ 
+ But if we get a should and did called on their own without a matching opposite
+ pair (shouldBegin/didBegin or shouldEnd/didEnd) then we are transitioning
+ into/out of our subviews from/to outside of ourselves
```
bdorfman-stripe: I can't tell if you're joking or not :P

Stripe/PublicHeaders/STPPaymentCardTextField.h
```diff
@@ -109,6 +64,11 @@ IB_DESIGNABLE
 @property(nonatomic, copy, nullable) IBInspectable NSString *cvcPlaceholder;
 
 /**
+ *  The placeholder for the postal code field. Defaults to @"ZIP/Postal".
```
bdorfman-stripe: Note to self: Change this to make clear that you get either "ZIP" or "Postal" and not "ZIP/Postal"

Stripe/STPPaymentCardTextField.m
```diff
@@ -421,11 +549,22 @@ - (NSString *)cvc {
     return self.viewModel.cvc;
 }
 
+- (NSString *)postalCode {
+    if ([self postalCodeFieldIsEnabled]) {
+        return self.viewModel.postalCode;
+
```
bdorfman-stripe: Note to self: Extra line break here

bg-stripe:
>> When you tap into a field that's already filled, typing a number does nothing (doesn't advance to the next field)

> I think this is correct? The previous version worked the same iirc.

Ah you're right, just checked the previous version.

bg-stripe:
> US zips are numeric-only (unless you count the separator between the first 5 and +4, which we auto fill in for you anyway).

Ah, whoops – I was just confused by my first reading of the US zip validation

bdorfman-stripe:
Also re: ZIP vs ZIP+4 - After talking to design, I added an "intended usage" (billing or shipping) to the postal code formatter. For US zip codes this changes whether it allows ZIP or also ZIP+4 (the validator allows either always). At some point we can use this swap out the ShippingVC implementation to do automatic ZIP+4 formatting. I didn't touch the AddCardVC or ShippingVC stuff at all in this PR, although they could both benefit from improved automatic zip formatting.

bg-stripe:
> Also re: ZIP vs ZIP+4 - After talking to design, I added an "intended usage" (billing or shipping) to the postal code formatter. For US zip codes this changes whether it allows ZIP or also ZIP+4 (the validator allows either always). At some point we can use this swap out the ShippingVC implementation to do automatic ZIP+4 formatting. I didn't touch the AddCardVC or ShippingVC stuff at all in this PR, although they could both benefit from improved automatic zip formatting.

Ah cool, that makes sense!

bdorfman-stripe:
@bg-stripe I think I'm ready for re-review now.

Stripe/STPPaymentCardTextField.m
bg-stripe: not joking! this would be really confusing without it

Stripe/PublicHeaders/STPPaymentCardTextField.h
```diff
@@ -144,6 +105,72 @@ IB_DESIGNABLE
...
+/**
+ *  Controls if a postal code entry field is displayed, and if so how its text
+ *  should be formatted and what input is valid.
+ *
+ *  Default is nil (no postal code entry is displayed).
+ *  Note that some non-nil values will still result in no postal code being
+ *  displayed (invalid country codes, or countries which do not commonly use
+ *  postal codes).
+ */
+@property(nonatomic, copy, nullable) NSString *countryCodeForPostalCodeFormattingAndValidation;
```
bg-stripe: (discussed in chat) can we update the API to be something like `postalCodeCollectionEnabled` + `countryCode` + use locale if `countryCode` is nil?

Stripe/STPPaymentCardTextField.m
```diff
@@ -1050,8 +1078,7 @@ - (void)layoutViewsToFocusField:(NSNumber *)focusedField
     self.focusedTextFieldForLayout = focusedField;
 
     void (^animations)() = ^void() {
-        [self setNeedsLayout];
-        [self layoutIfNeeded];
```
bg-stripe: ah, was this causing the jump?

Stripe/STPPaymentCardTextField.m
```diff
@@ -1186,29 +1192,74 @@ - (void)formTextFieldTextDidChange:(STPFormTextField *)formTextField {
...
+typedef NS_ENUM(NSInteger, STPFieldEditingTransitionCallSite) {
+    STPFieldEditingTransitionCallSiteShouldBegin,
+    STPFieldEditingTransitionCallSiteShouldEnd,
+    STPFieldEditingTransitionCallSiteDidBegin,
+    STPFieldEditingTransitionCallSiteDidEnd,
+};
+
+// Explanation of the logic here is with the definition of these properties
+// at the top of this file
+- (BOOL)getAndUpdateSubviewEditingTransitionStateFromCall:(STPFieldEditingTransitionCallSite)sendingMethod {
```
bg-stripe: nice, this is much cleaner :)

Example/Stripe iOS Example (Custom)/CardExampleViewController.m
```diff
@@ -37,17 +45,19 @@ - (void)viewDidLoad {
...
     UIActivityIndicatorView *activityIndicator = [[UIActivityIndicatorView alloc] initWithActivityIndicatorStyle:UIActivityIndicatorViewStyleGray];
     activityIndicator.hidesWhenStopped = YES;
     self.activityIndicator = activityIndicator;
-    [self.view addSubview:activityIndicator];
+    [scrollView addSubview:activityIndicator];
 }
 
 - (void)viewDidLayoutSubviews {
     [super viewDidLayoutSubviews];
+    self.scrollView.bounds = self.view.bounds;
```
bg-stripe: should be `self.scrollView.frame = self.view.bounds;`

Stripe/PublicHeaders/STPPaymentCardTextField.h
bdorfman-stripe: Ah right, I'll do that!

Stripe/STPPostalCodeValidator.m
```diff
@@ -7,40 +7,171 @@
 //
 
 #import "STPPostalCodeValidator.h"
+
 #import "STPCardValidator.h"
 #import "STPPhoneNumberValidator.h"
+#import "NSCharacterSet+Stripe.h"
+#import "NSCharacterSet+Stripe.h"
```
joeydong-stripe: double import 😉

Stripe/STPPostalCodeValidator.m
```diff
@@ -7,40 +7,171 @@
...
+        [mutableZip insertString:@"-" atIndex:5];
+        formattedString = mutableZip.copy;
+    }
+
+    return formattedString;
+}
+
+
++ (BOOL)postalCodeIsRequiredForCountryCode:(NSString *)countryCode {
+    if (countryCode == nil) {
```
joeydong-stripe: we should check the length here:

```objc
if (countryCode.length == 0) {
  return NO;
}
```

Stripe/STPPostalCodeValidator.m
```diff
@@ -7,40 +7,171 @@
...
+                return STPCardValidationStateIncomplete;
+            }
         }
-        case STPCountryPostalCodeTypeAlphanumeric:
-            return string.length > 0;
-        case STPCountryPostalCodeTypeNotRequired:
-            return YES;
+    }
+    else {
+        return STPCardValidationStateValid;
```
joeydong-stripe: I think `validationStateForPostalCode: countryCode:` and `validationStateForUSPostalCode:` readability could benefit from some early returns but this is totally valid code too.

Stripe/STPPostalCodeValidator.m
```diff
@@ -7,40 +7,171 @@
...
+                break;
+            }
+            else {
+                count += range.length;
+                lastPosition = NSMaxRange(range);
+            }
+        }
+    }
+
+    return count;
```
joeydong-stripe: this function is a little hairy but I'm guessing it's the right way to handle the various string encodings?

Stripe/STPPostalCodeValidator.m
```diff
@@ -7,40 +7,171 @@
...
+
+    NSString *formattedString = [[STPCardValidator sanitizedNumericStringForString:zipCode] stp_safeSubstringToIndex:maxLength];
+
+    if (formattedString.length > 5
+        || (formattedString.length == 5
+            && [[zipCode substringFromIndex:(zipCode.length - 1)] isEqualToString:@"-"])) {
+        NSMutableString *mutableZip = formattedString.mutableCopy;
+        [mutableZip insertString:@"-" atIndex:5];
+        formattedString = mutableZip.copy;
+    }
```
joeydong-stripe: I think a quick comment here would make this condition less confusing:

```
// Insert `-` separator if zip has more than 5 digits or original 5 digit zip ended with a `-`
```

or shuffle some code around and do an earlier if/else branch on `maxLength`

Example/Stripe iOS Example (Custom)/CardExampleViewController.m
```diff
@@ -15,15 +15,23 @@
...
 @implementation CardExampleViewController
 
 - (void)viewDidLoad {
     [super viewDidLoad];
+    UIScrollView *scrollView = [[UIScrollView alloc] initWithFrame:self.view.bounds];
+    scrollView.delegate = self;
+    scrollView.alwaysBounceVertical = YES;
+    scrollView.backgroundColor = [UIColor whiteColor];
+    [self.view addSubview:scrollView];
+    self.scrollView = scrollView;
```
bdorfman-stripe: Ah good idea.

Stripe/PublicHeaders/STPCardValidationState.h
```diff
@@ -9,10 +9,17 @@
...
 /**
- *  These fields indicate whether a card field represents a valid value, invalid value, or incomplete value.
+ These fields indicate whether a card field represents a valid value, invalid value, or incomplete value.
+
+ - STPCardValidationStateValid: The field's contents are valid. For example, a valid, 16-digit card number.
+    Note that valid values may not be complete. For example: a US Zip code can be 5 or 9 digits. A 5-digit code
+    is Valid, but more text could be entered to transition to incomplete again. American Express CVC codes
+    can be 3 or 4 digits and both will be treated as Valid.
+ - STPCardValidationStateInvalid: The field's contents are invalid. For example, an expiration date of "13/42".
+ - STPCardValidationStateIncomplete: The field's contents are not yet valid, but could be by typing additional characters. For example, a CVC of "1".
```
bdorfman-stripe: Weird, I just used the auto generated template Xcode made for me? I'll poke around and see if I can fix.

Stripe/STPPaymentCardTextField.m
```diff
@@ -476,24 +633,155 @@ - (void)setText:(NSString *)text inField:(STPCardFieldType)field {
...
+    }
+    else {
+        return MAX([self widthForCardNumber:self.viewModel.cardNumber],
+                   [self widthForCardNumber:self.viewModel.defaultPlaceholder]);
+    }
+}
+
+- (CGFloat)numberFieldCompressedWidth {
+    // Current longest possible longest pan fragment is 5 characters
+    return [self widthForText:@"88888"];
```
joeydong-stripe: I think this non-localized string will trigger an analyze warning once you update this branch to `master`.

Stripe/STPPaymentCardTextField.m
```diff
@@ -250,6 +316,60 @@ - (void)setCvcPlaceholder:(NSString * __nullable)cvcPlaceholder {
...
+- (void)setPostalCodePlaceholder:(NSString *)postalCodePlaceholder {
+    _postalCodePlaceholder = postalCodePlaceholder.copy;
+    [self updatePostalFieldPlaceholder];
+}
+
+- (void)setPostalCodeEntryEnabled:(BOOL)postalCodeEntryEnabled {
+    _postalCodeEntryEnabled = postalCodeEntryEnabled;
+    if (postalCodeEntryEnabled
+        && !self.countryCode) {
+        self.countryCode = [[NSLocale autoupdatingCurrentLocale] objectForKey:NSLocaleCountryCode];
```
joeydong-stripe: not sure where we draw the line here but maybe this can be a fallback value for the `self.viewModel.postalCodeCountryCode` getter so we don't need to remember to set it:

```objc
//  STPPaymentCardTextFieldViewModel.m

- (NSString *)postalCodeCountryCode {
  if (_postalCodeCountryCode.length) == 0 {
    _postalCodeCountryCode = [[NSLocale currentLocale] objectForKey:NSLocaleCountryCode];
  }

  return _postalCodeCountryCode;
}
```

Stripe/STPPaymentCardTextField.m
```diff
@@ -250,6 +316,60 @@ - (void)setCvcPlaceholder:(NSString * __nullable)cvcPlaceholder {
...
+        self.postalCodeField.placeholder = _postalCodePlaceholder;
+    }
+}
+
+- (NSString *)defaultPostalFieldPlaceholderForCountryCode:(NSString *)countryCode {
+    if ([countryCode.uppercaseString isEqualToString:@"US"]) {
+        return NSLocalizedString(@"ZIP", @"Short string for zip code (United States only)");
+    }
+    else {
+        return NSLocalizedString(@"Postal", @"Short string for postal code (text used in non-US countries)");
```
joeydong-stripe: I guess we should switch these to `STPLocalizedString` calls?

Stripe/STPPaymentCardTextField.m
joeydong-stripe: hmm I guess you're doing this to call the setter code 🤔

joeydong-stripe:
it's a little weird that "MM/YY" and "CVC" nudge over as soon as you start typing but maybe that's not a straightforward fix:

bdorfman-stripe:
Ah yeah, I think that is because of the differences between leaving enough space for the maximum text you could enter and wrapping to the actual text currently being shown and when we should switch from one to the other and for which fields. I tried to strike the right balance to minimize things randomly shifting, but it's complicated to get something that actually looks correct positioned and sized and spaced all the time given that the fields aren't actually consistent widths or character counts. It also changes depending on how wide your total view is and what's currently being shown. I can try to mess around with it more, or maybe you have suggestions on how to handle this better?

joeydong-stripe:
Given the fact that we need to support variable width because of various device screen widths, I think what you have here is good because it selectively hides fields entirely instead of clipping them. This nudging is probably just a minor bug somewhere in the layout code.

joeydong-stripe:
maybe you can break out `fieldsView` into a `STPPaymentCardTextFieldFieldsView` class and move a bunch of the layout code there

bdorfman-stripe:
It's not a "bug" in the layout code per se, it is because when you are not editing it spaces things out to what is currently being rendered if they are empty or have valid data, whereas when you start editing it (or they are incomplete) it resizes the fields to hold what it thinks the max possible width would be (otherwise they would resize and move around as you were entering text). It's possible this isn't the best way to handle it or there are improvements, but this behavior was an intentional decision by me.

Stripe/STPPostalCodeValidator.m
bdorfman-stripe: Nonnil codes of any length will get caught in the else branch here and so don't necessarily need special logic like this. While i use use `.length == 0` for string checking, I thought it more clear to call it out like this since it's specifically checking that nil values can't be passed to `containsObject:`

Stripe/STPPaymentCardTextField.m
bdorfman-stripe: I'm doing a setter instead of a callback in the getter because I don't want the country code to have any chance of changing between calls to the getter. This way if it is nil when you enable postal code entry, it will grab the current locale but cache it locally so that it can't change out from under us while the user is interacting with the view.

Stripe/STPPaymentCardTextField.m
bdorfman-stripe: Good catch

Stripe/STPPaymentCardTextField.m
```diff
@@ -443,9 +600,9 @@ - (void)setCardParams:(STPCardParams *)cardParams {
         [self setText:@"" inField:STPCardFieldTypeExpiration];
```
bdorfman-stripe: I think removing it would be a breaking API change and I'm not sure what the full consequences are. It is not wholly constructed from other data in this view, it can also save and return non-rendered data. Eg look at `testSetCard_addressFields` in `STPPaymentCardTextFieldTest.m`

Stripe/STPPaymentCardTextField.m
bdorfman-stripe: Ah maybe. I'm not sure how the linter detects these honestly. It may only trigger on strings literals that it can tell are passed directly to UIKit properties? I'll mark them as nonlocalized for clarity.

Stripe/STPPaymentCardTextField.m
joeydong-stripe: yeah you're right. i didn't realize STPCardParams had all of those non-card related fields.

bdorfman-stripe:
Ready for re-review I think @joeydong-stripe

joeydong-stripe:
lgtm (besides the build errors) :shipit:

bdorfman-stripe:
I think this is good to go now. @bg-stripe or @joeydong-stripe do you have any final feedback before I merge?

joeydong-stripe:
👍

PR #728, opened 2017-07-10
Add "Stripe iOS Objective-C Style Guide" document

joeydong-stripe:
## Summary

Add our first version of an objc style guide.

## Motivation

The style guide should help us write more coherent code and be a little less error prone.

r? @bdorfman-stripe 
cc @bg-stripe

### DIFF
--- a/StyleGuide.md
+++ b/StyleGuide.md
@@ -1,0 +1,265 @@
+# Stripe iOS Objective-C Style Guide
+
+## Ground Rules
+
+### Spacing
+
+- Indent using 4 spaces. No tabs.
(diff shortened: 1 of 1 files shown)


bdorfman-stripe:
LGTM!

Btw I think near the nullability section we should also say something about annotating collections with generics information in all headers when possible. (eg do `NSArray<STPSomeType *> *` instead of `NSArray *` for homogenous collections).

joeydong-stripe:
hmm yeah maybe we can add a full "nullability annotations" section next pass

PR #729, opened 2017-07-11
Improve usability for `build_documentation.sh`

joeydong-stripe:
## Summary & Motivation

Some minor improvements to make the script easier to use:

- Print help doc if jazzy is not installed
- Support executing from any directory
- Adjust clean step so that saving a copy of `index.html` is unnecessary
- Add informational logs

## Testing

Verified docs still generate successfully and tested various exit branches.

r? @bdorfman-stripe 
cc @bg-stripe


joeydong-stripe:
thx

PR #731, opened 2017-07-11
Minor Improvements to `STPAPIClient`

joeydong-stripe:
## Summary

Some minor improvements to the `STPAPIClient`. See individual commits.

## Motivation

Cleaning up the implementation in a separate PR before I add to it soon.

## Testing

Verified tests still pass.

r? @bdorfman-stripe 
cc @bg-stripe

### DIFF
--- a/Stripe/STPAPIClient+Private.h
+++ b/Stripe/STPAPIClient+Private.h
@@ -11,21 +11,26 @@
 #import "STPAPIClient.h"
 #import "STPAPIRequest.h"
 
-NS_ASSUME_NONNULL_BEGIN
-
 @class STPEphemeralKey;
 
-@interface STPAPIClient()
+NS_ASSUME_NONNULL_BEGIN
+
+@interface STPAPIClient ()
 
 + (NSString *)apiVersion;
 
 - (void)createTokenWithParameters:(NSDictionary *)parameters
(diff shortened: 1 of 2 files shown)


Stripe/STPAPIClient+Private.h
```diff
@@ -11,21 +11,26 @@
...
-- (NSURLSessionDataTask *)retrieveSourceWithId:(NSString *)identifier clientSecret:(NSString *)secret responseCompletion:(STPAPIResponseBlock)completion;
 
-@property (nonatomic, readwrite) NSURL *apiURL;
-@property (nonatomic, readwrite) NSURLSession *urlSession;
+@property (nonatomic, strong, readwrite) NSURL *apiURL;
+@property (nonatomic, strong, readwrite) NSURLSession *urlSession;
+
+@end
+
+@interface STPAPIClient (SourcesPrivate)
```
bdorfman-stripe: Why move to a separate category?

joeydong-stripe: We cant have both `STPAPIClient (Sources)` in both the public and private header so this is a workaround for that. What do you think?

bdorfman-stripe: Ah ok, makes sense.

PR #732, opened 2017-07-11
Add support for deleting sources from `STPPaymentMethodViewController`

joeydong-stripe:
# Summary

artboard

Adds support for allowing users to delete payment methods using the `STPPaymentMethodsViewController`. This is currently restricted so that the "selected" payment method cannot be deleted.

r? @bg-stripe 
cc @bdorfman-stripe

### DIFF
--- a/Stripe/PublicHeaders/STPPaymentMethodsViewController.h
+++ b/Stripe/PublicHeaders/STPPaymentMethodsViewController.h
@@ -75,6 +75,17 @@ NS_ASSUME_NONNULL_BEGIN
 @property(nonatomic, strong, nullable)STPUserInformation *prefilledInformation;
 
 /**
+ Determines whether or not the user is able to delete payment methods using this view controller
+
+ The user can tap the "Edit" button or swipe left on a payment method to delete it if enabled.
+
+ Currently, the user is not allowed to delete the selected payment method but this may change in the future.
+
+ Default is NO.
+ */
+@property (nonatomic, assign, readwrite) BOOL canDeletePaymentMethods;
+
+/**
  *  If you're pushing `STPPaymentMethodsViewController` onto an existing `UINavigationController`'s stack, you should use this method to dismiss it, since it may have pushed an additional add card view controller onto the navigation controller's stack.
  *
  *  @param completion The callback to run after the view controller is dismissed. You may specify nil for this parameter.
--- a/Stripe/PublicHeaders/STPPaymentMethodsViewController.h
+++ b/Stripe/PublicHeaders/STPPaymentMethodsViewController.h
@@ -75,17 +75,6 @@ NS_ASSUME_NONNULL_BEGIN
 @property(nonatomic, strong, nullable)STPUserInformation *prefilledInformation;
 
 /**
- Determines whether or not the user is able to delete payment methods using this view controller
-
- The user can tap the "Edit" button or swipe left on a payment method to delete it if enabled.
-
- Currently, the user is not allowed to delete the selected payment method but this may change in the future.
-
- Default is NO.
- */
-@property (nonatomic, assign, readwrite) BOOL canDeletePaymentMethods;
-
-/**
  *  If you're pushing `STPPaymentMethodsViewController` onto an existing `UINavigationController`'s stack, you should use this method to dismiss it, since it may have pushed an additional add card view controller onto the navigation controller's stack.
  *
  *  @param completion The callback to run after the view controller is dismissed. You may specify nil for this parameter.
--- a/Stripe/STPPaymentMethodTableViewCell.h
+++ b/Stripe/STPPaymentMethodTableViewCell.h
@@ -7,12 +7,20 @@
 //
 
 #import <UIKit/UIKit.h>
-#import "STPPaymentMethod.h"
-#import "STPTheme.h"
+
+@class STPTheme;
+
+@protocol STPPaymentMethod;
+
+NS_ASSUME_NONNULL_BEGIN
 
 @interface STPPaymentMethodTableViewCell : UITableViewCell
 
-- (void)configureWithPaymentMethod:(id<STPPaymentMethod>)paymentMethod selected:(BOOL)selected theme:(STPTheme *)theme;
+- (instancetype)initWithStyle:(UITableViewCellStyle)style reuseIdentifier:(nullable NSString *)reuseIdentifier;
+
 - (void)configureForNewCardRowWithTheme:(STPTheme *)theme;
+- (void)configureWithPaymentMethod:(id<STPPaymentMethod>)paymentMethod theme:(STPTheme *)theme selected:(BOOL)selected;
 
 @end
+
+NS_ASSUME_NONNULL_END
--- a/Stripe/STPPaymentMethodTableViewCell.h
+++ b/Stripe/STPPaymentMethodTableViewCell.h
@@ -16,8 +16,6 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface STPPaymentMethodTableViewCell : UITableViewCell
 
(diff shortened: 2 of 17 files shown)


Stripe/STPPaymentMethodsViewController.m
```diff
@@ -267,15 +288,14 @@ - (instancetype)initWithConfiguration:(STPPaymentConfiguration *)configuration
         self.navigationItem.title = STPLocalizedString(@"Loading…", @"Title for screen when data is still loading from the network.");
 
         WEAK(self);
-        [loadingPromise onSuccess:^(STPPaymentMethodTuple *tuple) {
-            STRONG(self);
-            self.paymentMethods = tuple.paymentMethods;
-            self.selectedPaymentMethod = tuple.selectedPaymentMethod;
```
joeydong-stripe: Looked like we were never reading these properties so just getting rid of this extra cached state.

Stripe/UITableViewCell+Stripe_Borders.m
```diff
@@ -15,42 +15,42 @@
...
-    UIView *view = [self.contentView viewWithTag:STPTableViewCellFakeSeparatorTag];
+    UIView *view = [self viewWithTag:STPTableViewCellFakeSeparatorTag];
     if (!view) {
         view = [[UIView alloc] initWithFrame:CGRectMake(0, self.bounds.size.height - 0.5f, self.bounds.size.width, 0.5f)];
         view.autoresizingMask = UIViewAutoresizingFlexibleWidth | UIViewAutoresizingFlexibleTopMargin;
         view.tag = STPTableViewCellFakeSeparatorTag;
         view.backgroundColor = self.backgroundColor;
         view.accessibilityIdentifier = @"stp_fakeSeparatorView";
-        [self.contentView addSubview:view];
+        [self addSubview:view];
```
joeydong-stripe: Ended up moving these separators out of the `contentView` so that they don't move rightward when we start the editing state:

simulator screen shot jul 17 2017 12 09 41 pm

joeydong-stripe:
Should I add `canDeletePaymentMethods` as a configuration option in `STPPaymentContext` too?

Stripe/STPPaymentMethodsViewController.m
```diff
@@ -190,6 +203,14 @@ - (void)internalViewControllerDidSelectPaymentMethod:(id<STPPaymentMethod>)payme
     [self finishWithPaymentMethod:paymentMethod];
 }
 
+- (void)internalViewControllerDidDeletePaymentMethod:(id<STPPaymentMethod>)paymentMethod {
+    if ([self.delegate isKindOfClass:[STPPaymentContext class]]) {
+        // Notify payment context to update its copy of payment methods
+        STPPaymentContext *paymentContext = (STPPaymentContext *)self.delegate;
+        [paymentContext removePaymentMethod:paymentMethod];
+    }
```
joeydong-stripe: this is a little bit of a workaround to avoid adding the "did delete payment method" to the STPPaymentMethodsViewController delegate API

Stripe/STPPaymentMethodTableViewCell.h
```diff
@@ -7,12 +7,20 @@
...
+@class STPTheme;
+
+@protocol STPPaymentMethod;
+
+NS_ASSUME_NONNULL_BEGIN
 
 @interface STPPaymentMethodTableViewCell : UITableViewCell
 
-- (void)configureWithPaymentMethod:(id<STPPaymentMethod>)paymentMethod selected:(BOOL)selected theme:(STPTheme *)theme;
+- (instancetype)initWithStyle:(UITableViewCellStyle)style reuseIdentifier:(nullable NSString *)reuseIdentifier;
```
bg-stripe: do we need to declare this if it's in super? or is this a style choice?

Stripe/STPPaymentMethodsInternalViewController.m
```diff
@@ -9,125 +9,315 @@
...
+
+        id<STPPaymentMethod> paymentMethodToDelete = [self.paymentMethods stp_boundSafeObjectAtIndex:indexPath.row];
+
+        if (![self isPaymentMethodDetachable:paymentMethodToDelete]) {
+            // Showed the user a delete option for a payment method when we shouldn't have
+            [tableView reloadData];
+            return;
+        }
+
+        if (![paymentMethodToDelete conformsToProtocol:@protocol(STPSourceProtocol)]) {
```
bg-stripe: i think this is covered by `isPaymentMethodDetachable` above

Stripe/STPPaymentMethodsViewController.m
bg-stripe: 👍

Stripe/PublicHeaders/STPPaymentMethodsViewController.h
```diff
@@ -75,6 +75,17 @@ NS_ASSUME_NONNULL_BEGIN
...
 /**
+ Determines whether or not the user is able to delete payment methods using this view controller
+
+ The user can tap the "Edit" button or swipe left on a payment method to delete it if enabled.
+
+ Currently, the user is not allowed to delete the selected payment method but this may change in the future.
+
+ Default is NO.
+ */
+@property (nonatomic, assign, readwrite) BOOL canDeletePaymentMethods;
```
bg-stripe: I think we should move this to `STPPaymentConfiguration` – it's where all our configuration options live right now, and we can avoid redefining it in a bunch of places / forwarding.

I also wonder whether we should default it to on. This seems like a pretty safe feature to introduce for everyone, and since we introduced ephemeral keys pretty recently, it won't affect that many people.

We should also add a note that this feature won't be available if you're using STPBackendAPIAdapter, and maybe point to https://github.com/stripe/stripe-ios/blob/master/MIGRATING.md#migration-from-versions--1100

Stripe/UITableViewCell+Stripe_Borders.m
bg-stripe: ah, was wondering – thanks for explaining!

Stripe/STPPaymentMethodTableViewCell.h
joeydong-stripe: hmm yeah we don't need this declaration. i was playing around with the `nullability` annotations but didn't get anywhere with it.

Stripe/STPPaymentMethodsInternalViewController.m
joeydong-stripe: yeah it is but i think it's not too much extra work to have an additional protocol check closer to the casting just in case the implementation of `isPaymentMethodDetachable` changes dramatically.

Stripe/PublicHeaders/STPPaymentMethodsViewController.h
joeydong-stripe: > I think we should move this to STPPaymentConfiguration – it's where all our configuration options live right now, and we can avoid redefining it in a bunch of places / forwarding.

👍  yeah that makes a lot more sense than having this configuration copied down the whole tree 😄 

> I also wonder whether we should default it to on. This seems like a pretty safe feature to introduce for everyone, and since we introduced ephemeral keys pretty recently, it won't affect that many people.

yeah i think we're enabling the user to do what they want and it's not something that they would do unintentionally.

> We should also add a note that this feature won't be available if you're using STPBackendAPIAdapter, and maybe point to https://github.com/stripe/stripe-ios/blob/master/MIGRATING.md#migration-from-versions--1100

got it!

Stripe/STPPaymentMethodsInternalViewController.m
bg-stripe: reasonable!

PR #735, opened 2017-07-13
Expose inputView similar to inputAccessoryView

theill:
## Summary

Allows inputView to be set on child text fields similar to how it's done on inputAccessoryView.

## Motivation

I need to do a custom keyboard and do not want the standard iOS keyboard to be shown. Overriding inputView with my custom keyboard is exactly what I need.

## Testing

I tested it by building the pod locally and setting the inputView property from my custom implementation.

### DIFF
--- a/Stripe/PublicHeaders/STPPaymentCardTextField.h
+++ b/Stripe/PublicHeaders/STPPaymentCardTextField.h
@@ -134,6 +134,11 @@ IB_DESIGNABLE
 @property(nonatomic, assign) UIKeyboardAppearance keyboardAppearance UI_APPEARANCE_SELECTOR;
 
 /**
+ *  This behaves identically to setting the inputView for each child text field.
+ */
+@property(nonatomic, strong, nullable) UIView *inputView;
+
+/**
  *  This behaves identically to setting the inputAccessoryView for each child text field.
  */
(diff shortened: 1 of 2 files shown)


joeydong-stripe:
Thanks for the PR @theill 😄 we really appreciate it. We're going to think about the API implications but will get back to you soon.

Also, if you don't mind, what use case do you have behind adding this hook? Just out of curiosity.

theill:
@joeydong-stripe it's because I'm doing a customized experience looking like https://goo.gl/oZ3pVE i.e. I have customized design of `STPPaymentCardTextField` (which is nicely supported) but I also want the keyboard to be customized so it's not showing the default iOS keypad keyboard.

joeydong-stripe:
Ah thanks for the example. That really helped. This looks good to us!

Just as a heads up, we're working on merging the "postal code" field into `STPPaymentCardTextField` (https://github.com/stripe/stripe-ios/pull/725). This means some regions will need an alphanumeric keyboard to enter their postal code information.

We're thinking in your case, you will be able to add hooks to `[STPPaymentCardTextFieldDelegate paymentCardTextFieldDidBeginEditingPostalCode]` and `[STPPaymentCardTextFieldDelegate paymentCardTextFieldDidEndEditingPostalCode]` and change the `inputView` appropriately if desired.

PR #738, opened 2017-07-18
Improve `STPPaymentConfiguration`

joeydong-stripe:
## Summary

Some cleanup for one of our most visible classes. See individual commit messages.

r? @bg-stripe 
cc @bdorfman-stripe

### DIFF
--- a/Stripe/PublicHeaders/STPPaymentConfiguration.h
+++ b/Stripe/PublicHeaders/STPPaymentConfiguration.h
@@ -90,7 +90,7 @@ NS_ASSUME_NONNULL_BEGIN
  one of these, see our guide at https://stripe.com/docs/mobile/apple-pay . You 
  must set this to a valid identifier in order to automatically enable Apple Pay.
  */
-@property (nonatomic, nullable, copy, readwrite) NSString *appleMerchantIdentifier;
+@property (nonatomic, copy, nullable, readwrite) NSString *appleMerchantIdentifier;
 
 /**
  Determines whether or not the user is able to delete payment methods
--- a/Stripe/PublicHeaders/STPPaymentConfiguration.h
+++ b/Stripe/PublicHeaders/STPPaymentConfiguration.h
@@ -25,62 +25,85 @@ NS_ASSUME_NONNULL_BEGIN
 @interface STPPaymentConfiguration : NSObject<NSCopying>
 
 /**
- This is a convenience singleton configuration that uses the default values
- for every property
+ This is a convenience singleton configuration that uses the default values for
+ every property
  */
 + (instancetype)sharedConfiguration;
 
 /**
- *  Your Stripe publishable key. You can get this from https://dashboard.stripe.com/account/apikeys .
+ Your Stripe publishable key
+ 
+ @see https://dashboard.stripe.com/account/apikeys
  */
-@property(nonatomic, copy)NSString *publishableKey;
+@property (nonatomic, copy, readwrite) NSString *publishableKey;
 
 /**
- *  An enum value representing which payment methods you will accept from your user in addition to credit cards. Unless you have a very specific reason not to, you should leave this at the default, `STPPaymentMethodTypeAll`.
+ An enum value representing which payment methods you will accept from your user
+ in addition to credit cards. Unless you have a very specific reason not to, you
+ should leave this at the default, `STPPaymentMethodTypeAll`.
  */
-@property(nonatomic)STPPaymentMethodType additionalPaymentMethods;
+@property (nonatomic, assign, readwrite) STPPaymentMethodType additionalPaymentMethods;
 
 /**
- *  The billing address fields the user must fill out when prompted for their payment details. These fields will all be present on the returned token from Stripe. See https://stripe.com/docs/api#create_card_token for more information.
+ The billing address fields the user must fill out when prompted for their 
+ payment details. These fields will all be present on the returned token from 
+ Stripe.
+ 
+ @see https://stripe.com/docs/api#create_card_token
  */
-@property(nonatomic)STPBillingAddressFields requiredBillingAddressFields;
+@property (nonatomic, assign, readwrite) STPBillingAddressFields requiredBillingAddressFields;
 
 /**
- *  The shipping address fields the user must fill out when prompted for their shipping info.
+ The shipping address fields the user must fill out when prompted for their
+ shipping info.
  */
-@property(nonatomic)PKAddressField requiredShippingAddressFields;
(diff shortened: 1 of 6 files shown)


Tests/Tests/STPPaymentConfigurationTest.m
```diff
@@ -0,0 +1,120 @@
...
+- (void)testInit {
+    STPPaymentConfiguration *paymentConfiguration = [[STPPaymentConfiguration alloc] init];
+
+    XCTAssertNil(paymentConfiguration.publishableKey);
+    XCTAssertEqual(paymentConfiguration.additionalPaymentMethods, STPPaymentMethodTypeAll);
+    XCTAssertEqual(paymentConfiguration.requiredBillingAddressFields, STPBillingAddressFieldsNone);
+    XCTAssertEqual(paymentConfiguration.requiredShippingAddressFields, PKAddressFieldNone);
+    XCTAssert(paymentConfiguration.verifyPrefilledShippingAddress);
+    XCTAssertEqual(paymentConfiguration.shippingType, STPShippingTypeShipping);
+    XCTAssertNil(paymentConfiguration.companyName); // Defaults to nil for test target
```
joeydong-stripe: Maybe I can use a mock to simulate the `kCFBundleNameKey`

Stripe/PublicHeaders/STPPaymentConfiguration.h
```diff
@@ -25,62 +25,85 @@ NS_ASSUME_NONNULL_BEGIN
...
+@property (nonatomic, copy, readwrite) NSString *publishableKey;
 
 /**
- *  An enum value representing which payment methods you will accept from your user in addition to credit cards. Unless you have a very specific reason not to, you should leave this at the default, `STPPaymentMethodTypeAll`.
+ An enum value representing which payment methods you will accept from your user
+ in addition to credit cards. Unless you have a very specific reason not to, you
+ should leave this at the default, `STPPaymentMethodTypeAll`.
  */
-@property(nonatomic)STPPaymentMethodType additionalPaymentMethods;
+@property (nonatomic, assign, readwrite) STPPaymentMethodType additionalPaymentMethods;
```
bg-stripe: Hmm. I like that leaving out the default values keeps us below the 80 char guide and makes our headers a bit more readable. I think I still prefer being explicit in the end though.

Stripe/PublicHeaders/STPPaymentConfiguration.h
```diff
@@ -25,62 +25,85 @@ NS_ASSUME_NONNULL_BEGIN
...
+@property (nonatomic, copy, readwrite) NSString *companyName;
 
 /**
- *  The Apple Merchant Identifier to use during Apple Pay transactions. To create one of these, see our guide at https://stripe.com/docs/mobile/apple-pay . You must set this to a valid identifier in order to automatically enable Apple Pay.
+ The Apple Merchant Identifier to use during Apple Pay transactions. To create 
+ one of these, see our guide at https://stripe.com/docs/mobile/apple-pay . You 
+ must set this to a valid identifier in order to automatically enable Apple Pay.
  */
-@property(nonatomic, nullable, copy)NSString *appleMerchantIdentifier;
+@property (nonatomic, nullable, copy, readwrite) NSString *appleMerchantIdentifier;
```
bg-stripe: nullable goes after copy in the guide :P

Stripe/STPPaymentConfiguration.m
```diff
@@ -7,13 +7,19 @@
...
 
 #import "NSBundle+Stripe_AppName.h"
 #import "STPAnalyticsClient.h"
-#import "STPPaymentConfiguration+Private.h"
 #import "STPTelemetryClient.h"
 #import "Stripe.h"
 
+@interface STPPaymentConfiguration ()
+
+// See STPPaymentConfiguration+Private.h
```
bg-stripe: nice touch

StyleGuide.md
```diff
@@ -66,6 +66,8 @@ type = dictionary[@"type"] ?: @"default";
 - (void)title;
 ```
 
+- Header documentation should wrap lines to 80 characters
```
bg-stripe: 👍

Tests/Tests/STPPaymentConfigurationTest.m
```diff
@@ -0,0 +1,120 @@
...
+    paymentConfiguration.additionalPaymentMethods = STPPaymentMethodTypeAll;
+
+    XCTAssertFalse([paymentConfiguration applePayEnabled]);
+}
+
+#pragma mark - Description
+
+- (void)testDescription {
+    STPPaymentConfiguration *paymentConfiguration = [[STPPaymentConfiguration alloc] init];
+    XCTAssert(paymentConfiguration.description);
```
bg-stripe: I kind of worry about this implementation drifting, since we add to PaymentConfiguration fairly regularly and it'd be easy to forget to update description. Maybe we can assert that all the properties are present in the string? 

💭  Maybe we should use `class_copyPropertyList` to do this for all `description` implementation tests?

Stripe/PublicHeaders/STPPaymentConfiguration.h
joeydong-stripe: mm okay, lets keep going down this explicitness path. I don't think it's ever going to be reasonable for objc code to fit under 80 characters because of the typical verbose naming conventions.

Stripe/PublicHeaders/STPPaymentConfiguration.h
joeydong-stripe: heheh

Tests/Tests/STPPaymentConfigurationTest.m
joeydong-stripe: yeah, I'll try some things out today 😄

joeydong-stripe: I just tried going down the path of `class_copyPropertyList` and it "works" but it's kind of sketchy because we have to:
- Do a bunch of runtime lookups (probably okay, just some ugly code)
- Derive from `class_copyIvarList` which is closer to what we log but then we go into matching implementation specifics
- Subtract any properties introduced by protocols with `class_copyProtocolList` and `protocol_copyPropertyList`
- Subtract any properties we intentionally don't log because of PII
- Add any derivations like logging `last4` instead of `accountNumber`

So in the end, we'll end up with a test that's going to fail more often than it should because of unexpected reasons.

bg-stripe: ah, ok. i'm fine with this test for now then! we'll just have to remember to update `description`, and i guess it's not the end of the world if it gets out of sync.

Issue #584, opened 2017-03-11
Reload the payment list view controller when a new method is added


irace:
To make it easier to diagnose your issue, please fill out the following:

## Summary

Currently, the payment list view controller is not reloaded when a new payment method is added.

## Code to reproduce

```swift
extension SettingsCoordinator: STPPaymentMethodsViewControllerDelegate {
    func paymentMethodsViewController(_ paymentMethodsViewController: STPPaymentMethodsViewController, didSelect paymentMethod: STPPaymentMethod) {
        // Do nothing
    }
    
    func paymentMethodsViewController(_ paymentMethodsViewController: STPPaymentMethodsViewController, didFailToLoadWithError error: Error) {
        paymentMethodsViewController.hideViewController()
        
        AlertPresenter.errorPresenter(message: NSLocalizedString("Unable to load payment options", comment: ""))
            .present()
    }
    
    func paymentMethodsViewControllerDidFinish(_ paymentMethodsViewController: STPPaymentMethodsViewController){
        /*
         This will pop both the list controller and “add card” controller, necessary when a new card is added because 
         the underlying list doesn’t reload on its own.
         */
        paymentMethodsViewController.dismiss(completion: nil)
    }
}
```

## iOS version

9-10

## Installation method

CocoaPods

## SDK version

10

bg-stripe:
Thanks for reporting this @irace ! Added in #585 – we're planning on shipping a new SDK version this week, and we'll include this in that release.

yujia917:
Looks like the issue is not fixed. The paymentMethod view controller is still not refreshed when a new card is added.

FahadIqbal:
i have same issue open stppaymentview controller then try to add new card viewcontroller not dismissing activty indicator on right side jus spinning

PR #741, opened 2017-07-19
Add UI Examples app

bg-stripe:
This adds a new example app, `UI Examples`, which demos our UI components and doesn't require any setup (it stubs `STPCustomerContext` and `STPAPIClient`).

It's pretty minimal at the moment – definitely open to feature / design suggestions!

I also noticed a couple bugs and fixed them here:
- 61bcffc3 customer.shippingAddress was nonnull, when it should be nullable (noticed this while mocking)
- 42d9239 `PaymentMethodsViewController` was refreshing its customer before selecting a default source. This wouldn't affect most users – if you're using `STPPaymentContext`, `PaymentMethods` gets dismissed immediately when you select a new source. However if you're using `PaymentMethods` on its own, and don't dismiss when selecting a new source (like we're doing in this example), you would end up with stale state. (see #584)

r? @joeydong-stripe 
cc @bdorfman-stripe 

Screenshots:

screenshot 2017-07-19 18 03 23

__

__


Example/UI Examples/MockCustomerContext.swift
```diff
@@ -0,0 +1,118 @@
...
+    var mockDefaultSource: STPSourceProtocol? = nil
+    var mockShippingAddress: STPAddress?
+
+    override init() {
+        // Preload the mock customer with saved cards
+        let visa = [
+            "id": "preloaded_visa",
+            "exp_month": "10",
+            "exp_year": "2020",
+            "last4": "1881",
```
bg-stripe: These last4s are from our test cards for these brands: https://stripe.com/docs/testing#cards
I thought it'd be better to not use "(42424242)4242" and "(55555555)4444" since those numbers are the easiest to type if you're testing adding a card to `PaymentMethods`.

Example/UI Examples/MockAPIClient.swift
```diff
@@ -0,0 +1,75 @@
...
+    let swizzledMethod = class_getInstanceMethod(forClass, swizzledSelector)
+    method_exchangeImplementations(originalMethod, swizzledMethod)
+}
+
+
+extension STPAddCardViewController {
+
+    // We can't swizzle in initialize because it's been deprecated in Swift 3.1.
+    // Instead, we have to call this method before STPAddCardViewController appears.
+    static func startMockingAPIClient() {
```
bg-stripe: this was fun 😝

Example/UI Examples/MockAPIClient.swift
```diff
@@ -0,0 +1,75 @@
...
+        let tokenJSON: [String: Any] = [
+            "id": "\(card.hashValue)",
+            "object": "token",
+            "livemode": false,
+            "created": NSDate().timeIntervalSince1970,
+            "used": false,
+            "card": cardJSON,
+        ]
+        let token = STPToken.decodedObject(fromAPIResponse: tokenJSON)
+        DispatchQueue.main.asyncAfter(deadline: .now() + 0.6) {
```
bg-stripe: added a delay so you can see the spinner in action

Example/UI Examples/CardFieldViewController.swift
```diff
@@ -0,0 +1,38 @@
...
+//  Created by Ben Guo on 7/19/17.
+//  Copyright © 2017 Stripe. All rights reserved.
+//
+
+import UIKit
+import Stripe
+
+class CardFieldViewController: UIViewController {
+
+    let cardField = STPPaymentCardTextField()
```
bg-stripe: I thought about demoing a few different styles a la https://stripe.com/docs/elements/examples , but couldn't think of enough interesting variations...

teich-stripe:
Before we ship, one suggestion is that we actual make the alternate themes look more attractive to sell the feature better.

bg-stripe:
@teich-stripe good suggestion! Updated in 4b8095e – we now have 2 custom themes that match the first two [Elements examples](https://stripe.com/docs/elements/examples)

___

bg-stripe:
(making the STPPaymentCardTextField example nicer – wip while I update)

Example/UI Examples/BrowseViewController.swift
```diff
@@ -0,0 +1,179 @@
...
+            switch row {
+            case 0: self = .STPPaymentCardTextField
+            case 1: self = .STPAddCardViewController
+            case 2: self = .STPPaymentMethodsViewController
+            case 3: self = .STPShippingInfoViewController
+            case 4: self = .ChangeTheme
+            default: return nil
+            }
+        }
+    }
```
joeydong-stripe: Hmm I think I would reverse this definition so that the enum is on Integer and the String is a method:

```swift
enum DemoTwo: Int {
    case STPPaymentCardTextField
    case STPAddCardViewController
    case STPPaymentMethodsViewController
    case STPShippingInfoViewController
    case ChangeTheme

    func title() -> String {
        switch self {
        case .STPPaymentCardTextField:
            return "Card Field"
        ...
    }
}
```

one thing I'm alluding to here is maybe we should detail labels to the rows that indicate the class name

```
Card Field
-> STPPaymentCardTextField

Card Form
-> STPAddCardViewController
...
```

Example/UI Examples/BrowseViewController.swift
```diff
@@ -0,0 +1,179 @@
...
+
+    let customerContext = MockCustomerContext()
+    let themeViewController = ThemeViewController()
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        self.title = "UI Examples"
+        self.tableView.tableFooterView = UIView()
+        self.navigationController?.navigationBar.isTranslucent = false
```
joeydong-stripe: I think we can get rid most of the `self.` references in this class. I think it's helpful to only specify self when we're disambiguating or explicitly capturing a reference to it.

Example/UI Examples/MockAPIClient.swift
joeydong-stripe: 👍

Example/UI Examples/MockCustomerContext.swift
joeydong-stripe: 👍  maybe add what you just said as a comment in the code too

joeydong-stripe:
Maybe not for this PR but I'm wondering if there's value in giving these components a little more context. Here's where I'm currently at:

artboard

Example/UI Examples/BrowseViewController.swift
bg-stripe: ah yeah, that's a good idea – i'll update

Example/UI Examples/BrowseViewController.swift
bg-stripe: My style has usually been explicit `self`, since it visually distinguishes from local vars (this is what we do in the dashboard), but I'm starting to come around – syntax highlighting provides enough clarity. I'll update.

joeydong-stripe: mm okay. we can come back to it. maybe once dan is spun-up and has an opinion 😄

bg-stripe:
Interesting – I was envisioning this app as just a simple visual demo of our UI components, with fuller reference integrations in other apps (and perhaps a playground demonstrating how to use non-UI things like STPAPIClient). Worth thinking about more though!

bg-stripe:
(I didn't end up updating the card field example – the variations I made didn't seem different enough to be useful).

Example/UI Examples/MockCustomerContext.swift
bg-stripe: 👍

bg-stripe:
Tweaked a bit in e3b716d:  
- update titles to be more descriptive (and added a subtitle with the class name)
- all examples are now presented in themed modals, including the STPPaymentCardTextField example

___

joeydong-stripe:
👍

PR #746, opened 2017-07-21
Add ephemeral key decoding error

bg-stripe:
We now return a more useful error when decoding an ephemeral key fails:

I also noticed we were missing a reference to `StripeError+Private.m` in our project, probably from a merge error. We weren't using anything in `StripeError+Private`, which is why this didn't turn up earlier. I've:
* added `StripeError+Private` to the project, and renamed it to `StripeError+Internal`. I think we should reserve `Private` for private headers, and this is a private category. Because we were using `Private` in the name, our `check_category_linking` script didn't catch this issue.
* removed some unused errors/codes: the checkout errors are from Remember Me, which we've removed, and CustomerContext can't be missing a key provider anymore.
* updates the `check_category_linking` script so that it catches our two categories without `+`s in their names

r? @joeydong-stripe


bg-stripe:
Since it's easy to overlook our error message, I've also added an NSAssert in `ea437ca`.

cc @fred-stripe

Stripe/StripeError+Internal.m
```diff
@@ -0,0 +1,24 @@
...
+    NSDictionary *userInfo = @{
+                               NSLocalizedDescriptionKey: [self stp_unexpectedErrorMessage],
+                               STPErrorMessageKey: @"Failed to decode the ephemeral key. Make sure your backend is sending the unmodified JSON of the ephemeral key to your app."
+                               };
+    return [[self alloc] initWithDomain:StripeDomain code:STPEphemeralKeyDecodingError userInfo:userInfo];
+}
+
+@end
+
+void linkNSErrorInternalCategory(void) {}
```
joeydong-stripe: I think we can actually move this code into a single `StripeError.m` implementation but we would will still need the `StripeError+Internal.h`.

I did this recently with the `STPAPIClient (Sources) / STPAPIClient (SourcesPrivate)`:

https://github.com/stripe/stripe-ios/blob/master/Stripe/PublicHeaders/STPAPIClient.h#L238
https://github.com/stripe/stripe-ios/blob/master/Stripe/STPAPIClient%2BPrivate.h

ci_scripts/check_category_linking.rb
```diff
@@ -22,9 +22,17 @@
 all_headers = (Dir.glob("Stripe/PublicHeaders/*.h") + Dir.glob("Stripe/*.h")).map do |h|
   filename = File.basename(h)
 end
-categories = (all_headers.select do |h|
-  h.include?("+") && !h.include?("Private") && !h.include?("Fabric")
-end).sort
+
+def is_category(h)
+  if h.include?("+")
+    return !h.include?("Private") && !h.include?("Fabric")
```
joeydong-stripe: I'm trying to understand what this category linking is doing and looking at https://medium.com/ios-os-x-development/categories-in-static-libraries-78e41f8ddb96

So, the problem is that whenever we have an implementation file that doesn't implement any classes, we have to manually link it? Which is why all of our `+Private.h` headers are fine? If that's it, then maybe this script needs to be updated to look at implementations instead of headers. Then we can get rid of these exceptions.

Backing up from that, it seems like asking developers to add `-ObjC` to their build flags is pretty standard and would let us avoid this problem. Our check script feels brittle and this problem is difficult to notice during our day-to-day development. I guess the hard part would be transitioning to this model. Maybe we can craft a check on initialization that would throw an assertion error.

Stripe/StripeError+Internal.h
```diff
@@ -0,0 +1,17 @@
...
+//  NSError+StripeInternal.h
+//  Stripe
+//
+//  Created by Ben Guo on 5/22/17.
+//  Copyright © 2017 Stripe, Inc. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+@interface NSError (StripeInternal)
```
joeydong-stripe: Should we just make `StripeError` a subclass of `NSError` instead of a category? I think I've seen that pattern before.

Stripe/STPEphemeralKeyManager.m
```diff
@@ -96,7 +96,16 @@ - (void)getCustomerKey:(STPEphemeralKeyCompletionBlock)completion {
...
+                    NSError *err;
+                    // the API request failed
+                    if (error) {
+                        err = error;
+                    }
+                    // the ephemeral key could not be decoded
+                    else {
+                        NSAssert(NO, @"Could not parse the ephemeral key response. Make sure your backend is sending the unmodified JSON of the ephemeral key to your app. For more info, see https://stripe.com/docs/mobile/ios/standard#prepare-your-api");
+                        err = [NSError stp_ephemeralKeyDecodingError];
+                    }
```
joeydong-stripe: We should add a test to verify when `stp_ephemeralKeyDecodingError` is returned

Stripe/StripeError+Internal.m
bg-stripe: ah yeah, that'd be simpler – I'll move

Stripe/STPEphemeralKeyManager.m
bg-stripe: 👍  good call

ci_scripts/check_category_linking.rb
bg-stripe: Yeah, the reliance on filenames in our current check is definitely brittle... updating to look at implementation files is a good call.

I think I'd still prefer not requiring adding `-ObjC` to use the static lib – Fabric uses the static lib, and I think adding an additional step to the "click to install" integration wouldn't be great.

Stripe/StripeError+Internal.h
bg-stripe: Ah, to avoid the category linking dance? Maybe! It'd be a breaking change to `StripeError.h` though, so I'd prefer punting.

bg-stripe:
@joeydong-stripe improved `check_category_loading` a bit in 988f72d so that it checks implementation files instead - re-r?

joeydong-stripe:
nice! 👍

PR #747, opened 2017-07-24
Remove Nulls Deeply

joeydong-stripe:
## Summary

Improve the NSDictionary / NSArray remove nulls helpers to scan deeply instead of only at the first level.

## Motivation

Nulls are particularly difficult to work with in any situation so just filter them out everywhere.

## Testing

Updated unit tests.


bg-stripe:
hmm. my guess is there probably aren't many users at the intersection of
- use `allResponseFields`
- explicitly check for `NSNull` in a nested dictionary, and don't also check for nil

seems like a worthwhile improvement to me, and worth the small risk of a subtle break as long as we  call it out in MIGRATING

joeydong-stripe:
I guess the trickiest thing you can end up doing is something like:

```objc
if (card.allResponseFields[@"metadata"][@"username"] != nil) {
    // Do something without actually checking the value of metadata.username
}
```

joeydong-stripe:
Oops forgot too add an entry to the CHANGELOG

PR #757, opened 2017-08-01
Improve jazzy version check to support multiple local installed versions

joeydong-stripe:
The output of `gem search —local` includes all historic versions
installed which does not compare nicely with `gem search —remote`

r? @bdorfman-stripe 
cc @bg-stripe


joeydong-stripe:
Hmm that's a good point. I'm not sure what it means to have older gems installed but we can use a better source of truth: `jazzy --version`.

bdorfman-stripe:
👍

PR #760, opened 2017-08-02
Split up cards and card params.

bdorfman-stripe:
Make STPCard not subclass STPCardParams.
Make many STPCard properties readonly.
Deprecate some public methods that didn't actually need to be public.
Deprecate broken out address methods in favor of STPAddress object.

This brings STPCard and STPCardParams more line with the sources API.

Note: I did a bunch of deprecation marks, but we could discuss just deleting those methods instead?


CHANGELOG
```diff
@@ -1,4 +1,5 @@
 == X.Y.Z 2017-XX-YY
+* Many methods on `STPCard` and `STPCardParams` have been deprecated or made readonly to bring card objects more in line with the rest of the API. See MIGRATING for further details.
```
joeydong-stripe: Just starting a discussion, what do we think about starting all of our changelog lines with "Adds / Improves / Fixes"? In this case, I think it would be more like "Fixes many methods..."

Stripe/PublicHeaders/STPCard.h
```diff
@@ -31,123 +31,151 @@ typedef NS_ENUM(NSInteger, STPCardFundingType) {
...
+ The cardholder's name.
  */
-+ (NSString *)stringFromBrand:(STPCardBrand)brand;
+@property (nonatomic, copy, nullable, readonly) NSString *name;
 
 /**
- This parses a string representing a card's funding type into the appropriate `STPCardFundingType` enum value, i.e. `[STPCard fundingFromString:@"prepaid"] == STPCardFundingTypePrepaid`.
+ The cardholder's address.
+ */
+@property (nonatomic, nonnull, readonly) STPAddress *address;
```
joeydong-stripe: +1. Also wrapping all header comments at the 80th column.

Stripe/PublicHeaders/STPCard.h
```diff
@@ -31,123 +31,151 @@ typedef NS_ENUM(NSInteger, STPCardFundingType) {
...
- @return an STPCard instance populated with the provided values.
+ You cannot directly instantiate an `STPCard`. You should only use one that has been returned from an `STPAPIClient` callback.
  */
-- (instancetype)initWithID:(NSString *)cardID
-                     brand:(STPCardBrand)brand
-                     last4:(NSString *)last4
-                  expMonth:(NSUInteger)expMonth
-                   expYear:(NSUInteger)expYear
-                   funding:(STPCardFundingType)funding;
+- (nonnull instancetype) init __attribute__((unavailable("You cannot directly instantiate an STPCard. You should only use one that has been returned from an STPAPIClient callback.")));
```
joeydong-stripe: nice guard!

bdorfman-stripe:
Updated a bunch of the property declarations. I went with leaving out the strong/copy/assign part from the readonly headers and just putting them in the readwrite versions in the implementation. Not sure if we have an opinion one way or another on that?

joeydong-stripe:
>  I went with leaving out the strong/copy/assign part from the readonly headers and just putting them in the readwrite versions in the implementation.

Yeah I wanted to lean that way too. We can update the style guide to reflect that.

bg-stripe:
> I went with leaving out the strong/copy/assign part from the readonly headers and just putting them in the readwrite versions in the implementation.

+1

CHANGELOG
bg-stripe: 👍 , I think changelog items should be declarative, like commit messages

Stripe/STPCard.m
```diff
@@ -13,20 +13,23 @@
 #import "STPImageLibrary+Private.h"
 #import "STPImageLibrary.h"
 
+NS_ASSUME_NONNULL_BEGIN
```
bg-stripe: 🎉

PR #761, opened 2017-08-03
Split bank account from params

bdorfman-stripe:
Similar to previous PR for STPCard.
Also add STPSourceProtocol to STPBankAccount because it theoretically is a customer payment source as well.

### DIFF
--- a/Stripe/PublicHeaders/STPBankAccount.h
+++ b/Stripe/PublicHeaders/STPBankAccount.h
@@ -56,7 +56,7 @@ typedef NS_ENUM(NSInteger, STPBankAccountStatus) {
  You cannot directly instantiate an `STPBankAccount`. You should only use one 
  that has been returned from an `STPAPIClient` callback.
  */
-- (nonnull instancetype) init __attribute__((unavailable("You cannot directly instantiate an STPBankAccount. You should only use one that has been returned from an STPAPIClient callback.")));
+- (instancetype)init __attribute__((unavailable("You cannot directly instantiate an STPBankAccount. You should only use one that has been returned from an STPAPIClient callback.")));
 
 /**
  The routing number for the bank account. This should be the ACH routing number,
--- a/Stripe/PublicHeaders/STPBankAccount.h
+++ b/Stripe/PublicHeaders/STPBankAccount.h
@@ -10,85 +10,108 @@
 
 #import "STPBankAccountParams.h"
 #import "STPAPIResponseDecodable.h"
+#import "STPSourceProtocol.h"
 
+NS_ASSUME_NONNULL_BEGIN
 
 /**
  Possible validation states for a bank account.
-
- - STPBankAccountStatusNew: The account has had no activity or validation performed
- - STPBankAccountStatusValidated: Stripe has determined this bank account exists.
- - STPBankAccountStatusVerified: Bank account verification has succeeded.
- - STPBankAccountStatusVerificationFailed: Verification for this bank account has failed.
- - STPBankAccountStatusErrored: A transfer sent to this bank account has failed.
  */
 typedef NS_ENUM(NSInteger, STPBankAccountStatus) {
+    /**
+     The account has had no activity or validation performed
+     */
     STPBankAccountStatusNew,
+
+    /**
+     Stripe has determined this bank account exists.
+     */
     STPBankAccountStatusValidated,
+
+    /**
+     Bank account verification has succeeded.
+     */
     STPBankAccountStatusVerified,
+
+    /**
+     Verification for this bank account has failed.
+     */
     STPBankAccountStatusVerificationFailed,
+
+    /**
+     A transfer sent to this bank account has failed.
+     */
     STPBankAccountStatusErrored,
 };
 
 /**
- Representation of a user's bank account details that have been tokenized with the Stripe API
+ Representation of a user's bank account details that have been tokenized with
+ the Stripe API.
 
  @see https://stripe.com/docs/api#bank_accounts
  */
-@interface STPBankAccount : STPBankAccountParams<STPAPIResponseDecodable>
+@interface STPBankAccount : NSObject<STPAPIResponseDecodable, STPSourceProtocol>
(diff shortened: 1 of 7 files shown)


Stripe/PublicHeaders/STPBankAccount.h
```diff
@@ -10,85 +10,108 @@
...
     STPBankAccountStatusVerified,
+
+    /**
+     Verification for this bank account has failed.
+     */
     STPBankAccountStatusVerificationFailed,
+
+    /**
+     A transfer sent to this bank account has failed.
+     */
```
joeydong-stripe: did documenting enums this way turn out to be better? i think we're flip-flop-ing on it.

Stripe/PublicHeaders/STPBankAccount.h
```diff
@@ -10,85 +10,108 @@
...
-@interface STPBankAccount : STPBankAccountParams<STPAPIResponseDecodable>
+@interface STPBankAccount : NSObject<STPAPIResponseDecodable, STPSourceProtocol>
 
 /**
- The last 4 digits of the bank account's account number.
+ You cannot directly instantiate an `STPBankAccount`. You should only use one 
+ that has been returned from an `STPAPIClient` callback.
  */
-- (nonnull NSString *)last4;
+- (nonnull instancetype) init __attribute__((unavailable("You cannot directly instantiate an STPBankAccount. You should only use one that has been returned from an STPAPIClient callback.")));
```
joeydong-stripe: we can drop the `nonnull` here. also is it normal to have a space before `init` when it's marked `unavailable`?

Stripe/PublicHeaders/STPBankAccount.h
```diff
@@ -10,85 +10,108 @@
...
 
 /**
- The routing number for the bank account. This should be the ACH routing number, not the wire routing number.
+ The routing number for the bank account. This should be the ACH routing number,
+ not the wire routing number.
  */
-@property (nonatomic, copy, nonnull) NSString *routingNumber;
+@property (nonatomic, nullable, readonly) NSString *routingNumber;
```
joeydong-stripe: heheh finally corrected these 👏

Stripe/STPBankAccount.m
```diff
@@ -49,31 +56,21 @@ + (STPBankAccountStatus)statusFromString:(NSString *)string {
     return STPBankAccountStatusNew;
 }
 
-+ (NSString *)stringFromStatus:(STPBankAccountStatus)status {
++ (nullable NSString *)stringFromStatus:(STPBankAccountStatus)status {
```
joeydong-stripe: a little bit of a leap but I think we can offer `nonnull` here?

Stripe/STPBankAccount.m
```diff
@@ -127,7 +124,7 @@ + (NSArray *)requiredFields {
              ];
 }
 
-+ (instancetype)decodedObjectFromAPIResponse:(NSDictionary *)response {
++ (nullable instancetype)decodedObjectFromAPIResponse:(nullable NSDictionary *)response {
```
joeydong-stripe: looking through the `requiredFields` list, should we add:
- account_holder_type
- fingerprint
- status
to support our nonnull annotation?

joeydong-stripe: i guess this goes back to our discussion on how do we determine what fields we actually non-optional from the API

joeydong-stripe:
oh snap, the basic `testDescription` case actually caught something 😆

Stripe/PublicHeaders/STPBankAccount.h
bdorfman-stripe: Yeah after playing around with it a bit, SourceKit (the Apple framework both Xcode and jazzy use for documentation) only recognizes enum docs written out this way, even though the Xcode doc generator doesn't do this. I'm gonna change everything to this as I audit the rest of our docs and will update our style guide.

Stripe/PublicHeaders/STPBankAccount.h
bdorfman-stripe: No, I think I just copy and pasted from somewhere else that had it wrong. I'll fix both things.

Stripe/STPBankAccount.m
bdorfman-stripe: Ah yeah good point. Theoretically as long as you only pass in valid enum values it will always be nonnull, I'll fix.

Stripe/STPBankAccount.m
bdorfman-stripe: I went the other way in most of this, where I treated whatever was in `requiredFields` as currently correct, and tried to change the property declaration nullability to match. I might have missed a few, I can fix the header if needed. I think for now that is a better solution than changing `requiredFields`

Stripe/STPBankAccount.m
```diff
@@ -11,14 +11,21 @@
 #import "NSDictionary+Stripe.h"
 #import "STPBankAccountParams+Private.h"
 
+NS_ASSUME_NONNULL_BEGIN
```
bg-stripe: 🎉

Stripe/STPBankAccount.m
bdorfman-stripe: Wait, maybe I misunderstand your initial comment? Status is already listed as a required field. The other two are marked as nullable in the header, so I think this is correct already?

joeydong-stripe: Oh yeah, just double checked. I was looking at the property definitions in the implementation. 

I think the only one out of sync is `fingerprint` is `nullable` in the header and `nonnull (default)` in the implementation.

Also the `accountHolderType` will default to `STPBankAccountHolderTypeIndividual` if it's not defined in the response which is a little sketchy but probably the right fallback anyway.

bdorfman-stripe: Yeah I agree that's sketchy, but not sure what to do about it.

I can fix the other nullability mismatch.

PR #762, opened 2017-08-04
Add all the possible file upload purposes

bdorfman-stripe:
Noticed that the API docs list a whole bunch more possible values for the file upload `purpose` key, so I added them.

The file upload guide doesn't state what the max file size values are for the other types (or really give any guidance on what they're for that I could find). I have an ask open and will update the branch with correct values before merging.

### DIFF
--- a/Stripe/STPFile.m
+++ b/Stripe/STPFile.m
@@ -11,16 +11,18 @@
 
 #import "NSDictionary+Stripe.h"
 
+NS_ASSUME_NONNULL_BEGIN
+
 @interface STPFile ()
 
-@property (nonatomic, readwrite) NSString *fileId;
-@property (nonatomic, readwrite) NSDate *created;
-@property (nonatomic, readwrite) STPFilePurpose purpose;
-@property (nonatomic, readwrite) NSNumber *size;
-@property (nonatomic, readwrite) NSString *type;
-@property (nonatomic, readwrite, copy) NSDictionary *allResponseFields;
+@property (nonatomic, copy, readwrite) NSString *fileId;
+@property (nonatomic, copy, readwrite) NSDate *created;
+@property (nonatomic, assign, readwrite) STPFilePurpose purpose;
+@property (nonatomic, copy, readwrite) NSNumber *size;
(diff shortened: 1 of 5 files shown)


Stripe/STPFile.m
```diff
@@ -11,16 +11,18 @@
...
-@property (nonatomic, readwrite) STPFilePurpose purpose;
-@property (nonatomic, readwrite) NSNumber *size;
-@property (nonatomic, readwrite) NSString *type;
-@property (nonatomic, readwrite, copy) NSDictionary *allResponseFields;
+@property (nonatomic, copy, readwrite) NSString *fileId;
+@property (nonatomic, copy, readwrite) NSDate *created;
+@property (nonatomic, assign, readwrite) STPFilePurpose purpose;
+@property (nonatomic, copy, readwrite) NSNumber *size;
+@property (nonatomic, copy, readwrite) NSString *type;
+@property (nonatomic, copy, readwrite, copy) NSDictionary *allResponseFields;
```
joeydong-stripe: Just glancing at this PR again, we should list the classes that should be `copy` instead of `strong` in the style guide. Currently it's just:

- NSString
- NSDictionary
- NSArray

Should we add NSDate and NSNumber? or is this just a copy/paste bug?

bdorfman-stripe: NSDate and NSNumber should be copied as well. Not sure if that is really a style guide thing or an ObjC language thing, but we can include it.

bdorfman-stripe: Actually hmmm, maybe they don't need to be copied I guess because there is no mutable version? Probably the actual list of things that should be copied are immutable properties that have mutable subclasses? https://developer.apple.com/library/content/documentation/General/Conceptual/CocoaEncyclopedia/ObjectMutability/ObjectMutability.html

joeydong-stripe: > Probably the actual list of things that should be copied are immutable properties that have mutable subclasses? 

Yeah exactly. Maybe we can link to a list somewhere but I think these three are the overwhelmingly common cases.

PR #763, opened 2017-08-04
100% Documentation coverage

bdorfman-stripe:
Also working on a CI test to error if jazzy doesnt report 100% coverage so that we'll stay this way.


Stripe/PublicHeaders/STPSourceEnums.h
```diff
@@ -0,0 +1,159 @@
+//
+//  STPSourceEnums.h
+//  Stripe
+//
+//  Created by Brian Dorfman on 8/4/17.
+//  Copyright © 2017 Stripe, Inc. All rights reserved.
+//
+
+#ifndef STPSourceEnums_h
```
bg-stripe: I think include guards aren't necessary, since people use `#import`

Stripe/PublicHeaders/StripeError.h
```diff
@@ -42,65 +63,117 @@ typedef NS_ENUM(NSInteger, STPErrorCode) {
...
+ The error type returned by the Stripe API.
+
+ @see https://stripe.com/docs/api#errors-code
+ */
 FOUNDATION_EXPORT NSString * __nonnull const STPStripeErrorTypeKey;
 
 #pragma mark STPCardErrorCodeKeys
 
+/**
+ Possible string values you may receive when there was an error tokenzing
```
bg-stripe: tokenzing -> tokenizing

Stripe/PublicHeaders/StripeError.h
```diff
@@ -42,65 +63,117 @@ typedef NS_ENUM(NSInteger, STPErrorCode) {
...
 FOUNDATION_EXPORT STPCardErrorCode __nonnull const STPExpiredCard;
+
+/**
+ The card was declined.
+ */
 FOUNDATION_EXPORT STPCardErrorCode __nonnull const STPCardDeclined;
-FOUNDATION_EXPORT STPCardErrorCode __nonnull const STPProcessingError;
+
+/**
+ The card has na incorrect CVC.
```
bg-stripe: na -> an

Stripe/NSError+Stripe.h
```diff
@@ -0,0 +1,36 @@
...
+
++ (NSString *)stp_cardErrorInvalidNumberUserMessage;
++ (NSString *)stp_cardInvalidCVCUserMessage;
++ (NSString *)stp_cardErrorInvalidExpMonthUserMessage;
++ (NSString *)stp_cardErrorInvalidExpYearUserMessage;
++ (NSString *)stp_cardErrorExpiredCardUserMessage;
++ (NSString *)stp_cardErrorDeclinedUserMessage;
++ (NSString *)stp_cardErrorProcessingErrorUserMessage;
++ (NSString *)stp_unexpectedErrorMessage;
++ (NSError *)stp_ephemeralKeyDecodingError;
```
joeydong-stripe: should this line be in the above section?

Stripe/PublicHeaders/STPBankAccountParams.h
```diff
@@ -10,8 +10,12 @@
 
 #import "STPFormEncodable.h"
 
+
 /**
  The type of entity that holds a bank account.
+
+ - STPBankAccountHolderTypeIndividual: An individual holds this bank account.
+ - STPBankAccountHolderTypeCompany: A company holds this bank account.
```
joeydong-stripe: do we need the extra copy of the enum case descriptions here?

bdorfman-stripe: Ah no, I think that came back from a merge. I'll delete it.

Stripe/NSError+Stripe.h
bdorfman-stripe: I'll move

bdorfman-stripe:
@joeydong-stripe I added a new script to check coverage and error Travis if its not 100% if you want to re-review with that.

ci_scripts/check_documentation.sh
```diff
@@ -0,0 +1,54 @@
...
+if ! command -v jazzy > /dev/null; then
+  if [[ "${CI}" != "true" ]]; then
+    echo "ERROR: Please install jazzy:"
+    echo "https://github.com/realm/jazzy"
+    exit 1
+  fi
+
+  echo "Installing jazzy..."
+
+  gem install jazzy || die "Executing \`gem install jazzy\` failed"
```
joeydong-stripe: need to copy the `die` function (probably bring in the `info` function too). see `export_builds.sh`.

additionally, you can replace all of the `echo "ERROR: ..." ; exit 1` with calls to `die`

ci_scripts/check_documentation.sh
```diff
@@ -0,0 +1,54 @@
...
+
+# Verify jazzy exit code
+jazzy_exit_code="$?"
+
+if [[ "${jazzy_exit_code}" != 0 ]]; then
+  echo "ERROR: Executing jazzy failed with status code: ${jazzy_exit_code}"
+  exit 1
+fi
+
+# Search for warnings in log file
```
joeydong-stripe: should update the comment 😉.

also, should we fail if there's warnings too? maybe we can check for both (warnings and documentation coverage) and fail the build if either check fails.

bdorfman-stripe: I'm not clear on what other warnings it might generate? I haven't seen any others I don't think.

ci_scripts/check_documentation.sh
bdorfman-stripe: Why `info` and `die` and not what I have? I copied this from the other Travis ci scripts like `check_fauxpas.sh` and `run_analyzer.sh`  (whereas export build doesn't actually need to trigger Travis failures so I wasn't sure it was really the same sort of thing).

joeydong-stripe: Sorry, I meant that `die` isn't a built in function, it's something we write as a helper to shorten implementations. Same reasoning behind `info` which attaches information about what script is generating the log line. Does that clear it up?

bdorfman-stripe: Ah I see what you mean, sorry. Will fix that up.

ci_scripts/check_documentation.sh
joeydong-stripe: Oh nvm, I was confusing the logic from `run_analyzer.sh` which isn't relevant here :)

Issue #756, opened 2017-07-31
Feature request: ability to supply STPPaymentMethodsViewController with a custom footer view


irace:
## Summary

I’d like to be able to supply a custom view in order to add a message akin to the “A standard 3% fee applies...” disclosure from the following Venmo screenshot.

![pasted image at 2017_07_31 09_15 am](https://user-images.githubusercontent.com/399560/28796350-3aa4247e-760b-11e7-8117-a5632c6a6b1b.png)

## Proposal

This is a little tricky since `STPPaymentMethodsViewController` wraps two other potential child view controllers that are each loaded asynchronously. Here’s a first pass at how this could work (note: my Objective-C is _super_ rusty):

`STPPaymentMethodsViewControllerDelegate` could have a new optional method that allows the delegate to supply a custom footer view (this delegate method would take the superview as an argument so that the superview’s width could be used to determine the view’s frame). Ideally this doesn’t leak the fact that it’s going to be installed as the `tableViewFooter`, but the documentation should indicate that Auto Layout is basically a no-go as a result, and that this view’s frame should already be set.

```objc
@optional

- (UIView *)paymentMethodsViewController:(STPPaymentMethodsViewController *)paymentMethodsViewController footerViewForSuperview:(UIView *)superview;
```

Next, both `STPPaymentMethodsInternalViewController` and `STPAddCardViewController` could get new methods like:

```objc
- (void)setFooterView:(UIView *)footerView {
    self.tableView.tableFooterView = footerView;
}
```

After installing a child view controller `STPPaymentMethodsViewController` could ask the delegate for a footer view, and if it gets one, pass it through to the child, whose view should be fully sized at this point.

```objc
// There might be a better way to do this. Again, my Objective-C has lapsed quite a bit
if ([self.delegate respondsToSelector@selector(paymentMethodsViewController:paymentMethodsViewController footerViewForSuperview:)] && [internal respondsToSelector:@selector(setFooterView:)]] {
    UIView *footerView = [self.delegate paymentMethodsViewController:self footerViewForSuperview:internal.view];
    [internal setFooterView:footerView];
}
```

I seem to *mostly* have this working, though had to make a couple of tweaks to avoid excess padding in the case where the “Remember me” cell is hidden (not feeling 100% confident that these are the “right” changes, but they seem to work for my needs):

[...]

PR #767, opened 2017-08-14
Add custom footer view support

irace:
## Summary

Adds the ability to provide a custom footer view that will be added to the view controllers managed by the Stripe SDK.

## Motivation

https://github.com/stripe/stripe-ios/issues/756

## Testing

Call `setFooterView:` on either `STPPaymentMethodsViewController` or `STPPaymentContext`


bdorfman-stripe:
Hey @irace,

Thanks for the pull request! We really appreciate the work you've put into this. Below is some feedback and questions/suggestions we have.

Why the separation between the footer view callbacks for payment methods VC and STPSizedViewProvider? (`paymentMethodsViewController:footerViewForWidth:` vs `sizedViewForWidth:`). Can these be combined into a single callback/protocol? Assuming we merge these two APIs, can we add `customFooterViewProvider` to the `STPFooterViewSupporting` protocol, and perhaps rename that protocol to `STPCustomFooterViewSupporting`?

e.g.
```@protocol STPCustomFooterViewSupporting
@property (nonatomic) id<STPSizedViewProvider> customFooterViewProvider;
- (void)setCustomFooterView:(UIView *)footerView;
@end
```

Should the STPPaymentContext callback provide some sort of info on what view is showing the footer? (Add Card Vs Payment Methods vs any other screen we might want to expand footer support to, such as shipping view). Perhaps an enum value?

It seems like this code assumes the width of the view controller will never change, as the width is only provided once. Perhaps just a comment mandating that the passed in views properly implement `sizeThatFits:` and use that to resize the footers in the our view controllers' `viewDidLayoutSubviews`?

Let us know if these proposed changes make sense based on how your app will be using this feature. Thanks again for your help!

irace:
> Why the separation between the footer view callbacks for payment methods VC and STPSizedViewProvider? (`paymentMethodsViewController:footerViewForWidth:` vs `sizedViewForWidth:`). Can these be combined into a single callback/protocol?

Yeah, I’m not thrilled with the differences here either. Basically, the view controller can ask for a sized footer synchronously via its delegate protocol, where `STPPaymentContext` doesn’t currently have a delegate protocol (and I didn’t want to add one), so I needed to come up with an abstraction (`STPSizedViewProvider`) that would allow it to ask asynchronously later on, when needed.

In the interest of uniformity, we could change the view controller delegate method to return an object conforming to `STPSizedViewProvider` instead of a `UIView`? In this case, the width wouldn’t even need to leak into the delegate API.

The downside is that this would be an additional layer of abstraction that could be somewhat confusing to someone who is only using the view controller delegate, and not `STPPaymentContext`. They might wonder why they need to create a new type that vends a `UIView` (`STPSizedViewProvider`) instead of just vending the `UIView` themselves. But this doesn’t seem like a huge deal to me. I‘m happy to make this change if you’d like 👍.

> Should the STPPaymentContext callback provide some sort of info on what view is showing the footer? (Add Card Vs Payment Methods vs any other screen we might want to expand footer support to, such as shipping view). Perhaps an enum value?

Sure, can’t hurt.

> It seems like this code assumes the width of the view controller will never change, as the width is only provided once. Perhaps just a comment mandating that the passed in views properly implement sizeThatFits: and use that to resize the footers in the our view controllers' viewDidLayoutSubviews?

Hmmm - in this case would we ditch `STPSizedViewProvider` altogether in favor of just passing around `UIView` instances that can size themselves?

bdorfman-stripe:
I think requiring people pass views around that size themselves solves several issues if you think that's a reasonable solution. I think it also avoids some of the delegate problems you describe (as now you can just have a property for the view and don't need a callback that passes the width).

irace:
Yeah, I totally agree. Let me see if I can make this work.

irace:
@bdorfman-stripe OK, I ended up simplifying this quite a bit and unifying on simply using setters and objects that all conform to `STPFooterViewSupporting` instead of having multiple ways to do this. 

In the interest of keeping this simple, I didn’t end up using delegation which means I didn’t end up having a method to pass a “screen type” parameter to after all. This could be added in the future but I didn’t want to prematurely add complexity that may not actually prove useful.

Feedback welcome!

bdorfman-stripe:
Looks pretty good to me! We can take it the rest of the way fixing it to pass our CI tests and adding new tests for the new API if you'd like. Thanks for your help!

bdorfman-stripe:
Closing this in favor of https://github.com/stripe/stripe-ios/pull/792. Still working on some tweaks and adding tests, but hopefully ready to ship soon. Thanks again for your work on this @irace !

PR #768, opened 2017-08-14
[WIP] Autocorrecting Linter

joeydong-stripe:
## Summary

Spent some time looking into linting solutions for ObjC but nothing in particular stood out in terms of ease of use and extensibility. Then started going down the path of writing our own so that we can add rules to it as issues come up.

Initially ruled out building this linter on an abstract syntax tree because Xcode uses clang but has many blackbox flags/extensions on it that we can't recreate. The current implementation basically:

- For each header/implementation file
  - For each rule
    - For each line
      - If line matches (regex) a problem
      - Then fix the problem by rewriting the line
      - Else do nothing
 
In the future, we could start piping non-autocorrectable warnings into Xcode using [mayday](https://github.com/marklarr/mayday)

You can test this out by running `./Tests/linter_tests/linter_tests.rb --force Stripe/` which should autocorrect some property attribute ordering violations we currently have in the codebase.

## Motivation

Help us avoid common mistakes / style consistencies through automation.

## Testing

There's a corresponding test suite for the linter itself at `./Tests/linter_tests/tests/run_tests.rb`

wdyt? @bdorfman-stripe , @bg-stripe , @danj-stripe

### DIFF
--- a/Tests/linter_tests/tests/property_rule.rb
+++ b/Tests/linter_tests/tests/property_rule.rb
@@ -1,0 +1,53 @@
+require 'test/unit'
+
+require_relative '../rules/property_rule'
+
+class PropertyRuleTest < Test::Unit::TestCase
+  def test_apply_property_spacing_rule_autocorrect_compact
+    # Should autocorrect property spacing (compact style)
+    line = '@property(nonatomic,strong,readwrite)NSString *string; // comment'
+    expected = '@property (nonatomic, strong, readwrite) NSString *string; // comment'
+    result = Rules::PropertyRule.apply_property_spacing_rule('file_name', [line])
+    assert_equal([expected], result)
+  end
+
+  def test_apply_property_spacing_rule_autocorrect_expanded
+    # Should autocorrect property spacing (expanded style)
+    line = '@property    ( nonatomic,   strong  , readwrite  )    NSString *string; // comment'
+    expected = '@property (nonatomic, strong, readwrite) NSString *string; // comment'
+    result = Rules::PropertyRule.apply_property_spacing_rule('file_name', [line])
+    assert_equal([expected], result)
+  end
+
+  def test_apply_property_attributes_ordering_rule_garbage
+    # Should throw on unrecognized property attribute
+    line = '@property (garbage) NSString *string;'
+    assert_raise do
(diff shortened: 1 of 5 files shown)


Tests/linter_tests/tests/property_rule.rb
```diff
@@ -0,0 +1,53 @@
...
+    assert_equal([expected], result)
+  end
+
+  def test_apply_property_attributes_ordering_rule_non_property
+    # Should do nothing for non property line
+    line = 'NSString *string;'
+    expected = 'NSString *string;'
+    result = Rules::PropertyRule.apply_property_attributes_ordering_rule('file_name', [line])
+    assert_equal([expected], result)
+  end
```
bg-stripe: - Maybe include a test for when a property declaration includes a suffix like `__attribute__...` or `DEPRECATED...`?
- 💭  We're currently inconsistent in how we deprecate. Maybe we should unify and add to the linter and our style guide, i.e. prefer the `DEPRECATED...` variants to `__attribute__...`?

bg-stripe:
Very exciting! This seems like a reasonable approach to me.

joeydong-stripe:
+1 on the `__attribute__` and `DEPRECATED` syntax. I'm starting to realize this regex approach isn't strong enough to meet our needs. We will be able to put together somewhat reliable rules that will warn us but it won't be able to autocorrect without a more complex understanding of the code.

For example, this PR currently parses the most common property syntax:

```
@property (nonatomic, readwrite, nonnull, copy) NSDictionary *allResponseFields;
```

But all of these other ones are valid syntax but look completely different:

```
@property (nonatomic, readonly) id<STPBackendAPIAdapter> apiAdapter __attribute__((deprecated));

@property (nonatomic, copy, nullable) NSString *addressLine1 DEPRECATED_MSG_ATTRIBUTE("Use address.line1");

@property (nonatomic, copy, nullable) IBInspectable NSString *numberPlaceholder;
```

Some interesting next steps that we can consider (suggested by @danj-stripe):
- Dig into OCLint and see if we can get it to autocorrect code (http://oclint.org/)
- Evaluate the actual complexity of building clang extensions (https://github.com/facebook/facebook-clang-plugins)

PR #771, opened 2017-08-15
Jazzy config improvements

joeydong-stripe:
## Summary

General improvements to our jazzy tooling. See individual commits.

## Motivation

Meant to look into the jazzy configuration to see if we can explicitly say something like "These docs are for our latest v11.2.0 release" in the doc pages but didn't find anything straightforward.

## Testing

Verified docs generated as expected. Will be run with next release.

r? @danj-stripe 
cc @stripe/mobile-ios

### DIFF
--- a/.jazzy.yaml
+++ b/.jazzy.yaml
@@ -1,13 +1,16 @@
-objc: true
+# Ordered by output of `jazzy --help config`
+
+output: docs/docs
 clean: true
-author: Stripe
-author_url: https://stripe.com
-github_url: https://github.com/stripe/stripe-ios
-github_file_prefix: https://github.com/stripe/stripe-ios/tree/master
+objc: true
 umbrella_header: Stripe/PublicHeaders/Stripe.h
 framework_root: .
+sdk: iphonesimulator
+author: Stripe
+author_url: https://stripe.com
 module: Stripe
+github_url: https://github.com/stripe/stripe-ios
+github_file_prefix: https://github.com/stripe/stripe-ios/tree/master
 skip_undocumented: true
 hide_documentation_coverage: true
 theme: fullwidth
-sdk: iphonesimulator
--- a/.jazzy.yaml
+++ b/.jazzy.yaml
@@ -10,7 +10,6 @@ author: Stripe
 author_url: https://stripe.com
 module: Stripe
 github_url: https://github.com/stripe/stripe-ios
-github_file_prefix: https://github.com/stripe/stripe-ios/tree/master
 skip_undocumented: true
 hide_documentation_coverage: true
 theme: fullwidth
--- a/.jazzy.yaml
+++ b/.jazzy.yaml
@@ -10,6 +10,7 @@ author: Stripe
 author_url: https://stripe.com
 module: Stripe
 github_url: https://github.com/stripe/stripe-ios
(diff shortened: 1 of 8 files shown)


ci_scripts/build_documentation.sh
```diff
@@ -2,31 +2,37 @@
 
 script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 
+function info {
+  echo "[$(basename ${0})] [INFO] ${1}"
+}
+
+function die {
+  echo "[$(basename ${0})] [ERROR] ${1}"
```
danj-stripe: Is it interesting to echo to `stderr` also?

.jazzy.yaml
```diff
@@ -10,7 +10,6 @@ author: Stripe
 author_url: https://stripe.com
 module: Stripe
 github_url: https://github.com/stripe/stripe-ios
-github_file_prefix: https://github.com/stripe/stripe-ios/tree/master
```
danj-stripe: Could you leave this in, in case someone tries to run jazzy w/out using `ci_scripts/build_documentation.sh`?

I don't know if it'd work, I'm curious.

ci_scripts/build_documentation.sh
```diff
@@ -2,39 +2,46 @@
 
 script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 
+function info {
+  echo "[$(basename ${0})] [INFO] ${1}"
```
danj-stripe: running `shellcheck`, this line introduces error [SC2086](https://github.com/koalaman/shellcheck/wiki/SC2086).

Looks like you need double quotes (unescaped) around the parameter to basename.
```
echo "[$(basename "${0}")] [INFO] ${1}"
```

Without the quotes, spaces in the path to the script will cause bad output.

It looks like the same mistake exists in a couple other scripts, want to fix them up too?

.jazzy.yaml
```diff
@@ -1,13 +1,15 @@
-objc: true
+# Ordered by output of `jazzy --help config`
+
+output: docs/docs
 clean: true
-author: Stripe
-author_url: https://stripe.com
-github_url: https://github.com/stripe/stripe-ios
-github_file_prefix: https://github.com/stripe/stripe-ios/tree/master
```
danj-stripe: Could you leave this in, in case someone tries to run jazzy w/out using `ci_scripts/build_documentation.sh`?

I don't know if it'd work, I'm curious.

ci_scripts/build_documentation.sh
joeydong-stripe: Oops, yup I'll do that. Didn't realize my shellcheck linter was disabled.

.jazzy.yaml
joeydong-stripe: Yup looks like the command line parameter will overwrite the config value 👍

ci_scripts/build_documentation.sh
danj-stripe: I'm sure it's fine, and now I'm not sure what would be more convenient for running and viewing script output.

If I were going to do it, maybe something like `echo "..." | tee /dev/stderr`

danj-stripe:
👍

PR #773, opened 2017-08-16
[WIP] Adding a playground to UI Examples

danj-stripe:
## Summary

Adding a playground to UI Examples

It has pages for STPPaymentCardTextField and STPAddCardViewController, which show off
some of the customization options, and allow users to play with the components directly
inside the playground's timeline.

## Motivation

I have a hypothesis that it'll be an interesting way to allow people to explore the library.

## Testing

Interact with the playground.

If the Stripe framework hasn't been built (new checkout or cleaned build folder), the playground fails in non-obvious ways. It also doesn't appear to recover well.

### DIFF
--- a/Example/Examples.playground/Pages/1. STPPaymentCardTextField.xcplaygroundpage/Contents.swift
+++ b/Example/Examples.playground/Pages/1. STPPaymentCardTextField.xcplaygroundpage/Contents.swift
@@ -1,0 +1,116 @@
+/*:
+ # UI Examples Playground
+ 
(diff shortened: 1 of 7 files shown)


joeydong-stripe:
Yeah looks like this "need to build UI examples before any playgrounds work" is a blocker for now. Maybe we'll come back to this some other time. Thanks for trying it out!

joeydong-stripe:
If this doesn't get fixed anytime soon by Xcode, I bet we can start another repository and check in a copy of `Stripe.framework` so that the dependency always exists.

danj-stripe:
Closing, but leaving the branch in case we revisit this in the near future and find it useful.

PR #779, opened 2017-08-25
Add p24 support to source bindings

bdorfman-stripe:


Stripe/PublicHeaders/STPSourceParams.h
```diff
@@ -248,12 +248,32 @@ NS_ASSUME_NONNULL_BEGIN
...
+ @param returnURL   The URL the customer should be redirected to after they have
+ @param name        The full name of the account holder (optional).
+
+ @return An STPSourceParams object populated with the provided values.
+ */
++ (STPSourceParams *)p24ParamsWithAmount:(NSUInteger)amount
+                                currency:(NSString *)currency
+                                   email:(NSString *)email
+                               returnURL:(NSString *)returnURL
+                                    name:(nullable NSString *)name;
```
joeydong-stripe: is there a pattern we're using for ordering these parameters?

bdorfman-stripe: I don't think so? I've been putting all the optional ones at the end but otherwise I don't think there is a set pattern. Do you have a suggestion?

joeydong-stripe: I guess I would copy what we have in the [P24 Documentation](https://stripe.com/docs/sources/p24) which looks alphabetical after `type`:

```
type
amount
currency
(owner)email
(owner)name
(redirect)returnURL
```

bdorfman-stripe: Hmm, that won't match the rest of the source params methods though? (which I think are basically alphabetical but with optional ones at the end?) Happy to change it going forward for all new apms starting here though.

joeydong-stripe: I think the important point is that `name` feels pretty separated from `email` now even though they're referring to the same "person".

bdorfman-stripe: Changed.

PR #783, opened 2017-09-01
Support native redirects

bdorfman-stripe:
Not going to merge immediately because I think there's still some question about where in the source return data the native url for Alipay is going to be, but this is a general framework for dealing with sources that support native redirects going forward.

### DIFF
--- a/Stripe/STPRedirectContext.m
+++ b/Stripe/STPRedirectContext.m
@@ -11,6 +11,7 @@
 #import "STPDispatchFunctions.h"
 #import "STPSource.h"
 #import "STPURLCallbackHandler.h"
+#import "STPWeakStrongMacros.h"
 #import "NSError+Stripe.h"
 
 #import <SafariServices/SafariServices.h>
@@ -18,6 +19,8 @@
 #define FAUXPAS_IGNORED_IN_METHOD(...)
 
 NS_ASSUME_NONNULL_BEGIN
+
+typedef void (^STPNativeRedirectCompletionBlock)(BOOL success);
 
 @interface STPRedirectContext () <SFSafariViewControllerDelegate, STPURLCallbackListener>
 @property (nonatomic, copy) STPRedirectContextCompletionBlock completion;
@@ -48,14 +51,72 @@ NS_ASSUME_NONNULL_BEGIN
     [self unsubscribeFromNotificationsAndDismissPresentedViewControllers];
 }
 
-- (void)startRedirectFlowFromViewController:(UIViewController *)presentingViewController {
+- (void)performAppRedirectIfPossibleWithCompletion:(STPNativeRedirectCompletionBlock)onCompletion {
     FAUXPAS_IGNORED_IN_METHOD(APIAvailability)
-    if ([SFSafariViewController class] != nil) {
-        [self startSafariViewControllerRedirectFlowFromViewController:presentingViewController];
+
+    if (self.state == STPRedirectContextStateNotStarted) {
+        NSString *nativeUrlString = nil;
+        switch (self.source.type) {
+            case STPSourceTypeAlipay:
+                nativeUrlString = self.source.details[@"native_url"];
+                break;
+            default:
+                // All other sources have no native url support
+                break;
+        }
+
+        NSURL *nativeUrl = nativeUrlString ? [NSURL URLWithString:nativeUrlString] : nil;
+        if (!nativeUrl) {
+            onCompletion(NO);
+            return;
+        }
+
+        // Optimistically start listening in case we get app switched away.
+        // If the app switch fails we'll undo this later
+        _state = STPRedirectContextStateInProgress;
+        [self subscribeToUrlAndForegroundNotifications];
+
+        UIApplication *application = [UIApplication sharedApplication];
+        if ([application respondsToSelector:@selector(openURL:options:completionHandler:)]) {
+
+            WEAK(self);
+            [application openURL:nativeUrl options:@{} completionHandler:^(BOOL success) {
+                if (!success) {
+                    STRONG(self);
+                    self->_state = STPRedirectContextStateNotStarted;
+                    [self unsubscribeFromNotifications];
+                }
+                onCompletion(success);
+            }];
+        }
+        else {
+            _state = STPRedirectContextStateInProgress;
+            BOOL opened = [application openURL:nativeUrl];
+            if (!opened) {
+                _state = STPRedirectContextStateNotStarted;
+                [self unsubscribeFromNotifications];
+            }
+            onCompletion(opened);
+        }
     }
     else {
-        [self startSafariAppRedirectFlow];
+        onCompletion(NO);
     }
+}
+
+- (void)startRedirectFlowFromViewController:(UIViewController *)presentingViewController {
+    FAUXPAS_IGNORED_IN_METHOD(APIAvailability)
+
+    [self performAppRedirectIfPossibleWithCompletion:^(BOOL success) {
+        if (!success) {
+            if ([SFSafariViewController class] != nil) {
+                [self startSafariViewControllerRedirectFlowFromViewController:presentingViewController];
+            }
+            else {
+                [self startSafariAppRedirectFlow];
+            }
+        }
+    }];
 }
 
 - (void)startSafariViewControllerRedirectFlowFromViewController:(UIViewController *)presentingViewController {
--- a/Stripe/STPRedirectContext.m
+++ b/Stripe/STPRedirectContext.m
@@ -30,12 +30,29 @@ typedef void (^STPNativeRedirectCompletionBlock)(BOOL success);
 
 @implementation STPRedirectContext
 
+
+- (nullable NSURL *)nativeRedirectURLForSource:(STPSource *)source {
+    NSString *nativeUrlString = nil;
+    switch (source.type) {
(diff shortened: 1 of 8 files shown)


Stripe/STPRedirectContext.m
```diff
@@ -19,6 +20,8 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+typedef void (^STPNativeRedirectCompletionBlock)(BOOL success);
```
bg-stripe: nit: can we call this `STPBoolCompletionBlock` to match our other block type names?

Stripe/STPRedirectContext.m
```diff
@@ -48,16 +68,64 @@ - (void)dealloc {
...
+        [self subscribeToUrlAndForegroundNotifications];
+
+        UIApplication *application = [UIApplication sharedApplication];
+        if ([application respondsToSelector:@selector(openURL:options:completionHandler:)]) {
+
+            WEAK(self);
+            [application openURL:nativeUrl options:@{} completionHandler:^(BOOL success) {
+                if (!success) {
+                    STRONG(self);
+                    self->_state = STPRedirectContextStateNotStarted;
```
bg-stripe: ooc, why `self->_state=` instead of just `_state=`?

Also, can't we just use the getters/setters in this method instead of ivars? I usually only use ivars in `init`, or if the getter/setter has side effects. Maybe we should add something to the style guide...

Stripe/STPRedirectContext.m
```diff
@@ -48,16 +68,64 @@ - (void)dealloc {
...
+                    [self unsubscribeFromNotifications];
+                }
+                onCompletion(success);
+            }];
+        }
+        else {
+            _state = STPRedirectContextStateInProgress;
+            BOOL opened = [application openURL:nativeUrl];
```
bg-stripe: OOC, do you know if this behaves differently from the new `openURL:options:completionHandler:` method? E.g. does the returned success value here always match the success value in the async completion block?

Tests/Tests/STPRedirectContextTest.m
```diff
@@ -282,4 +282,58 @@ - (void)testSafariAppRedirectFlow_noNotification {
...
+
+    id applicationMock = OCMClassMock([UIApplication class]);
+    OCMStub([applicationMock sharedApplication]).andReturn(applicationMock);
+
+    id mockVC = OCMClassMock([UIViewController class]);
+    [context startRedirectFlowFromViewController:mockVC];
+
+    OCMReject([sut startSafariViewControllerRedirectFlowFromViewController:[OCMArg any]]);
+    OCMReject([sut startSafariAppRedirectFlow]);
+    OCMVerify([applicationMock openURL:[OCMArg any]
```
bg-stripe: let's check the URL argument too with `OCMArg isEqual:`

Tests/Tests/STPRedirectContextTest.m
```diff
@@ -282,4 +282,58 @@ - (void)testSafariAppRedirectFlow_noNotification {
...
+                               options:[OCMArg any]
```
bg-stripe: nit: we can use `[OCMArg isNil]`

Tests/Tests/STPRedirectContextTest.m
```diff
@@ -282,4 +282,58 @@ - (void)testSafariAppRedirectFlow_noNotification {
...
+                                                                  }];
+    id sut = OCMPartialMock(context);
+
+    id applicationMock = OCMClassMock([UIApplication class]);
+    OCMStub([applicationMock sharedApplication]).andReturn(applicationMock);
+
+    id mockVC = OCMClassMock([UIViewController class]);
+    [context startRedirectFlowFromViewController:mockVC];
+
+    OCMVerify([sut startSafariViewControllerRedirectFlowFromViewController:[OCMArg any]]);
```
bg-stripe: nit: we can verify this arg is equal to `mockVC`

Stripe/STPRedirectContext.m
bdorfman-stripe: As far as I know it's always the same, it's described the same in the docs. The main difference in the new API afaict is the options and the asyncronicity.

Stripe/STPRedirectContext.m
bdorfman-stripe: You have to do `self->state` or you get a compiler warning about implicit retain of self.

Yeah there are some style guide stuff here I guess. I didn't actually make setters for these in the .m file, I generally prefer doing it this way, but we can change (And have a style guide conversation).

bg-stripe: oh right, forgot about that compiler warning. 👍

Tests/Tests/STPFixtures.h
```diff
@@ -55,6 +55,16 @@
 + (STPSource *)iDEALSource;
 
 /**
+ A Source object with type Alipay
+ */
++ (STPSource *)alipaySource;
+
+/**
+ A Source object with type Alipay
```
joeydong-stripe: "A Source object with type Alipay and native url"* 😄

Tests/Tests/STPRedirectContextTest.m
joeydong-stripe: ==

Tests/Tests/STPRedirectContextTest.m
joeydong-stripe: ==

Tests/Tests/STPRedirectContextTest.m
```diff
@@ -282,4 +282,58 @@ - (void)testSafariAppRedirectFlow_noNotification {
...
+    OCMStub([applicationMock sharedApplication]).andReturn(applicationMock);
+
+    id mockVC = OCMClassMock([UIViewController class]);
+    [context startRedirectFlowFromViewController:mockVC];
+
+    OCMReject([sut startSafariViewControllerRedirectFlowFromViewController:[OCMArg any]]);
+    OCMReject([sut startSafariAppRedirectFlow]);
+    OCMVerify([applicationMock openURL:[OCMArg any]
+                               options:[OCMArg any]
+                     completionHandler:[OCMArg any]]);
```
joeydong-stripe: In the past, I've been able to use an `[OCMArg checkWithBlock:]` to trigger the completion to test the entire flow:

```
OCMVerify([applicationMock openURL:options:completionHandler:[OCMArg checkWithBlock:^BOOL(void (^)completion) {
  completion();
}]]);
```

Stripe/STPRedirectContext.m
bdorfman-stripe: Hmm, should i move it to STPBlocks.h ? I wasn't sure about just naming it just BoolBlock because the bool is specifically success or failure and wasn't sure if that was generic across what all bool using blocks would be.

Tests/Tests/STPRedirectContextTest.m
bdorfman-stripe: I don't think any of these should be nil?

bdorfman-stripe:
Updated if people want to re-review.

Tests/Tests/STPRedirectContextTest.m
```diff
@@ -282,4 +282,62 @@ - (void)testSafariAppRedirectFlow_noNotification {
...
+                                                                  completion:^(__unused NSString *sourceID, __unused NSString *clientSecret, __unused NSError *error) {
+        XCTFail(@"completion called");
+    }];
+    id sut = OCMPartialMock(context);
+
+    id applicationMock = OCMClassMock([UIApplication class]);
+    OCMStub([applicationMock sharedApplication]).andReturn(applicationMock);
+    OCMStub([applicationMock openURL:[OCMArg any]
+                             options:[OCMArg any]
+                   completionHandler:([OCMArg invokeBlockWithArgs:@YES, nil])]);
```
bg-stripe: nice 👍

joeydong-stripe: sweet!!

Tests/Tests/STPRedirectContextTest.m
```diff
@@ -297,15 +298,18 @@ - (void)testNativeRedirectSupportingSourceFlow_validNativeURL {
...
 
     id mockVC = OCMClassMock([UIViewController class]);
     [context startRedirectFlowFromViewController:mockVC];
 
     OCMReject([sut startSafariViewControllerRedirectFlowFromViewController:[OCMArg any]]);
     OCMReject([sut startSafariAppRedirectFlow]);
-    OCMVerify([applicationMock openURL:[OCMArg any]
-                               options:[OCMArg any]
-                     completionHandler:[OCMArg any]]);
+    OCMVerify([applicationMock openURL:[OCMArg isEqual:sourceURL]
```
joeydong-stripe: I think you can just write `sourceURL` instead of `[OCMArg isEqual:sourceURL]` in the verify calls? Not completely sure.

bdorfman-stripe: Yeah it wasn't clear to me if they were the same. I thought maybe one way did `== sourceURL` and one way did `isEqual:sourceURL`? But I'm not sure.

bdorfman-stripe:
Some minor changes if anyone wants to re-review. Otherwise this has been tested with a live charge and is ready to merge.

PR #789, opened 2017-09-07
Remove custom back button in VCs

bdorfman-stripe:
Only use system back button or Cancel buttons everywhere.

Our fake Back button we had been using doesn't match iOS 11 back buttons and doesn't behave correctly. It will be even harder to fix given iOS 11's different nav bar sizes. Instead just use Cancel buttons or the system back button everywhere instead.


joeydong-stripe:
👍

bdorfman-stripe:
Yeah maybe? I don't think it's that big a deal tbh and adding extra text to the end will just lead to the back button text being truncated anyway.

joeydong-stripe:
There some other adjustments I would make to this page, anyway, some other time:

- Remove or shrink the top package image so that we're aligned with our add card view controller

- Move the shipping price into the lefthand side so that we're not showing both text and a checkmark on the righthand side

screen shot 2017-09-07 at 4 32 18 pm

- Adjust the view controller title to match its purpose (like @bg-stripe mentioned), make sure the button that we tapped to get to this page matches that title, and remove any redundant section headers if needed

PR #791, opened 2017-09-07
Fix addressbook tests for ios11

bdorfman-stripe:
iOS11 returns empty string (@"") instead of nil for empty address fields. Changed checks and STPAddress logic to accomodate this behavior.


joeydong-stripe:
:+1:

PR #792, opened 2017-09-07
Add custom footer view support

bdorfman-stripe:
Taking over from https://github.com/stripe/stripe-ios/pull/767

Adds the ability to provide a custom footer view that will be added to the view controllers managed by the Stripe SDK.


bdorfman-stripe:
Re-r with newest commit @bg-stripe ?

Issue #794, opened 2017-09-13
Disable code coverage


TofPlay:
## Summary

Code coverage has to be disabled for Carthage to work with Xcode 9.

Check Carthage/Carthage#2056 and Carthage/Carthage#2057.

## Code to reproduce

Try to submit an application with stripe
Apple send back an email with the message:
```
Dear developer,

We have discovered one or more issues with your recent delivery for "<your app>". To process your delivery, the following issues must be corrected:

Invalid Bundle - Disallowed LLVM instrumentation. Do not submit apps with LLVM profiling instrumentation or coverage collection enabled. Turn off LLVM profiling or code coverage, rebuild your app and resubmit the app.

Once these issues have been corrected, you can then redeliver the corrected binary.

Regards,

The App Store team
```

## iOS version

iOS 10 and iOS 11

## Installation method

Cartfile
```
github "stripe/stripe-ios" "ios11"
```

## SDK version

branche `ios11`

## Other information

None

bdorfman-stripe:
This should be fixed in the latest release, 11.3.0, which just went up. Please switch to using the standard releases if you were previously using the `ios11` branch. The mainline releases should now support iOS11 correctly and the other unofficial branch is no longer supported.

Please open a new issue if you have any further trouble. Thanks for your report!

PR #795, opened 2017-09-13
Disable code coverage for carthage xcode 9 compatibility

bdorfman-stripe:
Xcode 9 now includes code coverage for all except Archive command, instead of just Test command, which breaks Carthage[0]. They apparently don't want to work around this on their end, so all framework developers have to change their project settings -_-

Fixes https://github.com/stripe/stripe-ios/issues/794

[0] https://github.com/Carthage/Carthage/issues/2056


joeydong-stripe:
👍

PR #797, opened 2017-09-18
[WIP] Run travis on multiple xcode versions

bdorfman-stripe:
A little unclear how travis's `osx_image` property behaves and if you can have it matrix like this, but hopefully this will run our tests on both xcode8 and xcode9 (still both on iOS 10.3 though)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -80,7 +80,7 @@ matrix:
       - TEST_TYPE=analyzer
 
 before_install:
-- SIMULATOR_ID=$(xcrun instruments -s | grep -o $SIMULATOR_NAME \[.*\]" | grep -o
+- SIMULATOR_ID=$(xcrun instruments -s | grep -o "$SIMULATOR_NAME \[.*\]" | grep -o
   "\[.*\]" | sed "s/^\[\(.*\)\]$/\1/")
 script:
 - open -a "simulator" --args -CurrentDeviceUDID $SIMULATOR_ID
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: objective-c
-osx_image: xcode8.3
+osx_image:
+  - xcode8.3
+  - xcode9
 branches:
   only:
     - master
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: objective-c
-osx_image:
-  - xcode8.3
-  - xcode9
+os: osx
+osx_image: xcode8.3
 branches:
   only:
     - master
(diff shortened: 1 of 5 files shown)


joeydong-stripe:
I guess we'll find out in ~hours 😂

bdorfman-stripe:
Yeah 🙄 

Btw if you have any suggestions on a good way to update this yaml so that it tests on both ios 10 and ios 11 I'm all ears. Maybe we only need to actually do that for the build phase though? Or maybe we can just drop ios 10 building at some point?

joeydong-stripe:
> Btw if you have any suggestions on a good way to update this yaml so that it tests on both ios 10 and ios 11 I'm all ears. Maybe we only need to actually do that for the build phase though? Or maybe we can just drop ios 10 building at some point?

I think you would need to update the `run_test.sh` to take the xcodebuild `destination` parameter as a script parameter. Then pass in an iOS 10 simulator and an iOS 11 simulator. Alternatively, Xcode 9 should support running multiple simulators in parallel which may be faster overall.

bdorfman-stripe:
Ah yeah, maybe I'll just look into Xcode9's new command line arguments to get it building both at once.

danj-stripe:
It might be interesting to use the `matrix.include` key, and make `.travis.yml` more explicit. It also gives flexibility to do more combinations. Something like:
```
matrix:
  include:
    - osx_image: xcode8.3
      env: SIMULATOR_NAME='iPhone 6,OS=10.3.1'
    - osx_image: xcode9
      env: SIMULATOR_NAME='iPhone 6,OS=10.3.1'
    - osx_image: xcode9
      env: SIMULATOR_NAME='iPhone 6,OS=11.0'
```

along with the required changes to `run_tests.sh` to plug that into the `destination` parameter.

I think compiling with Xcode 9/iOS 11 SDK and running on iOS 10 has potential behavioral differences compared to running on iOS 11 and compared to building with Xcode 8.3, so it might be interesting to test all 3 combinations.

It might also be a big waste of time, since I'm not aware of meaningful differences, and it adds 50% more time to our builds.

I don't have a strong opinion on whether you'd want to explicitly put simulator strings into travis.yml (only one place to update in the future?) or put more logic into `run_tests.sh` (make it easier to run locally, maybe?).

----

edit: Something more abstract in the `.travis.yml` would make it easier to use new features of the Xcode 9 command line.

bdorfman-stripe:
Makes sense, although it's unclear to me if `osx_image` actually works properly inside matrix includes? See this still open issue: https://github.com/travis-ci/travis-ci/issues/7587 (linked commit from the reporter indicates that they tried a bunch of different ways of specifying osx_image in their matrix and none worked correctly).

bdorfman-stripe:
Ah nice. I'll play around with that then. I think ideally it tests 10.3 on Xcode 8 and 10.3 and 11 simultaneously on Xcode 9.

joeydong-stripe:
Actually, can we just drop Xcode 8 support and test Xcode 9 + iOS 8-11?

At this point, it's going to be increasingly more difficult to even find a copy of Xcode 8 since everyone should be using Xcode 9.

bdorfman-stripe:
We should probably wait a bit to drop Xcode8 officially, as 9 isn't even officially out on the real distribution channels I think. Also lots of places will continue to submit on 8 for a bit probably. I think last year we dropped the previous version with an official announcement after a month or so (we dropped the oldest iOS version we supported at the same time I think).

bdorfman-stripe:
Ok so there isn't a super easy way to do this with Travis it seems. Because of the way matrix include works, in order to use that solution we would have to manually list out every job we want afaict (every combination of xcode version, ios version, and test type). Which is possible, but annoying. Travis _should_ do the combinatorics for us, but doesn't because `osx_image` doesnt matrix itself properly when you list multiple options. This branch currently has them all written out like that but... not sure we want to stick with it.

joeydong-stripe:
I'm okay with this if it works. If we want to hold onto this longterm, we can work on reducing our build parallelism to 2-3 queues instead of one for each script and putting some logic in `test.sh` like:

- If Xcode 8
  - Then test with iOS 10.x
- Else
  - Test with iOS 10.x followed by iOS 11.x

PR #802, opened 2017-09-22
[WIP] Increase Test Device Coverage

joeydong-stripe:
Following up from our learnings in: https://github.com/stripe/stripe-ios/pull/797

# Summary

Adjust our `.travis.yml` and `run_tests.sh` to increase our test coverage. Previously we were testing:

- Xcode 8.x, iOS 10.x

Now we should be testing:

- Xcode 8.x, iOS 8.x
- Xcode 8.x, iOS 9.x
- Xcode 8.x, iOS 10.x
- Xcode 9.x, iOS 8.x
- Xcode 9.x, iOS 9.x
- Xcode 9.x, iOS 10.x
- Xcode 9.x, iOS 11.x

We're targeting the "iPhone 6" because it conveniently supports all of our target iOS versions: iOS 8.x to 11.x. We also removed the `-destination` specifier for `xcodebuild build` (not `xcodebuild test`) commands because it's unnecessary (only the `-sdk` is needed).

As a result, we're discovering a few groups of test case failures that need to be investigated:

- LocalizationTests
  - STPAddCardViewControllerLocalizationTests
  - STPPaymentMethodsViewControllerLocalizationTests
  - STPShippingAddressViewControllerLocalizationTests
- STPAddressTests https://github.com/stripe/stripe-ios/pull/805
- STPApplePayTest (`testPaymentRequestWithMerchantIdentifierCountryCurrency` only)
- STPPaymentCardTextFieldTest (`testIntrinsicContentSize` only)
- STPRedirectContextTest https://github.com/stripe/stripe-ios/pull/804

PR #804, opened 2017-09-25
[WIP] Fix strong reference cycles for STPRedirectContext, Improve tests

joeydong-stripe:
Fixing test failures found in: https://github.com/stripe/stripe-ios/pull/802

# Summary

I think the `STPRedirectContext` was failing tests quietly in our previous `Xcode 8, iOS 10.x` testing but shows up when we enable the other Xcode / iOS versions.

1. The first problem is some of the tests were intended to never call the `unsubscribeFromUrlAndForegroundNotifications`. In particular, because the `STPURLCallbackHandler` [held a strong reference](https://github.com/stripe/stripe-ios/pull/804/commits/534ba2875d2c83c161565f1894808e5ae9afeacc#diff-f5b882a2b86c0c0fce5b2f949b1abadaL35), the `STPRedirectContext<STPURLCallbackListener>` ended up staying alive between tests leading to unexpected results.

2. The second problem is the `startRedirectFlowFromViewController` was implicitly [holding a strong reference to itself](https://github.com/stripe/stripe-ios/pull/804/commits/534ba2875d2c83c161565f1894808e5ae9afeacc#diff-fbf1a3bd7e2932d541757062a7d430f4R103) in its completion block. Reproducing this is similar to the first problem, either put a breakpoint in `dealloc` and make sure it's called after every test, or send a valid url scheme to the `STPURLCallbackHandler` and see if any listener picks up.

3. Finally, we needed to [adjust the test cases](https://github.com/stripe/stripe-ios/pull/804/commits/534ba2875d2c83c161565f1894808e5ae9afeacc#diff-30df8fae75ff2f3bb3a3dbc5f5f28493R359) so that they're compatible with all tested iOS versions

I took a broader stroke in the second commit because I was often getting lost/confused when trying to understand the implementation of `STPRedirectContext`. I think some key method renaming is sufficient to solve that but definitely want to avoid introducing new bugs.

The tests are still really tricky to maintain but the implementation itself is good so a refactoring just to simplify tests is probably a bad idea right now.

TODO: Update documentation, add `CHANGELOG` and `MIGRATING` entry to warn about holding a strong reference to `STPRedirectContext` which was not needed before

### DIFF
--- a/Stripe/STPRedirectContext.m
+++ b/Stripe/STPRedirectContext.m
@@ -100,8 +100,10 @@ typedef void (^STPBoolCompletionBlock)(BOOL success);
 - (void)startRedirectFlowFromViewController:(UIViewController *)presentingViewController {
     FAUXPAS_IGNORED_IN_METHOD(APIAvailability)
 
+    WEAK(self);
     [self performAppRedirectIfPossibleWithCompletion:^(BOOL success) {
         if (!success) {
+            STRONG(self);
             if ([SFSafariViewController class] != nil) {
                 [self startSafariViewControllerRedirectFlowFromViewController:presentingViewController];
             }
--- a/Stripe/STPRedirectContext.m
+++ b/Stripe/STPRedirectContext.m
@@ -23,22 +23,18 @@ NS_ASSUME_NONNULL_BEGIN
 typedef void (^STPBoolCompletionBlock)(BOOL success);
 
 @interface STPRedirectContext () <SFSafariViewControllerDelegate, STPURLCallbackListener>
+
 @property (nonatomic, copy) STPRedirectContextCompletionBlock completion;
 @property (nonatomic, strong) STPSource *source;
 @property (nonatomic, strong, nullable) SFSafariViewController *safariVC;
 @property (nonatomic, assign, readwrite) STPRedirectContextState state;
+
 @end
 
 @implementation STPRedirectContext
 
-- (nullable instancetype)initWithSource:(STPSource *)source
-                             completion:(STPRedirectContextCompletionBlock)completion {
-
-    if (source.flow != STPSourceFlowRedirect
-        || source.status != STPSourceStatusPending
-        || source.redirect.returnURL == nil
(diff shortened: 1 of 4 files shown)


Stripe/STPRedirectContext.m
```diff
@@ -194,30 +245,23 @@ - (void)handleRedirectCompletionWithError:(nullable NSError *)error
 }
 
 - (void)subscribeToUrlAndForegroundNotifications {
-    [[NSNotificationCenter defaultCenter] addObserver:self
-                                             selector:@selector(handleWillForegroundNotification)
-                                                 name:UIApplicationWillEnterForegroundNotification
-                                               object:nil];
-    [[STPURLCallbackHandler shared] registerListener:self
-                                              forURL:self.source.redirect.returnURL];
+    [[NSNotificationCenter defaultCenter] addObserver:self selector:@selector(handleApplicationWillEnterForegroundNotification) name:UIApplicationWillEnterForegroundNotification object:nil];
```
bdorfman-stripe: Why did you go through and remove the line breaks from like, every method call in all these files? I find it much harder to read like this.

Stripe/STPURLCallbackHandler.m
```diff
@@ -32,7 +32,7 @@ + (BOOL)handleStripeURLCallbackWithURL:(NSURL *)url FAUXPAS_IGNORED_ON_LINE(Unus
 
 @interface STPURLCallback : NSObject
 @property (nonatomic) NSURLComponents *urlComponents;
-@property (nonatomic) id<STPURLCallbackListener> listener;
+@property (nonatomic, weak) id<STPURLCallbackListener> listener;
```
bdorfman-stripe: Is it potentially actually better to just leave this as it was and let the context retain itself for the duration of the redirect or until a cancel occurs? Or do you think that behavior is too unexpected.

bdorfman-stripe:
Note: The retain cycle issues have been fixed on their own in https://github.com/stripe/stripe-ios/pull/846

Issue #807, opened 2017-09-29
Feature request: Access to Card metadata from STPCard


davidsantiago:
To make it easier to diagnose your issue, please fill out the following:

## Summary

It's currently inconvenient to access the metadata on an STPCard, as there's no accessor on the class as far as we can tell. This metadata is stored and accessible when using another API (we do this with Java on the backend). 

## Code to reproduce

No code for this. 

## iOS version

iOS 10/11

## Installation method

Via Cocoapods

## SDK version

11.1.0

## Other information

We'd be glad to submit code for this, if that's agreeable to you.

joeydong-stripe:
Hi @davidsantiago, thanks for opening the issue.

That definitely sounds like a good idea! We will take a look.

In the meantime, if you haven't already done so, you should be able to read the `metadata` like so:

```
STPCard *card = ...;

NSDictionary *metadata = card.allResponseFields[@"metadata"];
```

davidsantiago:
Thanks so much!

PR #808, opened 2017-09-29
Add `metadata` accessors

joeydong-stripe:
## Summary

Add `metadata` accessors to `STPBankAccount` and `STPCard`

The data is populated on these two objects when working with customers fetched using ephemeral keys.

## Motivation

Following up from https://github.com/stripe/stripe-ios/issues/807

## Testing

Updated unit tests. Also added extra guards to ensure the metadata keys/values are strings.

### DIFF
--- a/Stripe/NSDictionary+Stripe.h
+++ b/Stripe/NSDictionary+Stripe.h
@@ -16,6 +16,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (NSDictionary *)stp_dictionaryByRemovingNulls;
 
+- (NSDictionary<NSString *, NSString *> *)stp_dictionaryByRemovingNonStrings;
+
 @end
 
 NS_ASSUME_NONNULL_END
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 11.x.x 2017-XX-XX
 
 * Restores `[STPCard brandFromString:]` method which was marked as deprecated in a recent version [#801](https://github.com/stripe/stripe-ios/pull/801)
+* Adds `[STPBankAccount metadata]` and `[STPCard metadata]` read-only accessors and improves annotation for `[STPSource metadata]` [#808](https://github.com/stripe/stripe-ios/pull/808)
 
 ## 11.3.0 2017-09-13
 * Adds support for creating `STPSourceParams` for P24 source [#779](https://github.com/stripe/stripe-ios/pull/779)
--- a/Stripe/NSDictionary+Stripe.m
+++ b/Stripe/NSDictionary+Stripe.m
@@ -10,6 +10,8 @@
 
 #import "NSArray+Stripe.h"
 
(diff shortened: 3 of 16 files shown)


Stripe/NSDictionary+Stripe.h
```diff
@@ -16,6 +16,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (NSDictionary *)stp_dictionaryByRemovingNulls;
 
+- (NSDictionary<NSString *, NSString *> *)stp_dictionaryByRemovingNonStrings;
```
bdorfman-stripe: I don't think we should do this. We don't currently enforce this type for other classes like STPSource and I don't think we should change that or add the restriction here. It is not clear to me from the API docs that they must in fact be string pairs (as opposed to having number, dictionary, or array values). I think we should just leave as `NSDictionary *` or at most `NSDictionary<NSString *, id> *`

bg-stripe: Hmm, I'm actually for this. It is an API constraint, and though the docs aren't super clear we do say:

> Note: You can specify up to 20 keys, with key names up to 40 characters long and values up to 500 characters long.

[0] https://stripe.com/docs/api#metadata

bdorfman-stripe: Oh interesting. I didn't realize we specified that. In that case I'm ok with changing our API to match this constraint 👍

bg-stripe:
Just did a run through of our docs – looks like bank account, PII, and file creations don't support metadata, but you can attach metadata to a card. Should we also add a `metadata` property to `STPCardParams`? 

https://stripe.com/docs/api#token_object-card-metadata

bdorfman-stripe:
Hmm, it looks like bank accounts do support metadata to me[0]. We should probably add to card params and bank account params, and forward through the metadata on `[STPSourceParams cardParamsWithCard:]`

[0] https://stripe.com/docs/api#customer_create_bank_account

bg-stripe:
oh hmm. I was looking at this page [0], which I think is different from the "attach a bank account source to a customer" API. Definitely possible that metadata on bank account tokenization is supported and just not documented – we should test it out.

[0] https://stripe.com/docs/api#create_bank_account_token

bdorfman-stripe:
Yeah I'm not sure. It's actually super unclear to me how this works, as the token API does not mention metadata at all (including for cards) but the Customer docs say that both cards and bank accounts fetched from them can have metadata (but how would you get those details on without having added them to the token?). Probably need to sync with someone to sort out how this works and get the docs updated.

bdorfman-stripe:
I guess actually the docs imply that metadata only is addable if you add the bank account/card _directly_ to the customer and not the "make a token then add that token" method, which seems odd and confusing. Let's sync offline on this.

bdorfman-stripe:
After further discussion, since metadata really only makes sense for / is settable by Customer-related endpoints which are generally not handled by the SDK (because you need a private key) we decided it's fine to just add read-only support for them to the models and not to add them into the params objects.

PR #810, opened 2017-10-09
Support saved card sources in UI elements

bdorfman-stripe:
This copies over a small subset of the `sources-ui` branch to allow for reading saved source objects from customer and using them in `STPPaymentMethodsViewController`.

It also fixes snapshot tests for that class which appear to have been broken and erroneously passing for quite a while.


Stripe/STPPaymentMethodTuple.m
```diff
@@ -27,22 +27,23 @@ + (instancetype)tupleWithPaymentMethods:(NSArray<id<STPPaymentMethod>> *)payment
...
 }
 
-+ (instancetype)tupleWithCardTuple:(STPCardTuple *)cardTuple
-                   applePayEnabled:(BOOL)applePayEnabled {
-    NSMutableArray *paymentMethods = [NSMutableArray array];
-    for (STPCard *card in cardTuple.cards) {
-        [paymentMethods addObject:card];
-    }
++ (instancetype)tupleWithPaymentMethods:(NSArray<id<STPPaymentMethod>> *)paymentMethods
+                  selectedPaymentMethod:(id<STPPaymentMethod>)selectedPaymentMethod
```
bg-stripe: nit: can we wrap this file with `NS_ASSUME_NONNULL`s and annotate this as `nullable`?

bdorfman-stripe: Yep I'll make the change.

Tests/Tests/STPPaymentMethodsViewControllerTest.m
```diff
@@ -0,0 +1,172 @@
...
+ */
+- (void)testInitWithSingleCardTokenSourceAndCardAvailable {
+    STPCustomer *customer = [STPFixtures customerWithSingleCardTokenSource];
+    STPPaymentConfiguration *config = [STPFixtures paymentConfiguration];
+    config.additionalPaymentMethods = STPPaymentMethodTypeAll;
+    id<STPPaymentMethodsViewControllerDelegate>delegate = OCMProtocolMock(@protocol(STPPaymentMethodsViewControllerDelegate));
+    STPPaymentMethodsViewController *sut = [self buildViewControllerWithCustomer:customer
+                                                                   configuration:config
+                                                                        delegate:delegate];
+    XCTAssertTrue([sut.internalViewController isKindOfClass:[STPPaymentMethodsInternalViewController class]]);
```
bg-stripe: can we also check that the internal `paymentMethods` and `selectedPaymentMethod` props are set to the expected values? 

alternatively, we could add some tests for `filteredSourceTupleForUIWithConfiguration`, and avoid exposing those internal properties to tests.

bdorfman-stripe: I'll just add some tests for filter 👍

bdorfman-stripe:
Ready for re-review I think. Also added some fixtures code that should help our Customer testing a bunch going forward.

Tests/Tests/STPCustomerSourceTupleTest.m
```diff
@@ -0,0 +1,148 @@
...
+                                                                           STPJSONKeySourceBitcoin]
+                                                           defaultSource:STPJSONKeySourceiDEAL];
+    [self performSourceTupleTestWithCustomer:customer
+                        expectedValidSources:0
+                      expectedSelectedSource:nil];
+}
+
+
+- (void)testSourceTupleCreationMixedValidAndInvalidSourcesWithInvalidDefaultCustomer {
+    STPCustomer *customer = [STPFixtures customerWithSourcesFromJSONKeys:@[STPJSONKeySource3DS,
```
bg-stripe: nice, these tests are really clean ✨

Tests/Tests/STPFixtures.m
```diff
@@ -10,6 +10,17 @@
 #import "STPTestUtils.h"
 #import "STPEphemeralKey.h"
 
+NSString *const STPJSONKeyCustomer = @"Customer";
```
bg-stripe: nit: not sure about `STPJSONKey` as a name, since it kinda looks like these are constants from the SDK. Maybe something like `STPTestJSON`?

bdorfman-stripe: Good idea. I wasn't sure what to call them. As I was working I renamed all of them 3-4 times before pushing this :P

bdorfman-stripe:
Constants renamed. Also fixed description of one of the test methods which I realized was slightly incorrect.

Issue #818, opened 2017-10-29
Shipping method dismissal prevents payment completion.


nikolas7892:
The only problem is that the payment cannot be completed after the successful dismissal (See the issue below). Every time I try to complete the purchase, I receive the shipping dialogue..

I previously opened an issue for this subject, although the solution produced an error that doesn't not allow the user to complete the checkout process. 
https://github.com/stripe/stripe-ios/issues/817

The following code will produce the problem:

` func paymentContext(_ paymentContext: STPPaymentContext, didUpdateShippingAddress address: STPAddress, completion: @Escaping STPShippingMethodsCompletionBlock) {

    guard let countryAddress = address.country else { return }
    guard let postalCodeAddress = address.postalCode else { return }
    guard let cityAddress = address.city else { return }
    guard let nameAddress = address.name else { return }
    guard let streetLineAddress = address.line1 else { return }
    guard let apartmentLineAddress = address.line2 else { return }
    guard let stateAddress = address.state else { return }
    guard let phoneAddress = address.phone else { return }
    
    shippingCountry = countryAddress
    shippingZipCode = postalCodeAddress
    shippingCity = cityAddress
    shippingName = nameAddress
    shippingStreetLine = streetLineAddress
    shippingApartmentLine = apartmentLineAddress
    shippingState = stateAddress
    shippingPhone = phoneAddress
    completion(.valid, nil, nil, nil)
  
}`

bg-stripe:
Ah, you're totally right, sorry about that! I was mistaken about my suggestion. You should instead not implement the `paymentContext:didUpdateShippingAddress` method (which is optional, and should only be used if you need to collect a shipping method – I'll make sure this is documented better). 

To get the shipping information entered by the user from `PaymentContext`, you should instead use the `STPPaymentContext` instance's `shippingAddress` property after the payment completes, e.g. in `paymentContext:didCreatePaymentResult:`, where you could send the shipping information to your backend along with other information about the order.

PR #821, opened 2017-10-30
Add comment to example app about optional shipping methods delegate

bg-stripe:
r? @bdorfman-stripe

Addresses some confusion in #818. This is currently documented in the header, but not clear in the example app.


bg-stripe:
Yeah, updating the API so that calling `completion` with empty or nil shipping methods doesn't show present the shipping methods VC is probably a good idea – I was confused and actually thought that was the contract until I read the docs.

PR #823, opened 2017-11-01
Disable availability warnings.

bdorfman-stripe:
This warning only knows about the new `@available` objc language feature (that matches the Swift one). All the cases it catches are already blocked by other methods, like selector testing, so it's not actually a problem. We can't add `@available` calls until we drop Xcode 8 support so disabling temporarily seems the best solution.


Tests/installation_tests/cocoapods/with_frameworks/Podfile
```diff
@@ -1,6 +1,7 @@
 use_frameworks!
 
 target 'CocoapodsTest' do
+	platform :ios, '8.0'
```
bg-stripe: 👍

bdorfman-stripe: Yeah I noticed our cocoapods test was throwing warnings because we didn't have this specified. Probably need to add it to the without_frameworks version also. I have a task on me to change a bunch of this stuff when we drop Xcode 8 / iOS 8 so I'll do it then.

Issue #826, opened 2017-11-03
Crash (unrecognized selector sent to instance) in 11.4.0 when deserialize STPCustomer


andproff:
To make it easier to diagnose your issue, please fill out the following:

## Summary

After update to 11.4.0 we got next problem:
When the SDK attempts to deserialize STPCustomer with sources, app even crash.

```
-[__NSArray0 stp_dictionaryByRemovingNonStrings]: unrecognized selector sent to instance 0x60c00000c650
*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[__NSArray0 stp_dictionaryByRemovingNonStrings]: unrecognized selector sent to instance 0x60c00000c650'
```
## Code to reproduce

Actually problem in your code:
STPCard.m line 201:
`card.metadata = [dict[@"metadata"] stp_dictionaryByRemovingNonStrings];` that means you try to work with the `NSDictionary`
but your API return `metadata = ( );` that means metadata is an Array

## iOS version

Found in iOS11 but I think this not matter

## Installation method

Cocoapods

## SDK version

11.4.0

## Other information

Anything else you can include that'll make it easier for us to help you!

bdorfman-stripe:
Hi @andproff,

Sorry you're having trouble. Can you send us an example of the raw JSON you are seeing? As far as we can tell from our end the Stripe API should always return a dictionary value for metadata so we are trying to track down what is happening for you. 

Also, how are you getting the data you are feeding into the customer parsing code? Are you using ephemeral key / STPCustomerContext or do you have a custom backend api adapter?

Thanks,
Brian

andproff:
Hello Brian
> Can you send us an example of the raw JSON you are seeing?

find below raw JSON
```
{
  "status": "ok",
  "stripeCustomer": {
    "id": "cus_B5D9SdDuEfqLkm",
    "object": "customer",
    "account_balance": 0,
    "created": 1500887973,
    "currency": null,
    "default_source": "card_1BGSkxHGJFW1N235uGxYRAA3",
    "delinquent": false,
    "description": "Leonardo DiCaprio",
    "discount": null,
    "email": null,
    "livemode": false,
    "metadata": {
      "UUID": "8C636BAD-82EE-4252-AEDE-8A304D164A2F"
    },
    "shipping": null,
    "sources": {
      "object": "list",
      "data": [
        {
          "id": "card_1BGSkxHGJFW1N235uGxYRAA3",
          "object": "card",
[...]

PR #827, opened 2017-11-03
Hotfix for crash (unrecognized selector sent to instance) (Updated STPCard.m)

andproff:
## Summary

Hotfix for crash (unrecognized selector sent to instance) in 11.4.0 when deserialize STPCustomer #826

## Motivation

Bug described in issue #826

## Testing

Was tested in real project.


bdorfman-stripe:
The issue referenced in this PR turned out to be a bug in merchant code, and not in the SDK. We are internally tracking adding SDK-wide json type checks to all of our code as a future feature to add.

PR #833, opened 2017-11-08
Fix redirect context dismissing SafariVC on foreground

bdorfman-stripe:
When using SafariVC don't dismiss on foreground notification, just on url received (or when the user manually closes). Continue listening for foreground notification when actually redirecting out of the app.

Also fixes a bunch of redirect test code. See commit desc for more details.

### DIFF
--- a/Cartfile.private
+++ b/Cartfile.private
@@ -1,1 +1,2 @@
 github "facebook/ios-snapshot-test-case"
+github "erikdoe/ocmock"
--- a/Tests/Tests/STPRedirectContextTest.m
+++ b/Tests/Tests/STPRedirectContextTest.m
@@ -13,6 +13,7 @@
 #import "STPFixtures.h"
 #import "STPRedirectContext.h"
 #import "STPURLCallbackHandler.h"
+#import "STPWeakStrongMacros.h"
 
 @interface STPRedirectContext (Testing)
 - (void)unsubscribeFromNotifications;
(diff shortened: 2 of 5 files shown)


Cartfile.private
```diff
@@ -1 +1,2 @@
 github "facebook/ios-snapshot-test-case"
+github "erikdoe/ocmock"
```
bg-stripe: oh weird, this line disappeared?

Tests/Tests/STPRedirectContextTest.m
```diff
@@ -13,18 +13,42 @@
...
+
+/**
+ Use this to unsubscribe a context from notifications without calling
+ `sut.unsubscrbeFromNotifications` if you have OCMReject'd that method and thus
+ can't call it.
+
+ Note: You MUST pass in the actual context object here and not the mock or the
+ unsubscibe will silently fail.
+ */
+- (void)unsubscribeContext:(STPRedirectContext *)context {
```
bg-stripe: nice

Tests/Tests/STPRedirectContextTest.m
```diff
@@ -13,18 +13,42 @@
...
 
 @interface STPRedirectContextTest : XCTestCase
-
+@property (nonatomic, weak) STPRedirectContext *weak_sut;
 @end
 
+/*
+ NOTE:
+
+ If you are adding a test unsubscribe from notifications is called before your
```
bg-stripe: maybe "If you are adding a test use `unsubscribeContext:` to unsubscribe the context being tested before the test ends"

Cartfile.private
bdorfman-stripe: Yeah I don't know what happened. It was still working with bootstrap cause it was in Cartfile.resolved

Tests/Tests/STPRedirectContextTest.m
bdorfman-stripe: Ah yeah I also accidentally a word here. Should be "make sure unsubscribe from notifications is called". You theoretically don't need to manually call it if your test is already verifying it is called (Which many do). You can also just directly call unsubscribe on it if you aren't explicitly OCMRejecting that method instead of using `unsubscribeContext:`. I'll rewrite this comment to be more clear.

bdorfman-stripe:
Comment updated. Will merge when ci passes again

Issue #831, opened 2017-11-07
Remove "Use Billing" button


odedharth:
## Summary

At the payment method view controller, we are not asking for a billing address. 
Nevertheless, In the standard shipping view controller, it shows the "Use Billing" button. When clicking on the "Use billing" button - the button simply disappears and nothing happens. 

## Code to reproduce

Setup the config like that:
```
        let config = STPPaymentConfiguration.shared()
        config.requiredBillingAddressFields = .none
        config.requiredShippingAddressFields = [.postalAddress, .phone]
        config.shippingType = .shipping

```

## iOS version

11

## Installation method

Github

## SDK version

11.4.0

danj-stripe:
Thanks for the bug report. I'm able to reproduce the bug, and we'll be looking into a fix for it.

danj-stripe:
@odedharth We've just released [11.5.0](https://github.com/stripe/stripe-ios/releases/tag/v11.5.0), which should fix this issue. (If it doesn't, please let us know!)

odedharth:
@danj-stripe yep, it's fixed now, thanks!

yonilevy:
This seems to still be happening in some scenarios. E.g. if I call `[paymentContext presentShippingViewController]` twice, the "Use Billing" button will show up on the 2nd open. Also if the shipping VC is presented by a call to `[paymentContext requestPayment]`. Using iOS 12.2, Stripe SDK 15.0.1 (on an iPad if it matters).

yonilevy:
btw from a little debugging, the button seems to show up because the `billingAddress` returns true to `hasPartialPostalAddress`, because its country is set to `"US"`. Not sure if it's a test-card behavior, but I'd say it's a bit unexpected.

PR #834, opened 2017-11-08
Adding STPAddress `containsContentFor...Fields:` methods to check for any useful data

danj-stripe:
## Summary

These are like an `isEmpty` check for the address, but they check the fields needed
for a particular billing or shipping address setting.

## Motivation

This addresses #831, which was introduced when
`STPShippingAddressViewController.billingAddress` started receiving an empty STPAddress
from STPCard instead of a nil one.

## Testing

Unit tests, and manual testing through the standard integration app.

## Other Notes

I'm not in love with the method names.

I had trouble deciding on the semantics of `containsContentFor...Fields: .None` (should it
return true or false?). I chose to always return NO. If the question is "Does this address
have useful data for me, and I'm interested in none of the data from this address?", then
false/NO makes sense. It also parallels with the behavior when messaging nil.

### DIFF
--- a/Stripe/PublicHeaders/STPAddress.h
+++ b/Stripe/PublicHeaders/STPAddress.h
@@ -182,7 +182,7 @@ typedef NS_ENUM(NSUInteger, STPBillingAddressFields) {
  @parameter desiredFields The billing address information the caller is interested in.
  @return YES if there is any data in this STPAddress that's relevant for those fields.
  */
-- (BOOL)containsContentForFields:(STPBillingAddressFields)desiredFields;
+- (BOOL)containsContentForBillingAddressFields:(STPBillingAddressFields)desiredFields;
 
 /**
  Checks if this STPAddress has the level of valid address information
--- a/Stripe/PublicHeaders/STPAddress.h
+++ b/Stripe/PublicHeaders/STPAddress.h
@@ -167,6 +167,24 @@ typedef NS_ENUM(NSUInteger, STPBillingAddressFields) {
 - (BOOL)containsRequiredFields:(STPBillingAddressFields)requiredFields;
 
 /**
+ Checks if this STPAddress has any content (possibly invalid) in any of the
+ desired billing address fields.
+
+ Where `containsRequiredFields:` validates that this STPAddress contains valid data in
+ all of the required fields, this method checks for the existence of *any* data.
+
+ For example, if `desiredFields` is `STPBillingAddressFieldsZip`, this will check
+ if the postalCode is empty.
+
+ Note: When `desiredFields == STPBillingAddressFieldsNone`, this method always returns
+ NO.
+
+ @parameter desiredFields The billing address information the caller is interested in.
+ @return YES if there is any data in this STPAddress that's relevant for those fields.
+ */
+- (BOOL)containsContentForFields:(STPBillingAddressFields)desiredFields;
+
+/**
  Checks if this STPAddress has the level of valid address information
  required by the passed in setting.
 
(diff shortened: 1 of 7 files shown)


Stripe/STPAddCardViewController.m
```diff
@@ -145,8 +145,11 @@ - (void)createAndSetupViews {
     }
     [addressHeaderView.button addTarget:self action:@selector(useShippingAddress:)
                        forControlEvents:UIControlEventTouchUpInside];
-    BOOL needsAddress = self.configuration.requiredBillingAddressFields != STPBillingAddressFieldsNone && !self.addressViewModel.isValid;
-    BOOL buttonVisible = (needsAddress && self.shippingAddress != nil && !self.hasUsedShippingAddress);
+    BOOL requiredFields = self.configuration.requiredBillingAddressFields;
+    BOOL needsAddress = requiredFields != STPBillingAddressFieldsNone && !self.addressViewModel.isValid;
+    BOOL buttonVisible = (needsAddress &&
```
danj-stripe: There wasn't actually a bug here (shippingAddress was still nil), but I chose to make a parallel change to keep things consistent and guard against future changes.

Stripe/PublicHeaders/STPAddress.h
```diff
@@ -167,6 +167,24 @@ typedef NS_ENUM(NSUInteger, STPBillingAddressFields) {
...
+ For example, if `desiredFields` is `STPBillingAddressFieldsZip`, this will check
+ if the postalCode is empty.
+
+ Note: When `desiredFields == STPBillingAddressFieldsNone`, this method always returns
+ NO.
+
+ @parameter desiredFields The billing address information the caller is interested in.
+ @return YES if there is any data in this STPAddress that's relevant for those fields.
+ */
+- (BOOL)containsContentForFields:(STPBillingAddressFields)desiredFields;
```
bdorfman-stripe: I think this should be `billingAddressFields`. It doesn't match the other one (which was written before billing address existed) but I think new methods going forward should make clear whether they are doing billing or shipping fields and at some point we can deprecate/rename the old ones. This is what we did for the newer Apple Pay billing address related method further down.

Stripe/STPShippingAddressViewController.m
```diff
@@ -137,8 +137,11 @@ - (void)createAndSetupViews {
                        forState:UIControlStateNormal];
     [headerView.button addTarget:self action:@selector(useBillingAddress:)
                 forControlEvents:UIControlEventTouchUpInside];
-    BOOL needsAddress = self.configuration.requiredShippingAddressFields & PKAddressFieldPostalAddress && !self.addressViewModel.isValid;
-    BOOL buttonVisible = (needsAddress && self.billingAddress != nil && !self.hasUsedBillingAddress);
+    PKAddressField requiredFields = self.configuration.requiredShippingAddressFields;
+    BOOL needsAddress = requiredFields & PKAddressFieldPostalAddress && !self.addressViewModel.isValid;
```
bdorfman-stripe: For clarity I think we should put parens around the expression that is a bitwise AND here because it's easy to misinterpret what's happening.

Tests/Tests/STPAddressTests.m
```diff
@@ -458,6 +503,81 @@ - (void)testContainsRequiredShippingAddressFields {
...
+    XCTAssertFalse([address containsContentForShippingAddressFields:PKAddressFieldNone]);
+    XCTAssertFalse([address containsContentForShippingAddressFields:PKAddressFieldName]);
+    XCTAssertFalse([address containsContentForShippingAddressFields:PKAddressFieldPhone]);
+    XCTAssertTrue([address containsContentForShippingAddressFields:PKAddressFieldEmail]);
+    XCTAssertFalse([address containsContentForShippingAddressFields:PKAddressFieldPostalAddress]);
+    address.email = @"";
+
+    // Test every property that contributes to the full address
+    // This is *not* refactoring-safe, but I think it's better than a bunch more duplicated code
+    for (NSString *propertyName in @[@"line1", @"line2", @"city", @"state", @"postalCode", @"country"]) {
```
bdorfman-stripe: Maybe do something like `NSStringFromSelector(@selector(line1))` instead of `@"line1"` for some more compiler safety?

PR #846, opened 2017-12-08
Fix an STPURLCallbackListener retain cycle.

bdorfman-stripe:
Previously there was a bug where STPURLCallbackListener retained its listeners, which meant that STPRedirectContext had a retain cycle and would not deallocate while it was actively listening for callbacks. Fixing this is in some ways a breaking API change.
This commit fixes the bug, and also updates the docs to explicitly mention that you need to retain the instances you make.
It also replaces referencing to STPAPIClient source polling (which is deprecated) and instead links to the sources best practices doc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## 12.0.0 2017-12-XX
+* `STPRedirectContext` will no longer retain itself for the duration of the redirect, you must explicitly maintain a reference to it yourself. If you 
+
 ## 11.5.0 2017-11-09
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## 12.0.0 2017-12-XX
+* `STPRedirectContext` will no longer retain itself for the duration of the redirect, you must explicitly maintain a reference to it yourself. If you
```
bg-stripe: incomplete :P

bdorfman-stripe: Oh weird. I typed more stuff locally into Atom and it seems to not have saved the file before I comitted -_- Will fix

PR #848, opened 2017-12-09
Replace PKAddressField with new STPContactField

bdorfman-stripe:
Convert STPContactField sets to either PKAddressField bitmasks or PKContactField sets depending on runtime iOS version.


Stripe/PublicHeaders/STPAddress.h
```diff
@@ -221,6 +247,26 @@ typedef NS_ENUM(NSUInteger, STPBillingAddressFields) {
...
+
+/**
+ Converts a set of STPContactField values into the closest equivalent
+ representation of PKContactField options
+
+ @param contactFields Stripe contact fields values to convert.
+ @return The closest represenation of the contact fields as
+ a PKContactField value.
+ */
++ (NSSet<PKContactField> *)pkContactFieldsFromStripeContactFields:(nullable NSSet<STPContactField> *)contactFields API_AVAILABLE(ios(11.0));
```
bg-stripe: looks like this `API_AVAILABLE` guard isn't working – I think it needs to be `API_AVAILABLE(ios(11))`

Stripe/STPAddress.m
```diff
@@ -347,6 +354,48 @@ + (PKAddressField)applePayAddressFieldsFromBillingAddressFields:(STPBillingAddre
...
+    for (STPContactField contactField in contactFields) {
+        NSNumber *boxedConvertedField = contactToAddressFieldMap[contactField];
+        if (boxedConvertedField) {
+            addressFields = (PKAddressField) (addressFields | [boxedConvertedField unsignedIntegerValue]);
+        }
+    }
+    return addressFields;
+}
+
++ (NSSet<PKContactField> *)pkContactFieldsFromStripeContactFields:(NSSet<STPContactField> *)contactFields API_AVAILABLE(ios(11.0)) {
```
bg-stripe: same as above, `API_AVAILABLE(ios(11))`

Stripe/PublicHeaders/STPAddress.h
```diff
@@ -221,6 +247,26 @@ typedef NS_ENUM(NSUInteger, STPBillingAddressFields) {
...
+ a PKAddressField value.
+ */
++ (PKAddressField)pkAddressFieldsFromStripeContactFields:(nullable NSSet<STPContactField> *)contactFields;
+
+/**
+ Converts a set of STPContactField values into the closest equivalent
+ representation of PKContactField options
+
+ @param contactFields Stripe contact fields values to convert.
+ @return The closest represenation of the contact fields as
```
bg-stripe: typo: `represenation`

bdorfman-stripe:
Cool. I thiiink we are going to disable faux pas availability checking and switch over to Xcode 9's `@available` feature, but I'll fix this so that the build will pass for now.

Stripe/STPAddress.m
bdorfman-stripe: Hmm, the documentation for `API_AVAILABLE` seems to say you should do it as `11.0` and not `11`. Maybe a bug in faux pas? I also notice we do `NS_AVAILABLE_IOS` everywhere else, so maybe I'll just switch to that.

bdorfman-stripe:
Oh I just realized this won't compile on Xcode 8.x  (which is what travis is using) because I added iOS 11+ APIs and it doesn't have that sdk. Going to hold this PR until I make/merge Xcode9+ changes.

bdorfman-stripe:
Updated with Xcode9/iOS9 changes

Issue #830, opened 2017-11-04
largeTitleDisplayMode on shipping and payments controller 


ReeceHumphreys:
**## Summary**

There is no method to override the default behavior on the shipping and payments controllers navigationItem to support largeTitleDisplayMode for iOS 11

**## Code to reproduce**

N/A

**## iOS version**

iOS 11.1

**## Installation method**

Cocoapods

**## SDK version**

11.4.0

danj-stripe:
Hi Reece,

Thanks for opening this issue. We'd like to fix this for you, but I need to know more about it. Can you provide some more details about how you're integrating with the SDK, and what you're trying to accomplish?

Are you using `STPPaymentContext` with the `present...` methods or the `push...` methods? Or are you using the Custom Integration Guide, and directly instantiating the VCs?

Thanks!
Dan

ReeceHumphreys:
Hi Dan,

I am currently utilizing the push method on STPPaymentContext. It is inheriting the current largeTitleDisplayMode behavior from the previous controller which is resulting in some layout issues on the Simulator. I have attached a screenshot of the simulator that demonstrates this issue.
It was produced on the iPhone X simulator running iOS 11.1.

As you can see in the screenshot there appears to be a large amount of space between the top of the box image and the navigation bar. I attempted to resolve this issue by setting largeTitleDisplayMode = .never on the payment controller and shipping controller but there is no way to access the navigationItem when presenting these controllers using STPPaymentContext.

![simulator screen shot - iphone x - 2017-11-11 at 16 50 22](https://user-images.githubusercontent.com/14209515/32693895-f3dd8354-c700-11e7-8218-68ca36516571.png)

bg-stripe:
Hey @ReeceHumphreys – thanks for the additional info. We'll try to get a fix out in an upcoming release.

PR #849, opened 2017-12-11
add largeTitleDisplayMode to PaymentContext

bg-stripe:
r? @bdorfman-stripe 

`PaymentContext` will set any pushed or presented VC's `largeTitleDisplayMode` to its own. 

re: #830 – I wasn't able to reproduce the layout issue, but this does fix not being able to set `largeTitleDisplayMode` of pushed PaymentContext view controllers to `never`.


Stripe/STPPaymentContext.m
```diff
@@ -292,9 +293,11 @@ - (void)presentPaymentMethodsViewControllerWithNewState:(STPPaymentContextState)
             paymentMethodsViewController.prefilledInformation = self.prefilledInformation;
             paymentMethodsViewController.paymentMethodsViewControllerFooterView = self.paymentMethodsViewControllerFooterView;
             paymentMethodsViewController.addCardViewControllerFooterView = self.addCardViewControllerFooterView;
+            paymentMethodsViewController.navigationItem.largeTitleDisplayMode = self.largeTitleDisplayMode;
 
             UINavigationController *navigationController = [[UINavigationController alloc] initWithRootViewController:paymentMethodsViewController];
             navigationController.navigationBar.stp_theme = self.theme;
+            navigationController.navigationBar.prefersLargeTitles = YES;
```
bg-stripe: Based on the docs, it looks like it isn't necessary to expose this property on `PaymentContext` too – just setting it to always `YES` always seems fine.

> When this property is set to true, the navigation bar allows the title to be displayed out-of-line and using a larger font. The navigation item used to build the bar must specify whether it wants its title displayed in the large or small format. Use the largeTitleDisplayMode property to configure the title's appearance.

> When the property is set to false, the navigation bar displays the title inline with the other bar button items.

Example/Standard Integration (Swift)/BrowseProductsViewController.swift
```diff
@@ -35,6 +35,7 @@ class BrowseProductsViewController: UITableViewController {
         super.viewDidLoad()
         self.navigationItem.title = "Emoji Apparel"
         self.navigationController?.navigationBar.isTranslucent = false
+        self.navigationController?.view.backgroundColor = .white
```
bg-stripe: This fixes the transition when `PaymentContext` pushes a small title view controller onto a large title nav controller (there's a black flash without this line).

Stripe/PublicHeaders/STPPaymentContext.h
```diff
@@ -233,6 +233,14 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic, assign) UIModalPresentationStyle modalPresentationStyle;
 
+/**
+ The mode to use when displaying the title of the navigation bar in all view
+ controllers presented by the context. The default value is `automatic`,
+ which causes the title to use the same styling as the previously displayed
+ navigation item.
+ */
+@property (nonatomic, assign) UINavigationItemLargeTitleDisplayMode largeTitleDisplayMode;
```
bdorfman-stripe: Might want to consider adding in some extra notes about how this property interacts with push vs present. Eg. the default of `automatic` won't work for present because there is no previous item. And for push we don't alter the owning navigation bar, you have to make sure it's navigation bar has `prefersLargeTitles = YES` in order for this to do anything.

bg-stripe: ah, good call – I'll expand

bg-stripe:
> Do we need to update our tests to cover this? Either regular unit tests or snapshot tests? (I made a bunch of snapshot test updates in the xcode9 branch if you want to wait for that to be merged).

Ah yeah, now's probably a good time to add some PaymentContext snapshot tests. I'll wait and update the PR after #851 gets in

bg-stripe:
I've updated docs, added availability guards, and added some snapshot tests (b10520a) for PaymentContext pushing payment methods / shipping onto a hostViewController.

re-r? @bdorfman-stripe

Tests/Tests/FBSnapshotTestCase+STPViewControllerLoading.h
```diff
@@ -9,5 +9,18 @@
 #import <FBSnapshotTestCase/FBSnapshotTestCase.h>
 
 @interface FBSnapshotTestCase (STPViewControllerLoading)
-- (UIView *)stp_preparedAndSizedViewForSnapshotTestFromViewController:(UIViewController *)viewController;
```
bdorfman-stripe: Might be nice to retain this, or a similarly named method, that encompasses both the other helpers (since everything except Pay Context immediately calls both in a row I think?)

bg-stripe: ah yeah, that's a good call, I'll update

Issue #839, opened 2017-11-21
Incomplete workflow when using STPShippingTypeDelivery as shipping type


flexaddicted:
## Summary

I'm using the standard integration example. If I set the shipping type as `STPShippingTypeDelivery`, the payment context does not complete correctly (i.e. when `paymentContext.requestPayment()` is invoked through the buy button). After having inserted the details of the card, a delivery view is presented. On tapping Done the delivery controller is presented again.

## Code to reproduce

Set the delivery type using Settings view. Choose a product to buy. Tap Buy button. Insert the card details. Tap Done. Insert delivery details. Tap Done. Delivery details are presented again.

## iOS version

iOS 11

## Installation method

Carthage and Dynamic Framework

## SDK version

11.5.0

danj-stripe:
@flexaddicted You're right, that's a bug. We'll try to get a fix into an upcoming version.

If you need a workaround in the mean time, deleting the implementation of `paymentContext:didUpdateShippingAddress:completion:` in the `STPPaymentContextDelegate` should work.

In our standard integration sample code, that's [implemented here](https://github.com/stripe/stripe-ios/blob/243da135f73840c781011c92610197be5ca41ffc/Example/Standard%20Integration%20(Swift)/CheckoutViewController.swift#L250-L252).

Issue #817, opened 2017-10-26
Dismiss shipping without shipping method


nikolas7892:
To make it easier to diagnose your issue, please fill out the following:

## Summary

I'm trying to (in Swift) simply record the user shipping address without the shipping method.

## Code to reproduce
I've attempted to dismiss and return on the "didUpdateShippingAddress" method.

` func paymentContext(_ paymentContext: STPPaymentContext, didUpdateShippingAddress address: STPAddress, completion: @escaping STPShippingMethodsCompletionBlock) {
        
        guard let countryAddress = address.country else { return }
        guard let postalCodeAddress = address.postalCode else { return }
        guard let cityAddress = address.city else { return }
        guard let nameAddress = address.name else { return }
        guard let streetLineAddress = address.line1 else { return }
        guard let apartmentLineAddress = address.line2 else { return }
        guard let stateAddress = address.state else { return }
        guard let phoneAddress = address.phone else { return }
        
        shippingCountry = countryAddress
        shippingZipCode = postalCodeAddress
        shippingCity = cityAddress
        shippingName = nameAddress
        shippingStreetLine = streetLineAddress
        shippingApartmentLine = apartmentLineAddress
        shippingState = stateAddress
        shippingPhone = phoneAddress
        return
      
    }`

or 

` func paymentContext(_ paymentContext: STPPaymentContext, didUpdateShippingAddress address: STPAddress, completion: @escaping STPShippingMethodsCompletionBlock) {
        
        guard let countryAddress = address.country else { return }
        guard let postalCodeAddress = address.postalCode else { return }
        guard let cityAddress = address.city else { return }
        guard let nameAddress = address.name else { return }
        guard let streetLineAddress = address.line1 else { return }
        guard let apartmentLineAddress = address.line2 else { return }
        guard let stateAddress = address.state else { return }
        guard let phoneAddress = address.phone else { return }
        
        shippingCountry = countryAddress
        shippingZipCode = postalCodeAddress
        shippingCity = cityAddress
        shippingName = nameAddress
        shippingStreetLine = streetLineAddress
        shippingApartmentLine = apartmentLineAddress
        shippingState = stateAddress
        shippingPhone = phoneAddress
        completion(.valid, nil, [], nil)
      
    }`

## iOS version

iOS11

## Installation method

cocoapods

bg-stripe:
Hey @nikolas7892 ,

[...]

PR #850, opened 2017-12-11
fix bug where PaymentContext gets stuck with shippingType Delivery

bg-stripe:
r? @danj-stripe 

This fixes the bug reported in #839 – shipping method shouldn't be required if `shippingType` is `delivery`.

### DIFF
--- a/Stripe/PublicHeaders/STPPaymentContext.h
+++ b/Stripe/PublicHeaders/STPPaymentContext.h
@@ -370,9 +370,10 @@ didCreatePaymentResult:(STPPaymentResult *)paymentResult
  Inside this method, you should verify that you can ship to the given address.
  You should call the completion block with the results of your validation
  and the available shipping methods for the given address. If you don't implement
- this method or call the completion block with nil or an empty array of shipping
- method, the user won't be prompted to select a shipping method and all
- addresses will be valid.
+ this method, the user won't be prompted to select a shipping method and all
+ addresses will be valid. If you call the completion block with nil or an
+ empty array of shipping methods, the user won't be prompted to select a
+ shipping method.
 
  @note If a user updates their shipping address within the Apple Pay dialog,
  this address will be anonymized. For example, in the US, it will only include the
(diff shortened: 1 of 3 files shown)


bg-stripe:
Update – after looking at #817, I've changed the fix. `requestPayment` actually shouldn't decide whether to present `shippingAddressVC` based on whether shipping methods are required. 

`ShippingAddressVC` already has logic to determine whether or not to push a shipping methods VC, based on whether the delegate method is implemented / if nil or empty shipping methods are passed to the completion block. Since you can only collect a shipping method if you first collect an address, there's no need for the extra check in `PaymentContext`. 

I've updated some documentation in `PaymentContext` to make it clearer that you can pass `nil` or empty shipping methods to skip collecting a shipping method.

Stripe/STPPaymentContext.m
```diff
@@ -510,9 +510,8 @@ - (void)appropriatelyDismissViewController:(UIViewController *)viewController
 - (BOOL)requestPaymentShouldPresentShippingViewController {
     BOOL shippingAddressRequired = self.configuration.requiredShippingAddressFields != STPBillingAddressFieldsNone;
     BOOL shippingAddressIncomplete = ![self.shippingAddress containsRequiredShippingAddressFields:self.configuration.requiredShippingAddressFields];
-    BOOL shippingMethodRequired = ([self.delegate respondsToSelector:@selector(paymentContext:didUpdateShippingAddress:completion:)] && !self.selectedShippingMethod);
     BOOL verificationRequired = self.configuration.verifyPrefilledShippingAddress && self.shippingAddressNeedsVerification;
-    return (shippingAddressRequired && (shippingAddressIncomplete || shippingMethodRequired || verificationRequired));
+    return (shippingAddressRequired && (shippingAddressIncomplete || verificationRequired));
```
danj-stripe: I think if they need the user to select a shipping method, turn off `verifyPrefilledShippingAddress` **and** pre-fill the shipping address, this will incorrectly skip selecting a shipping method.

Stripe/PublicHeaders/STPPaymentContext.h
```diff
@@ -370,7 +370,8 @@ didCreatePaymentResult:(STPPaymentResult *)paymentResult
  Inside this method, you should verify that you can ship to the given address.
  You should call the completion block with the results of your validation
  and the available shipping methods for the given address. If you don't implement
- this method, the user won't be prompted to select a shipping method and all
+ this method or call the completion block with nil or an empty array of shipping
+ method, the user won't be prompted to select a shipping method and all
```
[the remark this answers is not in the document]
bg-stripe: 60bed6a2

Stripe/STPPaymentContext.m
bg-stripe: ah, good catch! updated to handle that case in `601d2f71d6`

bg-stripe:
re-r? @danj-stripe

PR #851, opened 2017-12-12
Change minimum requirements to Xcode9 / iOS 9

bdorfman-stripe:
This is a pretty monster change because so many things needed to change to get it working, but I have tried to split each commit up into a logical chunk so you may want to review them one at a time.

cc @danj-stripe (in case you also wanted to skim the changes).

Summary of changes:
* Change all our settings to min ios version 9
* Remove all AddressBook framework references in favor of Contacts framework
* Turn FauxPas back on for static framework only (it seems to only be broken for checking dynamic frameworks and this gives us some linting at least)
* Replace all `respondsToSelector` checks and FauxPas `APIAvailability` rule with Xcode9's `@available` function and warning.
* Replace out custom scroll view insetting code with the system default (the custom code was broken on iOS 11 and there didn't appear to be a clear reason for not using the default behavior).
* Cleanup snapshot test logic. Made a new helper for preparing a VC for being snapshotted which works reliably on iOS11 and made everything use that.


Stripe/STPCoreScrollViewController.m
```diff
@@ -46,21 +54,4 @@ - (void)updateAppearance {
     }
 }
 
-- (void)viewWillAppear:(BOOL)animated {
```
bg-stripe: does removing this break things on iOS < 11?

MIGRATING.md
```diff
@@ -1,5 +1,9 @@
 ## Migration Guides
 
+### Migrating from versions < 12.0.0
+* The SDK now requires iOS 9+ andXcode version is now 9+. If you need to support iOS 8 or Xcode 8, the last supported version is [11.5.0](https://github.com/stripe/stripe-ios/releases/tag/v11.5.0)
```
bg-stripe: typo: `andXcode`

Stripe/STPCoreScrollViewController.m
bg-stripe: cool, sgtm – thanks for the explanation!

PR #854, opened 2017-12-12
iPhone X support

bg-stripe:
r? @bdorfman-stripe 

- Added a UIView category method that returns bounds adjusted to `safeAreaInsets`
- Updated cells and `STPPaymentCardTextField` to use adjusted bounds
- Updated cells to set `backgroundColor`, not `contentView.backgroundColor`. I think it looks better when our cells are full width on iPhone X (contentView is inset to the safe area).
- Updated the example app to use `safeAreaInsets`

simulator screen shot - iphone x - 2017-12-12 at 17 57 20

simulator screen shot - iphone x - 2017-12-12 at 18 00 14


Stripe/STPAddCardViewController.m
```diff
@@ -408,8 +408,7 @@ - (UITableViewCell *)tableView:(__unused UITableView *)tableView
         default:
             return [UITableViewCell new]; // won't be called; exists to make the static analyzer happy
     }
-    cell.backgroundColor = [UIColor clearColor];
-    cell.contentView.backgroundColor = self.theme.secondaryBackgroundColor;
+    cell.backgroundColor = self.theme.secondaryBackgroundColor;
```
bdorfman-stripe: If you don't set contentView background is it clear? Do you need to set both cell background and content view background to the theme color?

Stripe/UIView+Stripe_SafeAreaBounds.m
```diff
@@ -0,0 +1,38 @@
...
+NS_ASSUME_NONNULL_BEGIN
+
+@implementation UIView (Stripe_SafeAreaBounds)
+
+- (CGRect)boundsWithHorizontalSafeAreaInsets {
+    if (@available(iOS 11.0, *)) {
+        UIView *view = self;
+        UIEdgeInsets insets = UIEdgeInsetsZero;
+        while (view.superview != nil || UIEdgeInsetsEqualToEdgeInsets(insets, UIEdgeInsetsZero)) {
+            view = view.superview;
```
bdorfman-stripe: This seems like not how `safeAreaInsets` is meant to be used. Is there some other reason the insets aren't being propagated to subviews like described in the docs?

Stripe/STPAddCardViewController.m
bg-stripe: It seems like `contentView.backgroundColor` defaults to clear, but probably best to set it explicitly just it's different on other iOS versions. I'll update.

Stripe/UIView+Stripe_SafeAreaBounds.m
bg-stripe: Ah whoops, yeah this isn't necessary – I was confused by the docs at first, and thought it didn't propagate.

bg-stripe:
updated – re-r? @bdorfman-stripe

Stripe/UIView+Stripe_SafeAreaBounds.h
```diff
@@ -0,0 +1,26 @@
...
+//
+
+#import <UIKit/UIKit.h>
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface UIView (Stripe_SafeAreaBounds)
+
+/**
+ Finds the first non-zero `safeAreaInsets` in the view hierarchy, and returns
```
bdorfman-stripe: This comment is wrong now

bg-stripe: ah yep, will fix

bdorfman-stripe:
Also maybe add a note to the changelog that this was fixed?

bg-stripe:
👍  updated the comment for that method and added a CHANGELOG entry.

I don't think we need to worry about `viewSafeAreaInsetsDidChange` – whenever orientation changes we also get a layoutSubviews call.

PR #855, opened 2017-12-13
Fix bugs around `STPPaymentCardTextField becomeFirstResponder`

### LINKED ISSUE #841: STPPaymentCardTextField.becomeFirstResponder not working when STPPaymentCardTextField filled out
## Summary

When STPPaymentCardTextField is completely filled out and is not the first responder, trying to set it as first responder doesn't work.

## Code to reproduce

```
import Stripe
import UIKit

class ViewController: UIViewController {

    @IBOutlet weak var paymentCardTextField: STPPaymentCardTextField!
    
    @IBAction func editButtonTapped(_ sender: Any) {
        paymentCardTextField.becomeFirstResponder()
    }
    
    @IBAction func doneButtonTapped(_ sender: Any) {
        paymentCardTextField.resignFirstResponder()
    }
}
```

## iOS version

iOS 11

## Installation method

Pod

## SDK version

11.5.0

## Other information

Repro steps:
1. Fill out STPPaymentCardTextField
2. Tap Done button
3. Tap Edit button
4. STPPaymentCardTextField doesn't become the first responder. The user should be able to edit the payment card.

danj-stripe:
## Summary

Previous implementation of `nextFirstResponderField` would return `nil` if all fields were
valid, which blocked programatic calls to `becomeFirstResponder`.

It *also* used to cycle through the fields every time it was called, which doesn't seem
right. It feels more natural for the default first responder to be either the first
invalid field *or* the last field. This puts the insertion point at the right place
to start deleting.

We *do* want to move to the next field (ignoring whether fields are valid or not) when the user fills the current field with valid text. This matches the web elements behavior, and feels better as they proceed through, instead of jumping around (possibly earlier, possibly to the end)

## Motivation

Fixes #841

## Testing

Manual testing through the Standard Integration. I also added a unit test for this
behavior. Some of the things it verifies:

* becomeFirstResponder should not change the first responder if one of the STPPaymentCardTextField subfields is already first responder
* becomeFirstResponder should take the user to the first invalid field when called
* becomeFirstResponder should make the last sub-field first responder when all the fields are valid

### DIFF
--- a/Stripe/STPPaymentCardTextField.m
+++ b/Stripe/STPPaymentCardTextField.m
@@ -447,7 +447,7 @@ CGFloat const STPPaymentCardTextFieldMinimumPadding = 10;
 #pragma mark UIResponder & related methods
 
 - (BOOL)isFirstResponder {
-    return [self.currentFirstResponderField isFirstResponder];
+    return self.currentFirstResponderField != nil;
 }
 
 - (BOOL)canBecomeFirstResponder {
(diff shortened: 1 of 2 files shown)


bdorfman-stripe:
We sync'ed offline on this. Dan is going to make some tweaks.

PR #856, opened 2017-12-13
Fix versions in readme

bdorfman-stripe:
Missed this in the other Xcode9/iOS9 pull request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ If you're reading this on GitHub.com, please make sure you are looking at the [t
 
 ## Requirements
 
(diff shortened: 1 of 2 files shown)


danj-stripe:
I think you can also delete some docs from [STPRedirectContext.h#L122-L126](https://github.com/stripe/stripe-ios/blob/86bc2fa46f5a820ca862bc66f5da1a7d0918bf13/Stripe/PublicHeaders/STPRedirectContext.h#L122-L126)

bdorfman-stripe:
Oh good catch on redirect docs!

danj-stripe:
👍

PR #861, opened 2017-12-15
Fix comparison between localized file names and preferred localization

KevinMorais:
## Summary

Fix comparison between localized file names and preferred localization

## Motivation

Trying to find a real job (startups are hard man...)

## Testing

Tested in real case (production environment)

### DIFF
--- a/Stripe/STPLocalizationUtils.m
+++ b/Stripe/STPLocalizationUtils.m
@@ -35,7 +35,7 @@
     
     static dispatch_once_t onceToken;
     dispatch_once(&onceToken, ^{
-        if (![[STPBundleLocator stripeResourcesBundle].preferredLocalizations.firstObject isEqualToString:[NSBundle mainBundle].preferredLocalizations.firstObject]) {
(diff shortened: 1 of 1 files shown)


bdorfman-stripe:
Thanks for the submission, but I don't think this behavior change is correct. You are changing from comparing our localizations to the system locale, instead of comparing to the localizations supported by the containing app (which is the behavior we want).

Do you have some example code or project where this feature as it exists is not behaving correctly?

KevinMorais:
Hi Brian,

I’m not able to send you our project as it’s an app available on the Store.

We have this issue when we want to show errors from you to our users (we used error.localizedDescription). I understood what you try to do but the compared values were « fr » for you and « fr_FR » from the app bundle. As all the localized languages in your bundle only have « fr », « en » …etc, this comparison cannot work.

Let me know if it’s not enough clear for you, I can print out the variables values of what I got with your initial code.

Kevin

bg-stripe:
Hi @KevinMorais – thanks for the bug report and submission. We'll look into this a bit further and try to get a fix into an upcoming version.

PR #863, opened 2017-12-15
Always recreate loadingPromise when calling retryLoading

bg-stripe:
r? @danj-stripe 

Calling `retryLoading` currently does nothing if the promise has been completed with a value. The method was initially intended for use when PaymentContext's initial load fails, but you would also expect it to be useful for triggering a refresh of the stored customer object:

```
customerContext.clearCachedCustomer()
paymentContext.retryLoading()
paymentContext.presentPaymentMethodsViewController()
```

I can't think of any unintended side effects from this change. The method is only called once internally, in the `STPPaymentContext` initialization path. Apps that use this method to recover from initial loading errors won't be affected by the change.


bg-stripe:
note from IRL discussion: I'll update this so that `retryLoading` also clears the cached customer on `CustomerContext`

Issue #842, opened 2017-11-28
Request: Please re-enable enumeration of Apple Pay sources


davidsantiago:
To make it easier to diagnose your issue, please fill out the following:

## Summary

Previous to v11.4.0, the STPCustomer object would give you all the customer's cards and sources. With #810 in v11.4.0, the STPCustomer's deserializer filters out ApplePay cards and sources. This makes things very difficult for our custom integration, which is not using STPPaymentContext as its payment UI. We use a custom integration that works through STPCustomerContext because it works better for our app, which is more of a subscription service and we think our payment UI is better designed for our app's needs.

This change presents a problem for us because now we can't show the user which payment source they are using for their subscription if it is an ApplePay card. We can't even just present "ApplePay" in the UI, because it appears that the user has no default source set when their default source is an ApplePay card for subscription, so we can't even tell that that is the case. 

This is a particularly frustrating limitation, because the API itself is not filtering this at the backend. Subscription payments are also a [supported transaction type/business model for Apple Pay](https://developer.apple.com/apple-pay/get-started/), so this seems like something we should be able to do. We could recreate the client code with changes, or write our own code to hit that endpoint with the ephemeral key, or just fork the SDK and modify those lines. But of course, we'd rather just use the stock SDK and not have to maintain those things ourselves. 

Given that this change was surely made for good reason, it would be great if we could change the "filter ApplePay sources" feature to be optional in some way, perhaps as a flag to the `retrieveCustomer()` function, or with a second function that does not do the filtering. 

Thanks.

## Code to reproduce

```swift
customerContext.retrieveCustomer { customer, error in 
  let customerSources = customer.sources
  for source in customerSources {
    // customerSources has sources except the ApplePay ones. 

    if source.stripeID == customer.defaultSource?.stripeID {
      // Nope, this customer's default source in the dashboard is an ApplePay card, but
      // this if statement will never be entered. 
    }
  }
}
```

## iOS version

iOS 11, but almost certainly on earlier versions as well. 

## Installation method

CocoaPods

## SDK version

v11.5.0 (v11.4.0 also has the issue).

## Other information

Anything else you can include that'll make it easier for us to help you!

bg-stripe:
Hey @davidsantiago , thanks for the feedback! We'll try to get a change into the next release that allows disabling CustomerContext's Apple Pay filtering.

PR #864, opened 2017-12-18
add includeApplePaySources to CustomerContext

bg-stripe:
r? @danj-stripe 

Addresses #842 by adding a new `includeApplePaySources` property to `CustomerContext`

### DIFF
--- a/Stripe/PublicHeaders/STPCustomerContext.h
+++ b/Stripe/PublicHeaders/STPCustomerContext.h
@@ -48,8 +48,12 @@ NS_ASSUME_NONNULL_BEGIN
 
 /**
  By default, `STPCustomerContext` will filter Apple Pay sources when it retrieves
- a Customer object. If you are using `STPCustomerContext` to back your own UI and
- would like to disable Apple Pay filtering, set this property to YES.
+ a Customer object. Apple Pay sources should generally not be re-used and
+ shouldn't be offered to customers as a new payment source (Apple Pay sources may
+ only be re-used for subscriptions).
+
+ If you are using `STPCustomerContext` to back your own UI and would like to
+ disable Apple Pay filtering, set this property to YES.
 
  Note: If you are using `STPPaymentContext`, you should not change this property.
  */
--- a/Stripe/PublicHeaders/STPCustomerContext.h
+++ b/Stripe/PublicHeaders/STPCustomerContext.h
@@ -46,6 +46,15 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (void)clearCachedCustomer;
 
+/**
+ By default, `STPCustomerContext` will filter Apple Pay sources when it retrieves
+ a Customer object. If you are using `STPCustomerContext` to back your own UI and
+ would like to disable Apple Pay filtering, set this property to YES.
+
+ Note: If you are using `STPPaymentContext`, you should not change this property.
+ */
+@property (nonatomic, assign) BOOL includeApplePaySources;
+
 @end
 
 NS_ASSUME_NONNULL_END
--- a/Stripe/STPCustomer+Private.h
+++ b/Stripe/STPCustomer+Private.h
@@ -1,0 +1,26 @@
+//
+//  STPCustomer+Private.h
+//  Stripe
+//
+//  Created by Ben Guo on 12/18/17.
+//  Copyright © 2017 Stripe, Inc. All rights reserved.
+//
+
+#import "STPCustomer.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface STPCustomer ()
+
+/**
+ Updates the customer's sources and defaultSource with a Customer API response.
+
+ @param response            The Customer API response
+ @param filterApplePay      If YES, Apple Pay sources will be ignored
+ */
+- (void)updateSourcesWithResponse:(NSDictionary *)response
+                filteringApplePay:(BOOL)filterApplePay;
+
+@end
+
+NS_ASSUME_NONNULL_END
--- a/Stripe/STPCustomer+Private.h
+++ b/Stripe/STPCustomer+Private.h
@@ -13,7 +13,8 @@ NS_ASSUME_NONNULL_BEGIN
 @interface STPCustomer ()
 
 /**
- Updates the customer's sources and defaultSource with a Customer API response.
+ Replaces the customer's sources and defaultSource with the contents of a
+ Customer API response.
 
  @param response            The Customer API response
  @param filterApplePay      If YES, Apple Pay sources will be ignored
--- a/Stripe/STPCustomer.m
+++ b/Stripe/STPCustomer.m
@@ -18,11 +18,11 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface STPCustomer()
 
-@property (nonatomic, copy) NSString *stripeID;
-@property (nonatomic) id<STPSourceProtocol> defaultSource;
-@property (nonatomic) NSArray<id<STPSourceProtocol>> *sources;
-@property (nonatomic) STPAddress *shippingAddress;
-@property (nonatomic, readwrite, nonnull, copy) NSDictionary *allResponseFields;
+@property (nonatomic, copy, readwrite) NSString *stripeID;
+@property (nonatomic, strong, nullable, readwrite) id<STPSourceProtocol> defaultSource;
+@property (nonatomic, strong, readwrite) NSArray<id<STPSourceProtocol>> *sources;
+@property (nonatomic, strong, nullable, readwrite) STPAddress *shippingAddress;
+@property (nonatomic, copy, readwrite) NSDictionary *allResponseFields;
 
 @end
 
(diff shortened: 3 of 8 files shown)


Stripe/STPCustomerContext.m
```diff
@@ -48,6 +49,10 @@ - (void)clearCachedCustomer {
 }
 
 - (void)setCustomer:(STPCustomer *)customer {
+    if (self.includeApplePaySources) {
+        [customer updateSourcesWithResponse:customer.allResponseFields
```
danj-stripe: This is kind of funky/subtle/non-obvious.

* This mutates the `STPCustomer` object passed into the setter, which the caller still has a reference to.
* I didn't understand why you changed `customer` -> `self.customer` on L92 below.
  * I'm guessing because `setCustomer:` has side effects, but because they're just references to the same object, I don't think that change actually changes the behavior?

I see that this is a private property, and I think there're only three places that call the setter and none of them use the customer afterwords, so it's fine as-is (although maybe a well-placed & well-worded comment could help?).

I *think* it might be easier to understand & maintain if this was (duplicated and) moved up a level in the call stack to those places that receive `STPCustomer` objects from `STPAPI`. In each of them it's more clear that the `STPCustomerContext` completely owns the object & mutating it is fine, as well as not having (potentially) surprising behavior "hidden" in the setter.

Tests/Tests/STPCustomerContextTest.m
```diff
@@ -304,4 +304,45 @@ - (void)testDetachSourceFromCustomerCallsAPIClientCorrectly {
...
+- (void)testIncludeApplePaySources {
+    STPEphemeralKey *customerKey = [STPFixtures ephemeralKey];
+    STPCustomer *expectedCustomer = [STPFixtures customerWithCardAndApplePaySources];
+    [self stubRetrieveCustomerUsingKey:customerKey
+                     returningCustomer:expectedCustomer
+                         expectedCount:2];
+    id mockKeyManager = [self mockKeyManagerWithKey:customerKey];
+    STPCustomerContext *sut = [[STPCustomerContext alloc] initWithKeyManager:mockKeyManager];
+    sut.includeApplePaySources = YES;
+    [sut clearCachedCustomer];
```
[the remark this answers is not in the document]
danj-stripe: Or maybe it just calls the `updateSourcesWithResponse:filteringApplePay:` method, instead of hitting the server again?

Stripe/STPCustomer.m
```diff
@@ -85,43 +83,60 @@ + (instancetype)decodedObjectFromAPIResponse:(NSDictionary *)response {
...
+        return;
+    }
+    NSString *defaultSourceId;
+    if ([response[@"default_source"] isKindOfClass:[NSString class]]) {
+        defaultSourceId = response[@"default_source"];
+    }
+    NSMutableArray *sources = [NSMutableArray new];
+    for (id contents in data) {
+        if ([contents isKindOfClass:[NSDictionary class]]) {
+            if ([contents[@"object"] isEqualToString:@"card"]) {
```
danj-stripe: I recognize this is just the old code, but this line makes an assumption about the type of `contents[@"object]"` by using an `NSString`-only method (`isEqualToString:`), where the parsing code so far has been diligently checking types.

Maybe -> `isEqual:@"card"`

danj-stripe: side note: reading this method makes me want `as?`, `flatMap`, `filter` & `find` :)

Stripe/STPCustomer.m
```diff
@@ -85,43 +83,60 @@ + (instancetype)decodedObjectFromAPIResponse:(NSDictionary *)response {
...
+- (void)updateSourcesWithResponse:(NSDictionary *)response
+                filteringApplePay:(BOOL)filterApplePay {
+    NSArray *data;
+    if ([response[@"sources"] isKindOfClass:[NSDictionary class]]) {
+        data = response[@"sources"][@"data"];
+    }
+    if (![data isKindOfClass:[NSArray class]]) {
+        return;
+    }
+    NSString *defaultSourceId;
```
danj-stripe: I think you should probably reset `self.defaultSource = nil`.

ex: calling with `filterApplePay = NO` and then with `YES`, for a customer whose default source is ApplePay, will still have the apple pay `defaultSource` on the customer object.

Stripe/STPCustomer+Private.h
```diff
@@ -0,0 +1,26 @@
...
+//
+
+#import "STPCustomer.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface STPCustomer ()
+
+/**
+ Updates the customer's sources and defaultSource with a Customer API response.
```
danj-stripe: I wonder if it's worth changing this from "Updates" to "Replaces", since you're getting all new objects?

Stripe/PublicHeaders/STPCustomerContext.h
```diff
@@ -46,6 +46,15 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (void)clearCachedCustomer;
 
+/**
+ By default, `STPCustomerContext` will filter Apple Pay sources when it retrieves
+ a Customer object. If you are using `STPCustomerContext` to back your own UI and
```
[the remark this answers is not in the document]
bg-stripe: Ah yep, I'll add a note.

Stripe/STPCustomer+Private.h
bg-stripe: 👍

Stripe/STPCustomer.m
bg-stripe: ah yep, good catch!

Stripe/STPCustomerContext.m
bg-stripe: Ah yeah that's good feedback, putting it in the setter was kinda lazy. I'll move it up in the stack.

Tests/Tests/STPCustomerContextTest.m
```diff
@@ -304,4 +304,45 @@ - (void)testDetachSourceFromCustomerCallsAPIClientCorrectly {
...
+
+- (void)testFiltersApplePaySourcesByDefault {
+    STPEphemeralKey *customerKey = [STPFixtures ephemeralKey];
+    STPCustomer *expectedCustomer = [STPFixtures customerWithCardAndApplePaySources];
+    [self stubRetrieveCustomerUsingKey:customerKey
+                     returningCustomer:expectedCustomer
+                         expectedCount:2];
+    id mockKeyManager = [self mockKeyManagerWithKey:customerKey];
+    STPCustomerContext *sut = [[STPCustomerContext alloc] initWithKeyManager:mockKeyManager];
+    [sut clearCachedCustomer];
```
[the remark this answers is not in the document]
bg-stripe: ah yeah, I'll add a comment explaining the `2`. CustomerContext fetches a customer on initialization, but there's no callback we can use for assertions, so I just used another `retrieveCustomer` call.

Tests/Tests/STPCustomerContextTest.m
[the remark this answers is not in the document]
bg-stripe: ah interesting, it does seem reasonable for `includeApplePaySources` to update the current customer too. I'll update.

Tests/Tests/STPCustomerContextTest.m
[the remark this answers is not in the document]
bg-stripe: Oh actually, after changing the behavior of `includeApplePaySources` per below, we can just use the cached customer and change this to `1`

bg-stripe:
updated, re-r? @danj-stripe

Stripe/STPCustomer.m
```diff
@@ -85,43 +83,61 @@ + (instancetype)decodedObjectFromAPIResponse:(NSDictionary *)response {
...
+}
+
+- (void)updateSourcesWithResponse:(NSDictionary *)response
+                filteringApplePay:(BOOL)filterApplePay {
+    NSArray *data;
+    if ([response[@"sources"] isKindOfClass:[NSDictionary class]]) {
+        data = response[@"sources"][@"data"];
+    }
+    if (![data isKindOfClass:[NSArray class]]) {
+        return;
```
danj-stripe: When called from `decodedObjectFromAPIResponse:`, this will return a customer with `sources = nil` (if `sources.data` is not an `NSArray`), but the property declaration for `sources` is `nonnull`

I wasn't (and still am not) sure how we want to handle a malformed `response`. I like that we're not crashing! :)

I also thought I liked the idea that you're not resetting `sources` and `defaultSources` if this is not the initialization pass. However, this is private, and it looks like you're always just using the cached response, so maybe it'd be easier to just reset to empty array & nil?

Another option would be to set sources to the empty array in `decodedObjectFromAPIResponse:` before calling this method. I think I like this option slightly better.

bg-stripe: ah good catch, I agree that the second option is better – I'll update. Working on some related guards against malformed responses in my next PR :)

bg-stripe: updated in ffb5ff6

PR #867, opened 2017-12-19
Convert all Address Fields (except phone number) to use `STPValidatedTextField` instead of `STPFormTextField`

danj-stripe:
## Summary

Create `SPTValidatedTextField` and extract default/error/placeholder colors & `validText`
from `STPFormTextField`. This is now the `UITextField` superclass for the SDK, so this
also includes the bugfix for `editingRect` (#860)

Use `UIControlEventEditingChanged` instead of `STPFormTextFieldDelegate` to detect changes
to the address fields.

Change zip/postal code text field to enable selection and preserve contents on paste.

This *removes* several `STPFormTextField`-specific behaviors from these fields:

* Remove 'Go to previous field on delete' behavior.
* The implementation of `textField:shouldChangeCharactersInRange:replacementString:` 
that always returns NO, and uses `setText:` to apply the changes from the user action.
None of these fields had an `autoformattingBehavior`, and they all allowed
selection, so this method was (mostly) trying to replicate standard UITextField behavior.
* Custom `setText:` override that turned the text into an attributed string
and called `setAttributedText:`
* Custom `setAttributedText:` override that fires `UIControlEventEditingChanged` when called.
* Disabled all `UIMenuController` actions except `paste:`

The other parts of `STPFormTextField` were already disabled or not applicable to these
address fields. Things like sanitization, formatting, `selectionEnabled`,
`preservesContentsOnPaste`, etc.

## Motivation

Primary motivation was eliminating `textField:shouldChangeCharactersInRange:replacementString:`
always returning `NO`, because that breaks Japanese (and related) auto-complete views.

This SO post is actually trying *to* disable the autocomplete, we want the opposite:
https://stackoverflow.com/questions/8293648/disable-the-kanji-auto-complete-henkan-for-japanese-keyboards-in-ios

Customer support issue IOS-588

## Testing

Manual testing in the Standard Integration sample app, with a variety of required shipping
and billing address settings. Ensuring that the expected changes to behavior have occurred:
old `STPFormTextField` behaviors were removed. Ensuring that address entry in US & non-US
countries still behaves correctly.

Automated tests continue to pass.

### DIFF
--- a/Stripe/STPAddressFieldTableViewCell.m
+++ b/Stripe/STPAddressFieldTableViewCell.m
@@ -17,7 +17,7 @@
 
 @interface STPAddressFieldTableViewCell() <UITextFieldDelegate, UIPickerViewDelegate, UIPickerViewDataSource>
 
-@property (nonatomic, weak) STPFormTextField *textField;
+@property (nonatomic, weak) STPValidatedTextField *textField;
 @property (nonatomic) UIToolbar *inputAccessoryToolbar;
 @property (nonatomic) UIPickerView *countryPickerView;
 @property (nonatomic, strong) NSArray *countryCodes;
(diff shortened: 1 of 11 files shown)


danj-stripe:
This may be easiest to review a commit at a time. You may also want to review the full contents of [STPFormTextField.m](https://github.com/stripe/stripe-ios/blob/danj/bugfix/separate-payment-and-address-text-fields/Stripe/STPFormTextField.m) to make sure there's not something it's doing that we *do* want the address fields to keep doing.

Any tips on further testing are also very welcome.

Stripe/STPAddressFieldTableViewCell.m
```diff
@@ -139,9 +139,7 @@ - (void)updateTextFieldsAndCaptions {
             } else {
                 self.textField.keyboardType = UIKeyboardTypeASCIICapable;
             }
-            
-            self.textField.preservesContentsOnPaste = NO;
```
bg-stripe: 👍 , this seems like better behavior

Stripe/STPValidatedTextField.h
```diff
@@ -0,0 +1,30 @@
...
+//
+
+#import <UIKit/UIKit.h>
+
+/**
+ A UITextField that changes the text color, based on the validity of
+ its contents.
+
+ This does *not* (currently?) have any logic or hooks for determining whether
+ the contents are valid, that must be done by something else.
```
bg-stripe: 🏆  always leaving things more documented than before :)

PR #869, opened 2017-12-19
Decoding improvements

bg-stripe:
r? @danj-stripe 

I've added getters to `NSDictionary+Stripe` and updated our models to use them.

I've also removed `requiredFields` – I liked the declarative style, but there's no type validation. For now, it seems simpler to just validate required fields in the `decode` method of each model.

### DIFF
--- a/Stripe/STPCard.m
+++ b/Stripe/STPCard.m
@@ -197,7 +197,8 @@ NS_ASSUME_NONNULL_BEGIN
 #pragma clang diagnostic ignored "-Wdeprecated"
     // This is only intended to be deprecated publicly.
     // When removed from public header, can remove these pragmas
-    card.funding = [self.class fundingFromString:dict[@"funding"]];
+    NSString *rawFunding = [dict stp_stringForKey:@"funding"];
+    card.funding = [self.class fundingFromString:rawFunding];
 #pragma clang diagnostic pop
 
     card.country = [dict stp_stringForKey:@"country"];
--- a/Stripe/STPCard.m
+++ b/Stripe/STPCard.m
@@ -169,24 +169,30 @@ NS_ASSUME_NONNULL_BEGIN
     return @"card";
 }
 
-+ (NSArray *)requiredFields {
-    return @[@"id", @"last4", @"brand", @"exp_month", @"exp_year"];
-}
-
 + (nullable instancetype)decodedObjectFromAPIResponse:(nullable NSDictionary *)response {
-    NSDictionary *dict = [response stp_dictionaryByRemovingNullsValidatingRequiredFields:[self requiredFields]];
+    NSDictionary *dict = [response stp_dictionaryByRemovingNulls];
     if (!dict) {
+        return nil;
+    }
+
+    // required fields
+    NSString *stripeId = [dict stp_stringForKey:@"id"];
+    NSString *last4 = [dict stp_stringForKey:@"last4"];
+    NSString *rawBrand = [dict stp_stringForKey:@"brand"];
+    NSNumber *rawExpMonth = [dict stp_numberForKey:@"exp_month"];
+    NSNumber *rawExpYear = [dict stp_numberForKey:@"exp_year"];
+    if (!stripeId || !last4 || !rawBrand || !rawExpMonth || !rawExpYear) {
         return nil;
     }
 
     STPCard *card = [self new];
     card.address = [STPAddress new];
 
-    card.stripeID = dict[@"id"];
-    card.name = dict[@"name"];
-    card.last4 = dict[@"last4"];
-    card.dynamicLast4 = dict[@"dynamic_last4"];
-    card.brand = [self.class brandFromString:dict[@"brand"]];
+    card.stripeID = stripeId;
+    card.name = [dict stp_stringForKey:@"name"];
+    card.last4 = last4;
+    card.dynamicLast4 = [dict stp_stringForKey:@"dynamic_last4"];
+    card.brand = [self.class brandFromString:rawBrand];
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wdeprecated"
     // This is only intended to be deprecated publicly.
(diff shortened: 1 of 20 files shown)


Tests/Tests/NSDictionary+StripeTest.m
```diff
@@ -164,4 +164,82 @@ - (void)test_dictionaryByRemovingNonStrings_returnsImmutableCopy {
...
+                           @"f": @"10.5",
+                           @"g": @(10.0),
+                           @"h": @(10.5),
+                           @"i": @"foo",
+                           };
+
+    XCTAssertEqual([dict stp_intForKey:@"a" or:0], 1);
+    XCTAssertEqual([dict stp_intForKey:@"b" or:0], -1);
+    XCTAssertEqual([dict stp_intForKey:@"c" or:0], 1);
+    XCTAssertEqual([dict stp_intForKey:@"e" or:0], 10);
```
danj-stripe: Did you skip `@"d"` on purpose?

Stripe/STPCard.m
```diff
@@ -169,44 +169,50 @@ - (NSString *)stripeObject {
...
+    card.stripeID = stripeId;
+    card.name = [dict stp_stringForKey:@"name"];
+    card.last4 = last4;
+    card.dynamicLast4 = [dict stp_stringForKey:@"dynamic_last4"];
+    card.brand = [self.class brandFromString:rawBrand];
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wdeprecated"
     // This is only intended to be deprecated publicly.
     // When removed from public header, can remove these pragmas
     card.funding = [self.class fundingFromString:dict[@"funding"]];
```
danj-stripe: Is there a reason this isn't using `stp_stringForKey:`?

Stripe/STPCustomer.m
```diff
@@ -58,29 +58,34 @@ - (NSString *)description {
...
-        shipping.line1 = shippingDict[@"address"][@"line1"];
-        shipping.line2 = shippingDict[@"address"][@"line2"];
-        shipping.city = shippingDict[@"address"][@"city"];
-        shipping.state = shippingDict[@"address"][@"state"];
-        shipping.postalCode = shippingDict[@"address"][@"postal_code"];
-        shipping.country = shippingDict[@"address"][@"country"];
+        shipping.name = [shippingDict stp_stringForKey:@"name"];
+        shipping.phone = [shippingDict stp_stringForKey:@"phone"];
+        NSDictionary *addressDict = [shippingDict stp_dictionaryForKey:@"address"];
+        if (addressDict) {
```
danj-stripe: WDYT about refactoring/reordering this method to use `+[STPAddress decodedObjectFromAPIResponse:]` to extract the address lines? It's not totally straightforward because you have to handle the nil address dict case and still put name/phone into the shipping address, but...

Stripe/STPSource.m
```diff
@@ -214,34 +214,51 @@ - (NSString *)stripeObject {
...
-    source.verification = [STPSourceVerification decodedObjectFromAPIResponse:dict[@"verification"]];
-    source.details = dict[typeString];
+    source.stripeID = stripeId;
+    source.amount = [dict stp_numberForKey:@"amount"];
+    source.clientSecret = [dict stp_stringForKey:@"client_secret"];
+    source.created = [dict stp_dateForKey:@"created"];
+    source.currency = [dict stp_stringForKey:@"currency"];
+    NSString *rawFlow = [dict stp_stringForKey:@"flow"];
+    source.flow = [[self class] flowFromString:rawFlow];
+    source.livemode = [dict stp_boolForKey:@"livemode" or:NO];
```
danj-stripe: In `STPEphemeralKey` and `STPToken`, you coerce un-parseable `livemode` values to `YES`. Any reason for the discrepancy here?

Stripe/STPSourceOwner.m
```diff
@@ -29,26 +29,28 @@ @implementation STPSourceOwner
...
-    owner.verifiedPhone = dict[@"verified_phone"];
+    NSDictionary *rawAddress = [dict stp_dictionaryForKey:@"address"];
+    if (rawAddress) {
+        owner.address = [STPAddress decodedObjectFromAPIResponse:rawAddress];
+    }
+    owner.email = [dict stp_stringForKey:@"email"];
+    owner.name = [dict stp_stringForKey:@"name"];
+    owner.phone = [dict stp_stringForKey:@"phone"];
+    NSDictionary *rawVerifiedAddress = [dict stp_dictionaryForKey:@"verified_address"];
+    if (rawVerifiedAddress) {
```
danj-stripe: I believe all of the `decodedObjectFromAPIResponse:` methods gracefully handle `nil`.

Conversely, you don't check for non-`nil` strings before calling `fundingFromString:`, `brandFromString:`, etc. Those methods (so far) *also* gracefully handle `nil`.

I think this is totally fine as-is. I think I would slightly prefer *not* checking the dict against nil, because we have precedent that the decoding methods accept nil and it slightly reduces the amount of code.

I probably wouldn't change this (effort vs reward), but I will mention it in case there's something I'm missing or in case you feel more strongly about it.

Stripe/STPCard.m
bg-stripe: oh nope, good catch!

Stripe/STPCustomer.m
bg-stripe: ah, that's a good idea 👍

Stripe/STPSource.m
bg-stripe: hmm, i'm not sure what the best fallback is tbh (I think it's pretty unlikely for us to hit the default anyway). I'll make the default `YES` to be consistent.

Tests/Tests/NSDictionary+StripeTest.m
bg-stripe: whoops, good catch :P

Stripe/STPSourceOwner.m
bg-stripe: ah i think you're right. I can remove these checks

bg-stripe:
updated - re-r? @danj-stripe

Stripe/STPCustomer.m
```diff
@@ -78,13 +78,14 @@ + (nullable instancetype)decodedObjectFromAPIResponse:(nullable NSDictionary *)r
...
-            shipping.postalCode = [addressDict stp_stringForKey:@"postal_code"];
-            shipping.country = [addressDict stp_stringForKey:@"country"];
+        STPAddress *address = [STPAddress decodedObjectFromAPIResponse:addressDict];
+        if (address) {
+            shipping.line1 = address.line1;
+            shipping.line2 = address.line2;
+            shipping.city = address.city;
+            shipping.state = address.state;
+            shipping.postalCode = address.postalCode;
+            shipping.country = address.country;
```
danj-stripe: I was picturing something more like this:

```objective-c
if (shippingDict) {
    NSDictionary *addressDict = [shippingDict stp_dictionaryForKey:@"address"];
    STPAddress *shipping = [STPAddress decodedObjectFromAPIResponse:addressDict] ?: [STPAddress new];
    shipping.name = [shippingDict stp_stringForKey:@"name"];
    shipping.phone = [shippingDict stp_stringForKey:@"phone"];
    customer.shippingAddress = shipping;
}
```

This version keeps `addressDict` in the `allResponseFields` property, otherwise I think they're the same.

bg-stripe: oh derp yeah that's better, creating another address was silly

Issue #836, opened 2017-11-11
Add UITextContentType for any relevant text fields to support auto-fill on iOS 11


dmcgloin:
## Summary

Feature request: Add UITextContentType attribute for any relevant text fields to support auto-fill iOS 11.

Examples:

Address, City, State, Zip, Credit Card

## iOS version

iOS 11

## Installation method

UI Examples 

## SDK version

11.5.0

bg-stripe:
@dmcgloin Thanks for the feature request! We'll try to get this into an upcoming release.

PR #870, opened 2017-12-19
Add UITextContentType values to text fields provided by the SDK.

danj-stripe:
## Summary

This turns on QuickType suggestions for the name, email, and address fields. Also, using
a better keyboard for Payment Card fields.

For STPFormTextField, this was a lot harder than expected, due to *always* returning NO
from `textField:shouldChangeCharactersInRange:replacementString:`. There's a workaround
for the way UIKit behaves, which is only applicable for the Phone Number field right now.

Also using `UIKeyboardTypeASCIICapableNumberPad` in `STPPaymentCardTextField` if available

Per [WWDC 2016](http://asciiwwdc.com/2016/sessions/201), this was specifically added to
support things like credit card entry:

> We’ve also localized the keyboard number pads this year. [...]
> Now in some cases, you might want to specify ASCII capable number pad instead when you’re sure that the input that you have needs to be restricted to ASCII digits.
> Some examples of this are credit card numbers and IP addresses.

I think this is redundant for the Credit Card Number field (just setting the
`textContentType` is sufficient to use this keyboard), but it'll help w/the expiration
and CVC.

These text fields fall back to using .PhonePad, apparently to avoid localized & third
party keyboards (#358), and as far as I can tell `.ASCIICapableNumberPad` continues to
avoid them.

## Motivation

https://github.com/stripe/stripe-ios/issues/836

## Testing

Tested in the Standard Integration example app. Best tested on a device that knows an
address (simulator doesn't provide valid completions).

Both Shipping & Billing addresses are affected, and this also touches the Payment Card
text field.

Verified that enabling/using non-ASCII keyboards doesn't allow non-ASCII digits into
payment card fields.

Verified that 3rd party keyboards cannot be used for the credit card number entry.

### DIFF
--- a/Stripe/STPAddressFieldTableViewCell.m
+++ b/Stripe/STPAddressFieldTableViewCell.m
@@ -175,9 +175,7 @@
             break;
         case STPAddressFieldTypeCountry:
             self.textField.keyboardType = UIKeyboardTypeDefault;
-            if (@available(iOS 10.0, *)) {
-                self.textField.textContentType = UITextContentTypeCountryName;
-            }
+            // Don't set textContentType for Country, because we don't want iOS to skip the UIPickerView for input
             self.textField.inputView = self.countryPickerView;
             NSInteger index = [self.countryCodes indexOfObject:self.contents];
             if (index == NSNotFound) {
--- a/Stripe/STPAddressFieldTableViewCell.m
+++ b/Stripe/STPAddressFieldTableViewCell.m
@@ -127,24 +127,43 @@
     switch (self.type) {
         case STPAddressFieldTypeName: 
             self.textField.keyboardType = UIKeyboardTypeDefault;
+            if (@available(iOS 10.0, *)) {
+                self.textField.textContentType = UITextContentTypeName;
+            }
             break;
         case STPAddressFieldTypeLine1: 
             self.textField.keyboardType = UIKeyboardTypeDefault;
+            if (@available(iOS 10.0, *)) {
+                self.textField.textContentType = UITextContentTypeStreetAddressLine1;
+            }
             break;
         case STPAddressFieldTypeLine2: 
             self.textField.keyboardType = UIKeyboardTypeDefault;
+            if (@available(iOS 10.0, *)) {
+                self.textField.textContentType = UITextContentTypeStreetAddressLine2;
+            }
             break;
         case STPAddressFieldTypeCity:
             self.textField.keyboardType = UIKeyboardTypeDefault;
+            if (@available(iOS 10.0, *)) {
+                self.textField.textContentType = UITextContentTypeAddressCity;
+            }
             break;
         case STPAddressFieldTypeState:
             self.textField.keyboardType = UIKeyboardTypeDefault;
+            if (@available(iOS 10.0, *)) {
+                self.textField.textContentType = UITextContentTypeAddressState;
+            }
             break;
         case STPAddressFieldTypeZip:
             if ([self countryCodeIsUnitedStates]) { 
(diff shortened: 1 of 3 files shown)


danj-stripe:
Unfortunately, in English on my 11.2 test device, this puts a space after the inserted content. This breaks our validation for email & zip codes (as well as having unwanted spaces in the other fields).

I plan to submit a follow up PR that relaxes validation to allow those trailing spaces, and trims spaces from these fields.

danj-stripe:
This PR currently relies on the changes from #867, and has that branch as a base.

Stripe/STPAddressFieldTableViewCell.m
```diff
@@ -156,6 +175,9 @@ - (void)updateTextFieldsAndCaptions {
             break;
         case STPAddressFieldTypeCountry:
             self.textField.keyboardType = UIKeyboardTypeDefault;
+            if (@available(iOS 10.0, *)) {
+                self.textField.textContentType = UITextContentTypeCountryName;
```
bg-stripe: does this autofill the full name of the country? if so, i'm not sure if autofill will correctly update `contents` (which should be a country code).

Stripe/STPPaymentCardTextField.m
```diff
@@ -1106,7 +1114,12 @@ must be visible so they can be tapped over to (although
 - (STPFormTextField *)buildTextField {
     STPFormTextField *textField = [[STPFormTextField alloc] initWithFrame:CGRectZero];
     textField.backgroundColor = [UIColor clearColor];
-    textField.keyboardType = UIKeyboardTypePhonePad;
+    // setCountryCode: updates the postalCodeField keyboardType, this is safe
+    if (@available(iOS 10, *)) {
+        textField.keyboardType = UIKeyboardTypeASCIICapableNumberPad;
+    } else {
+        textField.keyboardType = UIKeyboardTypeNumberPad;
```
bg-stripe: should this be `PhonePad`?

Stripe/STPFormTextField.m
```diff
@@ -22,13 +22,42 @@ @interface STPTextFieldDelegateProxy : STPDelegateProxy<UITextFieldDelegate>
...
+        // This property is available starting in 10.0, but didn't offer in-app suggestions till 11.0
+        hasTextContentType = textField.textContentType != nil;
+    }
+
+    if (hasTextContentType && insertingIntoEmptyField && [string isEqualToString:@" "]) {
+        /* Observed behavior w/iOS 11.0 through 11.2.0 (latest):
+
+         1. UITextContentType suggestions are only available when textField is empty
+         2. When user taps a QuickType suggestion for the `textContentType`, UIKit *first*
+         calls this method with `range:{0, 0} replacementString:@" "`
```
bg-stripe: interesting, nice comment!

Stripe/STPAddressFieldTableViewCell.m
danj-stripe: Hmm. It does **not** provide quicktype suggestions (when tested on 11.2 and 11.0.1), and I think that's due to our use of the `UIPickerView` (which handles updating `contents`, etc).

This does seem potentially dangerous, because that *could* change in the future. I don't think this is adding anything, so I'm leaning toward removing it.

bg-stripe: Ah yeah, +1 for removing it then.

Stripe/STPPaymentCardTextField.m
danj-stripe: Yes, it should be, thanks for catching that!

PR #875, opened 2017-12-20
add stripeAccount to STPPaymentConfiguration

bg-stripe:
r? @danj-stripe 

Currently, there's no way to use PaymentContext/AddCardVC on behalf of a connected account.

IOS-597

### DIFF
--- a/Stripe/PublicHeaders/STPPaymentConfiguration.h
+++ b/Stripe/PublicHeaders/STPPaymentConfiguration.h
@@ -109,6 +109,15 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic, assign, readwrite) BOOL canDeletePaymentMethods;
 
+/**
+ In order to perform API requests on behalf of a connected account, e.g. to
+ create a source on a connected account, set this property to the ID of the
+ account for which this request is being made.
+
+ @see https://stripe.com/docs/connect/authentication#authentication-via-the-stripe-account-header
+ */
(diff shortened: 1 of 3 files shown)


Stripe/PublicHeaders/STPPaymentConfiguration.h
```diff
@@ -109,6 +109,15 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic, assign, readwrite) BOOL canDeletePaymentMethods;
 
+/**
+ In order to perform API requests on behalf of a connected account, e.g. to
+ create a source on a connected account, set this property to the ID of the
+ account for which this request is being made.
+
+ @see https://stripe.com/docs/connect/authentication#authentication-via-the-stripe-account-header
```
danj-stripe: That docs page says this authentication option works with the platform **secret** key (I'm nearly positive you can also use the platform publishable key too: DOCS-529).

Does the publishable key work? Should we update the docs?

fred-stripe: Yes, the publishable key does work—this is what Stripe.js does when you set the `stripeAccount` property: https://stripe.com/docs/stripe-js/reference#stripe-function

This is somewhat covered, since tokenization can & should be done with a publishable key. It's also discussed in the sources documentation: https://stripe.com/docs/sources/connect#creating-direct-charges

bg-stripe: thanks @fred-stripe ! :)

PR #876, opened 2017-12-21
[WIP] Connect Account tokenization

danj-stripe:
## Summary

Simple summary of what was changed.

## Motivation

Why are you making this change? If it's for fixing a bug, if possible, please include a code snippet or example project that demonstrates the issue.

## Testing

How was the code tested? Be as specific as possible.

### DIFF
--- a/Stripe/PublicHeaders/STPAPIClient.h
+++ b/Stripe/PublicHeaders/STPAPIClient.h
@@ -127,6 +127,19 @@ static NSString *const STPSDKVersion = @"11.5.0";
 
 @end
 
+#pragma mark Connect Accounts
+
+/**
+ Stripe extensions for working with Connect Accounts
+ */
+@interface STPAPIClient (ConnectAccounts)
+
+- (void)createTokenWithConnectAccount:(id)obj completion:(__nullable STPTokenCompletionBlock)completion;
+
+@end
+
+#pragma mark Upload
+
 /**
  STPAPIClient extensions to upload files.
  */
--- a/Stripe/STPAPIClient.m
+++ b/Stripe/STPAPIClient.m
@@ -287,6 +287,18 @@ static NSString * const FileUploadURL = @"https://uploads.stripe.com/v1/files";
 
 @end
 
+#pragma mark - Connect Accounts
+
+@implementation STPAPIClient (ConnectAccounts)
+
+- (void)createTokenWithConnectAccount:(__unused id)obj completion:(__nullable STPTokenCompletionBlock)completion {
+    if (completion) {
+        completion(nil, nil);
+    }
+}
+
+@end
+
 #pragma mark - Upload
 
 @implementation STPAPIClient (Upload)
--- a/Stripe/STPLegalEntityParams.h
+++ b/Stripe/STPLegalEntityParams.h
@@ -1,0 +1,48 @@
+//
+//  STPLegalEntityParams.h
+//  Stripe
+//
+//  Created by Daniel Jackson on 12/20/17.
+//  Copyright © 2017 Stripe, Inc. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+#import "STPFormEncodable.h"
+
+@class STPAddress;
+
+
+@interface STPVerificationParams: NSObject<STPFormEncodable>
+@property (nonatomic, copy, nullable) NSString *document;
+@end
+
+
+@interface STPPersonParams: NSObject<STPFormEncodable>
+@property (nonatomic, copy, nullable) NSString *firstName;
+@property (nonatomic, copy, nullable) NSString *lastName;
+@property (nonatomic, copy, nullable) NSString *maidenName;
+@property (nonatomic, strong, nullable) STPAddress *address;
+@property (nonatomic, copy, nullable) NSDateComponents *dateOfBirth;
+@property (nonatomic, strong, nullable) STPVerificationParams *verification;
+@end
+
+
+typedef NS_ENUM(NSUInteger, STPLegalEntityType) {
+    STPLegalEntityTypeIndividual,
+    STPLegalEntityTypeCompany,
+};
+
+
+@interface STPLegalEntityParams : STPPersonParams
+@property (nonatomic, strong, nullable) NSArray<STPPersonParams *> *additionalOwners;
+@property (nonatomic, copy, nullable) NSString *businessName;
+@property (nonatomic, copy, nullable) NSString *businessTaxId;
+@property (nonatomic, copy, nullable) NSString *businessVATId;
+@property (nonatomic, copy, nullable) NSString *gender;
+@property (nonatomic, strong, nullable) STPAddress *personalAddress;
+@property (nonatomic, copy, nullable) NSString *personalIdNumber;
+@property (nonatomic, copy, nullable) NSString *phoneNumber;
+@property (nonatomic, copy, nullable) NSString *ssnLast4;
+@property (nonatomic, copy, nullable) NSString *taxIdRegistrar;
+@end
+
--- a/Stripe/STPLegalEntityParams.m
+++ b/Stripe/STPLegalEntityParams.m
@@ -1,0 +1,48 @@
+//
+//  STPLegalEntityParams.m
+//  Stripe
+//
+//  Created by Daniel Jackson on 12/20/17.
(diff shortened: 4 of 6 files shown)


Stripe/PublicHeaders/STPAPIClient.h
```diff
@@ -127,6 +127,19 @@ static NSString *const STPSDKVersion = @"11.5.0";
...
 @end
 
+#pragma mark Connect Accounts
+
+/**
+ Stripe extensions for working with Connect Accounts
+ */
+@interface STPAPIClient (ConnectAccounts)
+
+- (void)createTokenWithConnectAccount:(id)obj completion:(__nullable STPTokenCompletionBlock)completion;
```
danj-stripe: obj needs to turn into:
```
account: {
  legal_entity: {...},
  tos_shown_and_accepted: true,
}
```
with either `legal_entity` **or** `tos_shown_and_accepted`, or both.

I don't know whether I want to create another `...Params` class, or change this method to take two nullable parameters.

I also think it's unfortunate that `Bool?` is `nullable NSNumber` in Obj-C

Stripe/STPLegalEntityParams.h
```diff
@@ -0,0 +1,48 @@
...
+@end
+
+
+typedef NS_ENUM(NSUInteger, STPLegalEntityType) {
+    STPLegalEntityTypeIndividual,
+    STPLegalEntityTypeCompany,
+};
+
+
+@interface STPLegalEntityParams : STPPersonParams
```
danj-stripe: Any problems foreseen by inheriting from `STPPersonParams`, which conforms to `STPFormEncodable`?

danj-stripe: Do `STPVerificationParams` and `STPPersonParams` belong in their own files? Or are they fine here?

They're *only* used here (for now?), **and** the inheritance also links `STPLegalEntityParams` with `STPPersonParams` pretty closely.

danj-stripe:
@bg-stripe WDYT? Does this look ~reasonable?

Stripe/PublicHeaders/STPAPIClient.h
bg-stripe: Hmm, I think it'd be better to make this method take parameters rather than make a parent `AccountParams` class, but don't feel strongly.

Is it necessary to model `tos_shown_and_accepted` as a `Bool?`, or can we treat `tos_shown_and_accepted: nil` as `tos_shown_and_accepted: false`?

Stripe/STPLegalEntityParams.h
bg-stripe: hmm, I can't think of any

bg-stripe: I think they're fine together!

bg-stripe:
this looks reasonable to me!

Stripe/PublicHeaders/STPAPIClient.h
danj-stripe: I don't think that would work. Valid values for `tos_shown_and_accepted` are `{true, nil}`. Some basic use cases:

* `account: { tos_shown_and_accepted: true }`: minimum required to set up a new account
* `account: { legal_entity: { ... }, tos_shown_and_accepted: true }`: setting up new account with some data about the legal entity
* `account: { legal_entity: { ... } }`: valid for updating a pre-existing account
* `account: { tos_shown_and_accepted: false, ... }`: rejected by the server as invalid

As I understand it, the initial purpose for this API is collecting positive affirmation that the user has seen & accepted the ToS, instead of the old world where the platform is allowed to vouch that fact as true. However, it's *also* just as usable for updating the legal entity info directly.


I'll think about `AccountParams` a little more. I like that it's easier to add new fields in the future, and the possibility of providing >1 constructors to help understand the `legal_entity/tos_shown_and_accepted` distinction. But it's a lot just for two optional fields

bg-stripe: ah interesting, if that's the case then perhaps an `AccountParams` object with multiple constructors makes sense. thanks for the explanation!

danj-stripe:
This includes the commit for #881, and I think it's now complete.

r? @bg-stripe

Stripe/STPAPIClient.m
```diff
@@ -288,6 +288,19 @@ - (void)createTokenWithPersonalIDNumber:(NSString *)pii completion:(__nullable S
 
 @end
 
+#pragma mark - Connect Accounts
+
+@implementation STPAPIClient (ConnectAccounts)
+
+- (void)createTokenWithConnectAccount:(__unused id)account completion:(__nullable STPTokenCompletionBlock)completion {
```
bg-stripe: replace `__unused id` with `STPConnectAccountParams *``

Stripe/STPAPIClient.m
```diff
@@ -288,6 +288,19 @@ - (void)createTokenWithPersonalIDNumber:(NSString *)pii completion:(__nullable S
...
+    NSMutableDictionary *params = [[STPFormEncoder dictionaryForObject:account] mutableCopy];
+    [[STPTelemetryClient sharedInstance] addTelemetryFieldsToParams:params];
```
bg-stripe: I think we probably don't need to add telemetry data.

Stripe/STPAddress.m
```diff
@@ -42,14 +44,8 @@ + (NSDictionary *)shippingInfoForChargeWithAddress:(nullable STPAddress *)addres
...
-    NSMutableDictionary *addressDict = [NSMutableDictionary new];
-    addressDict[@"line1"] = address.line1;
-    addressDict[@"line2"] = address.line2;
-    addressDict[@"city"] = address.city;
-    addressDict[@"state"] = address.state;
-    addressDict[@"postal_code"] = address.postalCode;
-    addressDict[@"country"] = address.country;
-    params[@"address"] = [addressDict copy];
+    // Re-use STPFormEncoder
+    params[@"address"] = [STPFormEncoder dictionaryForObject:address];
```
bg-stripe: 👍

Stripe/STPLegalEntityParams.m
```diff
@@ -0,0 +1,150 @@
...
+ Add STPFormEncodable conformance for `STPPersonParams.dateOfBirth`
+ */
+@interface NSDateComponents (STPFormEncodable) <STPFormEncodable> @end
+@implementation NSDateComponents (STPFormEncodable)
+
+- (NSDictionary *)additionalAPIParameters {
+    return @{};
+}
+
+- (void)setAdditionalAPIParameters:(__unused NSDictionary *)additionalAPIParameters {
```
bg-stripe: 👍

Stripe/STPConnectAccountParams.m
```diff
@@ -0,0 +1,65 @@
+//
+//  STPConnectAccountParams.m
+//  Stripe
+//
+//  Created by Daniel Jackson on 1/4/18.
+//  Copyright © 2018 Stripe, Inc. All rights reserved.
+//
+
+#import "STPConnectAccountParams.h"
+
```
bg-stripe: NS_ASSUM_NONNULL

Stripe/STPLegalEntityParams.m
```diff
@@ -0,0 +1,150 @@
+//
+//  STPLegalEntityParams.m
+//  Stripe
+//
+//  Created by Daniel Jackson on 12/20/17.
+//  Copyright © 2017 Stripe, Inc. All rights reserved.
+//
+
+#import "STPLegalEntityParams.h"
```
bg-stripe: NS_ASSUME_NONNULL

danj-stripe:
Also, cc @ksun-stripe, who implemented this in stripe/stripe-android#492

Stripe/STPAPIClient.m
danj-stripe: Papertrail: bg and I discussed in Slack. Android sends telemetry for all tokens. We think this is fine, but I have an open ask out to the team that owns telemetry.

danj-stripe:
@bg-stripe I've addressed your comments (thanks for all the help!), and fixed some other problems. Travis has a 1000+ build backlog and is running at half capacity. If you want to re-review now you can, or I'll assign to you again if/when the build has succeeded.

PR #879, opened 2018-01-09
Fix STPEncodable to work with collections of STPEncodable objects, add array index to encoded string

danj-stripe:
## Summary

Fix STPEncodable to work with arrays, dictionaries and sets of STPEncodable objects, and add array index to the encoded string.

## Motivation

It currently only works if the collection contains primitive items, and it's ambiguous if the array contains nested objects:

It won't work with just `[]`, because this particular test case could either deserialize
as a single object with two properties filled out, or two objects with one property
each.

## Testing

Added unit tests to expose the bugs & verify it works as expected.

Quick pass through the standard integration to ensure taking a payment still works.

Manually used the Connect Account Token API to verify that the Stripe API accepts `[0]=` instead of `[]=` for setting array contents.

### DIFF
--- a/Tests/Tests/STPFormEncoderTest.m
+++ b/Tests/Tests/STPFormEncoderTest.m
@@ -95,7 +95,22 @@
     STPTestFormEncodableObject *testObject = [STPTestFormEncodableObject new];
     testObject.testProperty = @"success";
     testObject.testArrayProperty = @[@1, @2, @3];
-    XCTAssertEqualObjects([self encodeObject:testObject], @"test_object[test_array_property][]=1&test_object[test_array_property][]=2&test_object[test_array_property][]=3&test_object[test_property]=success");
+    XCTAssertEqualObjects([self encodeObject:testObject], @"test_object[test_array_property][0]=1&test_object[test_array_property][1]=2&test_object[test_array_property][2]=3&test_object[test_property]=success");
+}
+
+- (void)testFormEncoding_arrayOfEncodable {
+    STPTestFormEncodableObject *testObject = [STPTestFormEncodableObject new];
+
+    STPTestFormEncodableObject *inner1 = [STPTestFormEncodableObject new];
+    inner1.testProperty = @"inner1";
+    STPTestFormEncodableObject *inner2 = [STPTestFormEncodableObject new];
+    inner2.testArrayProperty = @[@"inner2"];
+
+    testObject.testArrayProperty = @[inner1, inner2];
+
+    XCTAssertEqualObjects([self encodeObject:testObject],
+                          @"test_object[test_array_property][0][test_property]=inner1"
+                          "&test_object[test_array_property][1][test_array_property][0]=inner2");
 }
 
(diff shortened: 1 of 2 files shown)


danj-stripe:
I think stripe-android does serialize objects inside arrays, but I don't think it correctly handles this ambiguity.

https://github.com/stripe/stripe-android/blob/b8e390970eb6694522a259468e5baf5879463594/stripe/src/main/java/com/stripe/android/StripeApiHandler.java#L744

Tests/Tests/STPFormEncoderTest.m
```diff
@@ -98,6 +98,21 @@ - (void)testFormEncoding_arrayValue {
...
+    STPTestFormEncodableObject *testObject = [STPTestFormEncodableObject new];
+
+    STPTestFormEncodableObject *inner1 = [STPTestFormEncodableObject new];
+    inner1.testProperty = @"inner1";
+    STPTestFormEncodableObject *inner2 = [STPTestFormEncodableObject new];
+    inner2.testArrayProperty = @[@"inner2"];
+
+    testObject.testArrayProperty = @[inner1, inner2];
+
+    XCTAssertEqualObjects([self encodeObject:testObject],
```
danj-stripe: This fails, `encodeObject:` returns:

```
 "test_object[test_array_property][]=<STPTestFormEncodableObject: 0x7fb845d0ffd0>"
"&test_object[test_array_property][]=<STPTestFormEncodableObject: 0x7fb845d10010>"
```

danj-stripe: As far as I can tell, Android would serialize this as

```java
 "test_object[test_array_property][][test_property]=inner1" +
"&test_object[test_array_property][][test_array_property][]=inner2"
```

Which could either be an array with a single object `[{ test_property=inner1, test_array_property=[inner2] }]` **or** it could be an array with two objects `[{ test_property=inner1 }, { test_array_property=[inner2] }]`

Tests/Tests/STPFormEncoderTest.m
```diff
@@ -98,6 +98,21 @@ - (void)testFormEncoding_arrayValue {
...
+                          @"test_object[test_array_property][0][test_property]=inner1"
+                          "&test_object[test_array_property][1][test_array_property][0]=inner2");
```
danj-stripe: I'm not sure if `[0]` and `[1]` are the right solution here. Still need to test against the Stripe API.

danj-stripe: Talked to basta & tested against the API, this *is* correct.

danj-stripe:
@bg-stripe Dictionaries and Sets also have the problem that we won't deep encode into them. Should I include this commit, even though we don't need it (yet?)

https://github.com/stripe/stripe-ios/commit/f47f148695bc0b516741a84023e8ea5be46006cc

danj-stripe:
Ok, added Dictionary & Set support.

re-r? @bg-stripe

PR #881, opened 2018-01-10
Encode NSNumber-wrapped boolean values as `true`/`false`, not `1`/`0`

danj-stripe:
## Summary

Identify NSNumbers that were created via any of the `bool` constructors, and encode them
as `true`/`false`.

## Motivation

Trying to encode/serialize a BOOL/NSNumber field in Connect Account token creation failed
with invalid value because it was sending `1` instead of `true`.

## Testing

Adding a test to verify this works, and that this doesn't break other NSNumber instances
that store 0 or 1.

Verified that it works with the new Connect Account token API

### DIFF
--- a/Stripe/STPFormEncoder.m
+++ b/Stripe/STPFormEncoder.m
@@ -159,7 +159,16 @@ NSString * STPPercentEscapedStringFromString(NSString *string) {
     if (!self.value || [self.value isEqual:[NSNull null]]) {
         return STPPercentEscapedStringFromString([self.field description]);
     } else {
-        return [NSString stringWithFormat:@"%@=%@", STPPercentEscapedStringFromString([self.field description]), STPPercentEscapedStringFromString([self.value description])];
+        NSString *unescapedValue = [self.value description];
+
+        // coerce boxed BOOL values back into true/false
+        // https://stackoverflow.com/a/30223989/1196205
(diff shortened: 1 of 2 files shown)


Stripe/STPFormEncoder.m
```diff
@@ -159,7 +159,16 @@ - (NSString *)URLEncodedStringValue {
     if (!self.value || [self.value isEqual:[NSNull null]]) {
         return STPPercentEscapedStringFromString([self.field description]);
     } else {
-        return [NSString stringWithFormat:@"%@=%@", STPPercentEscapedStringFromString([self.field description]), STPPercentEscapedStringFromString([self.value description])];
+        NSString *unescapedValue = [self.value description];
+
+        // coerce boxed BOOL values back into true/false
+        // https://stackoverflow.com/a/30223989/1196205
+        if ([self.value isKindOfClass:[NSNumber class]]
+            && (CFBooleanGetTypeID() == CFGetTypeID((__bridge CFTypeRef)(self.value)))) {
```
bg-stripe: oh neat, I didn't know you could do this

danj-stripe:
I forgot to mention that I also audited all of our `STPFormEncodable` classes, and did not find any `BOOL` nor boolean `NSNumber` instances that would have already used or relied on the previous behavior.

PR #883, opened 2018-01-12
Fix `STPPaymentCardTextField` touch handling bugs for small devices on iOS 11

danj-stripe:
## Summary

The `maskView` over `STPPaymentCardTextField` numbers no longer blocks touches on iOS 11.
Also, making the `brandImage` tappable to move focus/first responder to the numbers text
field.

## Motivation

Found while testing #870, #855, etc. Filed as IOS-601

## Testing

Manual testing in Standard Integration sample app, on small & large phones.

### DIFF
--- a/Stripe/STPPaymentCardTextField.m
+++ b/Stripe/STPPaymentCardTextField.m
@@ -1070,6 +1070,7 @@ typedef NS_ENUM(NSInteger, STPCardTextFieldState) {
                                                                     fieldsHeight)];
         maskView.backgroundColor = [UIColor blackColor];
         maskView.opaque = YES;
+        maskView.userInteractionEnabled = NO;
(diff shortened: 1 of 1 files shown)


danj-stripe:
In the debugger view, you can see the frames of the fields:
screen shot 2018-01-11 at 4 58 20 pm

* `brandImage`: selected (blue highlight), *now* has a tap recognizer to move focus to the numbers text field
* `maskView`: shown in black. The only part of the numbers field that's drawn is what's covered by the `maskView`, which is once again a valid tap target for the numbers text field.
* Similarly, you can see the bounds on the expiration & CVC fields (they go full height, but hug horizontally pretty tightly).
* Taps in the white areas between the text fields will `resignFirstResponder`, which isn't great, but hasn't changed.

PR #893, opened 2018-02-02
Fix createsCardSources <> didCreatePaymentResult , Apple Pay

bg-stripe:
r? @danj-stripe 

This fixes two issues:
- didCreatePaymentResult wasn't called when createsCardSources = true
- selecting apple pay created a card token when createsCardSources = true

I tested Apple Pay and didCreatePaymentResult in the example app with sources. I've also updated the example app to always create card sources.

### DIFF
--- a/Stripe/PKPaymentAuthorizationViewController+Stripe_Blocks.m
+++ b/Stripe/PKPaymentAuthorizationViewController+Stripe_Blocks.m
@@ -10,7 +10,9 @@
 
 #import "PKPaymentAuthorizationViewController+Stripe_Blocks.h"
 #import "STPAPIClient+ApplePay.h"
-
+#import "STPCard.h"
+#import "STPSource.h"
+#import "STPToken.h"
 
 static char kSTPBlockBasedApplePayDelegateAssociatedObjectKey;
 
@@ -22,10 +24,11 @@ typedef void (^STPApplePayShippingAddressCompletionBlock)(PKPaymentAuthorization
 @property (nonatomic, copy) STPShippingAddressSelectionBlock onShippingAddressSelection;
 @property (nonatomic, copy) STPShippingMethodSelectionBlock onShippingMethodSelection;
 @property (nonatomic, copy) STPPaymentAuthorizationBlock onPaymentAuthorization;
-@property (nonatomic, copy) STPApplePayTokenHandlerBlock onTokenCreation;
+@property (nonatomic, copy) STPApplePaySourceHandlerBlock onSourceCreation;
 @property (nonatomic, copy) STPPaymentCompletionBlock onFinish;
 @property (nonatomic) NSError *lastError;
 @property (nonatomic) BOOL didSucceed;
+@property (nonatomic) BOOL createSource;
 @end
 
 typedef void (^STPPaymentAuthorizationStatusCallback)(PKPaymentAuthorizationStatus status);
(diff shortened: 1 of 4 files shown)


Stripe/PKPaymentAuthorizationViewController+Stripe_Blocks.m
```diff
@@ -35,22 +38,37 @@ @implementation STPBlockBasedApplePayDelegate
...
+    void(^tokenOrSourceCompletion)(id<STPSourceProtocol>, NSError *) = ^(id<STPSourceProtocol> result, NSError *error) {
         if (error) {
             self.lastError = error;
             completion(PKPaymentAuthorizationStatusFailure);
             return;
         }
-        self.onTokenCreation(token, ^(NSError *tokenCreationError){
-            if (tokenCreationError) {
-                self.lastError = tokenCreationError;
+        id<STPSourceProtocol> source;
```
danj-stripe: What is `source` used for? It kind of looks like you meant to pass `source` to the `onSourceCreation()` call, but it's getting `result` right now, so I think this is dead code.

edit: neat, static analyzer agrees with me and failed the build 😄

Stripe/STPPaymentContext.m
```diff
@@ -582,9 +582,10 @@ - (void)requestPayment {
         else if ([self requestPaymentShouldPresentShippingViewController]) {
             [self presentShippingViewControllerWithNewState:STPPaymentContextStateRequestingPayment];
         }
-        else if ([self.selectedPaymentMethod isKindOfClass:[STPCard class]]) {
+        else if ([self.selectedPaymentMethod isKindOfClass:[STPCard class]] ||
+                 [self.selectedPaymentMethod isKindOfClass:[STPSource class]]) {
```
danj-stripe: WDYT about checking `[self.selectedPaymentMethod conformsToProtocol:@protocol(STPSourceProtocol)]` instead?

I don't know: 
* whether there are other classes that conform to the protocol at we want excluded from this branch (making the check invalid)
* how likely it is that we'll add new classes that *do* conform to the protocol that should take this branch (updated check would future-proof us)
* how likely that we'll remove conformance to the protocol from these classes (which would break the check)

Stripe/PKPaymentAuthorizationViewController+Stripe_Blocks.m
danj-stripe: I'm more confident that you meant to pass `source` instead of `result`: down in the definition of `applePaySourceHandler`, the code used to be constructing a `STPPaymentResult` with `token.card`, and now it just passes `source` directly.

bg-stripe: oh derp, good catch!

Stripe/STPPaymentContext.m
bg-stripe: mm, interesting idea. I think it's unlikely that we'll create other class that conform to this protocol (the protocol is mostly a compatibility shim for card tokens to work with sources). I think we should leave it this way and explicitly reference `STPCard`, which we should start treating as a legacy class (card tokens will continue to be supported for a long time, but sources are the preferred integration).

Stripe/PKPaymentAuthorizationViewController+Stripe_Blocks.m
bg-stripe: so i think i understand why this worked (returning `STPToken` instead of the child `STPCard` for an apple pay source). we don't display apple pay sources in the UI, so it didn't matter that `STPToken` doesn't conform to `<STPPaymentMethod>` (the protocol defining how a payment method is displayed). i'll still fix for consistency with how we handle non-apple pay sources.

bg-stripe: fixed in eb09444, re-r?

Issue #895, opened 2018-02-14
Static analysis shows couple of issues in latest release


bielikb:
## Summary

Static code analysis unveils couple of issues in latest release

## Code to reproduce

Run static analysis

## iOS version

11.2

## Installation method

cocoa pods

## SDK version
12.1.0

## Other information

![screen shot 2018-02-14 at 15 55 56](https://user-images.githubusercontent.com/4924709/36210923-fe4f6b60-119f-11e8-8939-01bb0748a2a5.png)

bg-stripe:
Thanks for pointing this out @bielikb , we'll look into it. I think this is a new analyzer warning – are you using the Xcode 9.3 beta ?

bielikb:
@bg-stripe Im using Xcode version 9.2 (9C40b)

danj-stripe:
Thanks again for reporting, @bielikb. We just released v12.1.1, which has a fix for this.

PR #897, opened 2018-02-20
Match cocoapods static analyzer settings, fix issues

danj-stripe:
## Summary

`Misuse of 'nonnull'` and `'Suspicious Conversions of NSNumber'` are aggressive by default
when using cocoapods. Adopts the same settings for our project.

screen shot 2018-02-20 at 1 54 36 pm

## Motivation

#895

## Testing

Automated tests. There should be no functional changes from these commits.


joeydong-stripe:
👍

PR #902, opened 2018-02-26
Add missing nullability annotations to STPRedirectContextCompletionBlock

### LINKED ISSUE #901: STPRedirectContextCompletionBlock's error param should be optional.
I'am using Source  method for payment.

I Creating **STPRedirectContext** object using init method.
Below CompletionBlock is used for this.
`typedef void (^STPRedirectContextCompletionBlock)(NSString *sourceID, NSString *clientSecret, NSError *error);
`

## Code to reproduce
```
self.redirectContext = STPRedirectContext.init(source: source!, completion: { (sourceId, clientSecret, error) in
 if error != nil {
   debugPrint("Error:\(error)")
 }
```
getting warning in checking Error (

> Comparing non-optional value of type 'Error' to nil always returns true

).

## Summary
So I think Error param in **STPRedirectContextCompletionBlock** should be **_nullable_** (i.e in Swift Optional), Like below:
`typedef void (^STPSourceCompletionBlock)(STPSource * __nullable source, NSError * __nullable error);
`

## Installation method

How did you install our SDK?
Using **CocoaPod**.

danj-stripe:
## Summary

Marking `clientSecret` and `error` parameters as `__nullable`. Fixes #901

## Motivation

The header is marked with `NS_ASSUME_NONNULL_BEGIN`, so this fixes Swift compatibility when
checking whether the error is null.

## Testing

Unit tests for STPRedirectContext passing.

Wrote throwaway code in our swift integration example to verify the parameters are
imported as optional.

### DIFF
--- a/Stripe/PublicHeaders/STPRedirectContext.h
+++ b/Stripe/PublicHeaders/STPRedirectContext.h
@@ -47,7 +47,7 @@ typedef NS_ENUM(NSUInteger, STPRedirectContextState) {
  is if SFSafariViewController fails its initial load (like the user has no 
  internet connection, or servers are down).
  */
-typedef void (^STPRedirectContextCompletionBlock)(NSString *sourceID, NSString *clientSecret, NSError *error);
+typedef void (^STPRedirectContextCompletionBlock)(NSString *sourceID, NSString * __nullable clientSecret, NSError * __nullable error);
 
 /**
  This is a helper class for handling redirect sources.


danj-stripe:
As far as I can tell, `clientSecret` can be `nil`. It's marked as `nullable`:

https://github.com/stripe/stripe-ios/blob/e53d9704cd094c8cbe3d9e1b31734d701476ee58/Stripe/PublicHeaders/STPSource.h#L45

Also, it's not a required field when decoding `STPSource` objects:

https://github.com/stripe/stripe-ios/blob/e53d9704cd094c8cbe3d9e1b31734d701476ee58/Stripe/STPSource.m#L223-L234

PR #904, opened 2018-03-07
Mark invalid view controller initializers as unavailable

### LINKED ISSUE #903: STPAddCardViewController done button disabled
## Summary
When initializing `STPAddCardViewController` with a theme and filling out the credit card information, the done button doesn't become enabled upon finishing. It works, however, when not initializing with a theme.

## Code to reproduce
```swift
let theme = STPTheme()
theme.accentColor = UIColor(named: "Mocha")
let addCardViewController = STPAddCardViewController(theme: theme)
addCardViewController.delegate = self
navigationController?.pushViewController(addCardViewController, animated: true)
```

## iOS version
iOS 11.2

## Installation method
Cocoapods

## SDK version
12.1.1

bg-stripe:
r? @danj-stripe 
Fixes #903 

`STPCoreViewController` defines `initWithTheme`, which should be unavailable in all public subclasses (as should the `UIViewController`-defined initializers)

Autocomplete will now only show valid init options.

image

### DIFF
--- a/Stripe/PublicHeaders/STPAddCardViewController.h
+++ b/Stripe/PublicHeaders/STPAddCardViewController.h
@@ -64,6 +64,22 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic, strong, nullable) UIView *customFooterView;
 
+/**
+ Use init: or initWithConfiguration:theme:
+ */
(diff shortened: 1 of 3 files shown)


bg-stripe:
I think it has to be public because it's publicly subclassed. I think the initializers it defines could be in a private header – maybe making a private header would've been a better solution. Going to leave this way though in the interest of time.

PR #905, opened 2018-03-09
Update Diners Club and Discover BIN ranges

joeydong-stripe:
Looks like the only functional change is removing the "62..." range from Discover to be used by UnionPay later on.

### DIFF
--- a/Stripe/STPBINRange.m
+++ b/Stripe/STPBINRange.m
@@ -30,22 +30,32 @@
     static dispatch_once_t onceToken;
     dispatch_once(&onceToken, ^{
         NSArray *ranges = @[
-                            // Catch-all values
+                            // Unknown
                             @[@"", @"", @16, @(STPCardBrandUnknown)],
+
+                            // American Express
                             @[@"34", @"34", @15, @(STPCardBrandAmex)],
                             @[@"37", @"37", @15, @(STPCardBrandAmex)],
+
+                            // Diners Club
                             @[@"30", @"30", @14, @(STPCardBrandDinersClub)],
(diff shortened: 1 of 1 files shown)


Stripe/STPBINRange.m
```diff
@@ -30,22 +30,30 @@ @implementation STPBINRange
...
                             @[@"34", @"34", @15, @(STPCardBrandAmex)],
                             @[@"37", @"37", @15, @(STPCardBrandAmex)],
-                            @[@"30", @"30", @14, @(STPCardBrandDinersClub)],
-                            @[@"36", @"36", @14, @(STPCardBrandDinersClub)],
-                            @[@"38", @"39", @14, @(STPCardBrandDinersClub)],
-                            @[@"60", @"60", @16, @(STPCardBrandDiscover)],
-                            @[@"62", @"62", @16, @(STPCardBrandDiscover)],
-                            @[@"64", @"65", @16, @(STPCardBrandDiscover)],
+
+                            @[@"30000000", @"30599999", @14, @(STPCardBrandDinersClub)],
```
bg-stripe: I think we should keep these the way they were before (more permissive and matching the behavior of stripe.js), and then replace the "62" Discover range with UnionPay in #906 .

joeydong-stripe: Got it! Thanks for checking in on this.

joeydong-stripe:
I'll hold off on merging this until https://github.com/stripe/stripe-ios/pull/906 is ready in case people are grabbing from `master` so that we don't have an interm period where `62...` is invalid.

Issue #912, opened 2018-03-17
PKAddress Field Deprecation Warnings


dmerfeld:
## Summary

Compiler warnings of PKAddress Field Deprecation in iOS 11 when building project.

## Code to reproduce

<img width="271" alt="screen shot 2018-03-17 at 3 30 46 pm" src="https://user-images.githubusercontent.com/1771887/37559704-7b210ca0-29f8-11e8-8c56-c6ec26a4d06b.png">

Warnings appear when compiling a project after doing a "Clean Build Folder", archiving and building.

## iOS version

Occurs in projects targeting: iOS 11, 11.1 and 11.2

## Installation method

Installed SDK using Pods

## SDK version

Issue occurred on 12.1.1 and after upgrading to 12.1.2

## Other information
Xcode version 9.2 (9C40b).

Issue was also reported here: 
https://github.com/stripe/stripe-ios/issues/814

bg-stripe:
Hey @dmerfeld , thanks for reporting the issue. We've reproduced the deprecation warnings, and we'll silence them in our next release.

Ben

PR #913, opened 2018-03-20
Fix PKAddress warnings

bg-stripe:
r? @danj-stripe 

Silences the deprecation warnings in STPAddress.h and adds a new objc installation test to catch regressions.

I also did some cleanup on the cocoapods installation projects:
- renamed the directories `_objc`/`_swift`
- updated to Xcode 9 project settings
- turned on treat warnings as errors in the project + pods


Tests/installation_tests/cocoapods/with_frameworks_objc/CocoapodsTest/ViewController.m
```diff
@@ -0,0 +1,30 @@
+//
+//  ViewController.m
+//  CocoapodsTest
+//
+//  Created by Ben Guo on 3/20/18.
+//  Copyright © 2018 Stripe. All rights reserved.
+//
+
+#import "ViewController.h"
+#import <Stripe/Stripe.h>
```
danj-stripe: I wasn't sure whether we wanted to use `#import` or `@import` - they both seemed to reproduce the warning. Is there a good reason to prefer one or the other here?

Stripe/PublicHeaders/STPAddress.h
```diff
@@ -220,7 +220,10 @@ extern STPContactField const STPContactFieldName;
  @return The closest representation of the billing address requirement as
  a PKAddressField value.
  */
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated"
```
danj-stripe: Looks like this should go before the documentation comment, build is failing because jazzy thinks these two methods are undocumented.

Tests/installation_tests/cocoapods/with_frameworks_objc/CocoapodsTest/ViewController.m
bg-stripe: I think they're effectively identical – `#import` maps to `@import` if modules are enabled, which is on by default in new projects.

https://stackoverflow.com/a/18947634

Stripe/PublicHeaders/STPAddress.h
bg-stripe: 👍 fixed in da1cbea

bg-stripe:
Fixes #912

PR #921, opened 2018-04-11
Add Norwegian localization

Nailer:
## Summary

Added Norwegian Localization

## Motivation

We're using stripe for an app I'm working on, and figured I could contribute.

## Testing

No testing, just read through it a couple of times to ensure correctness. No guarantees though!


joeydong-stripe:
These look great @Nailer. The translations looks suitable to me (using Google Translate 😆).

I'm going to merge it into one of my branches first so I can do some finishing cleanup steps (not related to the translation strings themselves).

Thanks again!

Issue #917, opened 2018-04-08
Stripe 3D secure redirect issues


buntydey:
To make it easier to diagnose your issue, please fill out the following:

## Summary
Facing issue with 3DS redirect
A simple summary of the problems you're having.

I am facing 3 issues while incorporating 3DS redirect. 

Issue #1) When using redirectcontext, when I used redirectContext?.startRedirectFlow(from: self) or redirectContext?.startSafariViewControllerRedirectFlow(from: self), the safari view controller seems to come and go back in a flash without any intervention and brings me back to the app.

Issue #2) When I used redirectContext?.startSafariAppRedirectFlow(), the safari view controller page for payment authorization looks to open fine. However, when I click on authorize payment button, it does not automatically gets me back to the app. Instead shows a button saying "Return to App_name". Shouldnt I be automatically redirected to the app as I have mentioned the return URL when creating the 3Dsource?

Issue #3) With #2 mentioned above, when I click on "Return to App", the appdelegate boolean stripeHandled returns as false, even though I can see that the url returned to the app is the one that I set as the return URL while creating the 3D source, 

Please let me know what am I missing!

## Code to reproduce

If possible, please include a brief piece of code (or ideally, a link to an example project) demonstrating the problem you're having.

    func paymentContext(_ paymentContext: STPPaymentContext, didCreatePaymentResult paymentResult: STPPaymentResult, completion: @escaping STPErrorBlock) {
        let source = paymentResult.source.stripeID
        self.price = (Int(amountTextBox.text!)!) * 100
        createStripeSource(card: source)
        
    }

     func createStripeSource(card: String) {
        
        let threedparams = STPSourceParams.threeDSecureParams(withAmount: 1098, currency: "hkd", returnURL: "stripe-payment://stripe-redirect", card: (card))
            
            STPAPIClient.shared().createSource(with: threedparams) { source, error in
                if let err = error {
                    print(error?.localizedDescription)
                } else if let source = source {
                    let redirectContext = STPRedirectContext(source: source, completion: { (sourceID, clientSecret, error) in
                        print("User back !")
                        STPAPIClient.shared().startPollingSource(withId: sourceID, clientSecret: clientSecret!, timeout: 10, completion: { source, err in
                            
                            if let source: STPSource = source {
                                switch source.status {
                                case STPSourceStatus.chargeable:
                                    print("chargeable")
[...]

PR #937, opened 2018-05-17
Workaround for SafariVC reporting failure while redirecting to iDEAL

danj-stripe:
## Summary

Ignores reported errors from SafariVC unless we know the error was for a
`stripe.com` domain.

Adds missing error handling to the example app for errors from `STPRedirectContext`

## Motivation

Reported in #917 and to Stripe support, more details internally in IOS-729.

SafariVC is basically a black box: give it a URL and ask it to load. It'll
report via the delegate method when the initial load finishes, and whether
it was successful or not.

Prior to this commit, iDEAL (and others?) would fail and close SafariVC.
However, it hadn't actually failed, and just leaving the SafariVC open
would allow the user to finish the redirect workflow.

It's strange to me that SafariVC behaves this way. It's also a little
strange how the redirects are being handled by girogate.de. This works
around the interaction between the two, while still attempting to report
legitimate errors, so the host application can tell the user and allow them
to retry.

## Testing

This is very much dependent on the specific behavior of the 3rd party and SafariVC, and
only shows up with livemode transactions.

I was able to isolate this behavior in a test application that just used SafariVC, hitting
a sinatra web-app that mimicked the redirect behavior I observed from girogate. That allowed
me to eliminate any Stripe SDK-specific code, and experiment to verify hypotheses.

Then, applied this fix, and made sure that the "SafariVC opens and then closes before the 
page loads" bug wasn't happening any more.

Automated tests are passing.

### DIFF
--- a/Stripe/STPRedirectContext.m
+++ b/Stripe/STPRedirectContext.m
@@ -26,7 +26,7 @@ typedef void (^STPBoolCompletionBlock)(BOOL success);
 @property (nonatomic, strong, nullable) SFSafariViewController *safariVC;
 @property (nonatomic, assign, readwrite) STPRedirectContextState state;
 /// If we're on iOS 11+ and in the SafariVC flow, this tracks the latest URL loaded/redirected to during the initial load
-@property (nonatomic, strong, readwrite, nullable) NSURL *latestSafariVCUrl;
+@property (nonatomic, strong, readwrite, nullable) NSURL *lastKnownSafariVCUrl;
 
 @property (nonatomic, assign) BOOL subscribedToURLNotifications;
 @property (nonatomic, assign) BOOL subscribedToForegroundNotifications;
@@ -123,8 +123,8 @@ typedef void (^STPBoolCompletionBlock)(BOOL success);
     if (self.state == STPRedirectContextStateNotStarted) {
         _state = STPRedirectContextStateInProgress;
         [self subscribeToUrlNotifications];
-        self.latestSafariVCUrl = self.source.redirect.url;
-        self.safariVC = [[SFSafariViewController alloc] initWithURL:self.latestSafariVCUrl];
+        self.lastKnownSafariVCUrl = self.source.redirect.url;
+        self.safariVC = [[SFSafariViewController alloc] initWithURL:self.lastKnownSafariVCUrl];
         self.safariVC.delegate = self;
         [presentingViewController presentViewController:self.safariVC
                                                animated:YES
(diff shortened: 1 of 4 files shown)


Stripe/STPRedirectContext.m
```diff
@@ -154,14 +157,37 @@ - (void)safariViewControllerDidFinish:(__unused SFSafariViewController *)control
...
+     We can only track the latest URL loaded on iOS 11, because `safariViewController:initialLoadDidRedirectToURL:`
+     didn't exist prior to that.
+     */
     if (didLoadSuccessfully == NO) {
-        stpDispatchToMainThreadIfNecessary(^{
-            [self handleRedirectCompletionWithError:[NSError stp_genericConnectionError]
-                        shouldDismissViewController:YES];
-        });
+        if (@available(iOS 11, *)) {
+            stpDispatchToMainThreadIfNecessary(^{
```
teich-stripe: We don't do anything here on iOS < 11?

danj-stripe: If there was some way to distinguish between real failures and spurious ones, I'd love to. I wasn't able to find one though.

Stripe/STPRedirectContext.m
```diff
@@ -25,6 +25,8 @@ @interface STPRedirectContext () <SFSafariViewControllerDelegate, STPURLCallback
 @property (nonatomic, strong) STPSource *source;
 @property (nonatomic, strong, nullable) SFSafariViewController *safariVC;
 @property (nonatomic, assign, readwrite) STPRedirectContextState state;
+/// If we're on iOS 11+ and in the SafariVC flow, this tracks the latest URL loaded/redirected to during the initial load
+@property (nonatomic, strong, readwrite, nullable) NSURL *latestSafariVCUrl;
```
joeydong-stripe: Maybe `latestSafariVCUrl` -> `latestKnownSafariVCUrl` to better emphasize it's an approximation

Stripe/STPRedirectContext.m
joeydong-stripe: I had a similar initial reaction mostly by the way it's written. Maybe it needs an else clause with a doc:

```
  if (didLoadSuccessfully == NO) {
    if (@available(iOS 11, *) && [self.latestSafariVCUrl.host containsString:@"stripe.com"]) {
    // Report load errors that exist on `stripe.com` website
    ...
    } else {
      // Do nothing because `latestSafariVCUrl` is not accurate `< iOS 11`
    }
  }
}
```

In any case, this looks functional and don't have strong opinion on this one.

PR #940, opened 2018-05-18
Make iDEAL `name` parameter optional, and accept empty string as nil.

danj-stripe:
## Summary

I considered removing the `name:` parameter completely, but this maintains backwards
compatibility, by just marking it as nullable/optional.

Also handling the empty string the same as `nil` for these optional fields, so that there
isn't a surprise waiting. IMO, an optional field that contains the empty string shouldn't
cause the request to fail.

## Motivation

When iDEAL support was originally implemented in the iOS SDK, this was a required
parameter. It isn't any more.

IOS-739

## Testing

Added a couple of functional tests that exercise this case.


joeydong-stripe:
Maybe we need to add a changelog entry saying something like: "Creating iDEAL sources no longer requires `name`". I dont think the detail about mapping empty strings to nil is useful to mention.

Issue #942, opened 2018-05-22
Scroll bug when presenting addCardViewController


Klingefjord:
## Summary

Scroll bug on addCardViewController. When textfield becomes first responder, the view scrolls autmoatically so that the card animation goes underneath the navigation bar

## Code to reproduce

This is what I use to present the addCardViewController:

```
    func initializeStripe(_ sender: UIViewController, key: String) {
        let addCardViewController = STPAddCardViewController(configuration: StripeUtil.getPaymentConfiguration(key),
                                                             theme: StripeUtil.getTheme())
        addCardViewController.delegate = self
        sender.present(UINavigationController(rootViewController: addCardViewController), animated: true)
        }
    }
```

## iOS version

11.3

## Installation method

CocoaPods

## SDK version

13.0.1

## Other information

![gid](https://user-images.githubusercontent.com/22666776/40348093-d58546ce-5da2-11e8-9010-ddc957a73f00.gif)

joeydong-stripe:
Hi @Klingefjord, thanks for opening an issue and providing the gif.

This looks like the expected behavior as far as I know. The intended behavior is for the text field to be vertically center in the remaining space not taken up by the keyboard. As a result, the card can end up offscreen. What do you think?

Here is the code if you want to try experimenting with it: 

https://github.com/stripe/stripe-ios/blob/master/Stripe/UIViewController%2BStripe_KeyboardAvoiding.m#L50-L60

One option is to stop calling `stp_beginObservingKeyboardAndInsettingScrollView` but that could become an issue in this particular case if the screen is shorter and the text field can end up behind the keyboard frame.

Klingefjord:
Personally I think the textfield shouldn't get centered as it looks especially weird during the card flip animation. Maybe some tweaking to the offset if the textfield is the first one underneath the card? I tried changing https://github.com/stripe/stripe-ios/blob/master/Stripe/UIViewController%2BStripe_KeyboardAvoiding.m#L111 to ```offset.y = ((topOfScreenOffset + topOfKeyboardOffset) / 2) - scrollView.contentInset.top - 70;```, It looks like the following on iPhone 5 and iPhone 7:

<img width="328" alt="skarmavbild 2018-05-23 kl 09 41 27" src="https://user-images.githubusercontent.com/22666776/40410901-30183ba6-5e6f-11e8-8fb4-c35c733b350a.png">

[...]

PR #943, opened 2018-05-23
Adjust keyboard avoiding behavior

joeydong-stripe:
simulator screen shot - iphone se - 2018-05-23 at 10 25 23

## Summary

Stop adjusting `scrollView.contentOffset` and let the system natively scroll because of the `scrollView.contentInset` change. This will scroll the minimum amount until the text field is visible. Previously, we would scroll until the text field is vertically centered on screen.

## Testing

Did some basic testing with iPhone SE 11.2 Portrait + Landscape Simulator but we should do additional testing on other configurations since I don't have a confident understanding of how `UIViewController+Stripe_KeyboardAvoiding` works.

https://github.com/stripe/stripe-ios/issues/942

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+* Adjusts scroll view content offset behavior when focusing on a text field [#943](https://github.com/stripe/stripe-ios/pull/943)
+
 ## 13.0.1 2018-05-17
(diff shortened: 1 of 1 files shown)


Klingefjord:
👍 lgtm (only tested on iPhone 7 iOS 11 though)

PR #954, opened 2018-05-30
Fixes crash in `STPColorUtils.perceivedBrightnessForColor`

### LINKED ISSUE #951: Hard crash on STPColorUtils > perceivedBrightnessForColor
To make it easier to diagnose your issue, please fill out the following:

## Summary

There's a hard crash from this method if the navigation bar color is UIColor.white:

+ (CGFloat)perceivedBrightnessForColor:(UIColor *)color {
    const CGFloat *component = CGColorGetComponents(color.CGColor);
    CGFloat brightness = ((component[0] * 299) + (component[1] * 587) + (component[2] * 114)) / 1000;
    return brightness;
}

Basically if the color is defined using white/alpha rather than rgba, *component here will not have 3 values, and you get an exec bad acess

## Code to reproduce

Just try calling this method with a nav bar with the tint color set to UIColor.white

## iOS version

11

## Installation method

cocoapods

## SDK version

latest

danj-stripe:
## Summary

This is used for deciding whether to use light/dark scroll bars & status bar indicators,
and choosing which color should be used for the switch in `STPSwitchTableViewCell`.

This code assumed the color was always in the RGB color space, which is incorrect. It would
reliably crash with Address Sanitization turned on, and give inconsistent answers with
it turned off.

Now using the `UIColor` `getRed:green:blue:alpha:` method to convert the color to RGB
before calculating the luma value. Also added some links to documentation for why this
calculation is interesting.

## Motivation

Fixes #951

## Testing

Added tests for `STPColorUtils colorIsBright:`

This also tests `perceivedBrightnessForColor:`, but does so by checking something that's
more meaningful to me, rather than testing for specific luma numbers

### DIFF
--- a/Stripe/STPColorUtils.m
+++ b/Stripe/STPColorUtils.m
@@ -11,9 +11,19 @@
 @implementation STPColorUtils
 
 + (CGFloat)perceivedBrightnessForColor:(UIColor *)color {
-    const CGFloat *component = CGColorGetComponents(color.CGColor);
-    CGFloat brightness = ((component[0] * 299) + (component[1] * 587) + (component[2] * 114)) / 1000;
-    return brightness;
+    CGFloat red, green, blue;
+    if ([color getRed:&red green:&green blue:&blue alpha:nil]) {
+        // We're using the luma value from YIQ
+        // https://en.wikipedia.org/wiki/YIQ#From_RGB_to_YIQ
+        // recommended by https://www.w3.org/WAI/ER/WD-AERT/#color-contrast
+        return red * 0.299 + green * 0.587 + blue * 0.114;
+    } else {
+        // Couldn't get RGB for this color, device couldn't convert it from whatever
+        // colorspace it's in.
+        // Make it "bright", since most of the color space is (based on our current
+        // formula), but not very bright.
+        return 0.4;
+    }
 }
 
 + (UIColor *)brighterColor:(UIColor *)color1 color2:(UIColor *)color2 {
--- a/Stripe/STPColorUtils.m
+++ b/Stripe/STPColorUtils.m
@@ -16,13 +16,13 @@
         // We're using the luma value from YIQ
         // https://en.wikipedia.org/wiki/YIQ#From_RGB_to_YIQ
         // recommended by https://www.w3.org/WAI/ER/WD-AERT/#color-contrast
-        return red * 0.299 + green * 0.587 + blue * 0.114;
+        return red * (CGFloat)0.299 + green * (CGFloat)0.587 + blue * (CGFloat)0.114;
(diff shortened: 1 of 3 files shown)


Stripe/STPColorUtils.m
```diff
@@ -11,9 +11,19 @@
 @implementation STPColorUtils
 
 + (CGFloat)perceivedBrightnessForColor:(UIColor *)color {
-    const CGFloat *component = CGColorGetComponents(color.CGColor);
-    CGFloat brightness = ((component[0] * 299) + (component[1] * 587) + (component[2] * 114)) / 1000;
-    return brightness;
+    CGFloat red, green, blue;
+    if ([color getRed:&red green:&green blue:&blue alpha:nil]) {
+        // We're using the luma value from YIQ
+        // https://en.wikipedia.org/wiki/YIQ#From_RGB_to_YIQ
```
bg-stripe: 🎓

danj-stripe:
Thanks for the review, and for the chance to write a little more about this - mostly for posterity.

There's a couple things going on. We have a pre-existing `0.3` threshold for brightness, and it shows up in the `testGrayscaleColorsIsBright` for loops. I'm not sure it's worth the effort to make it a constant, I don't think it's likely that we'll change that without making more changes - and if we do it should be pretty easy to find in the failing tests.

I think it's (slightly?) more likely that in the future we'll want to change our algorithm without changing the behavior with respect to grayscale, so that 30% gray is still the transition point. I think this is an [updated algorithm](https://www.w3.org/WAI/GL/WCAG20-TECHS/G18.html#G18-tests) (plus a [example swift implementation](https://github.com/jathu/sweetercolor/blob/50969d76790fc61e26bb644629c0d4fa39259470/Sweetercolor/Sweetercolor.swift#L252-L262)). If I'm right (and I don't know how likely it that is), then having the value in the tests tied to the value in the implementation would be incorrect.

There's also a situation where we cannot convert a color to RGB (probably because it's in an exotic color space?). The `0.4` return value is somewhat based off the `0.3`, but I really have no idea what the right behavior is for an unknown color, and so I'm less sure that it needs to be calculated based on the threshold - even though it currently is. Maybe it should be `0.5`, since that's (I think) the mean/median value for standard RGB? If it was only for determining `colorIsBright` it'd be a little easier to reason about, but this is also used in `[STPColorUtils brighterColor:color2:]`

The `testAllColorSpaces` test case probably has the least value. It'll help prevent from regressing #951, and was also added so that I could verify (using the debugger/changing the code) that at least one color (`1 1 1 ...`) in each color space *could* be converted to RGB, and it wasn't hitting the `else return 0.4` branch.

Issue #957, opened 2018-06-05
3DSecure redirect issues


raulabamobile:
## Summary

Facing same behaviour as closed bug #917.

- startRedirectFlow and startSafariViewControllerRedirectFlow: the safari view controller seems to come and go back in a flash without any intervention and brings me back to the app.

- startSafariAppRedirectFlow: when I click on "Return to App", the appdelegate boolean stripeHandled returns as false, even though I can see that the url returned to the app is the one that I set as the return URL while creating the 3D source,

## Code to reproduce

<pre><code>
let cardParams = STPCardParams()
        cardParams.number = "4000000000003063"
        cardParams.expMonth = 12
        cardParams.expYear = 18
        cardParams.cvc = "424"       
        let sourceParams = STPSourceParams.cardParams(withCard: cardParams)
        STPAPIClient.shared().createSource(with: sourceParams) { (source, error) in
            if let s = source, s.cardDetails?.threeDSecure == STPSourceCard3DSecureStatus.required {
                let threeDSecureParams = STPSourceParams.threeDSecureParams(withAmount: 1000, currency: "eur", returnURL: "myexample://stripe-redirect", card: s.stripeID)                             STPAPIClient.shared().createSource(with: threeDSecureParams, completion: { (source, error) in
                    if let error = error {
                        NSLog("ERROR:")
                    } else {
                        let redirectContext = STPRedirectContext(source: source!, completion: { (sourceId, clientSecret, error) in
                            if let s = source, s.cardDetails?.threeDSecure == TPSourceCard3DSecureStatus.required {                                if s.status == STPSourceStatus.chargeable {
                                    NSLog("+++ STATE: CHARGEABLE +++")
                                } else if s.status == STPSourceStatus.consumed {
                                    NSLog("+++ STATE: CONSUMED +++")
                                } else if s.status == STPSourceStatus.canceled {
                                    NSLog("+++ STATE: CANCELED +++")
                                } else if s.status == STPSourceStatus.pending {
                                    NSLog("+++ STATE: PENDING +++")
                                } else if s.status == STPSourceStatus.failed {
                                    NSLog("+++ STATE: FAILED +++")
                                } else if s.status == STPSourceStatus.unknown {
                                    NSLog("+++ STATE: UNKNOW +++")
                                }
                            }
                        })                   
                        redirectContext?.startRedirectFlow(from: self)
                    }
                })
            } else {
                // Send to Server
                NSLog("Send to Server")
            }
        }</code></pre>

[...]

PR #962, opened 2018-06-08
Add `dispatch_async` to `[STPRedirectContext handleWillForegroundNotification]`

joeydong-stripe:
Add `dispatch_async` to `[STPRedirectContext handleWillForegroundNotification]` to allow `[STPRedirectContext handleURLCallback]` to be prioritized

We considered some other approaches and landed on this one:

Option 1:

Delete `STPURLCallback` completely

The downside to this approach is the host app will still get `[AppDelegate application:openURL:options:]` calls. If `STPURLCallback` was gone, the host app won't know whether or not that `[AppDelegate application:openURL:options:]` call should be ignored by their own app's routing logic.

Option 2:
Don't unschedule `STPURLCallback` and let the `STPRedirectContext` stay in the `STPRedirectContextStateCompleted` state

The downside to this approach is the `STPRedirectContext` ends up holding a non-negligible amount of memory in order to just store the `STPRedirectContextStateCompleted` state.

Option 3:

Similarly to the last one, we can avoid unscheduling the `STPURLCallback` for a second or so.

This is very similar to the approach we landed on but would require more significant code changes. Also the reasoning behind how much delay to have is unclear.

### DIFF
--- a/Stripe/STPRedirectContext.m
+++ b/Stripe/STPRedirectContext.m
@@ -193,7 +193,23 @@ typedef void (^STPBoolCompletionBlock)(BOOL success);
 #pragma mark - Private methods -
 
 - (void)handleWillForegroundNotification {
-    stpDispatchToMainThreadIfNecessary(^{
+    // Always `dispatch_async` the `handleWillForegroundNotification` function
+    // call to re-queue the task at the end of the run loop. This is so that the
+    // `handleURLCallback` gets handled first.
+    //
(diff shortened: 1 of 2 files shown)


joeydong-stripe:
Intended solution to the second problem described in https://github.com/stripe/stripe-ios/issues/957

joeydong-stripe:
> Looks good. I never feel good about using dispatch_async, but I think this is our best option.

Yeah same here. Surprising to see a case where it's appropriate.

PR #969, opened 2018-06-15
Reduce NSURLSession memory footprint

csabol-stripe:
## Summary
Don't create a new NSURLSession instance for each ephemeral key request

## Motivation
https://jira.corp.stripe.com/browse/IOS-773

Creating NSURLSessions induces a memory overhead because iOS allocates and saves cache space (ref. https://developer.apple.com/library/archive/qa/qa1727/_index.html). We were also not properly invalidating old sessions as we replaced them, aggravating the problem.

The other option which appears to remove all memory leaks is to use the shared instance of NSURLSession, but we want to keep a private instance. Using an ephemeral configuration (see link below) can also remove any of the reported links, but does not suit our use case.

More references:

https://stackoverflow.com/questions/42780244/swift-3-urlsession-memory-leak
https://stackoverflow.com/questions/28223345/memory-leak-when-using-nsurlsession-downloadtaskwithurl/35757989#35757989
http://footle.org/2015/10/10/fixing-a-swift-memory-leak/

## Testing
I used the standard integration demo and was able to easily reproduce the reported leaks just by viewing an item. With the changes all leaks attributed to the Stripe library were gone.

### DIFF
--- a/Stripe/STPAPIClient.m
+++ b/Stripe/STPAPIClient.m
@@ -146,21 +146,6 @@ static NSString * const FileUploadURL = @"https://uploads.stripe.com/v1/files";
     return [additionalHeaders copy];
 }
 
-- (void)setUrlSession:(NSURLSession *)urlSession
-{
-    if (urlSession != _urlSession) {
-        [_urlSession invalidateAndCancel];
-        _urlSession = urlSession;
-    }
-}
-
-- (void)setApiKey:(NSString *)apiKey {
-    _apiKey = apiKey;
-
-    // Regenerate url session configuration
-//    self.urlSession = [NSURLSession sessionWithConfiguration:[self sessionConfiguration]];
-}
-
 - (void)setPublishableKey:(NSString *)publishableKey {
     self.configuration.publishableKey = [publishableKey copy];
     self.apiKey = [publishableKey copy];
@@ -168,13 +153,6 @@ static NSString * const FileUploadURL = @"https://uploads.stripe.com/v1/files";
 
 - (NSString *)publishableKey {
     return self.configuration.publishableKey;
-}
-
-- (void)setStripeAccount:(NSString *)stripeAccount {
-    _stripeAccount = stripeAccount;
-
-    // Regenerate url session configuration
-//    self.urlSession = [NSURLSession sessionWithConfiguration:[self sessionConfiguration]];
 }
 
 - (void)createTokenWithParameters:(NSDictionary *)parameters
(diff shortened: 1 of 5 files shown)


Stripe/STPAPIClient.m
```diff
@@ -160,7 +174,7 @@ - (void)setStripeAccount:(NSString *)stripeAccount {
     _stripeAccount = stripeAccount;
 
     // Regenerate url session configuration
-    self.urlSession = [NSURLSession sessionWithConfiguration:[self sessionConfiguration]];
+//    self.urlSession = [NSURLSession sessionWithConfiguration:[self sessionConfiguration]];
```
danj-stripe: These should be deleted, not commented out.

Stripe/STPAPIClient.m
```diff
@@ -124,27 +124,41 @@ - (instancetype)initWithConfiguration:(STPPaymentConfiguration *)configuration {
...
     additionalHeaders[@"Stripe-Version"] = APIVersion;
     additionalHeaders[@"Authorization"] = [@"Bearer " stringByAppendingString:self.apiKey ?: @""];
     additionalHeaders[@"Stripe-Account"] = self.stripeAccount;
-    NSURLSessionConfiguration *sessionConfiguration = [NSURLSessionConfiguration defaultSessionConfiguration];
-    sessionConfiguration.HTTPAdditionalHeaders = additionalHeaders;
-    return sessionConfiguration;
+    return [additionalHeaders copy];
+}
+
+- (void)setUrlSession:(NSURLSession *)urlSession
```
danj-stripe: Is this needed/ever called? I saw you made the `urlSession` `readonly`.

Tests/Tests/STPAPIClientTest.m
```diff
@@ -63,7 +63,7 @@ - (void)testInitWithConfiguration {
     STPAPIClient *sut = [[STPAPIClient alloc] initWithConfiguration:config];
     XCTAssertEqualObjects(sut.publishableKey, config.publishableKey);
     XCTAssertEqualObjects(sut.stripeAccount, config.stripeAccount);
-    NSString *accountHeader = sut.urlSession.configuration.HTTPAdditionalHeaders[@"Stripe-Account"];
+    NSString *accountHeader = [sut configuredRequestForURL:[NSURL URLWithString:@"https://www.stripe.com"]].allHTTPHeaderFields[@"Stripe-Account"];
```
danj-stripe: Nit: Might be nice to extract this into a helper method on the test case.

PR #976, opened 2018-06-26
Make `STPSource` conform to `STPPaymentMethod`, exposing labels & images

joeydong-stripe:
## Summary
Adds `STPPaymentMethod` protocol implementation for `STPSource`. In particular, it enables `source.label`, `source.image`, and `source.templateImage`.

There's potentially room to improve the labels & images - right now non-cards just mirror whatever the Source type is.

## Motivation
Issue originally raised after extensive investigation between Joey & Fred. See individual commits. More details in IOS-816.

## Testing
Automated tests added for the new functionality.


Stripe/STPSource.m
```diff
@@ -289,15 +287,38 @@ - (UIImage *)templateImage {
...
+        case STPSourceTypeThreeDSecure:
+            return @"3D Secure";
+        case STPSourceTypeAlipay:
+            return @"Alipay";
+        case STPSourceTypeP24:
+            return @"P24";
+        case STPSourceTypeEPS:
+            return @"EPS";
+        case STPSourceTypeMultibanco:
+            return @"Multibanco";
```
danj-stripe: Do any of these need to be localizable strings?

My first thought was no, because they're brand/product names. On the other hand, I suspect it'd be weird to see "Alipay" instead of "支付宝" (according to wikipedia) in an app localized to Chinese.

Stripe/STPSource.m
```diff
@@ -289,15 +287,38 @@ - (UIImage *)templateImage {
...
+        case STPSourceTypeUnknown:
+            return [STPCard stringFromBrand:STPCardBrandUnknown];
```
danj-stripe: wdyt about using `@"Unknown"` here instead of this layer of indirection with `STPCardBrandUnknown`?

CHANGELOG.md
```diff
@@ -1,3 +1,5 @@
+* Adds `STPPaymentMethod` protocol implementation for `STPSource` [#976](https://github.com/stripe/stripe-ios/pull/976)
```
danj-stripe: wdyt about: mentioning what this accomplishes? I suspect this won't mean much to our users

Stripe/STPSource.m
```diff
@@ -289,15 +287,38 @@ - (UIImage *)templateImage {
...
+        case STPSourceTypeGiropay:
+            return @"Giropay";
+        case STPSourceTypeIDEAL:
+            return @"iDEAL";
+        case STPSourceTypeSEPADebit:
+            return @"SEPA Direct Debit";
+        case STPSourceTypeSofort:
+            return @"SOFORT";
+        case STPSourceTypeThreeDSecure:
+            return @"3D Secure";
```
danj-stripe: Did you consider including more details from the source in the label strings?

3DS is an obvious one: it's wrapping a card, and the customer is probably more likely to recognize the card that's used in the Source than the fact that it was charged through 3DS. It'd also be hard for them to disambiguate if they had more than one.

I suspect the `STPSourceParam` factory methods are good references for what fields might be interesting on a per-SourceType basis.

Stripe/STPSource.m
joeydong-stripe: Good point, we should probably wrap them in NSLocalizedString calls in case users want to localize it themselves in a particular way.

Stripe/STPSource.m
joeydong-stripe: Yeah why not. I first put this here because I was focused on card source types but it is a little out of place in the `STPSourceTypeUnknown` block.

Stripe/STPSource.m
joeydong-stripe: Yeah I did want to include more information if available but it looks like we don't make API Version guarantees on what information is included in the source specific blocks. If I'm wrong about the guarantees, then yeah we should add them appropriately.

Stripe/Resources/Localizations/en.lproj/Localizable.strings
```diff
@@ -46,34 +55,49 @@
...
+/* Source type brand name */
+"P24" = "P24";
+
 /* Title for Payment Method screen */
 "Payment Method" = "Payment Method";
 
 /* Caption for Phone field on address form */
 "Phone" = "Phone";
 
-/* Short string for postal code (text used in non-US countries). Please keep to 8 characters or less if possible. */
```
danj-stripe: Were these character count suggestions removed because you re-ran `genstrings`? I wonder if we need to move them into the `STPLocalizedString` call

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## XX.Y.Z
```
csabol-stripe: this is just a placeholder for when we make the next release?

danj-stripe: Yeah. I think it's better for us to update the CHANGELOG during the release process, to avoid conflicts while merging PRs, but sometimes we do it during the PR.

PR #978, opened 2018-06-29
Fix compile error in test target when running on 32 bit device

danj-stripe:
## Summary
Added a cast to `CGFloat`

## Motivation
Warning about implicit conversion that our compiler settings elevate to an error.

## Testing
Compiled/ran test suite on both iPhone 5 & iPhone 6

### DIFF
--- a/Tests/Tests/STPColorUtilsTest.m
+++ b/Tests/Tests/STPColorUtilsTest.m
@@ -31,7 +31,7 @@
         CGColorRelease(cgcolor);
     }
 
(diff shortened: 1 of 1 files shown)


danj-stripe:
This test was recently added by me, in #954

joeydong-stripe:
https://github.com/stripe/stripe-ios/pull/979/files#diff-26ce4e60c64c9c76adc4f3872cf284f9

danj-stripe:
@joeydong-stripe I noticed that too. I figured it could be a separate PR, since it's a separate issue.

PR #979, opened 2018-06-29
Expand Test Execution Coverage

joeydong-stripe:
Seems like Travis CI picks up our builds relatively quickly now so I think it's an appropriate time to expand our iOS version test coverage. Also would catch problems like https://github.com/stripe/stripe-ios/pull/978. See individual commits.

### DIFF
--- a/Tests/Tests/STPColorUtilsTest.m
+++ b/Tests/Tests/STPColorUtilsTest.m
@@ -75,12 +75,12 @@
 
 - (void)testAllColorSpaces {
     // block to create & check brightness of color in a given color space
-    void (^testColorSpace)(NSString *, BOOL) = ^(NSString *colorSpaceName, BOOL expectedToBeBright) {
+    void (^testColorSpace)(const CFStringRef, BOOL) = ^(const CFStringRef colorSpaceName, BOOL expectedToBeBright) {
         // this a bright color in almost all color spaces
         CGFloat components[] = {1.0, 1.0, 1.0, 1.0, 1.0, 1.0};
 
         UIColor *color = nil;
-        CGColorSpaceRef colorSpace = CGColorSpaceCreateWithName((__bridge CFStringRef)colorSpaceName);
+        CGColorSpaceRef colorSpace = CGColorSpaceCreateWithName(colorSpaceName);
 
         if (colorSpace) {
             CGColorRef cgcolor = CGColorCreate(colorSpace, components);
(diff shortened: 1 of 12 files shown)


Tests/Tests/STPColorUtilsTest.m
```diff
@@ -103,38 +103,42 @@ - (void)testAllColorSpaces {
...
+                                                                           (__bridge NSString *)kCGColorSpaceExtendedLinearGray,
+                                                                           ]];
+    }
 
-    int colorSpaceCount = sizeof(colorSpaceNames) / sizeof(colorSpaceNames[0]);
-    for (int i = 0; i < colorSpaceCount; ++i) {
+    for (NSString *name in colorSpaceNames) {
         // CMYK is the only one where all 1's results in a dark color
-        testColorSpace(colorSpaceNames[i], colorSpaceNames[i] != kCGColorSpaceGenericCMYK);
+        testColorSpace(name, name != (__bridge NSString *)kCGColorSpaceGenericCMYK);
```
joeydong-stripe: Interesting, I would have assumed we needed to use `[NSString isEqualToString:]` but the tests are passing anyway?

Tests/Tests/STPColorUtilsTest.m
```diff
@@ -103,38 +103,42 @@ - (void)testAllColorSpaces {
...
+      (__bridge NSString *)kCGColorSpaceExtendedGray,
+      (__bridge NSString *)kCGColorSpaceExtendedSRGB,
+      (__bridge NSString *)kCGColorSpaceGenericRGBLinear,
+      (__bridge NSString *)kCGColorSpaceExtendedLinearSRGB,
+      (__bridge NSString *)kCGColorSpaceGenericGrayGamma2_2,
+      ];
+
+    if (@available(iOS 10.0, *)) {
+        colorSpaceNames = [colorSpaceNames arrayByAddingObjectsFromArray:@[
+                                                                           (__bridge NSString *)kCGColorSpaceLinearGray,
```
danj-stripe: Instead of adding all this (ugly?) bridging, since it was only two iOS 10 cases, did you consider just following the pattern that the iOS 11-only case `kCGColorSpaceGenericLab` took?

Tests/Tests/STPColorUtilsTest.m
danj-stripe: I'm less confident about keeping this as a pointer equality check with the bridging involved. I think it's probable that `CFStringRef` is always toll-free bridged to `NSString` and this'll be fine, but I think it's also plausible that some future runtime changes that.

ci_scripts/run_builds.sh
```diff
@@ -0,0 +1,63 @@
...
+
+# Execute sample app builds (iPhone 6, iOS 11.x)
+info "Executing sample app builds (iPhone 6, iOS 11.x)..."
+
+xcodebuild build \
+  -workspace "Stripe.xcworkspace" \
+  -scheme "Standard Integration (Swift)" \
+  -sdk "iphonesimulator" \
+  -destination "platform=iOS Simulator,name=iPhone 6,OS=11.2" \
+  | xcpretty
```
danj-stripe: Was it intentional to drop the `-c` flag that colorizes the output?

ci_scripts/run_tests.sh
```diff
@@ -1,10 +1,98 @@
...
+carthage_exit_code="$?"
+
+if [[ "${carthage_exit_code}" != 0 ]]; then
+  die "Executing carthage failed with status code: ${carthage_exit_code}"
+fi
+
+# Execute tests (iPhone 6 @ iOS 11.2)
+info "Executing tests (iPhone 6 @ iOS 11.2)..."
+
+xcodebuild clean test \
```
danj-stripe: Why did you combine the build & test steps? They were intentionally separated last year: https://github.com/stripe/stripe-ios/commit/eb8009acf315e4b5c5b7fdfe2cba46fde9236ec1

I know I've run into this problem in the past (and have the radar to prove it), but I don't know if it's still an issue in Xcode 9.

ci_scripts/run_tests.sh
```diff
@@ -1,10 +1,98 @@
...
+# - Skips snapshot tests because they're recorded for the iPhone 6 on the newest iOS version only
+# - Not sure why tests STPImageLibraryTest fail on older iOS versions
+info "Executing tests on 32-bit device  (iPhone 4s @ iOS 9.x)..."
+
+xcodebuild clean test \
+  -workspace "Stripe.xcworkspace" \
+  -scheme "StripeiOS" \
+  -configuration "Debug" \
+  -sdk "iphonesimulator" \
+  -destination "platform=iOS Simulator,name=iPhone 4s,OS=9.3" \
```
danj-stripe: What happens when you try to stick this in the previous `xcodebuild` command? Does it fail because `Debug` only builds for the active architecture?

Did you consider DRYing this up instead of duplicating so many parameters?

Tests/Tests/STPColorUtilsTest.m
joeydong-stripe: Yeah I agree it's not smart to leave it as a pointer check. I'm more wondering why it still works. Maybe this constant is always mapped to the same memory address. Anyway, I'll switch it to the usual `isEqualToString`.

Tests/Tests/STPColorUtilsTest.m
joeydong-stripe: Oh yeah, I was so focused on struggling with this Core Foundation syntax I forgot to try that.

ci_scripts/run_builds.sh
joeydong-stripe: Yeah, to me, the default behavior seems to works so why bother specifying it:

screen shot 2018-07-09 at 3 05 30 pm

ci_scripts/run_tests.sh
joeydong-stripe: i combined it to reduce code duplication, and it seems to work? i think if we need it again in the future, we should leave the reasoning as a comment in the code because a bum like me is going to miss the commit blame

ci_scripts/run_tests.sh
joeydong-stripe: > What happens when you try to stick this in the previous xcodebuild command? Does it fail because Debug only builds for the active architecture?

Oops, forgot to leave a reason comment but yeah that's why. Never figured it out but your guess seems reasonable! The code duplication is definitely annoying. I haven't tried parameters in a variable before but I'll look into it.

joeydong-stripe: Duplication doesn't look too bad after adding the `ONLY_ACTIVE_ARCH=NO` usage

PR #980, opened 2018-06-29
Generalize STPRedirectContext's implementation to not rely on the STPSource

danj-stripe:
## Summary
Making `STPRedirectContext` less coupled to `STPSource`. Instead of storing & using the
Source and Source-specific completion block, the redirect context now stores just the 
URLs needed and a general closure to call when redirecting is finished.

This does *not* update any of the class documentation, nor does it change the external
API at all.

## Motivation
Clears the path for `STPRedirectContext` to be used with PaymentIntents, which similarly
want to redirect to a URL and wait for the user to return to the app.

## Testing
There's a pretty complete set of tests in `STPRedirectContextTest`, and they still pass.
Also, ran through the redirect flow in the Custom Integration sample app.

### DIFF
--- a/Stripe/STPRedirectContext.m
+++ b/Stripe/STPRedirectContext.m
@@ -47,21 +47,38 @@ typedef void (^STPBoolCompletionBlock)(BOOL success);
 
     if (source.flow != STPSourceFlowRedirect
         || !(source.status == STPSourceStatusPending ||
-             source.status == STPSourceStatusChargeable)
-        || source.redirect.returnURL == nil
-        || (source.redirect.url == nil
-            && [STPRedirectContext nativeRedirectURLForSource:source] == nil)) {
+             source.status == STPSourceStatusChargeable)) {
+        return nil;
+    }
+
+    self = [self initWithNativeRedirectUrl:[STPRedirectContext nativeRedirectURLForSource:source]
+                               redirectUrl:source.redirect.url
+                                 returnUrl:source.redirect.returnURL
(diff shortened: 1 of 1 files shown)


Stripe/STPRedirectContext.m
```diff
@@ -21,8 +22,15 @@
 typedef void (^STPBoolCompletionBlock)(BOOL success);
 
 @interface STPRedirectContext () <SFSafariViewControllerDelegate, STPURLCallbackListener>
-@property (nonatomic, copy) STPRedirectContextCompletionBlock completion;
-@property (nonatomic, strong) STPSource *source;
+/// Optional URL for a native app. This is passed directly to `UIApplication openURL:`, and if it fails this class falls back to `redirectUrl`
+@property (nonatomic, nullable, copy) NSURL *nativeRedirectUrl;
+/// The URL to redirect to, assuming `nativeRedirectUrl` is nil or fails to open. Cannot be nil if `nativeRedirectUrl` is.
+@property (nonatomic, nullable, copy) NSURL *redirectUrl;
```
joeydong-stripe: i would suggest calling it `webRedirectUrl` to reduce confusion with `nativeRedirectUrl`

joeydong-stripe:
Otherwise, looks reasonable. Will check again once you get tests passing :)

Stripe/STPRedirectContext.m
danj-stripe: I'm not totally comfortable doing that. It's *almost* certainly a web url, but it doesn't actually have to be.

PR #985, opened 2018-07-06
Add STPPaymentIntent, and method to retrieve from the API given a clientSecret

danj-stripe:
## Summary
Adding support for `/v1/payment_intent/:id`

## Motivation
IOS-792

## Testing
Added some tests. This is the first in a series of PRs that'll add full PaymentIntent
support, and this was tested in the sample app too.

PR #986, opened 2018-07-06
Add method to API that takes STPPaymentIntentParams and confirms them

danj-stripe:
## Summary
Adding support for `/v1/payment_intents/:id/confirm`

This method takes a STPPaymentIntentParams object, which may have STPSourceParams or a
Source id, among other fields.

## Motivation
IOS-792

## Testing
Added tests. This is part of a series of PRs, starting with #985 (which this relies on),
and this was also tested in the sample app.

### DIFF
--- a/Stripe/PublicHeaders/STPPaymentIntentParams.h
+++ b/Stripe/PublicHeaders/STPPaymentIntentParams.h
@@ -1,0 +1,87 @@
+//
+//  STPPaymentIntentParams.h
+//  Stripe
+//
+//  Created by Daniel Jackson on 7/3/18.
+//  Copyright © 2018 Stripe, Inc. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
+
+#import "STPFormEncodable.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@class STPSourceParams;
+
+/**
+ An object representing parameters used to confirm a PaymentIntent object.
+
+ A PaymentIntent must have a Source associated in order to successfully confirm it.
+
+ That Source can either be:
+
+ - created during confirmation, by passing in a `STPSourceParams` object in the `sourceParams` field
+ - a pre-existing Source can be associated by passing its id in the `sourceId` field
+ - or already set via your backend, either when creating or updating the PaymentIntent
+
+ @see https://stripe.com/docs/api#confirm_payment_intent
+ */
+@interface STPPaymentIntentParams : NSObject<STPFormEncodable>
+
+/**
+ Initialize this `STPPaymentIntentParams` with a `clientSecret`, which is the only required
+ field.
+
+ @param clientSecret the client secret for this PaymentIntent
+ */
+- (instancetype)initWithClientSecret:(NSString *)clientSecret;
+
+/**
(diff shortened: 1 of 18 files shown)


danj-stripe:
This also needs review.

Stripe/STPPaymentIntentParams.m
```diff
@@ -0,0 +1,74 @@
...
+//  STPPaymentIntentParams.m
+//  Stripe
+//
+//  Created by Daniel Jackson on 7/3/18.
+//  Copyright © 2018 Stripe, Inc. All rights reserved.
+//
+
+#import "STPPaymentIntentParams.h"
+#import "STPPaymentIntent+Private.h"
+
```
csabol-stripe: I find it useful to include NS_ASSUME_NONNULL_BEGIN/END in the .m files as well. It makes sure your implementation matches your interface!

Tests/Tests/STPPaymentIntentParamsTest.m
```diff
@@ -0,0 +1,66 @@
...
+@interface STPPaymentIntentParamsTest : XCTestCase
+
+@end
+
+@implementation STPPaymentIntentParamsTest
+
+- (void)testInit {
+    for (STPPaymentIntentParams *params in @[[[STPPaymentIntentParams alloc] initWithClientSecret:@"secret"],
+                                             [[STPPaymentIntentParams alloc] init],
+                                             [STPPaymentIntentParams new],
```
csabol-stripe: no personal preference but mixing `alloc] init]` and `new]` right next to each other reads a bit weird :P

csabol-stripe: Is this explicit to test that `new` actually calls `alloc] init]`? Seems like that's out of scope of something we need to test

Stripe/PublicHeaders/STPPaymentIntentParams.h
```diff
@@ -0,0 +1,87 @@
...
+ */
+@property (nonatomic, copy, nullable, readwrite) NSString *receiptEmail;
+
+/**
+ `@YES` to save this PaymentIntent’s Source to the associated Customer,
+ if the Source is not already attached.
+
+ This should be a boolean NSNumber, so that it can be `nil`
+ */
+@property (nonatomic, strong, nullable, readwrite) NSNumber *saveSourceToCustomer;
```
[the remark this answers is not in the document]
csabol-stripe: 👍

Stripe/STPPaymentIntentParams.m
```diff
@@ -0,0 +1,74 @@
...
+@implementation STPPaymentIntentParams
```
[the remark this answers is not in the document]
csabol-stripe: yeah noticed that later so that makes sense

Tests/Tests/STPPaymentIntentParamsTest.m
csabol-stripe: 👍

csabol-stripe:
> I think my answer to most of these is "that's the way the existing code does things", so my preference is to land as-is and write up any improvements as separate JIRAs to be prioritized against our other work, but let me know if you think any of them are high enough priority to include.

Sounds good to me. Land away!

PR #987, opened 2018-07-06
Add support for STPPaymentIntent to STPRedirectContext

danj-stripe:
## Summary
Adding init method on STPRedirectContext that takes an STPPaymentIntent.

This relies on #985 (adds STPPaymentIntent) and #980 (refactored STPRedirectContext to
make this possible), and so it has merge commits from each of those branches.

## Motivation
IOS-792

## Testing
STPRedirectContext has extensive tests, which (still) need to be updated for this new feature

PR #988, opened 2018-07-06
Add 3DS example VC that uses PaymentIntents to charge the user

danj-stripe:
## Summary
Add 3DS example VC that uses PaymentIntents to charge the user

This is built on top of the existing 3DS example VC (using the same UI),
but intercepts the `pay` button response to use a PaymentIntent instead.

This relies on #986 and #987.

## Motivation
IOS-792

## Testing
Ran the example VC and played around with it. This was used as the test app for the previous
PRs.

### DIFF
--- a/Example/Custom Integration (ObjC)/BrowseExamplesViewController.m
+++ b/Example/Custom Integration (ObjC)/BrowseExamplesViewController.m
@@ -96,7 +96,7 @@
         NSError *error = [NSError errorWithDomain:StripeDomain
                                              code:STPInvalidRequestError
                                          userInfo:@{NSLocalizedDescriptionKey: @"You must set a backend base URL in Constants.m to create a charge."}];
-        completion(STPBackendChargeResultFailure, error);
+        completion(STPBackendResultFailure, error);
         return;
     }
 
@@ -121,7 +121,7 @@
                                                                                       userInfo:@{NSLocalizedDescriptionKey: @"There was an error connecting to your payment backend."}];
                                                           }
                                                           if (error) {
-                                                              completion(STPBackendChargeResultFailure, error);
+                                                              completion(STPBackendResultFailure, error);
                                                           } else {
                                                               completion(STPBackendChargeResultSuccess, nil);
                                                           }
--- a/Example/Custom Integration (ObjC)/BrowseExamplesViewController.m
+++ b/Example/Custom Integration (ObjC)/BrowseExamplesViewController.m
@@ -13,6 +13,7 @@
 #import "Constants.h"
 #import "SofortExampleViewController.h"
 #import "ThreeDSExampleViewController.h"
+#import "ThreeDSPaymentIntentExampleViewController.h"
 
 /**
  This view controller presents different examples, each of which demonstrates creating a payment using a different payment method.
@@ -36,24 +37,27 @@
 }
 
 - (NSInteger)tableView:(UITableView *)tableView numberOfRowsInSection:(NSInteger)section {
-    return 4;
+    return 5;
 }
 
 - (UITableViewCell *)tableView:(UITableView *)tableView cellForRowAtIndexPath:(NSIndexPath *)indexPath {
     UITableViewCell *cell = [UITableViewCell new];
     switch (indexPath.row) {
         case 0:
-            cell.textLabel.text = @"Card";
-            break;
+        cell.textLabel.text = @"Card";
+        break;
         case 1:
-            cell.textLabel.text = @"Card + 3DS";
-            break;
+        cell.textLabel.text = @"Card + 3DS";
+        break;
         case 2:
-            cell.textLabel.text = @"Apple Pay";
-            break;
+        cell.textLabel.text = @"Apple Pay";
+        break;
         case 3:
-            cell.textLabel.text = @"Sofort";
-            break;
+        cell.textLabel.text = @"Sofort";
+        break;
+        case 4:
+        cell.textLabel.text = @"Card + 3DS + PaymentIntent";
+        break;
     }
     return cell;
 }
@@ -81,6 +85,12 @@
         }
         case 3: {
             SofortExampleViewController *exampleVC = [SofortExampleViewController new];
+            exampleVC.delegate = self;
+            viewController = exampleVC;
+            break;
+        }
+        case 4: {
+            ThreeDSPaymentIntentExampleViewController *exampleVC = [ThreeDSPaymentIntentExampleViewController new];
             exampleVC.delegate = self;
             viewController = exampleVC;
             break;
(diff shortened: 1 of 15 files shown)


danj-stripe:
Since this PR relies on several previous ones, they're all merged into here. The changes specific to this PR start at https://github.com/stripe/stripe-ios/pull/988/commits/33fd12c8f5f9f77a634621be6c17d2dabd780a77

Example/Custom Integration (ObjC)/BrowseExamplesViewController.m
```diff
@@ -36,24 +37,27 @@ - (NSInteger)numberOfSectionsInTableView:(UITableView *)tableView {
...
+        cell.textLabel.text = @"Apple Pay";
+        break;
         case 3:
-            cell.textLabel.text = @"Sofort";
-            break;
+        cell.textLabel.text = @"Sofort";
+        break;
+        case 4:
+        cell.textLabel.text = @"Card + 3DS + PaymentIntent";
+        break;
```
csabol-stripe: does `ctrl+i` give this indentation for obj-c now?

Example/Custom Integration (ObjC)/BrowseExamplesViewController.m
```diff
@@ -123,7 +133,69 @@ - (void)createBackendChargeWithSource:(NSString *)sourceID completion:(STPSource
                                                           if (error) {
                                                               completion(STPBackendResultFailure, error);
                                                           } else {
-                                                              completion(STPBackendChargeResultSuccess, nil);
+                                                              completion(STPBackendResultSuccess, nil);
+                                                          }
+                                                      }];
+
+    [uploadTask resume];
```
csabol-stripe: why was this added?

Example/Custom Integration (ObjC)/BrowseExamplesViewController.m
```diff
@@ -123,7 +133,69 @@ - (void)createBackendChargeWithSource:(NSString *)sourceID completion:(STPSource
...
+}
+
+
+/**
+ Ask the example backed to create a PaymentIntent with the specified amount.
```
csabol-stripe: typo, `backend`

Example/Custom Integration (ObjC)/BrowseExamplesViewController.m
```diff
@@ -123,7 +133,69 @@ - (void)createBackendChargeWithSource:(NSString *)sourceID completion:(STPSource
...
+
+ The implementation of this function is not interesting nor relevant to using PaymentIntents. The
```
csabol-stripe: grammar nit: `not interesting or relevant`

csabol-stripe: https://www.quickanddirtytips.com/education/grammar/when-use-nor?page=1

Example/Custom Integration (ObjC)/ThreeDSPaymentIntentExampleViewController.m
```diff
@@ -0,0 +1,133 @@
...
+                [self.delegate exampleViewController:self didFinishWithError:error];
+                return;
+            }
+
+            if (paymentIntent.status == STPPaymentIntentStatusRequiresSourceAction) {
+                self.redirectContext = [[STPRedirectContext alloc] initWithPaymentIntent:paymentIntent returnUrl:returnUrl completion:^(NSString * _Nonnull clientSecret, NSError * _Nullable error) {
+                    if (error) {
+                        [self.delegate exampleViewController:self didFinishWithError:error];
+                    }
+                    else {
```
csabol-stripe: this file is mixing `} else {` and `}\n else {` Do we have a standard in our sample code? I'd vote for `} else {`

Example/Custom Integration (ObjC)/ThreeDSPaymentIntentExampleViewController.m
```diff
@@ -0,0 +1,133 @@
...
+                        [stripeClient retrievePaymentIntentWithClientSecret:clientSecret completion:^(STPPaymentIntent * _Nullable paymentIntent, NSError * _Nullable error) {
+                            if (error) {
+                                [self.delegate exampleViewController:self didFinishWithError:error];
+                            } else {
+                                [self finishWithStatus:paymentIntent.status];
+                            }
+                        }];
+                    }
+                    self.redirectContext = nil; // break retain cycle
```
csabol-stripe: so it's an intentional retain cycle? Can we call that out in like L72?

csabol-stripe: Is there a not retain cycle way to accomplish this? Doesn't feel like we should be demonstrating to users that a retain cycle is the "right" way to do something

Example/Custom Integration (ObjC)/ThreeDSPaymentIntentExampleViewController.m
joeydong-stripe: We encoded `}\n else {` in our style guide a long time ago but I'm coming around to doing `} else {` everywhere. Should open a PR that fixes this across the codebase.

joeydong-stripe: https://github.com/stripe/stripe-ios/blob/master/STYLEGUIDE.md#control-flow

Example/Custom Integration (ObjC)/BrowseExamplesViewController.m
danj-stripe: This is pre-existing code, it's just the Github diff algorithm is being weird. Both methods end the same way (L202 has the same call).

Example/Custom Integration (ObjC)/ThreeDSPaymentIntentExampleViewController.m
danj-stripe: I don't think I've ever seen `}\n else {` in a style guide, and it wouldn't be my choice, but I try to adhere to it. Thanks for catching the mistake.

Example/Custom Integration (ObjC)/ThreeDSPaymentIntentExampleViewController.m
danj-stripe: I think it starts on L79 - `self` needs to retain the `redirectContext` (doesn't work otherwise), and none of the examples are using weak self, so the `redirectContext` retains `self`.

`SofortExampleViewController` and `ThreeDSExampleViewController` do the same thing, but without this note.

It's possible that it's also necessary so that `STPRedirectContext.dismissPresentedViewController` is called, and it might be intentional that we're releasing the memory early.

IMO our example code does not do a great job demonstrating best practices, instead preferring to do things compactly with very little indirection, so that "all the pieces" are in one place.

I think there are improvements that could be made, and I think there's value in keeping the examples in sync.

Example/Custom Integration (ObjC)/BrowseExamplesViewController.m
danj-stripe: It was, and I didn't like it but couldn't get it to do anything different. Now it looks like Xcode has changed its mind. So there must have been something I'd missed, maybe an error in a header or something.

danj-stripe:
In order to fix conflicts and pull in latest changes from #986 and #987, rebased onto master, merged in latest from those PRs, and then picked the 4 commits for this PR. Starting after the octopus merge (6b79bb7) has the changes to the example code, and they're the same as reviewed by csabol, plus f035539 with updates

Example/Custom Integration (ObjC)/ThreeDSPaymentIntentExampleViewController.m
csabol-stripe: makes sense. Thanks

danj-stripe:
This now includes #994 and #995, and has been rebased onto master. No new or un-reviewed code

PR #992, opened 2018-07-12
Fix invalid JSON error when deleting Card from Customer

danj-stripe:
## Summary
We use the same code path to delete/detach Card + Source + any `STPSourceProtocol` from
a Customer. Depending on the type of the object, the response varies, but the only field
that the response always has is `id`.

This updates the private `STPAPI deleteSource:fromCustomerUsingKey:completion:` method
to use a new class, `STPGenericStripeObject`, as the deserializer. Now the SDK doesn't
incorrectly report failures.

## Motivation
IOS-798

## Testing
Minor updates to the tests, but this was best tested manually in the Standard Integration.

1. In `CheckoutViewController.init`, change `config.createCardSources` to `false`.
2. In `[STPPaymentMethodsInternalViewController tableView:commitEditingStyle:forRowAtIndexPath:]`,
add a completion block to the `[self.apiAdapter detachSourceFromCustomer:completion:]` call.
I chose to log the error, but you could also set a breakpoint and inspect it.
3. Run standard integration
4. Add *two* cards in the payment information
5. Delete the non-default card.

We don't have an automated way to create customers & attach sources, because that requires
a secret key.

We also don't have an existing test pattern to mock network responses to `STPAPIClient`.


danj-stripe:
cc @bg-stripe: thanks for chatting about this.

PR #994, opened 2018-07-13
Use `STPPaymentIntent.returnUrl` instead of requiring app developer to provide the url

danj-stripe:
## Summary
When initially implemented, `PaymentIntent`'s `return_url` wasn't exposed via the
`client_secret`. It is now, and this simplifies the API.

## Motivation
This'll simplify the API and process for integrating. It makes it harder to make a mistake
too.

## Testing
Augmented tests for this

### DIFF
--- a/MIGRATING.md
+++ b/MIGRATING.md
@@ -1,5 +1,9 @@
 ## Migration Guides
 
+### Migrating from versions < XX.X.X
+ * The SDK now supports PaymentIntents with `STPPaymentIntent`, which use `STPRedirectContext` in the same way that `STPSource` does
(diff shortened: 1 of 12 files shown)


danj-stripe:
This is a proposed enhancement to #987. We could ship without it, but I think we want to get this into the release. It does cause churn in the example code and docs though.

danj-stripe:
I merged #995 into the branch for #988, and so this is part of that #988 too.

Got myself into a tangled little mess of branches, with the intention things would be easier to review piecemeal. Instead of sticking with that approach, going to bundle them up

PR #995, opened 2018-07-13
Update example code for #994: `returnUrl` from `STPPaymentIntent`

danj-stripe:
## Summary
Updates the example code from #988 with the proposed change from #994.

This *also* attaches the `returnUrl` to the PaymentIntent via `/confirm` instead of sending it to the backend during PaymentIntent creation.

## Motivation
Simplifies the example code.

## Testing
Made sure it worked in the Custom Integration example app.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ env:
   - TEST_TYPE=installation_cocoapods_frameworks
   - TEST_TYPE=lint
   - TEST_TYPE=tests
+  - TEST_TYPE=builds
   - TEST_TYPE=analyzer
   - TEST_TYPE=documentation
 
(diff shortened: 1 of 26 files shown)


danj-stripe:
The only part of this that needs reviewing is be688ff (and later, if necessary) - the rest is just merge spam: This relies on #994 and #988, and so the PR comes with all those commits via merges (including stuff from master)

danj-stripe:
The only part of this that needs reviewing is dcedc16 (and later, if necessary) - the rest is just merge spam: This relies on #994 and #988, and so the PR comes with all those commits via the merge commit be688ff (including stuff from master)

Issue #968, opened 2018-06-14
Standard integration Apple Pay cancel while processing


yliu342:
## Summary
Hi I'm using the standard integration of Stripe for Apple Pay. It worked well for the happy path and everything is good. But there is one scenario where I believe it is not handled by the SDK and the details are the following:

- User selected Apple Pay as payment method
- `PKPaymentAuthorizationViewController` is presented by the SDK. 
- User authorised the payment by using TouchID or FaceID
- Apple Pay will show processing, and `STPPaymentContextDelegate` method `didCreatePaymentResult` is called.
- Make service call to the backend to make a charge in side `didCreatePaymentResult`  method.
- Before the backend return any response, user click cancel button on the apple pay popup.
- `PKPaymentAuthorizationViewController` is dismissed `STPPaymentContextDelegate` method `didFinishWith status` is called with `userCancellation`.
- Response from backend returned and the charge is successful. Calling `completion(nil)` inside `didCreatePaymentResult` but `didFinishWith status` will not be called again after it received `userCancellation`

In this case, the user has authorised the payment and the payment went through correctly but  `didFinishWith status` does not show the correct status.

I have tried to see how Apple handles it in App Store and it seems after the user is authorised the payment by TouchID or FaceID, click on the cancel button does nothing. 

I would suggest observe the authorization status of the payment in `PKPaymentAuthorizationViewControllerDelegate` and if the payment has been authorised then click on cancel does nothing (same behaviour as App Store)

## iOS version
11.3

## Installation method
Cocoapods

## SDK version
12.1

yliu342:
Thank you @csabol-stripe

csabol-stripe:
I reviewed the Apple Human Interface Guidelines (https://developer.apple.com/design/human-interface-guidelines/apple-pay/architecture/error-handling/#data-validation) and found

> A user-driven event like a cancellation or a system-driven event like a timeout could cause an interruption in the payment flow, resulting in the payment sheet being dismissed. When such an event occurs, you must cancel any in-progress payment.

yliu342:
That make sense but I don't think Stripe back end has an api to cancel/revert a charge request. So in scenario where a user clicked on apple pay and authorized by using TouchID/FaceID then during process payment clicked cancel. The charge request has already being made to the Stripe backend.  The app can send a request to our backend system to cancel the charge but I don't think there is a way for our backend system to tell Stripe to cancel that charge request. (Please correct me if I'm wrong)

[...]

PR #996, opened 2018-07-13
Joeyg/apple pay fix

joeyg-stripe:
## Summary
I changed the STPPaymentContextDelegate so that the didCreatePaymentResult delegate method returns a block. That block is called if the user cancels apple pay after a payment token has already been created.

## Motivation
This is in response to issue #968. 

## Testing
I checked that the block returned by the diCreatePaymentToken delegate method gets called when it should and I checked that the didFinish delegate method is not called if the user cancels apple pay after a payment token has been created.


Stripe/PublicHeaders/STPPaymentContext.h
```diff
@@ -372,7 +372,20 @@ didCreatePaymentResult:(STPPaymentResult *)paymentResult
...
 /**
- This is invoked by an `STPPaymentContext` when it is finished. This will be called after the payment is done and all necessary UI has been dismissed. You should inspect the returned `status` and behave appropriately. For example: if it's `STPPaymentStatusSuccess`, show the user a receipt. If it's `STPPaymentStatusError`, inform the user of the error. If it's `STPPaymentStatusUserCanceled`, do nothing.
+ This method is only called when the following happens:
+    1. The user uses apple pay
+    2. A token for stripe is created and we call the delegate method with the parameter named didCreatePaymentResult is called
+    2. The user taps Apple pay's cancel button AFTER we already called the delegate method with didCreatePaymentResult
+ When this is called the user will expect you to cancel the payment but that may be difficult if you already sent the stripe token to your server. You could try
+ to give the user a refund or just tell them that they can't cancel.
+
+ Note that if the user cancel BEFORE the delegate method with didCreatePaymentResult is called then didFinish will be called and this method will not be called.
```
csabol-stripe: I think the documentation for this should remain as general as possible, with a note about the apple pay flow. How about something like:

> Some payment flows support user cancellation after they have initially authorized a payment but before it has been processed. 

> ApplePay is one such payment flow, where the ApplePay sheet remains displayed with a "Cancel" button until the payment is completed.

> In this case, the Stripe SDK will have already called `paymentContext:didCreatePaymentResult`. This method should be used to cancel a payment that is still processing if possible, or communicate to the user that the merchant was unable to cancel.

Stripe/PublicHeaders/STPPaymentContext.h
```diff
@@ -372,7 +372,20 @@ didCreatePaymentResult:(STPPaymentResult *)paymentResult
...
+ */
+- (void)paymentContext:(STPPaymentContext *)paymentContext
```
csabol-stripe: this seems like a pretty reasonable method name to me. Let's wait for a thumbs up from @danj-stripe as well though

Stripe/STPPaymentContext.m
```diff
@@ -648,15 +653,25 @@ - (void)requestPayment {
                             if ([source isKindOfClass:[STPToken class]]) {
                                 paymentResultSource = ((STPToken *)source).card;
                             }
-                            STPPaymentResult *result = [[STPPaymentResult alloc] initWithSource:paymentResultSource];
+                            result = [[STPPaymentResult alloc] initWithSource:paymentResultSource];
                             [self.delegate paymentContext:self didCreatePaymentResult:result completion:^(NSError * error) {
                                 // for Apple Pay, the didFinishWithStatus callback is fired later when Apple Pay VC finishes
+                                // The apple pay vc finishes when the user cancels and it finishes only once.
+                                // So if the user has already canceled calling completion will not cause the apple pay vc to finish
+                                // Therefore, we must directly call our did finish with status method
```
csabol-stripe: hmm, trying to understand this flow. Could you point out where `completion` is created?

Stripe/PublicHeaders/STPPaymentContext.h
joeyg-stripe: I changed the documentation but can you explain why we want this to be general. I understand that we made the api as general as possible so that we can avoid changing it in the future. But it seems to me that there is no cost to changing the documentation. So why make the documentation general?

Stripe/PublicHeaders/STPPaymentContext.h
```diff
@@ -372,7 +372,20 @@ didCreatePaymentResult:(STPPaymentResult *)paymentResult
             completion:(STPErrorBlock)completion;
 
 /**
- This is invoked by an `STPPaymentContext` when it is finished. This will be called after the payment is done and all necessary UI has been dismissed. You should inspect the returned `status` and behave appropriately. For example: if it's `STPPaymentStatusSuccess`, show the user a receipt. If it's `STPPaymentStatusError`, inform the user of the error. If it's `STPPaymentStatusUserCanceled`, do nothing.
+ Unfortunately, some payment flows support user cancellation after they have initially authorized a payment but before it has been processed.
+ ApplePay is one such payment flow, where the ApplePay sheet remains displayed with a "Cancel" button until the payment is completed.
```
jil-stripe: Nit: "Apple Pay"

Stripe/PublicHeaders/STPPaymentContext.h
```diff
@@ -372,7 +372,20 @@ didCreatePaymentResult:(STPPaymentResult *)paymentResult
...
+ As a result, the following sequence of events may occur:
+    1. The user authorizes the payment
+    2. the payment context object calls your degelate's paymentContext:didCreatePaymentResult
+    3. The user taps the cancel button before the payment is completed
+ In order to handle this case, the payment context object calls this delegate method whenever the user taps ApplePay's cancel button after paymentContext:didCreatePaymentResult. Note that paymentContext:didFinishWithStatus will not be called in this case.
```
jil-stripe: Same

jil-stripe: I would also advocate for some line length consistency, even in comments. Having one-off long lines can get kind of confusing.

Stripe/STPPaymentContext.m
```diff
@@ -626,39 +632,27 @@ - (void)requestPayment {
...
-                                if (error) {
-                                    completion(error);
-                                } else {
-                                    completion(nil);
-                                }
-                            }];
+                [self handleApplePaySource:source
+                                completion:^(NSError * error) {
+                    // for Apple Pay, the didFinishWithStatus callback is fired later when Apple Pay VC finishes
+                    // The apple pay vc finishes when the user cancels and it finishes only once.
```
jil-stripe: Nit: "Apple Pay." Also might want to spell out View Controller.

Stripe/STPPaymentContext.m
```diff
@@ -673,10 +667,17 @@ - (void)requestPayment {
                              onFinish:^(STPPaymentStatus status, NSError * _Nullable error) {
                                  [self.hostViewController dismissViewControllerAnimated:[self transitionAnimationsEnabled]
                                                                              completion:^{
-                                     [self didFinishWithStatus:status
-                                                         error:error];
+                                                                                 if (status == STPPaymentStatusUserCancellation &&
+                                                                                     self.paymentResult != nil) {
```
danj-stripe: I think you've got one or more race conditions in here. Here's my understanding of the happy path:

1. The user interacts with the `PKPaymentAuthorizationController`, authorizing the payment
2. stripe-ios creates a Source
3. stripe-ios attaches the Source to a Customer
4. The merchant's app receives `paymentContext:didCreatePaymentResult:completion:` and sends that payment result to their backend.
5. Merchant app calls `completion()` and then stripe-ios does some cleanup before calling `paymentContext:didFinishWithStatus:error:`

As far as I can tell, this PR handles the case where the user hits cancel during the processing of step 4.

However, I'm worried about what happens if the user hits cancel **earlier**. As far as I can tell, there's *nothing* that interrupts the subsequent steps. They're all chained together, and things are going to go off the rails.

As a result, the `STPPaymentContextDelegate` is going to get a very confusing set of callbacks. It'll be told that the context has finished with status canceled, and *then* it'll be told about the payment result, and presumably go send it to their backend.

Stripe/STPPaymentContext.m
```diff
@@ -626,39 +632,27 @@ - (void)requestPayment {
...
+                [self handleApplePaySource:source
+                                completion:^(NSError * error) {
+                    // for Apple Pay, the didFinishWithStatus callback is fired later when Apple Pay VC finishes
+                    // The apple pay vc finishes when the user cancels and it finishes only once.
+                    // So if the user has already canceled, calling completion will not cause the apple pay vc to finish
+                    // Therefore, we must directly call our did finish with status method
+                    if (error) {
+                        completion(error);
+                        if (isCanceledAfterPaymentResultCreated) {
+                            [self didFinishWithStatus:STPPaymentStatusError error:error];
```
danj-stripe: Your documentation on the delegate said the delegate method `paymentContext:didFinishWithStatus:error:` **will not** be called, but the code does the opposite.

I suspect we *do* want to keep the call to `paymentContext:didFinishWithStatus:error:` at the end, but I'm not sure.

Stripe/STPPaymentContext.m
```diff
@@ -626,39 +632,27 @@ - (void)requestPayment {
...
                         }
-                    });
+                    } else {
+                        completion(nil);
+                        // if isCanceledAfterPaymentResultCreated is true then we should call self.didFinishWithStatus immediately
+                        // because the user already canceled. Otherwise, self.didFinishWithStatus will be called later
+                        if (isCanceledAfterPaymentResultCreated) {
+                            [self didFinishWithStatus:STPPaymentStatusSuccess error:nil];
```
danj-stripe: How do you feel about DRYing up this method? I think it'll allow you to re-write and re-position the comments so it's easier to understand what's going on:

```objective-c
completion(error); // error might be nil

// ...
if (isCanceled...) {
  // ...
  [self didFinishWithStatus:(error != nil ? ...Error : ...Success) error:error];
}

File: Stripe/STPPaymentContext.m [hunk 10]
```diff
@@ -687,9 +688,45 @@ - (void)requestPayment {
     }];
 }
 
+/**
+ @param completion The completion block that gets passed to the paymentContext:didCreatePaymentResult
+ method on self.delegate.
```
danj-stripe: I think this method deserves better documentation, and maybe a better name.

What does it do to handle the source? (attaches the source to the customer, creates a payment result, passes it to the delegate)

When does the `completion` block get called? (if there's an error attaching the source, or once the `STPPaymentContextDelegate` has finished handling the `STPPaymentResult`, probably by telling their backed to charge it)

Stripe/STPPaymentContext.m
danj-stripe: Also, the `paymentResult` will stick around on the `STPPaymentContext`, and if they re-use the context to try again it's going to throw off your new logic.

Issue #999, opened 2018-07-17
Crash when adding a card


odedharth:
## Summary
There is a fatal exception that is crashing the app for some of our users when adding a credit card.

The error is:
```
Fatal Exception: NSInternalInconsistencyException
attempt to insert row 0 into section 1, but there are only 1 sections after the update
-[STPAddCardViewController addressViewModel:addedCellAtIndex:]
```
Happening at:
```
STPAddCardViewController.m line 362-[STPAddCardViewController addressViewModel:addedCellAtIndex:]
```

## Code to reproduce
It's happening for some of our users in production, and we don't know how to reproduce but we do have the stack trace of the crashes using Crashalytics: https://fabric.io/oded-harths-projects/ios/apps/oded-harth.mdacne/issues/5b1482136007d59fcdd040b0?time=last-ninety-days

## iOS version
iOS 11

## Installation method
Cocoapods

## SDK version
11.5.0

odedharth:
@cale-stripe Sorry send the wrong link: http://crashes.to/s/90e272acb88

odedharth:
@csabol-stripe Any ETA for the fix?

csabol-stripe:
Fix will go into our next SDK release which is our priority at the moment! :)

csabol-stripe:
We've just released v13.2.0 that fixes this crash. Thanks for the report and let us know if you run into any other issues!

PR #1004, opened 2018-08-03
Fixes crash in STPAddCardViewController with some prefilled billing addresses

csabol-stripe:
## Summary
Fixes STPAddCardViewController STPAddressViewModelDelegate methods to check for valid table view inserts and removes.

In cases where a zip code is required and the user's locale doesn't support zip codes but their billing address includes one (or vice-versa),
this was causing a crash because we were trying to add or remove cells that were already there or were never there respectively.

## Motivation
https://github.com/stripe/stripe-ios/issues/999

## Testing

I wrote automated tests for the cases that were previously causing crashes

### DIFF
--- a/Stripe/STPAddCardViewController.m
+++ b/Stripe/STPAddCardViewController.m
@@ -402,8 +402,8 @@ typedef NS_ENUM(NSUInteger, STPPaymentCardSection) {
     if (rowsInSection != NSNotFound && rowsInSection < [self tableView:self.tableView numberOfRowsInSection:STPPaymentCardBillingAddressSection]) {
         NSIndexPath *indexPath = [NSIndexPath indexPathForRow:index inSection:STPPaymentCardBillingAddressSection];
         [self.tableView insertRowsAtIndexPaths:@[indexPath] withRowAnimation:UITableViewRowAnimationAutomatic];
-        [self updateInputAccessoryVisiblity];
     }
+    [self updateInputAccessoryVisiblity];
 }
 
 - (void)addressViewModel:(__unused STPAddressViewModel *)addressViewModel removedCellAtIndex:(NSUInteger)index {
--- a/Stripe/STPAddCardViewController.m
+++ b/Stripe/STPAddCardViewController.m
@@ -115,10 +115,6 @@ typedef NS_ENUM(NSUInteger, STPPaymentCardSection) {
     STPPaymentCardTextFieldCell *paymentCell = [[STPPaymentCardTextFieldCell alloc] init];
     paymentCell.paymentField.delegate = self;
     self.paymentCell = paymentCell;
-
-    if (self.prefilledInformation.billingAddress != nil) {
-        self.addressViewModel.address = self.prefilledInformation.billingAddress;
-    }
     
     self.activityIndicator = [[STPPaymentActivityIndicatorView alloc] initWithFrame:CGRectMake(0, 0, 20.0f, 20.0f)];
     
@@ -127,6 +123,10 @@ typedef NS_ENUM(NSUInteger, STPPaymentCardSection) {
     [self updateInputAccessoryVisiblity];
     self.tableView.dataSource = self;
     self.tableView.delegate = self;
+
+    if (self.prefilledInformation.billingAddress != nil) {
+        self.addressViewModel.address = self.prefilledInformation.billingAddress;
+    }
 
     STPSectionHeaderView *addressHeaderView = [STPSectionHeaderView new];
     addressHeaderView.theme = self.theme;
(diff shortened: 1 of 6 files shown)


Tests/Tests/NSLocale+STPSwizzling.m
```diff
@@ -0,0 +1,59 @@
...
+
+@interface NSObject (STPSwizzling)
+
++ (void)stp_swizzleClassMethod:(SEL)original withReplacement:(SEL)replacement;
+
+@end
+
+@implementation NSObject (STPSwizzling)
+
++ (void)stp_swizzleClassMethod:(SEL)original withReplacement:(SEL)replacement
```
danj-stripe: Did you consider using the `STPAspects` code instead?

This is a really simple usage, and since it's test code I don't think it needs to get fancy, mostly just curious.

Stripe/STPAddCardViewController.m
```diff
@@ -398,14 +398,20 @@ - (void)paymentCardTextFieldDidEndEditingCVC:(__unused STPPaymentCardTextField *
 #pragma mark - STPAddressViewModelDelegate
 
 - (void)addressViewModel:(__unused STPAddressViewModel *)addressViewModel addedCellAtIndex:(NSUInteger)index {
-    NSIndexPath *indexPath = [NSIndexPath indexPathForRow:index inSection:STPPaymentCardBillingAddressSection];
-    [self.tableView insertRowsAtIndexPaths:@[indexPath] withRowAnimation:UITableViewRowAnimationAutomatic];
-    [self updateInputAccessoryVisiblity];
+    NSInteger rowsInSection = [self.tableView numberOfRowsInSection:STPPaymentCardBillingAddressSection];
+    if (rowsInSection != NSNotFound && rowsInSection < [self tableView:self.tableView numberOfRowsInSection:STPPaymentCardBillingAddressSection]) {
```
danj-stripe: Is there a way to (cleanly) push this down into the `STPAddressViewModel`? It looks like it's already trying to track `showingPostalCodeCell`, and has logic for whether it's being added or removed based on updates to the address.

Or maybe we should move the `_addressViewModel.delegate = self` line to *after* we set the pre-filled info, because we don't care about updates to the ViewModel until after we've finished creating the views? That might allow us to avoid this complexity completely?

Stripe/STPAddCardViewController.m
```diff
@@ -398,14 +398,20 @@ - (void)paymentCardTextFieldDidEndEditingCVC:(__unused STPPaymentCardTextField *
...
 
 - (void)addressViewModel:(__unused STPAddressViewModel *)addressViewModel removedCellAtIndex:(NSUInteger)index {
-    NSIndexPath *indexPath = [NSIndexPath indexPathForRow:index inSection:STPPaymentCardBillingAddressSection];
-    [self.tableView deleteRowsAtIndexPaths:@[indexPath] withRowAnimation:UITableViewRowAnimationAutomatic];
+    NSInteger rowsInSection = [self.tableView numberOfRowsInSection:STPPaymentCardBillingAddressSection];
+    if (rowsInSection != NSNotFound && index < (NSUInteger)rowsInSection) {
+        NSIndexPath *indexPath = [NSIndexPath indexPathForRow:index inSection:STPPaymentCardBillingAddressSection];
+        [self.tableView deleteRowsAtIndexPaths:@[indexPath] withRowAnimation:UITableViewRowAnimationAutomatic];
+    }
     [self updateInputAccessoryVisiblity];
```
danj-stripe: Is it important that the call to `updateInputAccessoryVisibility` is inside the `if` block for add, but outside it for remove?

I don't think so, but wanted to see if there was something I'm missing.

Tests/Tests/NSLocale+STPSwizzling.m
csabol-stripe: didn't even know that existed! I'll keep in mind for next time

Stripe/STPAddCardViewController.m
csabol-stripe: > Is there a way to (cleanly) push this down into the STPAddressViewModel?

I actually prefer the logic to live in the view controller because this situation is a result of layout passes/table view caching/etc. that I think the view controller is closer too than the view model (correct me if I'm wrong in my view model understanding :) ). 

> Or maybe we should move the _addressViewModel.delegate = self line to after we set the pre-filled info, because we don't care about updates to the ViewModel until after we've finished creating the views? That might allow us to avoid this complexity completely?

I thought about this. I believe that change would also fix the crash, but this solution is more robust

Stripe/STPAddCardViewController.m
csabol-stripe: no, I'll move it out of the if in remove for consistency :)

csabol-stripe:
>  I'm curious if there's something better we can do to handle the add/removed cell calls

I played around with the idea of a larger fix that avoids delegate calls that are too early, etc. I ended up deciding that this solution is not too hacky, pretty robust, and ultimately the best bang for our buck. Happy to reconsider/discuss if you feel differently!

csabol-stripe:
@danj-stripe pointed out that STPShippingAddressViewController has pretty much identical code. I'm going to try triggering the crash there and fix in this patch

Tests/Tests/STPShippingAddressViewControllerTest.m
```diff
@@ -0,0 +1,73 @@
...
+
+    STPShippingAddressViewController *sut = [[STPShippingAddressViewController alloc] initWithConfiguration:config
+                                                                                                      theme:[STPTheme defaultTheme]
+                                                                                                   currency:nil
+                                                                                            shippingAddress:address
+                                                                                     selectedShippingMethod:nil
+                                                                                       prefilledInformation:nil];
+
+    [sut loadView];
+    [sut viewDidLoad];
```
danj-stripe: WDYT about adding `XCTAssertNoThrow` calls around these?

I think that'd catch the `NSInternalInconsistencyException` that was being triggered. If so, it'd give a better failure message if it regresses, and it helps document what these tests are looking for. As it is, without  context this test looks like a no-op.

Same for the matching tests of the other class.

Tests/Tests/STPShippingAddressViewControllerTest.m
```diff
@@ -0,0 +1,73 @@
...
+}
+
+- (void)testPrefilledBillingAddress_addAddress {
+    [NSLocale stp_setCurrentLocale:[NSLocale localeWithLocaleIdentifier:@"en_ZW"]];
```
danj-stripe: WDYT about adding assertions (or a separate test?) that `postalCodeIsRequiredForCountryCode:` returns the expected values?

If ZW starts requiring a postal code, or we stop requiring it for US, we start testing a different code path, and the code could break silently.

It also helps document what's being tested, and why those countries were chosen.

Tests/Tests/STPAddCardViewControllerTest.m
```diff
@@ -41,6 +42,53 @@ - (STPAddCardViewController *)buildAddCardViewController {
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Warc-performSelector-leaks"
 
+- (void)testPrefilledBillingAddress_removeAddress {
```
danj-stripe: Do these need to be inside the `#pragma clang diagnostic ignored "-Warc-performSelector-leaks"` block? It might be slightly better to move them outside.

PR #1011, opened 2018-08-08
Adds 'CVV', 'CVC', and 'MM/YY' to localizable strings

csabol-stripe:
## Motivation
https://jira.corp.stripe.com/browse/IOS-229

## Testing
Verified strings still show up correctly in english (sending strings to our localization team out of band)


Stripe/STPPaymentCardTextField.m
```diff
@@ -161,7 +161,7 @@ - (void)commonInit {
     expirationField.alpha = 0;
     expirationField.accessibilityLabel = STPLocalizedString(@"expiration date", @"accessibility label for text field");
     self.expirationField = expirationField;
-    self.expirationPlaceholder = @"MM/YY";
+    self.expirationPlaceholder = STPLocalizedString(@"MM/YY", "label for text field to enter card expiry");
```
danj-stripe: Looks like CI failed because this is missing `@` at the beginning of the second string here.

PR #1017, opened 2018-08-15
Add document_back property to STPVerificationParams

rance-attack:
## Summary
Add document_back property 

## Motivation
Can see this property in go project (https://github.com/stripe/stripe-go/blob/master/account.go#L211)
Hope iOS SDK can also have it

### DIFF
--- a/Stripe/PublicHeaders/STPLegalEntityParams.h
+++ b/Stripe/PublicHeaders/STPLegalEntityParams.h
@@ -135,6 +135,10 @@ NS_ASSUME_NONNULL_BEGIN
  The file id for the uploaded verification document.
  */
 @property (nonatomic, copy, nullable) NSString *document;
+
+/**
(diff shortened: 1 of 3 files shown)


danj-stripe:
@rance-attack Yeah, we appreciate the contribution, and your follow up commit w/documentation that fixed the build!

FYI: we do have `additionalAPIParameters` on all of the params objects, so you can immediately start using properties (like `document_back`) that the Stripe API supports but that the SDK doesn't yet have: https://github.com/stripe/stripe-ios/blob/7116259c7b57b195d7d7f28c1521f8e65f27e99c/Stripe/PublicHeaders/STPFormEncodable.h#L28-L35

It won't have the type-safety that the property declaration does, so we definitely want to fix the SDK like you've done here, but it would allow you to use the SDK without forking it.

PR #1027, opened 2018-10-09
Set `internalCardParams` after setting fields to avoid params overriding

jeanelie:
## Summary
In `STPPaymentCardTextField`, set `internalCardParams` *after* setting text fields in `setCardParams:(STPCardParams *)cardParams`

## Motivation
With the current implementation, setting `cardParams` triggers `paymentCardTextFieldDidChange`. 
If you just access `textField.cardParams` in this delegate, it will modify params and thus only fill the first field `STPCardFieldTypeNumber`, not the other ones. 

## Testing
Setting `paymentCardTextField.cardParams = cardParams` with custom params. 
Accessing `textField.cardParams` in `paymentCardTextFieldDidChange` by printing them for instance.
Accessing will modify the data, and not fill all the Fields.

### DIFF
--- a/Stripe/STPPaymentCardTextField.m
+++ b/Stripe/STPPaymentCardTextField.m
@@ -638,7 +638,6 @@ CGFloat const STPPaymentCardTextFieldMinimumPadding = 10;
      */
     STPFormTextField *originalSubResponder = self.currentFirstResponderField;
 
-    self.internalCardParams = cardParams;
     [self setText:cardParams.number inField:STPCardFieldTypeNumber];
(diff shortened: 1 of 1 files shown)


danj-stripe:
Hi @jeanelie,

Thanks **so** much for tracking this down, and opening a PR. It's an interesting interaction that I imagine was frustrating for you to encounter.

I wrote a test that reproduces the problem, and then we talked it over internally. We'd rather take a different approach to fixing it, which I've done in #1031. So I'm going to close this PR in favor of that one, but I really do appreciate the work you put into this one.

You can watch #1031 for updates, or keep an eye out for our next release which should fix this issue.

Thanks,
Dan

PR #1031, opened 2018-10-23
Change `STPPaymentCardTextField` to copy the `cardParams` object when it's set or read.

danj-stripe:
## Summary
Exposing the internal `STPCardParams` object led to weird interactions, like #1027. It
also doesn't match our desired information hiding. Setting some fields on the `cardParams`
would persist (name, address, etc), while setting others had no effect (card number, etc).

By copying in both setter & getter, the internal state of this view is protected, and
must be changed by setting a new object on the property.

We expect well-behaved code to be backwards compatible, but this is a breaking change
that'll require a major version bump because it changes the interface & behavior.

## Motivation
#1027 and IOS-958

## Testing
Tests were written that illustrate the problem & current behavior, and then updated as the
behavior changed.

Sample projects were inspected and tested to make sure they weren't using the payment text
field in an unsupported way.

### DIFF
--- a/Stripe/STPPaymentCardTextField.m
+++ b/Stripe/STPPaymentCardTextField.m
@@ -618,7 +618,7 @@ CGFloat const STPPaymentCardTextFieldMinimumPadding = 10;
         // if we are not showing the postal code entry field.
         self.internalCardParams.address.postalCode = self.postalCode;
     }
-    return self.internalCardParams;
+    return [self.internalCardParams copy];
 }
 
 - (void)setCardParams:(STPCardParams *)cardParams {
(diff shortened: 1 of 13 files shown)


Stripe/STPPaymentCardTextField.m
```diff
@@ -638,7 +638,7 @@ behaves identically to user entering text (and will have the same forwarding
      */
     STPFormTextField *originalSubResponder = self.currentFirstResponderField;
 
-    self.internalCardParams = cardParams;
+    self.internalCardParams = [cardParams copy];
     [self setText:cardParams.number inField:STPCardFieldTypeNumber];
```
csabol-stripe: might be worth adding a comment here highlighting that it's important to use `cardParams` and not `internalCardParams` for the `sets`

danj-stripe:
re-review? @csabol-stripe

danj-stripe:
I'm investigating the failure of `-[STPCertTest testExpired]`

PR #1037, opened 2018-11-08
Change Styleguide around acronym capitalization, and fix recent PaymentIntent code

danj-stripe:
## Summary
Fixes the styleguide to match the codebase, and updates recent code that was written
to follow the (incorrect) styleguide re: URL vs Url.

Only one public-facing property needed changing. Deprecated it, along with a fixit to
make the upgrade path easier for users. Everything else was private or test-only.

## Motivation
I'd been following the STYLEGUIDE.md while working on the last PaymentIntent-related PRs.
While working on another PaymentIntent-related PR, I noticed the inconsistency, and
took the time to fix all of the `Url` cases.

## Testing
I specifically checked URL vs `Url`, and it was overwhelmingly `URL`.
`stripeID` beats `stripeId` by about 2 to 1, but I'm not going to change them.
You can see fully capitalized acronyms in class names too.

The automated tests verify no functionality change, and the fixit on the only
public-facing property worked to update the sample code.

### DIFF
--- a/STYLEGUIDE.md
+++ b/STYLEGUIDE.md
@@ -18,7 +18,7 @@
 
 - Avoid single letter variables. Try using `idx` / `jdx` instead of `i` / `j` in for loops.
 
-- Prefer `urlString` over `URLString` (acronym prefix), `baseUrlString` over `baseURLString` (acronym infix), and `stripeId` over `stripeID` (acronym suffix)
+- Acronyms should be all lowercase as a method prefix (ex:`url` or `urlString`). Otherwise, they should be all caps when occurring elsewhere in the method name, or as a class name (ex: `handleStripeURLCallbackWithURL`, `stripeID` or `STPAPIClient`)
 
 ### Control Flow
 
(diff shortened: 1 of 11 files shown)


Stripe/STPRedirectContext.m
```diff
@@ -69,10 +69,10 @@ - (nullable instancetype)initWithPaymentIntent:(STPPaymentIntent *)paymentIntent
...
     }
 
-    NSString *redirectUrl = nextSourceAction[@"value"][@"url"];
-    return [self initWithNativeRedirectUrl:nil
-                               redirectUrl:[NSURL URLWithString:redirectUrl]
-                                 returnUrl:paymentIntent.returnUrl
+    NSString *redirectURL = nextSourceAction[@"value"][@"url"];
+    return [self initWithNativeRedirectURL:nil
+                               redirectURL:[NSURL URLWithString:redirectURL]
+                                 returnURL:paymentIntent.returnUrl
```
danj-stripe: After this PR lands, the only case-sensitive search results for `Url` are the `returnUrl` property on `STPPaymentIntent`.

That property is being removed for IOS-1013, so I left it unchanged in this PR.

STYLEGUIDE.md
```diff
@@ -18,7 +18,7 @@
 
 - Avoid single letter variables. Try using `idx` / `jdx` instead of `i` / `j` in for loops.
 
-- Prefer `urlString` over `URLString` (acronym prefix), `baseUrlString` over `baseURLString` (acronym infix), and `stripeId` over `stripeID` (acronym suffix)
```
danj-stripe: I'm not sure where this came from. It doesn't match the code. I took a shot at re-writing it to what I believe is the dominant style of the code base, but feedback on what the replacement should say would be great.

PR #1038, opened 2018-11-08
Update PaymentIntents support

danj-stripe:
## Summary
* return_url removed from top-level PaymentIntent object
* parsing & exposing `nextSourceAction`, with the only possible type being `AuthorizeWithURL`
* using `returnURL` from the `nextSourceAction` instead of the top-level `returnURL` in
`STPRedirectContext`

## Motivation
IOS-1013

## Testing
Updated automated tests

### DIFF
--- a/Stripe/PublicHeaders/STPPaymentIntentSourceActionAuthorizeWithURL.h
+++ b/Stripe/PublicHeaders/STPPaymentIntentSourceActionAuthorizeWithURL.h
@@ -13,7 +13,7 @@
 NS_ASSUME_NONNULL_BEGIN
 
 /**
- The `STPPaymentIntentSourceAction` details when type is `authorize_with_url`.
+ The `STPPaymentIntentSourceAction` details when type is `STPPaymentIntentSourceActionTypeAuthorizeWithURL`.
 
  These are created & owned by the containing `STPPaymentIntent`.
  */
--- a/Stripe/PublicHeaders/STPPaymentIntentSourceActionAuthorizeWithURL.h
+++ b/Stripe/PublicHeaders/STPPaymentIntentSourceActionAuthorizeWithURL.h
@@ -1,0 +1,14 @@
+//
+//  STPPaymentIntentSourceActionAuthorizeWithURL.h
+//  Stripe
+//
+//  Created by Daniel Jackson on 11/7/18.
+//  Copyright © 2018 Stripe, Inc. All rights reserved.
+//
+
+#import <Foundation/Foundation.h>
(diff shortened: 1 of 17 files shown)


danj-stripe:
This PR relies on #1037, so it is not using master as the base branch and the CI tests haven't run against it. It looked good locally, and worked when I used the example app to create & confirm PaymentIntents.

I think the code's ready to review, but it'll probably require another 👍 after #1037 is merged and the base branch is changed.

Stripe/PublicHeaders/STPPaymentIntentSourceActionAuthorizeWithURL.h
```diff
@@ -0,0 +1,40 @@
...
+#import "STPAPIResponseDecodable.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ The `STPPaymentIntentSourceAction` details when type is `STPPaymentIntentSourceActionTypeAuthorizeWithURL`.
+
+ These are created & owned by the containing `STPPaymentIntent`.
+ */
+@interface STPPaymentIntentSourceActionAuthorizeWithURL: NSObject<STPAPIResponseDecodable>
```
csabol-stripe: What do you think of renaming this something like `STPPaymentIntentSourceActionURLAuthorizationData`? `AuthorizeWithURL` is a bit confusing to me

danj-stripe: I figured consistency with the API (where this shows up as `authorize_with_url`, plus the other stripe SDKs (which use essentially the same thing), was a good reason to go with this monstrosity of a name. It's my hope that users of the library never actually really interact with this class - I suspect they'll just access it's fields through the `STPPaymentIntent` object

* https://github.com/stripe/stripe-go/blob/4f11ac96bdae9a119b13ba82a5cc0e95d96dd5cc/paymentintent.go#L115-L118
* https://github.com/stripe/stripe-dotnet/blob/a9d958bf4a1ddafc8446461a47e6d04eef9301a1/src/Stripe.net/Entities/PaymentIntents/PaymentIntentSourceActionAuthorizeWithUrl.cs#L5
* https://github.com/stripe/stripe-java/blob/a8ff01ec123f8d0770a6248cc20a61fdae6c3b1a/src/main/java/com/stripe/model/PaymentIntentSourceActionValueAuthorizeWithUrl.java#L10

csabol-stripe: Makes sense

PR #1042, opened 2018-11-16
Don't percent encode entire form bodies

jack-stripe:
## Summary
On GET and POST request, we currently percent-escape the entire query - for example, trying to GET a URL like /v1/charges with a hash of parameters like `{foo: "bar]"}` will end up looking like `GET /v1/charges?foo=bar%5D` (which, to be clear, is what you want).

We are overzealously applying this logic, however, to array and dictionary-like structures. The same request above, with a param hash like `{foo: {bar: "baz"}}` will be serialized to `GET /v1/charges?foo%5Bbar%5D=baz`, when really what you want is `GET /v1/charges?foo[bar]=baz`. The thing you don't want is _user-provided_ characters like `]` not getting escaped. While our API is generally robust against encoding like this, certain components don't like it.

To do this, I've refactored our form encoder quite a bit. I've tried to make it a bit clearer how it works, and document some of my changes along the way. Basically, where the path used to be something like (NSDictionary of parameters) -> (form-encoded path) -> (percent-escaping) it is now (NSDictionary of parameters) -> (NSDictionary of percent-escaped parameters) -> (form-encoded path).

## Motivation
I'm working on an upcoming release that is finickier-than-usual about how requests are encoded.

## Testing
Added some automated tests.

### DIFF
--- a/Tests/Tests/STPFormEncoderTest.m
+++ b/Tests/Tests/STPFormEncoderTest.m
@@ -61,8 +61,7 @@
 // helper test method
 - (NSString *)encodeObject:(STPTestFormEncodableObject *)object {
     NSDictionary *dictionary = [STPFormEncoder dictionaryForObject:object];
-    NSString *queryString = [STPFormEncoder queryStringFromParameters:dictionary];
-    return [queryString stringByRemovingPercentEncoding];
+    return [STPFormEncoder queryStringFromParameters:dictionary];
 }
 
 - (void)testFormEncoding_emptyObject {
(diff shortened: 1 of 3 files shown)


Tests/Tests/STPFormEncoderTest.m
```diff
@@ -184,6 +183,15 @@ - (void)testFormEncoding_nilRootObject {
     XCTAssertEqualObjects([self encodeObject:testObject], @"test_property=success");
 }
 
+- (void)testQueryStringWithBadFields {
+    NSDictionary *params = @{
+                             @"foo]": @"bar",
+                             @"baz": @"qux[",
```
danj-stripe: Any chance you could add a `;` to this bad field test (it should encode to `%3B`)? We tripped over that in dashboard-ios#862

PR #1045, opened 2018-11-19
Push provisioning

jack-stripe:


Stripe/STPEphemeralKey.m
```diff
@@ -80,7 +86,8 @@ - (BOOL)isEqualToResourceKey:(STPEphemeralKey *)other {
     return [self.stripeID isEqualToString:other.stripeID] 
         && [self.secret isEqualToString:other.secret]
         && [self.expires isEqual:other.expires]
-        && [self.customerID isEqual:other.customerID];
+        && [self.customerID isEqual:other.customerID]
```
danj-stripe: I think `[nil isEqual: nil]` returns false, and so this breaks when you make `customerID` `nullable`

Stripe/PublicHeaders/STPEphemeralKeyProvider.h
```diff
@@ -40,7 +34,39 @@ NS_ASSUME_NONNULL_BEGIN
...
+@end
+
+/**
+ You should make your application's API client conform to this interface.
+ It provides a way for Stripe utility classes to request a new ephemeral key from
+ your backend, which it will use to retrieve and update Stripe API objects.
+ @deprecated use `STPCustomerEphemeralKeyProvider` or `STPIssuingCardEphemeralKeyProvider`
+ instead depending on the type of key that will be fetched.
+ */
+__attribute__ ((deprecated))
```
danj-stripe: idk if it works in this particular case, but it'd be great to use the deprecated msg & a replacement name so that Xcode can provide a fixit:

https://github.com/stripe/stripe-ios/blob/b787351d6139d4114cfd68d02583420683b21c5f/Stripe/PublicHeaders/STPRedirectContext.h#L55-L57

Stripe/STPPushProvisioningContext.m
```diff
@@ -0,0 +1,95 @@
...
+            if (![[PKPassLibrary new] canAddPaymentPassWithPrimaryAccountIdentifier:card.cardId]) {
+                NSError *error = nil; // TODO
+                completion(nil, error);
+            }
+            
+            PKAddPaymentPassRequestConfiguration *configuration = [[PKAddPaymentPassRequestConfiguration alloc] initWithEncryptionScheme:PKEncryptionSchemeECC_V2];
+            configuration.cardholderName = card.name;
+            configuration.primaryAccountSuffix = card.last4;
+            if (@available(iOS 12.0, *)) {
+                configuration.style = PKAddPaymentPassStylePayment;
```
danj-stripe: Offhand I don't remember the incantation necessary, but we're still supporting Xcode 9 and it doesn't know about this symbol.

I don't actually know what that'll mean for that static binary that we provide via github. Instructions are to generate it with the oldest supported major version of Xcode, which means this would only be set for users who integrate via one of the other methods...

Stripe/STPPushProvisioningContext.m
```diff
@@ -0,0 +1,95 @@
...
+            }
+            configuration.paymentNetwork = [STPCard stringFromBrand:card.brand];
```
danj-stripe: I don't know that I trust that our strings are the same as what Apple expects

PR #1046, opened 2018-11-20
Add custom asset to Add Card VC for entering Amex CVC 

danj-stripe:
## Summary
The payment text field has different assets for CVC, based on whether the card is an amex or not. Adding the same distinction to the card icon on `STPAddCardViewController`

simulator screen shot - iphone 5s - 2018-11-19 at 17 08 22
simulator screen shot - iphone 5s - 2018-11-19 at 17 08 14

amex-flip

## Motivation
IOS-83

## Testing
Manually tested in the example app.


danj-stripe:
This PR is waiting on new assets.

PR #1050, opened 2018-11-28
Adds new localizations and updated translations

csabol-stripe:
## Summary
All translations were audited and updated by new localization team.
Added new localizations for:
- Danish
- Spanish (Argentina/Latin America)
- French (Canada)
- Norwegian
- Portuguese (Brazil)
- Portuguese (Portugal)
- Swedish

## Motivation
Improve localization

## Testing
Working on LocalizationQA with Localization team


csabol-stripe:
I'm not adding additional snapshots for the new translations because, after meeting with Localization QA, they have an incoming patch by @aa-stripe that uses XCUITest and is considered preferable. Once that's integrated we can probalby get rid of these snapshots

PR #1131, opened 2019-02-07
New type of ephemeral keys

acavailhez-stripe:
## Summary
Allowing for ephemeral keys that are not customer

## Motivation
Stripe Issuing will need those

## Testing
Existing ephemeral keys test apply

### DIFF
--- a/Stripe/PublicHeaders/STPCustomerContext.h
+++ b/Stripe/PublicHeaders/STPCustomerContext.h
@@ -12,7 +12,7 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-@protocol STPEphemeralKeyProvider;
+@protocol STPCustomerEphemeralKeyProvider;
 @class STPEphemeralKey, STPEphemeralKeyManager;
 
 /**
@@ -35,7 +35,7 @@ NS_ASSUME_NONNULL_BEGIN
  @param keyProvider   The key provider the customer context will use.
  @return the newly-instantiated customer context.
  */
-- (instancetype)initWithKeyProvider:(id<STPEphemeralKeyProvider>)keyProvider;
+- (instancetype)initWithKeyProvider:(id<STPCustomerEphemeralKeyProvider>)keyProvider;
 
 /**
  `STPCustomerContext` will cache its customer object for up to 60 seconds.
--- a/Stripe/PublicHeaders/STPEphemeralKeyProvider.h
+++ b/Stripe/PublicHeaders/STPEphemeralKeyProvider.h
@@ -14,6 +14,11 @@ NS_ASSUME_NONNULL_BEGIN
 
 @class STPEphemeralKey;
 
+/**
+ You should make your application's API client conform to this interface.
+ It provides a way for Stripe utility classes to request a new ephemeral key from
+ your backend, which it will use to retrieve and update Stripe API objects.
+ */
 @protocol STPCustomerEphemeralKeyProvider <NSObject>
 /**
  Creates a new ephemeral key for retrieving and updating a Stripe customer.
@@ -36,6 +41,11 @@ NS_ASSUME_NONNULL_BEGIN
 - (void)createCustomerKeyWithAPIVersion:(NSString *)apiVersion completion:(STPJSONResponseCompletionBlock)completion;
 @end
 
+/**
+ You should make your application's API client conform to this interface.
+ It provides a way for Stripe utility classes to request a new ephemeral key from
+ your backend, which it will use to retrieve and update Stripe API objects.
+ */
 @protocol STPIssuingCardEphemeralKeyProvider <NSObject>
 /**
  Creates a new ephemeral key for retrieving and updating a Stripe Issuing Card.
--- a/Stripe/PublicHeaders/STPEphemeralKeyProvider.h
+++ b/Stripe/PublicHeaders/STPEphemeralKeyProvider.h
@@ -14,23 +14,17 @@ NS_ASSUME_NONNULL_BEGIN
 
 @class STPEphemeralKey;
 
-/**
- You should make your application's API client conform to this interface.
- It provides a way for `STPCustomerContext` to request a new ephemeral key from
- your backend, which it will use to retrieve and update a Stripe customer.
- */
-@protocol STPEphemeralKeyProvider <NSObject>
-
+@protocol STPCustomerEphemeralKeyProvider <NSObject>
 /**
  Creates a new ephemeral key for retrieving and updating a Stripe customer.
  On your backend, you should create a new ephemeral key for the Stripe customer
  associated with your user, and return the raw JSON response from the Stripe API.
  For an example Ruby implementation of this API, refer to our example backend:
  https://github.com/stripe/example-ios-backend/blob/v14.0.0/web.rb
-
+ 
  Back in your iOS app, once you have a response from this API, call the provided
  completion block with the JSON response, or an error if one occurred.
-
+ 
  @param apiVersion  The Stripe API version to use when creating a key.
  You should pass this parameter to your backend, and use it to set the API version
  in your key creation request. Passing this version parameter ensures that the
@@ -40,7 +34,39 @@ NS_ASSUME_NONNULL_BEGIN
  or `completion(nil, error)` if an error is returned.
  */
 - (void)createCustomerKeyWithAPIVersion:(NSString *)apiVersion completion:(STPJSONResponseCompletionBlock)completion;
+@end
 
+@protocol STPIssuingCardEphemeralKeyProvider <NSObject>
+/**
+ Creates a new ephemeral key for retrieving and updating a Stripe Issuing Card.
+ On your backend, you should create a new ephemeral key for your logged-in user's
+ primary Issuing Card, and return the raw JSON response from the Stripe API.
+ For an example Ruby implementation of this API, refer to our example backend:
+ https://github.com/stripe/example-ios-backend/blob/v14.0.0/web.rb
+ 
+ Back in your iOS app, once you have a response from this API, call the provided
+ completion block with the JSON response, or an error if one occurred.
+ 
+ @param apiVersion  The Stripe API version to use when creating a key.
+ You should pass this parameter to your backend, and use it to set the API version
+ in your key creation request. Passing this version parameter ensures that the
+ Stripe SDK can always parse the ephemeral key response from your server.
+ @param completion  Call this callback when you're done fetching a new ephemeral
+ key from your backend. For example, `completion(json, nil)` (if your call succeeds)
+ or `completion(nil, error)` if an error is returned.
+ */
+- (void)createIssuingCardKeyWithAPIVersion:(NSString *)apiVersion completion:(STPJSONResponseCompletionBlock)completion;
+@end
+
+/**
+ You should make your application's API client conform to this interface.
+ It provides a way for Stripe utility classes to request a new ephemeral key from
+ your backend, which it will use to retrieve and update Stripe API objects.
+ @deprecated use `STPCustomerEphemeralKeyProvider` or `STPIssuingCardEphemeralKeyProvider`
+depending on the type of key that will be fetched.
+ */
+__attribute__ ((deprecated("STPEphemeralKeyProvider has been renamed to STPCustomerEphemeralKeyProvider", "STPCustomerEphemeralKeyProvider")))
+@protocol STPEphemeralKeyProvider <STPCustomerEphemeralKeyProvider>
 @end
 
 NS_ASSUME_NONNULL_END
--- a/Stripe/STPEphemeralKey.m
+++ b/Stripe/STPEphemeralKey.m
@@ -17,7 +17,8 @@
 @property (nonatomic, readwrite) BOOL livemode;
 @property (nonatomic, readwrite) NSString *secret;
 @property (nonatomic, readwrite) NSDate *expires;
-@property (nonatomic, readwrite) NSString *customerID;
+@property (nonatomic, readwrite, nullable) NSString *customerID;
+@property (nonatomic, readwrite, nullable) NSString *issuingCardID;
 @property (nonatomic, readwrite, nonnull, copy) NSDictionary *allResponseFields;
 
 @end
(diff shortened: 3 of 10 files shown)


Stripe/STPEphemeralKey.m
```diff
@@ -73,14 +79,11 @@ - (BOOL)isEqual:(id)object {
     if (!object || ![object isKindOfClass:self.class]) {
         return NO;
     }
-    return [self isEqualToResourceKey:object];
+    return [self isEqualToEphemeralKey:object];
 }
 
-- (BOOL)isEqualToResourceKey:(STPEphemeralKey *)other {
-    return [self.stripeID isEqualToString:other.stripeID] 
-        && [self.secret isEqualToString:other.secret]
```
csabol-stripe: why were these removed?

Stripe/STPEphemeralKeyManager.m
```diff
@@ -64,18 +68,44 @@ - (void)handleWillForegroundNotification {
     // eager refreshses to once per hour.
     if (!self.currentKeyIsUnexpired && self.shouldPerformEagerRefresh) {
         self.lastEagerKeyRefresh = [NSDate date];
-        [self.keyProvider createCustomerKeyWithAPIVersion:self.apiVersion completion:^(NSDictionary *jsonResponse, __unused NSError *error) {
-            STPEphemeralKey *key = [STPEphemeralKey decodedObjectFromAPIResponse:jsonResponse];
-            if (key) {
-                self.customerKey = key;
-            }
+        [self getOrCreateKey:^(__unused STPEphemeralKey * _Nullable ephemeralKey, __unused NSError * _Nullable error) {
+
```
csabol-stripe: It looks like `self.ephemeralKey` gets set in the actual to `getOrCreateKey`. Could we add some comments here explaining why this is called and why we do nothing with the response here?

Stripe/STPEphemeralKeyManager.m
```diff
@@ -64,18 +68,44 @@ - (void)handleWillForegroundNotification {
...
+        STPEphemeralKey *key = [STPEphemeralKey decodedObjectFromAPIResponse:jsonResponse];
+        if (key) {
+            [self.createKeyPromise succeed:key];
+        } else {
+            // the API request failed
+            if (error) {
+                [self.createKeyPromise fail:error];
+            }
+            // the ephemeral key could not be decoded
+            else {
```
csabol-stripe: style-nit: I prefer to have the `else` on the same line as the closing brace, `} else {`

Stripe/STPEphemeralKeyManager.m
```diff
@@ -64,18 +68,44 @@ - (void)handleWillForegroundNotification {
...
+            }
+        }
+        self.createKeyPromise = nil;
+    };
+    
+    if ([self.keyProvider conformsToProtocol:@protocol(STPCustomerEphemeralKeyProvider)]) {
+        id<STPCustomerEphemeralKeyProvider> provider = self.keyProvider;
+        [provider createCustomerKeyWithAPIVersion:self.apiVersion completion:jsonCompletion];
+    }
+    else if ([self.keyProvider conformsToProtocol:@protocol(STPIssuingCardEphemeralKeyProvider)]) {
```
csabol-stripe: style-nit

Stripe/STPEphemeralKeyManager.m
```diff
@@ -64,18 +68,44 @@ - (void)handleWillForegroundNotification {
...
-                self.customerKey = key;
-            }
+        [self getOrCreateKey:^(__unused STPEphemeralKey * _Nullable ephemeralKey, __unused NSError * _Nullable error) {
+            
         }];
     }
 }
 
-- (void)getCustomerKey:(STPEphemeralKeyCompletionBlock)completion {
+- (void)createKey {
```
csabol-stripe: style-nit: (this is a new one I'm trying to enforce so a lot of existing code doesn't follow it). Internal methods should start with an underscore to differentiate from public methods `- (void)_createKey`

Stripe/STPEphemeralKeyManager.m
```diff
@@ -64,18 +68,44 @@ - (void)handleWillForegroundNotification {
     // eager refreshses to once per hour.
```
fred-stripe: Minor typo:
```    // eager refreshes to once per hour.```

File: Tests/Tests/STPEphemeralKeyManagerTest.m [hunk 7]
```diff
@@ -32,12 +31,16 @@ - (void)setUp {
     self.apiVersion = @"2015-03-03";
 }
 
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated"
```
csabol-stripe: what are we using here that's deprecated?

Tests/Tests/STPEphemeralKeyManagerTest.m
```diff
@@ -32,12 +31,16 @@ - (void)setUp {
...
+
 - (id)mockKeyProviderWithKeyResponse:(NSDictionary *)keyResponse {
     XCTestExpectation *exp = [self expectationWithDescription:@"createCustomerKey"];
     id mockKeyProvider = OCMProtocolMock(@protocol(STPEphemeralKeyProvider));
     OCMStub([mockKeyProvider createCustomerKeyWithAPIVersion:[OCMArg isEqual:self.apiVersion]
                                                   completion:[OCMArg any]])
     .andDo(^(NSInvocation *invocation) {
+        [invocation retainArguments];
```
csabol-stripe: why is this needed?

Stripe/STPEphemeralKeyManager.m
```diff
@@ -64,18 +68,44 @@ - (void)handleWillForegroundNotification {
...
+            [self.createKeyPromise succeed:key];
+        } else {
+            // the API request failed
+            if (error) {
+                [self.createKeyPromise fail:error];
+            }
+            // the ephemeral key could not be decoded
+            else {
+                [self.createKeyPromise fail:[NSError stp_ephemeralKeyDecodingError]];
+                NSAssert(NO, @"Could not parse the ephemeral key response. Make sure your backend is sending the unmodified JSON of the ephemeral key to your app. For more info, see https://stripe.com/docs/mobile/ios/standard#prepare-your-api");
```
fred-stripe: It might be handy to log the type of key here (Issuing vs Customer), not sure if that's available in this context. My thought is the user will contact us when they see that error, and it'd be nice to know which key provider is causing the issue.

Stripe/PublicHeaders/STPEphemeralKeyProvider.h
```diff
@@ -40,7 +39,44 @@ NS_ASSUME_NONNULL_BEGIN
...
  */
 - (void)createCustomerKeyWithAPIVersion:(NSString *)apiVersion completion:(STPJSONResponseCompletionBlock)completion;
+@end
 
+/**
+ You should make your application's API client conform to this interface.
+ It provides a way for Stripe utility classes to request a new ephemeral key from
+ your backend, which it will use to retrieve and update Stripe API objects.
+ */
+@protocol STPIssuingCardEphemeralKeyProvider <NSObject>
```
csabol-stripe: I'm not sure these need to be two separate protocols. The documentation says that the API client should conform to both of these, which means checking for `keyProvider` protocol conformance will always succeed for both. 

It's not totally clear to me how or why we intend to differentiate between creating customer and card ephemeral keys? Is the intention for the integrator to create two instances of `STPEphemeralKeyManager` and call the right one as needed?

Stripe/PublicHeaders/STPCustomerContext.h
```diff
@@ -35,7 +35,7 @@ NS_ASSUME_NONNULL_BEGIN
  @param keyProvider   The key provider the customer context will use.
  @return the newly-instantiated customer context.
  */
-- (instancetype)initWithKeyProvider:(id<STPEphemeralKeyProvider>)keyProvider;
+- (instancetype)initWithKeyProvider:(id<STPCustomerEphemeralKeyProvider>)keyProvider;
```
fred-stripe: Renaming this class means:

1. We're also going to have to update the Examples, the stripe.com docs, and the [generated docs](https://stripe.github.io/stripe-ios/docs/index.html)
2. We'll need coordinate a release between all three.
3. Users will have to update their existing code when they upgrade, even if they're not using Issuing.

I don't think any of this is blockers, just thinking through the user-facing consequences here.

acavailhez-stripe: @jack-stripe Just to make sure: is it in line with what you expected?

Stripe/STPEphemeralKey.m
acavailhez-stripe: @jack-stripe ?

Stripe/STPEphemeralKeyManager.m
acavailhez-stripe: 👍

Stripe/STPEphemeralKeyManager.m
acavailhez-stripe: 👍

Stripe/STPEphemeralKeyManager.m
acavailhez-stripe: 👍

Stripe/STPEphemeralKeyManager.m
acavailhez-stripe: 👍

Tests/Tests/STPEphemeralKeyManagerTest.m
acavailhez-stripe: @jack-stripe ?

Tests/Tests/STPEphemeralKeyManagerTest.m
acavailhez-stripe: There is a bug using `OCMock` that this avoids: https://github.com/erikdoe/ocmock/issues/147

Stripe/PublicHeaders/STPEphemeralKeyProvider.h
acavailhez-stripe: @jack-stripe ?

Stripe/STPEphemeralKeyManager.m
acavailhez-stripe: 👍

Stripe/PublicHeaders/STPCustomerContext.h
jack-stripe: I believe I safely deprecated `STPEphemeralKeyProvider` by changing it to subclass  `STPCustomerEphemeralKeyProvider` - this would trigger a deprecation warning in existing integrations but not a compiler error. Yes, this will mean a docs update though.

Stripe/STPEphemeralKey.m
jack-stripe: I renamed it to `isEqualToEphemeralKey` ("Resource Key" was an early and terrible name for ephemeral keys)

Tests/Tests/STPEphemeralKeyManagerTest.m
jack-stripe: My mistake, I added this accidentally in my initial testing, feel free to remove these ignores

Stripe/PublicHeaders/STPEphemeralKeyProvider.h
jack-stripe: That's the way I imagined it, yes. You might imagine one API that calls `POST https://myapi.com/users/:id/ephemeral_keys` and one that calls `POST https://myapi.com/cards/:id/ephemeral_keys`. Then you might have one API client client-side that calls the first, and another that calls the second.

The other alternative, as you suggest, is to make a single API - `POST https://myapi.com/ephemeral_keys` (with `user` or `card` parameters) and a single API client to talk to both. The issue is that now the method signature needs to change - we need to tell the app what they should be making an ephemeral key for exactly. So the protocol method would have to be something generic like `createKey:forAssociatedStripeObject:` which just feels like turning a thing to figure out at compile-time into a thing to figure out at runtime.

You raise a good point, though, that this breaks if someone's API client conforms to both protocols. Sadly, I think we'll need to add a property to `STPEphemeralKeyManager` to indicate what kind of key it is intended to make (not the worst thing, as it's a private class). Thoughts @csabol-stripe ?

Stripe/STPEphemeralKeyManager.m
acavailhez-stripe: 👍

Tests/Tests/STPEphemeralKeyManagerTest.m
acavailhez-stripe: 👍

Tests/Tests/STPEphemeralKeyManagerTest.m
csabol-stripe: 👍 a comment referencing this would be helpfule

Stripe/STPEphemeralKey.m
csabol-stripe: ah sorry, I mean why was it changed so it only compares the stripeIDs?

Stripe/PublicHeaders/STPEphemeralKeyProvider.h
csabol-stripe: hmm, what about adding a new method to `STPEphemeralKeyManager` so we have `getCustomerKey:` and `getCardKey:` (or similar)? That way we only have one key manager but the requests are still "typed" at compile time. 

Similarly, what are your thoughts about keeping `STPEphemeralKeyProvider` and just adding `createIssuingCardKeyWithAPIVersion:completion:` to the existing protocol (and marking as optional if we think that would be useful?)

Tests/Tests/STPEphemeralKeyManagerTest.m
acavailhez-stripe: Let me add this

Stripe/PublicHeaders/STPEphemeralKeyProvider.h
jack-stripe: I think I'd prefer to keep a property on `STPEphemeralKeyManager`, because I think it'd be a programmer error to call `getCustomerKey:` and then later `getCardKey:` on the same instance. Nice thing is, assuming that property is set in the initializer, we could assert that the given `keyProvider` conforms to the necessary protocol.

I don't feel as strongly about whether or not to split `STPEphemeralKeyProvider` into 2 protocols, but realistically users are going to be implementing only one of `createCustomerKeyWithAPIVersion:completion` or createIssuingCardKeyWithAPIVersion:completion`, so both would have to be optional. In that case, I think it's a little nicer to just have two protocols so that you can enforce at compile-time (instead of runtime) that the necessary methods are implemented.

jack-stripe: Thoughts @csabol-stripe? Can also discuss in slack on monday : )

csabol-stripe: closing the loop: @jack-stripe and I met to discuss this today and we are happy moving forward with the patch as is :)

Stripe/STPEphemeralKey.m
csabol-stripe: ping on this @jack-stripe ? (not going to block patch because id equivalence makes sense as sufficient to me but would be useful to know the reasoning

jack-stripe: Oh, sorry, missed your comment @csabol-stripe!
I honestly don't remember why I changed that. I don't see a world in which `stripeID` is the same but `secret` is different. I'm happy with either implementation.

acavailhez-stripe:
Thanks everyone

PR #1139, opened 2019-03-05
 [Breaking] Rename "payment method" to "payment option" 

yuki-stripe:
## Summary & Motivation
This change precedes adding `PaymentMethod`.  Unfortunately, we already have a type that collides with this name, so I'm renaming it all to PaymentOption.  

I've opted to aggressively rename most all references to 'payment method' - this means possibly more work for users to upgrade, but will make the upcoming usage of PaymentMethod clear and consistent.

## Testing
All StripeIOS Tests pass, all projects build.


yuki-stripe:
Should I update `MIGRATING.md` in this commit as well?

Example/UI Examples/BrowseViewController.swift
```diff
@@ -9,21 +9,21 @@
...
+        case STPPaymentOptionsViewController
         case STPShippingInfoViewController
         case ChangeTheme
 
         var title: String {
             switch self {
             case .STPPaymentCardTextField: return "Card Field"
             case .STPAddCardViewController: return "Card Form with Billing Address"
-            case .STPPaymentMethodsViewController: return "Payment Method Picker"
+            case .STPPaymentOptionsViewController: return "Payment Method Picker"
```
bg-stripe: nit: "Payment Option Picker"

Stripe/PublicHeaders/STPApplePay.h
```diff
@@ -1,24 +1,24 @@
...
  
  ```
- if ([paymentContext.selectedPaymentMethod isKindOfClass:[STPApplePayPaymentMethod class]]) {
+ if ([paymentContext.selectedPaymentOption isKindOfClass:[STPApplePay class]]) {
     // Don't ask the user for their card number; they want to pay with apple pay.
  }
  ```
  */
-@interface STPApplePayPaymentMethod : NSObject <STPPaymentMethod>
+@interface STPApplePay : NSObject <STPPaymentOption>
```
bg-stripe: Should this be `STPApplePayPaymentOption`?

Tests/Tests/STPPaymentContextSnapshotTests.m
```diff
@@ -54,7 +54,7 @@ - (void)testPushPaymentMethodsSmallTitle {
 
         self.hostViewController.navigationBar.prefersLargeTitles = NO;
         self.paymentContext.largeTitleDisplayMode = UINavigationItemLargeTitleDisplayModeAutomatic;
-        [self.paymentContext pushPaymentMethodsViewController];
+        [self.paymentContext pushPaymentOptionsViewController];
```
mshafrir-stripe: rename the method "testPushPaymentMethodsSmallTitle"

Tests/Tests/STPPaymentContextSnapshotTests.m
```diff
@@ -66,7 +66,7 @@ - (void)testPushPaymentMethodsLargeTitle {
 
         self.hostViewController.navigationBar.prefersLargeTitles = YES;
```
mshafrir-stripe: rename the method "testPushPaymentMethodsLargeTitle"

bg-stripe:
> Should I update MIGRATING.md in this commit as well?

@yuki-stripe ah yeah good call, let's do that too.

Stripe/PublicHeaders/STPApplePay.h
yuki-stripe: Only thought with `STPApplePay` is that the name might be more consistent with the other options - `STPCard`, `STPSource`.  

I don't have a strong opinion either way, LMK what you prefer

bg-stripe: I don't have a strong preference either way either, but since `STPApplePay` isn't backed by a Stripe API object, I'd lean towards being more specific and calling it `STPApplePayPaymentOption`

PR #1140, opened 2019-03-07
Add PaymentMethod API binding models

yuki-stripe:
## Summary
Adding models, according to https://site-admin.stripe.com/docs/api/payment_methods/object

- STPPaymentMethod
  - STPPaymentMethodCard
    - STPPaymentMethodCardChecks
    - STPPaymentMethodThreeDSecureUsage
    - STPPaymentMethodCardWallet
  - STPPaymentMethodCardPresent
  - STPPaymentMethodBillingDetails
    - STPPaymentMethodBillingDetailsAddress

Future PRs will add 
* CardPresent and Wallet on PaymentMethod
* Creating a PaymentMethod
* Using PaymentMethod to create a PaymentIntent

## Motivation
We're adding support for PaymentMethods!

## Testing
Unit tests.  
A separate branch for a future PR also confirms that the PaymentMethod object is created from an API response.


Stripe/PublicHeaders/STPPaymentMethodBillingDetails.h
```diff
@@ -0,0 +1,47 @@
...
+@class STPPaymentMethodBillingDetailsAddress;
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface STPPaymentMethodBillingDetails : NSObject <STPAPIResponseDecodable>
+
+/**
+ Billing information associated with a `STPPaymentMethod` that may be used or required by particular types of payment methods.
+ 
+ @see https://site-admin.stripe.com/docs/api/payment_methods/object#payment_method_object-billing_details
```
mshafrir-stripe: let's not include references to internal URLs (here and elsewhere)

Stripe/PublicHeaders/STPPaymentMethod.h
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ Billing information associated with the PaymentMethod that may be used or required by particular types of payment methods.
+ */
+@property (nonatomic, nullable, readonly) STPPaymentMethodBillingDetails *billingDetails;
+
+/**
+ If this is a card PaymentMethod (ie `self.type == @"card"`), this contains details about the card.
+ */
+@property (nonatomic, nullable, readonly) STPPaymentMethodCard *card;
```
mshafrir-stripe: will you also include card_present and ideal?

Stripe/PublicHeaders/STPPaymentMethodBillingDetails.h
yuki-stripe: Is there an alternative I can use?

Stripe/PublicHeaders/STPPaymentMethod.h
yuki-stripe: Yeah, I will add in a future PR

Stripe/PublicHeaders/STPPaymentMethodBillingDetails.h
mshafrir-stripe: you can drop the site-admin subdomain

Stripe/PublicHeaders/STPPaymentMethodCard.h
```diff
@@ -0,0 +1,76 @@
...
+@class STPPaymentMethodThreeDSecureUsage, STPPaymentMethodCardChecks;
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ Contains details about a user's credit card.
+ 
+ @see https://stripe.com/docs/api/payment_methods/object#payment_method_object-card
+ */
+@interface STPPaymentMethodCard : NSObject <STPAPIResponseDecodable>
```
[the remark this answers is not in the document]
ln-stripe: I think it's best to create a new class because we're renaming/adding a few fields. This way the legacy object isn't tied to the new object.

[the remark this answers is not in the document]
yuki-stripe: Yeah, it's similar with some differences...it feels like we should be able to reuse `STPCard`, but @csabol-stripe and I talked early on about just making a new class in the interest of expedience.  Android also built this out as a separate class.

Stripe/PublicHeaders/STPPaymentMethodBillingDetails.h
```diff
@@ -0,0 +1,46 @@
...
+@class STPPaymentMethodBillingDetailsAddress;
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ Billing information associated with a `STPPaymentMethod` that may be used or required by particular types of payment methods.
+ 
+ @see https://stripe.com/docs/api/payment_methods/object#payment_method_object-billing_details
+ */
+@interface STPPaymentMethodBillingDetails : NSObject <STPAPIResponseDecodable>
```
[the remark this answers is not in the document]
yuki-stripe: ah I should have called this out, I will indeed also use to send up data in a future PR.

Tests/Tests/STPPaymentMethodCardTest.m
```diff
@@ -0,0 +1,49 @@
...
+    NSArray<NSString *> *requiredFields = @[];
+    
+    for (NSString *field in requiredFields) {
+        NSMutableDictionary *response = [[STPTestUtils jsonNamed:STPTestJSONPaymentMethod][@"card"] mutableCopy];
+        [response removeObjectForKey:field];
+        
+        XCTAssertNil([STPPaymentMethodCard decodedObjectFromAPIResponse:response]);
+    }
+    
+    XCTAssert([STPPaymentMethodCard decodedObjectFromAPIResponse:[STPTestUtils jsonNamed:STPTestJSONPaymentMethod][@"card"]]);
```
danj-stripe: same
```suggestion
    XCTAssertNotNil([STPPaymentMethodCard decodedObjectFromAPIResponse:[STPTestUtils jsonNamed:STPTestJSONPaymentMethod][@"card"]]);
```

Stripe/PublicHeaders/STPPaymentMethodCardChecks.h
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ You cannot directly instantiate an `STPPaymentMethodCardChecks`.
+ */
+- (instancetype)init __attribute__((unavailable("You cannot directly instantiate an STPPaymentMethodCardChecks.")));
+
+/**
+ If a address line1 was provided, results of the check, one of ‘pass’, ‘failed’, ‘unavailable’ or ‘unchecked’.
+ */
+@property (nonatomic, nullable, readonly) NSString *addressLine1Check;
```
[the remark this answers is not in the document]
yuki-stripe: Good call.  I think I'll go with an explicit `unknown` case.

Tests/Tests/STPPaymentMethodCardChecksTest.m
```diff
@@ -0,0 +1,34 @@
...
+@interface STPPaymentMethodCardChecksTest : XCTestCase
+
+@end
+
+@implementation STPPaymentMethodCardChecksTest
+
+- (void)testDecodedObjectFromAPIResponse {
+    NSDictionary *response = @{@"address_line1_check": [NSNull null],
+                               @"address_postal_code_check": [NSNull null],
+                               @"cvc_check": [NSNull null]};
```
danj-stripe: It's probably worth updating/adding to this test so that it's checking the possible values (pass, failed, unavailable, unchecked), possibly along with the work to add an enum for them.

Tests/Tests/STPPaymentMethodTest.m
```diff
@@ -0,0 +1,57 @@
...
+@implementation STPPaymentMethodTest
+
+#pragma mark - STPAPIResponseDecodable Tests
+
+- (void)testDecodedObjectFromAPIResponseRequiredFields {
+    NSDictionary *fullJson = [STPTestUtils jsonNamed:STPTestJSONPaymentMethod];
+    
+    XCTAssertNotNil([STPPaymentMethod decodedObjectFromAPIResponse:fullJson], @"can decode with full json");
+    
+    NSArray<NSString *> *requiredFields = @[];
```
danj-stripe: Assuming `identifier`/`stripeId` *is* a required field, it should go in here.

Stripe/STPPaymentMethodThreeDSecureUsage.m
```diff
@@ -0,0 +1,35 @@
...
+#pragma mark - STPAPIResponseDecodable
+
++ (nullable instancetype)decodedObjectFromAPIResponse:(nullable NSDictionary *)response {
+    NSDictionary *dict = [response stp_dictionaryByRemovingNulls];
+    if (!dict || dict[@"supported"] == nil) {
+        return nil;
+    }
+    STPPaymentMethodThreeDSecureUsage *usage = [self new];
+    usage.allResponseFields = dict;
+    usage.supported = [dict stp_boolForKey:@"supported" or:NO];
```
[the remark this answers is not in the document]
yuki-stripe: I think I see what you're saying.  In the non-nil and not a bool case, I probably want to return nil in this method..

I'm going to leave this for now

Tests/Tests/STPPaymentMethodBillingDetailsTest.m
```diff
@@ -0,0 +1,45 @@
...
+    XCTAssert([STPPaymentMethodBillingDetails decodedObjectFromAPIResponse:[STPTestUtils jsonNamed:STPTestJSONPaymentMethod][@"billing_details"]]);
+}
+
+- (void)testDecodedObjectFromAPIResponseMapping {
+    NSDictionary *response = [STPTestUtils jsonNamed:STPTestJSONPaymentMethod][@"billing_details"];
+    STPPaymentMethodBillingDetails *billingDetails = [STPPaymentMethodBillingDetails decodedObjectFromAPIResponse:response];
+    XCTAssertEqualObjects(billingDetails.email, @"jenny@example.com");
+    XCTAssertEqualObjects(billingDetails.name, @"jenny");
+    XCTAssertEqualObjects(billingDetails.phone, @"+15555555555");
+
```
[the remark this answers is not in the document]
yuki-stripe: Good call

Stripe/PublicHeaders/STPPaymentMethod.h
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ `YES` if the object exists in live mode or the value `NO` if the object exists in test mode.
+ */
+@property (nonatomic, readonly) BOOL liveMode;
+
+/**
+ The type of the PaymentMethod, currently only @"card" is supported. The corresponding property (`card`) contains additional information specific to the PaymentMethod type.
+ */
+@property (nonatomic, nullable, readonly) NSString *type;
```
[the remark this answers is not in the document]
yuki-stripe: Ah yeah, I change this in a future PR when I actually do support more than one type.  If it works for you, I'll leave it like this in this PR.

Stripe/PublicHeaders/STPPaymentMethodCardChecks.h
```diff
@@ -25,19 +33,31 @@ NS_ASSUME_NONNULL_BEGIN
...
+/**
+@param string a string representing the check result as returned from
+the Stripe API
+ 
+@see https://stripe.com/docs/api/payment_methods/object#payment_method_object-card-checks
+
+@return an enum value mapped to that string. If the string is unrecognized or nil,
+returns STPPaymentMethodCardCheckResultUnknown.
+*/
++ (STPPaymentMethodCardCheckResult)checkResultFromString:(nullable NSString *)string;
```
danj-stripe: I don't think we need/want to make this part of the public API.

The only place I found where we're exposing our string/enum conversion function is `STPCard.h`, and that's marked as deprecated. The others are either only in the implementation, or are in a `+Private.h`

Stripe/STPPaymentMethod.m
```diff
@@ -14,7 +14,7 @@
 
 @interface STPPaymentMethod ()
 
-@property (nonatomic, nullable) NSString *identifier;
+@property (nonatomic, nullable) NSString *stripeId;
```
danj-stripe: I just noticed these properties don't have `strong`/`copy`. I believe the STYLEGUIDE.md wants it to always be included:
https://github.com/stripe/stripe-ios/blob/master/STYLEGUIDE.md#properties

I don't think we usually include `strong`/`copy`/`weak`/`assign` in the `.h` file for `readonly` properties, but I believe it's always included in the `readwrite` re-definition in the `.m` file.

It'd be good to update them all, but I especially think that all of the `NSString` properties in this PR should be `copy`. Faux Pas should catch that, but maybe it doesn't catch it if it's omitted?

Stripe/STPPaymentMethod.m
```diff
@@ -29,23 +29,50 @@ @interface STPPaymentMethod ()
...
 
+- (NSString *)description {
+    NSArray *props = @[
+                       // Object
+                       [NSString stringWithFormat:@"%@: %p", NSStringFromClass([self class]), self],
+                       
+                       // Identifier
+                       [NSString stringWithFormat:@"stripeId = %@", self.stripeId],
+                       
+                       // STPPaymentMethod details (alphabetical)
```
danj-stripe: These aren't actually alphabetical. It's probably worth alphabetizing them because I think it makes it easier to consume the logged value, but you could also just update the comment.

Stripe/STPPaymentMethodBillingDetailsAddress.m
```diff
@@ -16,9 +16,26 @@ @interface STPPaymentMethodBillingDetailsAddress ()
...
 
-
 @implementation STPPaymentMethodBillingDetailsAddress
 
+- (NSString *)description {
+    NSArray *props = @[
+                       // Object
+                       [NSString stringWithFormat:@"%@: %p", NSStringFromClass([self class]), self],
+                       
+                       // Properties
```
danj-stripe: This definitely makes sense as you have it: *not* alphabetized.

Stripe/STPPaymentMethod.m
yuki-stripe: Ahh thanks, I will make all the properties conform to the styleguide.  

Also agree `copy` should be here, I thought faux pas caught this for me before.

PR #1141, opened 2019-03-08
Add API to create PaymentMethod

yuki-stripe:
## Summary
Add `createPaymentMethod...` method to `STPAPIClient`

## Motivation
We're adding support for PaymentMethods!

## Testing
Added a unit test for STPPaymentMethodParams
Added a functional test that exercises `createPaymentMethod`

### DIFF
--- a/Stripe/PublicHeaders/STPAPIClient.h
+++ b/Stripe/PublicHeaders/STPAPIClient.h
@@ -371,9 +371,9 @@ static NSString *const STPSDKVersion = @"15.0.0";
 /**
  Creates a PaymentMethod object with the provided params object.
  
- @see https://site-admin.stripe.com/docs/api/payment_methods/create
+ @see https://stripe.com/docs/api/payment_methods/create
  
- @param paymentMethodParams  The `STPPaymentMethodParams` to pass to `/payment_methods`.  Cannot be nil.
+ @param paymentMethodParams  The `STPPaymentMethodParams` to pass to `/v1/payment_methods`.  Cannot be nil.
  @param completion           The callback to run with the returned PaymentMethod object, or an error.
  */
 - (void)createPaymentMethodWithParams:(STPPaymentMethodParams *)paymentMethodParams
--- a/Stripe/PublicHeaders/STPAPIClient.h
+++ b/Stripe/PublicHeaders/STPAPIClient.h
@@ -21,7 +21,7 @@ NS_ASSUME_NONNULL_BEGIN
 static NSString *const STPSDKVersion = @"15.0.0";
 
 @class STPBankAccount, STPBankAccountParams, STPCard, STPCardParams, STPConnectAccountParams;
-@class STPPaymentConfiguration, STPPaymentIntentParams, STPSourceParams, STPToken;
+@class STPPaymentConfiguration, STPPaymentIntentParams, STPSourceParams, STPToken, STPPaymentMethodParams;
 
 /**
  A top-level class that imports the rest of the Stripe SDK.
@@ -357,6 +357,27 @@ static NSString *const STPSDKVersion = @"15.0.0";
  */
 - (void)confirmPaymentIntentWithParams:(STPPaymentIntentParams *)paymentIntentParams
                             completion:(STPPaymentIntentCompletionBlock)completion;
+
+@end
+
+
+#pragma mark Payment Methods
+
+/**
+ STPAPIClient extensions for working with PaymentMethod objects.
+ */
+@interface STPAPIClient (PaymentMethods)
+
+/**
+ Creates a PaymentMethod object with the provided params object.
+ 
+ @see https://site-admin.stripe.com/docs/api/payment_methods/create
+ 
+ @param paymentMethodParams  The `STPPaymentMethodParams` to pass to `/payment_methods`.  Cannot be nil.
+ @param completion           The callback to run with the returned PaymentMethod object, or an error.
+ */
+- (void)createPaymentMethodWithParams:(STPPaymentMethodParams *)paymentMethodParams
+                           completion:(STPPaymentMethodCompletionBlock)completion;
 
 @end
 
--- a/Stripe/PublicHeaders/STPPaymentMethodBillingDetails.h
+++ b/Stripe/PublicHeaders/STPPaymentMethodBillingDetails.h
@@ -9,6 +9,7 @@
 #import <Foundation/Foundation.h>
 
 #import "STPAPIResponseDecodable.h"
+#import "STPFormEncodable.h"
 
 @class STPPaymentMethodBillingDetailsAddress;
 
@@ -19,7 +20,7 @@ NS_ASSUME_NONNULL_BEGIN
  
  @see https://stripe.com/docs/api/payment_methods/object#payment_method_object-billing_details
  */
-@interface STPPaymentMethodBillingDetails : NSObject <STPAPIResponseDecodable>
+@interface STPPaymentMethodBillingDetails : NSObject <STPAPIResponseDecodable, STPFormEncodable>
 
 /**
  Billing address.
--- a/Stripe/PublicHeaders/STPPaymentMethodBillingDetails.h
+++ b/Stripe/PublicHeaders/STPPaymentMethodBillingDetails.h
@@ -25,22 +25,22 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  Billing address.
  */
-@property (nonatomic, nullable) STPPaymentMethodBillingDetailsAddress *address;
+@property (nonatomic, strong, nullable, readwrite) STPPaymentMethodBillingDetailsAddress *address;
 
 /**
  Email address.
  */
-@property (nonatomic, copy, nullable) NSString *email;
+@property (nonatomic, copy, nullable, readwrite) NSString *email;
 
 /**
  Full name.
  */
-@property (nonatomic, copy, nullable) NSString *name;
+@property (nonatomic, copy, nullable, readwrite) NSString *name;
 
 /**
  Billing phone number (including extension).
  */
-@property (nonatomic, copy, nullable) NSString *phone;
+@property (nonatomic, copy, nullable, readwrite) NSString *phone;
 
 @end
 
--- a/Stripe/PublicHeaders/STPPaymentMethodBillingDetails.h
+++ b/Stripe/PublicHeaders/STPPaymentMethodBillingDetails.h
@@ -25,22 +25,22 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  Billing address.
  */
-@property (nonatomic, strong, nullable, readwrite) STPPaymentMethodBillingDetailsAddress *address;
+@property (nonatomic, strong, nullable) STPPaymentMethodBillingDetailsAddress *address;
 
 /**
  Email address.
  */
-@property (nonatomic, copy, nullable, readwrite) NSString *email;
+@property (nonatomic, copy, nullable) NSString *email;
 
 /**
  Full name.
  */
-@property (nonatomic, copy, nullable, readwrite) NSString *name;
+@property (nonatomic, copy, nullable) NSString *name;
 
 /**
  Billing phone number (including extension).
  */
-@property (nonatomic, copy, nullable, readwrite) NSString *phone;
+@property (nonatomic, copy, nullable) NSString *phone;
 
 @end
 
--- a/Stripe/PublicHeaders/STPPaymentMethodCardParams.h
+++ b/Stripe/PublicHeaders/STPPaymentMethodCardParams.h
@@ -22,12 +22,12 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  Two-digit number representing the card's expiration month.
  */
-@property (nonatomic, nullable) NSString *expMonth;
+@property (nonatomic) NSUInteger expMonth;
 
 /**
  Two- or four-digit number representing the card's expiration year.
  */
-@property (nonatomic, nullable) NSString *expYear;
+@property (nonatomic) NSUInteger expYear;
 
 /**
  For backwards compatibility, you can alternatively set this as a Stripe token (e.g., for apple pay)
--- a/Stripe/PublicHeaders/STPPaymentMethodCardParams.h
+++ b/Stripe/PublicHeaders/STPPaymentMethodCardParams.h
@@ -12,6 +12,9 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
+/**
+ The user's card details.
+ */
 @interface STPPaymentMethodCardParams : NSObject <STPFormEncodable>
 
 /**
--- a/Stripe/PublicHeaders/STPPaymentMethodCardParams.h
+++ b/Stripe/PublicHeaders/STPPaymentMethodCardParams.h
@@ -20,27 +20,27 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  The card number, as a string without any separators. Ex. @"4242424242424242"
  */
-@property (nonatomic, copy, nullable, readwrite) NSString *number;
+@property (nonatomic, copy, nullable) NSString *number;
 
 /**
  Two-digit number representing the card's expiration month.
  */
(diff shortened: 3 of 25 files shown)


Stripe/STPAPIClient.m
```diff
@@ -621,3 +623,25 @@ - (void)confirmPaymentIntentWithParams:(STPPaymentIntentParams *)paymentIntentPa
 }
 
 @end
+
+#pragma mark - Payment Methods
+
+@implementation STPAPIClient (PaymentMethods)
+
+- (void)createPaymentMethodWithParams:(STPPaymentMethodParams *)paymentMethodParams
```
mshafrir-stripe: will you also add a version with just a payment method id?

ln-stripe: that would just be a "retrieve" not "create" right? but yes, +1 to also adding

yuki-stripe: My current understanding is that we won't support 'retrieve' for now (https://stripe.slack.com/archives/C5522BZPS/p1551915861432300)

Stripe/PublicHeaders/STPPaymentMethodParams.h
```diff
@@ -0,0 +1,59 @@
...
+@property (nonatomic, nullable) NSDictionary<NSString *, NSString *> *metadata;
+
+/**
+ Creates params for a card PaymentMethod.
+ 
+ @param card                An object containing the user's card details.
+ @param billingDetails      An object containing the user's billing details.
+ @param metadata            Additional information to attach to the PaymentMethod.
+ */
++ (STPPaymentMethodParams *)cardParamsWithCard:(STPPaymentMethodCardParams *)card
```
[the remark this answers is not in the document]
yuki-stripe: Ah I was making this look like `STPSourceParams` (the convention there is like `cardParamsWithCard:...`, `idealParamsWithAmount:...`) but`paymentMethodParamsWithCard:` makes more sense to me.  

I might name it `paramsWithCard` instead because otherwise the call-site would look like `[STPPaymentMethodParams paymentMethodParamsWithCard:...]` which is redundant.

Stripe/STPAPIClient.m
ln-stripe: oh, sorry, i missed that. did we already build it for android? if yes, it seems unfortunate to have that gap (but if there's nothing we can do about it, that's ok)

mshafrir-stripe: Android has retrieve PaymentIntent

yuki-stripe: I'll add this to my TODO list, and add in a later PR.

Stripe/PublicHeaders/STPAPIClient.h
```diff
@@ -360,6 +360,27 @@ static NSString *const STPSDKVersion = @"15.0.0";
...
+
+/**
+ STPAPIClient extensions for working with PaymentMethod objects.
+ */
+@interface STPAPIClient (PaymentMethods)
+
+/**
+ Creates a PaymentMethod object with the provided params object.
+ 
+ @see https://site-admin.stripe.com/docs/api/payment_methods/create
```
ln-stripe: Let's remove the `site-admin` part of this and just use the normal "stripe.com" url

Stripe/PublicHeaders/STPAPIClient.h
```diff
@@ -360,6 +360,27 @@ static NSString *const STPSDKVersion = @"15.0.0";
...
+ 
+ @param paymentMethodParams  The `STPPaymentMethodParams` to pass to `/payment_methods`.  Cannot be nil.
```
ln-stripe: do we always document endpoints like this instead of saying `/v1/payment_methods`?

Tests/Tests/STPPaymentMethodFunctionalTest.m
```diff
@@ -0,0 +1,92 @@
...
+@import Stripe;
+
+@interface STPPaymentMethodFunctionalTest : STPNetworkStubbingTestCase
+
+@end
+
+@implementation STPPaymentMethodFunctionalTest
+
+- (void)setUp {
+//    self.recordingMode = YES;
```
ln-stripe: debugging?

Tests/Tests/STPPaymentMethodFunctionalTest.m
```diff
@@ -0,0 +1,92 @@
...
+    [super setUp];
+}
+
+- (void)testCreatePaymentMethod {
+    STPAPIClient *client = [[STPAPIClient alloc] initWithPublishableKey:@"pk_test_vOo1umqsYxSrP5UXfOeL3ecm"];
```
ln-stripe: is this key meant to be hardcoded?

Tests/Tests/STPPaymentMethodFunctionalTest.m
```diff
@@ -0,0 +1,92 @@
...
+    billingDetails.address = billingAddress;
+    billingDetails.email = @"email@email.com";
+    billingDetails.name = @"Isaac Asimov";
+    billingDetails.phone = @"555-555-5555";
+    
+    
+    STPPaymentMethodParams *params = [STPPaymentMethodParams paramsWithCard:card
+                                                                 billingDetails:billingDetails
+                                                                       metadata:@{@"test_key": @"test_value"}];
+    XCTestExpectation *expectation = [self expectationWithDescription:@"Payment Method retrieve"];
```
ln-stripe: this should be "PaymentMethod create" i think

Stripe/PublicHeaders/STPAPIClient.h
yuki-stripe: From what I can see it's not a convention, I just copied `confirmPaymentIntentWithParams`'s docs.  

I'll change to `/v1/payment_methods`

Tests/Tests/STPPaymentMethodFunctionalTest.m
yuki-stripe: I've seen this in a few other functional tests and it was helpful for me, so I'm leaning towards keeping it.  I don't know if there's a particular reason this is in some but not all tests.

Tests/Tests/STPPaymentMethodFunctionalTest.m
yuki-stripe: Yeah this seems to be the pattern in functional tests.

Stripe/STPPaymentMethodParams.m
```diff
@@ -0,0 +1,47 @@
...
++ (STPPaymentMethodParams *)paramsWithCard:(STPPaymentMethodCardParams *)card billingDetails:(STPPaymentMethodBillingDetails *)billingDetails metadata:(NSDictionary<NSString *,NSString *> *)metadata {
+    STPPaymentMethodParams *params = [STPPaymentMethodParams new];
+    params.type = STPPaymentMethodParamsTypeCard;
+    params.card = card;
+    params.billingDetails = billingDetails;
+    params.metadata = metadata;
+    return params;
+}
+
+- (NSString *)apiStringType {
```
danj-stripe: Nice to have would be to follow the pattern that `STPSourceParams` uses re: `type` & `rawTypeString`.

Tests/Tests/STPPaymentMethodFunctionalTest.m
danj-stripe: :+1: The key *is* publishable 😀 

IDK if it's as important with the network stubbing functionality that was recently added, but we'd certainly need a key if/when re-recording.

Stripe/PublicHeaders/STPPaymentMethodParams.h
```diff
@@ -0,0 +1,59 @@
...
+#import "STPFormEncodable.h"
+
+@class STPPaymentMethodBillingDetails, STPPaymentMethodCardParams;
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ Types of a PaymentMethod
+ */
+typedef NS_ENUM(NSUInteger, STPPaymentMethodParamsType) {
```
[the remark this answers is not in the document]
yuki-stripe: Oh great call, greed.  PaymentIntent has `STPPaymentIntentEnums`, I'll follow that pattern.

Stripe/PublicHeaders/STPPaymentMethodParams.h
```diff
@@ -0,0 +1,59 @@
...
+    STPPaymentMethodParamsTypeCard,
```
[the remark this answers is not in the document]
yuki-stripe: Thanks for the tip!

yuki-stripe:
Should have addressed all the feedback, ptal @danj-stripe

Stripe/PublicHeaders/STPPaymentMethodBillingDetails.h
```diff
@@ -19,27 +20,27 @@ NS_ASSUME_NONNULL_BEGIN
...
  @see https://stripe.com/docs/api/payment_methods/object#payment_method_object-billing_details
  */
-@interface STPPaymentMethodBillingDetails : NSObject <STPAPIResponseDecodable>
+@interface STPPaymentMethodBillingDetails : NSObject <STPAPIResponseDecodable, STPFormEncodable>
 
 /**
  Billing address.
  */
-@property (nonatomic, nullable) STPPaymentMethodBillingDetailsAddress *address;
+@property (nonatomic, strong, nullable, readwrite) STPPaymentMethodBillingDetailsAddress *address;
```
csabol-stripe: any particular reason for adding explicitly `strong` and `readwrite` (which are the defaults)?

Stripe/PublicHeaders/STPPaymentMethodCardParams.h
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ For backwards compatibility, you can alternatively set this as a Stripe token (e.g., for apple pay)
+ */
+@property (nonatomic, copy, nullable, readwrite) NSString *token;
+
+/**
+ Card security code. It is highly recommended to always include this value.
+ */
+@property (nonatomic, copy, nullable, readwrite) NSString *cvc;
```
csabol-stripe: same as above. Can we omit `readwrite` for clarity? (that would be my preference)

Stripe/PublicHeaders/STPPaymentMethodParams.h
```diff
@@ -0,0 +1,69 @@
...
+/**
+ An object representing parameters used to create a PaymentMethod object.
+ @see https://stripe.com/docs/api/payment_methods/create
+ */
+@interface STPPaymentMethodParams : NSObject <STPFormEncodable>
+
+/**
+ The type of payment method.  The associated property will contain additional information (e.g. `type == STPPaymentMethodTypeCard` means `card` should also be populated).
+ */
+@property (nonatomic, assign, readwrite) STPPaymentMethodType type;
```
csabol-stripe: same as above. `assign` and `readwrite` are the defaults

Stripe/STPPaymentMethod+Private.h
```diff
@@ -0,0 +1,20 @@
...
+//
+//  Created by Yuki Tokuhiro on 3/12/19.
+//  Copyright © 2019 Stripe, Inc. All rights reserved.
+//
+
+#import "STPPaymentMethod.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface STPPaymentMethod ()
```
csabol-stripe: nit: I find it useful to always give a name to a category, so something like `@interface STPPaymentMethod (Private)`

Tests/Tests/STPPaymentMethodFunctionalTest.m
csabol-stripe: yeah we're not really consistent. I'm fine with this if you found it helpful

Stripe/PublicHeaders/STPPaymentMethodBillingDetails.h
yuki-stripe: Haha ahh I added them after reading the Styleguide https://github.com/stripe/stripe-ios/blob/master/STYLEGUIDE.md#properties

I interpreted it as saying we should be explicit.

Stripe/STPPaymentMethod+Private.h
danj-stripe: Last week I learned that we usually *don't* want names here. Without a name it's a class extension, but adding a name turns it into a category.

The compiler will require all methods declared in the extension to have implementations. That restriction is relaxed for categories. It looks like the Faux Pas rule for that is disabled for stripe-ios, but I think it's still a good idea to prefer class extensions.

http://fauxpasapp.com/rules/#rule-PrivateCategory

Stripe/PublicHeaders/STPPaymentMethodEnums.h
```diff
@@ -0,0 +1,22 @@
+//
+//  Header.h
```
danj-stripe: nit:
```suggestion
//  STPPaymentMethodEnums.h
```

Stripe/STPPaymentMethodParams.m
```diff
@@ -7,25 +7,29 @@
...
-        case STPPaymentMethodParamsTypeCard:
-            return @"card";
+- (STPPaymentMethodType)type {
+    return [STPPaymentMethod typeFromString:self.rawTypeString];
+}
+
+- (void)setType:(STPPaymentMethodType)type {
+    // If setting unknown and we're already unknown, don't want to override raw value
+    if (type != self.type) {
+        self.rawTypeString = [STPPaymentMethod stringFromType:type];
```
danj-stripe: It looks like setting `type = .unknown` will set `rawTypeString = nil`, but `rawTypeString` is not `nullable`

Should it be `nullable`? Do we want to add an assert, that they never set `type = .unknown`?  Something else?

(It looks like this bug also exists in `STPSourceParams`, which I know I said would be a good model... sorry! Can you update that class too?)

yuki-stripe: Setting `type = .unknown` seems like user error to me, an assertion against that sounds ok.  

Come to think of it, why is `type` publicly writable anyways?

danj-stripe: Great question! I suspect if it was `readonly`, that'd work.

It's possible to create a `SourcesParam` using just alloc/init, and then manually setting any fields needed. The static constructors make it really easy to do the right thing for the common case, for all the various types.

The combination of the `rawTypeString` and `additionalAPIParameters` property allow full control over the Source/PaymentMethod being created, since we don't necessarily keep up with every possible change to available sources/paymentmethods.

Plausibly there's a use case for building a `type` that the SDK knows about, but one where the static constructor interferes, in which case being able to set the `type` using an enum *might* be nice. I think that's a sufficiently rare case that `readonly` is 👍. It'd be a breaking change on SourceParams, but I think that's probably fine.

csabol-stripe: Always a fan of more readonlys :)

yuki-stripe: Looks like the case described actually happens in the SDK already for `STPSourceParams`:
```
            STPSourceParams *params = [STPSourceParams new];
            params.type = STPSourceTypeCard;
            params.token = token.tokenId;
```

I'm thinking I'll make `type` readonly for `STPPaymentMethodParams` but add an assertion for `STPSourceParams`.

yuki-stripe: Looking at the tests, it really seems we intended to accommodate setting this value to Unknown.  e.g. one test does `XCTAssertNil(sourceParams.rawTypeString);`.  I'm inclined now to instead mark `rawTypeString` as `nullable` on `STPSourceParams`.

Stripe/STPPaymentMethod+Private.h
```diff
@@ -10,7 +10,7 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-@interface STPPaymentMethod ()
+@interface STPPaymentMethod (Private)
```
danj-stripe: Per my reply to @csabol-stripe's comment, I think this is a mistake. You're opting out of a compiler warning if methods declared in this header disappear/change names.

PR #1142, opened 2019-03-11
Add PaymentMethod to PaymentIntentParams

yuki-stripe:
## Summary
* Adds paymentMethod, paymentMethodId to STPPaymentIntentParams
* Deprecates `saveSourceToCustomer` on `STPPaymentIntentParams`, replaced by `savePaymentMethod`

## Motivation
Adding support for PaymentMethod

## Testing
Added another semi-manual test `disabled_testConfirmPaymentIntentWithCardPaymentMethodSucceeds` to STPPaymentIntentFunctionalTest and confirmed it passes with a manually created PaymentIntent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
   * Renames `STPPaymentMethodsViewController` to `STPPaymentOptionsViewController`
   * Renames all properties, methods, comments referencing 'PaymentMethod' to 'PaymentOption'
 * Adds `[STPAPI createPaymentMethodWithParams:completion:]`, which creates a PaymentMethod.
-* Adds paymentMethod, paymentMethodId to STPPaymentIntentParams
+* Adds paymentMethodParams, paymentMethodId to STPPaymentIntentParams.
 * Deprecates `saveSourceToCustomer` on `STPPaymentIntentParams`, replaced by `savePaymentMethod`
 
 ## 14.0.0 2018-11-14
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
   * Renames `STPPaymentMethodsViewController` to `STPPaymentOptionsViewController`
   * Renames all properties, methods, comments referencing 'PaymentMethod' to 'PaymentOption'
 * Adds `[STPAPI createPaymentMethodWithParams:completion:]`, which creates a PaymentMethod.
+* Adds paymentMethod, paymentMethodId to STPPaymentIntentParams
+* Deprecates `saveSourceToCustomer` on `STPPaymentIntentParams`, replaced by `savePaymentMethod`
 
 ## 14.0.0 2018-11-14
 * Changes `STPPaymentCardTextField`, which now copies the `cardParams` property. See [MIGRATING.md](/MIGRATING.md) for more details. [#1031](https://github.com/stripe/stripe-ios/pull/1031)
--- a/Stripe/PublicHeaders/STPPaymentIntentParams.h
+++ b/Stripe/PublicHeaders/STPPaymentIntentParams.h
@@ -12,17 +12,17 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-@class STPSourceParams;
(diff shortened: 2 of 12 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,8 @@
   * Renames `STPPaymentMethodsViewController` to `STPPaymentOptionsViewController`
   * Renames all properties, methods, comments referencing 'PaymentMethod' to 'PaymentOption'
 * Adds `[STPAPI createPaymentMethodWithParams:completion:]`, which creates a PaymentMethod.
+* Adds paymentMethod, paymentMethodId to STPPaymentIntentParams
```
ln-stripe: "Adds paymentMethodParams, paymentMethodId"?

Tests/Tests/STPPaymentIntentFunctionalTest.m
```diff
@@ -97,23 +99,25 @@ - (void)testConfirmCanceledPaymentIntentFails {
...
 
                                     [expectation fulfill];
                                 }];
     [self waitForExpectationsWithTimeout:5 handler:nil];
 }
 
+#pragma mark - Disabled Tests
 /*
- This test exists so that you can manually plug in an id + secret and verify that confirming a PaymentIntent
+ These tests exists so that you can manually plug in an id + secret and verify that confirming a PaymentIntent
```
ln-stripe: "These tests exist"

Stripe/PublicHeaders/STPPaymentIntentParams.h
```diff
@@ -90,6 +105,17 @@ NS_ASSUME_NONNULL_BEGIN
...
 @property (nonatomic, copy, nullable, readwrite) NSString *returnUrl __attribute__((deprecated("returnUrl has been renamed to returnURL", "returnURL")));
 
+
+/**
+ `@YES` to save this PaymentIntent’s Source to the associated Customer,
+ if the Source is not already attached.
+ 
+ This should be a boolean NSNumber, so that it can be `nil`
+ 
+ This property has been renamed to `savePaymentMethodToCustomer` and deprecated.
```
csabol-stripe: is this accurate? If I set `sourceParams` but not `paymentMethodParams` and then set `savePaymentMethod` will it save it the source? Reading the change to the comment above that seems accurate, but I'm wondering if we should keep them separate? Did you have a particular reason for combining?

Stripe/PublicHeaders/STPPaymentIntentParams.h
```diff
@@ -47,6 +47,21 @@ NS_ASSUME_NONNULL_BEGIN
...
+ @note alternative to `paymentMethodId`
+ */
+@property (nonatomic, strong, nullable, readwrite) STPPaymentMethodParams *paymentMethodParams;
+
+/**
+ Provide an already created PaymentMethod's id, and it will be used to confirm the PaymentIntent.
+ 
+ @note alternative to `paymentMethodParams`
+ */
+@property (nonatomic, copy, nullable, readwrite) NSString *paymentMethodId;
```
csabol-stripe: I much prefer `Identifier` to `Id` or `ID`. Since we already have `sourceId` on this class I guess `paymentMethodId` makes sense though so just consider this a note for any future/new work :)

Stripe/PublicHeaders/STPPaymentIntentParams.h
yuki-stripe: I think it's accurate - "save_source_to_customer has been renamed to save_payment_method." 
 (https://stripe.com/docs/upgrades#2019-02-11)

If you're thinking that `savePaymentMethod` really seems to imply it's..for PaymentMethod, not Source, I agree though.  I think combining makes sense because the semantics are basically the same to me and it matches the single parameter that the API expects, but the name is confusing.

ln-stripe: Yes, it is accurate!

csabol-stripe: okay sounds good! Thanks for the clarification

PR #1146, opened 2019-03-13
Update for mid-Feb API changes

yuki-stripe:
## Summary
Updating to accommodate some API changes https://stripe.com/docs/upgrades#2019-02-11
* Deprecates `STPPaymentIntentsStatusRequiresSource`, replaced by `STPPaymentIntentsStatusRequiresPaymentMethod`
* Deprecates `STPPaymentIntentsStatusRequiresSourceAction`, replaced by `STPPaymentIntentsStatusRequiresAction`
* Deprecates `STPPaymentIntentSourceAction` and `STPPaymentSourceActionAuthorizeWithURL` classes, replaced by `STPPaymentIntentAction` and `STPPaymentActionRedirectToURL`
* Deprecates `nextSourceAction` on `STPPaymentIntent`, replaced by `nextAction`

See https://paper.dropbox.com/doc/API-Changes-for-mid-Feb-version--AZT10970k8GfUgnOyuAeN~hOAg-PC7zkuMmx0i42AeypiP04

## Testing
Reworked `STPPaymentIntentSourceActionTest` -> `STPPaymentIntentActionTest`.
Updated `STPPaymentIntentFunctionalTest`, `STPPaymentIntentTest`

To see what it would look like for a user who was still using the deprecated classes, I did the following:
1. Turn off warnings as errors
2. Added back the old `STPPaymentIntentSourceActionTest`, but replaced the test JSON with the new version (ie "authorize_with_url" -> "redirect_to_url")
Result: The test passes, and you get warnings like this everywhere:
Screen Shot 2019-03-13 at 2 24 09 PM


Stripe/PublicHeaders/STPPaymentIntent.h
```diff
@@ -82,9 +82,16 @@ NS_ASSUME_NONNULL_BEGIN
 
 /**
  If `status == STPPaymentIntentStatusRequiresSourceAction`, this
- property contains the next action to take for this PaymentIntent.
+ property contains the next source action to take for this PaymentIntent.
+ @deprecated Use nextAction instead
  */
-@property (nonatomic, nullable, readonly) STPPaymentIntentSourceAction* nextSourceAction;
+@property (nonatomic, nullable, readonly) STPPaymentIntentAction* nextSourceAction DEPRECATED_MSG_ATTRIBUTE("Use nextAction instead");
```
danj-stripe: I think you can/should use `__attribute__((deprecated(...)))` instead, which would provide a fix-it.

I think this is true of every place that `DEPRECATED_MSG_ATTRIBUTE` is used (from AvailabilityMacros.h):

    #define DEPRECATED_MSG_ATTRIBUTE(s) __attribute__((deprecated(s)))

AFAIK, we're not interested in cross-compiler support, so might as well always use the fancy 2-argument `clang` attribute: https://clang.llvm.org/docs/AttributeReference.html#deprecated

Stripe/PublicHeaders/STPPaymentIntentAction.h
```diff
@@ -0,0 +1,50 @@
...
+
+#import "STPAPIResponseDecodable.h"
+#import "STPPaymentIntentEnums.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@class STPPaymentIntentActionRedirectToURL;
+
+/**
+ Source Action details for an STPPaymentIntent. This is a container for
```
danj-stripe: Should this be "PaymentMethod Action", or just "Action"?

Stripe/PublicHeaders/STPPaymentIntentAction.h
```diff
@@ -0,0 +1,50 @@
...
+ */
+@property (nonatomic, readonly) STPPaymentIntentActionType type;
+
+/**
+ The details for authorizing via URL, when `type == STPPaymentIntentActionTypeRedirectToURL`
+ */
+@property (nonatomic, strong, nullable, readonly) STPPaymentIntentActionRedirectToURL* redirectToURL;
+
+/**
+ The details for authorizing via URL, when `type == STPPaymentIntentSourceActionTypeAuthorizeWithURL`
```
danj-stripe: Same feedback as the doc comment on `nextSourceAction`.

My primary concern is that as a dev updating my code, if I don't know that these `types` are the same, is that I look at this and think I have to handle both cases. Then I get frustrated by the deprecation warning on the old one.

Stripe/PublicHeaders/STPPaymentIntentEnums.h
```diff
@@ -14,22 +14,35 @@ typedef NS_ENUM(NSInteger, STPPaymentIntentStatus) {
...
-    STPPaymentIntentStatusRequiresSource,
+    STPPaymentIntentStatusRequiresSource DEPRECATED_MSG_ATTRIBUTE("Use STPPaymentIntentStatusRequiresPaymentMethod"),
 
     /**
      This PaymentIntent needs to be confirmed
      */
     STPPaymentIntentStatusRequiresConfirmation,
+    
+    /**
+     The selected Source requires additional authentication steps.
```
danj-stripe: Might want to update/remove `Source` from here

Stripe/PublicHeaders/STPPaymentIntentEnums.h
```diff
@@ -93,22 +106,48 @@ typedef NS_ENUM(NSInteger, STPPaymentIntentConfirmationMethod) {
     STPPaymentIntentConfirmationMethodSecret,
 };
 
+/**
+ Types of Source Actions from a `STPPaymentIntent`, when the payment intent
```
danj-stripe: Should this not include the word `Source`?

Stripe/PublicHeaders/STPPaymentIntentEnums.h
```diff
@@ -14,22 +14,35 @@ typedef NS_ENUM(NSInteger, STPPaymentIntentStatus) {
...
+     This PaymentIntent requires a PaymentMethod or Source
+     */
+    STPPaymentIntentStatusRequiresPaymentMethod,
 
     /**
      This PaymentIntent requires a Source
+     @deprecated Use STPPaymentIntentStatusRequiresPaymentMethod instead.
      */
-    STPPaymentIntentStatusRequiresSource,
+    STPPaymentIntentStatusRequiresSource __attribute__((deprecated("Use STPPaymentIntentStatusRequiresPaymentMethod", "STPPaymentIntentStatusRequiresPaymentMethod"))) = STPPaymentIntentStatusRequiresPaymentMethod,
```
danj-stripe: 👍

Stripe/STPPaymentIntent.m
```diff
@@ -80,9 +81,14 @@ + (nullable NSString *)idFromClientSecret:(NSString *)clientSecret {
 
 + (STPPaymentIntentStatus)statusFromString:(NSString *)string {
     NSDictionary<NSString *, NSNumber *> *map = @{
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated"
                                                   @"requires_source": @(STPPaymentIntentStatusRequiresSource),
```
danj-stripe:
```suggestion
                                                  @"requires_source": @(STPPaymentIntentStatusRequiresPaymentMethod),
```

And delete the `#pragmas`, thanks to 5819a4c

Stripe/STPPaymentIntentAction.m
```diff
@@ -0,0 +1,63 @@
...
+
+@interface STPPaymentIntentAction()
+
+@property (nonatomic) STPPaymentIntentActionType type;
+@property (nonatomic, strong, nullable) STPPaymentIntentActionRedirectToURL* redirectToURL;
+@property (nonatomic, copy, nonnull, readwrite) NSDictionary *allResponseFields;
+
+@end
+
+@implementation STPPaymentIntentAction
```
danj-stripe: Looks like this is missing the `description` override from `STPPaymentIntentSourceAction`

Tests/Tests/STPPaymentIntentTest.m
```diff
@@ -32,20 +32,27 @@ - (void)testIdentifierFromSecret {
...
 
 - (void)testStatusFromString {
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated"
     XCTAssertEqual([STPPaymentIntent statusFromString:@"requires_source"],
-                   STPPaymentIntentStatusRequiresSource);
+                    STPPaymentIntentStatusRequiresSource);
     XCTAssertEqual([STPPaymentIntent statusFromString:@"REQUIRES_SOURCE"],
-                   STPPaymentIntentStatusRequiresSource);
+                    STPPaymentIntentStatusRequiresSource);
```
danj-stripe: These can be updated to `.requiresPaymentMethod` and the `#pragmas` removed

Tests/Tests/STPPaymentIntentTest.m
```diff
@@ -171,18 +178,35 @@ - (void)testDecodedObjectFromAPIResponseMapping {
     XCTAssertEqualObjects(paymentIntent.stripeDescription, @"My Sample PaymentIntent");
     XCTAssertFalse(paymentIntent.livemode);
     XCTAssertEqualObjects(paymentIntent.receiptEmail, @"danj@example.com");
+    
+    // Deprecated: nextSourceAction
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated"
     XCTAssertNotNil(paymentIntent.nextSourceAction);
```
danj-stripe: It *might* be interesting to also assert object identity equality, since we know that `nextSourceAction` should just call `nextAction` (and similar for the `authorizeWithUrl`/`redirectToURL` pair):

    XCTAssertEqual(paymentIntent.nextSourceAction, paymentIntent.nextAction, @"Should be same object");

Doing this kinda makes the parts of `STPRedirectContextTest.m` that test the deprecated properties *and* new properties superfluous.

Tests/Tests/STPRedirectContextTest.m
```diff
@@ -139,18 +139,30 @@ - (void)testInitWithPaymentIntent {
...
+    if (sut.completion) {
+        sut.completion(fakeError);
+    }
     XCTAssertTrue(completionCalled);
 }
 
 - (void)testInitWithPaymentIntentFailures {
+    // Note next_action has been renamed to next_source_action in the API, but both still get sent down in the 2015-10-12 API
     NSMutableDictionary *json = [[STPTestUtils jsonNamed:STPTestJSONPaymentIntent] mutableCopy];
     json[@"next_source_action"] = [json[@"next_source_action"] mutableCopy];
```
danj-stripe: I don't think you're parsing `next_source_action` anymore, I'd probably clean up this test & remove the mutation of that key.

Stripe/PublicHeaders/STPPaymentIntentAction.h
yuki-stripe: Changing to "Action"

Tests/Tests/STPPaymentIntentTest.m
yuki-stripe: Nice, I like it

Tests/Tests/STPRedirectContextTest.m
yuki-stripe: Yep, I didn't convert this test over properly.

Tests/Tests/STPPaymentIntentTest.m
yuki-stripe: I was too hasty changing this test - we don't parse "requires_source" anymore, so I'm just going to assert `STPPaymentIntentStatusRequiresSource` == `STPPaymentIntentStatusRequiresAction`.  

We *do* still parse "requires_source_action", so I'm going to add assertions that it produces `STPPaymentIntentStatusRequiresAction`, as well as another `STPPaymentIntentStatusRequiresAction` == `STPPaymentIntentStatusRequiresSourceAction`

Stripe/STPPaymentIntent.m
```diff
@@ -81,10 +81,6 @@ + (nullable NSString *)idFromClientSecret:(NSString *)clientSecret {
 
 + (STPPaymentIntentStatus)statusFromString:(NSString *)string {
     NSDictionary<NSString *, NSNumber *> *map = @{
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wdeprecated"
-                                                  @"requires_source": @(STPPaymentIntentStatusRequiresSource),
```
danj-stripe: Do you have other work I haven't seen that uses the newer API version? I thought it was still possible to receive the `requires_source` status string from the Stripe API, just that we'd want to transparently map that to the `.requiresPaymentMethod` case?

Tests/Tests/STPPaymentIntentTest.m
```diff
@@ -49,10 +46,18 @@ - (void)testStatusFromString {
...
+                   STPPaymentIntentStatusRequiresAction);
+    XCTAssertEqual([STPPaymentIntent statusFromString:@"REQUIRES_SOURCE_ACTION"],
+                   STPPaymentIntentStatusRequiresAction);
     XCTAssertEqual([STPPaymentIntent statusFromString:@"requires_action"],
                    STPPaymentIntentStatusRequiresAction);
     XCTAssertEqual([STPPaymentIntent statusFromString:@"REQUIRES_ACTION"],
                    STPPaymentIntentStatusRequiresAction);
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated"
+    XCTAssertEqual(STPPaymentIntentStatusRequiresAction, STPPaymentIntentStatusRequiresSourceAction);
```
danj-stripe: This looks the same as L37, just switched.

Stripe/STPPaymentIntent.m
yuki-stripe: You're right, I confused myself here.  We're still on the 2015 API, so the API will send down "requires_source" instead of "requires_payment_method"...

Issue #1032, opened 2018-10-24
Enable deleting selectedPaymentMethod.


wangjiejacques:
## Summary

From the stripe api [documentation](https://stripe.com/docs/api/cards/delete), the user can delete the selected card, but this is not allowed in the sdk, why this ?
```
if (paymentMethod == self.selectedPaymentMethod) {
    // Cannot detach selected payment method
    return YES;
}
```

## Code to reproduce
```
if (paymentMethod == self.selectedPaymentMethod) {
    // Cannot detach selected payment method
    return YES;
}
```

## iOS version
iOS 12

## Installation method
Pod

## SDK version
13.2.0

csabol-stripe:
Fix is in master. Thanks!

PR #1149, opened 2019-03-18
Allows deleting selected payment option

csabol-stripe:
## Summary
Allows deleting selected payment option
Will automatically end editing if the last deletable payment option is removed.

## Motivation

#1032

## Testing

Tested deleting selected and last payment option in standard integration test app

### DIFF
--- a/Stripe/STPPaymentOptionsInternalViewController.m
+++ b/Stripe/STPPaymentOptionsInternalViewController.m
@@ -307,7 +307,7 @@ static NSInteger const PaymentOptionSectionAddCard = 1;
             // in this function is not allowed)
             dispatch_async(dispatch_get_main_queue(), ^{
                 [self _endTableViewEditing];
-            })
+            });
             // manually set the value passed to reloadRightBarButtonItemWithTableViewIsEditing
             // below
             tableViewIsEditing = NO;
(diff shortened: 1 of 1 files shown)


Stripe/STPPaymentOptionsInternalViewController.m
```diff
@@ -301,8 +300,18 @@ - (void)tableView:(UITableView *)tableView commitEditingStyle:(UITableViewCellEd
         // Perform deletion animation for single row
         [tableView deleteRowsAtIndexPaths:@[indexPath] withRowAnimation:UITableViewRowAnimationAutomatic];
 
+        BOOL tableViewIsEditing = tableView.isEditing;
+        if (![self isAnyPaymentOptionDetachable]) {
+            // we deleted the last available payment option, stop editing
+            [self _endTableViewEditing];
```
yuki-stripe: The docs for this delegate method say we shouldn't do this: 

> You should not call setEditing:animated: within an implementation of this method. If for some reason you must, invoke it after a delay by using the performSelector:withObject:afterDelay: method.

csabol-stripe: Ah, thanks for checking! I'll update

PR #1156, opened 2019-04-09
Improve some STPRedirectContext docstrings to be clearer

yuki-stripe:
## Summary
Clarify some things in `STPRedirectContext`.

## Motivation
Feedback from some dogfooding PaymentIntents: https://paper.dropbox.com/doc/iOS-TODOs-for-bsharon-SCA-hackathon-on-iOS--Aa9DXzHlvLu1_3HXTOJi_nB0Ag-n5GOOs66dZE4lyjzuFwhS

Not too familiar with this code, please confirm these are in fact correct clarifications!

### DIFF
--- a/Stripe/PublicHeaders/STPRedirectContext.h
+++ b/Stripe/PublicHeaders/STPRedirectContext.h
@@ -60,8 +60,11 @@ typedef STPRedirectContextSourceCompletionBlock STPRedirectContextCompletionBloc
 
 /**
  A callback that is executed when the context believes the redirect action has been completed.
+ 
+ @note The STPPaymentIntent originally provided to this class may be out of date,
+ so you should re-fetch it using the clientSecret.
 
- @param clientSecret The client secret of the PaymentIntent
+ @param clientSecret The client secret of the PaymentIntent.
  @param error An error if one occured. Note that a lack of an error does not
  mean that the action was completed successfully, the presence of one confirms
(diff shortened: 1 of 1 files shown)


Stripe/PublicHeaders/STPRedirectContext.h
```diff
@@ -60,8 +60,11 @@ typedef STPRedirectContextSourceCompletionBlock STPRedirectContextCompletionBloc
 
 /**
  A callback that is executed when the context believes the redirect action has been completed.
+ 
+ @note The STPPaymentIntent originally provided to this class may be out of date,
+ so you should re-fetch it using the clientSecret.
```
danj-stripe: 👍  Love it. I'm pretty sure this is *why* the client secret is being provided, but I don't think I made that explicit.

Stripe/PublicHeaders/STPRedirectContext.h
```diff
@@ -91,6 +94,9 @@ typedef void(^STPRedirectContextPaymentIntentCompletionBlock)(NSString *clientSe
  was successful. Use Stripe webhooks on your backend server to listen for Source
  state changes and to make the charge.
  
+ @note You must retain this instance for the duration of the redirect flow.
+ This class dismisses any presented view controller upon deallocation.
```
danj-stripe: 👍

PR #1157, opened 2019-04-11
Disable code signing for framework target

sgruby-reformation:
When integrated via Carthage and targeting a device for testing (not the simulator), the automatic code signing setup in the framework causes the build to fail.

## Summary
Turned off automatic code signing on framework target.

## Motivation
The framework should be set to manual code signing as developers outside of Stripe don't have access to the automatic code signing for the framework. In addition, the framework does not need to be code signed except on deployment and that happens automatically with the application that is embedding the framework.

## Testing
None required.


csabol-stripe:
Sorry about that! #1172 has the same fix so I'm going to close this one :)

PR #1158, opened 2019-04-11
Add configurable support for JCB (Apple Pay)

yuki-stripe:
## Summary
Add configurable support for JCB Apple Pay Payment Network

## Motivation
https://jira.corp.stripe.com/browse/IOS-1212

## Testing
Added a unit test (STPApplePayTest.testJCBPaymentNetwork)

### DIFF
--- a/Stripe/PublicHeaders/STPAPIClient.h
+++ b/Stripe/PublicHeaders/STPAPIClient.h
@@ -227,8 +227,8 @@ static NSString *const STPSDKVersion = @"15.0.0";
  The Stripe supported Apple Pay card networks are:
  American Express, Visa, Mastercard, Discover.
  
- Japanese users can enable JCB, Diners Club and Discover via
- `setJCBPaymentNetworkSupported:`, after they have been approved by JCB.
+ Japanese users can enable JCB by setting `JCBPaymentNetworkSupported` to YES,
+ after they have been approved by JCB.
 
  @return YES if the device is currently able to make Apple Pay payments via one
  of the supported networks. NO if the user does not have a saved card of a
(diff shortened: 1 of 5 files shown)


Stripe/PublicHeaders/STPAPIClient.h
```diff
@@ -271,6 +274,13 @@ static NSString *const STPSDKVersion = @"15.0.0";
                                                    country:(NSString *)countryCode
                                                   currency:(NSString *)currencyCode;
 
+/**
+ Japanese users can enable JCB, Diners Club and Discover by calling this method with `YES`, after they have been approved by JCB.
+ 
+ The default value is NO.
+ */
++ (void)setJCBPaymentNetworkSupported:(BOOL)isSupported;
```
danj-stripe: WDYT about using a class property instead? I think it gives better interop with Swift, but I'm not positive: https://useyourloaf.com/blog/objective-c-class-properties/

yuki-stripe:
Made some changes, would appreciate another look @danj-stripe !

Tests/Tests/STPApplePayTest.m
```diff
@@ -49,6 +49,9 @@ - (void)testCanSubmitPaymentRequestReturnsNOIfMerchantIdentifierIsNil {
 }
 
 - (void)testJCBPaymentNetwork {
+    if (&PKPaymentNetworkJCB == NULL) {
+        return;
```
danj-stripe: If you wanted to be *really* complete you could instead test/assert that `setJCBPaymentNetworkSupported:YES` doesn't crash/add garbage/NULL to the `supportedPKPaymentNetworks` response.

Probably not worth it.

yuki-stripe: Good idea!  `[Stripe supportedPKPaymentNetworks]` crashed in an earlier commit because I forgot this check.

PR #1178, opened 2019-05-08
Fixed payment account API

realityworks:
## Summary

Connect Account data has changed in the Web API and the iOS SDK no longer correctly supports these functions for creating connect account. I have added calls to support the new connect accounts while still having the old Beta Connect calls available for backward compatiility.

## Motivation

The Connect API has been updated.

## Testing

It was test against a REST webapi that forward to the Stripe ConnectAccount API. Connect accounts were created.


csabol-stripe:
We've updated the API in https://github.com/stripe/stripe-ios/releases/tag/v17.0.0 . Thanks for your patience on that fix :)

PR #1185, opened 2019-05-21
Add PaymentMethod APIs to replace Source in UI

yuki-stripe:
## Summary
1 of 2 PRs to migrate the UI (`STPPaymentContext`, `STPOptionsViewController`, etc.) from Source to PaymentMethod
* Add attach, detach, list PaymentMethod APIs to `STPAPIClient`, `STPBackendAPIAdapter`, `STPCustomerContext`.
* Adds convenience methods that will be useful later
* Adds `STPPaymentMethod` conformance to `STPPaymentOption`

## Motivation
Migrating UI from Source to PaymentMethod

## Testing
* Tests the new attach, detach, list APIs in `STPCustomerContextTest`

### DIFF
--- a/Stripe/STPAPIClient+Private.h
+++ b/Stripe/STPAPIClient+Private.h
@@ -81,6 +81,24 @@ toCustomerUsingKey:(STPEphemeralKey *)ephemeralKey
 fromCustomerUsingKey:(STPEphemeralKey *)ephemeralKey
           completion:(STPErrorBlock)completion;
 
+/**
+ Attach a Payment Method to a customer
+ 
+ @see https://stripe.com/docs/api/payment_methods/attach
+ */
++ (void)attachPaymentMethod:(NSString *)paymentMethodID
+         toCustomerUsingKey:(STPEphemeralKey *)ephemeralKey
+                 completion:(STPErrorBlock)completion;
+
+/**
+ Detach a Payment Method from a customer
+ 
+ @see https://stripe.com/docs/api/payment_methods/detach
+ */
++ (void)detachPaymentMethod:(NSString *)paymentMethodID
+       fromCustomerUsingKey:(STPEphemeralKey *)ephemeralKey
+                 completion:(STPErrorBlock)completion;
+
 @end
 
 @interface Stripe (Private)
--- a/Stripe/STPAPIClient+Private.h
+++ b/Stripe/STPAPIClient+Private.h
@@ -99,6 +99,8 @@ fromCustomerUsingKey:(STPEphemeralKey *)ephemeralKey
        fromCustomerUsingKey:(STPEphemeralKey *)ephemeralKey
                  completion:(STPErrorBlock)completion;
 
++ (void)listPaymentMethodsForCustomerUsingKey:(STPEphemeralKey *)ephemeralKey
+                                   completion:(STPPaymentMethodsCompletionBlock)completion;
 @end
 
 @interface Stripe (Private)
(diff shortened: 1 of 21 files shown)


Stripe/STPPaymentMethod.m
```diff
@@ -117,4 +119,39 @@ + (nullable instancetype)decodedObjectFromAPIResponse:(nullable NSDictionary *)r
...
+    switch (self.type) {
+        case STPPaymentMethodTypeCard:
+            if (self.card != nil) {
+                NSString *brand = [STPPaymentMethodCard stringFromBrand:self.card.brand];
+                return [NSString stringWithFormat:@"%@ %@", brand, self.card.last4];
+            } else {
+                return [STPPaymentMethodCard stringFromBrand:STPCardBrandUnknown];
+            }
+        case STPPaymentMethodTypeiDEAL:
+            return STPLocalizedString(@"iDEAL", @"Source type brand name");
```
csabol-stripe: this probably doesn't need to be localized

Stripe/STPPaymentMethodAddress.m
```diff
@@ -18,6 +19,19 @@ @interface STPPaymentMethodAddress ()
...
 
+- (instancetype)initWithAddress:(STPAddress *)address {
+    self = [super init];
+    if (self) {
+        _city = address.city;
+        _country = address.country;
+        _line1 = address.line1;
+        _line2 = address.line2;
+        _postalCode = address.postalCode;
+        _state = address.state;
```
csabol-stripe: copy NSStrings

Stripe/STPPaymentMethodCard.m
```diff
@@ -74,4 +74,27 @@ + (nullable instancetype)decodedObjectFromAPIResponse:(nullable NSDictionary *)r
     return card;
 }
 
+#pragma mark - STPCardBrand
+
++ (NSString *)stringFromBrand:(STPCardBrand)brand {
```
csabol-stripe: we should really have this shared between STPPaymentMethodCard and STPCard (also we should probably rename the latter to STPSourceCard). Mind filing a followup?

Stripe/STPAPIClient+Private.h
```diff
@@ -81,6 +81,26 @@ toCustomerUsingKey:(STPEphemeralKey *)ephemeralKey
...
+/**
+ Detach a Payment Method from a customer
+ 
+ @see https://stripe.com/docs/api/payment_methods/detach
+ */
++ (void)detachPaymentMethod:(NSString *)paymentMethodID
+       fromCustomerUsingKey:(STPEphemeralKey *)ephemeralKey
+                 completion:(STPErrorBlock)completion;
+
++ (void)listPaymentMethodsForCustomerUsingKey:(STPEphemeralKey *)ephemeralKey
```
csabol-stripe: I think we should make this clear that it only fetches Cards (needs doc string)

Stripe/STPPaymentMethodListDeserializer.h
```diff
@@ -0,0 +1,25 @@
...
+#import "STPAPIResponseDecodable.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@class STPPaymentMethod;
+
+/**
+ Deserializes the response returned from https://stripe.com/docs/api/payment_methods/list
+ */
+@interface STPPaymentMethodListDeserializer : NSObject <STPAPIResponseDecodable>
```
csabol-stripe: hmm, kinda surprised we don't hit any other list endpoints. I suspect making this a template class would be useful, but feel free to file a followup for that since we don't have a use case now :P

Stripe/STPPaymentMethodCard.m
yuki-stripe: https://jira.corp.stripe.com/browse/IOS-1226

Stripe/STPPaymentMethodListDeserializer.h
yuki-stripe: ooh good call.  https://jira.corp.stripe.com/browse/IOS-1227

PR #892, opened 2018-02-01
Add ability to create card sources to UI

bg-stripe:
r? @danj-stripe 
cc @teich-stripe @thorsten-stripe @fred-stripe 

## Summary

- Adds a switch (default off) to create card sources to AddCardVC, PaymentMethodsVC, and PaymentContext.

- AddCardVC gets a new delegate method called when a card source is created, and `didCreateToken`/`didCreateSource` are now optional protocol methods (since you only need to implement one of the two).

- [x] add CHANGELOG entry

## Testing

- Added new unit tests for AddCardVC & PaymentMethodsVC. 
- Ran the example app and tested adding a new card & selecting a new default card. Dashboard logs show that sources are being created, and the SDK correctly displays card sources.

screen shot 2018-02-01 at 3 21 28 pm

PR #1186, opened 2019-05-22
Migrate UI components from Source to PaymentMethod

yuki-stripe:
## Summary
* The following have been migrated from Source/Token to PaymentMethod:
  * UI components
    * STPPaymentCardTextField
    * STPAddCardViewController
    * STPPaymentOptionViewController
  * PaymentContext
    * STPPaymentContext
    * STPCustomerContext
    * STPBackendAPIAdapter
    * STPPaymentResult
  * Standard Integration example project

See CHANGELOG and commits for more granular views of the changes here.  

TODO in another PR: Rename the Integration projects, directories. 

## Motivation
Migrating UI from Source to PaymentMethod 

## Testing
Manual Testing:
- [x] Use Apple Pay to buy something.  Confirm that the payment intent exists on the dashboard
- [ ] Add a new card.  
    - [x] Pay with it, confirm the PI exists
    - [x] Make sure it was attached successfully.  Wait 60 seconds, breakpoint in the list API to make sure it is hit again and still returns the card.
    - [ ] (Blocked on https://jira.corp.stripe.com/browse/DX-4013) Remove the card.  Wait 60 seconds, breakpoint in the list API and make sure it’s removed.


Stripe/PublicHeaders/STPAddCardViewController.h
```diff
@@ -50,11 +50,6 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic, strong, nullable) STPUserInformation *prefilledInformation;
 
-/**
- If you're using the token generated from STPAddCardViewController to make a Managed Account, you should set this property to the currency that account will use. Otherwise, you should leave it empty. For more information, see https://stripe.com/docs/api#create_card_token-card-currency
- */
-@property (nonatomic, copy, nullable) NSString *managedAccountCurrency;
```
csabol-stripe: I don't really understand this use case. Do you know how user's who were using token+managedAccountCurrency previously will migrate?

Stripe/PublicHeaders/STPBackendAPIAdapter.h
```diff
@@ -48,77 +48,20 @@ NS_ASSUME_NONNULL_BEGIN
...
+ customer.  For example, `completion(paymentMethods)` (if your call succeeds) or
  `completion(error)` if an error is returned.
-
- @see https://stripe.com/docs/api#update_customer
+ 
  */
-- (void)updateCustomerWithShippingAddress:(STPAddress *)shipping completion:(nullable STPErrorBlock)completion;
-
-#pragma mark - Payment Method
+- (void)listPaymentMethodsForCustomerWithCompletion:(nullable STPPaymentMethodsCompletionBlock)completion;
```
csabol-stripe: I think this completion block should be nonnull, because it doesn't do anything otherwise right?

Stripe/PKPaymentAuthorizationViewController+Stripe_Blocks.m
```diff
@@ -39,28 +38,23 @@ - (void)paymentAuthorizationViewController:(__unused PKPaymentAuthorizationViewC
...
             }
             self.didSucceed = YES;
             completion(PKPaymentAuthorizationStatusSuccess);
         });
     };
-    if (self.createSource) {
-        [self.apiClient createSourceWithPayment:payment completion:(STPSourceCompletionBlock)tokenOrSourceCompletion];
-    }
-    else {
-        [self.apiClient createTokenWithPayment:payment completion:(STPTokenCompletionBlock)tokenOrSourceCompletion];
```
csabol-stripe: wondering about the implications of removing token creation...Do you know what/if any use cases this was supporting?

Stripe/PublicHeaders/STPBackendAPIAdapter.h
yuki-stripe: `STPCustomerContext` passes nil to this and `retrieveCustomer` (also has a nullable completion block) because its implementation caches the response, and it calls them for that side effect in the init.  Conceivably someone could be using these in the same way..

Stripe/PKPaymentAuthorizationViewController+Stripe_Blocks.m
yuki-stripe: It looks like it was first written using tokens, then we added configurable support for Source https://github.com/stripe/stripe-ios/pull/892.  I'd guess we didn't remove tokens then to make this a not-breaking change.

Stripe/PublicHeaders/STPAddCardViewController.h
yuki-stripe: I talked to @danwang-stripe - users who are migrating should get the same functionality[0] for free with PaymentMethods.  I've called this out `CHANGELOG.md` but maybe it should be in `MIGRATING.md`.

[0] I'm not familiar with what it does beyond https://stripe.com/docs/api/tokens/create_card#create_card_token-card-currency

csabol-stripe: 👍

Stripe/PublicHeaders/STPBackendAPIAdapter.h
csabol-stripe: bleh -- kinda hacky to call like this for the side-effect, but if it's a potential breaking change we can keep as is for now

PR #1187, opened 2019-05-22
Update to latest Xcode version 10.2.1

yuki-stripe:
## Summary
* Require PassKit framework to remove Xcode and `export_builds.sh` warnings.  [This was set as optional](https://github.com/stripe/stripe-ios/commit/56338c28904b34c9be87281831bf3382a39bb44e) in v2.0.1).  I'm guessing we did that because the SDK at that time supported iOS 5.0 and PassKit was introduced in iOS 6.0.  
* Remove STPCustomerTest from StripeiOSStaticFramework Copy Files phase.  I accidentally added this in https://github.com/stripe/stripe-ios/commit/85aa37d2b431aa518c956f4077a297b5c9f1a1e5
* Bump travis xcode version from 9.2 to 10.2.1
* Don't use new Xcode 10 build system in `export_builds.sh`.
* Switch from facebook/ios-snapshot-test-case 2.1.4 to uber/ios-snapshot-test-case 6.0.3
* Update OHHTTPStubs from 2.1.4 to 6.0.3
* Workaround for iOS 9.3 erroring w/ “dyld: Library not loaded: /usr/lib/libauto.dylib”
* Update CarthageTest project to Swift 5
* Change travis.yml to open simulator from full path instead of -a "simulator"
* Fix STPColorUtilsTest not compiling on 32-bit
* Make fauxpas run on Xcode 9.3

## Motivation
Support the latest Xcode version!

## Testing
- `export_builds.sh` runs 
- Tests pass
- Set `recordingMode = YES` for a functional test to exercise OHHTTPStubs, ran the test, then set it back to `NO`.  Test passes.

### DIFF
--- a/ci_scripts/export_builds.sh
+++ b/ci_scripts/export_builds.sh
@@ -82,6 +82,7 @@ info "Compiling static framework..."
 cd "${root_dir}" || die "Executing \`cd\` failed"
 
 xcodebuild clean build \
+  -UseModernBuildSystem=NO \
   -workspace "Stripe.xcworkspace" \
   -scheme "StripeiOSStaticFramework" \
   -configuration "Release" \
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 language: objective-c
-osx_image: xcode10.2.1
+osx_image: xcode10.2
 branches:
   only:
     - master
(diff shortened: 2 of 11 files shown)


ci_scripts/export_builds.sh
```diff
@@ -82,6 +82,7 @@ info "Compiling static framework..."
 cd "${root_dir}" || die "Executing \`cd\` failed"
 
 xcodebuild clean build \
+  -UseModernBuildSystem=NO \
```
csabol-stripe: why not?

csabol-stripe: discussed on slack. Will investigate in followup

PR #1212, opened 2019-07-03
Updates example apps to accept only a client_secret

csabol-stripe:
## Summary
Updates sample API clients and adds fetches after calls to server confirm.

## Motivation
Previous example integration showing a full PaymentIntent JSON was unsecure.

## Testing
Verified charges with both example apps


Example/Custom Integration/BrowseExamplesViewController.h
```diff
@@ -30,8 +30,6 @@ typedef void (^STPConfirmPaymentIntentCompletionHandler)(STPBackendResult status
                                   paymentMethod:(NSString *)paymentMethodID
                                       returnURL:(NSString *)returnURL
                                      completion:(STPPaymentIntentCreateAndConfirmHandler)completion;
-- (void)confirmPaymentIntent:(STPPaymentIntent *)paymentIntent completion:(STPConfirmPaymentIntentCompletionHandler)completion;
```
yuki-stripe: `CardManualViewController` should be using this but currently doesn't, we should keep this method.  We don't need to return a `STPPaymentIntent` in the completion handler, it could just be `(STPBackendResult status, nullable NSError *error)`.

I'm fixing this in my last SetupIntents PR.

PR #1218, opened 2019-07-10
Send maxTimeout as a 2 digit string

yuki-stripe:
Also changed maxTimeout type from NSTimeInterval to NSInteger, since the precision is minutes not seconds.

## Motivation
https://jira.corp.stripe.com/browse/MOBILE3DS2-439

## Testing
Ran 3ds2 w/ card (automatic confirmation) example

### DIFF
--- a/Stripe/PublicHeaders/STPThreeDSCustomizationSettings.h
+++ b/Stripe/PublicHeaders/STPThreeDSCustomizationSettings.h
@@ -38,11 +38,11 @@ NS_ASSUME_NONNULL_BEGIN
 
 /**
  `authenticationTimeout` is the total time allowed for a user to complete a 3DS2 authentication
- interaction. This value *must* be at least 5 minutes.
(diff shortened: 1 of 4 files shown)


Stripe/PublicHeaders/STPThreeDSCustomizationSettings.h
```diff
@@ -40,6 +40,8 @@ NS_ASSUME_NONNULL_BEGIN
  `authenticationTimeout` is the total time allowed for a user to complete a 3DS2 authentication
  interaction. This value *must* be at least 5 minutes.
  
+ @note The value is rounded down to the nearest minute.
```
csabol-stripe: 🤦‍♂ I retract my previous statement. You were correct to change this from an NSTimeInterval

yuki-stripe: np!

PR #1229, opened 2019-07-13
Call STPRedirectContext completion handler after SFSafariVC dismissal has completed

joshavant:
## Summary
This changes `STPRedirectContext` so, when a `SFSafariViewController` was presented for the redirect flow, the flow's completion handler will get called once the `SFSafariViewController` has completed dismissal.

## Motivation
Currently, `STPRedirectContext` calls the flow's completion handler independent of the dismissal of `SFSafariViewController`. This breaks in use cases that depend on the `SFSafariViewController` being completely dismissed before continuing.

`SFSafariViewController` is known to break UIKit API expectations around dismissal. When the user taps the 'Done' button in the UI, `SFSafariViewController` manages its own dismissal and framework APIs provide only `-[SFSafariViewControllerDelegate safariViewControllerDidFinish:]` to notify callers that the user has dismissed the view controller and does not provide an opportunity for callers to submit a completion handler. In fact, the documentation for this method states:
> You can perform any necessary cleanup here. The view controller is dismissed afterwards.

For users of the Stripe SDK, this manifests as an unsynchronized call to the flow's completion handler and dismissal of the `SFSafariViewController` that the SDK implements. In our application, this ends up blocking SCA adoption and, therefore, the entire set of PaymentIntent features.

The minimal approach to implementing a dismissal completion handler for `SFSafariViewController` involves relying on `UIViewControllerTransitioning` APIs. A custom `UIViewControllerTransitioningDelegate` subclass with a custom `UIPresentationController` can be used to intercept the end of the dismissal transition via `-[UIPresentationController dismissalTransitionDidEnd:]`. From here, this hook can be delegated out to other layers depending on it.

A basic sample implementation of this approach is available here: https://gist.github.com/joshavant/acca2f1141cb19f062a49f0348c2f311

The changes in this PR apply this approach, which effectively calls the flow's completion handler in all redirects involving the dismissal of the `SFSafariViewController` instance.

## Testing

The new dependency on `UIViewControllerTransitioning` means that this changes testing of this behavior from a unit test to a UI test. This project does not appear to currently be tooled for UI testing. Therefore, no new tests were included. Broken unit tests were removed.

These changes were functionally verified in the following scenarios:
- user taps 'Done' button
- user completes authorization in `SFSafariViewController`
- user fails authorization in `SFSafariViewController`
- user pops out to Safari.app, completes authorization 
- user pops out to Safari.app, fails authorization authorization

## Additional Notes

A new localized string was introduced in the description for `+[NSError stp_cardRequiresAdditionalVerificationToProceedError]`.

### DIFF
--- a/Stripe/STPRedirectContext.m
+++ b/Stripe/STPRedirectContext.m
@@ -48,36 +48,9 @@ typedef void (^STPBoolCompletionBlock)(BOOL success);
 }
 @end
 
-@interface STPSafariViewControllerTransitioningDelegate : NSObject <UIViewControllerTransitioningDelegate>
-@property (nonatomic, weak) id<STPSafariViewControllerDismissalDelegate> dismissalDelegate;
-@end
-
-@implementation STPSafariViewControllerTransitioningDelegate
-- (instancetype)initWithDelegate:(id<STPSafariViewControllerDismissalDelegate>)delegate {
-    self = [super init];
-    if (self) {
-        _dismissalDelegate = delegate;
-    }
-    return self;
-}
-
-- (nullable UIPresentationController *)presentationControllerForPresentedViewController:(UIViewController *)presented
-                                                               presentingViewController:(nullable UIViewController *)presenting
-                                                                   sourceViewController:(__unused UIViewController *)source {
-    #pragma unused (source)
-    STPSafariViewControllerPresentationController *controller = [[STPSafariViewControllerPresentationController alloc] initWithPresentedViewController:presented
-                                                                                                                              presentingViewController:presenting];
-    controller.dismissalDelegate = self.dismissalDelegate;
-    return controller;
-}
-
-@end
-
-
-@interface STPRedirectContext () <SFSafariViewControllerDelegate, STPURLCallbackListener, STPSafariViewControllerDismissalDelegate>
+@interface STPRedirectContext () <SFSafariViewControllerDelegate, STPURLCallbackListener, STPSafariViewControllerDismissalDelegate, UIViewControllerTransitioningDelegate>
 
 @property (nonatomic, strong, nullable) SFSafariViewController *safariVC;
-@property (nonatomic, strong, nonnull) STPSafariViewControllerTransitioningDelegate *safariVCTransitioningDelegate;
 @property (nonatomic, assign, readwrite) STPRedirectContextState state;
 /// If we're on iOS 11+ and in the SafariVC flow, this tracks the latest URL loaded/redirected to during the initial load
 @property (nonatomic, strong, readwrite, nullable) NSURL *lastKnownSafariVCURL;
@@ -147,7 +120,6 @@ typedef void (^STPBoolCompletionBlock)(BOOL success);
 
         _subscribedToURLNotifications = NO;
         _subscribedToForegroundNotifications = NO;
-        _safariVCTransitioningDelegate = [[STPSafariViewControllerTransitioningDelegate alloc] initWithDelegate:self];
     }
     return self;
 }
@@ -222,7 +194,7 @@ typedef void (^STPBoolCompletionBlock)(BOOL success);
         self.lastKnownSafariVCURL = self.redirectURL;
         self.safariVC = [[SFSafariViewController alloc] initWithURL:self.lastKnownSafariVCURL];
         self.safariVC.delegate = self;
-        self.safariVC.transitioningDelegate = self.safariVCTransitioningDelegate;
+        self.safariVC.transitioningDelegate = self;
         self.safariVC.modalPresentationStyle = UIModalPresentationCustom;
         [presentingViewController presentViewController:self.safariVC
                                                animated:YES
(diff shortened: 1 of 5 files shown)


Stripe/STPRedirectContext.m
```diff
@@ -196,7 +249,7 @@ - (void)cancel {
 
 - (void)safariViewControllerDidFinish:(__unused SFSafariViewController *)controller {
     stpDispatchToMainThreadIfNecessary(^{
-        [self handleRedirectCompletionWithError:nil
+        [self handleRedirectCompletionWithError:[NSError stp_cardRequiresAdditionalVerificationToProceedError]
```
yuki-stripe: Doesn't this also get called in cases where this error doesn't apply, like if you successfully complete authentication and close the webview?  

AFAIK we can't know whether authentication failed or succeeded here without fetching the PaymentIntent and inspecting its status (this is what `STPPaymentHandler` does)

Stripe/STPRedirectContext.m
```diff
@@ -25,9 +25,59 @@
...
+    self = [super init];
+    if (self) {
+        _dismissalDelegate = delegate;
+    }
+    return self;
+}
+
+- (nullable UIPresentationController *)presentationControllerForPresentedViewController:(UIViewController *)presented
+                                                               presentingViewController:(nullable UIViewController *)presenting
+                                                                   sourceViewController:(UIViewController *)source {
```
yuki-stripe: nit: I think our convention is to do:
```suggestion
                                                                   sourceViewController:(__unused UIViewController *)source {
```

and delete the pragma on the next line

Stripe/STPRedirectContext.m
```diff
@@ -25,9 +25,59 @@
...
+@implementation STPSafariViewControllerPresentationController
+- (void)dismissalTransitionDidEnd:(BOOL)completed {
+    if ([self.presentedViewController isKindOfClass:[SFSafariViewController class]]) {
+        [self.dismissalDelegate safariViewControllerDidCompleteDismissal:(SFSafariViewController *)self.presentedViewController];
+    }
+    return [super dismissalTransitionDidEnd:completed];
+}
+@end
+
+@interface STPSafariViewControllerTransitioningDelegate : NSObject <UIViewControllerTransitioningDelegate>
```
yuki-stripe: Could we make `STPRedirectContext` conform to `UIViewControllerTransitioningDelegate` and remove this middleman?

Stripe/STPRedirectContext.m
```diff
@@ -235,6 +288,14 @@ - (void)safariViewController:(__unused SFSafariViewController *)controller initi
     });
 }
 
+#pragma mark - STPSafariViewControllerDismissalDelegate -
+
+- (void)safariViewControllerDidCompleteDismissal:(SFSafariViewController *)controller {
+    #pragma unused (controller)
```
yuki-stripe: same thing here, I think we should use `__unused`

Tests/Tests/STPRedirectContextTest.m
```diff
@@ -224,47 +224,6 @@ - (void)testSafariViewControllerRedirectFlow_foregroundNotification {
...
-    XCTestExpectation *exp = [self expectationWithDescription:@"completion"];
-    STPRedirectContext *context = [[STPRedirectContext alloc] initWithSource:source completion:^(NSString *sourceID, NSString *clientSecret, NSError *error) {
-        XCTAssertEqualObjects(sourceID, source.stripeID);
-        XCTAssertEqualObjects(clientSecret, source.clientSecret);
-        XCTAssertNil(error);
-        [exp fulfill];
-    }];
-    XCTAssertEqualObjects(source.redirect.returnURL, context.returnURL);
-    id sut = OCMPartialMock(context);
-
```
yuki-stripe: I'm guessing the issue with these tests is that the completion block isn't being called, because now that will only happen after `safariViewControllerDidCompleteDismissal:` is called?  

To keep these tests, could we try adding something like:

```
OCMStub([sut handleRedirectCompletionWithError:[OCMArg any] shouldDismissViewController:YES]).andForwardToRealObject().andDo(^(__unused NSInvocation *invocation) {
  [sut safariViewControllerDidCompleteDismissal:nil];
}
```

Stripe/STPRedirectContext.m
joshavant: At least in my own testing, all successful and unsuccessful flows ended with a call to `handleURLCallback`, which explicitly triggered the dismissal of the VC. I assumed `safariViewControllerDidFinish:` was implemented only for the use case where the user taps the Done button, mid-flow.

Are you saying we're unable to lean on all web flows to indicate when the VC should be dismissed?

yuki-stripe: Ah, I'm thinking of the case where the user taps the Done button after finishing authentication - my understanding is, at least in testmode, the user ends up here: 
image and they can hit Done if they want instead of the '<- Return to Merchant' button.

If all that's true, I think we just need to continue passing `nil` as the error here (and remove the related `stp_cardRequiresAdditionalVerificationToProceedError` method).

Tests/Tests/STPRedirectContextTest.m
joshavant: I fixed all the tests using your suggested approach.

Stripe/STPRedirectContext.m
joshavant: Ah, ok. I removed the error and resumed passing `nil` here.

PR #1254, opened 2019-07-26
[17.0.0, breaking] Update API version from 2015-10-12 to 2019-05-16

yuki-stripe:
## Summary

Relevant changes since then:

> The legal_entity property on the Account API resource has been replaced with individual, company, and business_type.

> - Some PaymentIntent statuses have been renamed
>     - `requires_source` is now `requires_payment_method`
>     - `requires_source_action` is now `requires_action`
>     - All other statuses are unchanged

> For a Source’s card[three_d_secure] property, adds recommended as a possible value. Previously, the only valid values were not supported, optional, and required.

> Adds not_required as a possible redirect[status] value on the Source object. Previously, optional redirects were marked as succeeded.

(Full changelog: https://paper.dropbox.com/doc/Update-iOS-API-version-to-2019-05-16--Ahqa8fWMvo2AGLYlL1kISzoTAg-hFrNrJFf1f4dL3ZlT62YK)
## Motivation

- https://jira.corp.stripe.com/browse/IOS-1116
- https://jira.corp.stripe.com/browse/IOS-686

## Testing
Re-ran all the functional tests
* except `STPPinManagementFunctionalTest`, which is in beta (so presumably has the same behavior for all API versions).  I couldn't re-record it anyways, I think it needs to be reworked to include manual steps like {Payment, Setup}IntentFunctionalTest.


Stripe/PublicHeaders/STPSourceCardDetails.h
```diff
@@ -25,7 +27,7 @@ typedef NS_ENUM(NSInteger, STPSourceCard3DSecureStatus) {
     STPSourceCard3DSecureStatusRequired,
 
     /**
-     3D Secure is optional. It is not required for successful charging,
+     3D Secure is optional. It is not required nor recommended for successful charging,
```
csabol-stripe: I think this should be "or" but I'm not positive

Stripe/STPConnectAccountParams.m
```diff
@@ -8,31 +8,54 @@
...
 
-#import "STPLegalEntityParams.h"
-
 NS_ASSUME_NONNULL_BEGIN
 
 @implementation STPConnectAccountParams
 
 @synthesize additionalAPIParameters;
 
 - (instancetype)initWithTosShownAndAccepted:(BOOL)wasAccepted
```
csabol-stripe: since we're changing these signatures anyway, any reason to even have the `wasAccepted` argument?

MIGRATING.md
```diff
@@ -1,5 +1,8 @@
 ## Migration Guides
 
+### Migrating from versions < 17.0.0
```
csabol-stripe: I'm kind of inclined not to have another major version bump so soon...Maybe `16.1.0`? This should only be a breaking change for Connect users right? Thoughts @yuki-stripe @davidme-stripe

davidme-stripe: I'm not familiar with how versioning has worked here in the past, but moving from the 2015 API sounds like an appropriate reason to increment the major version number.

yuki-stripe: Definitely wary of fatiguing our users...there are the (minor) enum changes that Source users should update with, too.  Outside of that, it *should* be unnoticed. 

Either way, I'd like to ship this only after some confidence that 16.* won't have more bugfixes.

csabol-stripe: 👍 Let's hold this patch until 16.0.* has settled and then go from there

Stripe/PublicHeaders/STPConnectAccountParams.h
```diff
@@ -26,16 +36,26 @@ NS_ASSUME_NONNULL_BEGIN
...
+@property (nonatomic, readonly) STPConnectAccountBusinessType businessType;
 
- At least one field in the legalEntity must have a value, otherwise the create token
- call will fail.
+/**
+ Information about the individual represented by the account.
+ 
+ @see https://stripe.com/docs/api/tokens/create_account#create_account_token-account-individual
+ */
+@property (nonatomic, nullable, readonly) NSDictionary *individual;
```
csabol-stripe: why are these dictionaries instead of objects?

yuki-stripe: Only because it's a lot of parameters and Android's API looks like this.

csabol-stripe: hmm, I think we should type these. https://github.com/stripe/stripe-ios/pull/1178 has this but I haven't reviewed it extensively

Stripe/PublicHeaders/STPSourceCardDetails.h
yuki-stripe: I'm not sure, I copied it from https://stripe.com/docs/sources/three-d-secure#check-requirement so gonna go with their grammar.

Stripe/STPConnectAccountParams.m
yuki-stripe: I think we need it to differentiate between if the customer accepted the ToS or if the user didn't show a ToS (ie YES vs nil).

Stripe/PublicHeaders/STPConnectAccountAddress.h
```diff
@@ -0,0 +1,69 @@
...
+#import <Foundation/Foundation.h>
+
+#import "STPFormEncodable.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ An address to use with `STPConnectAccountParams`.
+ */
+@interface STPConnectAccountAddress : NSObject <STPFormEncodable>
```
csabol-stripe: how many different address objects can we fit in one SDK?? 🙃

Stripe/STPConnectAccountCompanyParams.m
```diff
@@ -0,0 +1,62 @@
...
+             NSStringFromSelector(@selector(kanjiName)): @"name_kanji",
+             NSStringFromSelector(@selector(ownersProvided)): @"owners_provided",
+             NSStringFromSelector(@selector(phone)): @"phone",
+             NSStringFromSelector(@selector(taxID)): @"tax_id",
+             NSStringFromSelector(@selector(taxIDRegistrar)): @"tax_id_registrar",
+             NSStringFromSelector(@selector(vatID)): @"vat_id",
+             };
+}
+
++ (nullable NSString *)rootObjectName {
```
csabol-stripe: man we really need to autogen these

Tests/Tests/STPConnectAccountAddressTest.m
```diff
@@ -0,0 +1,37 @@
...
+//  STPConnectAccountAddressTest.m
+//  StripeiOS Tests
+//
+//  Created by Yuki Tokuhiro on 8/2/19.
+//  Copyright © 2019 Stripe, Inc. All rights reserved.
+//
+
+#import <XCTest/XCTest.h>
+
+@interface STPConnectAccountAddressTest : XCTestCase
```
csabol-stripe: is this incomplete or should we remove?

Stripe/PublicHeaders/STPConnectAccountAddress.h
yuki-stripe: 😓 wish I could nest this in the `STPConnectAccount` namespace

Tests/Tests/STPConnectAccountAddressTest.m
yuki-stripe: Whoops, incomplete.

PR #1264, opened 2019-07-30
ApplePay support in STPPaymentHandler, STPPaymentContext

yuki-stripe:
## Summary
STPPaymentHandler can't present on top of the Apple Pay sheet.  This PR adds a `authenticationWillPresent:` delegate method to `STPAuthenticationContext` to let the user dismiss the Apple Pay sheet.  

I also reworked users of `STPPaymentHandler` to handle this:
* `STPPaymentContext` now conforms to `STPAuthenticationContext`, and users should pass it to `STPPaymentHandler`
* Standard Integration updated to do ^
* ApplePayExampleViewController now uses `STPPaymentHandler`

## Motivation
Make Apple Pay work with STPPaymentHandler

## Testing
Manually tested the following combinations:

STPPaymentContext, Custom Integration
x
Apple Pay, manual card entry
x 
Test cards 3220, 3063, 4242 (3DS2, 3DS1, regular)

### DIFF
--- a/Stripe/PublicHeaders/STPAuthenticationContext.h
+++ b/Stripe/PublicHeaders/STPAuthenticationContext.h
@@ -30,9 +30,10 @@ NS_ASSUME_NONNULL_BEGIN
  This method is called before presenting a UIViewController for authentication.
 
  Implement this method if your customer is using Apple Pay.  For security, it's impossible to present UIViewControllers above the Apple Pay sheet.
- This method should dismiss the PKPaymentAuthorizationViewController and call `continueBlock` in the dismissal's completion block.
+ This method should dismiss the PKPaymentAuthorizationViewController and call `completion` in the dismissal's completion block.
  
- Note that `paymentAuthorizationViewControllerDidFinish` is not called after `PKPaymentAuthorizationViewController` is dismissed.
+ @note `STPPaymentHandler` will not proceed until `completion` is called.
+ @note `paymentAuthorizationViewControllerDidFinish` is not called after `PKPaymentAuthorizationViewController` is dismissed.
  */
 - (void)prepareAuthenticationContextForPresentation:(STPVoidBlock)completion;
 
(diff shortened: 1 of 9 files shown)


Stripe/PublicHeaders/STPAuthenticationContext.h
```diff
@@ -24,6 +25,17 @@ NS_ASSUME_NONNULL_BEGIN
...
+@optional
+/**
+ This method is called before presenting a UIViewController for authentication.
+
+ Implement this method if your customer is using Apple Pay.  For security, it's impossible to present UIViewControllers above the Apple Pay sheet.
+ This method should dismiss the PKPaymentAuthorizationViewController and call `continueBlock` in the dismissal's completion block.
+ 
+ Note that `paymentAuthorizationViewControllerDidFinish` is not called after `PKPaymentAuthorizationViewController` is dismissed.
+ */
+- (void)authenticationWillPresent:(STPVoidBlock)continueBlock;
```
csabol-stripe: hmm, if we want to enforce that the merchant continues the flow by calling the block, `willPresent` seems semantically wrong...What do you think about
`- (void)prepareAuthenticationContextForPresentation:(STPVoidBlock)completion`
or similar?
I think `willPresent` has a connotation that no matter what I do this is happening, but in this case nothing will happen until you call the block

yuki-stripe: Yeah, I think that works.  Definitely want to avoid that failure mode where the user doesn't call the block!

Stripe/PublicHeaders/STPAuthenticationContext.h
```diff
@@ -34,7 +34,7 @@ NS_ASSUME_NONNULL_BEGIN
```
csabol-stripe: call `completion`. I think we should call out that STPPaymentHandler will not proceed until completion is called

PR #1276, opened 2019-08-05
Add NS_SWIFT_NAME to STPPaymentHandler

yuki-stripe:
## Summary
Swift names for STPPaymentHandler interface

I almost don't want to do this b/c it will be a breaking change and I'll have to go and update all the sample swift code >_> . I don't know how much of an improvement I've made here, eager to hear feedback.  

## Motivation
https://jira.corp.stripe.com/browse/IOS-1267

### DIFF
--- a/Stripe/PublicHeaders/STPPaymentHandler.h
+++ b/Stripe/PublicHeaders/STPPaymentHandler.h
@@ -36,7 +36,7 @@ typedef NS_ENUM(NSInteger, STPPaymentHandlerActionStatus) {
      The action failed. See the error code for more details.
      */
     STPPaymentHandlerActionStatusFailed,
-};
+} NS_SWIFT_NAME(STPPaymentHandler.ActionStatus);
 
 /**
  The error domain for errors in `STPPaymentHandler`.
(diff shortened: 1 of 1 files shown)


Stripe/PublicHeaders/STPPaymentHandler.h
```diff
@@ -87,7 +87,7 @@ typedef NS_ENUM(NSInteger, STPPaymentHandlerErrorCode) {
      If you're using Apple Pay, you must implement `STPAuthenticationContext prepareAuthenticationContextForPresentation:`
      */
     STPPaymentHandlerRequiresAuthenticationContextErrorCode,
-};
+} NS_SWIFT_NAME(STPPaymentHandler.ErrorCode);
```
davidme-stripe: The Swift bridge is parsing these out as `STPPaymentHandler.ErrorCode.requiresPaymentMethodErrorCode`, so you might want to add something like `STPPaymentHandlerRequiresPaymentMethodErrorCode NS_SWIFT_NAME(requiresPaymentMethod),` to the values.

davidme-stripe:
Might want to pair this with #1254 to group the breaking changes together.

PR #1303, opened 2019-08-16
Fix STPThreeDSUICustomization not properly initializing defaults or t…

yuki-stripe:
…he STDS object it's wrapping.

## Summary
`STPThreeDSUICustomization` didn't initialize its properties to default values, and it didn't set up the `STDS` object it wraps with those default values.  

## Motivation
😞 Sample code doesn't actually work, old behavior is not ergonomic.  

## Testing
Added tests for the missing behavior and tried sample code in Standard Integration app.

### DIFF
--- a/Stripe/Payments/STPThreeDSUICustomization.m
+++ b/Stripe/Payments/STPThreeDSUICustomization.m
@@ -25,15 +25,39 @@
 - (instancetype)init {
     self = [super init];
     if (self) {
-        _uiCustomization = [STDSUICustomization new];
+        // Initialize defaults for all properties
+        _footerCustomization = [STPThreeDSFooterCustomization defaultSettings];
+        _labelCustomization = [STPThreeDSLabelCustomization defaultSettings];
+        _navigationBarCustomization = [STPThreeDSNavigationBarCustomization defaultSettings];
+        _selectionCustomization = [STPThreeDSSelectionCustomization defaultSettings];
+        _textFieldCustomization = [STPThreeDSTextFieldCustomization defaultSettings];
+        
+        STPThreeDSButtonCustomization *nextButton = [STPThreeDSButtonCustomization defaultSettingsForButtonType:STPThreeDSCustomizationButtonTypeNext];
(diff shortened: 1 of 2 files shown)


Stripe/Payments/STPThreeDSUICustomization.m
```diff
@@ -25,15 +25,39 @@ + (instancetype)defaultSettings {
...
+        [_uiCustomization setButtonCustomization:cancelButton.buttonCustomization forType:STDSUICustomizationButtonTypeCancel];
+        [_uiCustomization setButtonCustomization:resendButton.buttonCustomization forType:STDSUICustomizationButtonTypeResend];
+        [_uiCustomization setButtonCustomization:submitButton.buttonCustomization forType:STDSUICustomizationButtonTypeSubmit];
+        [_uiCustomization setButtonCustomization:continueButton.buttonCustomization forType:STDSUICustomizationButtonTypeContinue];
+        
+        _uiCustomization.footerCustomization = _footerCustomization.footerCustomization;
+        _uiCustomization.labelCustomization = _labelCustomization.labelCustomization;
+        _uiCustomization.navigationBarCustomization = _navigationBarCustomization.navigationBarCustomization;
+        _uiCustomization.selectionCustomization = _selectionCustomization.selectionCustomization;
+        _uiCustomization.textFieldCustomization = _textFieldCustomization.textFieldCustomization;
```
csabol-stripe: so the function on line 70 is what was messing me up. Mutating the returned `STPThreeDSButotnCustomization` wasn't syncing. Do the above `_uiCustomization setButtonCustomization` calls fix that because now both dictionaries reference the same STDSButtonCustomization instance?

yuki-stripe: Yep, exactly.  `_buttonCustomizationDictionary ` references `STPThreeDSUIButtonCustomization` instances which forward all their properties to the underlying `STDS` instance, which we are now properly setting on `_uiCustomization` like we should have been (like on line 67, which is why calling the setter synced things).

PR #1304, opened 2019-08-19
Add additionalSupportedApplePayNetworks. Replaces isJCBPaymentNetwork…

yuki-stripe:
…Supported

## Summary
We added support to enable JCB for Apple Pay here: https://github.com/stripe/stripe-ios/pull/1158
This PR replaces that with a generalized way to enable any additional Apple Pay networks.

## Motivation
https://jira.corp.stripe.com/browse/MOBILE-120

## Testing
Unit test.

### DIFF
--- a/Stripe/PublicHeaders/STPAPIClient.h
+++ b/Stripe/PublicHeaders/STPAPIClient.h
@@ -287,7 +287,15 @@ static NSString *const STPSDKVersion = @"16.0.6";
  
  The default value is NO.
  */
-@property (class, nonatomic, getter=isJCBPaymentNetworkSupported) BOOL JCBPaymentNetworkSupported;
+@property (class, nonatomic, getter=isJCBPaymentNetworkSupported) BOOL JCBPaymentNetworkSupported __attribute__((deprecated("Set additionalApplePayNetworks = @[PKPaymentNetworkJCB] instead")));
+
+/**
+ Set this to enable additional networks for Apple Pay if needed.
+ By default, the SDK supports Amex, MasterCard, Visa, and Discover.
+ 
+ For example, `Stripe.additionalSupportedApplePayNetworks = @[PKPaymentNetworkJCB];` enables JCB (note this requires onboarding from JCB and Stripe).
+ */
+@property (class, nonatomic, nonnull) NSArray<PKPaymentNetwork> *additionalSupportedApplePayNetworks;
 
 @end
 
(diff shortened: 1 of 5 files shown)


Stripe/PublicHeaders/STPAPIClient.h
```diff
@@ -287,7 +287,15 @@ static NSString *const STPSDKVersion = @"16.0.6";
...
-@property (class, nonatomic, getter=isJCBPaymentNetworkSupported) BOOL JCBPaymentNetworkSupported;
+@property (class, nonatomic, getter=isJCBPaymentNetworkSupported) BOOL JCBPaymentNetworkSupported __attribute__((deprecated("Set additionalApplePayNetworks = @[PKPaymentNetworkJCB] instead")));
+
+/**
+ Set this to enable additional networks for Apple Pay if needed.
+ By default, the SDK supports Amex, MasterCard, Visa, and Discover.
+ 
+ For example, `Stripe.additionalSupportedApplePayNetworks = @[PKPaymentNetworkJCB];` enables JCB (note this requires onboarding from JCB and Stripe).
+ */
+@property (class, nonatomic, nonnull) NSArray<PKPaymentNetwork> *additionalSupportedApplePayNetworks;
```
csabol-stripe: I'm worried that having `supportedPKPaymentNetworks ` and `additionalSupportedApplePayNetworks` will be confusing. What if we just make `supportedPKPaymentNetworks` writeable?

csabol-stripe: I'm actually not a huge fan of `supported` in this context. It seems to imply what Stripe supports, but we're allowing arbitrary editing?

yuki-stripe: Yeah I thought the same (even wrote out a `setSupportedPKPaymentNetworks`).  I decided against it b/c:
1. `supportedPKPaymentNetworks` isn't public. 
2. We don't want to let users remove the defaults (mastercard, etc.).  `setSupportedPKPaymentNetworks` implies arbitrary editing, but we just want to let them *add* to the defaults.  

Would rewording to `accept` be any better?  Maybe something like "The SDK accepts Amex, MasterCard, Visa, and Discover for Apple Pay.  Set this property to enable other card networks in addition to these." is clearer.

csabol-stripe: Makes sense. Maybe `enabled` instead of `supported` or `accept`? Defintely + on that docstring

PR #1307, opened 2019-08-20
Add support for iOS 13 Dark Mode

davidme-stripe:
## Summary
Adds full support for adaptable UIColors and Dark Mode in iOS 13. We've also updated the example apps to match.

## Motivation
We were doing some UIColor manipulation in STPTheme, which made the checkout experience look broken on apps compiled with the iOS 13 SDK.

## Testing
Tested with the various example apps on iOS 13.


Example/Standard Integration/BrowseProductsViewController.swift
```diff
@@ -77,13 +77,29 @@ class BrowseProductsViewController: UICollectionViewController {
...
         self.navigationItem.title = "Emoji Apparel"
         self.navigationController?.navigationBar.isTranslucent = false
+        #if canImport(CryptoKit)
+        if #available(iOS 13.0, *) {
+            self.navigationController?.view.backgroundColor = .systemBackground
+        } else {
+            self.navigationController?.view.backgroundColor = .white
+        }
+        #else
         self.navigationController?.view.backgroundColor = .white
```
yuki-stripe: oh maybe you could put these calls unconditionally before `#if canImport(CryptoKit)` and overwrite the property in the `#if` block, to avoid some duplication.

Don't think it matters much though, not a blocker.

davidme-stripe: I thought about that, but I didn't want the extra cycles of calling setBackgroundColor twice. Not sure if it really matters, the calls are probably all coalesced until the redraw.

PR #1312, opened 2019-08-22
Add support for iOS 13 Dark Mode to Custom Integration

davidme-stripe:
## Summary
The Custom Integration app (and STPPaymentCardTextField) should also support Dark Mode.

## Testing
On iOS 13 and in CI.

### DIFF
--- a/Stripe/PublicHeaders/STPPaymentCardTextField.h
+++ b/Stripe/PublicHeaders/STPPaymentCardTextField.h
@@ -36,7 +36,7 @@ IB_DESIGNABLE
 @property (nonatomic, copy, null_resettable) UIFont *font UI_APPEARANCE_SELECTOR;
 
 /**
- The text color to be used when entering valid text. Default is [UIColor blackColor]. 
+ The text color to be used when entering valid text. Default is [UIColor labelColor].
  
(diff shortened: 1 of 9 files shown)


Stripe/PublicHeaders/STPPaymentCardTextField.h
```diff
@@ -36,7 +36,7 @@ IB_DESIGNABLE
 @property (nonatomic, copy, null_resettable) UIFont *font UI_APPEARANCE_SELECTOR;
 
 /**
- The text color to be used when entering valid text. Default is [UIColor blackColor]. 
+ The text color to be used when entering valid text. Default is [UIColor labelColor].
```
yuki-stripe: possibly worth documenting both iOS 13+ and pre-iOS 13 defaults in here?

davidme-stripe: I did in my first draft, but I took it out for simplicity. I can't think of a situation where a developer would be confused by this, but I can put it back in if you think it would help.

yuki-stripe: 👍 eh I think it's fine to ship as-is

PR #1315, opened 2019-08-23
Fix typo'd Xcode version in Readme

jil-stripe:
## Summary
Our minimum supported Xcode version is 10.1; our Readme was out of date.

## Motivation
Noticed and confirmed typo during Slack discussion.

**NB:** Should it say "requires at least Xcode 10.1 to build from source"?

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ If you're reading this on GitHub.com, please make sure you are looking at the [t
 
 ## Requirements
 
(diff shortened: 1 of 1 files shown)


davidme-stripe:
> **NB:** Should it say "requires at least Xcode 10.1 to build from source"?

"At least" seems good. We may also leave out "to build from source" — I'm not sure if the older Xcode toolchain can process the embedded bitcode in our distributed library. `ld` would definitely fail, and some of the internal .app packaging and distribution bits may also have trouble.

PR #1327, opened 2019-08-29
Add an option to STPPaymentConfiguration to filter the country list

### LINKED ISSUE #1314: Limit country list in STPShippingAddressViewController
Hello,

I have 2 questions:

1. The STPShippingAddressViewController country field is "United States" by default. Is there a way to control this? For example, If I know what is the country of the user beforehand, can I prefill this with a different country?

2. The country field is showing plenty of countries. Can I limit this list to a specific few countries?

Thanks,
Oded

davidme-stripe:
## Summary
Add a new `availableCountries` property to STPPaymentConfiguration, allowing a developer to filter the list of countries.

This is nil, displaying all countries, by default. It can be set to a list of ISO country codes, which will filter our UI to only display those countries.
* If the user locale's country code is in that list, it will be selected by default. If not, the country field will be empty.
* If setAddress is called with a country code outside of this list, that country code will be added to the UI.

## Motivation
Fixes #1314 and many other out-of-band requests for this feature.

## Testing
Updated some related tests, though I'm going to write a specific UI test before shipping.

### DIFF
--- a/Stripe/PublicHeaders/STPPaymentConfiguration.h
+++ b/Stripe/PublicHeaders/STPPaymentConfiguration.h
@@ -78,6 +78,15 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, assign, readwrite) STPShippingType shippingType;
 
 /**
+ The set of countries supported when entering an address. This property accepts
+ a set of ISO 2-character country codes.
+
+ The default value is nil, which will display all known countries. Setting this
+ property will limit the available countries to your selected set.
+ */
+@property (nonatomic, copy, nullable, readwrite) NSSet<NSString *> *availableCountries;
+
+/**
  The name of your company, for displaying to the user during payment flows. For 
  example, when using Apple Pay, the payment sheet's final line item will read
  "PAY {companyName}". 
--- a/Stripe/PublicHeaders/STPPaymentConfiguration.h
+++ b/Stripe/PublicHeaders/STPPaymentConfiguration.h
@@ -81,10 +81,10 @@ NS_ASSUME_NONNULL_BEGIN
  The set of countries supported when entering an address. This property accepts
  a set of ISO 2-character country codes.
 
- The default value is nil, which will display all known countries. Setting this
- property will limit the available countries to your selected set.
+ The default value is all known countries. Setting this property will limit
+ the available countries to your selected set.
  */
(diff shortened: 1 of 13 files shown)


Stripe/STPAddressViewModel.h
```diff
@@ -25,10 +25,13 @@
 @property (nonatomic, readonly) NSArray<STPAddressFieldTableViewCell *> *addressCells;
 @property (nonatomic, weak) id<STPAddressViewModelDelegate>delegate;
 @property (nonatomic) STPAddress *address;
+@property (nonatomic, copy, readwrite) NSSet<NSString *> *availableCountries;
 @property (nonatomic, readonly) BOOL isValid;
 
 - (instancetype)initWithRequiredBillingFields:(STPBillingAddressFields)requiredBillingAddressFields;
 - (instancetype)initWithRequiredShippingFields:(NSSet<STPContactField> *)requiredShippingAddressFields;
+- (instancetype)initWithRequiredBillingFields:(STPBillingAddressFields)requiredBillingAddressFields availableCountries:(NSSet<NSString *> *)availableCountries;
```
csabol-stripe: can we add comments for the `availableCountries:nil` behavior here?

Stripe/PublicHeaders/STPPaymentConfiguration.h
```diff
@@ -77,6 +77,15 @@ NS_ASSUME_NONNULL_BEGIN
...
 @property (nonatomic, assign, readwrite) STPShippingType shippingType;
 
+/**
+ The set of countries supported when entering an address. This property accepts
+ a set of ISO 2-character country codes.
+
+ The default value is nil, which will display all known countries. Setting this
+ property will limit the available countries to your selected set.
+ */
+@property (nonatomic, copy, nullable, readwrite) NSSet<NSString *> *availableCountries;
```
csabol-stripe: what do you think about making this `null_resettable` and documenting that by default it's populated from `[NSLocale ISOCountryCodes]`?

davidme-stripe: It looked more like this originally, but I was worried about inadvertently copying availableCountries as we pass it down the chain. It feels ugly to have to check something like `if (newAvailableCountries == setOfAllCountries) { _availableCountries = nil }` in every setter. But agreed that it probably makes more sense from an API perspective.

PR #1338, opened 2019-08-30
Fix some CI issues

yuki-stripe:
## Summary
* Show undocumented log upon check_documentation failure
* Grep for warning not warning: because xcodebuild sometimes inserts spurious newlines

## Motivation
* It's annoying to run `check_documentation` locally, CI should just tell me why it failed
* `run_analyzer` is flaky b/c xcodebuild sometimes does this:
image

## Testing
Added code to trigger analyzer and doc warnings, checked these changes cause them to error usefully.  
Relying on CI to tell me I didn't introduce false positives

### DIFF
--- a/ci_scripts/run_analyzer.sh
+++ b/ci_scripts/run_analyzer.sh
@@ -33,7 +33,11 @@ fi
 
 # Search for warnings in log file
 echo "Searching for static analyzer warnings..."
-grep "warning:" "${log_file}" > "/dev/null"
+
+# Fun note:
+# xcodebuild outputs a line like "...SomeFile.m:36:1: warning: foo"
+# ...but sometimes, it inserts spurious newlines everywhere
+grep "warning" "${log_file}" > "/dev/null"
 
 if [[ "$?" != 1 ]]; then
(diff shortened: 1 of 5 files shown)


ci_scripts/run_analyzer.sh
```diff
@@ -33,7 +33,11 @@ fi
 
 # Search for warnings in log file
 echo "Searching for static analyzer warnings..."
-grep "warning:" "${log_file}" > "/dev/null"
+
+# Fun note:
+# xcodebuild outputs a line like "...SomeFile.m:36:1: warning: foo"
+# ...but sometimes, it inserts spurious newlines everywhere
```
danj-stripe: yuck.

Did you consider capturing the `xcpretty` output instead of `xcodebuild`? Seems like `The following commands produced analyzer issues:` could be a good alternative grep string, and might avoid the newline problem?

yuki-stripe: Yeah, I did see xcpretty still outputs that.  Perhaps we should log that or both, if the raw logs are too.  

I defaulted to optimizing for the smallest change here 🤷‍♂

Issue #1331, opened 2019-08-29
Can't set billingDetails with createPaymentMethodWithPayment


aurelhubert:
## Summary
I'm using the method `STPAPIClient.shared().createPaymentMethod(with: payment)` to create a payment with Apple Pay and I have set the billing details inside the payment object (in the field `billingContact`), but when the SDK receives the response from the API (in `STPAPIClient+ApplePay.m`), the `billingDetails` is set to nil and is `readOnly` (so it can't be updated after).

On Android, I do it with the next code for Google Pay:
```
val paymentData = PaymentData.getFromIntent(data!!)
val rawToken = paymentData?.paymentMethodToken!!.token
val stripeToken = Token.fromString(rawToken)

val paymentMethodCreateParams = PaymentMethodCreateParams.create(PaymentMethodCreateParams.Card.create(stripeToken.id), billingDetails)
val paymentIntentParams = ConfirmPaymentIntentParams.createWithPaymentMethodCreateParams(paymentMethodCreateParams, clientSecret, "...")
```

Is there currently a way to set it correctly?

## SDK version
16.0.7

davidme-stripe:
Thanks for filing, @aurelhubert! We're fixing this in #1331.

davidme-stripe:
Sorry for the auto-close. Try pointing your Podfile to our master branch, it should resolve this for you.

aurelhubert:
I've tested again with the code on the master branch and it doesn't work for now. I've maybe missed something.

The `billingContact` is correctly set on the `PKPaymentRequest` object, but when the transaction succeed, the `PKPayment` doesn't have the information.

This is the way I call Apple Pay:
```
let merchantIdentifier = StripeManager.appleMerchantID
let paymentRequest = Stripe.paymentRequest(withMerchantIdentifier: merchantIdentifier, country: "FR", currency: "EUR")
let totalAmount = NSDecimalNumber(value: cart.totalAmount)
        
// Configure the line items on the payment request
paymentRequest.paymentSummaryItems = [...]
paymentRequest.billingContact = self.billingContact

if Stripe.canSubmitPaymentRequest(paymentRequest) {
    if let paymentAuthorizationViewController = PKPaymentAuthorizationViewController(paymentRequest: paymentRequest) {
        paymentAuthorizationViewController.delegate = self

        print("paymentRequest: \(paymentRequest.billingContact?.name)") // => Not nil
[...]

PR #1343, opened 2019-08-30
Update Apple Pay to iOS 11 APIs

yuki-stripe:
## Summary
* Update ApplePayExampleViewController to use iOS 11+ APIs.  I also bumped the Custom Integration project to target iOS 11.  Which target to choose is open but for the purpose of this PR, iOS 11 or 12 doesn't make a difference.
* Update ApplePayExampleViewController automatic confirmation, this is the new recommended default.
* The new `didAuthorizePayment` API completion block has an errors param.  If you populate it, the apple sheet shows an error and doesn't dismiss, letting the user try again.  I added `pkPaymentErrorForStripeError` for this.  

TODO in another PR: Update STPPaymentContext's Apple Pay integration to the new APIs

## Motivation
Example code is probably more useful targeting iOS 11+.

## Testing
Tested ApplePayExampleViewController on my device w/ invalid zip test card:
image

"stolen" test card:
image

### DIFF
--- a/Example/Custom Integration.xcodeproj/project.pbxproj
+++ b/Example/Custom Integration.xcodeproj/project.pbxproj
@@ -443,7 +443,7 @@
 					"$(PROJECT_DIR)/**",
 				);
 				INFOPLIST_FILE = "$(SRCROOT)/Custom Integration/Info.plist";
-				IPHONEOS_DEPLOYMENT_TARGET = 9.0;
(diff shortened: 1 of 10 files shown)


davidme-stripe:
Great! I think another good TODO is being discussed in #1331: We should probably add a way to request the emailAddress and phoneNumber fields on `required(Billing|Shipping)AddressFields` and pass them along to the PaymentMethod.

PR #1363, opened 2019-09-09
Update examples

yuki-stripe:
## Summary
* Move STPPaymentContext to automatic confirmation - this is the default recommendation now.
* Update example APIs to match updated example backend.

## Motivation
Sync examples with what docs recommend https://github.com/stripe/example-ios-backend/pull/67

## Testing
All the Custom Integration Card example flows, Standard Integration.


davidme-stripe:
🎉woo, automatic confirmation

We should update [https://stripe.com/docs/mobile/ios/standard#paymentcontextdidcreatepaymentresultcompletion](the docs) to match.

Example/StandardIntegrationUITests/StandardIntegrationUITests.swift
```diff
@@ -17,7 +17,7 @@ class StandardIntegrationUITests: XCTestCase {
         // UI tests must launch the application that they test. Doing this in setup will make sure it happens for each test method.
         let app = XCUIApplication()
         let stripePublishableKey = "pk_test_6Q7qTzl8OkUj5K5ArgayVsFD00Sa5AHMj3"
-        let backendBaseURL = "https://stripe-mobile-test-backend.herokuapp.com/"
+        let backendBaseURL = "https://stripe-mobile-test-backend-17.herokuapp.com/"
```
davidme-stripe: Hah, forgot that versioning would be an issues on these tests. This seems to be an okay workaround, but if it gets too toilsome we should figure out something else.

PR #1375, opened 2019-09-13
Add BankSelectionViewController for FPX

davidme-stripe:
## Summary
* Adds BankSelectionViewController, which accepts a bank type and returns a PaymentMethodParams.
* PaymentMethodParams is now an STPPaymentOption, and can be treated as sort of a PaymentMethod promise, allowing a user to send it directly to the Stripe API when confirming their PaymentIntent. This saves a roundtrip to the server to create separate FPX PaymentMethod, which isn't necessary as the only metadata will generally be the name of the bank.
* The FPX example in Custom Integration has been modified to show how we expect developers to integrate this.

* I've also added a new `authenticationContextDidPresentSafariViewController` delegate method to give the developer a chance to configure their view underneath the SafariViewController. This allows a developer to remove the FPX bank selector from the view controller stack while the user is handling the authentication request, providing a better experience when the payment completes. You can see an example of how this is used in `FPXExampleViewController` — I'm not super happy with it, but it seems like the least gross option of the options I considered. Feedback is welcome!

## Testing
Tested using the Custom Integration app.


Example/Custom Integration/FPXExampleViewController.m
```diff
@@ -86,9 +86,14 @@ - (void)updateUIForPaymentInProgress:(BOOL)paymentInProgress {
...
-#pragma clang diagnostic ignored "-Wdeprecated"
-- (void)pay {
+- (void)selectBank {
+    STPBankSelectionViewController *vc = [[STPBankSelectionViewController alloc] initWithBankType:STPBankTypeFPX];
+    vc.delegate = self;
+    
+    [self.navigationController pushViewController:vc animated:YES];
+}
+
+- (void)payWithBankAccount:(STPFPXBankBrand)bank {
```
yuki-stripe: Refactor to take a `STPPaymentMethodParams` instead?

Stripe/PublicHeaders/STPFPXBankBrand.h
```diff
@@ -8,6 +8,21 @@
 
 #import <Foundation/Foundation.h>
 
+/**
+ The supported bank types.
+ */
+typedef NS_ENUM(NSInteger, STPBankType) {
```
csabol-stripe: it's not really clear what a "bank type" is...Does Stripe use this nomenclature anywhere else?

Stripe/STPBankSelectionTableViewCell.m
```diff
@@ -0,0 +1,115 @@
...
+
+    // Activity indicator
+    self.activityIndicator.center = CGPointMake(CGRectGetWidth(self.bounds) - padding - CGRectGetMidX(self.activityIndicator.bounds), midY);
+
+    // Title label
+    CGRect labelFrame = self.bounds;
+    // not every icon is `iconWidth` wide, but give them all the same amount of space:
+    labelFrame.origin.x = padding + iconWidth + padding;
+    labelFrame.size.width = CGRectGetMinX(self.activityIndicator.frame) - padding - labelFrame.origin.x;
+    self.titleLabel.frame = labelFrame;
```
csabol-stripe: why not autolayout?

Stripe/STPBankSelectionTableViewCell.m
```diff
@@ -0,0 +1,115 @@
...
+    self.titleLabel.font = theme.font;
+    self.titleLabel.text = STPStringFromFPXBankBrand(self.bank);
+    self.titleLabel.textColor = [self primaryColorForPaymentOptionWithSelected:self.selected enabled:enabled];
+
+    // Loading indicator
+    self.activityIndicator.tintColor = theme.accentColor;
+    if (selected) {
+        [self.activityIndicator startAnimating];
+    } else {
+        [self.activityIndicator stopAnimating];
```
csabol-stripe: maybe a `bringSubviewToFront` as well in case the hierarchy here changes?

csabol-stripe: er, on the start antimating I mean

Stripe/STPBankSelectionViewController.m
```diff
@@ -0,0 +1,136 @@
...
+    [self.tableView registerClass:[STPBankSelectionTableViewCell class] forCellReuseIdentifier:STPBankSelectionCellReuseIdentifier];
+
+    self.tableView.dataSource = self;
+    self.tableView.delegate = self;
+}
+
+- (void)updateAppearance {
+    [super updateAppearance];
+    
+    self.tableView.allowsSelection = YES;
```
csabol-stripe: is this needed?

Stripe/STPBankSelectionViewController.m
```diff
@@ -0,0 +1,136 @@
...
+}
+
+#pragma mark - UITableView
+
+- (NSInteger)numberOfSectionsInTableView:(__unused UITableView *)tableView {
+    return 1;
+}
+
+- (NSInteger)tableView:(__unused UITableView *)tableView numberOfRowsInSection:(__unused NSInteger)section {
+    return STPFPXBankBrandUnknown;
```
csabol-stripe: I'm not sure how consistent we are with always making Unknown the last item in an enumeration (hopefully we are) but 👍

Stripe/STPBankSelectionViewController.m
```diff
@@ -0,0 +1,136 @@
...
+    [cell stp_setFakeSeparatorLeftInset:15.0f];
+}
+
+- (CGFloat)tableView:(__unused UITableView *)tableView heightForFooterInSection:(__unused NSInteger)section {
+    return 27.0f;
+}
+
+- (void)tableView:(UITableView *)tableView didSelectRowAtIndexPath:(NSIndexPath *)indexPath {
+    if (self.loading) {
+        return; // Don't allow user interaction if we're currently setting up a payment method
```
csabol-stripe: I think this might be better implemented using `- (BOOL)tableView:(UITableView *)tableView 
shouldHighlightRowAtIndexPath:(NSIndexPath *)indexPath`, that way you can't select at all -- or did you explicitly want to make the cells still interactable?

Stripe/STPPaymentMethodParams.m
```diff
@@ -107,4 +108,53 @@ + (nonnull NSDictionary *)propertyNamesToFormFieldNamesMapping {
...
+        case STPPaymentMethodTypeCard:
+            if (self.card != nil) {
+                STPCardBrand brand = [STPCardValidator brandForNumber:self.card.number];
+                NSString *brandString = STPStringFromCardBrand(brand);
+                return [NSString stringWithFormat:@"%@ %@", brandString, self.card.last4];
+            } else {
+                return STPStringFromCardBrand(STPCardBrandUnknown);
+            }
+        case STPPaymentMethodTypeiDEAL:
+            return @"iDEAL";
```
csabol-stripe: hmm, we localized apm names for sources but not payment methods (see STPSource.m). Probably this and FPX should be wrapped in `STPLocalizedString`

csabol-stripe: looks like we're also not localizing in STPPaymentMethod.m

Example/Custom Integration/FPXExampleViewController.m
```diff
@@ -123,6 +128,27 @@ - (void)pay {
...
+
+#pragma mark - STPAuthenticationContext
+
+- (UIViewController *)authenticationPresentingViewController {
+    return self.navigationController.topViewController;
+}
+
+- (void)authenticationContextDidPresentSafariViewController:(SFSafariViewController *)viewController {
+    // If we're launching directly into the SFSafariViewController for authentication from the bank selector,
+    // we'll want to remove the bank selector from the VC stack and pop back directly to our main controller on dismiss.
```
yuki-stripe: Interesting problem!  Several thoughts

If we dropped it, the user would dismiss after `didCreatePaymentMethodParams`.  They could then immediately call `confirm...` ...that doesn't seem so bad, in my head - I might be missing something

---

What if the network request takes a while?  The customer would just be looking at the bank selector, right?  We should probably block them from selecting another bank.  

-----
If we keep it..

Any harm to making it `...didPresentViewController`?  Might come in handy in the future, for someone.

An alternative might be `authenticationContextWillFinish:withStatus:` or something, which would e.g. allow you to dismiss safariVC only if it was successful.  I'm thinking if a customer cancels/closes the webview, they want to go back to the previous screen (so the bank selector).

Stripe/PublicHeaders/STPPaymentMethodParams.h
```diff
@@ -23,7 +23,7 @@ NS_ASSUME_NONNULL_BEGIN
  
  @see https://stripe.com/docs/api/payment_methods/create
  */
-@interface STPPaymentMethodParams : NSObject <STPFormEncodable>
+@interface STPPaymentMethodParams : NSObject <STPFormEncodable, STPPaymentOption>
```
yuki-stripe: is this conformance needed for this PR?

Stripe/STPBankSelectionViewController.h
```diff
@@ -0,0 +1,69 @@
...
+#import "STPPaymentConfiguration.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@protocol STPBankSelectionViewControllerDelegate;
+@class STPPaymentMethodParams;
+
+/** This view controller displays a list of banks of the specified type, allowing the user to select one to pay from.
+    Once a bank is selected, it will return a PaymentMethodParams object, which you can use to confirm a PaymentIntent or
+    share the details with your backend.
```
yuki-stripe: We're not directly recommend it, but slight FUD around people passing up a paymentMethodParams object.  wdyt about leaving it at "...confirm a PaymentIntent".

Stripe/STPBankSelectionViewController.h
```diff
@@ -0,0 +1,69 @@
...
+/**
+An `STPBankSelectionViewControllerDelegate` is notified when a user selects a bank.
+*/
+@protocol STPBankSelectionViewControllerDelegate <NSObject>
+
+/**
+This is called when the user selects a bank.
+
+You can use the returned PaymentMethodParams to confirm a PaymentIntent, or inspect
+ it to pass along the details to your server for server-side confirmation.
```
yuki-stripe: same here.

Stripe/PublicHeaders/STPFPXBankBrand.h
davidme-stripe: It's a UI level concern of which bank list to show, but it currently only supports FPX. I left it in because I didn't want to change the initialization method declaration of BankSelectionViewController in the future. Stripe Elements has a specific "FPX" element instead of a generic bank selector, so as an alternative we could rename `BankSelectionViewController` to `FPXBankSelectionViewController` and have different bank selectors as we add support for other APMs.

Stripe/STPBankSelectionTableViewCell.m
davidme-stripe: This is a copypasta from our existing TableViewControllers. I don't think we use autolayout anywhere in the SDK.

Stripe/STPBankSelectionTableViewCell.m
davidme-stripe: I don't think the hierarchy will change in these cells: The views are all added at initialization and never added/removed again. 
Simulator Screen Shot - iPhone Xʀ - 2019-09-16 at 15 49 54

Stripe/STPBankSelectionViewController.m
davidme-stripe: No, `allowsSelection` is the default, and this method was copied from another VC that doesn't seem to need it either. I'll clean that one up too.

Stripe/STPBankSelectionViewController.m
davidme-stripe: We're pretty consistent.

Stripe/STPBankSelectionViewController.m
davidme-stripe: The cells are enabled/disabled individually by the `userInteractionDisabled` flag set by `configureWithBank:theme:selected:enabled:`, which is copying the style of our other ViewControllers. They could probably use some rethinking — there's a lot of overreliance on [tableView reloadData] in our VCs.

Stripe/STPPaymentMethodParams.m
davidme-stripe: Yeah, it looks like localization was largely dropped for PaymentMethod-related stuff. We need to audit more of it.

Stripe/STPBankSelectionTableViewCell.h
```diff
@@ -0,0 +1,24 @@
...
+//  Created by David Estes on 8/9/19.
+//  Copyright © 2019 Stripe, Inc. All rights reserved.
+//
+
+#import <UIKit/UIKit.h>
+#import "STPFPXBankBrand.h"
+
+@class STPTheme;
+
+@protocol STPPaymentOption;
```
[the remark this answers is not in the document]
davidme-stripe: nope

Stripe/STPBankSelectionTableViewCell.m
```diff
@@ -0,0 +1,115 @@
...
+
+#import "STPBankSelectionTableViewCell.h"
+
+#import "STPApplePayPaymentOption.h"
+#import "STPCard.h"
+#import "STPImageLibrary+Private.h"
+#import "STPLocalizationUtils.h"
+#import "STPPaymentMethod.h"
+#import "STPPaymentMethodCard.h"
+#import "STPPaymentOption.h"
```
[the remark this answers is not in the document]
davidme-stripe: not anymore

Example/Custom Integration/FPXExampleViewController.m
davidme-stripe: Hah, oops

Example/Custom Integration/FPXExampleViewController.m
davidme-stripe: > If we dropped it, the user would dismiss after didCreatePaymentMethodParams. They could then immediately call confirm... ...that doesn't seem so bad, in my head - I might be missing something
If they did that, they'd see the dismissal of the bank selector followed by an immediate animate-in of the SafariViewController. It doesn't look great.

> What if the network request takes a while? The customer would just be looking at the bank selector, right? We should probably block them from selecting another bank.
Yes, but they'll at least have a spinner and a disabled bank selector. It's on the app to provide more UI if they'd like.

> An alternative might be authenticationContextWillFinish:withStatus: or something, which would e.g. allow you to dismiss safariVC only if it was successful. I'm thinking if a customer cancels/closes the webview, they want to go back to the previous screen (so the bank selector).
Good idea, I'll look into that.

Stripe/STPBankSelectionViewController.h
davidme-stripe: I'd like the dev to still access the bank information somehow, and we're not encouraging them to send up the entire PaymentMethodParams object.

Maybe: "Once a bank is selected, it will return a PaymentMethodParams object, which you can use to confirm a PaymentIntent or inspect to obtain details about the selected bank."

Stripe/PublicHeaders/STPPaymentMethodParams.h
davidme-stripe: No, this was mistakenly left in from the Standard Integration support (which will be in another PR).

Stripe/STPBankSelectionViewController.m
```diff
@@ -0,0 +1,136 @@
...
+- (NSInteger)tableView:(__unused UITableView *)tableView numberOfRowsInSection:(__unused NSInteger)section {
+    return STPFPXBankBrandUnknown;
+}
+
+- (UITableViewCell *)tableView:(UITableView *)tableView
+         cellForRowAtIndexPath:(NSIndexPath *)indexPath {
+    STPBankSelectionTableViewCell *cell = [tableView dequeueReusableCellWithIdentifier:STPBankSelectionCellReuseIdentifier forIndexPath:indexPath];
+    STPFPXBankBrand bankBrand = indexPath.row;
+    BOOL selected = self.selectedBank == bankBrand;
+    [cell configureWithBank:bankBrand theme:self.theme selected:selected enabled:!self.loading];
```
yuki-stripe: Huh, every cell is configured with the same loading..I'd guess once `loading == YES` and you scroll the tableview, the cells that appear all have spinners 😃

Stripe/PublicHeaders/STPFPXBankBrand.h
csabol-stripe: What if we made this something like `STPBankSelectionContextType` and `STPBankSelectionContextTypeFPXPayment` or something? I can definitely see BankSerlectionViewController being reusable so we probably shouldn't create a bunch of subclasses

Stripe/STPBankSelectionTableViewCell.m
csabol-stripe: 😢

Stripe/STPBankSelectionTableViewCell.m
csabol-stripe: yeah I just mean more as a protection for in case someone comes in and changes the initializer without realizing. Not a blocker though

Stripe/STPBankSelectionViewController.m
csabol-stripe: sweet!

Stripe/STPBankSelectionViewController.m
csabol-stripe: yikes okay, mind filing a JIRA for us to take a look at this?

Stripe/STPPaymentMethodParams.m
csabol-stripe: presumably these are all the same as the localized values in `STPSource` so we should be able to just add the `STPLocalizedString` and then get translations (if there are any) for FPX. Can do as a followup

Example/Custom Integration/FPXExampleViewController.m
yuki-stripe: > What if the network request takes a while? The customer would just be looking at the bank selector, right? We should probably block them from selecting another bank.
Yes, but they'll at least have a spinner and a disabled bank selector. It's on the app to provide more UI if they'd like.

Ooh okay, didn't realize the BankSelectorVC already disabled UI after selection.  

Everything here sounds good to me.  I bet this AuthenticationContext API will be useful in the future for some other reason.

Stripe/STPBankSelectionViewController.h
yuki-stripe: +1 sounds good to me!

Stripe/STPBankSelectionViewController.m
davidme-stripe: I might try to take care of it here, this PR isn't super urgent.

Stripe/PublicHeaders/STPFPXBankBrand.h
yuki-stripe: mm yeah, maybe `STPBankSelectionType`?  

I would also be down to reduce the scope of this PR and only ship what we know we need now (so, just `FPXBankSelectionViewController`).  If the future is unexpected, it's probably easy/backwards-compatible to deprecate `FPXBankSelectionVC` vs. reworking a more generalized thing.

Stripe/STPBankSelectionViewController.m
davidme-stripe: It's enabled (vs greyed out and disabled) based on `!self.loading`, it's selected (which shows the spinner) based on `selected`, which is `self.selectedBank == bankBrand`.

Stripe/PublicHeaders/STPFPXBankBrand.h
davidme-stripe: I like that, but maybe `STPBankSelectionMethod` is more clear?

I think we're planning to ship similar methods in the coming weeks/months, so I'd like to avoid adding some API that we need to immediately deprecate.

Stripe/STPBankSelectionViewController.m
yuki-stripe: Ah my mistake

Example/Custom Integration/FPXExampleViewController.m
```diff
@@ -86,9 +86,14 @@ - (void)updateUIForPaymentInProgress:(BOOL)paymentInProgress {
...
-#pragma clang diagnostic ignored "-Wdeprecated"
-- (void)pay {
+- (void)selectBank {
+    STPBankSelectionViewController *vc = [[STPBankSelectionViewController alloc] initWithBankMethod:STPBankSelectionMethodFPX];
+    vc.delegate = self;
+    
+    [self.navigationController pushViewController:vc animated:YES];
+}
+
+- (void)payWithBankAccount:(STPPaymentMethodParams *)pmParams {
```
yuki-stripe: supernit: `paymentMethodParams`

Example/Custom Integration/FPXExampleViewController.m
```diff
@@ -123,6 +126,27 @@ - (void)pay {
...
+
+#pragma mark - STPAuthenticationContext
+
+- (UIViewController *)authenticationPresentingViewController {
+    return self.navigationController.topViewController;
+}
+
+- (void)authenticationContextWillDismissViewController:(UIViewController *)viewController {
+    // If we're launching directly into the SFSafariViewController for authentication from the bank selector,
+    // we'll want to remove the bank selector from the VC stack and pop back directly to our main controller on dismiss.
```
yuki-stripe: supernit: as a user I might not understand why we want to do this (is it necessary? what happens if I don't do it?)
Suggestion: "// remove the bank selector from the VC stack and pop back directly to our main controller on dismiss for better UX".

Issue #1030, opened 2018-10-22
Feature Request: improved VoiceOver support for STPPaymentCardTextField


provTheodoreNewell:
The STPPaymentCardField could use some improvements for VoiceOver support.

## Summary
For a user interacting with the `STPPaymentCardTextField` with VoiceOver enabled, the following is desired:
1. While editing the card number part of the field before a complete card number has been entered, the VoiceOver swipe gesture to move focus to the next onscreen element doesn't take the user to the expiration part of the field. Instead, the user is taken to whatever element follows the entire STPPaymentCardField.
2. If you enter an incorrect card number, expiration, CVC, or postal code, a user with VoiceOver would expect to hear that the field is invalid (corresponding to when the field changes color to the red/error color).
2.1. If you move your finger over any of the fields within STPPaymentCardField and they are invalid, the VoiceOver description should indicate they are invalid. 
3. If you enter a valid card number, the cursor focus automatically moves to the expiration part of the field. VoiceOver should activate to indicate the current field has changed to expiration date. Likewise after entering a valid expiration and cursor focus moves to the CVC, and CVC to postal code.

## iOS version
iOS 12

## Installation method
CocoaPods

## SDK version
13.2.0

WestonHanners:
👍 I wanna second this, this is affecting an app I am working on.

ThuggishNuggets:
Chiming in; also being affected by this issue. The current incarnation yields a really terrible experience for users that utilize VoiceOver accessibility features. Users have no way of knowing when focus moves from card number to expiration to CVC, or when validation failures occur.

Also.. Might want to bump up the priority on this issue since I'm pretty sure user-facing private/commercial SDKs are still governed by the Americans with Disabilities Act (ADA).

davidme-stripe:
Sorry that this took so long! We're addressing all the feedback in this issue in #1384, and we'll do a deeper VoiceOver audit in the near future.

davidme-stripe:
This should be fixed in v17.0.2.

PR #1384, opened 2019-09-18
Fix a few VoiceOver issues

davidme-stripe:
## Summary
* STPPaymentCardTextField should announce to the user when the insertion point moves between the different fields. This would ideally be handled by a targeted `UIAccessibilityLayoutChangedNotification` with the new targeted view, but that doesn't appear to work. `UIAccessibilityScreenChangedNotification` resolves this, alerting the user whenever the firstResponder changes. (But it's a big hammer. 😥)
* A VoiceOver user should be able to tell if a field contains invalid input. To implement this, we'll copy Safari's behavior for the [aria-invalid form attribute](https://www.w3.org/WAI/WCAG21/working-examples/aria-invalid-data-format/), which causes it to speak "Invalid Data" when the element is tapped after validation has failed.
* In iOS 13, VoiceOver can speak out a text field letter-by-letter when prompted, which is a behavior we'll want to adopt on all our card fields. (This seems preferable to having VoiceOver try to pronounce a British postal code, or read out a card number as "four hundred twenty four thousand etc".)
* In iOS 13, a VoiceOver user will face a crash deep in the Accessibility framework (related to mutating an array while iterating through it) when they open an STPAddCardViewController. This seems like an iOS bug, but we can work around it by explicitly defining the `accessibilityElementCount` for the STPPaymentCardTextFieldCell.
* VoiceOver users should be able to swipe between fields in an STPAddCardViewController. We'll fix this by implementing `accessibilityElementAtIndex` to access each field in the underlying `STPPaymentCardTextField` of `STPPaymentCardTextFieldCell`.
* Our shipping fields had a similar issue, with the STPAddressFieldTableViewCell being identified as a separate element from the underlying text field.

## Motivation
#1030

## Testing
Tested on my iOS 13 phone using VoiceOver.

### DIFF
--- a/Stripe/STPFormTextField.m
+++ b/Stripe/STPFormTextField.m
@@ -218,9 +218,11 @@ typedef NSAttributedString* (^STPFormTextTransformationBlock)(NSAttributedString
 
 - (NSAttributedString *)accessibilityAttributedValue {
     NSMutableAttributedString *attributedString = [self.attributedText mutableCopy];
+    #ifdef __IPHONE_13_0
     if (@available(iOS 13.0, *)) {
         [attributedString addAttribute:UIAccessibilitySpeechAttributeSpellOut value:@(YES) range:NSMakeRange(0, [attributedString length])];
     }
+    #endif
     if (!self.validText) {
         NSString *invalidData = STPLocalizedString(@"Invalid data.", @"Spoken during VoiceOver when a form field has failed validation.");
         NSMutableAttributedString *failedString = [[NSMutableAttributedString alloc] initWithString:invalidData attributes:@{UIAccessibilitySpeechAttributePitch: @(0.6)}];
--- a/Stripe/STPFormTextField.m
+++ b/Stripe/STPFormTextField.m
@@ -9,6 +9,7 @@
 #import "STPFormTextField.h"
 
 #import "NSString+Stripe.h"
+#import "STPLocalizationUtils.h"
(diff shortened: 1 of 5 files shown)


Stripe/STPFormTextField.m
```diff
@@ -206,6 +207,31 @@ - (void)setAttributedText:(NSAttributedString *)attributedText {
...
+- (NSAttributedString *)accessibilityAttributedValue {
+    NSMutableAttributedString *attributedString = [self.attributedText mutableCopy];
+    #ifdef __IPHONE_13_0
+    if (@available(iOS 13.0, *)) {
+        [attributedString addAttribute:UIAccessibilitySpeechAttributeSpellOut value:@(YES) range:NSMakeRange(0, [attributedString length])];
+    }
+    #endif
+    if (!self.validText) {
+        NSString *invalidData = STPLocalizedString(@"Invalid data.", @"Spoken during VoiceOver when a form field has failed validation.");
+        NSMutableAttributedString *failedString = [[NSMutableAttributedString alloc] initWithString:invalidData attributes:@{UIAccessibilitySpeechAttributePitch: @(0.6)}];
```
yuki-stripe: You can control the pitch, how neat!

davidme-stripe: Safari doesn't do this, but I thought it'd be nice to make it more obvious that this is a system message and not part of the text field's value.

PR #1386, opened 2019-09-20
Remove SCA + Apple Pay recommendation in docs, example code

yuki-stripe:
## Summary
Remove unnecessary Apple Pay + SCA code.

## Motivation
https://jira.corp.stripe.com/browse/IOS-1373

## Testing
Tested Apple Pay on device.  
Simulator is weird, delegate method doesn't get called.


mwoollard:
Re: the note on simulator testing of ApplePay, it broke with Xcode 10.3, worked prior to that. I've not checked Xcode 11.

PR #1390, opened 2019-09-26
Add FPX to Basic Integration

davidme-stripe:
## Summary
Adds FPX to STPPaymentOptionsViewController.
* To support sending FPX details without first creating a PaymentMethod on the Stripe end, this makes STPPaymentMethodParams conform to <STPPaymentOption> and implements the appropriate UI-level methods for display in an STPPaymentOptionsViewController. FPX STPPaymentOptions will be returned as STPPaymentMethodParams objects, and can be treated similarly to STPPaymentMethods in much of Basic Integration.
* Added FPX as an `STPPaymentOptionType`. `STPPaymentOptionTypeAll` is now `STPPaymentOptionTypeDefault` and does not include FPX. This is a breaking change, and we've added a deprecation warning with guidance to switch to `.Default`.
* `STPPaymentResult` is now initialized from an `<STPPaymentOption>` and filled with either an `STPPaymentMethod` or an `STPPaymentMethodParams`. This means `.paymentMethod` is now optional, which is a breaking change, but no one will encounter a nil `.paymentMethod` unless they're adopting FPX. We've added a convenience function, `[STPPaymentIntentParams configureWithPaymentResult:]`, to easily set up an STPPaymentIntentParams from an STPPaymentResult. (See [CheckoutViewController.swift](https://github.com/stripe/stripe-ios/pull/1390/files#diff-98301c9065e42b72872def95f81ed842L272) for usage. We'll want to update the stripe.com docs to match.)
* The Standard Integration app now has a `Country` selector in Settings, which allows the user to set the currency to `myr` for FPX testing. The selected country is also passed to example-ios-backend to configure the allowed_payment_methods and currency.
* Custom Integration now also supports sending a country to example-ios-backend.

## Testing
Tested in Custom Integration and Standard Integration apps.


Example/Custom Integration/MyAPIClient.h
```diff
@@ -33,10 +33,11 @@ typedef void (^STPCreateSetupIntentCompletionHandler)(MyAPIClientResult status,
  method signature is the most interesting part: you need some way to ask *your* backend to create
  a PaymentIntent with the correct properties, and then it needs to pass the client secret back.
  
+ @param additionalParameters additional parameters to pass to the example backend
  @param completion completion block called with status of backend call & the client secret if successful.
  @see https://stripe.com/docs/payments/payment-intents/ios
  */
-- (void)createPaymentIntentWithCompletion:(STPPaymentIntentCreationHandler)completion;
+- (void)createPaymentIntentWithCompletion:(STPPaymentIntentCreationHandler)completion additionalParameters:(NSString * _Nullable)additionalParameters;
```
csabol-stripe: Is this going to convert to swift strangely? I'd probalby expect `createPaymentIntentWithAdditionalParameters:(NSString * _Nullable)additionalParameters completion:(STPPaymentIntentCreationHandler)completion;`

I'd also probably expect `additionalParameters` to be an NSDictionary but since this is sample backend code don't feel strongly on either of these

Example/Standard Integration/BrowseProductsViewController.swift
```diff
@@ -55,6 +51,13 @@ class BrowseProductsViewController: UICollectionViewController {
             }, completion: nil)
         }
     }
+    
+    var numberFormatter : NumberFormatter = {
+        let numberFormatter = NumberFormatter()
+        numberFormatter.numberStyle = .currency
+        numberFormatter.usesGroupingSeparator = true
+        return numberFormatter
+    }()
```
csabol-stripe: 👏

Stripe/STPPaymentIntentParams.m
```diff
@@ -78,6 +81,14 @@ - (nullable NSString *)setupFutureUsageRawString {
     }
 }
 
+- (void)configureWithPaymentResult:(STPPaymentResult *)paymentResult {
+    if (paymentResult.paymentMethod) {
+        _paymentMethodId = paymentResult.paymentMethod.stripeId;
```
csabol-stripe: copy

Stripe/STPPaymentMethodParams.m
```diff
@@ -108,4 +108,53 @@ + (nonnull NSDictionary *)propertyNamesToFormFieldNamesMapping {
...
+    }
+}
+
+- (NSString *)label {
+    switch (self.type) {
+        case STPPaymentMethodTypeCard:
+            if (self.card != nil) {
+                STPCardBrand brand = [STPCardValidator brandForNumber:self.card.number];
+                NSString *brandString = STPStringFromCardBrand(brand);
+                return [NSString stringWithFormat:@"%@ %@", brandString, self.card.last4];
```
csabol-stripe: this should probably be localized. Okay for a follopwup

Stripe/STPPaymentOptionTableViewCell.m
```diff
@@ -117,6 +122,27 @@ - (void)configureWithPaymentOption:(id<STPPaymentOption>)paymentOption theme:(ST
...
+    self.leftIcon.image = [STPImageLibrary bankIcon];
+    self.leftIcon.tintColor = [self primaryColorForPaymentOptionWithSelected:NO];
+
+    // Title label
+    self.titleLabel.font = theme.font;
+    self.titleLabel.textColor = self.theme.primaryForegroundColor;
+    self.titleLabel.text = STPLocalizedString(@"Online Banking (FPX)", @"Button to pay with a Bank Account (using FPX).");
+
+    // Checkmark icon
+    self.checkmarkIcon.hidden = YES;
```
csabol-stripe: does this get reset if the cell is reused for a non-FPX option?

Stripe/STPPaymentOptionTuple.h
```diff
@@ -20,7 +20,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 + (instancetype)tupleWithPaymentOptions:(NSArray<id<STPPaymentOption>> *)paymentOptions
                   selectedPaymentOption:(nullable id<STPPaymentOption>)selectedPaymentOption
-                      addApplePayOption:(BOOL)applePayEnabled;
+                      addApplePayOption:(BOOL)applePayEnabled
```
csabol-stripe: can we get rid of the apple pay specific param and just use additionalOptions?

Stripe/STPPaymentOptionsInternalViewController.m
```diff
@@ -166,6 +169,31 @@ - (BOOL)isPaymentOptionDetachable:(id<STPPaymentOption>)paymentOption {
     return YES;
 }
 
+- (NSArray<id<STPPaymentOption>> *)cardPaymentOptions {
+    return [self.paymentOptions filteredArrayUsingPredicate:[NSPredicate predicateWithBlock:^BOOL(id<STPPaymentOption> _Nullable evaluatedObject, NSDictionary<NSString *,id> * __unused _Nullable bindings) {
+        if ([evaluatedObject isKindOfClass:[STPPaymentMethodParams class]]) {
+            STPPaymentMethodParams *paymentMethodParams = (STPPaymentMethodParams *)evaluatedObject;
+            if (paymentMethodParams.type == STPPaymentMethodTypeFPX) {
```
csabol-stripe: can this be like `if (paymentMethodParams.type != STPPaymentMethodTypeCard)` so we don't accidentally regress it when adding a new type'?

Stripe/STPPaymentOptionsInternalViewController.m
```diff
@@ -166,6 +169,31 @@ - (BOOL)isPaymentOptionDetachable:(id<STPPaymentOption>)paymentOption {
...
+        }
+        return YES;
+    }]];
+}
+
+- (NSArray<id<STPPaymentOption>> *)apmPaymentOptions {
+    return [self.paymentOptions filteredArrayUsingPredicate:[NSPredicate predicateWithBlock:^BOOL(id<STPPaymentOption> _Nullable evaluatedObject, NSDictionary<NSString *,id> * __unused _Nullable bindings) {
+        if ([evaluatedObject isKindOfClass:[STPPaymentMethodParams class]]) {
+            STPPaymentMethodParams *paymentMethodParams = (STPPaymentMethodParams *)evaluatedObject;
+            if (paymentMethodParams.type == STPPaymentMethodTypeFPX) {
```
csabol-stripe: i think `== STPPaymentMethodTypeFPX` makes sense here since we don't have basic integration support for other apms (would maybe add a comment to that effect)

Example/Custom Integration/MyAPIClient.h
davidme-stripe: Yeah, that was roughly my feeling. It should take a dictionary and convert all the parameters properly, but this was a quick change to make an Objective-C-only sample app work, so meh. I'm more worried about how this additional parameter adds incremental complexity to our sample app.

Stripe/STPPaymentOptionTuple.h
davidme-stripe: I did originally, but found that `applePayEnabled` is a little more complex than that. We need the `appleMerchantIdentifier` from STPPaymentConfiguration, which the STPPaymentOptionTuple isn't aware of, and I don't really want to pipe it all the way down here.

Stripe/STPPaymentOptionsInternalViewController.m
davidme-stripe: Sure, that'd be more forward-thinking. (Though a regression here would be very visible!)

Stripe/STPPaymentOptionsInternalViewController.m
davidme-stripe: Makes sense.

Stripe/STPPaymentIntentParams.m
davidme-stripe: good catch, thanks!

Stripe/STPPaymentOptionTableViewCell.m
davidme-stripe: It does! For selectable cells, the hidden field is populated with the selected state of the card.

Stripe/STPPaymentMethodParams.m
davidme-stripe: This isn't localized in STPPaymentMethod.m or STPCard.m either, though we do localize the similar "%@ Ending In %@" strings. We should probably localize all of them.

Stripe/STPPaymentOptionTuple.h
davidme-stripe: For reference:

```
- (BOOL)applePayEnabled {
    return self.appleMerchantIdentifier &&
    (self.additionalPaymentOptions & STPPaymentOptionTypeApplePay) &&
    [Stripe deviceSupportsApplePay];
}
```

Stripe/STPPaymentMethodParams.m
csabol-stripe: 👍

Stripe/STPPaymentOptionsInternalViewController.m
```diff
@@ -173,19 +173,19 @@ - (BOOL)isPaymentOptionDetachable:(id<STPPaymentOption>)paymentOption {
...
         if ([evaluatedObject isKindOfClass:[STPPaymentMethodParams class]]) {
             STPPaymentMethodParams *paymentMethodParams = (STPPaymentMethodParams *)evaluatedObject;
-            if (paymentMethodParams.type == STPPaymentMethodTypeFPX) {
-                return NO;
+            if (paymentMethodParams.type == STPPaymentMethodTypeCard) {
+                return YES;
             }
         }
-        return YES;
+        return NO;
```
csabol-stripe: couldn'e this still be yes in case it's an `STPPaymentMethod`?

Stripe/STPPaymentOptionsInternalViewController.m
```diff
@@ -173,11 +173,11 @@ - (BOOL)isPaymentOptionDetachable:(id<STPPaymentOption>)paymentOption {
...
         if ([evaluatedObject isKindOfClass:[STPPaymentMethodParams class]]) {
             STPPaymentMethodParams *paymentMethodParams = (STPPaymentMethodParams *)evaluatedObject;
-            if (paymentMethodParams.type == STPPaymentMethodTypeCard) {
-                return YES;
+            if (paymentMethodParams.type != STPPaymentMethodTypeCard) {
+                return NO;
             }
         }
-        return NO;
+        return YES;
```
csabol-stripe: ah there we go 🙃

PR #1052, opened 2018-12-01
[wip] Push provisioning

jack-stripe:
## Summary
Adds beta Issuing support for push provisioning. This will not do anything if your account is not in the necessary beta.

## Motivation
new feature

## Testing
Added automated tests

PR #1396, opened 2019-09-28
Push provisioning take 2

jack-stripe:
## Summary

This is a second take at the functionality in #1052.
- Adds wrappers for the (private) /v1/issuing/cards/:id/push_provisioning_details API, which will generate and spit out an encrypted card payload for use with `PKAddPaymentPassViewController`.
- Adds `STPPushProvisioningContext`, a stateful class that is used to handle fetching an ephemeral key from a user's server and then using that key to call the aforementioned API.
- Adds `STPFakeAddPaymentPassViewController`, a UI class that is API-compatible with `PKAddPaymentPassViewController` but can be more easily used for testing and debugging.

## Motivation
Enable push provisioning

## Testing
Added snapshot tests for the new API code.


davidme-stripe:
The code looks good, though I have some FUD around how App Review will treat references to PKAddPaymentPassViewController in apps that don't have the correct entitlement. We might want to send this through App Review at least once before we ship it out in an SDK.

Can you fix the errors in the CI build?
https://travis-ci.org/stripe/stripe-ios/jobs/591550245#L1426
https://travis-ci.org/stripe/stripe-ios/jobs/591550240#L1305
https://travis-ci.org/stripe/stripe-ios/jobs/591550246#L1045
https://travis-ci.org/stripe/stripe-ios/jobs/591550240#L1312

jack-stripe:
Ooh, good point on the app warnings. I'll try and submit something to the app store this week. Fixed the build issues in the meantime!

davidme-stripe:
Sounds good! I’m cool with this going in once we’ve tried that.

PR #1405, opened 2019-10-04
Update MIGRATING for 18.0.0

davidme-stripe:
## Summary
Update with migration instructions for 18.0.0.

## Testing
👀


MIGRATING.md
```diff
@@ -1,5 +1,17 @@
 ## Migration Guides
 
+### Migrating from versions < 18.0.0
+* Some error messages from the Payment Intents API are now localized to the user's display language. If your application's logic depends on specific `message` strings from the Stripe API, please use the error [`code`](https://stripe.com/docs/error-codes) instead.
```
csabol-stripe: We should probably add that you have to bump your min iOS version to at least 10.0

davidme-stripe: Done

PR #1422, opened 2019-10-23
Add FPX offline status endpoint

davidme-stripe:
## Summary
Our FPX implementation offers a status endpoint that we should use to inform the user that specific banks are offline. The FPX team says we should handle this by appending "Offline" to the bank's name: We shouldn't grey out the label or prevent a user from selecting it.

The endpoint is a little unreliable, so we should default to assuming the bank is online unless we hear otherwise.

## Motivation
Matches the behavior of the FPX Element.

## Testing
Tested in simulator.


Stripe/STPAPIClient.m
```diff
@@ -813,4 +815,16 @@ - (void)createPaymentMethodWithParams:(STPPaymentMethodParams *)paymentMethodPar
 
 }
 
+#pragma mark - FPX
+
+- (void)retrieveFPXBankStatusWithCompletion:(STPFPXBankStatusCompletionBlock)completion {
+    [STPAPIRequest<STPFPXBankStatusResponse *> getWithAPIClient:self
```
mshafrir-stripe: ooc, does this handle errors/exceptions?

Stripe/STPBankSelectionTableViewCell.m
```diff
@@ -95,6 +96,10 @@ - (void)configureWithBank:(STPFPXBankBrand)bankBrand theme:(STPTheme *)theme sel
     // Title label
     self.titleLabel.font = theme.font;
     self.titleLabel.text = STPStringFromFPXBankBrand(self.bank);
+    if (offline) {
+        self.titleLabel.text = [self.titleLabel.text stringByAppendingString:@" - "];
```
mshafrir-stripe: Would it be better to localize with the dash as part of the string, for example `"%s - Offline"`

I'm thinking of RTL translation scenarios.

Stripe/STPBankSelectionViewController.m
```diff
@@ -91,7 +115,11 @@ - (UITableViewCell *)tableView:(UITableView *)tableView
     STPBankSelectionTableViewCell *cell = [tableView dequeueReusableCellWithIdentifier:STPBankSelectionCellReuseIdentifier forIndexPath:indexPath];
     STPFPXBankBrand bankBrand = indexPath.row;
     BOOL selected = self.selectedBank == bankBrand;
-    [cell configureWithBank:bankBrand theme:self.theme selected:selected enabled:!self.loading];
+    BOOL offline = NO;
+    if (self.bankStatus && ![self.bankStatus bankBrandIsOnline:bankBrand]) {
```
mshafrir-stripe: nit, `BOOL offline = self.bankStatus && ![self.bankStatus bankBrandIsOnline:bankBrand]`?

Stripe/STPBankSelectionTableViewCell.m
davidme-stripe: Ah, good point.

Stripe/STPAPIClient.m
davidme-stripe: No, I don't think there's anything useful we could do with such an error. We could log them to the console, I guess.

Stripe/STPBankSelectionViewController.m
davidme-stripe: ... yes 🤦‍♂

Stripe/STPAPIClient.m
csabol-stripe: it does in the sense that we will generate an error if the response is not parseable for instance

Stripe/STPBankSelectionTableViewCell.m
```diff
@@ -95,6 +96,10 @@ - (void)configureWithBank:(STPFPXBankBrand)bankBrand theme:(STPTheme *)theme sel
     // Title label
     self.titleLabel.font = theme.font;
     self.titleLabel.text = STPStringFromFPXBankBrand(self.bank);
+    if (offline) {
+        NSString *format = STPLocalizedString(@"%@ - Offline", @"Bank name when bank is offline for maintenance.");
+        self.titleLabel.text = [NSString stringWithFormat:format, STPStringFromFPXBankBrand(self.bank)];
```
csabol-stripe: Do you think it's worthwhile to just always use `[NSString localizedStringWithFormat:]`?

Stripe/STPBankSelectionViewController.m
```diff
@@ -52,11 +54,27 @@ - (instancetype)initWithBankMethod:(STPBankSelectionMethod)bankMethod
...
         self.title = STPLocalizedString(@"Bank Account", @"Title for bank account selector");
     }
     return self;
 }
 
+- (void)dealloc {
+    [[NSNotificationCenter defaultCenter] removeObserver:self];
+}
+
+- (void)refreshFPXStatus {
```
csabol-stripe: style-nit: 
```suggestion
- (void)_refreshFPXStatus {
```

Stripe/STPBankSelectionViewController.m
```diff
@@ -76,6 +94,12 @@ - (BOOL)useSystemBackButton {
     return YES;
 }
 
+- (void)updateWithBankStatus:(STPFPXBankStatusResponse *)bankStatusResponse {
```
csabol-stripe: style-nit:
```suggestion
- (void)_updateWithBankStatus:(STPFPXBankStatusResponse *)bankStatusResponse {
```

Stripe/STPBankSelectionViewController.m
```diff
@@ -76,6 +94,12 @@ - (BOOL)useSystemBackButton {
...
+    self.bankStatus = bankStatusResponse;
+    
+    [self.tableView reloadData];
```
csabol-stripe: Any reason _not_ to do a `begin/EndUpdates` here rather than a `reloadData`?

Stripe/STPBankSelectionTableViewCell.m
davidme-stripe: STPLocalizedString does some bundle searching logic to allow for users to override our localizations. We'd want a similar macro for `localizedStringWithFormat:` if we wanted to start using that.

Stripe/STPAPIClient.m
davidme-stripe: I can't determine if we have any standardized way to communicate these sort of errors up to users. `retrieveFPXBankStatusWithCompletion` will return an error to the completion block if the response isn't parseable, but the BankSelectionViewController can't do anything sensible with that other than silently failing. We could `NSLog()` a warning, but we don't have any precedent for that.

Stripe/STPBankSelectionViewController.m
davidme-stripe: We're reloading all cells, but I suppose we could.

PR #1423, opened 2019-10-23
Move Travis to Xcode 11.2

davidme-stripe:
## Summary
The Travis Xcode 11 image has been updated to remove the iOS 10 simulators. Move to their 11.2 image to see if it resolves this.


davidme-stripe:
Looks like Xcode 11.2 contains the old simulators. It's technically still in beta, but this seems like a reasonable fix.

Issue #1421, opened 2019-10-21
Checkmark icon on top of Delete button


sushimatou:
## Summary
When a card is selected in the card options tableview, edit mode is triggered, just after tapping the "red icon circle" on the left side:

the checkmark of the row is misplaced, in front of the delete button. A weird animation occurs also on this checkmark when deletion is completed by swiping or clicking on the delete button. 

## Code to reproduce
`STPPaymentOptionTableViewCell`, the `checkmark` position is not the right one, probably when `layoutSubviews()` is called.

## iOS version
iOS 13.0

## Installation method
Cocoapods with Xcode 10.3

## SDK version
Stripe 10.2

## Screenshot
![IMG_7529](https://user-images.githubusercontent.com/10923684/67222150-0a98c300-f42d-11e9-87d7-b5b22bed2cca.jpg)

csabol-stripe:
@sushimatou thanks for the report! I've got a branch that fixes this, just dealing with some XCode issues then I'll get it merged in :)

davidme-stripe:
This was fixed in v18.1.0!

PR #1425, opened 2019-10-25
Fixes checkmark overlapping delete in iOS 13

csabol-stripe:
## Summary

Cell layout relative to contentView. Also converts to autolayout.

## Motivation

#1421 

## Testing

Snapshot tests for regressions. Tested deleted selected in Standard Integration


davidme-stripe:
> Also converts to autolayout.
:head-exploding:

csabol-stripe:
Ooo I had run the snapshot tests on the wrong device. We actually add room for an additional character in Japanese so updated that snapshot

PR #1427, opened 2019-10-25
Workaround iOS 13 TableViewRowData issue

### LINKED ISSUE #1426: Crash when adding a card
## Summary

The app crashes when clicking on "Add Card". It seems to happen with certain shipping addresses. 

## Code to reproduce
See in this video:
https://www.dropbox.com/s/5uofsjvpqoc11rf/stripe%20bug.MP4?dl=0

## iOS version
iOS 13

## Installation method
Cocoapods 

## SDK version
14.0.0

## Other information
The crash error is:
Fatal Exception: NSInternalInconsistencyException
Requested the number of rows for section (1) which is out of bounds

And it crashes on line 401 and 410 in STPAddCardViewController.
Line 401:
    NSInteger rowsInSection = [self.tableView numberOfRowsInSection:STPPaymentCardBillingAddressSection];

Line 410:
    NSInteger rowsInSection = [self.tableView numberOfRowsInSection:STPPaymentCardBillingAddressSection];

davidme-stripe:
## Summary
When we request numberOfRowsInSection from a UITableView while the UITableView's ViewController is being initialized, the internal UITableViewRowData is sometimes not in a good state and doesn't yet know how many sections we have. Work around this by requesting numberOfRowsInSection: from our controller instead of the tableView.

## Motivation
Fixes #1426.

## Testing
I can't reproduce the issue locally, but @odedharth says this fixed it.

### DIFF
--- a/Stripe/STPAddCardViewController.m
+++ b/Stripe/STPAddCardViewController.m
@@ -397,7 +397,7 @@ typedef NS_ENUM(NSUInteger, STPPaymentCardSection) {
 #pragma mark - STPAddressViewModelDelegate
 
 - (void)addressViewModel:(__unused STPAddressViewModel *)addressViewModel addedCellAtIndex:(NSUInteger)index {
-    NSInteger rowsInSection = [self.tableView numberOfRowsInSection:STPPaymentCardBillingAddressSection];
+    NSInteger rowsInSection = [self tableView:self.tableView numberOfRowsInSection:STPPaymentCardBillingAddressSection];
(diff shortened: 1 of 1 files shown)


shamilovtim:
Hello @davidme-stripe could we have this cherrypicked onto a 14.0.3 release? We're currently stuck with no viable upgrade path due to a monolithic implementation that will take a long time to fix, 14.x is the last version that works for `token` users and this is a common bug in production.

PR #1429, opened 2019-10-30
Fix build issue for Catalyst

davidme-stripe:
## Summary
We're not testing Catalyst in CI yet, so we didn't catch this issue.

## Testing
Tested building on Catalina and in CI.


davidme-stripe:
er wait this patch is totally wrong, one moment

PR #1439, opened 2019-11-08
Deprecate old methods in STPAddCardViewController

yuki-stripe:
## Summary
Not deprecating optional protocol methods is particularly bad because there is no indication an upgrading user's integration is now broken; the old methods are simply never called.

## Motivation
https://jira.corp.stripe.com/browse/IOS-1447

## Testing
They show up as deprecated when implemented in objc, but not in Basic Integration (swift).  Not sure why.  The methods show up as crossed out in the autocomplete.


davidme-stripe:
Weird that they're not marked as deprecated in Swift. I guess we could check `respondsToSelector:` for these and provide guidance at runtime if someone has implemented them, but that could be annoying if someone accidentally ships code like that. :|

Issue #1441, opened 2019-11-11
Amex CVC check


iKorich:
Hi guys,

continue thread https://github.com/stripe/stripe-ios/issues/890,

in STPCardValidator.m you have:

```
+ (NSUInteger)minCVCLength {
    return 3;
}
```

and that doesn't work for Amex card
please change it to 
```
+ (NSUInteger)minCVCLengthForCardBrand:(STPCardBrand)brand {
    switch (brand) {
        case STPCardBrandAmex:
        case STPCardBrandUnknown:
            return 4;
        default:
            return 3;
    }
}
```

yuki-stripe:
Hi @iKorich,

It's possible for Amex cards to have a 3 digit CVC equivalent that is accepted by the Stripe API.  As far as I know this has not changed, and the current code is correct.

See https://github.com/stripe/stripe-ios/issues/103

iKorich:
Hi @yuki-stripe ,

please check https://www.cvvnumber.com/cvv.html

yuki-stripe:
@iKorich This value is referred to by American Express as a "Three digit Card Security Code (CSC)" here: https://global.americanexpress.com/faq?question=Do%20I%20need%20to%20do%20anything%20before%20using%20my%20Gift%20Card?

If you find the Stripe API does not accept this as the "CVC" value, please reopen this issue!

PR #1442, opened 2019-11-12
update min CVC check

iKorich:
## Summary
move minCVCLength to minCVCLengthForCardBrand

## Motivation
issue #1441 

## Testing
Tests are updated


davidme-stripe:
Closing this out: As mentioned in #1441, Amex will accept 3-digit CVCs in certain situations, so we'd rather not prevent a user from entering one. Feel free to reach out if you think we should reconsider this!

PR #1443, opened 2019-11-14
Calls source_cancel endpoint for web-based 3DS challenge flow cancelations

csabol-stripe:
## Summary

When we determine that a webview closed before the action was completed, we cancel the 3DS source identified in the URL using the new `source_cancel` endpoints.

## Motivation

Sources were being left in an unresolved state.

## Testing

Tested PaymentIntent path with "Basic Integration" sample. Tested SetupIntent with https://github.com/stripe-samples/mobile-saving-card-without-payment

### DIFF
--- a/Stripe/Payments/STPPaymentHandler.m
+++ b/Stripe/Payments/STPPaymentHandler.m
@@ -22,7 +22,7 @@
 #import "STPPaymentIntentParams.h"
 #import "STPPaymentHandlerActionParams.h"
 #import "STPIntentAction+Private.h"
-#import "STPIntentActionRedirectToURL.h"
+#import "STPIntentActionRedirectToURL+Private.h"
 #import "STPIntentActionUseStripeSDK.h"
 #import "STPSetupIntent.h"
 #import "STPSetupIntentConfirmParams.h"
@@ -500,7 +500,7 @@ withAuthenticationContext:(id<STPAuthenticationContext>)authenticationContext
                                                                                           intentID:_currentAction.intentStripeID];
                     
                     [_currentAction.apiClient authenticate3DS2:authRequestParams
-                                              sourceIdentifier:authenticationAction.useStripeSDK.threeDS2SourceID
+                                              sourceIdentifier:authenticationAction.useStripeSDK.threeDSSourceID
                                                      returnURL:_currentAction.returnURLString
                                                     maxTimeout:_currentAction.threeDSCustomizationSettings.authenticationTimeout
                                                     completion:^(STP3DS2AuthenticateResponse * _Nullable authenticateResponse, NSError * _Nullable error) {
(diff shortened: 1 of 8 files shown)


Stripe/STPAPIClient.m
```diff
@@ -52,7 +52,7 @@
 #import "STPCategoryLoader.h"
 #endif
 
-static NSString * const APIVersion = @"2019-05-16";
+static NSString * const APIVersion = @"2019-11-05";
```
csabol-stripe: will revert this. I _think_ it's okay but I haven't tested fully and it's not needed

Stripe/Payments/STPPaymentHandler.m
```diff
@@ -816,8 +823,59 @@ - (void)transactionDidPresentChallengeScreen:(STDSTransaction *)transaction {
                                                                                uiType:transaction.presentedChallengeUIType];
 }
 
+- (void)_markChallengeCanceledWithCompletion:(STPBooleanSuccessBlock)completion {
```
yuki-stripe: I'm guessing not b/c of the PR title, but do we want to call this for native 3DS2 cancels as well?  Not obvious why we don't need to

Stripe/STPIntentActionRedirectToURL.m
```diff
@@ -5,14 +5,16 @@
...
+#import "STPIntentActionRedirectToURL+Private.h"
 
 #import "NSDictionary+Stripe.h"
+#import "NSURLComponents+Stripe.h"
 
 @interface STPIntentActionRedirectToURL()
 
 @property (nonatomic, nonnull) NSURL *url;
 @property (nonatomic, nullable) NSURL *returnURL;
+@property (nonatomic, nullable) NSString *threeDSSourceID;
```
yuki-stripe: nit, copy

Stripe/Payments/STPPaymentHandler.m
csabol-stripe: We call this for hosted 3DS2, but not native since native 3DS2 cancels already go to the ACS

yuki-stripe: Could you add a comment saying something to this effect?

Stripe/STPIntentActionRedirectToURL.m
```diff
@@ -51,6 +53,7 @@ + (nullable instancetype)decodedObjectFromAPIResponse:(nullable NSDictionary *)r
     redirect.url = url;
     redirect.returnURL = [dict stp_urlForKey:@"return_url"];
     redirect.allResponseFields = dict;
+    redirect->_threeDSSourceID = [[[NSURLComponents componentsWithURL:url resolvingAgainstBaseURL:NO] stp_queryItemsDictionary][@"client_secret"] copy];
```
yuki-stripe: thought I added a comment for this but I didn't - can we do `redirect.threeDSSourceID = ...` here for consistency?

PR #1444, opened 2019-11-19
Add initial support for Klarna

davidme-stripe:
## Summary
Add basic support for Klarna Sources. This attempts to provide a simple interface to create a Klarna transaction. More complex integrations are possible by using the `additionalAPIParameters` and `allResponseFields` fields of the STPSource[Params]. 

## Testing
In Non-Card Examples, unit tests, and CI.


Stripe/STPSourceKlarnaDetails.m
```diff
@@ -0,0 +1,48 @@
...
+//  Created by David Estes on 11/19/19.
+//  Copyright © 2019 Stripe, Inc. All rights reserved.
+//
+
+#import "STPSourceKlarnaDetails.h"
+
+#import "NSDictionary+Stripe.h"
+
+@interface STPSourceKlarnaDetails()
+@property (nonatomic, copy, readwrite) NSString *clientToken;
```
mshafrir-stripe: is there anything from our docs or Klarna docs we can use to document this field?

Stripe/PublicHeaders/STPKlarnaLineItem.h
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ An object representing a line item in a Klarna source.
+ @see https://stripe.com/docs/sources/klarna#create-source
+ */
+
+@interface STPKlarnaLineItem : NSObject
+
+/**
+ The line item's type. This is generally `sku` for a product, `tax` for taxes, and `shipping` for shipping.
```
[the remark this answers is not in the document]
davidme-stripe: Totally missed that page, thanks! Done.

Stripe/STPSourceKlarnaDetails.m
davidme-stripe: Added a link to Klarna's site.

Stripe/PublicHeaders/STPSourceParams.h
```diff
@@ -208,6 +208,65 @@ NS_ASSUME_NONNULL_BEGIN
...
+
+ @param returnURL            The URL the customer should be redirected to after
+                             they have successfully verified the payment.
+ @param currency             The currency the payment is being created in.
+ @param purchaseCountry      The ISO-3166 2-letter country code of the customer's location.
+ @param items                An array of STPKlarnaLineItems. Klarna will present these on the confirmation
+                             page. The total amount charged will be a sum of the `totalAmount` of each of these items.
+ @param customPaymentMethods Required for customers located in the US. This determines whether Pay Later and/or Slice It
+                             is offered to a US customer.
+ @param address              An STPAddress for the customer. At a minimum, an email, line1, postal code, city, and country must be provided.
```
mshafrir-stripe: `billingAddress`

Stripe/PublicHeaders/STPSourceParams.h
```diff
@@ -208,6 +208,65 @@ NS_ASSUME_NONNULL_BEGIN
...
+ @param firstName            The customer's first name.
```
mshafrir-stripe: `billingFirstName`

Stripe/PublicHeaders/STPSourceParams.h
```diff
@@ -208,6 +208,65 @@ NS_ASSUME_NONNULL_BEGIN
...
+ @param lastName             The customer's last name.
```
mshafrir-stripe: `billingLastName`

Stripe/PublicHeaders/STPSourceParams.h
```diff
@@ -208,6 +208,65 @@ NS_ASSUME_NONNULL_BEGIN
...
+
+ If the provided information is missing a line1, postal code, city, email, or first/last name, or if the country code is
+ outside the specified country, no address information will be sent to Klarna, and Klarna will prompt the customer to provide their address.
+
+ @param dateOfBirth           The customer's date of birth. This will be used by Klarna for a credit check in some EU countries.
```
mshafrir-stripe: `billingDOB`

Stripe/PublicHeaders/STPSourceParams.h
mshafrir-stripe: Should we mention that `STPAddress#name` will be ignored?

davidme-stripe: The compiler enforces that these docstrings match the name of the parameter, not the name of the selector.

And yeah, we should!

mshafrir-stripe: thanks for clarifying

davidme-stripe:
@yuki-stripe @csabol-stripe I'd like some extra iOS eyes on this, could one of you take a look?

Stripe/PublicHeaders/STPKlarnaLineItem.h
```diff
@@ -0,0 +1,73 @@
...
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+ The type of the Klarna line item.
+*/
+typedef NS_ENUM(NSUInteger, STPKlarnaLineItemType) {
+    /**
+     The line item for a product
+     */
+    STPKlarnaLineItemTypeSku,
```
csabol-stripe:
```suggestion
    STPKlarnaLineItemTypeSKU,
```

Stripe/PublicHeaders/STPKlarnaLineItem.h
```diff
@@ -0,0 +1,73 @@
...
+ The quantity to display for this line item.
+ */
+@property (nonatomic, copy) NSNumber *quantity;
+
+/**
+ The total price of this line item.
+ Note: This is the total price after multiplying by the quantity, not
+ the price of an individual item.
+ */
+@property (nonatomic, copy) NSNumber *totalAmount;
```
csabol-stripe: surprised that this isn't paired with a `currency` field. Is that intentional?
EDIT: currency is top-level on the source params. Maybe just add a comment saying that?

Stripe/PublicHeaders/STPSourceEnums.h
```diff
@@ -162,9 +162,35 @@ typedef NS_ENUM(NSInteger, STPSourceType) {
...
+
+    /**
+     Offer payments over 4 installments. (a.k.a. Pay Later in 4)
+     */
+    STPKlarnaPaymentMethodsPayIn4 = 1 << 0,
+
+    /**
+     Offer payments over an arbitrary number of installments. (a.k.a. Slice It)
+     */
+    STPKlarnaPaymentMethodsInstallments = 1 << 1,
```
csabol-stripe: Why are these options instead of an enum? Does it make sense to combine `PayIn4` and `Installments`?

Stripe/PublicHeaders/STPKlarnaLineItem.h
davidme-stripe: It exists in the underlying API, but I was trying to reduce boilerplate by making individual objects not have a currency, as it isn't valid for the individual items to have a different currency from the Source. If you found it surprising, maybe I should re-add currency here?

Stripe/PublicHeaders/STPKlarnaLineItem.h
davidme-stripe: I had it this way originally and changed it, but I forgot why. I'll try it again!

Stripe/PublicHeaders/STPSourceEnums.h
davidme-stripe: The API specifies that you can provide both, which will offer the user a choice of PayIn4 and Installments. It doesn't appear to do this in practice. :/

csabol-stripe: gotcha

Stripe/PublicHeaders/STPKlarnaLineItem.h
csabol-stripe: hmm yeah I'd probably lean towards adding it, I guess just with the comment that all the currencies need to match...Maybe we can say if it's left null it will default to the currency of the source?

csabol-stripe: @yuki-stripe @mshafrir-stripe thoughts? 
1) It _feels_ weird to set a Money Amount without specifying currency to me
2) But having a separate currency param on each line item feels like an easy way to make mistakes or at the very least write a bunch of redundant code
3) But the API has the currency param on each

Solutions:
1) Leave as is (no currency param)
2) Optional currency param (default to currency of owning Source)
3) "Factory" line items generator? This isn't really typical in objective-c, but I'm trying to think through true use cases for this. Like maybe you can't instantiate a line item separately, only through `SourceParams.addKlarnaLineItem(type, description, ...)`?

yuki-stripe: I suspect users won't be as surprised at the lack of currency.  I think I prefer not exposing it (1) with a note in the docstring ("Note: The amount is denominated in the currency of the top-level STPSourceParams")

mshafrir-stripe: I am in favor of not including `currency` for the guardrails benefit. Assuming that this class will only be used in the context of a `STPSourceParams`, I don't think it will result in confusion. A docstring will help as well as @yuki-stripe mentioned.

Is `STPKlarnaLineItem` used for both creation (i.e. params) and for the response object, or only creation?

PR #1445, opened 2019-11-20
Store last selected payment method

yuki-stripe:
## Summary
`STPPaymentContext` and `STPPaymentOptionsViewController` pre-select the last selected payment method via a `{customer_id: last_selected_payment_method_id}` store in NSUserDefaults.

* This aims to return the lost feature, not implement https://paper.dropbox.com/doc/Saving-last-selected-payment-method--Ao40PI6L1boSl8DXBN0S3zUxAg-YiSHZtquibTm2xrQ9me7O
## Motivation
https://jira.corp.stripe.com/browse/IOS-1365

## Testing
Added a UI test that covers reading, writing a PM id, writing nil, and multiple customers.

### DIFF
--- a/Stripe/PublicHeaders/STPPaymentOptionsViewController.h
+++ b/Stripe/PublicHeaders/STPPaymentOptionsViewController.h
@@ -102,14 +102,13 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, strong, nullable) STPUserInformation *prefilledInformation;
 
 /**
+ @note This is no longer recommended as of v18.3.0 - the SDK automatically saves the Stripe ID of the last selected
+ payment method using NSUserDefaults and displays it as the default pre-selected option.  You can override this behavior
+ by setting this property.
+ 
  The Stripe ID of a payment method to display as the default pre-selected option.
 
- Customer doesn't have a default payment method property, but you can store one
- (in its metadata, for example) and set this property accordingly.
-
  @note Setting this after the view controller's view has loaded has no effect.
- @note This class automatically saves the Stripe ID of the last selected payment method using NSUserDefaults
- and displays it as the default pre-selected option.  Settting this value overrides that behavior.
  */
 @property (nonatomic, copy, nullable) NSString *defaultPaymentMethod;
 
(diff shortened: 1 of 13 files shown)


Stripe/STPPaymentOptionsViewController.m
```diff
@@ -166,6 +175,19 @@ - (void)updateAppearance {
 }
     
 - (void)finishWithPaymentOption:(id<STPPaymentOption>)paymentOption {
+    BOOL isReusablePaymentMethod = [paymentOption isKindOfClass:[STPPaymentMethod class]] && ((STPPaymentMethod *)paymentOption).type == STPPaymentMethodTypeCard;
```
davidme-stripe: Why not `[paymentOption isKindOfClass:[STPPaymentMethod class]] && [paymentOption isReusable]`?

Stripe/PublicHeaders/STPPaymentOptionsViewController.h
```diff
@@ -103,11 +103,13 @@ NS_ASSUME_NONNULL_BEGIN
 
 /**
  The Stripe ID of a payment method to display as the default pre-selected option.
```
csabol-stripe: so do we not populate this with the value stored in NSUserDefaults?

yuki-stripe: right, I'm worried that would be a breaking change.  It would behave a little weirdly if we populated it, it would be nil until the network call finishes.

I've half a mind to deprecate the property, too.

WDYT?

csabol-stripe: hmm, maybe deprecate and make `retrieveLastSelectedPaymentMethodIDForCustomerWithCompletion` public?

dskibin:
Thank you for the fix and merge. Could you tell when this will be released?

Ariandr:
Hi @yuki-stripe 
By the way, could you please explain how to trigger this behavior of storing the last selected payment method?
Do we need to set a property or it will be done automatically after this update?

yuki-stripe:
@dskibin - You can expect a release within the next couple days.

@Ariandr - This will happen automatically.

Ariandr:
@yuki-stripe Great, thank you for the quick response!

PR #1464, opened 2019-12-17
Adds client secret format validation

csabol-stripe:
## Summary

Adds client secret validation utility methods for PaymentIntents and SetupIntents. Asserts if an invalid client secret is passed to `STPAPIClient`. Fails with new error type if invalid client secret is passed to `STPPaymentHandler`.

## Motivation

In response to dev support engineering request.

## Testing

Added automated unit tests.


Stripe/STPPaymentIntentParams.m
```diff
@@ -183,4 +184,20 @@ + (nonnull NSDictionary *)propertyNamesToFormFieldNamesMapping {
              };
 }
 
+#pragma mark - Utilities
+
++ (BOOL)isClientSecretValid:(NSString *)clientSecret {
+    static dispatch_once_t onceToken;
+    static NSRegularExpression *regex = nil;
+    dispatch_once(&onceToken, ^{
+        regex = [[NSRegularExpression alloc] initWithPattern:@"^pi_[^_]+_secret_[^_]+$"
```
mshafrir-stripe: This will match any non-underscore character, so `pi_    _secret_  ` will match.

I ended up using `^pi_([a-zA-Z0-9])+_secret_([a-zA-Z0-9])+$`

csabol-stripe: that seems reasonable to me, I'm unsure the source of truth here but stripe.js just uses the `[^_]` https://git.corp.stripe.com/stripe-internal/stripe-js-v3/blob/733bc8be/src/payment_intent/shared/index.js#L211

Stripe/Payments/STPPaymentHandler.m
```diff
@@ -122,7 +127,11 @@ - (void)handleNextActionForPayment:(NSString *)paymentIntentClientSecret
         NSAssert(NO, @"Should not handle multiple payments at once.");
         completion(STPPaymentHandlerActionStatusFailed, nil, [self _errorForCode:STPPaymentHandlerNoConcurrentActionsErrorCode userInfo:nil]);
         return;
+    } else if (![STPPaymentIntentParams isClientSecretValid:paymentIntentClientSecret]) {
```
csabol-stripe: @davidme-stripe overnight I was thinking of changing this to just be an assert in case for some reason the format changes in the future...what do you think? There are other places where we are depending on splitting the key

davidme-stripe: I'd rather not add asserts that could trigger in other folks' apps in production if we change the format. Best to give them an error and let them handle it as they'd like.

csabol-stripe: I mean debug only asserts so shipped sdks don't fail if there's a client secret format change 🙃 probably overly cautious

mshafrir-stripe: I like the idea of debug-only asserts. Interesting idea to consider.

PR #1469, opened 2020-01-02
Use STPAPIClient.shared by default

yuki-stripe:
## Summary
Make `STPPaymentContext`, `STPCustomerContext`, `STPPinManagementService`, `STPPushProvisioningContext`, `STPAddCardViewController`:
1. Use an explicit APIClient property to make API requests
2. Expose that property publicly
3. Default that property value to `STPAPIClient.shared`

Previously, they used various APIClient instances, leading to inconsistent and unclear configurations of APIClient properties like `stripeAccount`.

## Motivation
https://paper.dropbox.com/doc/psyduck-why-iOS-SDK-API-Configuration-LTvIMxj6CBy1Mw4ntaU0l#:uid=436909486618175148450143&h2=Proposal

https://jira.corp.stripe.com/browse/IOS-1437

## Testing
Manually test basic integration app and make sure customer things work.

### DIFF
--- a/Stripe/STPAPIClient.m
+++ b/Stripe/STPAPIClient.m
@@ -548,12 +548,11 @@ static NSArray<PKPaymentNetwork> *_additionalEnabledApplePayNetworks;
 
 @implementation STPAPIClient (Customers)
 
-+ (void)retrieveCustomerUsingKey:(STPEphemeralKey *)ephemeralKey completion:(STPCustomerCompletionBlock)completion {
-    STPAPIClient *client = [STPAPIClient sharedClient];
+- (void)retrieveCustomerUsingKey:(STPEphemeralKey *)ephemeralKey completion:(STPCustomerCompletionBlock)completion {
     NSString *endpoint = [NSString stringWithFormat:@"%@/%@", APIEndpointCustomers, ephemeralKey.customerID];
-    [STPAPIRequest<STPCustomer *> getWithAPIClient:client
+    [STPAPIRequest<STPCustomer *> getWithAPIClient:self
                                           endpoint:endpoint
-                                 additionalHeaders:[client authorizationHeaderUsingEphemeralKey:ephemeralKey]
+                                 additionalHeaders:[self authorizationHeaderUsingEphemeralKey:ephemeralKey]
                                         parameters:nil
                                       deserializer:[STPCustomer new]
                                         completion:^(STPCustomer *object, __unused NSHTTPURLResponse *response, NSError *error) {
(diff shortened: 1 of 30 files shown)


Stripe/STPAPIClient.m
```diff
@@ -163,7 +166,12 @@ - (NSMutableURLRequest *)configuredRequestForURL:(NSURL *)url {
     NSMutableDictionary *additionalHeaders = [NSMutableDictionary new];
     additionalHeaders[@"X-Stripe-User-Agent"] = [self.class stripeUserAgentDetailsWithAppInfo:self.appInfo];
     additionalHeaders[@"Stripe-Version"] = APIVersion;
-    additionalHeaders[@"Authorization"] = [@"Bearer " stringByAppendingString:self.apiKey ?: @""];
+    NSString *authorizationBearer = self.apiKey ?: @"";
+    if (self.useEphemeralKeyForNextRequest && self.ephemeralKey.secret) {
+        self.useEphemeralKeyForNextRequest = NO;
```
csabol-stripe: why is this reset after the first request?

yuki-stripe: the desired behavior is that we use the ephemeral key only when hitting the Customer API. 

There are a bunch of different ways to implement this and I have no fondness for this one though.

csabol-stripe: hmm, I don't like that this has the built-in assumption that this is coming from a short-lived instance. 

What do you think about adding `+ (NSURLSessionTask *)<action>WithAPIClient:endpoint:parameters:additionalRequestHeaders:(nullable NSDictionary *)...` methods?

csabol-stripe: instance meaning the shared client configured to use an ephemeral key. Then we would call the new STPAPIRequest methods from the customer STPAPIClient methods

yuki-stripe: Gave it a shot, I like this better.

yuki-stripe:
@csabol-stripe Ready for another pass.  I discovered 2 more classes that make API requests, both using ephemeral keys.  I think it makes sense that they use sharedClient too.  

Manually tested Basic Integration again.

yuki-stripe:
Expanded scope of PR after discussion, see revised summary.

Stripe/STPAPIRequest.h
```diff
@@ -21,20 +21,48 @@ typedef void(^STPAPIResponseBlock)(ResponseType object, NSHTTPURLResponse *respo
                                deserializer:(ResponseType)deserializer
                                  completion:(STPAPIResponseBlock)completion;
 
++ (NSURLSessionDataTask *)postWithAPIClient:(STPAPIClient *)apiClient
+                                   endpoint:(NSString *)endpoint
+                          additionalHeaders:(NSDictionary<NSString *, NSString *> *)additionalHeaders
```
csabol-stripe: I think the `additionalHeaders` param should be nullable for all of these

Stripe/STPAddCardViewController.m
```diff
@@ -91,7 +90,7 @@ - (void)commonInitWithConfiguration:(STPPaymentConfiguration *)configuration {
     _configuration = configuration;
     _shippingAddress = nil;
     _hasUsedShippingAddress = NO;
-    _apiClient = [[STPAPIClient alloc] initWithConfiguration:configuration];
+    _apiClient = [STPAPIClient sharedClient];
```
csabol-stripe: hmm, these changes are the most concerning to me. If `stripeAccount` (and other STPAPIClient configurations) is still a property on `STPPaymentConfiguration` this should probably respect those. Is the plan to deprecate on `STPPaymentConfiguration`?

yuki-stripe: Yes, plan is to kill `PaymentConfiguration.{stripeAccount, publishableKey}` in a follow-up PR

csabol-stripe: mind putting up that patch in parallel with this one so I can review both together?

PR #1472, opened 2020-01-10
Allow to hide CVC field

mpodeszwa:
## Summary

This PR adds possibility to hide CVC field on `STPPaymentCardTextField`.

## Motivation

I want to present `STPPaymentCardTextField` in read-only mode with prefilled data (for screen where user can edit address or other card's details). I can't present CVC of existing card so I just want to present masked credit card number with expiry date.

## Testing

I tested this code by modifying `CardFieldViewController` in UI Examples.

### DIFF
--- a/Stripe/PublicHeaders/STPPaymentCardTextField.h
+++ b/Stripe/PublicHeaders/STPPaymentCardTextField.h
@@ -206,6 +206,13 @@ The curent brand image displayed in the receiver.
 @property (nonatomic, readonly, nullable) NSString *postalCode;
 
 /**
+ Controls if a CVC entry field can be displayed to the user.
+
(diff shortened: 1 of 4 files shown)


csabol-stripe:
Hi @mpodeszwa! Thanks for the PR but I don't think this is something we can pull into master for now. In general, only the `last4` field should be stored in memory or displayed to the user.

mpodeszwa:
Hi @csabol-stripe 
I understand, I don't store full credit card number. I simply couldn't find a way to create `STPPaymentCardTextField` in read-only mode (field `last4` in `STPPaymentMethodCardParams` is read only). Do you have any idea how I could achieve it? :) In worst case I can still use my fork

PR #1474, opened 2020-01-15
Deprecate non-APIClient sources of publishableKey, stripeAccount

yuki-stripe:
## Summary
Part 2 of https://github.com/stripe/stripe-ios/pull/1469

1. Preserve `Stripe.setDefaultPublishableKey` behavior

This is the default value (in the programmatic sense) of publishable key for any new APIClient instance.  

2. Deprecate `STPPaymentConfiguration.{publishableKey, stripeAccount}`

For legacy reasons, make this a proxy for `STPAPIClient.shared.{publishableKey, stripeAccount}` if using the singleton.

3. Remove `apiKey` private property APIClient
4. Simplify the APIClient initializers by removing the dependency on PaymentConfiguration.  PaymentConfiguration is just another property in APIClient, instead of the sorta-source-of-truth of publishableKey.  It defaults to [PaymentConfiguration sharedConfiguration].

## Motivation

https://paper.dropbox.com/doc/psyduck-why-iOS-SDK-API-Configuration-LTvIMxj6CBy1Mw4ntaU0l#:uid=989210153385098517970797&h2=Risks

If users ignore the deprecation warnings, the updated behavior of these deprecated properties should result in no change for most (~everyone except for Connect users cloning PMs and direct charging their connected accounts).

## Testing
Existing tests still pass.

### DIFF
--- a/MIGRATING.md
+++ b/MIGRATING.md
@@ -15,7 +15,7 @@ You are affected by this change if:
 2. You use one of the above affected classes
 3. You are setting different `stripeAccount` values on `STPPaymentConfiguration` and `STPAPIClient`, i.e. `STPPaymentConfiguration.shared.stripeAccount != STPAPIClient.shared.stripeAccount`
 
-If these are all true, you must update your integration!  The SDK used to send the `STPPaymentConfiguration.shared.stripeAccount`, and will now send `STPAPIClient.shared.stripeAccount`.  
+If all three of the above conditions are true, you must update your integration!  The SDK used to send the `STPPaymentConfiguration.shared.stripeAccount`, and will now send `STPAPIClient.shared.stripeAccount`.  
 
 For example, if you are a Connect user who stores Payment Methods on your platform, but clones PaymentMethods to a connected count and create direct charges on that connected account ie if:
 
--- a/MIGRATING.md
+++ b/MIGRATING.md
@@ -17,7 +17,7 @@ You are affected by this change if:
 
 If all three of the above conditions are true, you must update your integration! The SDK used to send `STPPaymentConfiguration.shared.stripeAccount`, and will now send `STPAPIClient.shared.stripeAccount`.  
 
-For example, if you are a Connect user who stores Payment Methods on your platform, but clones PaymentMethods to a connected account and create direct charges on that connected account ie if:
+For example, if you are a Connect user who stores Payment Methods on your platform, but clones PaymentMethods to a connected account and creates direct charges on that connected account i.e. if:
 
 1. You never set `STPPaymentConfiguration.shared.stripeAccount`
 2. You set `STPAPIClient.shared.stripeAccount`
--- a/MIGRATING.md
+++ b/MIGRATING.md
@@ -11,7 +11,7 @@ This changes the behavior of the following classes, which previously used API cl
 
 You are affected by this change if:
 
-1. You set `stripeAccount`
+1. You use `stripeAccount` to work with your Connected accounts
 2. You use one of the above affected classes
 3. `STPPaymentConfiguration.shared.stripeAccount != STPAPIClient.shared.stripeAccount`
 
--- a/MIGRATING.md
+++ b/MIGRATING.md
@@ -1,5 +1,48 @@
 ## Migration Guides
 
+### Migrating from versions < 19.0.0
+
+* Deprecates `publishableKey` and `stripeAccount` properties of `STPPaymentConfiguration`. 
+  * If you used `STPPaymentConfiguration.sharedConfiguration` to set `publishableKey` and/or `stripeAccount`, use `STPAPIClient.sharedClient` instead. 
+  * If you passed a STPPaymentConfiguration instance to an SDK component, you should instead create an STPAPIClient, set publishableKey on it, and set the SDK component's APIClient property.
+* The SDK now uses `STPAPIClient.sharedClient` to make API requests by default.
+
+This changes the behavior of the following classes, which previously used API client instances configured from `STPPaymentConfiguration.shared`: `STPCustomerContext`, `STPPaymentOptionsViewController`, `STPAddCardViewController`, `STPPaymentContext`, `STPPinManagementService`, `STPPushProvisioningContext`. 
+
+You are affected by this change if:
+
+1. You set `stripeAccount`
+2. You use one of these classes
+3. `STPPaymentConfiguration.shared.stripeAccount != STPAPIClient.shared.stripeAccount`
+
+If these are all true, you must update your integration!  The SDK used to send the `STPPaymentConfiguration.shared.stripeAccount`, and will now send `STPAPIClient.shared.stripeAccount`.  
+
+For example, if you are a Connect user who stores Payment Methods on your platform, but clones PaymentMethods to a connected count and create direct charges on that connected account ie if:
+
+1. You never set `STPPaymentConfiguration.shared.stripeAccount`
+2. You set `STPAPIClient.shared.stripeAccount`
+
+We recommend you do the following:
+
+```
+    // By default, you don't want the SDK to pass stripeAccount
+    STPAPIClient.shared().publishableKey = "pk_platform"
+    STPAPIClient.shared().stripeAccount = nil
+
+    // You do want the SDK to pass stripeAccount when it makes payments directly on your connected account, so 
+    // you create a separate APIClient instance...
+    let connectedAccountAPIClient = STPAPIClient(publishableKey: "pk_platform")
+
+    // ...set stripeAccount on it...
+    connectedAccountAPIClient.stripeAccount = "your connected account's id"
+    
+    // ...and either set the relevant SDK components' apiClient property to your custom APIClient instance:
+    STPPaymentHandler.shared().apiClient = connectedAccountAPIClient // e.g. if you are using PaymentIntents
+
+    // ...or use it directly to make API requests with `stripeAccount` set:
+    connectedAccountAPIClient.createToken(withCard:...) // e.g. if you are using Tokens + Charges
+```
+
 ### Migrating from versions < 18.0.0
 * Some error messages from the Payment Intents API are now localized to the user's display language. If your application's logic depends on specific `message` strings from the Stripe API, please use the error [`code`](https://stripe.com/docs/error-codes) instead.
 * `STPPaymentResult` may contain a `paymentMethodParams` instead of a `paymentMethod` when using single-use payment methods such as FPX. Because of this, `STPPaymentResult.paymentMethod` is now nullable. Instead of setting the `paymentMethodId` manually on your `paymentIntentParams`, you may now call `paymentIntentParams.configure(with result: STPPaymentResult)`:
--- a/MIGRATING.md
+++ b/MIGRATING.md
@@ -13,11 +13,11 @@ You are affected by this change if:
 
 1. You use `stripeAccount` to work with your Connected accounts
 2. You use one of the above affected classes
-3. You are setting different `stripeAccount` values on `STPPaymentConfiguration` and `STPAPIClient`, i.e. `STPPaymentConfiguration.shared.stripeAccount != STPAPIClient.shared.stripeAccount`
+3. You set different `stripeAccount` values on `STPPaymentConfiguration` and `STPAPIClient`, i.e. `STPPaymentConfiguration.shared.stripeAccount != STPAPIClient.shared.stripeAccount`
 
-If all three of the above conditions are true, you must update your integration!  The SDK used to send the `STPPaymentConfiguration.shared.stripeAccount`, and will now send `STPAPIClient.shared.stripeAccount`.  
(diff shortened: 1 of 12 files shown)


Stripe/PublicHeaders/STPAPIClient.h
```diff
@@ -30,10 +30,12 @@ static NSString *const STPSDKVersion = @"18.3.0";
 @interface Stripe : NSObject FAUXPAS_IGNORED_ON_LINE(UnprefixedClass);
 
 /**
- Set your Stripe API key with this method. New instances of STPAPIClient will be initialized with this value. You should call this method as early as
```
csabol-stripe: why did this behavior change?

Stripe/PublicHeaders/STPAPIClient.h
```diff
@@ -30,10 +30,12 @@ static NSString *const STPSDKVersion = @"18.3.0";
...
  possible in your application's lifecycle, preferably in your AppDelegate.
-
- @param   publishableKey Your publishable key, obtained from https://stripe.com/account/apikeys
+ 
+ This is the same as setting STPAPIClient.shared.publishableKey.
+ 
+ @param   publishableKey Your publishable key, obtained from https://dashboard.stripe.com/apikeys
  @warning Make sure not to ship your test API keys to the App Store! This will log a warning if you use your test key in a release build.
  */
 + (void)setDefaultPublishableKey:(NSString *)publishableKey;
```
csabol-stripe: since we're documenting to use `STPAPIClient.shared.publishableKey` instead should this just be deprecated?

Stripe/STPAPIClient.m
```diff
@@ -115,33 +113,34 @@ + (instancetype)sharedClient {
...
     return self;
 }
 
+- (instancetype)initWithPublishableKey:(NSString *)publishableKey {
+    STPAPIClient *apiClient = [self init];
+    apiClient.publishableKey = publishableKey;
+    return apiClient;
+}
+
+- (instancetype)initWithConfiguration:(STPPaymentConfiguration *)configuration {
```
csabol-stripe: since all the behavior here is deprecated, should this initializer be deprecated as well?

Stripe/PublicHeaders/STPAPIClient.h
yuki-stripe: Hmm. I didn't deliberately keep this behavior, but I don't see why we shouldn't, and it will be less breaking.  Will re-add.

Stripe/STPAPIClient.m
yuki-stripe: I think that's fair.  I wouldn't expect many users of this, anyways.

Stripe/PublicHeaders/STPAPIClient.h
yuki-stripe: I leaned towards keeping it b/c:
- less breaking for users (esp with the above comment's change)
- no need to update docs recommendation
- `Stripe.setDefaultPublishableKey` reads better than `STPAPIClient.shared.publishableKey`

Stripe/STPAPIClient.m
```diff
@@ -115,33 +116,35 @@ + (instancetype)sharedClient {
...
     if (self) {
-        _apiKey = publishableKey;
         _apiURL = [NSURL URLWithString:APIBaseURL];
-        _configuration = configuration;
-        _stripeAccount = configuration.stripeAccount;
+        _configuration = [STPPaymentConfiguration sharedConfiguration];
         _sourcePollers = [NSMutableDictionary dictionary];
         _sourcePollersQueue = dispatch_queue_create("com.stripe.sourcepollers", DISPATCH_QUEUE_SERIAL);
         _urlSession = [NSURLSession sessionWithConfiguration:[self.class sharedUrlSessionConfiguration]];
+        _publishableKey = [Stripe defaultPublishableKey];
```
csabol-stripe: copy

csabol-stripe: actually wouldn't we not want to do this so we return the static _defaultPublishableKey?

Stripe/STPAPIClient+ApplePay.m
```diff
@@ -126,7 +126,7 @@ + (NSDictionary *)parametersForPayment:(PKPayment *)payment {
     payload[@"pk_token"] = paymentString;
     payload[@"card"] = [self addressParamsFromPKContact:payment.billingContact];
 
-    NSCAssert(!(paymentString.length == 0 && [[Stripe defaultPublishableKey] hasPrefix:@"pk_live"]), @"The pk_token is empty. Using Apple Pay with an iOS Simulator while not in Stripe Test Mode will always fail.");
+    NSCAssert(!(paymentString.length == 0 && [[STPAPIClient sharedClient].publishableKey hasPrefix:@"pk_live"]), @"The pk_token is empty. Using Apple Pay with an iOS Simulator while not in Stripe Test Mode will always fail.");
```
csabol-stripe: I think this should still reference [Stripe defaultPublishableKey]`

Stripe/STPAnalyticsClient.m
```diff
@@ -410,7 +410,7 @@ + (NSMutableDictionary *)commonPayload {
 
 + (NSDictionary *)serializeConfiguration:(STPPaymentConfiguration *)configuration {
     NSMutableDictionary *dictionary = [NSMutableDictionary dictionary];
-    dictionary[@"publishable_key"] = configuration.publishableKey ?: @"unknown";
+    dictionary[@"publishable_key"] = [STPAPIClient sharedClient].publishableKey ?: @"unknown";
```
csabol-stripe: I think this should be [Stripe defulatPublishableKey]

Stripe/STPAPIClient.m
```diff
@@ -69,14 +69,14 @@
...
 
 + (void)setDefaultPublishableKey:(NSString *)publishableKey {
-    [STPAPIClient validateKey:publishableKey];
-    [STPPaymentConfiguration sharedConfiguration].publishableKey = publishableKey;
+    _defaultPublishableKey = [publishableKey copy];
 }
 
 + (NSString *)defaultPublishableKey {
-    return [STPPaymentConfiguration sharedConfiguration].publishableKey;
+    return _defaultPublishableKey;
```
csabol-stripe: I might actually go so far as to make this non-nil (initial value of empty string) so that we can have true `null_resettable` behavior in STPAPIClient. Thoughts @davidme-stripe ?

Stripe/STPAPIClient.m
yuki-stripe: Good call, added a test to catch this.

Stripe/STPAPIClient+ApplePay.m
yuki-stripe: As it stands, `STPAPIClient sharedClient` is still more likely to be accurate of the two, since if they differ, it means the user set `STPAPIClient.sharedClient.publishableKey`.

e.g. You could never call `Stripe.setDefaultPublishableKey` and only set `STPAPIClient.sharedClient.publishableKey`.

Stripe/STPAPIClient.m
yuki-stripe: It's a bit weird to have that as an explicit default value, because an empty string isn't a valid value. Like, we fire an assertion if you try to set publishableKey to an empty string.

Stripe/STPAnalyticsClient.m
yuki-stripe: As it stands, STPAPIClient sharedClient is still more likely to be accurate of the two, since if they differ, it means the user set STPAPIClient.sharedClient.publishableKey.

e.g. You could never call Stripe.setDefaultPublishableKey and only set STPAPIClient.sharedClient.publishableKey.

csabol-stripe: mm, I have the opposite understanding. You should _always_ call `stripe.setDefaultPublishableKey` and only set `STPAPIClient.sharedClient.publishableKey` for very specific use cases wher eyou know what you're doing. Otherwise what's the point of Stripe.defaultPublishableKey?

MIGRATING.md
```diff
@@ -1,5 +1,48 @@
...
+  * If you used `STPPaymentConfiguration.sharedConfiguration` to set `publishableKey` and/or `stripeAccount`, use `STPAPIClient.sharedClient` instead. 
+  * If you passed a STPPaymentConfiguration instance to an SDK component, you should instead create an STPAPIClient, set publishableKey on it, and set the SDK component's APIClient property.
+* The SDK now uses `STPAPIClient.sharedClient` to make API requests by default.
+
+This changes the behavior of the following classes, which previously used API client instances configured from `STPPaymentConfiguration.shared`: `STPCustomerContext`, `STPPaymentOptionsViewController`, `STPAddCardViewController`, `STPPaymentContext`, `STPPinManagementService`, `STPPushProvisioningContext`. 
+
+You are affected by this change if:
+
+1. You set `stripeAccount`
+2. You use one of these classes
```
csabol-stripe:
```suggestion
2. You use one of the above affected classes
```

MIGRATING.md
```diff
@@ -1,5 +1,48 @@
...
+* Deprecates `publishableKey` and `stripeAccount` properties of `STPPaymentConfiguration`. 
+  * If you used `STPPaymentConfiguration.sharedConfiguration` to set `publishableKey` and/or `stripeAccount`, use `STPAPIClient.sharedClient` instead. 
+  * If you passed a STPPaymentConfiguration instance to an SDK component, you should instead create an STPAPIClient, set publishableKey on it, and set the SDK component's APIClient property.
+* The SDK now uses `STPAPIClient.sharedClient` to make API requests by default.
+
+This changes the behavior of the following classes, which previously used API client instances configured from `STPPaymentConfiguration.shared`: `STPCustomerContext`, `STPPaymentOptionsViewController`, `STPAddCardViewController`, `STPPaymentContext`, `STPPinManagementService`, `STPPushProvisioningContext`. 
+
+You are affected by this change if:
+
+1. You set `stripeAccount`
```
csabol-stripe:
```suggestion
1. You use `stripeAccount` to work with your Connected accounts
```

MIGRATING.md
```diff
@@ -1,5 +1,48 @@
...
+2. You use one of these classes
+3. `STPPaymentConfiguration.shared.stripeAccount != STPAPIClient.shared.stripeAccount`
```
csabol-stripe:
```suggestion
3. You are setting different `stripeAccount` values on `STPPaymentConfiguration` and `STPAPIClient`, i.e. `STPPaymentConfiguration.shared.stripeAccount != STPAPIClient.shared.stripeAccount`
```

MIGRATING.md
```diff
@@ -1,5 +1,48 @@
...
+
+If these are all true, you must update your integration!  The SDK used to send the `STPPaymentConfiguration.shared.stripeAccount`, and will now send `STPAPIClient.shared.stripeAccount`.
```
csabol-stripe:
```suggestion
If all three of the above conditions are true, you must update your integration!  The SDK used to send the `STPPaymentConfiguration.shared.stripeAccount`, and will now send `STPAPIClient.shared.stripeAccount`.  
```

MIGRATING.md
```diff
@@ -1,5 +1,48 @@
...
+
+If these are all true, you must update your integration!  The SDK used to send the `STPPaymentConfiguration.shared.stripeAccount`, and will now send `STPAPIClient.shared.stripeAccount`.  
+
+For example, if you are a Connect user who stores Payment Methods on your platform, but clones PaymentMethods to a connected count and create direct charges on that connected account ie if:
+
+1. You never set `STPPaymentConfiguration.shared.stripeAccount`
+2. You set `STPAPIClient.shared.stripeAccount`
+
+We recommend you do the following:
+
```
csabol-stripe: Maybe rewrite as:

For example, if you only want to use a Connected account for a single transaction, your integration could look like:

yuki-stripe: That sounds different from the use case described.  These users want to use a Connected account for all transactions, if transactions means payments, but they want to retrieve Customers and their PMs on the platform.

csabol-stripe: okay, as written it was confusing to me but I think you understand the use case better :)

MIGRATING.md
```diff
@@ -1,5 +1,48 @@
...
+You are affected by this change if:
+
+1. You use `stripeAccount` to work with your Connected accounts
+2. You use one of the above affected classes
+3. You set different `stripeAccount` values on `STPPaymentConfiguration` and `STPAPIClient`, i.e. `STPPaymentConfiguration.shared.stripeAccount != STPAPIClient.shared.stripeAccount`
+
+If all three of the above conditions are true, you must update your integration! The SDK used to send `STPPaymentConfiguration.shared.stripeAccount`, and will now send `STPAPIClient.shared.stripeAccount`.  
+
+For example, if you are a Connect user who stores Payment Methods on your platform, but clones PaymentMethods to a connected account and create direct charges on that connected account ie if:
```
csabol-stripe:
```suggestion
For example, if you are a Connect user who stores Payment Methods on your platform, but clones PaymentMethods to a connected account and create direct charges on that connected account i.e. if:
```

Issue #1475, opened 2020-01-15
Setting defaultPaymentMethod in STPPaymentOptionsViewController has no effect


jose-t:
## Summary
Setting `defaultPaymentMethod` from `STPPaymentOptionsViewController` seems to have no effect. The pre-selected payment method is always fetched from NSUserDefault.

## Code to reproduce
```
let paymentOptionsViewController = STPPaymentOptionsViewController(configuration: config, theme: STPTheme.default(), customerContext: customerContext, delegate: delegate)
paymentOptionsViewController.defaultPaymentMethod = defaultPaymentMethod
```

## iOS version
13.2

## Installation method
CocoaPods

## SDK version
18.3.0

## Other information
It works fine when I use `STPPaymentContext` to initialize the `STPPaymentOptionsViewController`.

davidme-stripe:
Thanks for the report, we're working on a fix.

yuki-stripe:
@jose-t This should be merged into master, and you can expect a release with this fix today!

I'm curious to learn more about your use case - why does the default behavior not work for you?

jose-t:
It was fast, thanks you !

My application use the Stripe SDK on both iOS and Android. The user can add/remove cards and select his default payment method from the `STPPaymentOptionsController` and `PaymentMethodsActivity`.

Suppose the same user has two phones on iOS and Android, if he selects a default payment method on iOS, it will be registered in `NSUserDefaults`. Now he opens his Android phone and he select a new default payment method. If he reopens his iPhone and go back to the `STPPaymentOptionsController`, the pre-selected payment method will be the same stored in `NSUserDefaults` and not the default payment method stored on the stripe Customer object.

I think it could be dangerous to keep the default payment method state locally because it does not represent the real state.

yuki-stripe:
Aha makes perfect sense, thank you!  This is helpful information as we plan future improvements here.

davidme-stripe:
We released this in SDK v18.4.0.

PR #1476, opened 2020-01-15
Fix PaymentOptionsVC race condition where payment method loading race…

yuki-stripe:
…s against setting defaultPaymentMethod.  

We end up checking `defaultPaymentMethod` before initialization returns, leaving no room for the user to set the property.

This happens b/c
- We make a network request for payment methods in init
- That request can call the completion block synchronously if cached
- We don't clear the cache in PaymentOptionsVC (unlike PaymentContext)

I find all of this surprising but to avoid breaking other things, I'm leaving all of these things as-is and making the smallest fix.

## Motivation
https://github.com/stripe/stripe-ios/issues/1475

## Testing
Tested on UI Examples app, which always synchronously returns. Repro'd pre-fix, cant repro post-fix.


davidme-stripe:
A note from our offline chat about this: `stpDispatchToMainThreadIfNecessary` is kinda dangerous, as it isn't immediately obvious to the user whether it will execute immediately or on the next runloop. Changing `stpDispatchToMainThreadIfNecessary` to always call `dispatch_async` could cause side effects for existing usage of the SDK, so we'll leave it alone for now, but we should probably reduce our usage of it.

PR #1479, opened 2020-01-23
Update postal code logic

davidme-stripe:
## Summary
* Request postal codes by default in most situations. This is not a breaking change to our public API, though it may cause trouble if folks are messing with our internals.
* When `STPBillingAddressFieldsZip` is selected, `STPAddCardViewController` will now embed the postal code collection into the `STPPaymentCardTextField` instead of splitting it out into a separate row.
* Postal codes will now be validated for non-US markets. We're assuming that they're always uppercase, and include only alphanumeric characters, dashes, and spaces. These rules will also apply in the US, as we can't be 100% sure that a US region user will have a US ZIP code. Shipping/Billing address entry forms will still validate and enforce US-style ZIP codes for US addresses.
* All countries will use the NumbersAndPunctuation keyboard.
* STPBillingAddressFieldsZip has been deprecated in favor of STPBillingAddressFieldsPostalCode, which is identical. I'm not sure whether this is worth the churn, but it didn't seem appropriate to use "ZIP" now that it's the default option globally.

## Motivation
We're now recommending that all users in the US collect postal codes.

## Testing
* On a device, switching between a few regions and exercising different `STPBillingAddressFields` settings.
* I've also added a new set of end-to-end tests using a French device in Belize, covering both the "different language" and "doesn't use postal codes" case.


Stripe/NSCharacterSet+Stripe.m
```diff
@@ -28,6 +28,24 @@ + (instancetype)stp_invertedAsciiDigitCharacterSet {
     return cs;
 }
 
++ (instancetype)stp_postalCodeCharacterSet {
+    static NSCharacterSet *cs;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        cs = [self characterSetWithCharactersInString:@"abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789- "];
```
csabol-stripe: just to confirm we're confident this shouldn't include, for example, Chinese characters?

Stripe/STPAddCardViewController.m
```diff
@@ -279,10 +284,16 @@ - (void)nextPressed:(__unused id)sender {
...
-    billingDetails.address = [[STPPaymentMethodAddress alloc] initWithAddress:self.addressViewModel.address];
-    billingDetails.email = self.addressViewModel.address.email;
-    billingDetails.name = self.addressViewModel.address.name;
-    billingDetails.phone = self.addressViewModel.address.phone;
+    if (self.configuration.requiredBillingAddressFields == STPBillingAddressFieldsPostalCode) {
+        STPAddress *address = [[STPAddress alloc] init];
+        address.postalCode = self.paymentCell.paymentField.postalCode;
+        billingDetails.address = [[STPPaymentMethodAddress alloc] initWithAddress:address];
+    } else {
+        billingDetails.address = [[STPPaymentMethodAddress alloc] initWithAddress:self.addressViewModel.address];
```
csabol-stripe: does this include the postal code?

Stripe/STPAddressFieldTableViewCell.m
```diff
@@ -185,38 +175,31 @@ - (void)updateTextFieldsAndCaptions {
...
-            
-            if (!self.lastInList) {
-                self.textField.inputAccessoryView = self.inputAccessoryToolbar;
-            } else {
-                self.textField.inputAccessoryView = nil;
-            }
             break;
         case STPAddressFieldTypePhone:
-            self.textField.keyboardType = UIKeyboardTypePhonePad;
+            self.textField.keyboardType = UIKeyboardTypeNumbersAndPunctuation;
```
csabol-stripe: is even `PhonePad` broken for autofill?

Stripe/NSCharacterSet+Stripe.m
davidme-stripe: I did check China (they only use numbers) and have yet to find a country that uses characters outside of this set.

Stripe/STPAddCardViewController.m
davidme-stripe: It does: The STPAddressViewModel handles the ZIP when using the full address entry form, but the STPPaymentCardTextFieldViewModel handles it when we're only requesting the ZIP. It's not the cleanest implementation, but the UX is better than breaking out ZIP into a separate field below the CC entry field.

Stripe/STPAddressFieldTableViewCell.m
davidme-stripe: Yes. :(

davidme-stripe:
@csabol-stripe How's this? https://github.com/stripe/stripe-ios/blob/davidme/update-zip-logic/MIGRATING.md#migrating-from-versions--1900

PR #1482, opened 2020-01-29
Support pasting in four-digit expiration years

davidme-stripe:
## Summary
Pasting in expiration dates in the format of "12/2024" (from 1Password, for example) enters "12/20" in the expiration date field instead of the expected "12/24." Stripe Elements handles this correctly, so we'll do something similar here.

## Testing
Wrote a unit test for the formatter and tested by pasting from 1Password on my phone.

### DIFF
--- a/Tests/Tests/STPStringUtilsTest.m
+++ b/Tests/Tests/STPStringUtilsTest.m
@@ -77,5 +77,20 @@
                                }];
 }
 
+- (void)testExpirationDateStrings {
+    XCTAssertEqualObjects([STPStringUtils expirationDateStringFromString:@"12/1995"], @"12/95");
+    XCTAssertEqualObjects([STPStringUtils expirationDateStringFromString:@"12 / 1995"], @"12 / 95");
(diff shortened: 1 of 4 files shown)


Tests/Tests/STPStringUtilsTest.m
```diff
@@ -77,5 +77,20 @@ - (void)testParseRangesMultiTag2 {
                                }];
 }
 
+- (void)testExpirationDateStrings {
+    XCTAssertEqualObjects([STPStringUtils expirationDateStringFromString:@"12/1995"], @"12/95");
+    XCTAssertEqualObjects([STPStringUtils expirationDateStringFromString:@"12 / 1995"], @"12 / 95");
+    XCTAssertEqualObjects([STPStringUtils expirationDateStringFromString:@"12 /1995"], @"12 /95");
+    XCTAssertEqualObjects([STPStringUtils expirationDateStringFromString:@"1295"], @"1295");
+    XCTAssertEqualObjects([STPStringUtils expirationDateStringFromString:@"08/2001"], @"08/01");
```
yuki-stripe: could throw in something like "12/95" for good measure

PR #1485, opened 2020-02-03
Fix STPPaymentMethodiDEAL naming on case-sensitive file systems

glyuck:
## Summary
Fix library to compile on case-sensitive file systems.

## Motivation
Here is the error from xcodebuild output:
```
error: Build input file cannot be found: '/Users/glyuck/workspace/myproject/Carthage/Checkouts/stripe-ios/Stripe/STPPaymentMethodiDEALParams.m' (in target 'StripeiOS' from project 'Stripe')
```

## Testing
I've only ensured that the code now compiles on my case-sensitive file system, no additional testing performed.


davidme-stripe:
Thank you! I do appreciate that you submitted this patch, though I think we'd rather keep the filename as `STPPaymentMethodiDEAL.m`. I've opened #1491 to fix this.

PR #1492, opened 2020-02-13
Send payment_method_type analytic when you attempt to create a PM, co…

yuki-stripe:
…nfirm a PI, confirm an SI

## Summary
Create a PM, confirm a PI, confirm an SI all send `payment_method_type` now.
Confirm PI no longer sends `bancontact` as `source_type` when there is no source.

## Motivation
https://jira.corp.stripe.com/browse/IOS-1610

## Testing
Ran basic integration example, created a PM and confirmed a PI.  Saw the expected analytics.

### DIFF
--- a/Stripe/STPAPIClient.m
+++ b/Stripe/STPAPIClient.m
@@ -743,9 +743,11 @@ toCustomerUsingKey:(STPEphemeralKey *)ephemeralKey
     NSCAssert([STPPaymentIntentParams isClientSecretValid:paymentIntentParams.clientSecret], @"`paymentIntentParams.clientSecret` format does not match expected client secret formatting.");
 
     NSString *identifier = paymentIntentParams.stripeId;
-    NSString *sourceType = [STPSource stringFromType:paymentIntentParams.sourceParams.type];
+    NSString *sourceType = paymentIntentParams.sourceParams.rawTypeString;
+    NSString *paymentMethodType = paymentIntentParams.paymentMethodParams.rawTypeString;
     [[STPAnalyticsClient sharedClient] logPaymentIntentConfirmationAttemptWithConfiguration:self.configuration
-                                                                                 sourceType:sourceType];
+                                                                                 sourceType:sourceType
+                                                                          paymentMethodType:paymentMethodType];
 
     NSString *endpoint = [NSString stringWithFormat:@"%@/%@/confirm", APIEndpointPaymentIntents, identifier];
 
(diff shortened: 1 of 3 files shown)


Stripe/STPAPIClient.m
```diff
@@ -847,14 +846,13 @@ - (void)createPaymentMethodWithParams:(STPPaymentMethodParams *)paymentMethodPar
...
     NSCAssert(paymentMethodParams != nil, @"'paymentMethodParams' is required to create a PaymentMethod");
     NSCAssert(paymentMethodParams.rawTypeString != nil, @"Set the `type` or `rawTypeString` property on paymentMethodParams.");
+    [[STPAnalyticsClient sharedClient] logPaymentMethodCreationAttemptWithConfiguration:self.configuration paymentMethodType:paymentMethodParams.rawTypeString];
+    
     [STPAPIRequest<STPPaymentMethod *> postWithAPIClient:self
                                                endpoint:APIEndpointPaymentMethods
                                              parameters:[STPFormEncoder dictionaryForObject:paymentMethodParams]
                                            deserializer:[STPPaymentMethod new]
                                              completion:^(STPPaymentMethod *paymentMethod, __unused NSHTTPURLResponse *response, NSError *error) {
-                                                 if (error == nil && paymentMethod != nil) {
```
yuki-stripe: I don't know why this didn't match the other analytics, which fire regardless of success.

Stripe/STPAnalyticsClient.m
```diff
@@ -252,7 +252,7 @@ - (void)logSetupIntentConfirmationAttemptWithConfiguration:(STPPaymentConfigurat
     NSMutableDictionary *payload = [self.class commonPayload];
     [payload addEntriesFromDictionary:@{
                                         @"event": @"stripeios.setup_intent_confirmation",
-                                        @"payment_method_type": paymentMethodType ?: @"unknown",
+                                        @"source_type": paymentMethodType ?: @"unknown",
```
yuki-stripe: deliberately overloading `source_type` here like android does.

mshafrir-stripe: We could also pass this as `payment_method_type` from the client and figure out what to do with it on the backend. For this PR maybe simpler to go with what you have.

Stripe/STPAPIClient.m
yuki-stripe: @csabol-stripe Can you confirm I'm not clobbering something intentional here?

Stripe/STPAnalyticsClient.m
```diff
@@ -218,13 +218,13 @@ - (void)logSourceCreationAttemptWithConfiguration:(STPPaymentConfiguration *)con
...
-- (void)logPaymentMethodCreationSucceededWithConfiguration:(STPPaymentConfiguration *)configuration
-                                           paymentMethodID:(NSString *)paymentMethodID {
+- (void)logPaymentMethodCreationAttemptWithConfiguration:(STPPaymentConfiguration *)configuration
+                                       paymentMethodType:(NSString *)paymentMethodType {
     NSDictionary *configurationDictionary = [self.class serializeConfiguration:configuration];
     NSMutableDictionary *payload = [self.class commonPayload];
     [payload addEntriesFromDictionary:@{
                                         @"event": @"stripeios.payment_method_creation",
-                                        @"payment_method_id": paymentMethodID,
+                                        @"source_type": paymentMethodType,
```
yuki-stripe: @csabol-stripe mind looking at this one too?

Issue #973, opened 2018-06-22
Standard integration Apple Pay issue after user cancelation


pierrem:
## Summary

I'm using Stripe (13.0.2) and I'm testing on an iPhone simulator

The user selects Apple Pay 
I call 	[self.paymentContext requestPayment];
The Apple Pay payment controller is presented
I tap on the cancel button
paymentContext:didFinishWithStatus:error: is called with status STPPaymentStatusUserCancellation 
and the Apple Pay payment controller is dismissed

Next time I call [self.paymentContext requestPayment]
paymentContext:didFinishWithStatus:error: is called immediately with STPPaymentStatusUserCancellation status
And the Apple Pay payment controller is not presented

2018-06-22 18:09:03.928227+0200 DermTel BETA[17368:9730145] Connection to remote alert view service failed
2018-06-22 18:09:08.670170+0200 DermTel BETA[17368:9730145] Payment canceled
2018-06-22 18:09:08.670552+0200 DermTel BETA[17368:9735927] Failed to present in-app payment interface: Error Domain=PKPassKitErrorDomain Code=-1 "Payment interface not available" UserInfo={NSLocalizedDescription=Payment interface not available}

And next time I call [self.paymentContext requestPayment], everything is ok !

## iOS version
11.4

## Installation method
pod

## SDK version
13.0.2

## Other information
Tested on simulator

csabol-stripe:
Thanks for the report @pierrem We're looking at our ApplePay integration right now with #968 and will be sure to cover this issue as well!

PR #1499, opened 2020-02-21
STPApplePayContext

yuki-stripe:
## Summary
A helper class that encapsulates `PKPaymentAuthorizationViewController`/ `PKPaymentAuthorizationViewControllerDelegate` and implements the boilerplate, state management, and tricky cancel/timeout logic for the user.

TODO in future PR
[] Use this in PaymentContext 

Handles:
* Automatic confirmation
* Manual confirmation
* Server-side confirmation (either auto or manual)
* Shipping
* Cancel/timeout (will solve https://github.com/stripe/stripe-ios/issues/968, https://github.com/stripe/stripe-ios/issues/973 when we use this in PaymentContext)

Doesn't handle:
* SetupIntents 
* If you want to respond to other `PKPaymentAuthorizationViewControllerDelegate` methods e.g. add a surcharge in `didSelectPaymentMethod`

## Motivation
https://paper.dropbox.com/doc/iOS-SDK-Apple-Pay-Wrapper-Proposal--Auyzacgn1O3XTdFO4U8DepqCAg-ter8eunlOgf9cEw9szQxm

## Testing
See unit and functional tests


Stripe/PublicHeaders/STPApplePayContext.h
```diff
@@ -0,0 +1,132 @@
...
+- (void)applePayContext:(STPApplePayContext *)context
+didSelectShippingContact:(PKContact *)contact
+                handler:(void (^)(PKPaymentRequestShippingContactUpdate *update))completion API_AVAILABLE(ios(11.0), watchos(4.0));
+
+/**
+ A pre-iOS 11 version of paymentAuthorizationViewController:didSelectShippingContact:handler:
+ */
+- (void)applePayContext:(STPApplePayContext *)context
+didSelectShippingContact:(PKContact *)contact
+             completion:(void (^)(PKPaymentAuthorizationStatus status, NSArray<PKShippingMethod *> *shippingMethods,
```
csabol-stripe: we're inconsistent on completion/handler in these delegate methods. I think the argument name and label should match, i.e. `completion:(type)completion` or `handler:(type)handler`. I'd vote for `completion` but am fine with either

Stripe/PublicHeaders/STPApplePayContext.h
```diff
@@ -0,0 +1,132 @@
...
+
+/**
+ Use initWithPaymentRequest:delegate: instead.
+ */
+- (instancetype)init NS_UNAVAILABLE;
+
+/**
+ Presents the Apple Pay sheet, starting the payment process.
+ 
+ @note This method should only be called once; create a new instance every time you present Apple Pay.
```
csabol-stripe: create a new instance of STPApplePayContext?

Stripe/STPApplePayContext.m
```diff
@@ -0,0 +1,283 @@
...
+
+#import "STPAPIClient+ApplePay.h"
+#import "STPPaymentMethod.h"
+#import "STPPaymentIntentParams.h"
+#import "STPPaymentIntent+Private.h"
+#import "STPPaymentHandler.h"
+#import "NSError+Stripe.h"
+
+static char kSTPApplePayContextAssociatedObjectKey;
+
```
csabol-stripe: we should add `NS_ASSUME_NONNULL_BEGIN/END` to .m files as well

Stripe/STPApplePayContext.m
```diff
@@ -0,0 +1,283 @@
...
+    
+    // This instance must live so that the apple pay sheet is dismissed; until then, the app is effectively frozen.
+    objc_setAssociatedObject(self.viewController, &kSTPApplePayContextAssociatedObjectKey, self, OBJC_ASSOCIATION_RETAIN_NONATOMIC);
+
+    [viewController presentViewController:self.viewController animated:YES completion:completion];
+}
+
+- (void)forwardInvocation:(NSInvocation *)invocation {
+    // Called for the methods we additionally respond YES to in `respondsToSelector`, letting us forward directly to self.delegate
+    // We could alternatively implement the PKPaymentAuthorizationViewControllerDelegate methods to call their respective STPApplePayContextDelegate methods
```
csabol-stripe: hmm, I think I'd actually prefer this approach. It seems less complicated and would also respond to delegate method changes better (e.g. if one gets marked as deprecated, etc.)

Stripe/STPApplePayContext.m
```diff
@@ -0,0 +1,283 @@
...
+                if (self.didCancelOrTimeoutWhilePending) {
+                    [self.viewController dismissViewControllerAnimated:YES completion:^{
+                        [self.delegate applePayContext:self didCompleteWithStatus:STPPaymentStatusSuccess error:nil];
+                        [self _end];
+                    }];
+                } else {
+                    completion(PKPaymentAuthorizationStatusSuccess, nil);
+                }
+                return;
+            default:
```
csabol-stripe: can we list out the other values here so compiler complains if they change?

Stripe/STPApplePayContext.m
```diff
@@ -0,0 +1,283 @@
...
+                        [self _end];
+                    }];
+                } else {
+                    completion(PKPaymentAuthorizationStatusFailure, error);
+                }
+                return;
+            case STPPaymentStateSuccess:
+                self.paymentState = STPPaymentStateSuccess;
+                
+                if (self.didCancelOrTimeoutWhilePending) {
```
csabol-stripe: how would we get into this state? success + canceled or timed out

Stripe/STPApplePayContext.m
```diff
@@ -0,0 +1,283 @@
...
+                }
+                return;
+            default:
+                NSAssert(NO, @"Invalid final state");
+                return;
+        }
+    };
+    
+    // 1. Create PaymentMethod
+    [[STPAPIClient sharedClient] createPaymentMethodWithPayment:payment completion:^(STPPaymentMethod *paymentMethod, NSError *paymentMethodCreationError) {
```
csabol-stripe:
```suggestion
    [self.apiClient createPaymentMethodWithPayment:payment completion:^(STPPaymentMethod *paymentMethod, NSError *paymentMethodCreationError) {
```

Stripe/STPApplePayContext.m
```diff
@@ -0,0 +1,283 @@
...
+                        } else {
+                            handleFinalState(STPPaymentStateError, confirmError);
+                        }
+                    }];
+                } else if (paymentIntent.status == STPPaymentIntentStatusSucceeded || paymentIntent.status == STPPaymentIntentStatusRequiresCapture) {
+                    handleFinalState(STPPaymentStateSuccess, nil);
+                } else {
+                    NSDictionary *userInfo = @{
+                        NSLocalizedDescriptionKey: [NSError stp_unexpectedErrorMessage],
+                        STPErrorMessageKey: @"The PaymentIntent is in an unexpected state. If you pass confirmation_method = manual when creating the PaymentIntent, also pass confirm = true.  If server-side confirmation fails, double check you passing the error back to the client."
```
csabol-stripe:
```suggestion
                        STPErrorMessageKey: @"The PaymentIntent is in an unexpected state. If you pass confirmation_method = manual when creating the PaymentIntent, also pass confirm = true.  If server-side confirmation fails, double check you are passing the error back to the client."
```

Stripe/STPApplePayContext.m
csabol-stripe: 👍

Stripe/PublicHeaders/STPApplePayContext.h
csabol-stripe: weird. I'd vote for changing to `handler:(type)handler` and `completion:(type)completion` but won't block

PR #1507, opened 2020-03-02
Use UOB0228 as the B2B bank for UOB

joelchoo-stripe:
## Summary
This is a follow up to https://github.com/stripe/stripe-ios/pull/1496 - we now want to set the B2B bank for UOB to UOB0228.

## Testing
Existing tests


joelchoo-stripe:
@davidme-stripe bumping this!

PR #1508, opened 2020-03-02
Adds dynamic type support to Basic Integration

csabol-stripe:
## Summary

STPTheme scales default font values.
Rewrites STPAddressFieldTableViewCell to use autolayout for correct autosizing.
Fixes missing theme propagation to subviews.

## Motivation

Accessibility fixes.

## Testing

Tested changing text size accesisbility setting on simulator with Basic Integration example app. Snapshot tests not regressed.


Stripe/STPTheme.m
```diff
@@ -201,8 +217,8 @@ - (id)copyWithZone:(__unused NSZone *)zone {
     copyTheme.secondaryForegroundColor = self.secondaryForegroundColor;
     copyTheme.accentColor = self.accentColor;
     copyTheme.errorColor = self.errorColor;
-    copyTheme.font = self.font;
-    copyTheme.emphasisFont = self.emphasisFont;
+    copyTheme->_font = _font;
+    copyTheme->_emphasisFont = _emphasisFont;
```
yuki-stripe: I was gonna ask to use dot-access to set these but I think that would break the dynamic getter?  

I guess we should `copy` these since that's what the property attributes say.

Stripe/STPTheme.m
```diff
@@ -159,11 +157,29 @@ - (UIColor *)errorColor {
...
 }
 
 - (UIFont *)emphasisFont {
-    return _emphasisFont ?: STPThemeDefaultMediumFont;
+    if (_emphasisFont != nil) {
+        return _emphasisFont;
+    } else {
+        if (@available(iOS 11.0, *)) {
+            UIFontMetrics *fontMetrics = [UIFontMetrics metricsForTextStyle:UIFontTextStyleBody];
+            return [fontMetrics scaledFontForFont:STPThemeDefaultMediumFont];
```
yuki-stripe: What happens if we also called this for `_emphasisFont` e.g. remove the if statement above and replace this line with `scaledFontForFont:_emphasisFont ?: STPThemeDefaultMediumFont`?

More broadly I'm wondering if we could support dynamic type even for user-supplied fonts.

Stripe/STPTheme.m
csabol-stripe: ah good idea

Stripe/STPTheme.m
csabol-stripe: yeah that _might_ be something we want to do, but I'm not sure. It seems simpler to not mess with custom set fonts to me so I was thinking have that as an initial solution and see what the response is like?

Stripe/STPTheme.m
csabol-stripe: and yes re the dynamic getter :)

Issue #1486, opened 2020-02-06
Crash when entering bank card number


Goodman333:
## Summary

I encountered a problem, we have two users crash when entering bank card number

## iOS version

iPhone 7 Plus, iOS 13.3
iPhone XS, iOS 13.3

## Installation method

Cocoapods

## SDK version

Stripe (18.0.0)

## Other information

<img width="1079" alt="WechatIMG7" src="https://user-images.githubusercontent.com/9305248/73905129-4458e100-48d9-11ea-95e1-b336a4c1e687.png">

davidme-stripe:
Thanks for reporting this, I'm trying various inputs to reproduce the issue. Any additional information you can provide would be helpful!

Goodman333:
> Thanks for reporting this, I'm trying various inputs to reproduce the issue. Any additional information you can provide would be helpful!

Thank you for replying, I will provide you with a complete crash file, hope this will be helpful
[010927ed81eb474c981e97964668216b-symbolicated.crash.zip](https://github.com/stripe/stripe-ios/files/4209398/010927ed81eb474c981e97964668216b-symbolicated.crash.zip)

davidme-stripe:
I think this was fixed by #1519! Please re-open if you see it again.

PR #1519, opened 2020-03-17
Adds payment details entry UI for AU BECS Debit

csabol-stripe:
## Summary

Adds STPAUBECSDebitFormView that can be used to collect the required data to create an AU BECS Debit PaymentMethod.

## Motivation

Support for new payment method AU BECS Debit.

## Testing

Added examples to UI Examples and Non-Card Payment.
Added snapshot tests for new UI.
Added unit tests for view models and validators.
Tested on iOS 13, 12.2, and 10.2 simulators.


Example/UI Examples/AUBECSDebitFormViewController.swift
```diff
@@ -0,0 +1,47 @@
...
+            ])
+        }
+
+        @objc func done() {
+            dismiss(animated: true, completion: nil)
+        }
+
+        override func viewDidAppear(_ animated: Bool) {
+            super.viewDidAppear(animated)
+//            becsFormView.becomeFirstResponder()
```
davidme-stripe: Should this be removed?

davidme-stripe:
* `return` in the keyboard doesn't switch to the next field.
* When I finish typing the bank account number after typing an incorrect email address (with an accidental space in it), I get redirected to the email field. This seems weird?
* The selection/tap handlers are broken for all of these fields. When I try to move the cursor, it always highlights the entire field. I also can't seem to trigger the cut/copy/paste menu.

csabol-stripe:
> Hitting return on the keyboard when typing doesn't switch to the next field.
Is this expected? Tabbing works

> When I finish typing the bank account number after typing an incorrect email address (with an accidental space in it), I get redirected to the email field. This seems weird?
This was intentional, copying the behavior of STPPaymentCardTextField moving between subfields, but I could see how it's more jarring in this layout. Do you think we should change to only ever auto-advance to next field if it's incomplete?

> The selection/tap handlers are broken for all of these fields. When I try to move the cursor, it always highlights the entire field. I also can't seem to trigger the cut/copy/paste menu.
Uh weird, I'll take a look. Thanks!

Example/UI Examples/AUBECSDebitFormViewController.swift
csabol-stripe: oops thanks!

davidme-stripe:
> > Hitting return on the keyboard when typing doesn't switch to the next field.
> 
> Is this expected? Tabbing works

It is: Try using it on a real iPhone without a physical keyboard attached, you'll see that hitting return is the usual UX for "next field".

> > When I finish typing the bank account number after typing an incorrect email address (with an accidental space in it), I get redirected to the email field. This seems weird?
> 
> This was intentional, copying the behavior of STPPaymentCardTextField moving between subfields, but I could see how it's more jarring in this layout. Do you think we should change to only ever auto-advance to next field if it's incomplete?
> 

Possibly!

> > The selection/tap handlers are broken for all of these fields. When I try to move the cursor, it always highlights the entire field. I also can't seem to trigger the cut/copy/paste menu.
> 
> Uh weird, I'll take a look. Thanks!

Looks like this happens for our other text fields, too, so maybe this could be fixed separately.

csabol-stripe:
> Whoa, does this mean you managed to reproduce #1486?

Oh yeah, looks like it!

csabol-stripe:
@davidme-stripe I'm gonna close this PR and open a new one because I think travis is broken on this one

PR #1524, opened 2020-03-24
% Addressed Xcode 11.4 compile issue

sgruby:
## Summary
Changed how completion handler is called in order to allow Xcode 11.4 to compile the framework.

## Motivation
When compiling with Xcode 11.4, the compiler throws an error when trying to cast the completion handle to an STPAPIResponseBlock. This change explicitly calls the completion handler with the STPSource parameter.

## Testing
This was visually inspected to verify that the completion handler was called with the correct parameters.


yuki-stripe:
@sgruby @hovaks Thanks for the PR - we're looking into this now.

sgruby:
Looks like #1526 is a cleaner fix than what I have, so I'm going to close this.

Issue #1525, opened 2020-03-24
Stripe pod does not compile on Xcode 11.4


stephengilroy-grx:
## Summary
After just recently updating to the new Xcode 11.4 I am unable to build my project withe Stripe SDK pod. Compiler error given is:
```Incompatible block pointer types sending '__strong STPAPIResponseBlock _Nonnull' (aka 'void (^__strong)(ResponseType  _Nullable __strong, NSHTTPURLResponse * _Nullable __strong, NSError * _Nullable __strong)') to parameter of type 'void (^ _Nonnull)(STPSource * _Nullable __strong, NSHTTPURLResponse * _Nullable __strong, NSError * _Nullable __strong)'```

## Code to reproduce
I was able to repro with a newly created empty project. Just run `pod init` add the latest stripe pod and try to build. 

## iOS version
13.4

## Installation method
Cocoapods

## SDK version
19.0.0 (also observed this behavior on 18.x versions)

## Other information
I've tried everything imaginable: nuking pods, restarting my computer, deleting derived data, etc.  As soon as I downgrade to 11.3.1 I am able to successfully build again.

From the error response above it seems this might be related to ObjC generics?

bryanbartow:
Looks like this might be fixed in #1524

stephengilroy-grx:
https://github.com/stripe/stripe-ios/pull/1526 was just merged in so I believe this should be resolved soon 🤞

yuki-stripe:
Thanks for the report, we're working on releasing a new version that fixes this now.

csabol-stripe:
We've just published a release with this fix https://github.com/stripe/stripe-ios/releases/tag/v19.0.1
It should also be live on cocoapods

prgorasiya:
My app still supports iOS 9.
Therefore I have to use V17.0.2 and thats throwing issues with Xcode 11.4
Any way to fix this?

mhergon:
> My app still supports iOS 9.
> Therefore I have to use V17.0.2 and thats throwing issues with Xcode 11.4
> Any way to fix this?

Same here!

mikedizon:
Issued a PR for v14.0.0/v14.0.1

https://github.com/stripe/stripe-ios/pull/1534

davidme-stripe:
[...]

PR #1526, opened 2020-03-24
Fixes completion block type definition

csabol-stripe:
## Summary

We were using a typedef-ed block param that depended on a class generic outside of that class. Updating to 11.4 caught the error.

## Motivation

Fixes build on Xcode 11.4

## Testing

Made sure build and analyzer passed with change

### DIFF
--- a/Stripe/STPAPIClient+Private.h
+++ b/Stripe/STPAPIClient+Private.h
@@ -41,7 +41,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (NSURLSessionDataTask *)retrieveSourceWithId:(NSString *)identifier
                                   clientSecret:(NSString *)secret
-                            responseCompletion:(STPAPIResponseBlock)completion;
+                            responseCompletion:(void (^)(STPSource * _Nullable, NSHTTPURLResponse * _Nullable, NSError * _Nullable))completion;
 
 @end
 
--- a/Stripe/STPAPIClient.m
+++ b/Stripe/STPAPIClient.m
@@ -515,7 +515,9 @@ static NSString *_defaultPublishableKey;
     }];
 }
 
-- (NSURLSessionDataTask *)retrieveSourceWithId:(NSString *)identifier clientSecret:(NSString *)secret responseCompletion:(STPAPIResponseBlock)completion {
+- (NSURLSessionDataTask *)retrieveSourceWithId:(NSString *)identifier
+                                  clientSecret:(NSString *)secret
+                            responseCompletion:(void (^)(STPSource * _Nullable, NSHTTPURLResponse * _Nullable, NSError * _Nullable))completion {
     NSString *endpoint = [NSString stringWithFormat:@"%@/%@", APIEndpointSources, identifier];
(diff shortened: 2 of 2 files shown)


AshleyDeng:
Thanks for the fix! 
When we are making the release, could we also make an update for v17.0.2? 
I think there are a lot of projects that still use v17.0.2 because they need to support iOS 9.

Thank you again!

mikedizon:
When can this release go out?

jwaldrip:
We are stuck as well...

csabol-stripe:
We've published a release with this fix https://github.com/stripe/stripe-ios/releases/tag/v19.0.1

AshleyDeng:
@csabol-stripe Thank you! 
Do we have a fix for v17.0.1 as well? We have lots of projects support iOS 9.

mikedizon:
@csabol-stripe would be great if v14 could be patched as well since it appears that tipsi-stripe isn’t actively maintained

borut-t:
@csabol-stripe I also have a request to patch v15 since our company hasn't migrated yet to the new API.

Stripe/STPAPIClient+Private.h
```diff
@@ -41,7 +41,7 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (NSURLSessionDataTask *)retrieveSourceWithId:(NSString *)identifier
                                   clientSecret:(NSString *)secret
-                            responseCompletion:(STPAPIResponseBlock)completion;
+                            responseCompletion:(void (^)(STPSource * _Nullable, NSHTTPURLResponse * _Nullable, NSError * _Nullable))completion;
```
facetious: If you're sure about this breaking change, you should at least define this new type so that it can be reused by name rather than requiring a full signature match.

Consider: https://github.com/stripe/stripe-ios/blob/7152797d123b20fd3bee8e29ab84b9bf3e9227a2/Stripe/PublicHeaders/STPBlocks.h

Stripe/STPAPIClient.m
```diff
@@ -515,7 +515,9 @@ - (void)retrieveSourceWithId:(NSString *)identifier clientSecret:(NSString *)sec
     }];
 }
 
-- (NSURLSessionDataTask *)retrieveSourceWithId:(NSString *)identifier clientSecret:(NSString *)secret responseCompletion:(STPAPIResponseBlock)completion {
+- (NSURLSessionDataTask *)retrieveSourceWithId:(NSString *)identifier
+                                  clientSecret:(NSString *)secret
+                            responseCompletion:(void (^)(STPSource * _Nullable, NSHTTPURLResponse * _Nullable, NSError * _Nullable))completion {
```
facetious: This completion block is passed straight through into `getWithAPIClient`, whose signature is expecting a different block structure.

See: https://github.com/stripe/stripe-ios/blob/47a13b20f155afe759563ee0c0b3161f6337026c/Stripe/STPAPIRequest.h#L50

davidme-stripe:
Just FYI for folks following this patch, fixes went out for v14, v15, v17, and v19. See #1525.

PR #1538, opened 2020-04-03
Rework, add ApplePayContext to analytics

yuki-stripe:
## Summary
* Rework how STPAnalyticsClient calculates productUsage FKA apiUsage.  
* Also track ApplePayContext usage in analytics

## Motivation
Swizzling the init has a problem where `STPAnalyticsClient initializeIfNeeded` isn't always called before the init.  If we always have to call this method before init in every class we want to track, we might as well not swizzle.

I put the `addClassToAPIUsageIfNecessary` calls in `alloc` because it's easier than identifying the designated initializer and stable (what if we change the designated init in the future?).

## Testing
See tests

### DIFF
--- a/Stripe/STPAnalyticsClient.m
+++ b/Stripe/STPAnalyticsClient.m
@@ -15,7 +15,6 @@
 #import "STPAPIClient+Private.h"
 #import "STPAddCardViewController+Private.h"
 #import "STPAddCardViewController.h"
-#import "STPAspects.h"
 #import "STPCard.h"
 #import "STPCardIOProxy.h"
 #import "STPFormEncodable.h"
@@ -31,7 +30,7 @@
 
 @interface STPAnalyticsClient()
 
-@property (nonatomic) NSSet *apiUsage;
+@property (nonatomic) NSMutableSet *apiUsage;
 @property (nonatomic) NSSet *additionalInfoSet;
 @property (nonatomic, readwrite) NSURLSession *urlSession;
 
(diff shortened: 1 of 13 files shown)


csabol-stripe:
> Swizzling the init has a problem where STPAnalyticsClient initializeIfNeeded isn't always called before the init.

@yuki-stripe could you explain this some more? I don't think putting these calls in `alloc` feels right -- I would expect them either in the designated initializer or in the `+ (void)initialize` function which is documented as `The runtime sends initialize to each class in a program just before the class, or any class that inherits from it, is sent its first message from within the program.` Are you seeing behavior inconsistent with that?

Stripe/STPAnalyticsClient.m
```diff
@@ -144,12 +76,16 @@ - (instancetype)init {
...
         _urlSession = [NSURLSession sessionWithConfiguration:config];
-        _apiUsage = [NSSet set];
+        _productUsage = [NSMutableSet set];
         _additionalInfoSet = [NSSet set];
     }
     return self;
 }
 
+- (void)addClassToAPIUsageIfNecessary:(Class)klass {
+    [self.productUsage addObject:NSStringFromClass(klass)];
```
davidme-stripe, line 86: You might want to add some sort of lock around your reads/writes of productUsage: This has a chance of blowing up if someone initializes multiple classes at once.

yuki-stripe:
> > Swizzling the init has a problem where STPAnalyticsClient initializeIfNeeded isn't always called before the init.
> 
> @yuki-stripe could you explain this some more? I don't think putting these calls in `alloc` feels right -- I would expect them either in the designated initializer or in the `+ (void)initialize` function which is documented as `The runtime sends initialize to each class in a program just before the class, or any class that inherits from it, is sent its first message from within the program.` Are you seeing behavior inconsistent with that?

Thanks for the suggestion - I think I tried that with the old swizzling way and ran into an issue, but looks to work perfectly with the refactor.  It also solves the thread safety issue @davidme-stripe mentions.

> The runtime sends the initialize message to classes in a thread-safe manner. That is, initialize is run by the first thread to send a message to a class, and any other thread that tries to send a message to that class will block until initialize completes.

davidme-stripe:
> That is, initialize is run by the first thread to send a message to a class, and any other thread that tries to send a message to that class will block until initialize completes.

The concern would be with multiple simultaneous threads trying to initialize _different_ classes: For example, someone creating an `STPAddCardViewController` on the main thread, while simultaneously creating an `STPCustomerContext` on a background thread. There's a very small chance that this would lead to `[productUsage addObject:]` being called on two threads at once.

yuki-stripe:
> > That is, initialize is run by the first thread to send a message to a class, and any other thread that tries to send a message to that class will block until initialize completes.
> 
> The concern would be with multiple simultaneous threads trying to initialize _different_ classes: For example, someone creating an `STPAddCardViewController` on the main thread, while simultaneously creating an `STPCustomerContext` on a background thread. There's a very small chance that this would lead to `[productUsage addObject:]` being called on two threads at once.

Ahhh now I understand, thanks!  I _believe_ the last patch fixes it.

PR #1541, opened 2020-04-08
Adds localization helper scripts

csabol-stripe:
## Summary

Adds new scripts to push new localized strings to lokalise, pull translations from lokalise, detect missing translations, and detect newly added localized strings.

CI will now include a check for any localized strings that were not added to en.lproj/Localizable.strings

## Motivation

Make it easier to keep track of our localizations

## Testing

Tested uploading and downloading from lokalise. Tested finding missing translations. Tested detecting new localized strings and removing one's that aren't used in code base.


csabol-stripe:
I can't find documentation on lokalise's encoding, so I'm gonna hold off commiting this one until I get a response from them so I can determine if we need to use recode to convert to utf-8 or not

csabol-stripe:
companion wiki page on when to use these coming ~tomorrow

ci_scripts/download_localized_strings_from_lokalise.sh
```diff
@@ -0,0 +1,37 @@
...
+
+if [[ -z $(which recode) ]]; then
+    echo "Installing recode via homebrew..."
+    brew install recode
+fi
+
+API_TOKEN=$(fetch-password mobile/lokalise/token -q)
+PROJECT_ID=$(fetch-password mobile/lokalise/ios -q)
+LANGUAGES="da,de,en-GB,en,es-AR,es,fi,fr-CA,fr,it,ja,ko,nb,nl,nn-NO,pt-BR,pt-PT,ru,sv,tr,zh-HANS,zh-HK,zh-TW"
+# Chinese (Traditional) is zh-TW by default in lokalise so we have to map it to zh-Hant
```
mkaplan9, line 17: If you want you can change this in Lokalise

Click the flag, then `Language Settings`
image

Change `Custom language code` to zh-Hant
image

csabol-stripe:
> All our exported files are encoded in UTF-8.

👍

ci_scripts/download_localized_strings_from_lokalise.sh
```diff
@@ -0,0 +1,37 @@
...
+lokalise2 --token $API_TOKEN \
+          --project-id $PROJECT_ID \
+          file download \
+          --format strings \
+          --filter-langs $LANGUAGES \
+          --language-mapping $LANG_MAP \
+          --export-sort "a_z" \
+          --directory-prefix "%LANG_ISO%.lproj" \
+          --original-filenames=true \
+          --unzip-to Stripe/Resources/Localizations/
```
mkaplan9, line 29: The workflow we have with translators has than go through a few steps, and the string is only complete when it's in the `Final` state. To pull only `Final` keys you can add
```
--custom-translation-status-ids final_status_id
```

ios sdk final status id is 587
ios 3ds2 sdk final status id is 583

ci_scripts/upload_localized_strings_to_lokalise.sh
```diff
@@ -0,0 +1,22 @@
...
+fi
+
+API_TOKEN=$(fetch-password mobile/lokalise/token -q)
+PROJECT_ID=$(fetch-password mobile/lokalise/ios -q)
+
+lokalise2 --token $API_TOKEN \
+          --project-id $PROJECT_ID \
+          file upload \
+          --file Stripe/Resources/Localizations/en.lproj/Localizable.strings \
+          --lang-iso "en"
```
mkaplan9, line 22: Will the key name always == the string? If yes, this looks good, if not there are a few extra params you may want.

mkaplan9:
> I can't find documentation on lokalise's encoding, so I'm gonna hold off commiting this one until I get a response from them so I can determine if we need to use recode to convert to utf-8 or not

Did you reach out to them yet? I not I can today

ci_scripts/download_localized_strings_from_lokalise.sh
csabol-stripe: oh cool, I'll add!

ci_scripts/upload_localized_strings_to_lokalise.sh
csabol-stripe: yeah for the time being. I think in the future we'll probably want key,value,comment but right now we only support key/value, comment

ci_scripts/download_localized_strings_from_lokalise.sh
csabol-stripe: Yeah I like that better. Updated in lokalise

mkaplan9:
good from i18n standpoint

PR #1548, opened 2020-04-17
Handle redirect completion after app is active

vineet-stripe:
## Summary
Ensures that the redirect completion is handled _after_ the app is in the foreground; moves completion handler from `applicationWillEnterForeground` to `applicationDidBecomeActive`.

## Motivation
IOS-1641

## Testing
CI
Non-Card Payment Examples app


davidme-stripe:
LGTM! You mentioned earlier that you were testing this in the example apps too — you should probably add that to the `# Testing` section.

vineet-stripe:
Updated a unit test, ptal @davidme-stripe

PR #1549, opened 2020-04-17
Replaces es-AR localization with es-419

csabol-stripe:
## Summary

Phrase only gave us es-AR which we used for ~Spanish - Latin America. Now we can use the actual designation of es-419

## Motivation

Better localization

## Testing

Built basic integration with scheme set to spanish Argentia


csabol-stripe:
> Cool. Do we need to request that they're retranslated, given it's now the localization for all of latin america?

No they were always intended for all of Latin America

PR #1550, opened 2020-04-18
Improve accessibility of checkout in Basic Integration example

vineet-stripe:
## Summary

`CheckoutRowView` is now a top-level accessibility element instead of exposing its subviews as individual accessibility elements.

## Motivation

IOS-1542
IOS-1547

## Testing

Accessibility Inspector on macOS running Basic Integration example app in iOS Simulator

### DIFF
--- a/Example/BasicIntegrationUITests/BasicIntegrationUITests.swift
+++ b/Example/BasicIntegrationUITests/BasicIntegrationUITests.swift
@@ -53,7 +53,7 @@ class BasicIntegrationUITests: XCTestCase {
         selectItems(app)
         
         app.buttons["Buy Now"].tap()
-        app.tables.otherElements.containing(.staticText, identifier:"Pay from").children(matching: .button).element.tap()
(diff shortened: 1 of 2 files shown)


Example/BasicIntegrationUITests/BasicIntegrationUITests.swift
```diff
@@ -185,10 +183,11 @@ class BasicIntegrationUITests: XCTestCase {
         selectItems(app)
         let buyNowButton = app.buttons["Buy Now"]
         buyNowButton.tap()
-        let tablesQuery = app.tables
-        let payFromButton = tablesQuery.otherElements.containing(.staticText, identifier:"Pay from").children(matching: .button).element
+        let payFromButton =         app.buttons.matching(identifier: "Pay from").element
```
yuki-stripe, line 186: Extra whitespace

vineet-stripe: @yuki-stripe Thanks! Updated

PR #1553, opened 2020-04-22
Giropay PaymentMethod bindings

csabol-stripe:
## Summary

Adds bindings for giropay on PaymentMethods

## Motivation

Migrating Sources->PaymentMethods

## Testing

Automated tests and added to non-card examples

PR #1556, opened 2020-04-24
Przelewy24 PaymentMethod bindings

vineet-stripe:
## Summary
Adds bindings for Przelewy24 on PaymentMethods

## Motivation
Migrating Sources->PaymentMethods

## Testing
Automated tests and added to non-card examples

### DIFF
--- a/Example/Non-Card Payment Examples/Przelewy24ExampleViewController.m
+++ b/Example/Non-Card Payment Examples/Przelewy24ExampleViewController.m
@@ -61,7 +61,7 @@ NS_ASSUME_NONNULL_BEGIN
         STPPaymentMethodPrzelewy24Params *przelewy24 = [[STPPaymentMethodPrzelewy24Params alloc] init];
 
         // Przelewy24 does not require additional parameters so we only need to pass the init-ed
-        // STPPaymentMethoPrzelewy24Params instance to STPPaymentMethodParams
+        // STPPaymentMethodPrzelewy24Params instance to STPPaymentMethodParams
         paymentIntentParams.paymentMethodParams = [STPPaymentMethodParams paramsWithPrzelewy24:przelewy24
                                                                                 billingDetails:billingDetails
                                                                                       metadata:nil];
--- a/Example/Non-Card Payment Examples/Przelewy24ExampleViewController.m
+++ b/Example/Non-Card Payment Examples/Przelewy24ExampleViewController.m
@@ -1,0 +1,90 @@
+//
+//  Przelewy24ExampleViewController.m
+//  Non-Card Payment Examples
+//
+//  Created by Vineet Shah on 4/23/20.
+//  Copyright © 2020 Stripe. All rights reserved.
+//
+
+#import "Przelewy24ExampleViewController.h"
+
+#import "MyAPIClient.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface Przelewy24ExampleViewController ()
+
+@end
+
+@implementation Przelewy24ExampleViewController {
+    UITextField *_emailField;
+}
+
+- (void)viewDidLoad {
+    [super viewDidLoad];
+
+    self.title = @"Przelewy24";
+
+    _emailField = [[UITextField alloc] init];
+    _emailField.borderStyle = UITextBorderStyleRoundedRect;
+    _emailField.textContentType = UITextContentTypeEmailAddress;
+    _emailField.placeholder = @"Email";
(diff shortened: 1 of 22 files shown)


Stripe/PublicHeaders/STPPaymentMethodPrzelewy24.h
```diff
@@ -0,0 +1,36 @@
...
+#import <Foundation/Foundation.h>
+
+#import "STPAPIResponseDecodable.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+/**
+A Przelewy24 Payment Method.
+
+@see https://stripe.com/docs/sources/p24
```
yuki-stripe, line 18: This link should be https://stripe.com/docs/payments/p24

Stripe/PublicHeaders/STPPaymentMethodPrzelewy24.h
```diff
@@ -0,0 +1,36 @@
...
+*/
+@interface STPPaymentMethodPrzelewy24 : NSObject <STPAPIResponseDecodable>
```
yuki-stripe, line 20: Do we know if the API ref is just incomplete, or is this intentionally empty?

vineet-stripe: My understanding is that P24 doesn't have any addtional params besides the required email in billing details, could you help confirm? https://site-admin.stripe.com/docs/payments/p24

yuki-stripe: Ah maybe @mshafrir-stripe or @stulentsev-stripe has context:

It's weird that this hash is [empty](https://site-admin.stripe.com/docs/api/payment_methods/object#payment_method_object-p24) in the [API ref](https://site-admin.stripe.com/docs/api/payment_methods/create#create_payment_method-p24) and the corresponding SDK objects are empty. Why not delete them?

mshafrir-stripe: In the Android SDK we didn't create an object to represent P24 params.

stulentsev-stripe: Documentation is auto-generated. Guess, it doesn't have a notion of omitting empty sub-hashes. And removing them would break symmetry with `type` field. We'd then have to explain why `type` accepts value `p24`, but there's no subhash for it. Was it oversight on our part? Etc.

vineet-stripe: Are these the first payment methods we've implemented that don't have any params?

FWIW this PR follows what @csabol-stripe did with giropay in #1553 -- perhaps he has some thoughts?

csabol-stripe: I think we should have these (empty classes)
1) For parity with the actual API response which contains `<payment_method>: {}`
2) For parity with the docs
3) Parsing the empty hash is a reasonable broken window test
4) Easier to iterate on/add to in the future

mshafrir-stripe: That sounds reasonable. FYI this is what we're doing in the Android SDK - https://github.com/stripe/stripe-android/blob/e4f8ac699422a18f3aba8d471363a6980710c039/stripe/src/main/java/com/stripe/android/model/PaymentMethodCreateParams.kt#L488

Example/Non-Card Payment Examples/Przelewy24ExampleViewController.m
```diff
@@ -0,0 +1,90 @@
...
+        STPPaymentIntentParams *paymentIntentParams = [[STPPaymentIntentParams alloc] initWithClientSecret:clientSecret];
+
+        STPPaymentMethodBillingDetails *billingDetails = [[STPPaymentMethodBillingDetails alloc] init];
+        billingDetails.email = self->_emailField.text;
+
+
+        STPPaymentMethodPrzelewy24Params *przelewy24 = [[STPPaymentMethodPrzelewy24Params alloc] init];
+
+        // Przelewy24 does not require additional parameters so we only need to pass the init-ed
+        // STPPaymentMethoPrzelewy24Params instance to STPPaymentMethodParams
```
vineet-stripe, line 64:
```suggestion
        // STPPaymentMethodPrzelewy24Params instance to STPPaymentMethodParams
```

vineet-stripe:
@csabol-stripe This is rebased on `master` now, needs another ✅

PR #1578, opened 2020-05-14
EPS PaymentMethod Bindings

shengwei-stripe:
## Summary
Adding EPS PaymentMethod

## Motivation
Migrating Sources->PaymentMethods

## Testing
Automated tests and added EPS to non-card examples


Example/Non-Card Payment Examples/EPSExampleViewController.m
```diff
@@ -0,0 +1,94 @@
...
+        STPPaymentIntentParams *paymentIntentParams = [[STPPaymentIntentParams alloc] initWithClientSecret:clientSecret];
+
+        STPPaymentMethodBillingDetails *billingDetails = [[STPPaymentMethodBillingDetails alloc] init];
+        billingDetails.name = self->_nameField.text;
+
+
+        STPPaymentMethodEPSParams *eps = [[STPPaymentMethodEPSParams alloc] init];
+
+        // EPS does not require additional parameters so we only need to pass the init-ed
+        // STPPaymentMethoEPSParams instance to STPPaymentMethodParams
```
yuki-stripe, line 67: typo: STPPaymentMethoEPSParams -> STPPaymentMethodEPSParams

Stripe/PublicHeaders/STPPaymentMethodParams.h
```diff
@@ -184,6 +190,17 @@ NS_ASSUME_NONNULL_BEGIN
...
+/**
+ Creates params for a giropay PaymentMethod;
+
+ @param eps   An object containing additional EPS details.
+ @param billingDetails  An object containing the user's billing details. Note that `billingDetails.name` is required for giropay PaymentMethods.
+ @param metadata     Additional information to attach to the PaymentMethod.
+ */
++ (nullable STPPaymentMethodParams *)paramsWithEPS:(STPPaymentMethodEPSParams *)eps
+                                        billingDetails:(STPPaymentMethodBillingDetails *)billingDetails
+                                              metadata:(nullable NSDictionary<NSString *, NSString *> *)metadata;
```
yuki-stripe, line 202: nit: misaligned lines.  Select the lines and hit control-I in Xcode

Stripe/PublicHeaders/STPPaymentMethodParams.h
```diff
@@ -184,6 +190,17 @@ NS_ASSUME_NONNULL_BEGIN
...
                                               metadata:(nullable NSDictionary<NSString *, NSString *> *)metadata;
 
+/**
+ Creates params for a giropay PaymentMethod;
+
+ @param eps   An object containing additional EPS details.
+ @param billingDetails  An object containing the user's billing details. Note that `billingDetails.name` is required for giropay PaymentMethods.
+ @param metadata     Additional information to attach to the PaymentMethod.
+ */
++ (nullable STPPaymentMethodParams *)paramsWithEPS:(STPPaymentMethodEPSParams *)eps
```
yuki-stripe, line 200: How come this is `nullable`?

Stripe/PublicHeaders/STPPaymentMethodParams.h
```diff
@@ -184,6 +190,17 @@ NS_ASSUME_NONNULL_BEGIN
                                         billingDetails:(STPPaymentMethodBillingDetails *)billingDetails
                                               metadata:(nullable NSDictionary<NSString *, NSString *> *)metadata;
 
+/**
+ Creates params for a giropay PaymentMethod;
```
yuki-stripe, line 194: giropay -> EPS

Stripe/PublicHeaders/STPPaymentMethodParams.h
```diff
@@ -184,6 +190,17 @@ NS_ASSUME_NONNULL_BEGIN
...
+
+ @param eps   An object containing additional EPS details.
+ @param billingDetails  An object containing the user's billing details. Note that `billingDetails.name` is required for giropay PaymentMethods.
```
yuki-stripe, line 197: giropay -> EPS

Also, is the Note true?  Is billingDetails.name the single required field?

csabol-stripe: https://stripe.com/docs/sources/eps#create-source

Note's correct

Stripe/PublicHeaders/STPPaymentMethodParams.h
csabol-stripe: most of the other methods here are nullable. I think they should be since the depend on a [[self alloc] init]?

yuki-stripe: Huh I didn't think about that.

> The init method defined in the NSObject class does no initialization; it simply returns self. In terms of nullability, callers can assume that the NSObject implementation of init does not return nil.

I think this means we can safely mark them as nonnull unless we add logic to return nil.  This is good because it improves the Swift ergonomics!

shengwei-stripe: change to `nonnull` as suggested

PR #1580, opened 2020-05-18
Fixes tabbing behavior in STPAUBECSFormView

csabol-stripe:
## Summary

* Makes STPMultiFormTextField a UIView instead of UIControl
* Removes UIKeyInput conformance so system calls directly to STPFormTextField subviews
* Cleans up UIResponder method implementations

## Motivation

Enables tabbing with external keyboard

## Testing

Tested in simulator

### DIFF
--- a/Stripe/PublicHeaders/STPMultiFormTextField.h
+++ b/Stripe/PublicHeaders/STPMultiFormTextField.h
@@ -56,7 +56,7 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  STPMultiFormTextField is a lightweight UIControl that wraps a collection of STPFormTextFields and can automatically move to the next form field when one is completed.
  */
-@interface STPMultiFormTextField : UIControl <STPFormTextFieldContainer, UIKeyInput>
+@interface STPMultiFormTextField : UIView <STPFormTextFieldContainer>
 
 /**
  The collection of STPFormTextFields that this instance manages.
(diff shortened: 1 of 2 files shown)


Stripe/PublicHeaders/STPMultiFormTextField.h
```diff
@@ -56,7 +56,7 @@ NS_ASSUME_NONNULL_BEGIN
 /**
  STPMultiFormTextField is a lightweight UIControl that wraps a collection of STPFormTextFields and can automatically move to the next form field when one is completed.
```
yuki-stripe, line 57: Update the comment

Stripe/STPMultiFormTextField.m
```diff
@@ -48,27 +48,43 @@ - (void)focusNextFormField {
...
+        return [[self _currentFirstResponderField] resignFirstResponder];
+    } else {
+        return YES;
+    }
 }
 
 - (BOOL)isFirstResponder {
-    return [self _currentFirstResponderField] != nil;
+    return [super isFirstResponder] || [[self _currentFirstResponderField] isFirstResponder];
+//    return [self _currentFirstResponderField] != nil;
```
yuki-stripe, line 69: Delete?

Stripe/PublicHeaders/STPMultiFormTextField.h
csabol-stripe: 👍 thanks

PR #1608, opened 2020-07-20
Adds support for dynamic card metadata and >16 digit PANs

csabol-stripe:
## Summary

* Adds API bindings for new, private card metadata endpoint
* Disables card number truncation in `STPPaymentCardTextField`
* Disables auto-advancing out of the number entry field in `STPPaymentCardTextField`
* Updates card number validation to wait for response from card metadata endpoint
* Updates tests with new card number behavior and asynchronicity

## Motivation

Support for CUP cards with > 16 digits in some cases.

## Testing

Added automated tests for new API binding. Updated existing STPPaymentCardTextField tests with new behavior.


Stripe/STPPaymentCardTextField.m
```diff
@@ -1287,50 +1303,72 @@ - (NSAttributedString *)formTextField:(STPFormTextField *)formTextField
...
-            formTextField.validText = NO;
+
+    switch (fieldType) {
+
+        case STPCardFieldTypeNumber: {
+            NSString *number = [self.viewModel.cardNumber copy];
+            
+            // Changing the card number field can invalidate the cvc, e.g. going from 4 digit Amex cvc to 3 digit Visa
+            // it is not expected that the brand will change based on network response so we update this immediately
+            // as well as in the completion just in case
```
yuki-stripe, line 1316: We don't update brand in the completion block - should we? Or should we remove the last comment?

Stripe/STPBINRange.m
```diff
@@ -142,4 +203,56 @@ + (instancetype)mostSpecificBINRangeForNumber:(NSString *)number {
...
+    
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        sRetrievalQueue = dispatch_queue_create("com.stripe.retrieveBINRangesForPrefix", DISPATCH_QUEUE_SERIAL);
+        sPendingRequests = [NSMutableDictionary new];
+        sRetrievedRanges = [NSMutableDictionary new];
+    });
+    
+    dispatch_async(sRetrievalQueue, ^{
+        NSString *binPrefixKey = [binPrefix stp_safeSubstringToIndex:6];
```
yuki-stripe, line 219: can we comment what 6 signifies here? Maybe make it a constant too, since it's used twice?

Stripe/STPPaymentCardTextFieldViewModel.m
```diff
@@ -106,40 +100,42 @@ - (STPCardBrand)brand {
...
+    if (self.postalCode.length > 0) {
+        return STPCardValidationStateValid;
+    } else {
+        return STPCardValidationStateIncomplete;
     }
 }
 
+- (void)validationStateForCardNumberWithHandler:(void (^)(STPCardValidationState))handler {
+    [STPBINRange retrieveBINRangesForPrefix:self.cardNumber completion:^(__unused NSArray<STPBINRange *> * _Nullable ranges, __unused NSError * _Nullable error) {
+        handler([STPCardValidator validationStateForNumber:self.cardNumber validatingCardBrand:YES]);
```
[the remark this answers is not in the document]
csabol-stripe: I think I slightly prefer `retrieveBINRanges` because the expectation is that you always get the BIN ranges with this method even if it doesn't have to make a network request to update them, but don't feel strongly.

I could see it on `STPCardValidator`, but I'd want to track the retrieved ranges in the same place as the hard-coded ranges so `STPBINRange` seemed more sensible. Open to ideas for that though

Stripe/STPPaymentCardTextField.m
csabol-stripe: brand is a computed property on the viewmodel so it's implicitly updated when we make the call to update the cvc placeholder again

Stripe/STPBINRange.m
csabol-stripe: updated in second patch

davidme-stripe (approved):

PR #1619, opened 2020-08-07
Sofort PaymentMethod bindings

davidme-stripe:
## Summary
Adds Sofort as a PaymentMethod.

## Motivation
Sofort should be supported by the Payment Intents API.

## Testing
In the Non-Card Payment Examples app and tests in CI.


Stripe/PublicHeaders/STPPaymentMethodSofort.h
```diff
@@ -0,0 +1,41 @@
...
+/**
+You cannot directly instantiate an `STPPaymentMethodSofort`.
+You should only use one that is part of an existing `STPPaymentMethod` object.
+*/
++ (instancetype)new NS_UNAVAILABLE;
+
+/**
+ Two-letter ISO code representing the country the bank account is located in. Required.
+ */
+@property (nonatomic, nullable, readonly) NSString *country;
```
csabol-stripe, line 37: if this is required it shouldn't be nullable right?

Stripe/Resources/Localizations/zh-Hant.lproj/Localizable.strings
```diff
@@ -241,8 +241,9 @@
 /* Label for shipping method form */
 "Shipping Method" = "送貨方式";
 
-/* Source type brand name */
-"SOFORT" = "SOFORT";
+/* Payment Method type brand name
+   Source type brand name */
+"Sofort" = "Sofort";
```
csabol-stripe, line 246: we should update these in lokalise...

Stripe/STPPaymentMethodSofort.m
```diff
@@ -0,0 +1,53 @@
...
+    if (!dict) {
+        return nil;
+    }
+    return [[self alloc] initWithDictionary:dict];
+}
+
+- (nullable instancetype)initWithDictionary:(NSDictionary *)dict {
+    self = [super init];
+    if (self) {
+        _country = [[dict stp_stringForKey:@"country"] copy];
```
csabol-stripe, line 45: missing required params should return nil

Stripe/Resources/Localizations/zh-Hant.lproj/Localizable.strings
davidme-stripe: Will do, I just requested access.

csabol-stripe: you should be able to just run `./ci_scripts/upload_localized_strings_to_lokalise.sh`, but we should manually log in to see if it removes the caps version

PR #1660, opened 2020-10-05
Removes invalid mandate ID confirm params

csabol-stripe:
## Summary

Deprecates mandate properties and removes their usage in SetupIntent and PaymentIntent confirmation

## Motivation

Mandate ID is not usable with publishable keys. They must be set using a secret key.
see https://stripe.com/docs/api/setup_intents/confirm https://stripe.com/docs/api/payment_intents/confirm#confirm_payment_intent-mandate

## Testing

Existing tests.


yuki-stripe (approved):
Given that this never worked and would presumably always result in failure, I favor hard-removing this without deprecation.

yuki-stripe (approved):

Issue #1659, opened 2020-10-03
Failed to setup camera to scan credit card


unchartedworks:
## Summary

I tried to add a card. When I tapped "Scan card" button, I got an error immediately. It failed to create VNRecognizeTextRequest.

NSError *	domain: @"com.apple.vis" - code: 3	0x00000002803d40c0
@"NSLocalizedDescription" : @"Operation failed due to attempt to crop zero or near zero dimensioned area"	

## Code to reproduce
STPCardScanner.m

```
- (void)setupCamera {
    __weak typeof(self) weakSelf = self;
    self.textRequest = [[VNRecognizeTextRequest alloc] initWithCompletionHandler:^(VNRequest * _Nonnull request, NSError * _Nullable error) {
        __strong typeof(self) strongSelf = weakSelf;
        if (!strongSelf.isScanning) {
            return;
        }
        if (error) {
            [strongSelf stopWithError:[STPCardScanner stp_cardScanningError]];
            return;
        }
        [strongSelf processVNRequest:request];
    }];
```

## iOS version
iOS 14.1

## Installation method
Swift Package Manager

## SDK version
20.0.0

davidme-stripe:
Thanks for the report! I *think* I see the problem: We're not setting up the cropping rect properly when the device orientation is UIDeviceOrientationUnknown. I'll put out a patch soon. Does it work when you rotate the device to landscape?

If that's not it, it could help if we had more information about your configuration:
* What device are you running this on?
* Does the app have camera permissions?
* Do other camera-using apps work?

unchartedworks:
Thanks! The UI only supports portrait mode and it was in portrait mode.
What device are you running this on? iPhone XR
Does the app have camera permissions? Yes
Do other camera-using apps work? Yes

davidme-stripe:
I think this was fixed by #1661, which we'll try to get in a release soon. Can you try the latest `master` branch?

davidme-stripe:
This should be fixed in SDK 20.1.0.

PR #1661, opened 2020-10-06
Fix STPCardScanner in unknown orientations

davidme-stripe:
## Summary
When the device orientation is unknown, STPCardScanner might fail to crop the camera image properly.

## Motivation
#1659

## Testing
Tested on device.


yuki-stripe (approved):

csabol-stripe (approved):
ci will fail because of metadata tests. I have a fix in https://github.com/stripe/stripe-ios/pull/1660

PR #1694, opened 2020-10-26
Swift conversion review: Delete Objective-C files and update build configs

davidme-stripe:
Update the `Stripe` folder, deleting the Objective-C files and updating the build config for Swift.


csabol-stripe (commented):
so much red! This one is also reverting the version back to 20.0.0 though

davidme-stripe:
That was ~okay, as we'll need to update the version number to the real one once we merge. (And we'll probably have 20.1.2 before this is out.) But I moved it to 20.1.1 for now.

davidme-stripe:
Flipped (though we'll need to update it again once this lands).

csabol-stripe (approved):

PR #1753, opened 2021-01-25
Fix the theme of button items in the navigation bar

rbenna:
## Summary
Accessing `navigationController` property of `UIViewController` from `viewDidLoad` will always return `nil`. So, `navigationController?.navigationBar.stp_theme ?? theme` in `updateAppearance` in `STPCoreViewController` will always use `theme` property and not the theme that has been set on the navigation bar.

## Motivation
I wanted to customize the navigation bar items with a different theme than the default one, but without success because of the bug described above.

### DIFF
--- a/Stripe/STPCoreViewController.swift
+++ b/Stripe/STPCoreViewController.swift
@@ -89,14 +89,13 @@ public class STPCoreViewController: UIViewController {
   @objc
   public override func viewDidLoad() {
     super.viewDidLoad()
-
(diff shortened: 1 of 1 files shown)


davidme-stripe:
Thank you for the pull request! We'll incorporate this into the next version of the SDK.

davidme-stripe:
We landed this in [21.2.1](https://github.com/stripe/stripe-ios/releases/tag/21.2.1).

PR #1777, opened 2021-03-19
Fix: add default mandate data for all applicable payment methods

thorsten-stripe:
## Summary
Add default mandate data for all applicable payment methods.

## Motivation
There are a bunch of payment methods that support set up for future off_session usage and they need to have `mandate_data` set when confirming the PI/SI

## Testing
Only tested manually via my test integration. Would appreciate guidance on what tests need to be written.


yuki-stripe:
This has been merged and will be in the next release of the SDK

PR #1815, opened 2021-06-01
Right UIBarButtonItem theme issues inside the STPPaymentOptionsInternalViewController

hico17:
## Summary
The SDK should not override the provided navigation bar stp_theme with the default theme.

## Motivation
We found on our app that inside the STPPaymentOptionsInternalViewController, inside the reloadRightBarButtonItem function, this code is overriding the previously set navigation bar stp_theme with the default theme: 
`barButtonItem?.stp_setTheme(theme)`

We think this is not correct, indeed this is causing some problems with rightBarButtonItem which is taking the color of the `navigationBar.stp_theme` when in edit mode, then when pressed it become the "done" button but it takes the default theme.
It should keep the color provided with the `navigationBar.stp_theme`

### DIFF
--- a/Stripe/STPPaymentOptionsInternalViewController.swift
+++ b/Stripe/STPPaymentOptionsInternalViewController.swift
@@ -159,7 +159,7 @@ class STPPaymentOptionsInternalViewController: STPCoreTableViewController, UITab
                 action: #selector(handleDoneButtonTapped(_:)))
         }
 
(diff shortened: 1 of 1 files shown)


Cordelro:
Payout

Cordelro:
Payout

davidme-stripe:
Thank you for investigating this, and sorry for the delayed response! We'll test this out and integrate it into a future update.

PR #1819, opened 2021-06-14
Remove invalid exclude from Package.swift

JonathanDowning:
## Summary
Xcode 13 now shows issues with packages in the issue navigator.
One issue popped up for this SDK, namely that there is an invalid exclude in the Package.swift, this PR removes it.

## Motivation
This change will make stripe compile with no warnings in Xcode 13

## Testing
No explicit testing other than CI


csabol-stripe:
This has been updated in master. Thanks!

PR #1834, opened 2021-07-06
STPPaymentCardTextField postalCodeField accessibility label

romanilchyshyndepop:
## Summary
Add missed accessibility label for `STPPaymentCardTextField.postalCodeField` with value equals to field's placeholder by analogy with other fields.

## Motivation
The motivation is to make missed field accessible.

The issue with missed `accessibilityLabel` is reproducible in `UI Examples` project with `Card Field` demo.

## Testing
The change is trivial and doesn't seem that need automated testing coverage. Also, AFAIK project doesn't contain tests on accessibility fields.


csabol-stripe:
Thanks so much for your help making our codebase more accessible @romanilchyshyndepop :) I'm including this change in our next release and will credit you in the CHANGELOG!

davidme-stripe:
This landed in SDK 21.8.0. Thanks again for your contribution!

PR #1846, opened 2021-07-20
Simplify expression in STPSectionHeaderView.swift

JonathanDowning:
## Summary
I've simplified an expression in this view to reduce the compile time from 13 seconds to 6 milliseconds.

## Motivation
I analyze the compile time of my projects from time to time and discovered Stripe currently has a hotspot.

After making this innocuous change, the time to compile goes from:
`12899.38ms stripe-ios/Stripe/STPSectionHeaderView.swift:124:10    instance method height(forButtonText:width:)`
to:
`6.05ms stripe-ios/Stripe/STPSectionHeaderView.swift:123:18    instance method height(forButtonText:width:)`

## Testing
The code is identical in functionality.

### DIFF
--- a/Stripe/STPSectionHeaderView.swift
+++ b/Stripe/STPSectionHeaderView.swift
@@ -56,7 +56,7 @@ class STPSectionHeaderView: UIView {
         }
     }
     private weak var label: UILabel?
-    private var buttonInsets: UIEdgeInsets!
+    private let buttonInsets = UIEdgeInsets(top: 5, left: 5, bottom: 5, right: 15)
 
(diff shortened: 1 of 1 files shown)


davidme-stripe:
Thanks for catching this and cleaning it up! We'll publish it and credit you in the next update.

This code was automatically converted from Objective-C using a tool, and the results were a little odd in some places. We'll run through with `-debug-time-function-bodies` and try to fix a few of the other hotspots.

JonathanDowning:
It's a pleasure @davidme-stripe! Thanks for all the hard work y'all do.

PR #1851, opened 2021-07-24
Attempt to fix Apple Pay callback for server-confirmed PaymentIntents

nickdnk:
## Summary
Added the ability to pass `nil` to both `clientSecret` *and* `intentCreationError` on the Apple Pay callback.

## Motivation
When using server-confirmed PaymentIntents, there is nothing for this delegate to do in an Apple Pay situation where the payment has already been completed. Currently, passing `nil` to both parameters leaves the application in a no-op state and the Apple Pay sheet just times out even though the payment was processed successfully in the background.

Why not just confirm the intent on the client? Because when using pending updates with subscriptions, there is no option to do this. It means that using Apple Pay to pay for a change in subscription *will* mess up your subscription if the payment fails, which is what pending updates are meant to prevent. I discussed this on the Stripe developer Discord before submitting this.

## Testing
I tested it manually by simply unlocking the dependency and running the code in XCode. It works as expected. I have no idea how to properly test this as I never worked with Carthage, so I couldn't get it to run at all. This is more of a suggestion than a viable PR.

### DIFF
--- a/StripeIdentity/StripeIdentity/Source/NativeComponents/Views/IdentityHTMLView/HTMLViewWithIconLabels.swift
+++ b/StripeIdentity/StripeIdentity/Source/NativeComponents/Views/IdentityHTMLView/HTMLViewWithIconLabels.swift
@@ -136,15 +136,12 @@ extension HTMLViewWithIconLabels {
         addAndPinSubview(vStack)
         vStack.addArrangedSubview(separatorView)
         vStack.addArrangedSubview(textView)
+        vStack.setCustomSpacing(Styling.separatorVerticalSpacing, after: separatorView)
     }
 
     fileprivate func installConstraints() {
         NSLayoutConstraint.activate([
             separatorView.heightAnchor.constraint(equalToConstant: IdentityUI.separatorHeight),
-            separatorView.bottomAnchor.constraint(
-                equalTo: textView.topAnchor,
-                constant: -Styling.separatorVerticalSpacing
-            ),
             separatorView.leadingAnchor.constraint(equalTo: vStack.leadingAnchor),
             separatorView.trailingAnchor.constraint(equalTo: vStack.trailingAnchor),
             textView.leadingAnchor.constraint(equalTo: vStack.leadingAnchor),


nickdnk:
Ideally, and I know this would be a breaking change, you'd pass something like a `STPIntentClientSecretCompletion` enum to the callback instead of two arguments:

```swift
enum STPIntentClientSecretCompletion {
    case success(clientSecret: String?)
    case error(error: Error)
}
```

This would utilize type enforcement to make the method more clear, prevent passing invalid combinations (such as this case of double `nil`) and still allow for a non-optional error and an optional client secret on success.

brendanm-stripe:
Hey @nickdnk - thanks for raising this. Looking at this with some folks on the iOS SDK team, we think we'll have a fix for this in the next release, but we want to make sure we understand your flow correctly. Can you please summarize the sequence of events that lead to the issue for you?

For example:
1. Start a subscription with some payment method (Apple Pay?)
2. Make a update to that subscription that requires payment, using `pending_if_incomplete`
3. ...

Exactly what type of payment method, what type of update and when/how you interact with the Apple Pay Context would help us confirm we're on the sage page.

Thanks!

nickdnk:
Hello @brendanm-stripe 

1. Have a subscription running already. Method of payment is irrelevant.
2. Want to make a change to this subscription, but with pending updates to avoid messing up the billing cycle if the payment were to fail.
3. Create a PaymentMethod using Apple Pay, using the default subscription price as authorization amount (given proration has not taken place yet) on the client. The actual amount charged is likely lower; I just tell the user this and send an invoice after payment to clarify.
4. Attach the PaymentMethod to the customer that owns the subscription.
5. Update the subscription using pending updates, i.e. `pending_if_incomplete`.
6. Tell the client (the iOS app) that the update succeeded.

At this time, step 6 is impossible with Apple Pay because the client does not need to do anything, yet it expects a client secret to confirm the payment intent, which has already been confirmed on the server. If you were to pay with a regular card, you'd send back a client secret in step 6 and have the client run through SCA, but for Apple Pay there is never such challenge by design and hence nothing to send back; and the app stalls.

brendanm-stripe:
OK thanks @nickdnk - and if you included `expand[]=latest_invoice.payment_intent` in the subscription update to get the PI `client_secret` of the successful payment, passing that to the `completion()` is where you encounter the error about manual confirmation?

nickdnk:
@brendanm-stripe I have not tried this. We only send back the client secret if the payment status has `requires_action`.

brendanm-stripe:
Ah, I see. There was some mention of that passed on from your discussions in Discord earlier today. but that may have been part of the confusion.

In this case, I'd suggest you do pass back the PI `client_secret` even when successful already, and provide that to the completion. At this point, if I understand, you will encounter the error I alluded to.

nickdnk:
I see how that could play out, but I can't really do that as all other clients except Apple Pay expect that a returned client secret means they should start the SCA flow. Making this change would require us to make quite a few modifications and would also increase processing time as the clients would then have to fetch the payment intents from Stripe to determine if they need to do something. Not a very clean solution to this problem in my humble opinion, at least not compared to just dismissing the Apple Pay sheet.

brendanm-stripe:
As it stands, what you've suggested here requires special casing for Apple Pay completion, and the flow should be supportable today given these changes. I think sending back the status + client secret would be a good solution here. Note that as I mentioned you can get this back immediately in the subscription update response by leveraging expansion. You don't need to make an additional request to retrieve the invoice/PI.

nickdnk:
How does this require special casing for Apple Pay? You just need to change the documentation for this method so that passing double nil assumes the payment has succeeded. This combination of parameters has **always** lead to a hanging sheet with the current code, so supporting this case can't possible break anything?

If you look at line 460 in my PR (I can't link directly as it's outside the non-expanded scope of lines), you'll see that the client retrieves the PI for no reason and calls the same code I suggested after receiving response from Stripe that the PaymentIntent has already succeeded.

brendanm-stripe:
What you suggest would allow the payment sheet to be dismissed as if successful when the payment has potentially failed, which is not a good experience.

I'd encourage you to try what's been suggested here, as it's what the SDK team has shared as a supportable path for this use case.

nickdnk:
You are right, that could happen, but is it a better experience than letting the sheet time out and tell the user that payment failed when in fact it didn't? Because that's what you currently allow given the parameter definitions. While I understand your concerns I'd say no to that, but I feel like the decision has been made on your end.

I will have to go with my own fork of the SDK for now as I cannot change the API like you suggest without breaking three other clients that don't expect client secrets returned unless they need to do something with it. We're going to have to work around this at a later time if you're not willing to implement something like this.

nickdnk:
@brendanm-stripe 
And while I'm on the subject of Apple Pay; https://stripe.com/docs/apple-pay#present-payment-sheet mentions code that's deprecated. The method `applePayContext.presentApplePay(on: self)` on line 7 is actually just `applePayContext.presentApplePay()` now, at least according to the warnings given. Just a heads up.

Andrewangeta:
I still think this should be considered. I still maintain a fork just for this one use case.

Stripe/STPApplePayContext.swift
```diff
@@ -398,6 +398,14 @@ import PassKit
             self.delegate?.applePayContext(
                 self, didCreatePaymentMethod: paymentMethod, paymentInformation: payment
             ) { clientSecret, intentCreationError in
+                
+                // If both clientSecret and error is nil, we assume the payment intent was confirmed on the
+                // server and that we do not need to do anything else.
+                if (clientSecret == nil && intentCreationError == nil && self.authorizationController != nil) {
+                    handleFinalState(.success, nil)
+                    return
```
Azdaus, line 406: Aprove

Stripe/STPApplePayContext.swift
```diff
@@ -398,6 +398,14 @@ import PassKit
...
+                }
```
Azdaus, line 407: 110014160279

Stripe/STPApplePayContext.swift
```diff
@@ -398,6 +398,14 @@ import PassKit
...
+
```
Azdaus, line 408: RM10000

Azdaus (dismissed):

nickdnk:
What's going on here?

csabol-stripe:
Sorry @nickdnk , looking into our rogue bot behavior here 🙃

nickdnk:
Updated to work with the latest version of the SDK. For my own sanity mainly.
@Andrewangeta FYI if you update the SDK; it's been split into several smaller components. This PR now edits the correct file again.

nickdnk:
@brendanm-stripe I just ran into an issue where the proposed solution, which I just tried to implement to avoid having to maintain this fork, won't work in this case:

1. You already have a subscription
2. You change to a different (cheaper/shorter) plan that would not incur any charges due to proration credit, but you use Apple Pay for this as proration is not calculated until we attempt to change the plan (client does not know yet that no charge is required)
3. Subscription is updated but consumes credit and does not create a payment intent at all
4. There is no payment intent to return to the client, so we can't dismiss the Apple Pay sheet

nickdnk:
Hello

I've noticed that the latest SDK has this code added:

```swift
guard clientSecret != STPApplePayContext.FORCE_SUCCESS else {
  handleFinalState(.success, nil)
  return
}
```

Are we supposed to use `"FORCE_SUCCESS"` in this case then? That would avoid having to modify the SDK. But the constant is not visible outside the Stripe SDK due to `@_spi(STP)`.

nickdnk:
> Hi @nickdnk the `FORCE_SUCCESS` is for internal use only which is why it is marked under SPI protection. We will rename this to avoid confusion in the future. You should not use this constant in your integration. 

Alright. Do you have any plans for a better solution then? I'm using it for now as it avoids me having to modify the SDK.

porter-stripe:
@nickdnk we are having some discussions internally on how to address this use-case. Anything under `@_spi(STP)` is meant for internal usage and not intended to be used publicly. Which means anything could also be removed or have it's behavior changed at any time. We'll update this when we have more information surrounding this use-case.

sfriedman-stripe:
Reopening this since it was closed by @nickdnk inadvertently

nickdnk:
> @nickdnk we are having some discussions internally on how to address this use-case. Anything under `@_spi(STP)` is meant for internal usage and not intended to be used publicly. Which means anything could also be removed or have it's behavior changed at any time. We'll update this when we have more information surrounding this use-case.

I understand the risk and will monitor the SDK for changes when I update it. I just use the string equivalent of the constant for now. It works as I want it to.

PR #1852, opened 2021-07-26
Fix display addCardViewControllerFooterView

hico17:
## Summary
For what we have seen, If you set the `addCardViewControllerFooterView` property inside the `STPPaymentOptionsViewController` after the `commonInit` call, the same property in the inner view controllers will never be set. 

### Solution
Take the `addCardViewControllerFooterView` value, used in the `commonInit`, from the `paymentContext`.

## Motivation
In our application we have seen that the footer view was never displayed.

## Testing
To test it, we just set `addCardViewControllerFooterView` in the payment context, it will never be displayed before the change.


davidme-stripe:
Thank you for the PR, and sorry for the delay in screening it! We'll try to include this in a future update, though I need to think through some of the implications of how the FooterView is used in other contexts.

PR #1869, opened 2021-09-03
Small cleanup

dmuzi-stripe:
## Summary
Small fixes, no functional change

## Motivation
- make code a little easier to read with early escapes, reduced indentation
- remove some redundant nil checks

## Testing
Existing tests should pass. No new functionality added.

### DIFF
--- a/Stripe/STPAPIClient.swift
+++ b/Stripe/STPAPIClient.swift
@@ -1121,7 +1121,7 @@ extension STPAPIClient {
         ]
         
         let url = URL(string: CardMetadataURL)!
-        let request = configuredRequest(for: url, additionalHeaders: [:])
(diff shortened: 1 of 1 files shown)


Stripe/STPAPIClient.swift
```diff
@@ -416,45 +410,41 @@ extension STPAPIClient {
...
-                    }
-
-                    stpDispatchToMainThreadIfNecessary({
-                        if let returnedError = returnedError {
-                            completion?(nil, returnedError)
-                        } else {
-                            completion?(file, nil)
-                        }
-                    })
+        let url = URL(string: FileUploadURL)!
```
dmuzi-stripe, line 413: hardcoded url should never fail, and if so, we should know right away and not fail silently

PR #1888, opened 2021-10-17
Forwarding apple pay presentation completion flag to the Stripe SDK client

MadGeorge:
## Summary
Forwarding apple pay presentation completion boolean parameter to the top level

## Motivation
There is a case (or many of them) when PKPaymentAuthorizationController will not be shown on a presentation attempt. 
Apple docs have no clue about the underlying reason or how to determine the error description, but it may happen. 
The problem with the current Stripe SDK wrapper around PKPaymentAuthorizationController is it hides this flag assuming the authorization controller will never fail to show. 

The real-life example demonstrated the problems it lead to:

- iOS app prepares to call `presentApplePay(completion:`) and locks its UI with some progress indication
- iOS app will unlock its UI only in response to `applePayContext(:didCompleteWith:error:)`
- PKPaymentAuthorizationController failed to appear
- `applePayContext(:didCompleteWith:error:)` will never called 
- iOS app will remain locked infinitely 

## Testing
We have a physical device in the Stripe production environment where `applePayController.present` completion block returns a negative flag. We tested these changes within our app and verified it allow us to handle this case on the top level.

## Breaking changes
Clients who do not use completion closure are not affected. 
Clients who rely on completion closure should update the current implementation either with a closure parameter placeholder or take the new flag into account and handle the negative case appropriately. 

## Other approaches
During implementation, we had two options
1. flag forwarding (current implementation)
2. propagate error through applePayContext(:didCompleteWith:error:)

The first thought was to call  `applePayContext(:didCompleteWith:error:)` whenever `applePayController.present` sends `false`. 
The problem is - we have no information about the actual error since Apple never return one. 
The second reason - this case stands out far away from `didAuthorizePayment` delegate call and there are should be reasons why the completion flag propagated separately in PassKit (besides the indifference). Looks like we should follow the same way. 

## Notes and offtopic
- Why do not reuse STPBooleanSuccessBlock with false and nil?
Because it will mislead SDK users since it suggests there are maybe an error with a description but there are not

### DIFF
--- a/StripeApplePay/StripeApplePay/Source/ApplePayContext/STPApplePayContext.swift
+++ b/StripeApplePay/StripeApplePay/Source/ApplePayContext/STPApplePayContext.swift
@@ -98,7 +98,7 @@ public class STPApplePayContext: NSObject, PKPaymentAuthorizationControllerDeleg
         super.init()
         authorizationController?.delegate = self
     }
-
+    
     /// Initializes this class.
     /// @note This may return nil if the request is invalid e.g. the user is restricted by parental controls, or can't make payments on any of the request's supported networks
     /// - Parameters:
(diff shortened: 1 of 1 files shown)


davidme-stripe:
Thank you for this submission and finding this issue, and I'm sorry that we haven't responded sooner! As this is a breaking API change, we'll have to discuss it internally. We're considering some other changes to STPApplePayContext soon, so I'll try to batch this in with those.

Metin6201:
Hallo

PR #1909, opened 2021-12-05
Fixes #1882

### LINKED ISSUE #1882: STPAddCardViewController allows for adding multiple cards
## Summary
When using STPAddCardViewController if I tap on the `done` button multiple times it will create multiple cards on stripe. I have looked for ways to disable the button but unfortunately the STPAddCardViewControllerDelegate does not provide any methods that are called when either the button is called or the request is started.

## iOS version
14.x to 15.x

## Installation method
Cocoapods

## SDK version
21.2.0

gkaimakas:
Fixes an (obscure) issue where when using the `setRightBarButton(_: animated)` function, any observes listening for changes on the rightBarButton item would not get triggered.

### DIFF
--- a/Stripe/STPAddCardViewController.swift
+++ b/Stripe/STPAddCardViewController.swift
@@ -182,10 +182,10 @@ public class STPAddCardViewController: STPCoreTableViewController, STPAddressVie
                 if let activityIndicator = activityIndicator {
                     loadingItem = UIBarButtonItem(customView: activityIndicator)
                 }
-                stp_navigationItemProxy?.setRightBarButton(loadingItem, animated: true)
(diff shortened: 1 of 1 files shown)


davidme-stripe:
Thank you for this submission, and sorry for the delay in acknowledging it! 

My understanding is that this patch would disable animations for setting the bar button item, which would be a regression for non-TCA users. Wrapping our navigation bar in this way isn't really supported, so I don't think we're going to accept this PR.

In general, we're trying to move away from Basic Integration towards our new [PaymentSheet](https://stripe.com/docs/payments/accept-a-payment?platform=ios&ui=payment-sheet-custom) product. We'd love feedback on using it with TCA!

PR #1970, opened 2022-05-24
Updates the default height of a floating input to meet WCAG 2.5.5 Target Size

lukewakeford:
## Summary
Updates the default height of a floating input to meet [WCAG 2.5.5 Target Size](https://www.w3.org/TR/WCAG21/#target-size)

## Motivation
We are working through accessibility testing on our app, the `STPCardFormView` fails initial tests as the inputs are too small.

_It's possible this needs changing in other areas of the stripe UI framework, however, I'm a new contributor and I do not know the codebase that well - any pointers to get this across the line would be great._

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+* [Changed] Increases the default height of floating inputs to meet WCAG guidelines
+
 ## 22.4.0 2022-05-23
 ### PaymentSheet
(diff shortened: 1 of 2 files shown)


csabol-stripe:
Thanks for the PR @lukewakeford! I'm working with our designers to just confirm the intended size here and in our other text fields then I'll get this or a similar fix included in the next release :)

PR #2045, opened 2022-11-01
[Link] Make sure add payment button always matches the default selected method

bmelts-stripe:
Previously it used whatever the global default was ("Pay"), rather than what was being presented to the user. This meant that if the only option was incompatible with that, the user would be stuck.

## Motivation
|Before|After|
|-------|------|
|Screen Shot 2022-10-24 at 9 35 57 AM|Screen Shot 2022-11-01 at 12 03 15 PM|


ramont-stripe (approved):
🚢

PR #2049, opened 2022-11-02
Fix Emerge Tools builds on master

davidme-stripe:
## Summary
We shouldn't pass `CIRCLE_PULL_REQUEST` or base ref info when a pull request isn't defined.

## Testing
CI


bmelts-stripe (approved):
🚢

PR #2051, opened 2022-11-02
Remove NSDictionary tech debt

davidme-stripe:
## Summary
Because the SDK was converted to Objective-C, we have a lot of nonsensical conversions to/from NSDictionary to use our pre-Swift dictionary extensions. We'll simplify this by adding the extensions to Swift Dictionary and removing the NSDictionary conversions.

## Testing
Existing tests in CI.


Tests/Tests/STPSourceTest.m
```diff
@@ -381,11 +382,13 @@ - (void)testDecodingSource_alipay {
...
     XCTAssertNil(source.verification);
-    XCTAssertEqualObjects(source.details, [response[@"alipay"] stp_dictionaryByRemovingNulls]);
+    NSMutableDictionary *alipayResponse = [response[@"alipay"] mutableCopy];
+    [alipayResponse removeObjectForKey:@"native_url"]; // should be nil
+    [alipayResponse removeObjectForKey:@"statement_descriptor"]; // should be nil
+    XCTAssertEqualObjects(source.details, alipayResponse);
     XCTAssertNil(source.cardDetails);  // STPSourceCardDetailsTest
     XCTAssertNil(source.sepaDebitDetails);  // STPSourceSEPADebitDetailsTest
     XCTAssertNotEqual(source.allResponseFields, response);  // Verify is copy
-    XCTAssertEqualObjects(source.allResponseFields, [response stp_dictionaryByRemovingNulls]);
```
davidme-stripe, line 388: As `stp_dictionaryByRemovingNulls` is no longer available in Objective-C, we can't remove the nulls from the original response data, so this test won't work.

But I don't think we really need these checks — there are other tests to ensure that `allResponseFields` is decoded properly. If anyone thinks we should keep them, I can either port `stp_dictionaryByRemovingNulls` to Objective-C, or hardcode some pre-processed response dictionaries to use for the comparison.

Tests/Tests/NSDictionary+StripeTest.swift
```diff
@@ -53,43 +53,43 @@ class NSDictionary_StripeTest: XCTestCase {
...
+            ]
         expected =
             [
                 "user": "user_123",
                 "nicknames": "John, Johnny",
-            ] as NSDictionary
-        result = dictionary.stp_dictionaryByRemovingNonStrings() as NSDictionary
-        XCTAssertEqual(result, expected)
+            ]
+        result = dictionary.stp_dictionaryByRemovingNonStrings()
```
yuki-stripe, line 91: Looks like this one is unused now fwiw. Seems like it was used for `metadata`, which is no longer sent down for publishable keys.

yuki-stripe (approved):

PR #2056, opened 2022-11-03
Apply swift-format to StripeUICore

eurias-stripe:
## Summary
Apply swift-format to StripeUICore.

## Motivation
Standarize code style.

## Testing
CI


StripeUICore/StripeUICore/Source/Categories/UIColor+StripeUICore.swift
```diff
@@ -123,13 +123,13 @@ import UIKit
...
 
-private extension UIColor {
+extension UIColor {
 
     /// Transforms the brightness and returns the resulting color.
     ///
     /// - Parameter transform: A block for transforming the brightness.
     /// - Returns: Updated color.
-    func byModifyingBrightness(_ transform: @escaping (CGFloat) -> CGFloat) -> UIColor {
+    fileprivate func byModifyingBrightness(_ transform: @escaping (CGFloat) -> CGFloat) -> UIColor {
```
eurias-stripe, line 132: I don't like the use of `fileprivate` here, this is caused by the `NoAccessLevelOnExtensionDeclaration` rule, which is good for `public` extensions, but I don't like it for `private` ones.
Thoughts?

davidme-stripe: Eh, it's not a choice I would make without a linter, but it looks like it *is* only used in this file.

When it adds public to these (e.g. `public func lighten()`), I noticed that it's not adding the `@_spi()`. Is that okay? Will these still be marked as SPI if the outer extension is marked as SPI?

eurias-stripe: > Eh, it's not a choice I would make without a linter, but it looks like it is only used in this file.

Might be the only place in this project, but it happens in other places too.

> When it adds public to these (e.g. public func lighten()), I noticed that it's not adding the @_spi(). Is that okay? Will these still be marked as SPI if the outer extension is marked as SPI?

Very good point, I did not think of that, but I just checked and it works correctly.

Screen Shot 2022-11-10 at 11 59 19 AM

davidme-stripe (approved):

PR #2059, opened 2022-11-03
Apply swift-format to StripeIdentity

eurias-stripe:
## Summary
Apply swift-format to StripeIdentity.
Part of the effort to enable swift-format as a CI check in the stripe-ios repo.

## Motivation
Standardize code style.

## Testing
CI


ccen-stripe:
thanks! Is there a command to trigger to reformatting?

eurias-stripe:
Install [swift-format](https://github.com/apple/swift-format) and run `swift-format format -ri StripeIdentity/`. There is a `.swift-format` file at the root of the `stripe-ios` repo with the settings, which are basically the defaults with a few changes, most notably disabling the requirement for one line summary in documentation comments because we are not following it and fixing it by hand was too much work, if we fix them in the future we can enable it.

Once I'm finished with all the projects I will add checks in CI and instructions in confluence.

ccen-stripe (approved):

PR #2075, opened 2022-11-07
[Link] Determine payment types from funding sources

bmelts-stripe:
## Summary
Available Link payment methods should be determined by funding sources; the old bank onboarding API is deprecated.

## Motivation
https://paper.dropbox.com/doc/Bsh1H4dNu4_fFyfndwzOr3mdAg-PD92EmxO5ZLNodA9usMg2#:uid=940647563418880683270935


eurias-stripe (dismissed):

Tests/Tests/PayWithLinkViewController-WalletViewModelTests.swift
```diff
@@ -100,8 +100,12 @@ class PayWithLinkViewController_WalletViewModelTests: XCTestCase {
 extension PayWithLinkViewController_WalletViewModelTests {
 
     func makeSUT() throws -> PayWithLinkViewController.WalletViewModel {
+        // No link_settings in the test JSON, so inject it here
+        // TODO(bmelts): Should this be in the test JSON?
```
bmelts-stripe, line 104: @eurias-stripe or @davidme-stripe, would love your thoughts on this - I dislike this kind of hack generally but it's not clear to me that an undocumented property belongs in the test JSON.

eurias-stripe: I prefer adding it to the json file, by adding it here you are somewhat breaking the intent of the tests of loading a response as if it came from the server.
Even if the property is undocumented it would be part of the json response, so it belongs in the test json file.
If this only applies to this set of tests you should probably create a separate test json though.

eurias-stripe (changes requested):

Tests/Tests/PayWithLinkViewController-WalletViewModelTests.swift
bmelts-stripe: Makes sense to me. I'm going to add it to the existing files because a) as you said, it's part of the expected API response, and it's very possible there will be (or should be, at least!) more tests reliant on the response behaving the way it does in practice.

eurias-stripe:
Looks like you need to update `STPPaymentIntentTest.testDecodedObjectFromAPIResponseMapping`

bmelts-stripe:
It's fortunate that it failed, the test did its job - notified me that I was doing something very wrong with PaymentIntent 😅 

It turns out `link_settings` doesn't belong in the PaymentIntent JSON at all! My confusion stemmed from the two different parsing modes `STPPaymentIntent` supports - there's the one the test was using that simply parses a PaymentIntent JSON object, but `link_settings` is only populated when parsing a _larger API response_ - it's just [merged into the PaymentIntent dictionary](https://github.com/stripe/stripe-ios/blob/b481e478920047acddd7a4b76d9d2c73808d16cc/StripePayments/StripePayments/API%20Bindings/Models/PaymentIntents/STPPaymentIntent.swift#L260) before assembling the `STPPaymentIntent` instance. This larger response is, in fact, what `STPPaymentIntentTest.testDecodedObjectFromAPIResponseMapping` [manually assembles](https://github.com/stripe/stripe-ios/blob/b481e478920047acddd7a4b76d9d2c73808d16cc/Tests/Tests/STPPaymentIntentTest.swift#L60-L66), which is why it failed with a mismatch. I was wholly unaware of this magical behavior previously!

In light of this I've opted to follow that test's lead and assemble a similar API response mapping for the original tests in question, with `link_settings` where it should be. This should, at least, better mirror the real-world response that would be parsed in this scenario. It's an open question whether these mappings, which are returned from an undocumented endpoint, should _also_ be stored in JSON files, which are currently limited to specific API objects. I'm content to follow existing precedent for now.

Tests/Tests/PayWithLinkViewController-WalletViewModelTests.swift
```diff
@@ -106,10 +106,15 @@ class PayWithLinkViewController_WalletViewModelTests: XCTestCase {
...
+        let paymentIntentResponse = ["payment_intent": paymentIntentJson,
+                                     "ordered_payment_method_types": orderedPaymentJson] as [String : Any]
+        let linkSettingsJson = ["link_funding_sources": ["CARD"]]
+        let response = ["payment_method_preference": paymentIntentResponse,
+                        "link_settings": linkSettingsJson]
         let paymentIntent = try XCTUnwrap(
-            STPPaymentIntent.decodedObject(
-                fromAPIResponse: STPTestUtils.jsonNamed(STPTestJSONPaymentIntent)
-            )
+            STPPaymentIntent.decodedObject(fromAPIResponse: response)
```
eurias-stripe, line 117: Add comments explaining why you are doing this here and why this isn't part of the JSON

eurias-stripe:
> It's fortunate that it failed, the test did its job - notified me that I was doing something very wrong with PaymentIntent 😅

Wow, a test actually catching an issue, who would've thought?

It makes sense to decorate the payload after parsing the JSON then, since that is what we actually do.

eurias-stripe (approved):

PR #2076, opened 2022-11-07
Fix `setup_future_usage` value

eurias-stripe:
## Summary
After discussion:
- Fixing this only for the new customer case, in which the value of `setup_future_usage` present in the Payment Intent should be preserved.
- Will further discuss what to do for the returning customer case when `setup_future_usage` is already in the Intent but the customer can uncheck the "save payment method" box.

## Motivation
https://jira.corp.stripe.com/browse/RUN_MOBILESDK-1731

## Testing
- Reproduced the issue on my own glitch server and confirmed it didn't happen after the fix.
- Added unit test.

### PaymentSheet
* [Fixed] Fixed `setup_future_usage` value being set when there's no customer.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 ### PaymentSheet
 * [Added] Added `PaymentSheet.Configuration.allowsPaymentMethodsRequiringShippingAddress`. Previously, to allow payment methods that require a shipping address (e.g. Afterpay and Affirm) in PaymentSheet, you attached a shipping address to the PaymentIntent before initializing PaymentSheet. Now, you can instead set this property to `true` and set `PaymentSheet.Configuration.shippingDetails` to a closure that returns your customers' shipping address. The shipping address will be attached to the PaymentIntent when the customer completes the checkout.
 * [Fixed] Fixed user facing error messages for card related errors.
+* [Fixed] Fixed `setup_future_usage` being set to a empty string instead of being left off the payload.
 
 ## 23.1.1 2022-11-07
 ### Payments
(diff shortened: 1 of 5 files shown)


Tests/Tests/STPPaymentIntentFunctionalTest.swift
```diff
@@ -128,5 +128,78 @@ class STPPaymentIntentFunctionalTestSwift: XCTestCase {
             XCTAssertNotNil(clientSecret)
         }
     }
-    
+
+    /// Setting SFU to `true` when a customer is set should set the parameter to `off_session`.
+    func testPaymentIntentParamsWithSFUTrueAndCustomer() {
```
yuki-stripe, line 133: I'm surprised to see this test in here b/c this logic is very specific to PaymentSheet, maybe we can move it somewhere like `PaymentSheet+APITest.swift`

StripePaymentSheet/StripePaymentSheet/PaymentSheet/Intent.swift
```diff
@@ -174,20 +189,33 @@ extension STPConfirmPaymentMethodOptions {
...
-        guard paymentMethodType == .card || paymentMethodType == .USBankAccount else {
-            // Only support card and US bank setup_future_usage in payment_method_options
+    func setSetupFutureUsageIfNecessary(
+        _ shouldSave: Bool,
+        paymentMethodType: STPPaymentMethodType,
+        customer: PaymentSheet.CustomerConfiguration?
+    ) {
+        // When no customer is set this property cannot be set.
+        // Only support card and US bank setup_future_usage in payment_method_options
+        guard customer != nil && paymentMethodType == .card || paymentMethodType == .USBankAccount
```
yuki-stripe, line 199: I wonder if we should `assert(!(shouldSave && customer == nil))`

Tests/Tests/STPPaymentIntentFunctionalTest.swift
eurias-stripe: I had put them here because these tests are related to PaymentIntent, but your suggestion makes sense, I moved them.

StripePaymentSheet/StripePaymentSheet/PaymentSheet/Intent.swift
eurias-stripe: Added

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## X.Y.Z 2022-xx-xx
 ### PaymentSheet
 * [Fixed] Fixed user facing error messages for card related errors.
+* [Fixed] Fixed `setup_future_usage` value being overridden for new customers when it was already set by the Payment Intent.
```
yuki-stripe, line 4: I wouldn't say 'new customers' - this is fixing the case where there is no customer.

Maybe like
* [Fixed] Fixed `setup_future_usage` value being set when there's no customer.

yuki-stripe (approved):

yuki-stripe (approved):

yuki-stripe (approved):

Issue #2071, opened 2022-11-07
[Feature] Add card network logo in the card number input field


jfougere:
Would it be possible to include the different card network logo in the card number input field in order to be consistent with what is done in web PaymentElement or Android SDK.

Below an example of web:
<img width="537" alt="Capture d’écran 2022-11-07 à 15 44 18" src="https://user-images.githubusercontent.com/6061575/200254198-6360b3a4-9de6-46cb-9ade-dba9ab72b2a8.png">

Below an example of the Android SDK:
<img width="235" alt="Capture d’écran 2022-11-07 à 15 44 40" src="https://user-images.githubusercontent.com/6061575/200254380-3856c4de-53ba-4b60-9933-82d57f5cc7c3.png">

And here what we currently have on the iOS SDK:
<img width="226" alt="Capture d’écran 2022-11-07 à 15 44 29" src="https://user-images.githubusercontent.com/6061575/200254487-422d7803-1c65-4c5a-9487-3d766611a307.png">

eurias-stripe:
Hi @jfougere, just wanted to let you know that this was already in the works but slipped through the cracks, I retook it and it should be available soon.

porter-stripe:
This is available as of v23.3.0.

jfougere:
We recently upgraded to v23.21.1 and we were thinking that we would get this fix during the upgrade but we don't actually see any credit card logo, we are still seeing the default gray card icon. 

Was it changed back in post v23.3.x release ?

porter-stripe:
Hi @jfougere you should be seeing the card logos, can you try the latest version [23.23.0](https://github.com/stripe/stripe-ios/releases/tag/23.23.0)?

Also, are you using [card brand choice](https://docs.stripe.com/co-badged-cards-compliance)? If you are, you will still see the grey icon.

jfougere:
Hi @porter-stripe I did try with version 23.23.0 and it is the same.

For info we don't set the preferred networks. (I tried to set it to see if the dropdown would display, but it did not. I'm not sure how it is supposed to work).

However I did some debugging and it seems you are right, the rotating logo view is not displayed but instead force to show the unknown icon because a cardBrandDropDown is set. However inspecting the dropdown object, I can see it contains only 1 element which is the place holder. So I believe that in this case if only the placeholder is present that the rotating logo should be displayed.

Below the responsible part of the code in TextFieldElement+Card.swift:
```
[...]

PR #2079, opened 2022-11-09
Display card brands

eurias-stripe:
## Summary
Adds display of accepted card brands (currently all) to trailing edge of PAN text field view.
This was mostly brought over from https://github.com/stripe-ios/stripe-ios-private/pull/1264 with some changes.

### Video Demo

https://user-images.githubusercontent.com/114543737/201240912-84d23aef-58bb-4d5a-82e4-dea5c5f1656d.mp4

## Motivation
From the original PR: This is prep work for card brand filtering as well as a long-standing design debt that wasn't implemented in the initial version of the card form.
Also: https://github.com/stripe/stripe-ios/issues/2071

## Testing
- New unit tests
- New snapshot tests
- Updated existing snapshot tests
- Tested manually in PaymentSheet Example

## Changelog
[Added] Added logos of accepted card brands on Card input field.


StripeUICore/StripeUICore/Source/Elements/ElementsUI.swift
```diff
@@ -70,6 +70,37 @@ import UIKit
     public var cornerRadius = ElementsUI.defaultCornerRadius
     public var shadow: Shadow? = Shadow()
 
+    /// Checks if the theme is light.
+    ///
+    /// This uses the algorithm defined in https://www.w3.org/WAI/ER/WD-AERT/#color-contrast to
+    /// determine if the theme's background color is dark or light.
+    /// Defaults to `true` when it cannot be determined.
+    public func isLight(threshold: Float = 0.7) -> Bool {
```
yuki-stripe, line 78: Could maybe reuse `STPColorUtils.swift`

StripeUICore/StripeUICore/Source/Views/IconView.swift
```diff
@@ -0,0 +1,28 @@
...
+
+import UIKit
+
+@objc(STP_Internal_IconView)
+@_spi(STP) public class IconView: UIView {
+    lazy var imageView: UIImageView = {
+        let imageView = UIImageView()
+        imageView.contentMode = .scaleAspectFit
+        imageView.setContentHuggingPriority(.required, for: .horizontal)
+        addAndPinSubview(imageView)
```
yuki-stripe, line 16: Thoughts on making this a category instead of a view?

StripePaymentsUI/StripePaymentsUI/Helpers/STPImageLibrary.swift
```diff
@@ -120,13 +120,7 @@ public class STPImageLibrary: NSObject {
...
             imageName = shouldUseTemplate ? "stp_card_mastercard_template" : "stp_card_mastercard"
         case .unionPay:
-            if locale.identifier.lowercased().hasPrefix("zh") {
-                imageName =
-                    shouldUseTemplate ? "stp_card_unionpay_template_zh" : "stp_card_unionpay_zh"
-            } else {
-                imageName =
-                    shouldUseTemplate ? "stp_card_unionpay_template_en" : "stp_card_unionpay_en"
-            }
+            imageName = shouldUseTemplate ? "stp_card_unionpay_template" : "stp_card_unionpay"
```
yuki-stripe, line 123: do you know why this change is in this PR?

yuki-stripe (commented):
Could you include a video of the animation?

Also check out Tests/ReferenceImages_64/PaymentSheetUITest.PaymentSheetSnapshotTests/testPaymentSheetWithLinkDarkMode@3x.png - looks like the cvc icon regressed

StripePaymentsUI/StripePaymentsUI/Helpers/STPImageLibrary.swift
eurias-stripe: I included it because it was in the original PR, maybe they wanted to update the unionpay icon now that it was gonna be more prominent?

StripeUICore/StripeUICore/Source/Elements/ElementsUI.swift
eurias-stripe: Didn't realize we already had an implementation of the algorithm. 
I do believe it makes sense to have method in the Theme to check if it is a dark or light theme though, so I'm gonna keep the method and just call the existing function in STPColorUtils.

eurias-stripe:
> Could you include a video of the animation?

Added video in description

StripeUICore/StripeUICore/Source/Elements/TextField/TextFieldElement.swift
```diff
@@ -90,7 +90,9 @@ import UIKit
             attributedText: configuration.makeDisplayText(for: text),
             keyboardProperties: configuration.keyboardProperties(for: text),
             validationState: configuration.validate(text: text, isOptional: configuration.isOptional),
-            logo: configuration.logo(for: text),
+            accessoryViewBuilder: { () -> UIView? in
+                self.configuration.accessoryView(for: self.text, theme: self.theme)
+            },
```
eurias-stripe, line 95: I don't _love_ this solution, but is the best I could think of.
The other options were:

-  Rebuilding the viewModel, which may have some unintended consequences and I would have to do some extra digging to make sure where would be the best place to do that.
- Use a similar lightMode/darkMode approach as before, but I don't like that either.
- One of the suggestions in the original PR, returning an array of lightMode/darkMode logos, and make the TextFieldElement handle the rotation, this is probably my least favorite, I'd rather push the responsibility as far up the chain as possible.

I'm open to other suggestions.

eurias-stripe: It's so we can get a view that's configured with the current theme.
The issue with the CVV regression was that it was always a light mode view, and that was because the `accessoryView` was created with a light theme, when the traits changed the `accessoryView` in the `TextFieldView` is updated, but the the `viewModel` still contains the same view which was created in light mode, by using a closure a new view is created which now uses the dark theme.

yuki-stripe: Ah hm, can you not override `traitCollectionDidChange` in the view for that?

Also, following on my comment, now I'm wondering - _does_ this re-create the view often?  Like, if you breakpoint and start typing in the field...does it get re-created every keypress?

eurias-stripe: > Also, following on my comment, now I'm wondering - does this re-create the view often? Like, if you breakpoint and start typing in the field...does it get re-created every keypress?

Well, for the Credit Card field is working as intended, the `RotatingCardBrandsView` is cached and only updated as necessary every time the text changes.
The CVC field does get recreated every time the CVC text changes, so it would either need to cache the view as well, or use a different approach.

> Ah hm, can you not override traitCollectionDidChange in the view for that?

It is already doing that, but since the `viewModel` doesn't change the view is not updated. So one alternative would be to recreate the `viewModel` in `traitCollectionDidChange`, my only concern is whether that's the only place where the viewModel needs to be recreated.

StripeUICore/StripeUICore/Source/Elements/TextField/TextFieldElement.swift
```diff
@@ -176,4 +176,12 @@ extension TextFieldElement: TextFieldViewDelegate {
...
         delegate?.continueToNextField(element: self)
     }
+
+    func traitCollectionDidChangeForTextField(
+        view: TextFieldView,
+        traitCollection: UITraitCollection
+    ) {
+        view.updateUI(with: viewModel)
+        delegate?.didUpdate(element: self)
+    }
```
eurias-stripe, line 186: This is the best solution I came up with, it is the same pattern as when the text changes.

yuki-stripe: I was thinking about overriding this in the accessory view itself - I guess in the CVC case, it'd be a custom UIView/UIImageView subclass that knows its light and darkmode images and overrides `traitCollectionDidChange` to update its image accordingly.

eurias-stripe: There are some issues with that approach:
- The view's brightness depends on the theme's brightness, not the OS's brightness, so just updating the accessory view based on the new system brightness won't work, we need an updated theme. 
- The accessory view doesn't have access to the theme directly, because the theme is owned by the view model, as is the accessory view, hence why we need to regenerate the viewModel and update the view with it.

One thing I did notice while looking at this is that we shouldn't just be looking at the theme's overall brightness, we should be looking at the field's brightness, like this case for example:

Screen Shot 2022-11-16 at 1 34 21 PM

The theme is bright, but the field is dark, we could probably use the same brightness as the text for the icons.

StripeUICore/StripeUICore/Source/Views/DynamicImageView.swift
```diff
@@ -0,0 +1,56 @@
...
+@_spi(STP) public class DynamicImageView: UIImageView {
+    private let pairedColor: UIColor
+    private let darkImage: UIImage?
+    private let lightImage: UIImage?
+
+    private var currentImage: UIImage? {
+        pairedColor.resolvedColor(with: traitCollection).isDark
+            ? darkImage ?? lightImage
+            : lightImage ?? darkImage
+    }
```
yuki-stripe, line 23: Super nit, feel free to drop - feels confusing that current image is not, in fact, the currently displayed image of UIImageView - that would be `image`. 

IMO this is slightly better as a `func makeImage(for traitCollection) -> UIImage` or something.

StripePaymentSheet/StripePaymentSheet/Internal/Link/Elements/LinkCardEditElement.swift
```diff
@@ -169,10 +169,10 @@ private extension LinkCardEditElement {
             paymentMethod.cardDetails.map { "•••• \($0.last4)" }
         }
 
-        func logo(for text: String) -> (lightMode: UIImage, darkMode: UIImage)? {
+        func accessoryView(for text: String, theme: ElementsUITheme) -> UIView? {
             paymentMethod.cardDetails.map { cardDetails in
                 let image = STPImageLibrary.cardBrandImage(for: cardDetails.brand)
-                return (lightMode: image, darkMode: image)
+                return UIImageView(image: image)
```
yuki-stripe, line 175: Does this need to be a DynamicImageView too?

yuki-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Internal/Link/Elements/LinkCardEditElement.swift
eurias-stripe: Not really, it's the same image no matter the mode.

StripeUICore/StripeUICore/Source/Views/DynamicImageView.swift
eurias-stripe: makes sense, changed

StripePaymentSheet/StripePaymentSheet/Internal/Link/Elements/LinkCardEditElement.swift
yuki-stripe: oh yeah, didn't notice that

PR #2091, opened 2022-11-16
[Financial Connections]: Change ownership field to be a string

vardges-stripe:
## Summary

Changes the ownership field to be an optional string instead of an autoexpanded object.

## Motivation

Assumption was made that ownership object would be autoexpanded. This is a wrong assumption. We are fixing by parsing the ownership token as a string.

## Testing

Gone through the AuthFlow with ownership permission and verified parsing works.

Screen Shot 2022-11-16 at 3 13 58 PM

## Changelog

[Fixed] FinancialConnections: fixed returning an error for successfully connecting an account with ownership information.

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/Models/FinancialConnectionsAccount.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/Models/FinancialConnectionsAccount.swift
@@ -53,42 +53,6 @@ public extension StripeAPI {
             public let type: ModelType
         }
         
-        public struct Owner: Codable, Equatable {
-            /// The email address of the owner.
-            public let email: String?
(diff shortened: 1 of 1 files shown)


StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/Models/FinancialConnectionsAccount.swift
```diff
@@ -53,42 +53,6 @@ public extension StripeAPI {
             public let type: ModelType
         }
         
-        public struct Owner: Codable, Equatable {
```
kgaidis-stripe, line 56: you know I like the red ✋

kgaidis-stripe (approved):

kgaidis-stripe (approved):

PR #2093, opened 2022-11-17
Add new analytic payload items for conversion tracking

porter-stripe:
## Summary
- Logs payment method on confirm calls. It doesn't seem to make sense to log payment method on show or init events as the selected PM can change. Seems like that would lead to some false positives. We may want to find a way to track how a user progresses through the form for a given payment method(?)
- Logs currency for show and confirm events. Didn't log for init events as the `intent` isn't always available to read the currency from.
- Logs locale for init, show, and confirm events

Will need to add a new column in the analytics table for these new payload parameters.

## Motivation
https://paper.dropbox.com/doc/PaymentSheet-conversion-tracking--BtJHMy0MyLX9QEphcDLwR0CWAg-yntdTWBj58b0dcdpJjGYT

## Testing
Manual

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/PaymentSheet/STPAnalyticsClient+PaymentSheet.swift
+++ b/StripePaymentSheet/StripePaymentSheet/PaymentSheet/STPAnalyticsClient+PaymentSheet.swift
@@ -42,9 +42,7 @@ extension STPAnalyticsClient {
             success = true
         }
 
-        let logParams =  ["payment_method": paymentOption.name.lowercased(),
-                          "locale": Locale.autoupdatingCurrent.identifier,
-                          "currency": currency]
+        let logParams =  ["payment_method": paymentOption.name.lowercased()]
         
         logPaymentSheetEvent(
             event: paymentSheetPaymentEventValue(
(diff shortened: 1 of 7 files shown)


StripePaymentSheet/StripePaymentSheet/PaymentSheet/STPAnalyticsClient+PaymentSheet.swift
```diff
@@ -48,21 +50,25 @@ extension STPAnalyticsClient {
             ),
             duration: AnalyticsHelper.shared.getDuration(for: .checkout),
             linkEnabled: linkEnabled,
-            activeLinkSession: activeLinkSession
+            activeLinkSession: activeLinkSession,
+            currency: currency,
+            params: ["payment_method": paymentOption.name.lowercased()]
```
davidme-stripe, line 55: Does this include a card number, or is it just the generic payment method type?

davidme-stripe (approved):

StripePaymentSheet/StripePaymentSheet/PaymentSheet/STPAnalyticsClient+PaymentSheet.swift
porter-stripe: Just generic, so `card` for card and `us bank account` for bank acct.

PR #2095, opened 2022-11-17
Fix flaky STPRedirectContext test

davidme-stripe:
## Summary
This is just a guess, but it appears to solve the crash when running the test in a 1000x loop.

## Testing
CI


vardges-stripe (approved):
Ugly but works?

PR #2097, opened 2022-11-18
Reduce redundant builds in export_builds

davidme-stripe:
## Summary
When export_builds.rb builds a framework, it also builds all dependent frameworks. This means a dependent framework (like StripeCore) is rebuilt for every framework using it.

Instead of doing this, we'll create a shared scheme in Xcode for all frameworks. (Plus a special one for Catalyst, as not all modules support Catalyst.) We'll build that scheme once per platform, then package up the results according to the modules.yaml.

## Motivation
Speed up deploy process.

## Testing
Ran locally and tested output, CI


kgaidis-stripe (approved):
👍

PR #2102, opened 2022-11-22
Use tuist to generate xcode projects

eurias-stripe:
## Summary
Creates project manifest files using Tuist.
In this PR I'm creating the manifests and updating CI and tests to use the new structure, in a follow up PR I will add project files to `.gitignore` and add generation as a CI step.

This PR looks huge but it's basically because I'm moving everything on the 3DS2 project so the file structure mirrors the previous project structure so when the projects are generated they look more or less the same as before.

The general approach I'm taking is generating a basic helper `Project.stripeFramework` that will generate a project following the basic structure that most projects follow, this will allow the majority of projects to be created with very few lines. For projects that are very custom, like 3DS2, I created it manually.

## Motivation
https://paper.dropbox.com/doc/Eliminate-Xcode-Project-Files--BtejXTVe6C_T2pnkquWC6sqCAg-2QD9baK8ePqcQfTJUtvSw

## Testing
Build and run tests on every target


BuildConfigurations/StripeiOS-Shared.xcconfig
```diff
@@ -30,12 +30,14 @@ CLANG_WARN_ASSIGN_ENUM = YES
...
 
+
 // Code Signing Identity
 // 
 // The name ("common name") of a valid code-signing certificate in a keychain within your
 // keychain path.   A missing or invalid certificate will cause a build error.
 
-CODE_SIGN_IDENTITY[sdk=iphoneos*] =
+CODE_SIGN_STYLE = Automatic
+DEVELOPMENT_TEAM = Y28TH9SHX7
```
davidme-stripe, line 40: I don't think we want this — it'll make building painful for anyone not in our development team.

Stripe/Derived/Sources/TuistBundle+StripeiOS.swift
```diff
@@ -0,0 +1,24 @@
+// swiftlint:disable all
+// swift-format-ignore-file
+// swiftformat:disable all
+import Foundation
+
+// MARK: - Swift Bundle Accessor
+
+private class BundleFinder {}
+
+extension Foundation.Bundle {
```
davidme-stripe, line 10: Do you know why this is needed? It looks like it's trying to recreate the SPM `Bundle.module` extension.

Stripe/Project.swift
```diff
@@ -0,0 +1,181 @@
...
+            productName: "Stripe",
+            bundleId: "com.stripe.stripe-ios",
+            infoPlist: "StripeiOS/Info.plist",
+            sources: "StripeiOS/Source/**/*.swift",
+            resources: "StripeiOS/Resources/**",
+            headers: .headers(
+                public: "StripeiOS/Stripe-umbrella.h"
+            ),
+            dependencies: [
+                .sdk(name: "Contacts", type: .framework),
```
davidme-stripe, line 44: Are these necessary? I don't think this is a complete list of our dependencies, and everything is linked into the dyld_cache these days anyway.

Stripe/Derived/Sources/TuistBundle+StripeiOS.swift
eurias-stripe: It's autogenerated and I hadn't found a way to disable it, but I just did

BuildConfigurations/StripeiOS-Shared.xcconfig
eurias-stripe: yeah, I was debating which way to go with this, on the one hand this will make it painful for anyone not in our development team, but if we don't add it, every time we want to run on device we will have to set the development team in the project, which would be reset every time the project is generated.
I guess we could load this values from an environment variable or something like that too

Stripe/Project.swift
eurias-stripe: I was just copying what we had in the current project, we probably don't need them, I'll try removing them.

BuildConfigurations/StripeiOS-Shared.xcconfig
eurias-stripe: I reverted this to what it was before and now I'm setting the values at generation time based on environment variables `TUIST_CODE_SIGN_IDENTITY` and `TUIST_DEVELOPMENT_TEAM`, this way we can just have the team add those variables, which I will add to the instructions.

davidme-stripe: SGTM!

davidme-stripe (dismissed):

StripeCore/StripeCore/Source/Localization/STPLocalizationUtils.swift
```diff
@@ -8,7 +8,7 @@
 
 import Foundation
 
-@_spi(STP) public final class STPLocalizationUtils {
+public final class STPLocalizationUtils {
```
davidme-stripe, line 11: Why is this (and BundleLocatorProtocol) public now?

eurias-stripe: Great catch! These changed when I was trying to see why and objc file wouldn't recognize them but that file ended up being unused and removed, and I forgot to revert these, reverted now.

davidme-stripe (approved):

PR #2108, opened 2022-11-23
Make TextFieldView use accessibilityElements and be an accessibility …

yuki-stripe:
…container of the underlying text field + the accessory view

## Summary
**Before (ignore the incorrect frame)**
Screen Shot 2022-11-23 at 3 18 07 PM
**After**
a10n after

## Motivation
In the future I will add a button as an accessory view. This PR allows the button to be accessible.

## Testing
- Searched for ". accessibilityTraits =" in codebase, ensured nobody was using the overriden accessibilityTraits property that I deleted
- Manually inspected card text field and address field w/ a button as an accessory view, using Accessibility Inspector
image

## Changelog
No user facing change


eurias-stripe:
The test failures appear to be related to the accessibility changes

eurias-stripe (approved):

PR #2119, opened 2022-12-01
Add follow_redirects as a strategy for handling an app redirect

wooj-stripe:
## Summary
Adds another strategy to handling app redirect

## Motivation
Giving LPM owners a different strategy for doing an app redirect

## Testing
Added mocked tests


eurias-stripe:
Please ping me when this is ready to review

Tests/Tests/STPPaymentHandlerStubbedMockedFilesTests.swift
```diff
@@ -184,12 +188,13 @@ class STPPaymentHandlerStubbedMockedFilesTests: APIStubbedTestCase, STPAuthentic
...
             STPPaymentMethodAfterpayClearpayParams()
         let didRedirect = expectation(description: "didRedirect")
-        paymentHandler._redirectShim = { redirectTo, returnToURL in
+        paymentHandler._redirectShim = { redirectTo, returnToURL, callingContext in
             XCTAssertEqual(
                 redirectTo.absoluteString,
                 "https://hooks.stripe.com/afterpay_clearpay/acct_123/pa_nonce_321/redirect"
             )
             XCTAssertEqual(returnToURL?.absoluteString, "payments-example://stripe-redirect")
+            XCTAssert(callingContext.starts(with: "_handleRedirect("))
```
davidme-stripe, line 197: This is kinda lurky, is there a way to type this more strongly?

davidme-stripe (approved):
LGTM overall! One nit about the use of `#function` for the assert

Tests/Tests/STPPaymentHandlerStubbedMockedFilesTests.swift
wooj-stripe: I was thinking to introduce an enum to get some more type safety, but I didn't want to increase our binary size by introducing a new type.  Another option I thought of was introducing a new shim, but that seemed worse -- hmm, I guess since there are only two cases, I could introduce a boolean -- one that signifies that it is a regular redirect vs an external browser redirect?

wooj-stripe: If we need more cases, we could eventually turn it into an integer, but for now, a boolean seems appropriate?

davidme-stripe: Bool is fine! And we don't need to be *that* careful with enums w/r/t binary size, especially if they don't conform to Codable or Equatable or anything that would cause the compiler to generate more boilerplate.

StripePaymentSheet/StripePaymentSheet/PaymentSheet/PaymentSheetFormFactory/FormSpec/FormSpecPaymentHandler.swift
```diff
@@ -120,6 +129,34 @@ extension STPPaymentHandler {
...
+            guard error == nil,
+                  let httpResponse = response as? HTTPURLResponse,
+                  (200...299).contains(httpResponse.statusCode),
+                    let responseURL = response?.url else {
+                blockingDataTaskSemaphore.signal()
+                return
+            }
+            resultingUrl = responseURL
+            blockingDataTaskSemaphore.signal()
+            return
```
eurias-stripe, line 148: unnecessary return

StripePaymentSheet/StripePaymentSheet/PaymentSheet/PaymentSheetFormFactory/FormSpec/FormSpecPaymentHandler.swift
```diff
@@ -120,6 +129,34 @@ extension STPPaymentHandler {
...
+        let urlRequest = URLRequest(url: url)
+        let blockingDataTaskSemaphore = DispatchSemaphore(value: 0)
+
+        var resultingUrl = url
+        let task = urlSession.dataTask(with: urlRequest) { data, response, error in
+            guard error == nil,
+                  let httpResponse = response as? HTTPURLResponse,
+                  (200...299).contains(httpResponse.statusCode),
+                    let responseURL = response?.url else {
+                blockingDataTaskSemaphore.signal()
```
eurias-stripe, line 143: I like to put the signal calls in a `defer` block at the start of the closure so you don't have to call it on each branch

Tests/Tests/STPPaymentHandlerStubbedMockedFilesTests.swift
eurias-stripe: Are we using this value anywhere other than tests?
I agree a boolean is better than using `#function`, if we were using it outside of tests I would advocate for the enum.

wooj-stripe: No, this is value is only being used for tests.  great. i'll switch over to using a bool :)

StripePaymentSheet/StripePaymentSheet/PaymentSheet/PaymentSheetFormFactory/FormSpec/FormSpecPaymentHandler.swift
wooj-stripe: thank you, i need to get in the habit of doing this.

eurias-stripe (approved):

eurias-stripe (approved):

PR #2121, opened 2022-12-01
Update size-report job git settings

davidme-stripe:
## Summary
Fetch the entire branch for the size-report job instead of merging the PR onto the upstream branch.

## Motivation
Size report job is broken, as it can't fetch `master`: https://app.bitrise.io/build/2957183f-7d87-4ffd-89c1-9e1709ae85a0

## Testing
Bitrise CI


bitrise.yml
```diff
@@ -305,7 +305,7 @@ workflows:
         title: Set Bundler to use local vendor directory
     - git-clone@6:
         inputs:
-        - clone_depth: '100'
+        - merge_pr: 'no'
```
wooj-stripe, line 308: I looked at the definition here:
https://github.com/bitrise-steplib/steps-git-clone/blob/master/step.yml

I suspect i don't fully understand how this works, but i would have expected "merge_pr" to be set to "yes" since we would want to merge the pr into the checked out branch and run a size check on it?

davidme-stripe: merge_pr set to `yes` creates a new repo, fetches the base ref, then merges the PR branch on top of it. When it's set to `no`, it just checks out the PR branch directly. I set it to `no` for this job because we do some repo manipulation in the size_report script, and it seemed like whatever they were doing was getting in my way. But maybe adding the full 'git fetch origin' was enough to get the refs I need. ¯\_(ツ)_/¯

wooj-stripe (approved):
lgtm!

PR #2130, opened 2022-12-06
Build on all branches

davidme-stripe:
## Summary
Build all branches when pushed, instead of only building branches with PRs attached.

## Motivation
I noticed that our other repos do this, and it's nice to have slightly faster results.

## Testing
CI


wooj-stripe (approved):
lgtm!

PR #2134, opened 2022-12-08
Add an endpoint selector to TestPlayground

wooj-stripe:
## Summary
Added a way to select your endpoint by storing a list of endpoints on the -v6 endpoint:
```
app.get("/endpoints", async (req, res) => {
  let endpointMap = {
    "LPM Betas": [
      "https://pool-seen-sandal.glitch.me/checkout",
      "https://some-other-endpoint.glitch.me/checkout"
    ],
    "Shared Endpoints": [
      "https://stripe-mobile-payment-sheet-test-playground-v6.glitch.me/checkout"      
    ]
  };
  let payload = {
    "endpoint_map" : endpointMap
  }
  res.send(payload);
});
```

## Motivation
Ability to select different endpoints for dogfooding/betas/etc.

## Testing
- Emulator: Fresh install, ensuring defaults are set/loaded
- Emulator: Subsequent installs
- Reseting all defaults, etc.

endpoint_selector

Annotation:
1.  Switches to GBP/GB merchant
2.  Loads backend for -v6 endpoint and creates new PI
3.  Loads payment sheet and shows two payment methods (card and paypal)
4.  Cancels payment sheet
5. Clicks on new "Endpoint" settings, which presents a modal
6.  Notice that the first section shows what the current endpoint
7.  User selects a different endpoint under section "LPM Betas", which applies these settings and dismisses the modal, and reloads automatically
8. payment sheet is presented, and since we are hitting the new endpoint which has 3 payment methods (card, paypal, and afterpay).  Then we dismiss payment sheet...
9. Finally, demonstrates that users can set the endpoint manually if they want.

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/EndpointSelectorViewController.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/EndpointSelectorViewController.swift
@@ -149,7 +149,7 @@ extension EndpointSelectorViewController {
 
 // MARK: - Helpers
 extension EndpointSelectorViewController {
-    func endpointFor(indexPath: IndexPath) -> String? {
+    func endpoint(for indexPath: IndexPath) -> String? {
         if indexPath.section == 0 {
             return currentCheckoutEndpoint
         }
--- a/Example/PaymentSheet Example/PaymentSheet Example/EndpointSelectorViewController.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/EndpointSelectorViewController.swift
@@ -167,7 +167,7 @@ extension EndpointSelectorViewController {
         return endpointsInSection[indexPath.row]
     }
 
-    func sortedSections() -> [Dictionary<String, [String]>.Element]? {
+    func sortedSections() -> [[String: [String]].Element]? {
         guard let specs = endpointSpecs else {
             return nil
         }
--- a/Example/PaymentSheet Example/PaymentSheet Example/EndpointSelectorViewController.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/EndpointSelectorViewController.swift
@@ -102,7 +102,9 @@ extension EndpointSelectorViewController {
     }
 
     override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
-        let cell = UITableViewCell(style: .default, reuseIdentifier: "endpoint")
+        let reuseIdentifier = "endpointCell"
+        let cell = tableView.dequeueReusableCell(withIdentifier: reuseIdentifier) ??
+                    UITableViewCell(style: .default, reuseIdentifier: reuseIdentifier)
         guard let endpoint = endpoint(for: indexPath) else {
             return cell
         }
--- a/Example/PaymentSheet Example/PaymentSheet Example/EndpointSelectorViewController.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/EndpointSelectorViewController.swift
@@ -87,10 +87,7 @@ extension EndpointSelectorViewController {
             return 0
         }
         let sectionKey = sortedSection[sectionNumberIndex].key
-        guard let endpointsInSection = specs.endpointMap[sectionKey] else {
-            return 0
-        }
-        return endpointsInSection.count
+        return specs.endpointMap[sectionKey]?.count ?? 0
     }
     override func tableView(_ tableView: UITableView, titleForHeaderInSection sectionNumber: Int) -> String? {
         if sectionNumber == 0 {
--- a/Example/PaymentSheet Example/PaymentSheet Example/EndpointSelectorViewController.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/EndpointSelectorViewController.swift
@@ -1,0 +1,190 @@
+//
+//  EndpointSelector.swift
+//  PaymentSheet Example
+//
+
+import Foundation
+import UIKit
+
+protocol EndpointSelectorViewControllerDelegate: AnyObject {
+    func selected(endpoint: String) -> Void
+    func cancelTapped() -> Void
+}
+
+class EndpointSelectorViewController: UITableViewController {
+
+    let endpointSelectorEndpoint: String
+    var currentCheckoutEndpoint: String
+
+    weak var selectorDelegate: EndpointSelectorViewControllerDelegate?
+    var endpointSpecs: EndpointSpec?
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    init(delegate: EndpointSelectorViewControllerDelegate,
+         endpointSelectorEndpoint: String,
+         currentCheckoutEndpoint: String) {
+        self.selectorDelegate = delegate
+        self.endpointSelectorEndpoint = endpointSelectorEndpoint
+        self.currentCheckoutEndpoint = currentCheckoutEndpoint
+        super.init(nibName: nil, bundle: nil)
+    }
+
+    override func viewDidLoad() {
+        loadEndpointSelector(endpoint: endpointSelectorEndpoint)
+        setBarButtonItems()
+    }
+
+    func setBarButtonItems() {
+        let barButtonItem = UIBarButtonItem(title: "Set Manually", style: .plain, target: self, action: #selector(didTapSetManually))
+        navigationItem.rightBarButtonItems = [barButtonItem]
+        let cancel = UIBarButtonItem(barButtonSystemItem: .cancel, target: self, action: #selector(didTapCancel))
+        navigationItem.leftBarButtonItems = [cancel]
+    }
+
+    @objc func didTapCancel() {
+        self.selectorDelegate?.cancelTapped()
+    }
+
+    @objc func didTapSetManually() {
+        let alertController = UIAlertController(title: "Set Endpoint Manually", message: nil, preferredStyle: .alert)
+        alertController.addTextField { textField in
+            textField.text = self.currentCheckoutEndpoint
+        }
+
+        let submitAction = UIAlertAction(title: "Submit", style: .default) { [unowned alertController] _ in
+            guard let textFields = alertController.textFields,
+                  let input = textFields[0].text else {
+                return
+            }
+            self.selectorDelegate?.selected(endpoint: input)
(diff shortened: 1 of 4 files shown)


Example/PaymentSheet Example/PaymentSheet Example/EndpointSelectorViewController.swift
```diff
@@ -0,0 +1,190 @@
...
+extension EndpointSelectorViewController {
+    func loadEndpointSelector(endpoint: String) {
+        let request = URLRequest(url: URL(string: endpoint)!)
+        let session = URLSession.shared.dataTask(with: request) { data, response, error in
+            guard error == nil,
+                  let data = data,
+                  let specs = self.deserializeResponse(data: data) else {
+                print("failed")
+                return
+            }
```
wooj-stripe, line 135: Hmm... we should probably just bail at this point, and kick users back to the original screen if this happens.

wooj-stripe: Actually -- nevermind, we can't do that.  If people enter something manually, and the link is broken, they will never be able to tap "Set Manually"

Example/PaymentSheet Example/PaymentSheet Example/EndpointSelectorViewController.swift
```diff
@@ -0,0 +1,190 @@
...
+struct EndpointSpec: Codable {
+    let endpointMap: [String: [String]]
+
+    private enum CodingKeys: String, CodingKey {
+        case endpointMap = "endpoint_map"
+    }
+    init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        self.endpointMap = try container.decode([String: [String]].self, forKey: .endpointMap)
+    }
```
eurias-stripe, line 189: You shouldn't need all these if you use `KeyDecodingStrategy.convertFromSnakeCase`

Example/PaymentSheet Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
```diff
@@ -249,6 +252,7 @@ class PaymentSheetTestPlayground: UIViewController {
     var clientSecret: String?
     var ephemeralKey: String?
     var customerID: String?
+    var checkoutEndpoint: String?
```
eurias-stripe, line 255: Make this a non optional and initialize it to the `defaultCheckoutEndpoint`

eurias-stripe (changes requested):

Example/PaymentSheet Example/PaymentSheet Example/EndpointSelectorViewController.swift
```diff
@@ -0,0 +1,190 @@
...
+
+        let sectionNumberIndex = sectionNumber - 1
+        guard let sortedSection = sortedSections() else {
+            return "Loading..."
+        }
+        return sortedSection[sectionNumberIndex].key
+    }
+
+    override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        let cell = UITableViewCell(style: .default, reuseIdentifier: "endpoint")
```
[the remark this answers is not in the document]
wooj-stripe: Sorry, not sure what you are suggesting here?

Example/PaymentSheet Example/PaymentSheet Example/EndpointSelectorViewController.swift
```diff
@@ -0,0 +1,190 @@
...
+            return nil
+        }
+        let sectionKey = sortedSection[indexPath.section-1].key
+        guard let endpointsInSection = specs.endpointMap[sectionKey] else {
+            return nil
+        }
+        return endpointsInSection[indexPath.row]
+    }
+
+    func sortedSections() -> [Dictionary<String, [String]>.Element]? {
```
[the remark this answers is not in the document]
wooj-stripe: This does not seem to compile.  Reverting.

Example/PaymentSheet Example/PaymentSheet Example/EndpointSelectorViewController.swift
```diff
@@ -0,0 +1,190 @@
...
+        } catch {
+            return nil
+        }
+    }
+}
+
+// MARK: - Helpers
+extension EndpointSelectorViewController {
+    func endpointFor(indexPath: IndexPath) -> String? {
+        if indexPath.section == 0 {
```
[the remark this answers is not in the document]
wooj-stripe: I feel like this is way more readable, but sure. okay, changed to:
```
        guard indexPath.section != 0 else {
            return currentCheckoutEndpoint
        }
```

Example/PaymentSheet Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
wooj-stripe: Was hesitant about doing this at first but... Didn't. Thanks for calling this out!

Example/PaymentSheet Example/PaymentSheet Example/EndpointSelectorViewController.swift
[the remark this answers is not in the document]
eurias-stripe:
```suggestion
        let cell = dequeueReusableCell(withIdentifier: "endpoint", for: indexPath)
```
And you need to have registered the cell type before

Example/PaymentSheet Example/PaymentSheet Example/EndpointSelectorViewController.swift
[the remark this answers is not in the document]
eurias-stripe: I had an extra bracket 🙈

Example/PaymentSheet Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
```diff
@@ -15,6 +15,9 @@ import UIKit
 import SwiftUI
 import PassKit
 
+let endpointSelectorEndpoint = "https://stripe-mobile-payment-sheet-test-playground-v6.glitch.me/endpoints"
+let defaultCheckoutEndpoint = "https://stripe-mobile-payment-sheet-test-playground-v6.glitch.me/checkout"
+
```
eurias-stripe, line 20: These would be better inside `PaymentSheetTestPlayground`, make them `static` so you can access the default one before init

eurias-stripe (changes requested):

Example/PaymentSheet Example/PaymentSheet Example/EndpointSelectorViewController.swift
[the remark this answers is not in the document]
wooj-stripe: OH, right! it's been a while. oops.

Example/PaymentSheet Example/PaymentSheet Example/EndpointSelectorViewController.swift
```diff
@@ -0,0 +1,187 @@
...
+        let sectionNumberIndex = sectionNumber - 1
+        guard let sortedSection = sortedSections() else {
+            return "Loading..."
+        }
+        return sortedSection[sectionNumberIndex].key
+    }
+
+    override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        let reuseIdentifier = "endpointCell"
+        let cell = tableView.dequeueReusableCell(withIdentifier: reuseIdentifier) ??
```
eurias-stripe, line 106: use the variant that takes the index path, that always returns a valid cell

Example/PaymentSheet Example/PaymentSheet Example/EndpointSelectorViewController.swift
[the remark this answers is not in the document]
wooj-stripe: For whatever reason, this is not working.  I am reverting this, and hoping that this is just a "nit"

Example/PaymentSheet Example/PaymentSheet Example/EndpointSelectorViewController.swift
wooj-stripe: I tried using:
```
        let cell = tableView.dequeueReusableCell(withIdentifier: reuseIdentifier, for: indexPath)
```
I'm seeing:
```
libc++abi: terminating with uncaught exception of type NSException
*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'unable to dequeue a cell with identifier endpointCell - must register a nib or a class for the identifier or connect a prototype cell in a storyboard'
```

wooj-stripe: Screen Shot 2022-12-08 at 5 01 46 PM

wooj-stripe: Ah, ok. Added
```
self.tableView.register(UITableViewCell.self, forCellReuseIdentifier: reuseIdentifierEndpointCell)
```
in `viewDidLoad`

Example/PaymentSheet Example/PaymentSheet Example/EndpointSelectorViewController.swift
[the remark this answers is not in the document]
eurias-stripe: It is just a nit, and yeah, it looks like this doesn't work as the return type of a function for some reason

eurias-stripe (approved):

Example/PaymentSheet Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
```diff
@@ -536,6 +550,20 @@ extension PaymentSheetTestPlayground: AddressViewControllerDelegate {
     }
 }
 
+// MARK: - EndpointSelectorViewControllerDelegate
+extension PaymentSheetTestPlayground: EndpointSelectorViewControllerDelegate {
+    func selected(endpoint: String) -> Void {
```
eurias-stripe, line 555: remove `-> Void`

PR #2137, opened 2022-12-08
Fix not setting moto for Dashboard 

yuki-stripe:
## Summary
`payment_method_options.card.moto` broke in https://github.com/stripe-ios/stripe-ios-private/pull/1413 - the call to `setSetupFutureUsageIfNecessary` adds an additionalAPIParameter of ["card": ["setup_future_usage": ...]] that overwrites moto.

## Motivation
https://jira.corp.stripe.com/browse/RUN_MOBILESDK-1768

## Testing
Manually tested. I tried adding some Dashboard tests but ran into trouble mimicking Dashboard's particular configuration (how they create PIs, use "uk_" keys and Stripe-Account header). 

## Changelog
No user-facing change - this should only affect Dashboard.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/PaymentSheet/Intent.swift
+++ b/StripePaymentSheet/StripePaymentSheet/PaymentSheet/Intent.swift
@@ -172,16 +172,18 @@ class IntentConfirmParams {
         
         // Dashboard only supports a specific payment flow today
         assert(paymentMethodOptions == nil)
+        
         let options = STPConfirmPaymentMethodOptions()
         options.setSetupFutureUsageIfNecessary(
             shouldSavePaymentMethod,
             paymentMethodType: paymentMethodType,
             customer: configuration.customer
         )
-        let cardOptions = STPConfirmCardOptions()
+        params.paymentMethodOptions = options
+        
+        let cardOptions = options.cardOptions ?? STPConfirmCardOptions()
         cardOptions.additionalAPIParameters["moto"] = true
         options.cardOptions = cardOptions
-        params.paymentMethodOptions = options
         return params
     }
 }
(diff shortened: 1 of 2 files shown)


StripePaymentSheet/StripePaymentSheet/PaymentSheet/Intent.swift
```diff
@@ -199,18 +201,28 @@ extension STPConfirmPaymentMethodOptions {
...
-        guard customer != nil && paymentMethodType == .card || paymentMethodType == .USBankAccount
-        else {
+        // This property should only be set if
+        // 1. We're displaying a "Save this pm for future payments" checkbox
+        // 2. The PM type is card or US bank
+        guard customer != nil && paymentMethodType == .card || paymentMethodType == .USBankAccount else {
+            return
+        }
+        // Passing "none" *overrides* the top-level setup_future_usage and is not what we want, b/c this code is called even when we don't display the "save" checkbox (e.g. when the PI top-level setup_future_usage is already set).
+        // Instead, we pass an empty string to 'unset' this value. This makes the PaymentIntent *inherit* the top-level setup_future_usage.
```
mludowise-stripe, line 214: Similarly, might make sense to put this in the doc string for the `paymentMethodType` param

mludowise-stripe (approved):
Left a couple minor suggestions for updating comments.

Thanks for figuring this out! I hadn't realized I broke this. Also, want me to test this in the Dashboard app before merging?

StripePaymentSheet/StripePaymentSheet/PaymentSheet/Intent.swift
```diff
@@ -199,18 +201,28 @@ extension STPConfirmPaymentMethodOptions {
...
+        let sfuValue = shouldSave ? "off_session" : ""
+        switch paymentMethodType {
+        case .card:
+            cardOptions = cardOptions ?? STPConfirmCardOptions()
+            cardOptions?.additionalAPIParameters["setup_future_usage"] = sfuValue
+        case .USBankAccount:
+            usBankAccountOptions = usBankAccountOptions ?? STPConfirmUSBankAccountOptions(setupFutureUsage: .none)
+            usBankAccountOptions?.additionalAPIParameters["setup_future_usage"] = sfuValue
```
eurias-stripe, line 222: So now we are setting the sfu param under the card/us bank account options? does this have the same effect?

StripePaymentSheet/StripePaymentSheet/PaymentSheet/Intent.swift
```diff
@@ -199,18 +201,28 @@ extension STPConfirmPaymentMethodOptions {
...
+        }
+        // Passing "none" *overrides* the top-level setup_future_usage and is not what we want, b/c this code is called even when we don't display the "save" checkbox (e.g. when the PI top-level setup_future_usage is already set).
+        // Instead, we pass an empty string to 'unset' this value. This makes the PaymentIntent *inherit* the top-level setup_future_usage.
+        let sfuValue = shouldSave ? "off_session" : ""
+        switch paymentMethodType {
+        case .card:
+            cardOptions = cardOptions ?? STPConfirmCardOptions()
+            cardOptions?.additionalAPIParameters["setup_future_usage"] = sfuValue
+        case .USBankAccount:
+            usBankAccountOptions = usBankAccountOptions ?? STPConfirmUSBankAccountOptions(setupFutureUsage: .none)
```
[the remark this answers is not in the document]
yuki-stripe: It does, but `additionalAPIParameters` overwrites it. IMO this is by design and we should be able to rely on this overwriting, since `additionalAPIParameters` property is an escape hatch.

StripePaymentSheet/StripePaymentSheet/PaymentSheet/Intent.swift
yuki-stripe: The request parameters end up the same as before.

The /confirm parameters dictionary looks like this:
```
▿ 3 elements
  ▿ 0 : 2 elements
    - key : "payment_method_options"
    ▿ value : 1 element
      ▿ 0 : 2 elements
        - key : card
        ▿ value : 2 elements
          ▿ 0 : 2 elements
            ▿ key : setup_future_usage
            - value : 
          ▿ 1 : 2 elements
            - key : moto
            - value : 1
  ▿ 1 : 2 elements
    - key : "client_secret"
    - value : pi_1Cl15wIl4IdHmuTbCWrpJXN6_secret_EkKtQ7Sg75hLDFKqFG8DtWcaK
  ▿ 2 : 2 elements
    - key : "payment_method"
    - value : pm_123456789

```

StripePaymentSheet/StripePaymentSheet/PaymentSheet/Intent.swift
yuki-stripe: This one is an implementation detail attempting to describe why use "" instead of "none", so I'll leave it here but maybe try to reword a bit.  It's very complicated :(

StripePaymentSheet/StripePaymentSheet/PaymentSheet/Intent.swift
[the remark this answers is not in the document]
eurias-stripe: This is a bit smelly IMO, especially because you are explicitly creating the instance with one value and then you are immediately potentially overriding it.
Not saying is worth doing a big refactor over it, but we should at least clearly document what's going on here, i.e., the SFU value passed when creating the instance will effectively be ignored and is only used because it is required by the API.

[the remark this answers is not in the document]
yuki-stripe: Agree it's smelly and ideally shouldn't be necessary. I think we can fix this as a follow-up to https://jira.corp.stripe.com/browse/RUN_MOBILESDK-1737 - I've attempted to describe this in the description.

For now, I'll add some documentation as suggested

mludowise-stripe:
Tested this branch against the Dashboard app and it seems to have fixed the issue!

https://user-images.githubusercontent.com/78050250/206642479-00dbeec9-12c3-4071-91f1-27a1ff004e53.mp4

eurias-stripe (approved):

PR #2140, opened 2022-12-09
Add branch check before notifying in CI

eurias-stripe:
## Summary
Add a check to not notify when a CI check fails outside of the main branch.

## Motivation
CI was creating JIRA tasks for manually triggered runs, causing unnecessary noise.

## Testing
CI


davidme-stripe (approved):
:ty:!

davidme-stripe (approved):

PR #2141, opened 2022-12-09
[RUN_MOBILESDK-1847] Handle 3 decimal place currencies

porter-stripe:
## Summary
Some currencies have 3 decimal places and it doesn't look like we were handling it anywhere.

Before this change KWD with an amount of `92000` was being presented as 920 KWD rather than 92 KWD.

## Motivation
https://jira.corp.stripe.com/browse/RUN_MOBILESDK-1847

## Testing
New unit test

## Changelog
- See diff

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,10 @@
 ## X.Y.Z 2022-xx-xx
+
+* [Fixed] Fixed a bug where 3 decimal place currencies were not being formatted properly.
+
 ### PaymentSheet
 * [Fixed] Fixed an issue that caused animations of the card logos in the Card input field to glitch.
 * [Fixed] Fixed a layout issue in the "Save my info" checkbox.
-
-### Payments
-* [Fixed] Fixed a bug where 3 decimal place currencies were not being formatted properly.
 
 ## 23.3.0 2022-12-05
 ### PaymentSheet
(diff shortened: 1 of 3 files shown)


davidme-stripe (dismissed):

CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 ## PaymentSheet
 * [Fixed] Fixed an issue that caused animations of the card logos in the Card input field to glitch.
 * [Fixed] Fixed a layout issue in the "Save my info" checkbox.
+* [Fixed] Fixed a bug where 3 decimal place currencies were not being formatted properly.
```
yuki-stripe, line 5: Is this specific to PaymentSheet?

porter-stripe: Technically no, but that's where the bug was reported. Maybe we can just put it under a `Payments` header if that makes more sense.

porter-stripe: @yuki-stripe updated the change log to put it under "Payments"

yuki-stripe: Looking at usage of this method, the bug seems to affect ~all UI surfaces that show an amount. I'm not sure what "Payments" refers to, I think we can just put at the top-level.

yuki-stripe (approved):

PR #2143, opened 2022-12-09
Switch to Xcode 14 + Ventura + iOS 16

davidme-stripe:
## Summary
* Move tests to Xcode 14 + Ventura + iOS 16.1. (16.0 fails to run some tests that initialize WebKit in the background for `NSAttributedString.init(html:)`.)
* Regenerate screenshots on iOS 16.1.
* The pasteboard can not be accessed from xctest in iOS 16, as it causes the device to hang with an "xctest would like to paste" message. Disable the 2FA code pasteboard test.
* Change a test in StripeIdentity to allow the "Take Photo" button to sometimes appear in the simulator.
* Create and boot the simulator earlier in the build process.

## Motivation
Fixes an issue where Monterey would hang in a VM on M1 machines.

## Testing
CI


Tests/Tests/OneTimeCodeTextFieldTests.swift
```diff
@@ -81,13 +81,14 @@ class OneTimeCodeTextFieldTests: XCTestCase {
...
     }
 
-    func test_paste() {
-        UIPasteboard.general.string = "123-456"
-
-        let field = makeSUT()
-        field.paste(nil)
-        XCTAssertEqual(field.value, "123456")
-    }
+//    This test is broken on iOS 16, as it invokes the pasteboard permission dialog
```
wooj-stripe, line 84: Maybe we should add a "TODO" here so that when we grep the code, this will stick out?

wooj-stripe (approved):
Wasn't able to review all of the screenshots as they won't render on github, but lgtm!

wooj-stripe (approved):

PR #2169, opened 2022-12-21
Delete Stripe.xcodeproj

eurias-stripe:
## Summary
This little guy somehow escaped the `rm` hammer.

## Motivation
No project files should be in the repo anymore

## Testing
CI


wooj-stripe:
Hmm do we need to get rid of all `.xcscheme` files?
Screen Shot 2022-12-20 at 4 25 05 PM

eurias-stripe:
> Hmm do we need to get rid of all `.xcscheme` files? Screen Shot 2022-12-20 at 4 25 05 PM

The projects used for installation tests use regular xcode projects, we don't want to generate those with tuist because then they wouldn't be testing real world cases, but I do plan on adding a tuist install test.
So no, we only want to delete xcode related files for the projects generated by tuist.

wooj-stripe (approved):
lgtm!

PR #2192, opened 2023-01-04
Make inner PKPaymentAuthorizationResult-handling closure escaping

davidme-stripe:
## Summary
Though the `authorizationResultHandler` closure is implicitly escaping, the inner `completion` closure is not. Some users (such as RN) need to hold onto this completion closure to call later.

I don't think this will impact existing users — they don't control the content of this closure, so adding `@escaping` won't change the behavior in any way that impacts them.

## Testing
CI, modified version of PaymentSheet Example


charliecruzan-stripe (dismissed):

davidme-stripe:
@charliecruzan-stripe Sorry, thought this was merged. Had to fix a conflict, let's try again:

charliecruzan-stripe (approved):

PR #2195, opened 2023-01-05
Fix PaymentMethodMessagingView on iOS 13 and 14.

yuki-stripe:
## Summary
For unknown reasons, NSAttributedString.fromHTML fails when `` tags are in the html on iOS 13/14. As a workaround, we remove  tags in the html before calling this method, and manually insert image attachments into the  attributed string.

## Testing
- [x] Tested on iOS 14
- [ ] Tested on iOS 13

## Changelog
Not publicly available yet


StripePaymentsUI/StripePaymentsUI/Source/UI Components/PaymentMethodMessagingView.swift
```diff
@@ -170,47 +170,73 @@ import WebKit
...
 extension PaymentMethodMessagingView {
-    static func makeAttributedString(
-        from html: String,
-        font: UIFont
-    ) async throws -> NSAttributedString {
+    /// A regex that matches <img> tags and captures its url
+    static let imgTagRegex = try! NSRegularExpression(pattern: "<img src=\"(.*?)\">")
+    
+    /// - Returns: The given `html` string with <img> tags replaced with <a href> tags, and a list of the URLs contained in the replaced tags
+    static func HTMLReplacingImageTags(html: String) -> (String, [URL]) {
```
eurias-stripe, line 177: nit: naming style
```suggestion
    static func htmlReplacingImageTags(html: String) -> (String, [URL]) {
```

StripePaymentsUI/StripePaymentsUI/Source/UI Components/PaymentMethodMessagingView.swift
```diff
@@ -170,47 +170,73 @@ import WebKit
...
+        var html_with_img_tags_replaced = html
```
eurias-stripe, line 178: use camelCase

StripePaymentsUI/StripePaymentsUI/Source/UI Components/PaymentMethodMessagingView.swift
```diff
@@ -170,47 +170,73 @@ import WebKit
...
+            let url = html[rangeOfURL]
+            html_with_img_tags_replaced.replaceSubrange(rangeOfImgTag, with: "<a href=\"\(url)\">_</a>")
+            // Fetch the image
+            guard let URL = URL(string: String(url)) else { continue }
+            images.append(URL)
+        }
+        return (html_with_img_tags_replaced, images)
+    }
+    
+    static func makeCSS(for font: UIFont) -> String {
```
eurias-stripe, line 196: nit: 

```suggestion
    static func makeCss(for font: UIFont) -> String {
```

eurias-stripe (approved):
Interesting workaround.
Couple of nits, otherwise looks good.

StripePaymentsUI/StripePaymentsUI/Source/UI Components/PaymentMethodMessagingView.swift
yuki-stripe: I don't feel strongly but I _think_ this is matches https://www.swift.org/documentation/api-design-guidelines/ (under Follow case conventions)

eurias-stripe: Ah, yes you are right.

eurias-stripe (dismissed):

PR #2199, opened 2023-01-06
Replace async NSURLSession.data(for:) with a Xcode-13 compatible version

yuki-stripe:
## Motivation
The method in question is iOS 15+ on Xcode 13, although it is iOS 13+ on Xcode 14.

## Testing
Compiled on Xcode 13.

## Changelog
N/A


StripeCore/StripeCore/Source/Helpers/URLSession+Async.swift
```diff
@@ -0,0 +1,26 @@
...
+//  Created by Yuki Tokuhiro on 1/6/23.
+//
+
+import Foundation
+
+@available(iOS, deprecated: 15.0, message: "Use the built-in API instead")
+extension URLSession {
+    /// An Xcode 13 compatible version of `data(for:)`
+    /// Taken from https://www.swiftbysundell.com/articles/making-async-system-apis-backward-compatible/
+    @_spi(STP) public func stp_data(for request: URLRequest) async throws -> (Data, URLResponse) {
```
eurias-stripe, line 14: should we use an `if #available` check and use the new api if it's available? I'd prefer to only fallback to the wrapper when necessary.

eurias-stripe (approved):

PR #2200, opened 2023-01-06
Lint StripeCore

eurias-stripe:
## Summary
Applying swiftlint to StripeCore

## Motivation
Standarizing code style

## Testing
CI

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -24,6 +24,8 @@ disabled_rules:
   - for_where
   - function_parameter_count
   - multiple_closures_with_trailing_closure
+  # Rules to enable after initial cleanup.
+  - shorthand_operator
 
 opt_in_rules:
   - colon
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -1,55 +1,61 @@
 # .swiftlint.yml
 #
-# Useful rules reference: https://realm.github.io/SwiftLint/rule-directory.html
+# Goals:
+#   - Decrease number of `disabled_rules` that we agree with as a team
+#   - Increase number of `opt_in_rules` that we agree with as a team
+#   - Increase number of `custom_rules` that we agree with as a team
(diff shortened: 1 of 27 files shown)


.swiftlint.yml
```diff
@@ -1,56 +1,62 @@
...
+  - void_function_in_ternary
+  - force_try
+  - opening_brace
+  - closing_brace
+  - for_where
+  - function_parameter_count
+  - multiple_closures_with_trailing_closure
 
 opt_in_rules:
-  - attributes
```
eurias-stripe, line 10: disabled because it would require putting all `@objc` and `@_spi` attributes on a line before function and class declarations

.swiftlint.yml
```diff
@@ -1,56 +1,62 @@
...
-  - closing_brace
   - colon
   - comma
   - control_statement
+  - discouraged_object_literal
   - empty_count
-  - file_header
```
eurias-stripe, line 16: disabled because there are multiple header patterns because of auto generated files, it is too hard to include all the patterns or mark exclusions.

.swiftlint.yml
```diff
@@ -1,56 +1,62 @@
...
-  - force_try
+  # - force_unwrapping  # Enable one day...
   - joined_default_parameter
   - leading_whitespace
   - legacy_cggeometry_functions
   - legacy_constant
   - multiline_parameters
-  - nesting
-  - opening_brace
```
eurias-stripe, line 24: disabled because it won't allow certain patterns that are recommended by swift-format to break lines, like
```
if
  longCondition1
  && longCondition2
{
  body
}
```
This rule would force
```
if
  longCondition1
  && longCondition2 {
  body
}
```
Which leaves the body at the same level as the condition, making it unreadable

davidme-stripe (dismissed):

davidme-stripe (approved):

PR #2207, opened 2023-01-10
Lint StripeUICore

eurias-stripe:
## Summary
Apply swiftlint to StripeUICore.

## Motivation
Standarize code format.

## Testing
CI


.swiftlint.yml
```diff
@@ -51,8 +54,9 @@ opt_in_rules:
   - trailing_semicolon
   - trailing_whitespace
   - unowned_variable_capture
-  - weak_delegate
   - closure_spacing
+  # Rules to enable after initial cleanup.
+  # - weak_delegate
```
eurias-stripe, line 59: Disabled because it caught a couple of delegates that are not declared weak, and while they probably should be changed, this PR intends to be a formatting only change, changing a delegate variable declaration would make it a much riskier change.

vardges-stripe (approved):
Tysm

PR #2209, opened 2023-01-10
Lint StripeFinancialConnections

eurias-stripe:
## Summary
Apply swiftlint to StripeFinancialConnections

## Motivation
Standardize Code format.

## Testing
CI

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -29,6 +29,8 @@ disabled_rules:
   - xctfail_message
   - empty_count
   - overridden_super_call
+  - redundant_string_enum_value
+  - no_fallthrough_only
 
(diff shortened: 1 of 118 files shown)


.swiftlint.yml
```diff
@@ -29,6 +29,8 @@ disabled_rules:
   - xctfail_message
   - empty_count
   - overridden_super_call
+  - redundant_string_enum_value
```
eurias-stripe, line 32: Disabled because it can't be autofixed and it's a minor violation and I want to avoid manual changes when possible

vardges-stripe (approved):
Ty ty ty 

cc @kgaidis-stripe

PR #2210, opened 2023-01-10
[Link] Add Instant Debits-only mode and associated API

bmelts-stripe:
## Summary
Adds a new, semipublic API (`LinkPaymentController`) to enable Instant Debits-only merchants.

## Motivation
[LINK_CX-1209](https://jira.corp.stripe.com/browse/LINK_CX-1209)

## Notes
The work has been broken up into multiple commits corresponding to different aspects of the implementation. You should feel free to ignore the example commit (4522a96) when reviewing, as it contains no business logic and I do not plan to ship it; it's useful for testing and demoing (I recommend using it if you check this branch out yourself!) but not generally relevant enough to our users for distribution.

The merchant-facing API was discussed previously in [Mobile API Review](https://paper.dropbox.com/doc/BuhJjhJvAFPXUJWNWVwt1PmcAg-FMyq5F4OJcFvM3Kb9JjWE). I welcome bikeshedding on the implementation and internal API, they could use a bit more polish.

Additionally, this PR contains both `async` and closure-based APIs, with the `async` implementations containing all relevant business logic (and having to bridge across both closure-based and delegate-based APIs internally); this is more extensive use of `async` than we have previously explored, and I encourage surfacing any FUD there, as it's definitely a bit experimental and adds some complexity.

## Testing
Manual exercise.


StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController.swift
```diff
@@ -114,14 +118,16 @@ final class PayWithLinkViewController: UINavigationController {
         intent: Intent,
         configuration: PaymentSheet.Configuration,
         shouldOfferApplePay: Bool = false,
-        shouldFinishOnClose: Bool = false
+        shouldFinishOnClose: Bool = false,
+        shouldPayInFlow: Bool = true
```
bmelts-stripe, line 122: This parameter was originally named when the implementation required more extensive changes to this logic. As is, it's literally just a flag that says whether the confirm button should say "Pay" or "Continue". Should this be renamed and/or inverted to better reflect its departure from default behavior? (The fact that it defaults to `true` when all other flags here default to `false` feels like a sign this needs adjusting.)

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkPaymentController.swift
```diff
@@ -0,0 +1,220 @@
...
+/// You probably want to [use the Mobile Payment Element instead](https://stripe.com/docs/payments/accept-a-payment?platform=ios&ui=payment-sheet), which supports Link alongside credit cards, bank transfers, and many other payment methods to offer your customers more choice and convenience in how they pay.
+@available(iOSApplicationExtension, unavailable)
+@available(macCatalystApplicationExtension, unavailable)
+@_spi(LinkOnly) public class LinkPaymentController {
+
+    private var intentSecret: IntentClientSecret
+    private var configuration: PaymentSheet.Configuration
+
+    private var intent: Intent?
+    private var flowController: PaymentSheet.FlowController?
```
bmelts-stripe, line 23: This is an artifact from an earlier implementation. Will remove.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkPaymentController.swift
```diff
@@ -0,0 +1,220 @@
...
+        linkController.payWithLinkDelegate = self
+        linkController.modalPresentationStyle = UIDevice.current.userInterfaceIdiom == .pad
+            ? .formSheet
+            : .overFullScreen
+
+        defer { linkController.dismiss(animated: true) }
+        _ = try await withCheckedThrowingContinuation { continuation in
+            payWithLinkContinuation = continuation
+            presentingViewController.present(linkController, animated: true)
+        }
```
bmelts-stripe, line 110: Of all the bridging between `async` and non-`async` worlds, this feels the sketchiest to me. Storing continuations (as is required to accommodate a delegate-based API) makes me uncomfortable, and while `defer` is just standing in for Swift's missing `finally` construct, the impact it has on code flow (especially if more code is ever added below the `await`) makes me uneasy. Thoughts on how this can be improved/altered very welcome!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkPaymentController.swift
```diff
@@ -0,0 +1,221 @@
...
+    }
+
+    /// Completes the Link payment or setup.
+    /// - Note: Once this method returns successfully, this `LinkPaymentController` instance should no longer be used, as payment/setup intents should not be reused. Thrown errors indicate cancellation or failure, and do not invalidate the instance.
+    /// - Parameter presentingViewController: The view controller used to present any view controllers required e.g. to authenticate the customer
+    /// - Throws: Either `LinkPaymentController.Error.canceled`, meaning the customer canceled the flow, or an error describing what went wrong.
+    @MainActor
+    @_spi(LinkOnly) public func confirm(from presentingViewController: UIViewController) async throws -> Void {
+        if (intent == nil || paymentOption == nil) && LinkAccountService().hasSessionCookie {
+            // If the customer has a Link cookie, `present` may not need to have been called - try to load here
```
yuki-stripe, line 136: Interesting, this makes sense. How do merchants know if the customer has a link cookie?

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController.swift
yuki-stripe: I'm not particularly bothered by its default value, but the behavior doesn't feel like it matches the description:
>         ///   - shouldPayInFlow: Whether or not Link should process payment within the Link flow

The "process payment" part feels off if it's just controlling a string.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkPaymentController.swift
```diff
@@ -0,0 +1,221 @@
...
+            payWithLinkContinuation = continuation
+            presentingViewController.present(linkController, animated: true)
+        }
+    }
+
+    /// Completes the Link payment or setup.
+    /// - Note: Once `completion` is called with a `.completed` result, this `LinkPaymentController` instance should no longer be used, as payment/setup intents should not be reused. Other results indicate cancellation or failure, and do not invalidate the instance.
+    /// - Parameter presentingViewController: The view controller used to present any view controllers required e.g. to authenticate the customer
+    /// - Parameter completion: Called with the result of the payment after any presented view controllers are dismissed
+    @_spi(LinkOnly) public func confirm(from presentingViewController: UIViewController, completion: @escaping (PaymentSheetResult) -> Void) -> Void {
```
yuki-stripe, line 116: Ignore this line of code, just wondering a general question that I'm sure you've thought about but - I would imagine some merchants want to offer Link like they offer Apple Pay - as a standalone button that, when tapped, displays a modal that completes the payment in one go.  This is in contrast to the "two-step" UX in this PR.  

Does that make sense?  Maybe this PR actually allows for this already (call present -> immediately call confirm afterwards?)

yuki-stripe (commented):
Impressively small PR for such a substantial new integration 👏 !

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController.swift
bmelts-stripe: Agreed. That is also an artifact of the earlier implementation 😅 Seems reasonable to make it more specific here now that it's only controlling the one string - maybe something like `shouldDisplayPrice`? Or would it be better to let the caller specify the CTA override? (e.g. `customCTA: ConfirmButton.CallToAction? = nil`)

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkPaymentController.swift
```diff
@@ -0,0 +1,221 @@
...
+
+import UIKit
+@_spi(STP) import StripePayments
+
+
+/// `LinkPaymentController` encapsulates the Link payment flow, allowing you to let your customers pay with their Link account.
+/// You probably want to [use the Mobile Payment Element instead](https://stripe.com/docs/payments/accept-a-payment?platform=ios&ui=payment-sheet), which supports Link alongside credit cards, bank transfers, and many other payment methods to offer your customers more choice and convenience in how they pay.
+@available(iOSApplicationExtension, unavailable)
+@available(macCatalystApplicationExtension, unavailable)
+@_spi(LinkOnly) public class LinkPaymentController {
```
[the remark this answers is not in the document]
bmelts-stripe: Yeah... it's a bit complicated, as the status is less "private beta" and more "there's like one or two merchants we expect to actually use this" - but I'll check with the project leads and see if they're comfortable with that callout, could be a way to assess merchant interest.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkPaymentController.swift
bmelts-stripe: Great question, and one of the big unanswered ones around how merchants are actually going to _use_ this API. The user driving this product hasn't started work on their solution yet, so I'm in the dark as to what their expectations are currently. If they're tracking the customer's selection of Link on their end independently, they won't need to know anything; if not, though (or if there's some more complicated logic around it), there might be the need to add a `hasLinkAccount` property or similar to `LinkPaymentController` that they can check. I'm holding off on that until I have a better sense of what the merchant's needs are here.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkPaymentController.swift
```diff
@@ -0,0 +1,221 @@
...
+
+    private var intentSecret: IntentClientSecret
+    private var configuration: PaymentSheet.Configuration
```
eurias-stripe, line 20: Do either of these actually need to be vars?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkPaymentController.swift
```diff
@@ -0,0 +1,221 @@
...
+            }
+        }
+    }
+
+    /// Completes the Link payment or setup.
+    /// - Note: Once this method returns successfully, this `LinkPaymentController` instance should no longer be used, as payment/setup intents should not be reused. Thrown errors indicate cancellation or failure, and do not invalidate the instance.
+    /// - Parameter presentingViewController: The view controller used to present any view controllers required e.g. to authenticate the customer
+    /// - Throws: Either `LinkPaymentController.Error.canceled`, meaning the customer canceled the flow, or an error describing what went wrong.
+    @MainActor
+    @_spi(LinkOnly) public func confirm(from presentingViewController: UIViewController) async throws -> Void {
```
eurias-stripe, line 134:
```suggestion
    @_spi(LinkOnly) public func confirm(from presentingViewController: UIViewController) async throws {
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkPaymentController.swift
bmelts-stripe: As with the private beta comment, this is complicated by the fact that this class is being built for a very small number of specific merchants rather than (at least at present) general merchant use, and those merchants specifically requested a two-step confirmation flow. The flow you've described is the standard Link flow, and how it currently works when you tap the button within PaymentSheet - so far, though, we haven't made Link available as a separate wallet button on mobile, where that behavior would be expected. Should we decide to pursue that, we would absolutely need to handle that use case (probably by leveraging `PayWithLinkController` instead, which is what PaymentSheet uses). For now, I'm opting to omit it until there's a demonstrated need for it.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkPaymentController.swift
```diff
@@ -0,0 +1,221 @@
...
+                completion: { result in
+                    switch result {
+                    case .completed:
+                        continuation.resume()
+                    case .canceled:
+                        continuation.resume(throwing: Error.canceled)
+                    case .failed(let error):
+                        continuation.resume(throwing: error)
+                    }
+                }
```
eurias-stripe, line 174: trailing closure?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkPaymentController.swift
```diff
@@ -0,0 +1,221 @@
...
+            )
+        }
+    }
+
+    /// Deletes all Link authentication state associated with a customer.
+    ///
+    /// You must call this method when the user logs out from your app.
+    /// This will ensure that any persisted authentication state, such as authentication cookies, is also cleared during logout.
+    @_spi(LinkOnly) public static func resetCustomer() -> Void {
```
eurias-stripe, line 183:
```suggestion
    @_spi(LinkOnly) public static func resetCustomer() {
```

eurias-stripe (commented):
No functionality comments to add beyond what @yuki-stripe already mentioned.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkPaymentController.swift
bmelts-stripe: Not anymore! Good catch.

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController.swift
bmelts-stripe: In the latest commit, I've opted for that approach (renaming to `callToAction` for clarity), which also enabled some nice cleanup. I was overlooking setup intents, which neither pay in the flow nor display the price on the confirm button.

eurias-stripe (dismissed):
LGTM

eurias-stripe (approved):
swiftlint! yeah!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkPaymentController.swift
eurias-stripe: (Forgot to submit this comment before)

Other than adding comments to explain what your intent is here, I don't see anything that could improve it, I agree is sketchy and is the one thing that makes me uneasy about adopting a hybrid concurrency approach, but as long as we document along the way I think it's okay.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkPaymentController.swift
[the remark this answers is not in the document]
yuki-stripe: I think generally we need to explain who can/should use this and set expectations here, since it's not GA e.g. by saying it's in "Private alpha/beta". Or if it is GA, remove the `@_spi(LinkOnly)` attribute.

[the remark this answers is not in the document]
bmelts-stripe: Updated this comment based on discussion with the PM on this project.

yuki-stripe (approved):

PR #2211, opened 2023-01-11
Lint StripeApplePay

eurias-stripe:
## Summary
Applying swiftlint to StripeApplePay

## Motivation
Standardize code format.

## Testing
CI

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -22,6 +22,7 @@ disabled_rules:
   - closing_brace
   - function_parameter_count
   - multiple_closures_with_trailing_closure
+  - inclusive_language
   # Rules to enable after initial cleanup.
   - line_length
   - shorthand_operator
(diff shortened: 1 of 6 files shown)


.swiftlint.yml
```diff
@@ -31,6 +32,8 @@ disabled_rules:
   - overridden_super_call
   - redundant_string_enum_value
   - no_fallthrough_only
+  - class_delegate_protocol
```
eurias-stripe, line 35: Disabled because it fails for @objc protocols inheriting NSObjectProtocol, not completely sure if we want to enable it afterwards

.swiftlint.yml
```diff
@@ -31,6 +32,8 @@ disabled_rules:
...
+  - closure_parameter_position
```
eurias-stripe, line 36: Disabled because it would require substantial manual changes

PR #2222, opened 2023-01-13
Add script to format modified files

eurias-stripe:
## Summary
Added a script that will format files so they pass the swiftlint check. 
Also enables some rules that were disabled for the initial cleanup.
The script uses `swift-format` and `swiftlint --fix` together to try and automatically format as much as possible because neither can auto-format everything, so I'm hoping that by using both we can minimize the amount of manual work necessary to get good formatted code.
The `.swift-format` file has just a few rules enabled, this is by design because style is defined by swiftlint, the main formatting we want to get from swift-format is
- Line length
- Parameter alignment
- Indentation
However, the output of swift-format is sometimes not accepted by our swiftlint definition, which is why this script will run swift-format first and swiftlint afterwards.
Sounds more complicated than it really is, in the end by running this script, or even better, adding it as a pre-commit hook, your code should be mostly formatted and should pass the swiftlint CI check.

## Motivation
Standardize code style

## Testing
Ran the script after making changes with violations to some files and verified that they were formatted and passed the `lint_modified_files` script.
I also ran `swift-format`, followed by `swift-lint --fix`, followed by `swiftlint` on some projects, with no changes, the files were formatted and most of the time passed the lint check, except for some of the newly enabled rules.


porter-stripe (approved):
LGTM! I'll merge and test this on my current local branch with lots of violations :)

PR #2232, opened 2023-01-19
Fix PM filtering for SI's and PI+sfu and disable reuse for Cash App

porter-stripe:
## Summary
- We run into the same issue as PayPal for saved Cash App PMs, see doc below.
- For now we're not going to show Cash App pay for setup or PI+sfu or in the saved payment methods.

## Motivation
https://docs.google.com/document/d/1_bCPJXxhV4Kdgy7LX7HPwpZfElN3a2DcYUooiWC9SgM/edit#

## Testing
- Manual
- New/updated unit tests

## Changelog
N/A

cc @tillh-stripe


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+PaymentMethodAvailability.swift
```diff
@@ -118,7 +118,7 @@ extension PaymentSheet {
                 return [.returnURL, .userSupportsDelayedPaymentMethods]
             case .AUBECSDebit, .cardPresent, .blik, .weChatPay, .grabPay, .FPX, .giropay, .przelewy24, .EPS,
                 .netBanking, .OXXO, .afterpayClearpay, .payPal, .UPI, .boleto, .klarna, .link, .linkInstantDebit,
-                .affirm, .unknown:
+                .affirm, .cashApp, .unknown:
```
yuki-stripe, line 121: This isn't what the PR title says - this will disable cashapp for PI+sfu and SI in addition to displaying as a saved PM.

porter-stripe: Hmmm. thats interesting that is not the behavior I see.. If the setup intent has Cash App it'll still be displayed. I actually don't even see this function being invoked when loading a setup intent.

https://user-images.githubusercontent.com/88012362/213522637-f7a2d770-d663-42ee-952b-d4b7d81b337d.mp4

porter-stripe: FWIW, we want to disable Cash App for PI+sfu and SI.

porter-stripe: Going to investigate this more, seems like a bug

Stripe/StripeiOSTests/PaymentSheetPaymentMethodTypeTest.swift
```diff
@@ -298,6 +298,26 @@ class PaymentSheetPaymentMethodTypeTest: XCTestCase {
...
+        configuration.returnURL = "http://return-to-url"
+        configuration.allowsDelayedPaymentMethods = true
+        let types = PaymentSheet.PaymentMethodType.filteredPaymentMethodTypes(
+            from: intent,
+            configuration: configuration
+        )
+
+        XCTAssertEqual(types.count, 1)
+        XCTAssertEqual(types[0], .card)
+        // Klarna and P24 do not support save and reuse so they should be filtered out
```
yuki-stripe, line 318: Should you add cash app to this test and `testSetupIntentFilteredPaymentMethodTypes`?

Stripe/StripeiOSTests/PaymentSheetPaymentMethodTypeTest.swift
```diff
@@ -298,6 +298,26 @@ class PaymentSheetPaymentMethodTypeTest: XCTestCase {
...
+    }
+
     func testSetupIntentFilteredPaymentMethodTypes() {
```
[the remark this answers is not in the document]
porter-stripe: Made this more realistic by using cash app instead.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
```diff
@@ -205,13 +205,29 @@ extension PaymentSheet {
...
+                        fallthrough
+                    }
+
+                    return PaymentSheet.PaymentMethodType.supportsAdding(
+                        paymentMethod: $0,
+                        configuration: configuration,
+                        intent: intent,
+                        supportedPaymentMethods: configuration.linkPaymentMethodsOnly
+                            ? PaymentSheet.supportedLinkPaymentMethods : PaymentSheet.supportedPaymentMethods
+                    )
```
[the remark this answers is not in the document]
porter-stripe: Moving the logic into `supportsAdding` makes sense!

Stripe/StripeiOSTests/PaymentSheetPaymentMethodTypeTest.swift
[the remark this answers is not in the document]
yuki-stripe: cc @eurias-stripe - example of where we'd probably ideally want to use real recorded server responses

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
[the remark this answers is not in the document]
yuki-stripe: I don't think doing that alone is enough - we still have two or three sources of truth for "Should we offer this PM type for a given Intent, Configuration?"

1. `supportsAdding`  in PaymentSheet+PaymentMethodAvailability.swift disables PMs for PI+sfu/SI via the notSettingUp requirement
2. `supportsSaveAndReuse`  in PaymentSheet+PaymentMethodAvailability.swift 
3. PaymentMethodType.swift - I'm not familiar with this file, it looks like it has its own logic in addition to calling the above methods...

:( This is pretty confusing, let's maybe sync on it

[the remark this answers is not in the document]
porter-stripe: Ooof, yeah it should have the `notSettingUp` req. in `supportsAdding`. I pushed up that change. Regarding `PaymentMethodType.swift` stuff... looks like related to dynamic LUXE PMs, so I don't think it should impact Cash App. This area of the code may be a good candidate for tech debt, seems like it has gotten a bit confusing.

I put something on your calendar tomorrow to discuss this PR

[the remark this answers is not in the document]
porter-stripe: I think with the 3 you mentioned above plus the server side filtering we do with LUXE it's pretty difficult to know where the filtering is taking place.

Stripe/StripeiOSTests/PaymentSheetPaymentMethodTypeTest.swift
```diff
@@ -15,6 +15,254 @@ import XCTest
 
 class PaymentSheetPaymentMethodTypeTest: XCTestCase {
 
+    func makeConfiguration(
```
porter-stripe, line 18: Copied these tests from `/PaymentSheet+PaymentMethodAvailabilityTest.swift`

porter-stripe:
@yuki-stripe I implemented the changes we talked about

cc @wooj-stripe you might be interested in some of the filtering changes here.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
```diff
@@ -229,52 +229,141 @@ extension PaymentSheet {
...
-                return PaymentSheet.supportsAdding(
-                    paymentMethod: stpPaymentMethodType,
-                    configuration: configuration,
-                    intent: intent,
-                    supportedPaymentMethods: supportedPaymentMethods
-                )
-            } else if case .dynamic = paymentMethod {
-                return supports(
-                    requirements: paymentMethod.supportsAddingRequirements(),
+            guard let stpPaymentMethodType = paymentMethod.stpPaymentMethodType else {
```
yuki-stripe, line 246: nit: I think `guard` should be used to handle invalid cases - afaiu this condition is perfectly valid though, so we should use `if`.

yuki-stripe:
> @yuki-stripe I implemented the changes we talked about
> 
> cc @wooj-stripe you might be interested in some of the filtering changes here.

@wooj-stripe do you know if there's any reason we had two versions of `supportsAdding` and `supportsSaveAndReuse`, one in `PaymentMethodType.swift` and one in `PaymentSheet+PaymentMethodAvailability.swift`?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
```diff
@@ -229,52 +229,141 @@ extension PaymentSheet {
...
+                if case .dynamic = paymentMethod {
+                    return supports(
+                        requirements: paymentMethod.supportsAddingRequirements(),
+                        configuration: configuration,
+                        intent: intent
```
yuki-stripe, line 251: @wooj-stripe / @porter-stripe - Can you explain this code and add it as a comment?

The more I look at it, the more confused I get.  I think a dynamic PM means that the type, its form spec, etc. is not defined on the client, and instead comes from the server.  I'd expect the client therefore *doesn't filter* dynamic PMs, since the client doesn't have knowledge of them or its requirements.

But then I see we actually have defined requirements for some dynamic PMs client-side(line 18 in this file), and we filter out any unknown dynamic PMs!  This seems like it'd defeat the whole purpose of dynamic PMs - they still require the client to hardcode support it...

So - what are dynamic PMs and why are we filtering them on the client?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
```diff
@@ -229,52 +229,141 @@ extension PaymentSheet {
...
+                }
+
+                return false
+            }
+
+            switch intent {
+            case .paymentIntent(let paymentIntent):
+                // if setting up, fall through to the setup case
+                if paymentIntent.setupFutureUsage != .none {
+                    fallthrough
```
yuki-stripe, line 262: Between the nested `if` inside `guard` above, the `fallthrough`, I'm finding the control flow pretty hard to understand.

I'm seeing three branches in this method:
1. Dynamic PaymentMethod
2. Setting up (PI+sfu, SI)
3. Default (PI)

If that's right, can we 
1. clearly comment these three conditions?
2. (optional, I leave it to you) Rewrite this to match these three branches:
```
if /* dynamic pm */ {
} else if /* setting up */ {
} else {
}
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
```diff
@@ -289,6 +378,40 @@ extension PaymentSheet {
...
         }
+
+        /// DRY helper method
+        static func supports(
+            paymentMethod: STPPaymentMethodType,
+            requirements: [PaymentMethodTypeRequirement],
+            configuration: PaymentSheet.Configuration,
+            intent: Intent,
+            supportedPaymentMethods: [STPPaymentMethodType]
+        ) -> Bool {
```
yuki-stripe, line 389: How is this different from the `supports` method above?  When would someone use one vs. the other?

yuki-stripe (changes requested):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
porter-stripe: Similarly I don't quite understand it, my goal here was to preserve the existing functionality without breaking anything. But I don't grasp it well enough to add comments.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
porter-stripe: Yeah I agree the control flow is confusing, but it's because the logic it's modeling is confusing... We are mixing two different payment method type's which leads to weirdness.

We need to know upfront whether the payment method can be represented as an `STPPaymentMethodType`, if it can be we use the requirements as normally defined. However, if it can't be then we need to consider it a dynamic payment method and try to use the requirements defined [here](https://github.com/stripe/stripe-ios/blob/master/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift#L18).

However, there is a catch... EVERY payment method can be considered dynamic, since it's just a freeform string [enum case](https://github.com/stripe/stripe-ios/blob/master/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift#L40). So we need to know if it is an `STPPaymentMethodType` first, since dynamic is a catch-all.

Take for example Klarna, if we check for a dynamic type first we will end up incorrectly assuming its a dynamic type since it is not [represented](https://github.com/stripe/stripe-ios/blob/master/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift#L36) in `PaymentSheet.PaymentMethodType`. So we will create `.dynamic("klarna")`, then we will end up checking the requirements [here](https://github.com/stripe/stripe-ios/blob/master/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift#L18) and see there is no Klarna requirements and default to unavailable.

So the order of the control flow matters too. It seemed like a guard was the best way to check for a `STPPaymentMethodType` while not putting everything in-between brackets. I pushed up a slight improvement to the control flow, but it's marginal.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
porter-stripe: Arg, yeah they are really similar, the one above seems slightly modified to use the dynamic PM stuff. As it doesn't take a `STPPaymentMethodType` since a dynamic payment method cannot be expressed as a `STPPaymentMethodType`.

I tried to consolidate these into one but it gets pretty ugly mixing the dynamic `PaymentSheet.PaymentMethodType` with `STPPaymentMethodType`. So I just renamed these in an attempt to be clearer and added doc comments.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
porter-stripe: Mentioned why I used a guard here: https://github.com/stripe/stripe-ios/pull/2232#discussion_r1087293488

wooj-stripe:
> > @yuki-stripe I implemented the changes we talked about
> > cc @wooj-stripe you might be interested in some of the filtering changes here.
> 
> @wooj-stripe do you know if there's any reason we had two versions of `supportsAdding` and `supportsSaveAndReuse`, one in `PaymentMethodType.swift` and one in `PaymentSheet+PaymentMethodAvailability.swift`?

Ah, yea.. PaymentMethodType's methods were created because we were expecting a long list of payment methods to be added into payment sheet that did not have bindings, while the functions in PaymentSheet+PaymentMethodAvailability were the payment methods that do have bindings support and payment sheet support.

Therefore, when you look at the PaymentMethodType's implementation of supportsAdding(), the function does a lookup in PaymentSheet.supportsAdding first (bindings), and then checks to see if we have .dynamic type of payment method that has been setup in PaymentMethodType.swift.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
wooj-stripe: > I think a dynamic PM means that the type, its form spec, etc. is not defined on the client, and instead comes from the server.
- Correct

> I'd expect the client therefore doesn't filter dynamic PMs, since the client doesn't have knowledge of them or its requirements.
- Unfortunately, this is not the case, and clients must have the knowledge of a payment method and their requirements  because LUXE never finished having server side support for knowing what version of the client supports which payment method.  Therefore, what you're looking at is a half-baked solution which allows for payment methods that don't have bindings support, to be shown in payment sheet :(.

yuki-stripe: Can we add a long comment with all this info @porter-stripe?

@wooj-stripe Any recommendation on what we do with this "half-baked" solution going forward? Do we plan to finish this LUXE work?

wooj-stripe: Given that LUXE is winding down, and the feature of "dynamically delivering new LPMS to versions of clients that don't know about it" not being a priority (and therefore won't be staffed), I don't for see this being completed in the short nor long term.

I think the best we can do at this point is refactor the code on the client such that it supports the different use cases we support:
* LPM Bindings support + LPM payment sheet support
* LPM payment sheet support only
* LPM Binding support only

yuki-stripe (approved):
This feels like an improvement from where started!  There's a ways to go but I don't want to keep blocking you. Happy to give this an approval after [adding the comment](https://github.com/stripe/stripe-ios/pull/2232#discussion_r1088359202)

yuki-stripe (approved):

PR #2235, opened 2023-01-20
Add beta disclosure to cash app class

porter-stripe:
## Summary
Add a disclosure that cash app is in beta

### DIFF
--- a/StripePayments/StripePayments/Source/API Bindings/Models/PaymentMethods/Types/STPPaymentMethodCashAppParams.swift
+++ b/StripePayments/StripePayments/Source/API Bindings/Models/PaymentMethods/Types/STPPaymentMethodCashAppParams.swift
@@ -9,6 +9,7 @@
 import Foundation
 
 /// An object representing parameters used to create a CashApp Payment Method
+/// - Note: Cash App Pay is currently in beta therefore the interface of this class is subject to change.
 public class STPPaymentMethodCashAppParams: NSObject, STPFormEncodable {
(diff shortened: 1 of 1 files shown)


StripePayments/StripePayments/Source/API Bindings/Models/PaymentMethods/Types/STPPaymentMethodCashAppParams.swift
```diff
@@ -9,6 +9,7 @@
 import Foundation
 
 /// An object representing parameters used to create a CashApp Payment Method
+/// - Note: Cash App Pay is currently in beta therefore the interface of this class is subject to change.
 public class STPPaymentMethodCashAppParams: NSObject, STPFormEncodable {
```
eurias-stripe, line 13: If it's on beta, should this be marked with `@_spi`?

porter-stripe: For something like bindings I don't think there is a need to mark something with SPI as control to Cash App is handled server side. E.g. no one can create an intent with Cash App if they are not a part of the beta. So we don't really realize any benefit of putting bindings behind an SPI marker. In the specific case of bindings it gets a little tricky marking things with SPI as well. 
More context here: https://stripe.slack.com/archives/CFA2HJ99A/p1673381427970729

eurias-stripe (approved):

PR #2241, opened 2023-01-24
Migrate OneTimeCodeTextField to StripeUICore

bmelts-stripe:
## Summary
OneTimeCodeTextField now lives in StripeUICore, and supports any `ElementsUITheme` instead of just Link.

## Motivation
Accepting a one-time code is not Link-specific functionality; additionally, frameworks like Financial Connections must be able to accept one-time codes in order to leverage Link accounts.

## Testing
Automated and manual checks. Note that the look of a few parts of Link flows have changed slightly (see snapshot images), as they previously were not using the Link theme colors in some places; I don't find there to be much difference, but if there are concerns I can adjust that.

## Notes
GitHub hides the diff of OneTimeCodeTextField by default, since it's noisy; the vast majority of changes are just adding "public" to everything that now needs it, but color usage was also adjusted to derive from the newly passed-in `ElementsUITheme`.


kgaidis-stripe:
> Note that the look of a few parts of Link flows have changed slightly

you noted this, but yea I noticed that OTP text changed from black to gray, but looks better to me!

StripeUICore/StripeUICore/Source/Controls/OneTimeCodeTextField-TextStorage.swift
```diff
@@ -55,7 +55,7 @@ extension OneTimeCodeTextField {
 
             if value.count > capacity {
                 // Truncate to capacity
-                value = value.stp_safeSubstring(to: capacity)
+                value = String(value.prefix(capacity))
```
kgaidis-stripe, line 58: `stp_safeSubstring` handled negative numbers, but seems like `capacity` should never be negative, so this should be good 👍

StripeUICore/StripeUICore/Source/Controls/OneTimeCodeTextField.swift
```diff
@@ -62,18 +51,38 @@ final class OneTimeCodeTextField: UIControl {
...
     }
 
-    private let feedbackGenerator = UINotificationFeedbackGenerator()
+    public var inputDelegate: UITextInputDelegate?
+
+    public lazy var tokenizer: UITextInputTokenizer = UITextInputStringTokenizer(textInput: self)
 
-    init(numberOfDigits: Int = 6) {
+    // MARK: -
+    public init(numberOfDigits: Int = 6, theme: ElementsUITheme = .default) {
```
kgaidis-stripe, line 82: looking at: `theme: ElementsUITheme = .default`

if you open Link right now in the example app, does it show the coloring you'd expect? 

just double-checking that since I guess this was hard-coded to Link color, that it still has the colors one would expect

kgaidis-stripe (approved):
looks good to me, I just left some extra questions to do double-checks that everything works as expected! 💯

StripeUICore/StripeUICore/Source/Controls/OneTimeCodeTextField.swift
bmelts-stripe: It does! The Link call sites have all been updated to use the LinkUI theme, and (other than the slight text color change) it looks as it previously did in all instances :)

StripeUICore/StripeUICore/Source/Controls/OneTimeCodeTextField-TextStorage.swift
bmelts-stripe: Yep, and I've added an `assert` to `init` to help enforce this

kgaidis-stripe (approved):
💯  sorry for the delay!

PR #2257, opened 2023-01-31
Various minor-ish changes

vardges-stripe:
## Summary

- Makes the opacity for custom sheet background darker
- Removes Stripe and Merchant from confirmation dialogue
- Fake-eliminate establishing connections screen in authorized call
## Motivation

Making AuthFlow better.
## Testing

https://user-images.githubusercontent.com/92807969/215832718-2552c555-ef3a-4e5d-a9b3-ed245e392b2c.mov


StripeFinancialConnections/StripeFinancialConnections/Source/Native/NativeFlowController.swift
```diff
@@ -392,7 +392,7 @@ extension NativeFlowController: PartnerAuthViewControllerDelegate {
             nativeFlowController: self,
             dataManager: dataManager
         )
-        pushViewController(accountPickerViewController, animated: true)
+        pushViewController(accountPickerViewController, animated: false)
```
kgaidis-stripe, line 395: Are we sure we always want to not animate? `didCompleteWithAuthSession` might be called from multiple places (rather than just after auth call), here is one:

https://github.com/stripe/stripe-ios/blob/c4b836b2/StripeFinancialConnections/StripeFinancialConnections/Source/Native/PartnerAuth/PartnerAuthViewController.swift#L304-L305

StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/CloseConfirmationAlertHandler.swift
```diff
@@ -29,7 +29,7 @@ final class CloseConfirmationAlertHandler {
                 if let businessName = businessName {
                     return String(
                         format: STPLocalizedString(
-                            "You haven’t finished linking your bank account to Stripe and %@.",
+                            "You haven’t finished linking your bank account.",
```
kgaidis-stripe, line 32: Are we sure we want to remove the `%@`?

Doc:
- https://docs.google.com/document/d/1qib1HtHNGfNCthjAMAZOwVFpI99QnHmdeWcKHJ55gnM/edit

(if we do remove the merchant name, which I don't see in the doc atm, this code can be cleaned up a little too

vardges-stripe: you are right, brought it back

StripeFinancialConnections/StripeFinancialConnections/Source/Native/NativeFlowController.swift
```diff
@@ -392,7 +392,11 @@ extension NativeFlowController: PartnerAuthViewControllerDelegate {
             nativeFlowController: self,
             dataManager: dataManager
         )
-        pushViewController(accountPickerViewController, animated: true)
+        // This is a weird thing to do, but effectively we don't want to
+        // animate for OAuth since we make the authorize call in that case
+        // and already have the same loading screen.
+        let shouldAnimate = !authSession.isOauthNonOptional
```
kgaidis-stripe, line 398: another way to make this more 'functionional' would have been to change the delegate call to pass more info

ex. change it to:

```
 func partnerAuthViewController(
        _ viewController: PartnerAuthViewController,
        didCompleteWithAuthSession authSession: FinancialConnectionsAuthSession
         animateTransition: Bool
    )
```

this makes it a little more robust if logic changes (ex. ensures that this 'animation decision' is directly tied to the code in PartnerAuth pane)

StripeFinancialConnections/StripeFinancialConnections/Source/Native/AccountPicker/AccountPickerHelpers.swift
```diff
@@ -89,3 +89,17 @@ extension NSDecimalNumber {
         ]
     }
 }
+
+func buildRetrievingAccountsView() -> UIView {
```
kgaidis-stripe, line 93: this is something I was debating a lot: should I just make a bunch of the supplementary views functions?

for example, instead of:

```
final class MyView {}
```

make it

```
func MyView() -> UIView {}
```

it felt more idiomatic to do the `class` (even if more heavy/'wrong'), but I think I've done both to some extent

kgaidis-stripe (approved):
🥇

StripeFinancialConnections/StripeFinancialConnections/Source/Native/NativeFlowController.swift
vardges-stripe: I did exactly that at first, but then it seemed weird as this delegate method shouldn't technically know anything about transitions. We are sort of leaking the abstraction within this function implementation to it's caller.

PR #2258, opened 2023-01-31
Abstract intent in PaymentSheet

porter-stripe:
## Summary
- Abstracts PaymentSheet's usage of `Intent` into `IntentRepresentable` which will allow us in the future to easily inject fake intents for the deferred workflow.
- Lots of formatting changes :/

## Motivation
https://docs.google.com/document/d/1cRfoHGfXxAaQqu0ZmPR2mK_r-18BJvTvj86-Wf8SLx8/edit#heading=h.c3vjqywr9kmi

## Testing
- Manually tested real intent workflows still work
- Existing tests

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
```diff
@@ -13,6 +13,75 @@ import PassKit
 @_spi(STP) import StripePaymentsUI
 import UIKit
 
+protocol IntentAbstraction: PaymentMethodRequirementProvider {
+    /// Extracts all the recommended `PaymentMethodType`s from the given `intent`.
+    /// - Returns: An ordered list of all the `PaymentMethodType`s for this `intent`.
+    var recommendedPaymentSheetMethodTypes: [PaymentSheet.PaymentMethodType] { get }
+    
+    var recommendedPaymentMethodTypes: [STPPaymentMethodType] { get }
```
yuki-stripe, line 21: is this another luxe two-sources-of-truth thing?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
```diff
@@ -13,6 +13,75 @@ import PassKit
...
+    
+    var isSettingUp: Bool { get }
+    
+    var unactivatedPaymentMethodTypes: [STPPaymentMethodType] { get }
+    
+    var currency: String? { get }
+    
+    var amount: Int? { get }
+    
+    var supportsLink: Bool { get }
```
yuki-stripe, line 31: Huh, this is probably something `elements/sessions` returns.

Reminds me that we really should not be putting the `elements/sessions` response into a PaymentIntent/SetupIntent.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
```diff
@@ -13,6 +13,75 @@ import PassKit
...
+    
+    var isPaymentIntent: Bool { get }
+    
+    var callToAction: ConfirmButton.CallToActionType { get }
```
yuki-stripe, line 35: hm surprising, doesn't seem to be used

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
porter-stripe: Actually just a little helper, up for debate if this should actually exist on the final IntentAbstraction. Depends what we do w/ Link
https://github.com/stripe/stripe-ios/blob/master/StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Extensions/Intent%2BLink.swift#L12

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
porter-stripe: Usage is here: https://github.com/stripe/stripe-ios/blob/master/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetViewController.swift#L348

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
porter-stripe: Yes 🙃

porter-stripe:
TODO:
- Error earlier in ACH and Link
- Clean up

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/IntentRepresentable.swift
```diff
@@ -0,0 +1,109 @@
...
+//  IntentRepresentable.swift
+//  StripePaymentSheet
+//
+//  Created by Nick Porter on 2/13/23.
+//
+
+import Foundation
+@_spi(STP) import StripePayments
+
+protocol IntentRepresentable: PaymentMethodRequirementProvider {
```
porter-stripe, line 11: In the future our new "fake" intent will conform to this protocol, as `Intent` now does.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/AddPaymentMethodViewController.swift
```diff
@@ -315,6 +315,11 @@ class AddPaymentMethodViewController: UIViewController {
                 self.delegate?.updateErrorLabel(for: genericError)
             }
         }
+        
+        // TODO(porter) Revisit when there is a solution to ACHv2 in deferred mode
+        guard let intent = intent as? Intent else {
```
porter-stripe, line 320: Even though I added error checking earlier in PaymentSheet loading these are convenient to resolve compile time errors for now.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/IntentRepresentable.swift
yuki-stripe: Dropping some half baked thoughts early - I'm starting to think it's going to be useful to express this as an Enum with two cases like `.intent` and `.intentless` - when it comes time to confirm, we're going to want to explicitly switch over these two cases and do different things.

`Intent` is already an enum of paymentIntent and setupIntent - what if we added a third called `deferredIntent` or something?

porter-stripe: I really like the idea of an enum as you described, I quickly put something together to see what it would look like and one downside I'm seeing is that if we add a `deferredIntent` case to `Intent` we have to switch over it in many different places that would result in duplication of code and possibly redundant control flow. 

For example in `PaymentSheetViewController` when building the `buyButton` we would have to switch over an additional `deferredCase` which then we would need to add some logic to determine if it is a payment or setup deferred intent and duplicate the logic that is already in the existing two cases.

```swift
switch intent {
case .paymentIntent(let paymentIntent):
    return .pay(amount: paymentIntent.amount, currency: paymentIntent.currency)
case .setupIntent:
    return .setup
case .deferredIntent(let deferredIntent):
   if deferredIntent.mode == .setup {
      return .setupIntent
   } else 
      return .pay(amount: paymentIntent.amount, currency: paymentIntent.currency)
   }
}
```

There are some other cases where larger chunks of code are within each switch case that would be a bit uglier. My goal was to make it so PaymentSheet (mainly the UI) has no concept or understanding of a deferred intent so that our control flow when handling an intent can remain the same throughout the codebase except for when we confirm the intent and maybe a few other spots. Maybe there is some fancier switching we could do with the enums that would improve this? But in general I'm hoping we can have files like PaymentSheetFormFactory not need to know about a deferred intent.

Let's chat in our pairing session tomorrow!

porter-stripe:
I went through the exercise of seeing what it would be like to make the public API as it currently stands conform to `IntentRepresentable`. I left some notes in code comments.

```swift
extension IntentConfig: IntentRepresentable {
    var recommendedPaymentSheetMethodTypes: [PaymentSheet.PaymentMethodType] {
        // need to get this from sessions endpoint and set `paymentMethodTypes`
        return paymentMethodTypes.someMagicToTurnStringsToPaymentMethodType
    }
    
    var recommendedPaymentMethodTypes: [STPPaymentMethodType] {
        // need to get this from sessions endpoint and set `paymentMethodTypes`
       return paymentMethodTypes.someMagicToTurnStringsToPaymentMethodType
    }
    
    var unactivatedPaymentMethodTypes: [STPPaymentMethodType] {
        // might need to add something to the public api or at least an internal property for this
        return [] // need to get this from sessions endpoint
    }
    
    var currency: String? {
        switch mode {
        case .payment(_, let currency):
            return currency
        case .setup(let currency):
            return currency
        }
    }
        
    
    var amount: Int? {
        switch mode {
        case .payment(let amount, _):
            return amount
        case .setup:
            return nil
        }
    }
    
    // renamed the internal abstraction name to avoid naming collisions
    var sfu: STPPaymentIntentSetupFutureUsage? {
        switch setupFutureUsage {
        case .offSession:
            return STPPaymentIntentSetupFutureUsage.offSession
        case .onSession:
            return STPPaymentIntentSetupFutureUsage.onSession
        }
    }
    
    var fulfilledRequirements: [PaymentMethodTypeRequirement] {
        // deferred intents cannot currently fufill requirements
        return []
    }
    
}
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/IntentRepresentable.swift
yuki-stripe: Some raw notes ahead of pairing:

I think in this specific case I prefer the explicitness over using amount and currency as a proxy for payment vs setup mode. e.g. it's possible that SI will contain amount and currency in the future (there is already a proposal to add currency to it).  

But I get your point in the general. I think we can kind of combine both worlds - add `deferredIntent` _and_ add vars that are in `IntentRepresentable` like `setupFutureUsage`, `recommendedPaymentSheetMethodTypes` etc. to `Intent` directly to avoid switching unless it's really necessary.  In the example above, perhaps we add a `var mode: Payment or Setup` to `Intent` to help with that.

PR #2262, opened 2023-02-03
Remove /elements/sessions (LUXE) filtering of payment methods

yuki-stripe:
## Summary
Disable server-side filtering by removing the "merchant_support_*" parameters in the `v1/elements/sessions` call.  

The endpoint doesn't filter unless "merchant_support_async" is passed in, see: https://git.corp.stripe.com/stripe-internal/pay-server/blob/95e8cc37c1a07a6a864d046d1937ff57d1a29d72/lib/lpm_ui_platform/command/filter/filter_mobile_lpms.rb#L55-L58 

## Motivation
We never fully transitioned to server-side filtering, and today filter on both the client and the server redundantly.  

We're going to encounter cases where we need to change the filtering logic. Since we don't plan to complete the transition, we'll revert server-side filtering and go back to having the client be solely responsible for filtering.

## Testing
Relying on existing tests.

## Changelog
No user facing change.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
@@ -204,7 +204,7 @@ extension PaymentSheet {
                 }
             }
 
-            let paymentTypes = recommendedPaymentMethodTypes.filter {
+            return recommendedPaymentMethodTypes.filter {
                 PaymentSheet.PaymentMethodType.supportsAdding(
                     paymentMethod: $0,
                     configuration: configuration,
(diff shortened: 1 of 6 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
```diff
@@ -493,22 +479,3 @@ extension STPPaymentMethodParams {
         }
     }
 }
-
-extension Array where Element == PaymentSheet.PaymentMethodType {
```
wooj-stripe, line 497: the original commit is here:
https://github.com/stripe/stripe-ios/commit/8e453258ec5750734432047e257b3615e90e238b
it looks like, we added `Hashable` to `PaymentSheet. PaymentMethodType` -- i wonder if we can get rid of `Hashable` ?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
```diff
@@ -330,13 +330,7 @@ extension PaymentSheet {
...
                 intentPromise.resolve(with: .paymentIntent(paymentIntent))
             }
-            let additionalParameters = [
-                "merchant_support_async": configuration.allowsDelayedPaymentMethods,
-                "merchant_support_shipping": configuration.allowsPaymentMethodsRequiringShippingAddress,
-            ]
-            configuration.apiClient.retrievePaymentIntentWithPreferences(
-                withClientSecret: clientSecret,
-                additionalParameters: additionalParameters) { result in
+            configuration.apiClient.retrievePaymentIntentWithPreferences(withClientSecret: clientSecret) { result in
```
wooj-stripe, line 333: in `retrievePaymentIntentWithPreferences`, we added `additionalParameters` as a parameter, can we get rid of that as well?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
yuki-stripe: This is still needed for `static var analyticLogForIcon: Set<PaymentMethodType> = []` - I won't try removing that in this PR.

wooj-stripe (approved):
lgtm!

PR #2263, opened 2023-02-03
Restore customerId comment to original location

bmelts-stripe:
## Summary
The new bindings accidentally misplaced this comment.


porter-stripe (approved):
good catch!

PR #2266, opened 2023-02-04
[Link] Migrate ConsumerPaymentDetails to modern bindings

bmelts-stripe:
## Summary
Migrate `ConsumerPaymentDetails` (and its constituent types) off of `STPAPIResponseDecodable` and `StripePayments.APIRequest`, and onto modern `Decodable` API bindings.

## Motivation
In addition to being the way of the future, modern bindings have the advantage of not having any dependency on StripePayments, which allows us to migrate `ConsumerPaymentDetails` into a shared framework at a future date.

## Testing
Automated tests to ensure JSON parsing behaves as expected, manual exercise to ensure user-facing behavior is unchanged.

## Notes
I'm really happy with the modern bindings in my limited usage so far - much less code, much cleaner code, no Objective-C compatibility required. I hope we can migrate more of the SDK onto them!

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/PaymentDetails.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/PaymentDetails.swift
@@ -18,98 +18,62 @@ typealias ConsumerSessionWithPaymentDetails = (session: ConsumerSession, payment
  
  For internal SDK use only
  */
-@objc(STP_Internal_ConsumerPaymentDetails)
-class ConsumerPaymentDetails: NSObject, STPAPIResponseDecodable {
-
+final class ConsumerPaymentDetails: Decodable {
     let stripeID: String
     let details: Details
     var isDefault: Bool
 
     // TODO(csabol) : Billing address
 
-    let allResponseFields: [AnyHashable: Any]
-
     init(stripeID: String,
          details: Details,
-         isDefault: Bool,
-         allResponseFields: [AnyHashable: Any]) {
+         isDefault: Bool) {
         self.stripeID = stripeID
         self.details = details
         self.isDefault = isDefault
-        self.allResponseFields = allResponseFields
-        super.init()
     }
 
-    static func decodedObject(fromAPIResponse response: [AnyHashable: Any]?) -> Self? {
-        guard let dict = response?["redacted_payment_details"] as? [AnyHashable: Any] ?? response, // When this is from a list endpoint it isn't nested in redacted_payment_details
-              let stripeID = dict["id"] as? String,
-              let details = Details(dict),
-              let isDefault = dict["is_default"] as? Bool else {
-            return nil
-        }
-
-        return ConsumerPaymentDetails(stripeID: stripeID,
-                                      details: details,
-                                      isDefault: isDefault,
-                                      allResponseFields: dict) as? Self
+    private enum CodingKeys: String, CodingKey {
+        case stripeID = "id"
+        case isDefault
     }
 
-    class ShareResponse: NSObject, STPAPIResponseDecodable {
-        let paymentMethodID: String
-        let allResponseFields: [AnyHashable: Any]
-
-        init(paymentMethodID: String,
-             allResponseFields: [AnyHashable: Any]) {
-            self.paymentMethodID = paymentMethodID
-            self.allResponseFields = allResponseFields
-            super.init()
-        }
-
-        static func decodedObject(fromAPIResponse response: [AnyHashable: Any]?) -> Self? {
-            guard let dict = response,
-                  let id = dict["payment_method"] as? String else {
-                return nil
-            }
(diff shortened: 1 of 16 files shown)


StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/PaymentDetails.swift
```diff
@@ -18,98 +18,62 @@ typealias ConsumerSessionWithPaymentDetails = (session: ConsumerSession, payment
...
-            default:
-                return nil
+        private enum CodingKeys: String, CodingKey {
+            case type
+            case card = "cardDetails"
+            case bankAccount = "bankAccountDetails"
+        }
+
+        // Our JSON structure doesn't align with Swift's expected structure for enums with associated values, so we do custom decoding.
+        init(from decoder: Decoder) throws {
```
davidme-stripe, line 69: I've seen this pattern when decoding a few other `Oneof`-shaped objects in our API. I wonder if it'd be worth making this more generic...

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/PaymentDetails.swift
```diff
@@ -128,9 +92,8 @@ extension ConsumerPaymentDetails {
 
 extension ConsumerPaymentDetails.Details {
     /// For internal SDK use only
-    @objc(STP_Internal_ConsumerPaymentDetails_CardChecks)
-    class CardChecks: NSObject, STPAPIResponseDecodable {
-        enum State: String {
+    final class CardChecks: Codable {
+        enum CVCCheck: String, Codable {
```
davidme-stripe, line 96: I might recommend using `SafeEnumCodable` for some of these. That way, a new enum case will fall through to the `.unparsable` case instead of causing the entire struct to fail decoding.

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/PaymentDetails.swift
```diff
@@ -140,82 +103,45 @@ extension ConsumerPaymentDetails.Details {
...
+    final class Card: Codable {
         let expiryYear: Int
         let expiryMonth: Int
-        let brand: STPCardBrand
+        let brand: String
         let last4: String
         let checks: CardChecks?
 
-        let allResponseFields: [AnyHashable: Any]
+        private enum CodingKeys: String, CodingKey {
```
davidme-stripe, line 123: Are these coming from the service as `expYear` or `exp_year`? If you use the camelCase version of the server representation (`exp_year` or `expYear`), you don't have to implement CodingKeys at all.

davidme-stripe (dismissed):
Thanks, glad the new bindings worked well for you!

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/PaymentDetails.swift
bmelts-stripe: `exp_year`, though a bit selfishly I prefer the unabbreviated version, hence the mapping 😅 Probably not worth it just to have a slightly nicer Swift API, though!

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/PaymentDetails.swift
bmelts-stripe: TIL! That's an excellent call, I'll use that here. Does that require making the parent objects `UnknownFieldsCodable`?

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/PaymentDetails.swift
bmelts-stripe: Seems useful! I'll file a ticket.

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/PaymentDetails.swift
davidme-stripe: It shouldn't be needed: `UnknownFieldsCodable` is an optional way of accessing extra fields from the raw JSON at runtime. It's somewhat documented in StripeCodable.swift, but I'll write up some more complete docs with examples.

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/PaymentDetails.swift
davidme-stripe: Meh, up to you! Explicitly defining the `CodingKeys` can make the code a little easier to understand at the expense of a little more boilerplate.

PR #2269, opened 2023-02-06
[Identity] Update example app for ID and address

ccen-stripe:
## Summary

* Update the example app for ID and Address type
* Add an option for returning a link instead of starting the verification

## Motivation

Support ID and Address

## Testing

- [x] Manually verified

| Before  | After |
| before | after |

### DIFF
--- a/Example/IdentityVerification Example/IdentityVerification Example/PlaygroundViewController.swift
+++ b/Example/IdentityVerification Example/IdentityVerification Example/PlaygroundViewController.swift
@@ -15,23 +15,30 @@ class PlaygroundViewController: UIViewController {
     let verifyEndpoint = "/create-verification-session"
 
     // Outlets
+    @IBOutlet weak var nativeOrWebSelector: UISegmentedControl!
     @IBOutlet weak var verificationTypeSelector: UISegmentedControl!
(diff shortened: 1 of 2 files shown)


Example/IdentityVerification Example/IdentityVerification Example/PlaygroundViewController.swift
```diff
@@ -244,11 +259,20 @@ class PlaygroundViewController: UIViewController {
...
-    @IBAction func didChangeUseNativeComponentsToggle(_ sender: Any) {
-        nativeComponentsOptionsContainerView.isHidden = !useNativeComponentsSwitch.isOn
+    @IBAction func didChangeNativeOrWeb(_ sender: Any) {
+        switch invocationType {
+        case .native:
+            nativeComponentsOptionsContainerView.isHidden = false
+        case .web:
+            nativeComponentsOptionsContainerView.isHidden = true
+        case .link:
+            nativeComponentsOptionsContainerView.isHidden = true
```
sgrant-stripe, line 274: Nit: Collapse the two cases here and in the switch above?

sgrant-stripe (approved):

PR #2273, opened 2023-02-07
FC Networking: created V0 of NetworkingLinkVerification pane

kgaidis-stripe:
## Summary

^ V0 of NetworkingLinkVerification pane. All the TODO's are expected. Lots more polish to later!

[Designs](https://www.figma.com/file/A4T5gNxmyRYmY02Y8YKgA2/Connections---Networking?node-id=8002%3A125413&t=3952VU5ofb1Mpvsc-0)

## Testing

Simulator Screen Shot - iPhone 14 Pro - 2023-02-10 at 13 01 32

https://user-images.githubusercontent.com/105514761/218164277-4055f973-326a-4344-98a0-64eff4905752.mov

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
@@ -98,6 +98,14 @@ protocol FinancialConnectionsAPIClient {
     func consumerSessionLookup(
         emailAddress: String
     ) -> Future<LookupConsumerSessionResponse>
+
+    func consumerSessionStartVerification(
+        emailAddress: String,
+        otpType: String,
+        customEmailType: String?,
+        connectionsMerchantName: String?,
+        consumerSessionClientSecret: String
+    ) -> Future<ConsumerSessionResponse>
 }
 
 extension STPAPIClient: FinancialConnectionsAPIClient {
@@ -470,6 +478,26 @@ extension STPAPIClient: FinancialConnectionsAPIClient {
                 .lowercased(),
         ]
         return post(resource: "consumers/sessions/lookup", parameters: parameters)
+    }
+
+    func consumerSessionStartVerification(
+        emailAddress: String,
+        otpType: String, // TODO(kgaidis): consider whether this should be an enum type SMS + EMAIL
+        customEmailType: String?, // TODO(kgaidis): consider whether this should be an enum type
+        connectionsMerchantName: String?,
+        consumerSessionClientSecret: String
+    ) -> Future<ConsumerSessionResponse> {
+        var parameters: [String: Any] = [
+            "request_surface": "web_connections",  // TODO(kgaidis): request backend to add ios_connections
+            "type": otpType,
+            "credentials": [
+                "consumer_session_client_secret": consumerSessionClientSecret,
+            ],
+            "locale": Locale.current.identifier,
+        ]
+        parameters["custom_email_type"] = customEmailType
+        parameters["connections_merchant_name"] = connectionsMerchantName
+        return post(resource: "consumers/sessions/start_verification", parameters: parameters)
     }
 }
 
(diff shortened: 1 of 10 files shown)


StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
```diff
@@ -471,6 +514,42 @@ extension STPAPIClient: FinancialConnectionsAPIClient {
         ]
         return post(resource: "consumers/sessions/lookup", parameters: parameters)
     }
+
+    func consumerSessionStartVerification(
+        emailAddress: String,
+        otpType: String, // TODO(kgaidis): consider whether this should be an enum type SMS + EMAIL
+        customEmailType: String?, // TODO(kgaidis): consider whether this should be an enum type
```
vardges-stripe, line 521: yes to enum?

StripeFinancialConnections/StripeFinancialConnections/Source/Native/NativeFlowController.swift
```diff
@@ -727,8 +752,21 @@ private func CreatePaneViewController(
...
+                apiClient: dataManager.apiClient,
+                clientSecret: dataManager.clientSecret,
+                analyticsClient: dataManager.analyticsClient
+            )
+            let networkingLinkVerificationViewController = NetworkingLinkVerificationViewController(dataSource: networkingLinkVerificationDataSource)
+            networkingLinkVerificationViewController.delegate = nativeFlowController
+            viewController = networkingLinkVerificationViewController
+        } else {
+            assertionFailure("Code logic error. Missing parameters for \(pane).")
+            viewController = nil
```
vardges-stripe, line 768: what should actually happen here? if exactly this (aka showing terminal error) maybe we log a better message?

StripeFinancialConnections/StripeFinancialConnections/Source/Native/NetworkingLinkVerification/NetworkingLinkVerificationViewController.swift
```diff
@@ -0,0 +1,192 @@
...
+        didEnterValidOTPCode otpCode: String
+    ) {
+        view.otpTextField.text = "CONFIRMING OTP..."
+
+        dataSource.confirmVerificationSession(otpCode: otpCode)
+            .observe { [weak self] result in
+                guard let self = self else { return }
+                switch result {
+                case .success:
+                    view.otpTextField.text = "SUCCESS! CALLING markLinkVerified..."
```
vardges-stripe, line 126: these are debug/interim code right?

vardges-stripe (approved):

StripeFinancialConnections/StripeFinancialConnections/Source/Native/NetworkingLinkVerification/NetworkingLinkVerificationViewController.swift
kgaidis-stripe: @vardges-stripe yes, it will be later replaced by a nicer OTP UI component

StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
kgaidis-stripe: this API should be replaced by what bmelts is building, so I probably can even delete this comment (as this whole API code should be deleted), but I kept it there just to be extra 'careful'

StripeFinancialConnections/StripeFinancialConnections/Source/Native/NativeFlowController.swift
kgaidis-stripe: Ideally, we never encounter this as this should never happen (hence the assert, but the assert can be debatable); we have been doing this since the inception of Auth Flow (there's precedence), so LMK on other ideas on how to handle. The thought is that we _think_ this will never happen, but in case it does, if we run in it DEBUG, we can catch it spectacularly failing rather than brushing it aside as "oh, just a random terminal error happened."

And yes, a normal user would see a terminal error:
- https://github.com/stripe/stripe-ios/blob/c4f05f52/StripeFinancialConnections/StripeFinancialConnections/Source/Native/NativeFlowController.swift#L137-L139

vardges-stripe: makes sense

PR #2277, opened 2023-02-09
Financial Connections: fixed an end-to-end test breaking

kgaidis-stripe:
## Summary

^

failure: https://stripe.slack.com/archives/C04G5PZM71T/p1675977162907139

the context is that server-driven prepane PR changed "alt" text for the "Continue" button's image and now the testing API thinks the button is called "Continue Continue"

Screen Shot 2023-02-09 at 4 57 24 PM

## Testing

```
Test Case '-[FinancialConnectionsUITests.FinancialConnectionsUITests testDataLiveModeOAuthWebAuthFlow]' passed (29.401 seconds).
```

### DIFF
--- a/Example/FinancialConnections Example/FinancialConnectionsUITests/FinancialConnectionsUITests.swift
+++ b/Example/FinancialConnections Example/FinancialConnectionsUITests/FinancialConnectionsUITests.swift
@@ -291,7 +291,10 @@ final class FinancialConnectionsUITests: XCTestCase {
         }
         institutionButton.tap()
 
-        let prepaneContinueButton = app.webViews.buttons["Continue"]
+        let prepaneContinueButton = app.webViews
+            .buttons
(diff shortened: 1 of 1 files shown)


Example/FinancialConnections Example/FinancialConnectionsUITests/FinancialConnectionsUITests.swift
```diff
@@ -291,7 +291,10 @@ final class FinancialConnectionsUITests: XCTestCase {
         }
         institutionButton.tap()
 
-        let prepaneContinueButton = app.webViews.buttons["Continue"]
+        let prepaneContinueButton = app.webViews
+            .buttons
+            .containing(NSPredicate(format: "label CONTAINS 'Continue'"))
```
kgaidis-stripe, line 296: This makes it more robust where instead of expecting it to always be "Continue" it is now flexible to having "Continue" anywhere in the button

vardges-stripe (approved):

PR #2290, opened 2023-02-14
FC Networking: V0 of NetworkingSaveToLinkVerification pane

kgaidis-stripe:
## Summary

^ see title

Note that this has a bunch of copied logic from `NetworkingLinkVerification` pane. That is per-design. The plan is:
1. Create the 3 verification panes that have OTP by copying code
2. Learn through the process
3. Then, refactor & combine (IF needed) after the lessons

Note that _all_ these panes _do_ have differences. It's a matter of figuring out what is the sweet-spot for reusability.

## Testing

Simulator Screen Shot - iPhone 14 Pro - 2023-02-15 at 14 56 21

https://user-images.githubusercontent.com/105514761/219139092-98f021ac-4a11-4ab2-99e8-fbc9f33386c4.mp4

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
@@ -80,9 +80,9 @@ protocol FinancialConnectionsAPIClient {
     // MARK: - Networking
 
     func saveAccountsToLink(
-        emailAddress: String,
-        phoneNumber: String,
-        country: String,
+        emailAddress: String?,
+        phoneNumber: String?,
+        country: String?,
         selectedAccountIds: [String],
         consumerSessionClientSecret: String?,
         clientSecret: String
@@ -453,24 +453,24 @@ extension STPAPIClient: FinancialConnectionsAPIClient {
     // MARK: - Networking
 
     func saveAccountsToLink(
-        emailAddress: String,
-        phoneNumber: String,
-        country: String,
+        emailAddress: String?,
+        phoneNumber: String?,
+        country: String?,
         selectedAccountIds: [String],
         consumerSessionClientSecret: String?,
         clientSecret: String
     ) -> Future<FinancialConnectionsSessionManifest> {
         var body: [String: Any] = [
             "client_secret": clientSecret,
-            "email_address":
-                emailAddress
-                .trimmingCharacters(in: .whitespacesAndNewlines)
-                .lowercased(),
-            "phone_number": phoneNumber,
-            "country": country,
-            "locale": Locale.current.identifier,
             "selected_accounts": selectedAccountIds,
+            "expand": ["active_auth_session"],
         ]
+        body["email_address"] = emailAddress?
+            .trimmingCharacters(in: .whitespacesAndNewlines)
+            .lowercased()
+        body["phone_number"] = phoneNumber
+        body["country"] = country
+        body["locale"] = (phoneNumber != nil) ? Locale.current.identifier : nil
         body["consumer_session_client_secret"] = consumerSessionClientSecret
         return post(resource: APIEndpointSaveAccountsToLink, parameters: body)
     }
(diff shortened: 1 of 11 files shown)


StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
```diff
@@ -453,24 +453,24 @@ extension STPAPIClient: FinancialConnectionsAPIClient {
...
-            "locale": Locale.current.identifier,
             "selected_accounts": selectedAccountIds,
+            "expand": ["active_auth_session"],
         ]
+        body["email_address"] = emailAddress?
+            .trimmingCharacters(in: .whitespacesAndNewlines)
+            .lowercased()
+        body["phone_number"] = phoneNumber
+        body["country"] = country
+        body["locale"] = (phoneNumber != nil) ? Locale.current.identifier : nil
```
vardges-stripe, line 473: any specific reason for not sending locale for all requests? Existing js logic?

StripeFinancialConnections/StripeFinancialConnections/Source/Native/NativeFlowController.swift
```diff
@@ -814,6 +843,27 @@ private func CreatePaneViewController(
             assertionFailure("Code logic error. Missing parameters for \(pane).")
             viewController = nil
         }
+    case .networkingSaveToLinkVerification:
+        if
+            let consumerSession = dataManager.consumerSession,
+            let selectedAccountIds = dataManager.linkedAccounts?.map({ $0.id })
```
vardges-stripe, line 849: should we also add a condition of selectedAccountIds not being empty?

StripeFinancialConnections/StripeFinancialConnections/Source/Native/NetworkingSaveToLinkVerification/NetworkingSaveToLinkBodyView.swift
```diff
@@ -0,0 +1,142 @@
...
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    @objc private func otpTextFieldDidChange() {
+        guard let otp = otpTextField.text else {
+            return
+        }
+
+        if otp.count == 6 && Int(otp) != nil {
```
vardges-stripe, line 68: should we check if all characters are digits? otherwise users can enter "-12345" and it would be a valid int. Doesn't really change that much since backend will fail but probably better to catch on the client?

vardges-stripe (approved):
lgtm. Does seem that some unification would be beneficial for not having same code in many places and would help with binary size.

StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
kgaidis-stripe: yes, just using Stripe.js precedent, this is the logic:

https://git.corp.stripe.com/stripe-internal/stripe-js-v3/blob/eccfa561/src/linkedAccounts/inner/components/panes/LinkLogin/useSaveToNetworkAndLink.ts#L172

StripeFinancialConnections/StripeFinancialConnections/Source/Native/NetworkingSaveToLinkVerification/NetworkingSaveToLinkBodyView.swift
kgaidis-stripe: yea this is just temporary, the OTP UI will change by introducing a reusable OTP component, which will maybe (or maybe not) provide us with new validation

StripeFinancialConnections/StripeFinancialConnections/Source/Native/NativeFlowController.swift
kgaidis-stripe: hmm great point! thank you

kgaidis-stripe: I will change it to actually just pass a single account; that's actually the case

I was debating this the whole time (should I pass the array or single account), and if I had just done the simple thing, this wouldn't have been an issue (I was being fancy)

vardges-stripe: simplicity wins again

PR #2291, opened 2023-02-15
[Link] Refactor ConsumerSession to support Codable and modern bindings

bmelts-stripe:
## Summary
Building off the work in #2266, this converts much of `ConsumerSession` off of `STPAPIResponseDecodable` and `StripePayments.APIRequest`, and onto modern `Decodable` API bindings.

## Motivation
In addition to being the way of the future, modern bindings have the advantage of not having any dependency on StripePayments, which allows us to migrate `ConsumerSession` into a shared framework at a future date.

## Testing
Automated tests to ensure JSON parsing behaves as expected, manual exercise to ensure user-facing behavior is unchanged.


StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/ConsumerSession.swift
```diff
@@ -12,81 +12,50 @@ import UIKit
...
 /// For internal SDK use only
-@objc(STP_Internal_ConsumerSession)
-class ConsumerSession: NSObject, STPAPIResponseDecodable {
-
+final class ConsumerSession: Decodable {
     let clientSecret: String
     let emailAddress: String
     let redactedPhoneNumber: String
     let verificationSessions: [VerificationSession]
-    let authSessionClientSecret: String?
```
bmelts-stripe, line 22: Hoisting this out into a separate response type (see [`SessionResponse` in STPAPIClient+Link.swift](https://github.com/stripe/stripe-ios/pull/2291/files#diff-37b1d3022a0265c5ae1a48d7f4f2027426df1271f4c2b03bacb2ce6bc5b7e3b4)) was informed by a couple of factors, but one was the lack of support in `StripeJSONDecoder` for nested containers - using a `nestedKeyedContainer` was the most sensible way to implement `init(from:)` here while still parsing `authSessionClientSecret` correctly, but that wasn't viable. A previous implementation for this PR actually added support for them to enable this, but looking over the [original PR](https://github.com/stripe-ios/stripe-ios-private/pull/979), it seems like some of these limitations may have actually been intentional, so I took another approach. I'm happier with this one overall—smaller surface area, better matching of object shapes to API shapes—but @davidme-stripe would love to get some insight around this!

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/ConsumerSession.swift
```diff
@@ -12,81 +12,50 @@ import UIKit
...
 
-// MARK: - Cookie Management
-
-extension ConsumerSession {
-    func updateCookie(withStore store: LinkCookieStore) {
-        store.updateSessionCookie(with: authSessionClientSecret)
+extension ConsumerSession: Equatable {
+    static func ==(lhs: ConsumerSession, rhs: ConsumerSession) -> Bool {
+        // NSObject-style equality
+        return ObjectIdentifier(lhs) == ObjectIdentifier(rhs)
```
davidme-stripe, line 58: Do we really want to compare the reference identifiers here? I'd expect this comparison to be based on the underlying data (or maybe the `allResponseFields`).

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/ConsumerSession.swift
davidme-stripe: This approach seems good to me! Re: Nested container support, I probably shouldn't have said *intentionally* in the original PR — I was just noting that it wasn't an oversight. We didn't have any plans to use it, so I skipped it in the interest of time.

davidme-stripe (approved):
Sorry for the delay here! I've gotta fix my GitHub notifications...

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/ConsumerSession.swift
bmelts-stripe: Gotcha! Happy to tackle adding it back in at some future date, but feels a touch too scope-creepy for this PR.

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/ConsumerSession.swift
bmelts-stripe: I think that would be the superior approach, honestly - this was to preserve the behavior from when this type was an NSObject and we got identifier-based equality by default. Auditing all places we compare two session objects to ensure that content-based equality wouldn't break things felt out of scope, but I'm happy to add a followup ticket for it!

PR #2298, opened 2023-02-16
Add STPElementsSession to Intent.deferredIntent

yuki-stripe:
## Summary
Adds STPElementsSession to Intent.deferredIntent and replaces some fatalErrors with real implementations based on the session. 

Also moves STPElementSession.swift to the PaymentSheet module

## Motivation
Decoupling

## Testing
No tests yet, we'll test this whole codepath later.

## Changelog
None


StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/STPElementsSession.swift
```diff
@@ -6,6 +6,7 @@
 //
 
 import Foundation
+@_spi(STP) import StripePayments
```
porter-stripe, line 9: Why is this needed? I don't see any changes to the file.

porter-stripe: Oh some file was moved?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Intent.swift
```diff
@@ -21,7 +21,7 @@ import UIKit
 enum Intent {
     case paymentIntent(STPPaymentIntent)
     case setupIntent(STPSetupIntent)
-    case deferredIntent
+    case deferredIntent(STPElementsSession)
 
     var clientSecret: String {
```
porter-stripe, line 26: Guessing this will become a `String?` eventually.

porter-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/STPElementsSession.swift
yuki-stripe: Yeah I think STPElementsSession should be in the StripePaymentSheet module.

porter-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Intent.swift
yuki-stripe: I'm guessing we'll delete this and make its users explicitly handle all cases
```
switch intent {
case paymentIntent(let pi):
  pi.clientSecret...
case setupIntent(let si):
  si.clientSecret...
case deferredIntent:
  ...ask merchant to create the intent client secret...
```

porter-stripe (approved):

PR #2299, opened 2023-02-16
Deferred public API + wire it up

porter-stripe:
## Summary
- Added our current definition for the `IntentConfiguration`
- Adds new inits to PaymentSheet and FlowController for the deferred workflow
- Uses the values from the public API to make the request to elements/sessions
- Renames a few things

## Motivation
Decoupled flow

## Testing
- Existing tests and some new

## Changelog
N/A


Stripe/StripeiOSTests/STPIntentWithPreferencesTest.swift
```diff
@@ -93,19 +93,64 @@ class STPIntentWithPreferencesTest: XCTestCase {
         wait(for: [expectation], timeout: STPTestingNetworkRequestTimeout)
     }
 
-    func testRetrieveElementSession() {
+    func testRetrieveElementSession_deferredPayment() {
```
porter-stripe, line 96: Added a test for both payment and setup mode for better coverage.

Stripe/StripeiOSTests/STPIntentWithPreferencesTest.swift
```diff
@@ -93,19 +93,64 @@ class STPIntentWithPreferencesTest: XCTestCase {
...
-                XCTAssertNotNil(deferredIntent.linkSettings)
-                XCTAssertNotNil(deferredIntent.paymentMethodSpecs)
-                XCTAssertFalse(deferredIntent.orderedPaymentMethodTypes.isEmpty)
-                XCTAssertFalse(deferredIntent.paymentMethodTypes.isEmpty)
+            case .success(let elementsSession):
+                XCTAssertNotNil(elementsSession)
+                XCTAssertEqual(elementsSession.countryCode, "US")
+                XCTAssertNotNil(elementsSession.linkSettings)
+                XCTAssertNotNil(elementsSession.paymentMethodSpecs)
+                XCTAssertEqual(
```
porter-stripe, line 113: Since we can specify the payment methods now I think it makes sense to test these specifically.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Intent.swift
```diff
@@ -92,15 +92,18 @@ enum Intent {
     }
 }
 
-// MARK: - IntentClientSecret
+// MARK: - RetrievableIntent
 
-/// An internal type representing a PaymentIntent or SetupIntent client secret
-enum IntentClientSecret {
+/// An internal type representing types of intents that can be retrieved from Stripe
+enum RetrievableIntent {
```
porter-stripe, line 98: Open to name suggestions here, but `IntentClientSecret` doesn't make sense any more.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Intent.swift
```diff
@@ -92,15 +92,18 @@ enum Intent {
...
     /// The [client secret](https://stripe.com/docs/api/payment_intents/object#payment_intent_object-client_secret) of a Stripe PaymentIntent object
     case paymentIntent(clientSecret: String)
 
     /// The [client secret](https://stripe.com/docs/api/setup_intents/object#setup_intent_object-client_secret) of a Stripe SetupIntent object
     case setupIntent(clientSecret: String)
+
+    /// The `IntentConfiguration` of a Stripe deferred intent
+    case defferedIntent(intentConfig: PaymentSheet.IntentConfiguration)
```
yuki-stripe, line 106: Typo: deferred

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Intent.swift
yuki-stripe: Can we delete this enum and replace it with `InitializationMode`?

Stripe/StripeiOSTests/STPIntentWithPreferencesTest.swift
yuki-stripe: I'm not 100% sure whether or not the endpoint can change the provided payment method types fyi but this seems fine.

Stripe/StripeiOSTests/PaymentSheet+APITest.swift
```diff
@@ -287,7 +287,7 @@ class PaymentSheetAPITest: XCTestCase {
             }
 
             PaymentSheet.load(
-                clientSecret: IntentClientSecret.setupIntent(clientSecret: clientSecret),
+                retrievableIntent: RetrievableIntent.setupIntent(clientSecret: clientSecret),
```
yuki-stripe, line 290: Can we add a test to load w/ a deferred Intent here?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Intent.swift
porter-stripe: Ah yea good catch! Now with this we don't need to add new init's but rather make our existing internal ones public (under spi for now)

Stripe/StripeiOSTests/PaymentSheet+APITest.swift
porter-stripe: Added!

yuki-stripe (approved):

PR #2301, opened 2023-02-17
Handle fatal errors for deferred intent

porter-stripe:
## Summary
- Handles all the fatal errors for `deferredIntent` excluding those at confirm and ACH time.

## Motivation
Decouple workflow

## Testing
- existing tests

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Intent.swift
```diff
@@ -21,7 +21,7 @@ import UIKit
 enum Intent {
     case paymentIntent(STPPaymentIntent)
     case setupIntent(STPSetupIntent)
-    case deferredIntent(STPElementsSession)
+    case deferredIntent(elementsSession: STPElementsSession, intentConfig: PaymentSheet.IntentConfiguration)
```
porter-stripe, line 24: Updated this to also contain an intent config

yuki-stripe: Nice, yep.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Intent.swift
```diff
@@ -86,8 +96,13 @@ enum Intent {
             return paymentIntent.setupFutureUsage != .none
         case .setupIntent:
             return true
-        case .deferredIntent:
-            fatalError("TODO(DeferredIntent): Needs to switch on the init object mode")
+        case .deferredIntent(_, let intentConfig):
+            switch intentConfig.mode {
+            case .payment(_, _, let setupFutureUsage):
+                return setupFutureUsage != .none
```
yuki-stripe, line 102: Supernit: this made me think setupFutureUsage had a `.none` case. Can we instead do `setupFutureUsage != nil`

yuki-stripe (approved):

yuki-stripe (approved):

PR #2311, opened 2023-02-21
Retry fetching cash app intent status after redirect

porter-stripe:
## Summary
-  A change to the redirect flow made the intent status not update fast enough once the SFSafariViewController was dismissed. I made a change so Cash App intent status is retried a few times similar to 3DS2 which fixed the issue.
- We also no longer need the follow redirects flow as the redirect flow has been improved.
- Bug fix in payment sheet where the spinner would not restart spinning after coming back to the foreground.

New flow:

https://stripe.slack.com/archives/C038SFN5Q3Z/p1677018621376119?thread_ts=1677012194.769999&cid=C038SFN5Q3Z

## Motivation
Cash app beta

## Testing
Manual live mode

## Changelog
N/A


sfriedman-stripe (approved):
LGTM

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
```diff
@@ -1311,6 +1311,14 @@ public class STPPaymentHandler: NSObject {
             defer {
                 blockingDataTaskSemaphore.signal()
             }
+
+            // Check if the request failed due to attempting to redirect to an app universal link for Cash App
```
porter-stripe, line 1315: @wooj-stripe this is a bit hacky.. when we do follow redirects we get an error back saying the cash app universal link is invalid, so we extract it out of the `error`. We never make a direct request to the universal link via a URL session so this seems like the best way to read it out. thoughts?

The link looks like: `cashme://cash.app/f/...`

wooj-stripe: This is odd, this means that the URL that is vended to us in `authenticationAction.cashAppRedirectToApp?.mobileAuthURL` resulting in an error if we attempt to fetch it?  Why are then vending us a URL which will result in an error?

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
```diff
@@ -1284,7 +1284,7 @@ public class STPPaymentHandler: NSObject {
 
             if let mobileAuthURL = authenticationAction.cashAppRedirectToApp?.mobileAuthURL {
                 let resultingRedirectURL = self.followRedirects(to: mobileAuthURL, urlSession: URLSession.shared)
-                _handleRedirect(to: resultingRedirectURL, withReturn: returnURL)
+                _handleRedirectToExternalBrowser(to: resultingRedirectURL, withReturn: returnURL)
 
             } else {
```
wooj-stripe, line 1289: Do we really want to use `_handleRedirectToExternalBrowser(to: resultingRedirectURL, withReturn: returnURL)` ?

If the app is not installed or safari isn't set as their default browser, then the user will end up in web browser outside of the merchant app.

porter-stripe: I actually have Firefox set as my default in the video above, since we are opening a universal link the external browser is just skipped and it goes straight to Cash App! If the app isn't installed it goes to the app store. The behavior between SFSafariViewController and the external flow is actually the same, but the callbacks for the external browser are a little more favorable for a fast redirect PM from my testing.

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
porter-stripe: I think this is because URLSession automatically follows redirects and eventually errors when it gets to the universal link. The URL we get back is `https://pm-redirects.stripe.com/authorize...` which redirects to the cash app universal link.

wooj-stripe: `followRedirects` should make a call to the value of  `authenticationAction.cashAppRedirectToApp?.mobileAuthURL`, which should be: `https://pm-redirects.stripe.com/authorize...`, and because that link is expected to be a redirect, the responseURL should be the cash app URL, however, it sounds doing an HTTP GET on `https://pm-redirects.stripe.com/authorize...` is resulting in an error.

1.  Is it erroring because cash app isn't using universal links? (e.g. their link is `cashme://` which is not a universal link.  `cashme://` is a "custom URL Scheme").  Is there any additional info on the error we get back?
  
2.  I thought cash app was using universal links (https://cash.app), which is far superior with regards to security (and provides a much better fall back experience in the event the app isn't installed).  Is there any reason they moved away from universal links?

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
```diff
@@ -1311,6 +1311,14 @@ public class STPPaymentHandler: NSObject {
             defer {
                 blockingDataTaskSemaphore.signal()
             }
+
+            // Check if the request failed due to attempting to redirect to an app universal link for Cash App
+            if let nsError = error as? NSError, let errorUrl = nsError.userInfo[NSURLErrorFailingURLStringErrorKey] as? String {
+                if errorUrl.contains("cashme://cash.app"), let universalLink = URL(string: errorUrl) {
+                    resultingUrl = universalLink
```
wooj-stripe, line 1328: Assuming we are okay with introducing a regression for users who don't have the app installed...

1. We should rename `universalLink` to something like `customSchemeAppURL`
2. Not ideal that we're having to hard-code this in here, but it does seem safer than checking for wildcards.

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
wooj-stripe, line 1325: Change:
```
// Check if the request failed due to attempting to redirect to an app universal link for Cash App
```

To reference "Custom Scheme URL" instead of Universal link... maybe something like:
```
// Check if the request failed due to attempting to redirect to a custom Scheme URL for Cash App
```

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
wooj-stripe: Ah, that's interesting -- probably because it is not a universal link... in this case, I still don't think we should be using redirect to external browser, we should probably do:
```
            if let mobileAuthURL = authenticationAction.cashAppRedirectToApp?.mobileAuthURL {
                let resultingRedirectURL = self.followRedirects(to: mobileAuthURL, urlSession: URLSession.shared)
                _handleRedirect(to: resultingRedirectURL, fallbackURL: mobileAuthURL, return: returnURL)
             }

```
This path seems to use `UIApplication.shared.open(` and assume the link is a customURL.  If it succeeds, then great, the app should open.  If this fails, because the app is not installed, open up the original value of `authenticationAction.cashAppRedirectToApp?.mobileAuthURL` in an in-app web modal.

porter-stripe: Swapped, also just saw the `_handleRedirect` function throws the `requiredAppNotAvailable` error if it can't open the link :)

wooj-stripe (approved):
Approving assuming that cash app is not going to adopt universal links in their app.

wooj-stripe (approved):

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
porter-stripe: @sfriedman-stripe I asked yesterday but haven't heard back. We only became aware of this within the last 24 hours when we started sending the IN_APP channel. Before that we didn't have insight into the links they were using.

PR #2312, opened 2023-02-22
Deferred intent snapshot tests

porter-stripe:
## Summary
- Adds snapshot tests for deferred intents. This allows us test that payment sheet can load via the deferred intent code path. We stub the sessions endpoint response and don't specify payment method types in our tests since they are hardcoded in the stubbed response.
- Fixes a bug where we were not reading ordered payment methods correctly causing PaymentSheet loading to fail
- Moves `InitializationMode` under the `PaymentSheet` namespace
- In a future PR when we can confirm a deferred intent we will add automated UI tests

## Motivation
Deferred intents

## Testing
- New tests

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetIntentConfiguration.swift
```diff
@@ -10,6 +10,18 @@ import Foundation
 @_spi(STP) public extension PaymentSheet {
     /// 🚧 Under construction
     struct IntentConfiguration {
+
+        /// Creates a `PaymentSheet.IntentConfiguration` with the given values
+        /// - Parameters:
+        ///   - mode: The mode of this intent, either payment or setup
+        ///   - captureMethod: The capture method of this intent, either automatic or manual
+        ///   - paymentMethodTypes: The payment method types for the intent
+        public init(mode: Mode, captureMethod: CaptureMethod, paymentMethodTypes: [String]? = nil) {
```
yuki-stripe, line 19: Let's give captureMethod a default value of `.automatic`, it should be very rarely used

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetSnapshotTests.swift
```diff
@@ -483,6 +483,56 @@ class PaymentSheetSnapshotTests: FBSnapshotTestCase {
...
+        let intentConfig = PaymentSheet.IntentConfiguration(mode: .payment(amount: 1000, currency: "USD", setupFutureUsage: .onSession),
+                                                            captureMethod: .automatic)
+
+        preparePaymentSheet(intentConfig: intentConfig)
+        presentPaymentSheet(darkMode: true)
+        verify(paymentSheet.bottomSheetViewController.view!)
+    }
+
+    func testPaymentSheetAppearance_deferredIntent() {
+        stubNewCustomerResponse()
```
yuki-stripe, line 511: ooc why test darkmode or appearance + deferred Intent?  They feel unrelated

porter-stripe: yeah probably not needed, added for good measure but removed them

yuki-stripe (approved):

Issue #2302, opened 2023-02-20
[BUG]  paymentOptionsViewControllerFooterView is not working


alwayswith:
## Summary

When I use the method of presentPaymentOptionsViewController(withNewState:) or pushPaymentOptionsViewController() in STPPaymentContext to present/push the build in payment options UI, I found that the paymentOptionsViewControllerFooterView property is not working.

## Code to reproduce

`STPPaymentContext.swift`

    func presentPaymentOptionsViewController(withNewState state: STPPaymentContextState) {
        assert(
            hostViewController != nil && hostViewController?.viewIfLoaded?.window != nil,
            "hostViewController must not be nil on STPPaymentContext when calling pushPaymentOptionsViewController on it. Next time, set the hostViewController property first!"
        )
        if self.state == STPPaymentContextState.none {
            self.state = state
            let paymentOptionsViewController = STPPaymentOptionsViewController(paymentContext: self) //  onSuccess is invoked, paymentOptionsViewControllerFooterView is nil
            self.paymentOptionsViewController = paymentOptionsViewController
            paymentOptionsViewController.prefilledInformation = prefilledInformation
            paymentOptionsViewController.defaultPaymentMethod = defaultPaymentMethod
            paymentOptionsViewController.paymentOptionsViewControllerFooterView =
                paymentOptionsViewControllerFooterView  // assign property after onSuccess
            paymentOptionsViewController.addCardViewControllerFooterView =
                addCardViewControllerFooterView
            paymentOptionsViewController.navigationItem.largeTitleDisplayMode =
                largeTitleDisplayMode

            let navigationController = UINavigationController(
                rootViewController: paymentOptionsViewController)
            navigationController.navigationBar.stp_theme = theme
            navigationController.navigationBar.prefersLargeTitles = true
            navigationController.modalPresentationStyle = modalPresentationStyle
            hostViewController?.present(
                navigationController,
                animated: transitionAnimationsEnabled())
        }
    }

`STPPaymentOptionsViewController.swift`

// strongSelf.paymentOptionsViewControllerFooterView is  always nil, because you assign  the property after the onSuccess callback invoked
                if strongSelf.paymentOptionsViewControllerFooterView != nil { 
                    payMethodsInternal?.customFooterView =
                        strongSelf.paymentOptionsViewControllerFooterView
                }

## iOS version
14.5

## Installation method
CocoaPods 

## SDK version
22.8.4

wooj-stripe:
[...]

PR #2314, opened 2023-02-22
Fix race condition between fetching payment options and setting properties after init

wooj-stripe:
## Summary
Fix race condition in footer view for basic integration payment methods

## Motivation
Issue reported: #2302

## Testing
Manual testing & Added unit tests for xcode 14 only.

This gif demonstrates that the footer within `STPPaymentOptionsInternalViewController` now shows up when a user taps on the payment method, before the response has fully loaded:
2314_footerIsShown

This gif demonstrates the `STPAddCardViewController` class being displayed instead of `STPPaymentOptionsInternalViewController`, because we have configured it to have no payment methods.  It demonstrates that the prefilledInformation is being properly passed down to these subview controllers and populating the UI with that information.
2314_addCard

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## x.x.x yyyy-mm-dd
+### Basic Integration
+* [Fixed] Race condition reported in #2302
+
 ## 23.4.0 2023-02-21
 ### PaymentSheet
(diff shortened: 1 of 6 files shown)


davidme-stripe (approved):

alwayswith:
Great! Thank you!

PR #2316, opened 2023-02-22
Add logging when a payment method is filtered out in PaymentSheet

porter-stripe:
## Summary
Developers are often confused as to why payment methods are not showing up. This PR adds logging detailing why a PaymentMethod is not being rendered in PaymentSheet.

Example log:
```
[Stripe SDK]: US Bank Account is not being displayed because one or more requirements are not being met or this payment method is not supported by PaymentSheet. userSupportsDelayedPaymentMethods: PaymentSheet.Configuration.allowsDelayedPaymentMethods must be set to true. 

[Stripe SDK]: Afterpay is not being displayed because one or more requirements are not being met or this payment method is not supported by PaymentSheet. shippingAddress: A shipping address must be present on the Intent or collected through the Address Element and populated on PaymentSheet.Configuration.shippingDetails. See https://stripe.com/docs/api/payment_intents/object#payment_intent_object-shipping and https://stripe.com/docs/elements/address-element/collect-addresses?platform=ios#ios-pre-fill-billing 

[Stripe SDK]: Affirm is not being displayed because one or more requirements are not being met or this payment method is not supported by PaymentSheet. shippingAddress: A shipping address must be present on the Intent or collected through the Address Element and populated on PaymentSheet.Configuration.shippingDetails. See https://stripe.com/docs/api/payment_intents/object#payment_intent_object-shipping and https://stripe.com/docs/elements/address-element/collect-addresses?platform=ios#ios-pre-fill-billing 

[Stripe SDK]: Alipay is not being displayed because one or more requirements are not being met or this payment method is not supported by PaymentSheet. Not currently supported by PaymentSheet.

[Stripe SDK]: WeChat Pay is not being displayed because one or more requirements are not being met or this payment method is not supported by PaymentSheet. Not currently supported by PaymentSheet.
```

Cons: 
- Noisier logs from our SDK.

## Motivation
Developer happiness

## Testing
Manual

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+PaymentMethodAvailability.swift
```diff
@@ -132,5 +132,63 @@ extension PaymentSheet {
...
+            case .financialConnectionsSDK:
+                return "financialConnectionsSDK: The FinancialConnections must be linked."
+            case .validUSBankVerificationMethod:
+                return "validUSBankVerificationMethod: Requires a valid us bank verification method."
+            }
+        }
+
+    }
+
+    enum PaymentMethodExclusionReason: Equatable {
```
eurias-stripe, line 155: Can you change the name? "ExclusionReason" doesn't make sense if one of the options is "supported", something like PaymentMethodValidationResult would make more sense

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+PaymentMethodAvailability.swift
```diff
@@ -132,5 +132,63 @@ extension PaymentSheet {
...
+            case .unavailable:
+                return "unavailable: This payment method is not available."
+            case .returnURL:
+                return "returnURL: A return URL must be set, see https://stripe.com/docs/payments/accept-a-payment?platform=ios&ui=payment-sheet#ios-set-up-return-url"
+            case .shippingAddress:
+                return "shippingAddress: A shipping address must be present on the Intent or collected through the Address Element and populated on PaymentSheet.Configuration.shippingDetails. See https://stripe.com/docs/api/payment_intents/object#payment_intent_object-shipping and https://stripe.com/docs/elements/address-element/collect-addresses?platform=ios#ios-pre-fill-billing"
+            case .userSupportsDelayedPaymentMethods:
+                return "userSupportsDelayedPaymentMethods: PaymentSheet.Configuration.allowsDelayedPaymentMethods must be set to true."
+            case .financialConnectionsSDK:
+                return "financialConnectionsSDK: The FinancialConnections must be linked."
```
eurias-stripe, line 147:
```suggestion
                return "financialConnectionsSDK: The FinancialConnections SDK must be linked."
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+PaymentMethodAvailability.swift
```diff
@@ -132,5 +132,63 @@ extension PaymentSheet {
...
+            case .validUSBankVerificationMethod:
+                return "validUSBankVerificationMethod: Requires a valid us bank verification method."
```
eurias-stripe, line 149:
```suggestion
                return "validUSBankVerificationMethod: Requires a valid US bank verification method."
```

not sure if we need to capitalize "bank" as well

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
```diff
@@ -211,14 +211,23 @@ extension PaymentSheet {
...
                     intent: intent,
                     supportedPaymentMethods: configuration.linkPaymentMethodsOnly
                         ? PaymentSheet.supportedLinkPaymentMethods : PaymentSheet.supportedPaymentMethods
                 )
+
+                if exclusionReason != .supported {
+                    // This payment method is being flitered out, log the reason/s why
+                    #if DEBUG
+                    print("[Stripe SDK]: \(paymentMethodType.displayName) is not being displayed because one or more requirements are not being met or this payment method is not supported by PaymentSheet. \(exclusionReason.description)\n")
+                    #endif
```
eurias-stripe, line 227: Can you move this into a helper function so we don't pollute our codebase with `#if DEBUG` checks?, it would also make it easier to change in the future

eurias-stripe (changes requested):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+PaymentMethodAvailability.swift
porter-stripe: Yeah makes sense, renamed to `PaymentMethodAvailabilityStatus`.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
porter-stripe: Agree it would be nice to re-use in the future but it's only being used in one place for now an a function that is 
```swift
func debugPrint(str) {
   #if DEBUG
   print(str)
   #endif
 }
```
seems a bit overkill for one use, and there really isn't a place for this function to live that makes sense now. I think it's something we should tackle when we add a logging framework.

eurias-stripe: it could live in StripeCore -> Helpers, and available through @_spi. The function could add the "[Stripe SDK]" prefix as well for example.
I agree that since it is a single use is not strictly necessary, but I'd rather add it now while you are in there, but either way is okay.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
```diff
@@ -211,14 +211,23 @@ extension PaymentSheet {
                 }
             }
 
-            return recommendedPaymentMethodTypes.filter {
-                PaymentSheet.PaymentMethodType.supportsAdding(
-                    paymentMethod: $0,
+            return recommendedPaymentMethodTypes.filter { paymentMethodType in
+                let exclusionReason = PaymentSheet.PaymentMethodType.supportsAdding(
```
eurias-stripe, line 215: rename this

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
```diff
@@ -229,13 +238,13 @@ extension PaymentSheet {
         ///   - requirementProviders: a list of [PaymentMethodRequirementProvider] who satisfy payment requirements
         ///   - intent: a intent object
         ///   - supportedPaymentMethods: the payment methods that PaymentSheet can display UI for
-        /// - Returns: true if `paymentMethod` should be available in the PaymentSheet, false otherwise
+        /// - Returns: a `PaymentMethodExclusionReason` detailing why or why not this payment method can be added
```
eurias-stripe, line 241: rename

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
```diff
@@ -373,18 +382,23 @@ extension PaymentSheet {
         ///   - requirements: a list of requirements to be satisfied
         ///   - configuration: a configuration to satisfy requirements
         ///   - intent: an intent object
-        /// - Returns: true if the configuration satisfies the requirements
+        /// - Returns: a `PaymentMethodExclusionReason` detailing why or why not this payment method can be added
```
eurias-stripe, line 385: rename

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
```diff
@@ -396,21 +410,21 @@ extension PaymentSheet {
         ///   - configuration: a configuration to satisfy requirements
         ///   - intent: an intent object
         ///   - supportedPaymentMethods: a list of supported payment method types
-        /// - Returns: true if the configuration satisfies the requirements and if `paymentMethod` is contained in `supportedPaymentMethods`
+        /// - Returns: a `PaymentMethodExclusionReason` detailing why or why not this payment method is supported
```
eurias-stripe, line 413: rename

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
porter-stripe: I'm leaning towards not adding it because I don't think we want people using a helper logging function, we want people using a logging framework (that doesn't exist). Creating a new file in StripeCore e.g. `Logger.swift` just for this  function will indicate to folks that this is the blessed way to do logging in our SDK when it is not. If we find a need for a helper like this I think we should definitely add a logging framework. 

https://paper.dropbox.com/doc/2021-10-11-Shared-Logging-APIs-Mobile-API-Review-R2HlpenOCskNdl37HWHY8

eurias-stripe: that makes sense, let's skip this

eurias-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
```diff
@@ -211,14 +211,23 @@ extension PaymentSheet {
...
+                let availabilityStatus = PaymentSheet.PaymentMethodType.supportsAdding(
+                    paymentMethod: paymentMethodType,
                     configuration: configuration,
                     intent: intent,
                     supportedPaymentMethods: configuration.linkPaymentMethodsOnly
                         ? PaymentSheet.supportedLinkPaymentMethods : PaymentSheet.supportedPaymentMethods
                 )
+
+                if availabilityStatus != .supported {
+                    // This payment method is being flitered out, log the reason/s why
```
sfriedman-stripe, line 224:
```suggestion
                    // This payment method is being filtered out, log the reason/s why
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
```diff
@@ -373,18 +382,23 @@ extension PaymentSheet {
...
+        ) -> PaymentMethodAvailabilityStatus {
             let fulfilledRequirements = [configuration, intent].reduce([]) {
                 (accumulator: [PaymentMethodTypeRequirement], element: PaymentMethodRequirementProvider) in
                 return accumulator + element.fulfilledRequirements
             }
             let supports = Set(requirements).isSubset(of: fulfilledRequirements)
-            return supports
+            if !supports {
+                let missingRequirements = Set(requirements).subtracting(fulfilledRequirements)
+                return .missingRequirements(Array(missingRequirements))
```
sfriedman-stripe, line 398: Could be missing something, but if we're converting to a `Set` and then back to an `Array` in a couple places, why not just make the associated value on `PaymentMethodAvailabilityStatus.missingRequirements` a `Set`? Do we care about the ordering somewhere?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+PaymentMethodAvailability.swift
```diff
@@ -132,5 +132,63 @@ extension PaymentSheet {
...
+        }
+
+        static func ==(lhs: PaymentMethodAvailabilityStatus, rhs: PaymentMethodAvailabilityStatus) -> Bool {
+            switch (lhs, rhs) {
+            case (.notSupported, .notSupported):
+                return true
+            case (.supported, .supported):
+                return true
+            case (.unactivated, .unactivated):
+                return true
```
sfriedman-stripe, line 185: Nit: I don't have Xcode open, but if memory serves we can condense this to:

```suggestion
            case (.notSupported, .notSupported),
                 (.supported, .supported),
                 (.unactivated, .unactivated):
                return true
```

sfriedman-stripe (approved):
Seems like a great dev exp improvement. Thanks for doing this!

PR #2319, opened 2023-02-23
Deferred intent playground

porter-stripe:
## Summary
- Adds a deferred selector to the payment sheet playground
- Adds a few tests
- Fixes a few assertions for Klarna and Afterpay to support deferred intents
Simulator Screen Shot - iPhone 12 mini - 2023-02-23 at 14 13 31

## Motivation
- Deferred workstream

## Testing
- Manual
- New tests

## Changelog
N/A


Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
```diff
@@ -321,12 +357,23 @@ class PaymentSheetTestPlayground: UIViewController {
...
         let mc: PaymentSheet
-        switch intentMode {
-        case .payment, .paymentWithSetup:
-            mc = PaymentSheet(paymentIntentClientSecret: clientSecret!, configuration: configuration)
-        case .setup:
-            mc = PaymentSheet(setupIntentClientSecret: clientSecret!, configuration: configuration)
+        let mode: PaymentSheet.InitializationMode
+
+        switch self.initMode {
+
```
yuki-stripe, line 363: ultranit: delete this line

yuki-stripe: wait is this what the linter wants?

Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
```diff
@@ -247,6 +265,24 @@ class PaymentSheetTestPlayground: UIViewController {
         return configuration
     }
 
+    var intentConfig: PaymentSheet.IntentConfiguration {
+
+        switch intentMode {
+        case .payment:
+            return PaymentSheet.IntentConfiguration(mode: .payment(amount: 1000, currency: "USD",
+                                                                   setupFutureUsage: nil),
+                                                                captureMethod: .automatic)
```
yuki-stripe, line 274: All of these properties and payment_method_types should match what the toggles say. 
e.g. if the toggle says 'currency = EUR', the currency here should be 'EUR'.

In practice it's probably simpler to not look at the toggles to determine the values, but instead look at the Intent returned by our backend.

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
```diff
@@ -221,6 +221,54 @@ class PaymentSheetUITest: XCTestCase {
         XCTAssertNotNil(successText.label.range(of: "Your order is confirmed!"))
     }
 
+    func testDeferredPaymentIntent() {
```
yuki-stripe, line 224: With these tests, are the snapshot tests you added earlier still useful?

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
```diff
@@ -221,6 +221,54 @@ class PaymentSheetUITest: XCTestCase {
...
+        )
+
+        app.buttons["Checkout (Complete)"].tap()
+        let setupButton = app.buttons["Set up"]
+        XCTAssertTrue(setupButton.waitForExistence(timeout: 10.0))
+
+        // TODO(porter) Finish test when we can confirm server side
+    }
+
+    func testDeferredPaymentIntent_FlowController() {
```
yuki-stripe, line 255: why these particular tests and not others e.g. `testDeferredSetupIntent_FlowController`?

in general I haven't thought about how to test the deferred flows, eager to hear your thoughts!

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
porter-stripe: Yeah, these are essentially snapshot tests now but really they will be confirm tests.

If these tests make the snapshot tests not useful, then I would imagine that would apply for many of the other snapshots as well.

Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
porter-stripe: no, just bad formatting on my end

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
porter-stripe: No rhyme or reason really, I haven't given much thought to how we will test the deferred flows. I think that will become more clear when we understand the nuance of server side confirmation better.

For now I'm considering these stubs to make writing the real tests easy while still providing some value. I'm sure we will end up adding more tests as the server side confirmation stuff becomes clearer.

I did add `testDeferredSetupIntent_FlowController` for consistency.

Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
porter-stripe: Good catch, I changed this to look at the currency toggle.

I figure we shouldn't be looking at an intent returned by our sever for the deferred workflow 🙈

yuki-stripe: Currency was just one example

- amount, setup_future_usage, capture method needs to match the values on the Intent 
- p_m_types needs to match the Intent if automatic pm types is off

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
yuki-stripe: It might, yeah!  The way I'm thinking about it, snapshot tests are for testing UI, and deferred workflow is unrelated to UI, so we don't need snapshot tests for it.

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
yuki-stripe: Ok, makes sense

Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
porter-stripe: Still doesn't make sense to me... my understanding is in a real-world deferred case the user would not create an intent until confirm time. Why would we make the playground create an intent at loading time (and read it), then presumably create another intent at confirm time? I think the most correct thing here to do would not even make a request to glitch to create an intent when we load and only make a request to glitch when we confirm. The only reason I left the loading in there was because the entire state of the playground is so tied to that function (`load`) e.g. re-enabling toggles. Maybe we need to change that to not even create an intent when we load in the deferred case to closely match the real world use case.

re: p_m_types, shouldn't that come from the elements session response or the `IntentConfiguration.paymentMethodTypes`, why would we read that off the intent?

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
porter-stripe: The way I viewed those snapshot tests were "We correctly render PaymentSheet UI when provided an `IntentConfiguration`". These automated UI tests may miss subtle unintended changes to the UI as they only really look for specific buttons/text that are critical to the UI flow. It seems plausible that a UI bug could crop up only when loading from a deferred intent/intent config that would be missed during automated tests since we don't look at every UI element on screen. The snapshot gives us 100% coverage of what we expect the UI to look like when loaded from an intent config while these do not.

Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
```diff
@@ -464,6 +521,8 @@ extension PaymentSheetTestPlayground {
             self.clientSecret = json["intentClientSecret"]
             self.ephemeralKey = json["customerEphemeralKeySecret"]
             self.customerID = json["customerId"]
+            self.paymentMethodTypes = json["paymentMethodTypes"]?.components(separatedBy: ",")
```
porter-stripe, line 524: Left this as a string for easier decoding, also didn't want to possibly break Android  or master by changing our server from returning [String: String] to [String: Any]

porter-stripe:
@yuki-stripe this is ready for another look

yuki-stripe (dismissed):

yuki-stripe (approved):

PR #2321, opened 2023-02-24
Enable billing details collection for cards

eurias-stripe:
## Summary
Enabling billing details collection configuration for Credit Cards.

## Details
This is the first step to enable billing details collection configuration for all PMs.
In this PR I'm making the API additions (gated behind `@_spi`), and enabling for credit cards.
Here is screencast showing multiple permutations of the possible options: 

https://user-images.githubusercontent.com/114543737/221060718-b3005ef4-694d-45fc-b1da-ed6d816fde11.mp4

|                                                                                                |                                                                                                                                                                                                |                                                                                                                                                                                                |
| ---------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| **Settings**                                                                                   | **UI**                                                                                                                                                                                         | **Payment result**                                                                                                                                                                             |
| name = .always<br/>phone = .always<br/>email = .always<br/>address = .full                                    |  |  |
| name = .always<br/>phone = .always<br/>email = .always<br/>address = .never                                   |  |  |
| name = .never<br/>phone = .never<br/>email = .never<br/>address = .never<br/>attachDefaultsToPaymentMethod = true  |  |  |
| name = .never<br/>phone = .never<br/>email = .never<br/>address = .never<br/>attachDefaultsToPaymentMethod = false |  |  |
| Link is a special case because it cannot collect contact information, since it already collected it. <br/><br/> name = .always<br/>phone = .always<br/>email = .always<br/>address = .full |  |  <br/><br/>I cannot find the data attached to the actual PM since dashboard only shows Link and I couldn't find it through the object browser either. |

## Motivation
https://docs.google.com/document/d/1eBKVqbdhky_b8ux5f_eKp1HdHxgbV1WdF_XQWp0kdk0/edit#heading=h.dofjidvfxcoc

## Testing
- Added unit and UI tests.
- Tested multiple permutations and verified that billing details were attached as expected.

## Changelog
A changelog entry will be added once this is ready to go public


eurias-stripe:
I also made some changes to the playground to add the new options and since it was getting way too long I made the Checkout (complete) button sticky

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
```diff
@@ -307,4 +328,52 @@ extension PaymentSheet {
...
+
+        /// How to collect the billing address.
+        /// Defaults to `automatic`.
+        public var address: AddressCollectionMode = .automatic
+
+        /// Whether the values included in `Configuration.defaultBillingDetails` should be attached to the payment
+        /// method, this includes fields that aren't displayed in the form.
+        ///
+        /// If `false` (the default), those values will only be used to prefill the corresponding fields in the form.
+        public var attachDefaultsToPaymentMethod = false
```
eurias-stripe, line 377: I added this field as a temporary way to not attach the defaults to avoid breaking the current functionality while this is in development, but the more I think about it, the more I think it is a good idea to add this in the final product so people have to opt-into this behavior, if people think this is a good idea I'll add it to the API review.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
```diff
@@ -409,6 +424,22 @@ extension PaymentSheetFormFactory {
...
+
+    func makeContactInformation(includeName: Bool, includeEmail: Bool, includePhone: Bool) -> SectionElement? {
+        let nameElement = includeName ? makeName() : nil
+        let emailElement = includeEmail ? makeEmail() : nil
+        let phoneElement = includePhone ? makePhone() : nil
+
+        let allElements: [Element?] = [nameElement, emailElement, phoneElement]
+        let elements = allElements.compactMap { $0 }
+
+        guard !allElements.isEmpty else { return nil }
```
bmelts-stripe, line 436: This should be checking `elements`, right?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/PaymentMethodElement/PaymentMethodElement.swift
```diff
@@ -16,15 +16,37 @@
...
 protocol PaymentMethodElement: Element {
-    /**
-     Modify the params according to your input, or return nil if invalid.
-     - Note: This is called on the Element hierarchy in depth-first search order.
-     */
+    /// Modify the params with default values.
+    ///
+    /// This method is called before `updateParams(params:)` and should be used to populate `params` with any necessary
+    /// default values, these can include values for fields not included in the element hierarchy.
+    func applyDefaults(params: IntentConfirmParams) -> IntentConfirmParams
```
bmelts-stripe, line 23: This addition ends up having a ton of surface area, so I'd love to learn a little more about it. Right now only cards are passing in a `defaultsApplier` - do we expect different payment methods to want to apply different defaults? Nothing in that closure seems specific to cards, or to the context the element's being created in, so it's not clear to me at a glance the value of this over just initializing the params you pass with the defaults (if the configuration allows for it). Some clarity on that would be super helpful!

Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
```diff
@@ -11,7 +11,7 @@
 //  This exposes internal functionality which may cause unexpected behavior if used directly.
 import Contacts
 import PassKit
-import StripePaymentSheet
+@_spi(STP) import StripePaymentSheet
```
wooj-stripe, line 14: I'm guessing this is temporary? We gaveled on a way to do betas in the mobile SDK -- i will link you via slack

eurias-stripe: This won't be a beta, I'm just using `@_spi` to gate it during development.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/PaymentMethodElement/PaymentMethodElement.swift
eurias-stripe: This will be used by all payment methods, in this PR I'm only adding support for credit cards, but I need a way to inject this into the flow.
The values will depend on the context, they don't in this case because cards are the simplest case, but for example sofort has an special field for country, is not part of the address.
I'm open to alternatives on how to do this, I spend a good chunk of time thinking of a good solution is this is the best I came up with, but happy to discuss more.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
eurias-stripe: After discussion, this field will be part of the final public API

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/STPAPIClient+Link.swift
```diff
@@ -434,6 +435,9 @@ extension STPPaymentMethodBillingDetails {
 
     var consumersAPIParams: [String: Any] {
         var params = STPFormEncoder.dictionary(forObject: self)
+        // Consumers API doesn't support email or phone.
+        params["email"] = nil
+        params["phone"] = nil
```
wooj-stripe, line 440: I think this only removes the value, not the key, right? Seems like we should remove both key and value?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/CardSectionWithScanner/CardSectionWithScannerElement.swift
```diff
@@ -32,14 +32,31 @@ final class CardSection: ContainerElement {
...
+        theme: ElementsUITheme = .default
+    ) {
         self.theme = theme
-        let panElement = PaymentMethodElementWrapper(TextFieldElement.PANConfiguration(), theme: theme) {  field, params in
+        let nameElement = collectName
+            ? PaymentMethodElementWrapper(
+                TextFieldElement.NameConfiguration(
+                    type: .full,
+                    defaultValue: defaultName,
+                    label: STPLocalizedString("Name on card", "Label for name on card field")),
```
wooj-stripe, line 52: I found this surprising because I would have expected name to be a part of the billing section, not showing up as "name on card", but i suspect there is good reason for this.  Any documentation you can point me to on this?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
```diff
@@ -145,6 +145,21 @@ extension PaymentSheetFormFactory {
         }
     }
 
+    func makePhone(apiPath: String? = nil) -> PaymentMethodElementWrapper<PhoneNumberElement> {
```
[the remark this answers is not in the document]
eurias-stripe: That's what I thought too, but I like adding it for consistency, and it doesn't add any significant overhead.

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/STPAPIClient+Link.swift
eurias-stripe: setting to nil removes the value, if there is no value there is no key

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/CardSectionWithScanner/CardSectionWithScannerElement.swift
eurias-stripe: There isn't a good reason for this, other than it's what I think looks best, I experimented with multiple ways and showed them to @jarvis-stripe, for context, checkout in web shows the name field on its own and it looks kinda weird, but it is one of the options.

This is web checkout:

image (1)

And these are some of the options I tried:
image (2)
image (3)
image (4)

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
eurias-stripe: Good catch

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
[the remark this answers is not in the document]
sfriedman-stripe: It does seem like additional cognitive overhead that isn't needed though right? Why add a param we're not using?

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetBillingCollectionUITests.swift
```diff
@@ -0,0 +1,163 @@
...
+//
+
+import XCTest
+
+final class PaymentSheetBillingCollectionUITests: XCTestCase {
+    var app: XCUIApplication!
+
+    override func setUpWithError() throws {
+        try super.setUpWithError()
+        // Put setup code here. This method is called before the invocation of each test method in the class.
```
sfriedman-stripe, line 15: Nit: We should remove boilerplate comments like this

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetBillingCollectionUITests.swift
```diff
@@ -0,0 +1,163 @@
...
+                "collect_name": "auto",
+                "collect_email": "auto",
+                "collect_phone": "auto",
+                "collect_address": "auto",
+            ]
+        )
+        app.buttons["Checkout (Complete)"].tap()
+
+        guard let card = scroll(collectionView: app.collectionViews.firstMatch, toFindCellWithId: "card") else {
+            XCTFail()
```
sfriedman-stripe, line 45: Should we fail with a message to make the error more diagnosable, or does Xcode already give us enough diagnostic info here?

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetBillingCollectionUITests.swift
```diff
@@ -0,0 +1,163 @@
...
+            return
+        }
+        card.tap()
+        try! fillCardData(app)
+
+        // Complete payment
+        app.buttons["Pay $50.99"].tap()
+        let successText = app.alerts.staticTexts["Success!"]
+        XCTAssertTrue(successText.waitForExistence(timeout: 10.0))
```
sfriedman-stripe, line 54: This is a long timeout. Seems like we should shorten to keep tests snappier when they fail

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/STPAPIClient+Link.swift
```diff
@@ -448,8 +452,34 @@ extension STPPaymentMethodAddress {
...
+        }
+        if let locality = params["locality"] as? String, locality.isEmpty {
+            params["locality"] = nil
+        }
+        if let administrativeArea = params["administrative_area"] as? String, administrativeArea.isEmpty {
+            params["administrative_area"] = nil
+        }
+        if let countryCode = params["country_code"] as? String, countryCode.isEmpty {
+            params["country_code"] = nil
+        }
```
sfriedman-stripe, line 482: This would be a lot cleaner as a `map`, e.g.

```swift
let consumerKeyMap = [
  "line1": "line_1",
  "line2": "line_2",
  "city": "locality",
  "state": "administrative_area",
  "country": "country_code"
]

return STPFormEncoder.dictionary(forObject: self).map { key, value in
  let newKey = consumerKeyMap[key] ?? key
  let newValue: Any?

  // Consumers API doesn't like empty strings. Turn any empties into nils
  if let stringValue = value as? String, stringValue.isEmpty {
    newValue = nil
  } else {
    newValue = value
  }

  return (newKey, newValue)
}
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/PaymentMethodElement/PaymentMethodElement.swift
```diff
@@ -43,10 +65,28 @@ extension ContainerElement {
...
+extension SectionElement: PaymentMethodElement {
+    func applyDefaults(params: IntentConfirmParams) -> IntentConfirmParams {
+        applyDefaultsHierarchichally(params: params)
+    }
+}
+
 extension StaticElement: PaymentMethodElement {
+    func applyDefaults(params: IntentConfirmParams) -> IntentConfirmParams {
+        return params
+    }
```
sfriedman-stripe, line 83: Do we need this since it's already defined in the `PaymentMethodElement` extension?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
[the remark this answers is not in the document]
eurias-stripe: That's fair, IMO it doesn't really add much cognitive load, I might actually think that not having the param could add confusion, i.e., why does every other helper has an `apiPath` param and this one doesn't?.
I don't feel strongly either way, and given that there are two comments mentioning that this is not necessary I think it makes sense to remove it.

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetBillingCollectionUITests.swift
eurias-stripe: This particular case does need a significant timeout because there is a network request going on, if we make it shorter, say 5 seconds, it might work in most cases, but there is also a chance that it would cause flakiness due to the request taking longer.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/PaymentMethodElement/PaymentMethodElement.swift
eurias-stripe: Unfortunately yes, because `applyDefaults` is implemented in the `PaymentMethodElement` extension and `ContainerElement` extension, so it becomes ambiguous.
I tried a few ways to make it work without reimplementing but none worked, so that's why I extracted the actual implementation to a helper method so all the extensions could call it.

eurias-stripe: Scratch that, I just realized that this comment is on the `StaticElement` extension, here it _can_ be removed, my previous comment stands for `FormElement` and `SectionElement`

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/STPAPIClient+Link.swift
eurias-stripe: Nice one, didn't think of it

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetBillingCollectionUITests.swift
eurias-stripe: Xcode will fail on this line, so that should be enough context, especially given that this is only meant to safely unwrap a value that should always exist, i.e. we're not testing whether there is a card option, there should be another test for that, we're just trying to get that option and tap it, but the result is an optional so it needs to be unwrapped.

eurias-stripe:
FYI, @jarvis-stripe took a look at the different options and agrees that putting it inside the card info is the most natural.

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
```diff
@@ -823,7 +823,84 @@ extension PaymentSheetUITest {
         try loginAndPay()
     }
 
-    private func loginAndPay() throws {
+    func testLinkAddCard_CollectingBillingDetails() throws {
```
wooj-stripe, line 826: Because cards are so important, seems like we should add a similar test with link off?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/PaymentMethodElement/PaymentMethodElement.swift
```diff
@@ -16,15 +16,37 @@
...
+
+    /// Modify the params according to your input, or return nil if invalid.
+    /// - Note: This is called on the Element hierarchy in depth-first search order.
     func updateParams(params: IntentConfirmParams) -> IntentConfirmParams?
 }
 
 // MARK: - Default implementations
 extension ContainerElement {
+    func applyDefaults(params: IntentConfirmParams) -> IntentConfirmParams {
+        applyDefaultsHierarchichally(params: params)
```
wooj-stripe, line 33: s/applyDefaultsHierarchichally/applyDefaultsHierarchically/ or maybe "recursively" is a more descriptive term?

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
eurias-stripe: That's the existing behavior and it is already covered by the link tests

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/PaymentMethodElement/PaymentMethodElement.swift
eurias-stripe: renamed

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/PaymentMethodElement/PaymentMethodElement.swift
eurias-stripe: If there are no concerns or suggestions regarding this I'll resolve the comment

wooj-stripe (approved):
lgtm!

eurias-stripe:
I'm gonna merge this to unblock myself, @bmelts-stripe if you have any comments regarding the Link parts please let me know and I can make changes in a follow up PR

bmelts-stripe:
Link parts lg2m, thanks @eurias-stripe!

PR #2344, opened 2023-03-01
Parallelize StripeiOSTests

eurias-stripe:
## Summary
This was dropped when moving to test plans.

## Motivation
Trying to improve test time.

## Testing
CI

### DIFF
--- a/Stripe/StripeiOSTests.xctestplan
+++ b/Stripe/StripeiOSTests.xctestplan
@@ -20,10 +20,12 @@
       "containerPath" : "container:Stripe.xcodeproj",
       "identifier" : "ADF894AA8F6022D9BED17346",
       "name" : "StripeiOS"
-    }
+    },
+    "testExecutionOrdering" : "random"
   },
   "testTargets" : [
     {
+      "parallelizable" : true,
       "skippedTests" : [
         "PaymentMethodMessagingViewSnapshotTests",
         "STPCardBINMetadataTests",


sfriedman-stripe (dismissed):
So we were already doing this before? Don't see any issues turning it back on then

davidme-stripe (dismissed):
Yep, thanks for fixing this!

davidme-stripe:
Looks like this isn't actually parallelizing the tests, we'll need to also pass a different number of executors into the job. On our call with Bitrise today, they mentioned ~3-5 might be good for M1 Max.

eurias-stripe:
This change seems to have cut the execution time of StripeiOSTests in half.

eurias-stripe:
image

vs a random execution

image

eurias-stripe:
> Looks like this isn't actually parallelizing the tests, we'll need to also pass a different number of executors into the job. On our call with Bitrise today, they mentioned ~3-5 might be good for M1 Max.

Hmm, it does seem to have improved times though, why would that be if it isn't actually parallelizing?

eurias-stripe:
> That's true, I didn't dig into the individual tests. But when I look at the total time for `framework-tests`, I see 11min 40s vs 11min 33s. If you look at the individual items in your screenshot, tests that were taking 5-7s with parallelism disabled are now taking 18-30s with parallelism enabled, so it's a wash.
> 
> Maybe we should only enable parallelism on the main StripeiOS build? Or try it on the M1 Max?

I noticed that, I am only enabling it on the StripeiOS target, which is the only one using test plans so I don't see how would that affect the other tests. I'll do more tests to figure it out.

eurias-stripe:
No significant improvement on M1 Max https://app.bitrise.io/build/e40a325f-fe36-4dbe-9ee6-061d4c4caea3.
The StripeiOS Tests decreased by another ~20 seconds, but overall time stays the same.
Note that the total time of 9:31 is what it took on the regular M1 after the first build, I assume subsequent builds are shorter because of caching.

eurias-stripe:
Reverted to original state, just enabling parallelization, we can look into overall improvements later.

eurias-stripe:
Hmm, checking again it seems like builds normally take ~11:30, I'm sure I saw some build taking ~9:30 but looks like I'm wrong, if this is the case this would be a 2 minute improvement by using the M1 Max with parallelization, that's significant and might be worth doing.

davidme-stripe:
I'm seeing 10m30s-10m45s on a few randomly selected PR builds: https://app.bitrise.io/app/b220f4ba85d134a7/pipelines/000b4c7c-d098-4c79-bc8d-0186b772479d
https://app.bitrise.io/app/b220f4ba85d134a7/pipelines/653ae60d-ed9f-473e-a3e4-4e8c3a5c9c6a

So for a 1min difference, I don't think it's worth ~2x-ing our costs. I'm surprised enabling parallelization on the first StripeiOS tests slows down the subsequent tests so much...

eurias-stripe:
> Hmm, checking again it seems like builds normally take ~11:30, I'm sure I saw some build taking ~9:30 but looks like I'm wrong, if this is the case this would be a 2 minute improvement by using the M1 Max with parallelization, that's significant and might be worth doing.

Worth noting that the credit difference is 80 vs 48, so almost double for ~20% build time reduction.
Also after this the largest bottle neck becomes size-report

eurias-stripe:
> I'm seeing 10m30s-10m45s on a few randomly selected PR builds: https://app.bitrise.io/app/b220f4ba85d134a7/pipelines/000b4c7c-d098-4c79-bc8d-0186b772479d https://app.bitrise.io/app/b220f4ba85d134a7/pipelines/653ae60d-ed9f-473e-a3e4-4e8c3a5c9c6a
> 
> So for a 1min difference, I don't think it's worth ~2x-ing our costs. I'm surprised enabling parallelization on the first StripeiOS tests slows down the subsequent tests so much...

The 11:30 time comes from the latest build on master https://app.bitrise.io/app/b220f4ba85d134a7/pipelines/486b71aa-9ba7-487c-b330-51c76c9ba8c7.

I agree it is not worth it, especially because the overall build time only improves by about a minute, with size-report becoming the new bottleneck.

eurias-stripe:
Reverted to the original version only enabling parallelization on StripeiOS Tests.

davidme-stripe:
Closing this out, we handled parallelization recently in another PR.

PR #2349, opened 2023-03-03
FC Networking: polish around LinkAccountPicker

kgaidis-stripe:
## Summary

Two things:
1. Corrected data access language in LinkAccountPicker to say "with Link" instead of "with Stripe"
2. Added support for disabled accounts

### Corrected Data Merchant Access 

iOS_After_MerchantDataAccess

### Disabled Account via SwiftUI Previews

Screen Shot 2023-03-03 at 11 57 51 AM

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/Native/LinkAccountPicker/LinkAccountPickerViewController.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/Native/LinkAccountPicker/LinkAccountPickerViewController.swift
@@ -173,7 +173,7 @@ final class LinkAccountPickerViewController: UIViewController {
                         } else {
                             // this shouldn't happen, but in case it does, we navigate to `institutionPicker` so user
                             // could still have a chance at successfully connecting their account
-                            self.delegate?.linkAccountPickerViewController(self, didRequestNextPane: .institutionPicker)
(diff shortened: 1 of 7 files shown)


StripeFinancialConnections/StripeFinancialConnections/Source/Native/LinkAccountPicker/LinkAccountPickerViewController.swift
```diff
@@ -45,7 +45,7 @@ final class LinkAccountPickerViewController: UIViewController {
     private weak var bodyView: LinkAccountPickerBodyView?
     private lazy var footerView: LinkAccountPickerFooterView = {
         return LinkAccountPickerFooterView(
-            isStripeDirect: dataSource.manifest.isStripeDirect ?? false,
+            isStripeDirect: false,
```
kgaidis-stripe, line 48: Stripe.js does this (likely because we don't want to show Stripe in one place, and Link in another):
- https://git.corp.stripe.com/stripe-internal/stripe-js-v3/blob/957a6a341820b5872640c38d3caec7022d7a04c3/src/linkedAccounts/inner/components/panes/LinkAccountPicker/index.tsx#L51

vardges-stripe (approved):

PR #2320, opened 2023-02-23
[Link] Add initial StripeLinkCore module

bmelts-stripe:
## Summary
Add StripeLinkCore as a module.

## Motivation
https://docs.google.com/document/d/17jTE9n9ekPZnoQv7cJlTaAeiuq01dW6vi3w8o8vASVU/edit?pli=1#heading=h.r03jhtm7a53r

## Changelog
Once the dependencies get added, we'll need to figure out a versioning/changelog strategy for communicating this to Carthage/manual integration users, as it will require some effort from them to update.

PR #2351, opened 2023-03-03
[Link] StripePaymentSheet depends on StripeLinkCore

bmelts-stripe:
## Summary
Added StripeLinkCore as a dependency of StripePaymentSheet, and updated documentation. cc @kgaidis-stripe 

## Motivation
Part of the ongoing module creation process (see: https://github.com/stripe/stripe-ios/pull/2320, https://github.com/stripe/stripe-ios/pull/2335).

## Testing
Ran through wiki instructions, everything works as expected.

## Notes
I copied the wording in the documentation updates from the wiki, but I'm open to discussion on if there are better ways we should communicate this. I've also left the version undefined for `MIGRATING.md` - we should decide on what kind of bump we want to do for this (major or minor).


kgaidis-stripe:
@bmelts-stripe 

Does the merging of this PR mean that people will have to perform migration steps IF they use `StripePaymentSheet`?

Since this is a little bit of a migration commitment, I wonder what you think about trying to do this once for _any_ framework that will eventually depend on `StripeLinkCore`? Is it just `StripePaymentSheet` and then eventually `FinancialConnections`? For example, I wonder if Financial Connections should do this _now_ (even if we don't use it at the moment) because we will have to anyway. CC: @vardges-stripe

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 24.0.0 yyyy-mm-dd
```
eurias-stripe, line 1: leave this as `x.x.x yyyy-mm-dd` the release script will replace it with the right version/date

eurias-stripe (approved):

davidme-stripe (changes requested):
We're holding off on a major version bump for now.

PR #2355, opened 2023-03-06
Verify xcode version before proposing release

eurias-stripe:
## Summary
Verify xcode version before proposing release

## Motivation
I keep forgetting to switch versions before proposing.

## Testing
Ran the scripts with wrong xcode version, the scripts aborted

### DIFF
--- a/ci_scripts/common.rb
+++ b/ci_scripts/common.rb
@@ -3,6 +3,21 @@
 require 'colorize'
 require 'English'
 
+# This should generally be the minimum Xcode version supported by the App Store, as the
+# compiled XCFrameworks won't be usable on older versions.
+# We sometimes bump this if an Xcode bug or deprecation forces us to upgrade early.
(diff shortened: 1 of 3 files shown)


eurias-stripe:
> Does the proposer need that Xcode version? That wasn't the case before, but is it required because of tuist?

Correct, because the projects need to be generated for xcode 13, specifically the package.resolved file

davidme-stripe:
Do we need to check in the Package.resolved? It only impacts tests, so I don't think it's relevant to the release itself

eurias-stripe:
> Do we need to check in the Package.resolved? It only impacts tests, so I don't think it's relevant to the release itself.

Package.resolved is inside the xcworkspace, I guess we could technically not include it, but then it would cause the project to fail building by default because of the missing packages, xcode is not very smart, it won't resolve them again unless you resolve them manually.

davidme-stripe (approved):

PR #2362, opened 2023-03-07
Fix deferred tests

porter-stripe:
## Summary
- A few deferred tests were failing and were disabled in the deploy yesterday
- One was failing due to billing details not being present for Apple Pay. I'm guessing this passed originally because a test before it may have ran and added the billing details but the test order is random causing it to fail intermediately.
- The payment methods returned from elements session randomly changed to just `card`.  It's pretty unclear why, but I changed it to just card and cash app which fixes the issue but we may want to just check if the returned payment methods are not empty since the behavior seems flaky. Something we should follow up on.

## Motivation
Deferred workflow

## Testing
N/A

## Changelog
N/A

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
+++ b/Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
@@ -776,8 +776,7 @@ extension PaymentSheetUITest {
         // TODO(porter) Finish test when we can confirm server side
     }
 
-    // TODO(porter): Fix test.
-    func disabled_testDeferredPaymentIntent_ApplePay() {
+    func testDeferredPaymentIntent_ApplePay() {
         loadPlayground(
             app,
             settings: [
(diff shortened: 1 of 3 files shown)


Stripe/StripeiOSTests/PaymentSheet+APITest.swift
```diff
@@ -218,8 +218,8 @@ class PaymentSheetAPITest: XCTestCase {
         let confirmExpectation = XCTestExpectation(description: "Confirm deferred intent")
         let callbackExpectation = XCTestExpectation(description: "Confirm callback invoked")
 
-        let types = ["ideal", "card", "bancontact", "sofort"]
-        let expected: [STPPaymentMethodType] = [.card, .iDEAL, .bancontact, .sofort]
+        let types = ["card", "cashapp"]
+        let expected: [STPPaymentMethodType] = [.card, .cashApp]
```
porter-stripe, line 222: Maybe we just check if the returned payment methods are not empty?

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
```diff
@@ -799,6 +798,38 @@ extension PaymentSheetUITest {
         XCTAssertTrue(cardButton.waitForExistence(timeout: 10.0))
         cardButton.forceTapElement()
 
+        // Fill out billing details if required
+        let addBillingDetailsButton = applePay.buttons["Add Billing Address"]
```
yuki-stripe, line 802: do we do this elsewhere?  I'm not finding it.

Stripe/StripeiOSTests/PaymentSheet+APITest.swift
yuki-stripe: that also makes sense

yuki-stripe (approved):

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
porter-stripe: I can't find it anywhere, but that was my best guess as to how it passed in the first place...

porter-stripe: Somehow the simulator had a billing address attached to Apple Pay when CI passed

Stripe/StripeiOSTests/PaymentSheet+APITest.swift
porter-stripe: Going to leave as-is for now since we know the root cause

PR #2364, opened 2023-03-07
Client and server side confirmation for deferred intents

porter-stripe:
## Summary
- Adds confirmation for client and server side confirmation for deferred intents
- Updates tests
- Still outstanding: UI tests for server side confirmation blocked by glitch updates and update playground for server side confirmation.

## Motivation
Deferred intents

## Testing
- Updated tests
- Manual

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
```diff
@@ -35,6 +35,17 @@ extension PaymentSheet {
         paymentHandler: STPPaymentHandler,
         completion: @escaping (PaymentSheetResult) -> Void
     ) {
+
+        // If processing a deferred intent, store state for later processing
```
porter-stripe, line 39: Open to suggestions here on the singleton. Since the load/confirm is static we need to save it in a static variable for later use.

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
```diff
@@ -749,14 +741,6 @@ extension PaymentSheetUITest {
             nameField.typeText("Jane Done")
         }
 
-        // Terms and privacy policy
-        for linkText in ["Terms", "Privacy Policy"] {
```
porter-stripe, line 753: Flaky and not needed.

porter-stripe:
Closing for now, going to re-open once the playground is updated for SSC now that I'm unblocked.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
```diff
@@ -582,8 +591,24 @@ extension PaymentSheet {
...
     }
 
+    /// - Note: Should only be invoked by merchant code via the `IntentConfiguration.ConfirmHandler.intentCreationCallback`
     private static func _handleIntentCreation(_ result: Result<String, Error>) {
-        // TODO(porter) Deferred intent confirmation handling
+        guard let deferredContext = DeferredIntentContext.current else {
+            fatalError("DeferredIntentContext.current unexpectedly nil")
+        }
+
+        DispatchQueue.main.async {
```
porter-stripe, line 600: Link was crashing if not called on main thread, I think this is better than asking merchant to invoke the callback on the main thread.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
```diff
@@ -616,8 +641,57 @@ extension PaymentSheet {
             } else {
                 fatalError("Either confirmHandler or confirmHandlerForServerSideConfirmation must be non-nil when using deferred intents")
             }
+        }
+    }
+
+    private static func confirmClientSide(for deferredContext: DeferredIntentContext, with clientSecret: String) {
+        // Retrieve either the payment or setup intent, then confirm it client side
+        deferredContext.configuration.apiClient.retrieveIntent(for: deferredContext.intentConfig,
+                                                                  withClientSecret: clientSecret) { result in
```
yuki-stripe, line 650: We can follow up in another PR, but we should try to avoid this retrieval.

Looks like besides the client_secret we inspect the PI `shipping`, but I don't think we need to do that for the deferred case.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
yuki-stripe: Hrm I can see how this is convenient, but we shouldn't need it - we can flow everything we need across the various methods called in the deferred flow.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
```diff
@@ -616,8 +641,57 @@ extension PaymentSheet {
             } else {
                 fatalError("Either confirmHandler or confirmHandlerForServerSideConfirmation must be non-nil when using deferred intents")
             }
+        }
+    }
+
+    private static func confirmClientSide(for deferredContext: DeferredIntentContext, with clientSecret: String) {
```
yuki-stripe, line 647: I'm trying to think how to make this whole flow easier to follow in the code.  I think one 'master' method that all the `.deferred` cases in `confirm` call would help.  The highlighted method here is almost that, except we don't use it in the saved PM case.

The pattern we have here where callbacks call other methods makes it hard to follow the code. I think we should take advantage of async/await to eliminate that.

This 'master' method would look roughly like
```
func handleDeferredIntentConfirmation(paymentMethod: STPPaymentMethod?, completionHandler: ..) {
  do {
    // Create PM if necessary
    let paymentMethod = paymentMethod ? try await createPaymentMethod(...)
    // Get Intent client secret from merchant
    let clientSecret = try await fetchIntentClientSecretFromMerchant(...)
    if isServerSideConfirmation {
       // Handle next actions
    } else {
       // Confirm payment
    }
  } catch {
    completionHandler(.failure(error))
  }
}
```

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
```diff
@@ -749,12 +741,221 @@ extension PaymentSheetUITest {
...
+
+        app.buttons["Checkout (Complete)"].tap()
+        let applePayButton = app.buttons["apple_pay_button"]
+        XCTAssertTrue(applePayButton.waitForExistence(timeout: 4.0))
+        applePayButton.tap()
+
+        let applePay = XCUIApplication(bundleIdentifier: "com.apple.PassbookUIService")
+        _ = applePay.wait(for: .runningForeground, timeout: 10)
+
+        let predicate = NSPredicate(format: "label CONTAINS 'Simulated Card - AmEx, ‪•••• 1234‬'")
```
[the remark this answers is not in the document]
porter-stripe: Yeah we might be able to, FWIW, we have a similar test: https://github.com/stripe/stripe-ios/blob/master/Testers/IntegrationTester/IntegrationTesterUITests/IntegrationTesterUITests.swift#L173

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
```diff
@@ -749,12 +741,221 @@ extension PaymentSheetUITest {
...
+            stateCell.typeText("CA")
+
+            let zipCell = applePay.tables.cells["ZIP"]
+            zipCell.tap()
+            zipCell.typeText("95014")
+
+            applePay.buttons["Done"].tap()
+        }
+
+        cardButton = applePay.buttons["Simulated Card - AmEx, ‪•••• 1234‬"].firstMatch
```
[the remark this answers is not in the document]
porter-stripe: We don't need to overwrite necessarily but we do need to re-query. It is a different button on a different screen with the same text.

I renamed it for clarity (although not sure it's actually more clear now haha)

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
porter-stripe: This looks good! But I don't understand what `fetchIntentClientSecretFromMerchant` would be doing. We can't fetch it from the merchant, only wait for it via the two callbacks.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
porter-stripe: This implies we can confirm an intent w/o actually having the intent object...?

porter-stripe: My understanding was that we split these into two handlers so we could avoid this retrieval for SSC but it was needed for CSC.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
yuki-stripe: It'd look something like
```
func fetchIntentClientSecretFromMerchant(...) async throws -> String {
  try await withCheckedThrowingContinuation { continuation in
    intentConfiguration.confirmCallback(..) { result in
     continuation.resume(with: result)
    }
  }
}
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
yuki-stripe: > This implies we can confirm an intent w/o actually having the intent object...?
Right. You won't find the object being passed into the `confirmPaymentIntent` call, just the client secret inside the `STPPaymentIntentParams` object.

> My understanding was that we split these into two handlers so we could avoid this retrieval for SSC but it was needed for CSC.
It's the other way around.  If you look at `STPPaymentHandler.handleNextAction`, the first thing it does is retrieve the Intent to inspect the status.  But per ^, no need to do this for CSC.

porter-stripe: I introduced `Intent.Confirmable` which is a minimum subset of a `Intent` to confirm an intent.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
porter-stripe: Switched to async APIs!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
```diff
@@ -108,27 +117,35 @@ extension PaymentSheet {
...
+                            configuration: configuration
+                        )
+                    }
 
-                    // Paypal requires mandate_data if setting up
-                    if confirmParams.paymentMethodType.stpPaymentMethodType == .payPal
-                        && paymentIntent.setupFutureUsage == .offSession
-                    {
-                        paymentIntentParams.mandateData = .makeWithInferredValues()
+                    if let paymentIntent = paymentIntent {
```
porter-stripe, line 131: Besides shipping we also read SFU, we could read it from the IntentConfig and attach it to the `Intent.Confirmable`. Let me know your thoughts.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
```diff
@@ -290,6 +317,14 @@ extension PaymentSheet {
 
             switch confirmOption {
             case .wallet:
+                guard let intent = confirmableIntent.asIntent else {
+                    let message =
+                        "Cannot present Link without an Intent"
+                    assertionFailure(message)
+                    completion(.failed(error: PaymentSheetError.unknown(debugDescription: message)))
```
yuki-stripe, line 324: does this mean Link doesn't work right now w/ decoupled?

We could `let intent = confirmableIntent.asIntent ?? await fetchIntent(..)` here for now and leave it as a future optimization to remove the Payment/SetupIntent object dependency from PayWithLinkController.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
```diff
@@ -591,75 +626,87 @@ extension PaymentSheet {
...
-                fatalError("Either confirmHandler or confirmHandlerForServerSideConfirmation must be non-nil when using deferred intents")
-            }
+    static func finishDeferredConfirmation(deferredIntentContext: DeferredIntentContext,
+                                           clientSecret: String) {
+        if deferredIntentContext.isServerSideConfirmation {
+            handleServerConfirmedIntent(deferredContext: deferredIntentContext, clientSecret: clientSecret)
+        } else {
+            confirmClientSide(for: deferredIntentContext, with: clientSecret)
         }
     }
```
yuki-stripe, line 691: I'm not seeing the value in extracting this code into a helper method.  It can only be called from exactly one specific place in our code, right - as step 3 of `handleDeferred..`

We should inline this if so.

yuki-stripe: Same applies to the inner `handleServer..` and `confirmClientSide`.. methods.  Can we inline it all? I don't want to pass control over to other helper methods - it's confusing to follow. 

We can extract work out of the 'master' method into helper methods, but if we do so they should return something and be callable from any point in the code.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
```diff
@@ -73,6 +62,13 @@ extension PaymentSheet {
         switch paymentOption {
         // MARK: - Apple Pay
         case .applePay:
+            guard let intent = confirmableIntent.asIntent else {
+                let message =
+                    "Cannot present Apple Pay without an Intent"
+                assertionFailure(message)
+                completion(.failed(error: PaymentSheetError.unknown(debugDescription: message)))
+                return
+            }
```
yuki-stripe, line 71: Does this mean Apple Pay isn't supported?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
porter-stripe: oof yes and no, it works in complete mode but not flow controller, going to fix this and add tests.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
porter-stripe: See Link comment.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
```diff
@@ -691,25 +654,20 @@ extension PaymentSheet {
...
-                            confirmableIntent: confirmableIntent,
+                            intent: intent,
                             paymentOption: deferredIntentContext.paymentOption,
                             paymentHandler: deferredIntentContext.paymentHandler,
                             completion: deferredIntentContext.completion)
                 }
 
             } catch {
-                deferredIntentContext.completion(.failed(error: error))
+                DispatchQueue.main.async {
```
porter-stripe, line 668: This is all we need to resolve the threading crash

porter-stripe: I've also manually verified that when the merchant invokes the `intentCreationCallback` from the background thread that all tests pass.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
```diff
@@ -582,44 +591,103 @@ extension PaymentSheet {
...
+                                                                        with: deferredIntentContext.authenticationContext,
+                                                                        returnURL: deferredIntentContext.configuration.returnURL,
+                                                                        completion: { status, _, error in
+                            handleStatus(status: status, error: error)
+                        })
+                    }
+                } else {
+                    // Client-side confirmation
+                    let intent = try await deferredIntentContext.configuration.apiClient.retrieveIntent(for: deferredIntentContext.intentConfig,
+                                                                                                        withClientSecret: clientSecret)
```
yuki-stripe, line 658: Let's add a comment indicating that this retrieval isn't strictly necessary and we could remove it in the future as an optimization

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/DeferredIntentContext.swift
```diff
@@ -0,0 +1,36 @@
...
+//  DeferredIntentContext.swift
+//  StripePaymentSheet
+//
+//  Created by Nick Porter on 3/3/23.
+//
+
+import Foundation
+import StripePayments
+
+class DeferredIntentContext {
```
yuki-stripe, line 11: Add a comment explaining why we've made this - it's just a convenience/helper to avoid passing every single one of its properties across the static methods in `PaymentSheet+API.swift`.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
yuki-stripe: Why only call the completion block on the main thread in the error case, and not the success cases?

After the completion block is called here, the control flow will eventually return to view controllers or PaymentSheet.FlowController etc. To be a good citizen to the rest of the codebase, it should always return control on the main thread. 

As a shortcut, you could take advantage of `DeferredIntentContext` to wrap the completion block in a `DispatchQueue.main.async` call in its initializer.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
```diff
@@ -582,44 +591,103 @@ extension PaymentSheet {
...
-                return
-            }
+    private static func handleDeferredIntentConfirmation(deferredIntentContext: DeferredIntentContext,
+                                                         paymentMethod: STPPaymentMethod?,
+                                                         paymentMethodParams: STPPaymentMethodParams?,
+                                                         shouldSavePaymentMethod: Bool) {
+        Task {
+            do {
+                // Create PM if necessary
+                let _paymentMethod = try await createPaymentMethodIfNeeded(apiClient: deferredIntentContext.configuration.apiClient,
```
yuki-stripe, line 615: Why _paymentMethod?

Ah to avoid overwriting `paymentMethod`?  Can you do `let paymentMethod = paymentMethod ?? try await ...`

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/DeferredIntentContext.swift
yuki-stripe, line 11: In terms of code organization, wdyt about making a `PaymentSheet+DeferredAPI.swift` file or something and putting this and all the deferred helper methods in there?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
porter-stripe: I'd really like to do this but putting the complier doesn't like it, that's why I opted to create a helper method. There are some other ways to write it but I prefer this, the suggestions in the SO post are confusing to read IMO.

https://stackoverflow.com/questions/68625044/swift-5-5-async-operation-as-optional-default
CleanShot 2023-03-09 at 22 45 08

yuki-stripe: Ah okay.  Let's just remove the `_`?

porter-stripe: removed

yuki-stripe (approved):

PR #2367, opened 2023-03-08
Enable BDCC for LUXE PMs

eurias-stripe:
Note: making this PR into the previous PR branch is my attempt at creating a PR stack to make reviews easier and avoid being blocked while waiting for review.

## Summary
Enabling billing details collection configuration for LUXE PMs.

## Details
As explained [here](https://docs.google.com/document/d/109d-foiy8KiKljV71gTtvit-phwmAYPOmaGf0UXz2fA/edit#heading=h.nygsesmz42ga), to enable BDCC for LUXE PMs I needed to add the concept of a "placeholder" field, so the client can detect where to insert billing details fields if necessary.

## Motivation
https://docs.google.com/document/d/1eBKVqbdhky_b8ux5f_eKp1HdHxgbV1WdF_XQWp0kdk0/edit#heading=h.dofjidvfxcoc

## Testing
- Added UI and unit tests
- Tested multiple LPMs, specially the most finicky ones, Afterpay and Klarna, because of their billing details requirements.


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory+FormSpec.swift
```diff
@@ -78,8 +153,88 @@ extension PaymentSheetFormFactory {
...
+        case .placeholder(let spec):
+            return makePlaceholder(for: spec)
         case .unknown:
             return nil
         }
     }
+
+    func makePlaceholder(for spec: FormSpec.PlaceholderSpec) -> Element? {
+        let field = spec.field
+        guard field != .unknown else { return nil }
```
sfriedman-stripe, line 165: Should we have any metrics for how often we're getting unknown fields? Feels like we shouldn't just be swallowing this

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory+FormSpec.swift
```diff
@@ -78,8 +153,88 @@ extension PaymentSheetFormFactory {
...
+        }
+    }
+
+    private func connectBillingDetailsFields(
+        countryElement: Element?,
+        addressElement: Element?,
+        phoneElement: Element?
+    ) {
+        // Using a closure because a function would require capturing self, which will be deallocated by the time
+        // the closures below are called.
```
sfriedman-stripe, line 195: I don't follow. If it's capturing self, wouldn't that mean it's not deallocated when the closures are called?

StripeUICore/StripeUICore/Source/Elements/Factories/Address/AddressSectionElement.swift
```diff
@@ -16,12 +16,14 @@ import UIKit
  In addition to the physical address, it can collect other related fields like name.
  */
 @_spi(STP) public class AddressSectionElement: ContainerElement {
+    public typealias DidUpdateAddress = (AddressDetails) -> Void
```
sfriedman-stripe, line 19: Why the typealias? The naming doesn't feel Swifty. Can we just use the closure type directly?

StripeUICore/StripeUICore/Source/Elements/Factories/Address/AddressSectionElement.swift
```diff
@@ -335,7 +350,12 @@ import UIKit
             }
         }
 
-        let initialElements: [Element?] = [name, country, autoCompleteLine]
+        var initialElements = [Element?]()
+        initialElements.append(name)
+        if collectionMode != .noCountry {
+            initialElements.append(country)
+        }
+        initialElements.append(autoCompleteLine)
```
sfriedman-stripe, line 358:
```suggestion
        var initialElements = [name]
        if collectionMode != .noCountry {
            initialElements.append(country)
        }
        initialElements.append(autoCompleteLine)
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory+FormSpec.swift
```diff
@@ -7,38 +7,111 @@
...
 
     private func makeFormElements(from spec: FormSpec) -> [Element] {
+        // These fields may be added according to `configuration.billingDetailsCollectionConfiguration` if they
+        // aren't already present.
+        var billingDetailsFields: [FormSpec.PlaceholderSpec.PlaceholderField] = [
+            .name,
+            .email,
+            .phone,
+            .billingAddress,
+        ]
```
sfriedman-stripe, line 40: Why can't we use a `Set` here?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory+FormSpec.swift
```diff
@@ -7,38 +7,111 @@
...
+                    case .phone:
+                        billingDetailsFields.remove(.phone)
+                        phoneElement = element
+                    case .billingAddress, .billingAddressWithoutCountry:
+                        billingDetailsFields.remove(.billingAddress)
+                        billingAddressElement = element
+                    default: break
+                    }
+                default: break
+                }
```
sfriedman-stripe, line 76: Can we not simplify this? Something like:

```suggestion
        for fieldSpec in spec.fields {
            if let element = fieldSpecToElement(fieldSpec: fieldSpec) {
                if billingDetailsFields.contains(fieldSpec) {
                    billingDetailsFields.remove(fieldSpec)
                }
            
                // Check if this is a billing details field and process it.
                switch fieldSpec {
                case .billing_address:
                    billingAddressElement = element
                case .country, .klarna_country:
                    countryElement = element
                case .placeholder(let placeholder):
                    if billingDetailsFields.contains(fieldSpec) {
                        billingDetailsFields.remove(fieldSpec)
                    }
                    
                    switch placeholder.field {
                    case .phone
                        phoneElement = element
                    case .billingAddress, .billingAddressWithoutCountry:
                        billingAddressElement = element
                    default: break
                    }
                default: break
                }
```

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetBillingCollectionUITests.swift
```diff
@@ -598,4 +598,318 @@ final class PaymentSheetBillingCollectionUITests: XCTestCase {
...
+        XCTAssertTrue(app.textFields["Email"].exists)
+        XCTAssertTrue(app.textFields["Full name"].exists)
+        XCTAssertFalse(app.textFields["Phone"].exists)
+        XCTAssertTrue(app.staticTexts["Billing address"].exists)
+        XCTAssertEqual(app.textFields["Country or region"].value as? String, "United States")
+        XCTAssertEqual(app.textFields["Address line 1"].value as? String, "510 Townsend St.")
+        XCTAssertEqual(app.textFields["Address line 2"].value as? String, "")
+        XCTAssertEqual(app.textFields["City"].value as? String, "San Francisco")
+        XCTAssertEqual(app.textFields["State"].value as? String, "California")
+        XCTAssertEqual(app.textFields["ZIP"].value as? String, "94102")
```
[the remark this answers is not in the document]
eurias-stripe: They are slightly different in different cases (capitalization, Name/Full name, Country/Country or Region) so using helpers wouldn't totally solve the problem.
Also the DAMP principle is usually preferred over DRY for test code

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/FormSpec/FormSpec.swift
```diff
@@ -229,6 +234,27 @@ extension FormSpec {
...
+        enum CodingKeys: String, CodingKey {
+            case field = "for"
+        }
+
+        enum PlaceholderField: String, Decodable, Equatable {
+            case name
+            case email
+            case phone
+            case billingAddress = "billing_address"
+            case billingAddressWithoutCountry = "billing_address_without_country"
```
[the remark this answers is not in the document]
eurias-stripe: It is [already set](https://github.com/stripe/stripe-ios/blob/3edcd2f3/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/FormSpec/FormSpecProvider.swift#L31) to `.convertFromSnakeCase` but it isn't working, I assume because of the custom `init(from decoder:)` implementation.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory+FormSpec.swift
```diff
@@ -7,38 +7,111 @@
...
-    func makeFormElementFromSpec(spec: FormSpec) -> FormElement {
+    func makeFormElementFromSpec(
+        spec: FormSpec,
+        additionalElements: [Element] = []
+    ) -> PaymentMethodElementWrapper<FormElement> {
         let elements = makeFormElements(from: spec)
-        return FormElement(autoSectioningElements: elements, theme: theme)
+        let formElement = FormElement(
+            autoSectioningElements: elements + additionalElements,
+            theme: theme)
```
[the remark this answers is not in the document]
eurias-stripe: SwiftLint doesn't correct this and it isn't a violation. SwiftFormat does enforce the latter style but we are not using it. I'm ambivalent on which style to use and I go with what feels better depending on the case, in this case for example, since it is only two lines I feel like keeping the parenthesis in the same line looks better.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory+FormSpec.swift
```diff
@@ -7,38 +7,111 @@
...
+
         return elements
     }
 
     private func fieldSpecToElement(fieldSpec: FormSpec.FieldSpec) -> Element? {
         switch fieldSpec {
         case .name(let spec):
-            return makeName(label: spec.translationId?.localizedValue, apiPath: spec.apiPath?["v1"])
+            return configuration.billingDetailsCollectionConfiguration.name != .never
+                ? makeName(label: spec.translationId?.localizedValue, apiPath: spec.apiPath?["v1"])
```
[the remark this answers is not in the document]
eurias-stripe: It probably should be, but this is pre-existing code and would be out of scope to change in this PR.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory+FormSpec.swift
eurias-stripe: I plan on adding analytics in another PR, I can include this if it makes sense

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory+FormSpec.swift
eurias-stripe: `self` should be captured weakly to avoid a retain cycle, and since it is not retained strongly it gets deallocated.
That's how I implemented it first and it wasn't working until I realized that was the issue, which is why I added the comment.

StripeUICore/StripeUICore/Source/Elements/Factories/Address/AddressSectionElement.swift
eurias-stripe: It's following the naming convention from [here](https://github.com/stripe/stripe-ios/blob/3edcd2f3/StripeUICore/StripeUICore/Source/Elements/DropdownFieldElement.swift#L20), and I'm using a typealias to follow the same convention.

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetBillingCollectionUITests.swift
```diff
@@ -598,4 +598,318 @@ final class PaymentSheetBillingCollectionUITests: XCTestCase {
...
+        XCTAssertTrue(saveAddressButton.isEnabled)
+        saveAddressButton.tap()
+
+        app.buttons["Checkout (Complete)"].tap()
+
+        guard let cell = scroll(collectionView: app.collectionViews.firstMatch, toFindCellWithId: "afterpay_clearpay")
+        else {
+            XCTFail()
+            return
+        }
```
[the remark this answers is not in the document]
eurias-stripe: I always forget that `XCTUnwrap` exists. I'm changing the instances in this PR but not changing any pre-existing ones.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory+FormSpec.swift
eurias-stripe: Order matters.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory+FormSpec.swift
eurias-stripe: Not really
1. The array contains `PlaceholderSpec`s, not `FieldSpec`s, so they cannot be compared directly.
2. I need to check whether the FieldSpec exists either as a FieldSpec or a PlaceholderSpec, which makes it check the array twice.
3. I still need the switch to assign the elements accordingly, so the change would only remove two lines and IMO it is easier to follow this way.

StripeUICore/StripeUICore/Source/Elements/Factories/Address/AddressSectionElement.swift
eurias-stripe: I was gonna do it this way, but since I needed to specify the type explicitly and would need to append anyway I went the other way. Changed.

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetBillingCollectionUITests.swift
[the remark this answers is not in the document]
sfriedman-stripe: Just feels really cumbersome if we ever need to change the field names in our UI, which is what all of these tests are inherently coupled to. I'm not suggesting we move the "510 Townsend St"-type strings into constants, which would definitely violate DAMP. Up to you

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory+FormSpec.swift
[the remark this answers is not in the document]
sfriedman-stripe: I'm pretty sure it's supposed to though? Look at the 2nd Triggering Example [here](https://realm.github.io/SwiftLint/multiline_arguments_brackets.html)

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory+FormSpec.swift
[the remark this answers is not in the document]
sfriedman-stripe: Fair, fine with this not being part of this PR, but what's the process for improving it then? In my experience, if these minor things don't get improved along the way by the people touching the existing code, they don't get improved at all. Am a general proponent of the Boy Scout Rule here

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory+FormSpec.swift
sfriedman-stripe: For my benefit, where's the retain cycle? Is the `PaymentSheetFormFactory` holding a strong reference to the element somewhere?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory+FormSpec.swift
sfriedman-stripe: I don't follow points (2) and (3) since the example code I wrote is still doing both of those things, but point (1) makes sense, thanks for clarifying

sfriedman-stripe: I don't follow point (3) since the example code I wrote is still doing that, but points (1) and (2) make sense, thanks for clarifying

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory+FormSpec.swift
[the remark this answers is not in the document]
eurias-stripe: I agree with making changes along the way, but this change is not completely localized, it's all over in here and in tests, I found 58 instances on a quick search.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory+FormSpec.swift
eurias-stripe: No, the closures on line 213 and 231 would need to capture `self` in order to call a function on it, that way the closure would hold a reference to `self` and `self` holds a reference to the closure, it's the most classic example of a retain cycle, so you normally capture `self` weakly to avoid it, but in this case the factory will be deallocated before the closures are called.
I also thought about capturing a reference to the function, but IMO it looks clearer if you use an inline closure an capture that, which is basically the same effect.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory+FormSpec.swift
eurias-stripe: In point 3 I'm referring to still needing that switch, so in the end you are only removing 4 total cases, in LOCs the total difference is 3 lines, and IMO it is easier to follow doing what you need to do in each case, than removing the specs in two separate contains/remove than doing it on each case when needed.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory+FormSpec.swift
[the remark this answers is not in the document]
eurias-stripe: hmm, you're right, but we don't have that rule enabled and I hadn't seen it, I'll enable it and run an autofix on the repo in a separate PR.

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetBillingCollectionUITests.swift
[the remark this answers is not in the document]
eurias-stripe: Ok, I reconsidered and added helpers for the fields that repeat the most, the one-offs don't have helpers

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory+FormSpec.swift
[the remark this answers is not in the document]
sfriedman-stripe: Ack. Adding this to the run queue

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetBillingCollectionUITests.swift
```diff
@@ -20,6 +20,24 @@ final class PaymentSheetBillingCollectionUITests: XCTestCase {
...
+    var countryField: XCUIElement { app.textFields["Country or region"] }
+    var line1Field: XCUIElement { app.textFields["Address line 1"] }
+    var line2Field: XCUIElement { app.textFields["Address line 2"] }
+    var cityField: XCUIElement { app.textFields["City"] }
+    var stateField: XCUIElement { app.textFields["State"] }
+    var zipField: XCUIElement { app.textFields["ZIP"] }
+    var checkoutButton: XCUIElement { app.buttons["Checkout (Complete)"] }
+    var payButton: XCUIElement { app.buttons["Pay $50.99"] }
+    var successText: XCUIElement { app.alerts.staticTexts["Success!"] }
+    var okButton: XCUIElement { app.alerts.scrollViews.otherElements.buttons["OK"] }
```
sfriedman-stripe, line 39: Shouldn't these all be private?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory+FormSpec.swift
sfriedman-stripe: > IMO it looks clearer if you use an inline closure an capture that, which is basically the same effect.

I agree that the closure looks cleaner either way, so just having a theoretical conversation at this point to improve my understanding of our code, but: `self` _doesn't_ hold a reference to the closure since it's being set on a property of the element, and I don't see where `self` has a strong reference to the element. So I'm not understanding where the retain cycle would come from here?

StripeUICore/StripeUICore/Source/Elements/Factories/Address/AddressSectionElement.swift
sfriedman-stripe: I was originally thinking this was public API, but just noticed it's `@_spi`, so I'm less concerned here. Still, we should probably look at improving this at some point. The naming convention is very Java, and I'm not convinced the typealiases add value since we're not referencing them in a bunch of places

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory+FormSpec.swift
eurias-stripe: Here is the memory leak 
Screen Shot 2023-03-11 at 11 56 52 AM

It is an indirect cycle in this case, or maybe not a cycle but it causes a memory leak since the factory would be retained far longer than it should.

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetBillingCollectionUITests.swift
eurias-stripe: These are tests, so access level doesn't really matter, but yeah, they can be private. Changed.

Stripe/StripeiOSTests/FormSpecProviderTest.swift
```diff
@@ -384,10 +384,10 @@ class FormSpecProviderTest: XCTestCase {
 
         // Validate that we were not able to override the spec read in from disk
         let epsUpdated = try XCTUnwrap(sut.formSpec(for: paymentMethodType))
-        XCTAssertEqual(epsUpdated.fields.count, 2)
+        XCTAssertEqual(eps.fields.count, 5)
         XCTAssertEqual(
-            epsUpdated.fields.first,
-            .name(FormSpec.NameFieldSpec(apiPath: nil, translationId: nil))
+            eps.fields.first,
+            .name(FormSpec.NameFieldSpec(apiPath: ["v1": "billing_details[name]"], translationId: nil))
```
wooj-stripe, line 390: These need to reference `epsUpdated`, not the original `eps` variable.

eurias-stripe: Oops, copy-pasta

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory+FormSpec.swift
```diff
@@ -7,38 +7,111 @@
...
             return nil
         }
         return provider.formSpec(for: paymentMethodType)
     }
 
-    func makeFormElementFromSpec(spec: FormSpec) -> FormElement {
+    func makeFormElementFromSpec(
+        spec: FormSpec,
+        additionalElements: [Element] = []
+    ) -> PaymentMethodElementWrapper<FormElement> {
```
wooj-stripe, line 24: Hm.. it's not immediately obvious to me why we are needing to wrap the FormElement in a PaymentMethodElementWrapper. I'm guessing it's an elegant solution to a problem, but not seeing it -- Could you please explain why we need to do so?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
```diff
@@ -100,17 +102,17 @@ class PaymentSheetFormFactory {
...
-            return FormElement(autoSectioningElements: [makePaypalMandate(intent: intent)], theme: theme)
+            additionalElements = [makePaypalMandate(intent: intent)]
         }
 
         // 2. Element-based forms defined in JSON
         guard let spec = specFromJSONProvider() else {
             fatalError()
         }
-        return makeFormElementFromSpec(spec: spec)
+        return makeFormElementFromSpec(spec: spec, additionalElements: additionalElements)
```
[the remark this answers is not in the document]
eurias-stripe: Maybe we could use the placeholder field to hold an optional mandate for these cases

sfriedman-stripe (dismissed):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory+FormSpec.swift
```diff
@@ -7,38 +7,111 @@
...
-
     func specFromJSONProvider(provider: FormSpecProvider = FormSpecProvider.shared) -> FormSpec? {
         guard let paymentMethodType = PaymentSheet.PaymentMethodType.string(from: paymentMethod) else {
             return nil
         }
         return provider.formSpec(for: paymentMethodType)
     }
 
-    func makeFormElementFromSpec(spec: FormSpec) -> FormElement {
+    func makeFormElementFromSpec(
```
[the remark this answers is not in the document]
eurias-stripe: Maybe, but we don't use "wrapped" for all the other functions that return wrapped elements, I don't know if it's really necessary. I can go either way though

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory+FormSpec.swift
```diff
@@ -7,38 +7,111 @@
...
+    ) -> PaymentMethodElementWrapper<FormElement> {
         let elements = makeFormElements(from: spec)
-        return FormElement(autoSectioningElements: elements, theme: theme)
+        let formElement = FormElement(
+            autoSectioningElements: elements + additionalElements,
+            theme: theme)
+        return makeDefaultsApplierWrapper(for: formElement)
     }
 
     private func makeFormElements(from spec: FormSpec) -> [Element] {
```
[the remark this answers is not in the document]
eurias-stripe: I wanted to do that, the problem is that if I want to just split out the `switch` I would either need to return a tuple with the billing details fields, where only one of them would be set or use out parameters, with the same issue.
I could do a pass on the elements array after generating them, but then I would need to check the type of the element instead of using the enum, which is possible but I try to avoid that if possible since it van be flimsy, i.e. if we change the element type, or if add a wrapper or something, it would be very easy to miss when refactoring, so I prefer switching on the enum.
In summary, I didn't find a good clean way of splitting this out, so I decided to leave it in the same function for that reason.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/FormSpec/FormSpec.swift
```diff
@@ -39,6 +39,9 @@ struct FormSpec: Decodable {
 
         case iban(BaseFieldSpec)
         case sepa_mandate
+
+        case placeholder(PlaceholderSpec)
+
```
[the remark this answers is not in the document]
eurias-stripe: I considered doing something like that, originally that's what I wanted to do with the field type directly, but there are some aspects of it that I don't like.
1. I don't like deserializations that don't map fields one to one, you need to know about the serialization code in order to understand what the json file will look like in code.
2. The `FieldSpec` enum would grow a lot by doing that.

I would be okay with it if we have a compelling reason for it, but as it is right now I don't think it's necessary.

wooj-stripe (approved):

PR #2370, opened 2023-03-09
Connect billing details elements for non LUXE PMs

eurias-stripe:
## Summary
Connecting billing details elements for non LUXE PMs so the phone country can change when the billing country changes.
I previously created this helper for LUXE and now I'm using it for all PMs.

## Motivation
Part of BDCC

## Testing
Checked that phone number country changed when country changed only when the field didn't have a default value and had not been modified.


Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
```diff
@@ -803,27 +803,28 @@ extension PaymentSheetUITest {
         if addBillingDetailsButton.waitForExistence(timeout: 4.0) {
             addBillingDetailsButton.tap()
 
-            let firstNameCell = applePay.tables.cells["First Name"]
+            print("Apple pay: \(applePay.debugDescription)")
```
wooj-stripe, line 806: remove print

wooj-stripe (approved):

PR #2373, opened 2023-03-10
Add API for Card Brand Choice

davidme-stripe:
## Summary
Add a `possibleBrands(forCard:completion:)` API, enabling users to detect the available brands for a card. A user could then offer the customer a choice between available card networks for a card, e.g. Cartes Bancaires or Visa.
Add some images and strings for CB as well.

## Motivation
User request.

## Testing
Added tests.

## Changelog
Done.


StripePayments/StripePayments/Source/Helpers/STPCardValidator.swift
```diff
@@ -414,6 +414,43 @@ public class STPCardValidator: NSObject {
...
+    static let cbcBinController = STPBINController()
+
+    /// Returns available brands for the provided card details.
+    /// - Parameter card: The card details to validate.
+    /// - Parameter completion: Will be called with the set of available STPCardBrands or an error.
+    /// - seealso: https://stripe.com/docs/card-brand-choice
+    public class func possibleBrands(forCard cardParams: STPPaymentMethodCardParams,
+                                     completion: @escaping (Result<Set<STPCardBrand>, Error>) -> Void) {
+        guard let cardNumber = cardParams.number else {
+            completion(.success([]))
```
sfriedman-stripe, line 433: Is it a success if no number is passed in?

StripePayments/StripePayments/Source/Helpers/STPCardValidator.swift
```diff
@@ -414,6 +414,43 @@ public class STPCardValidator: NSObject {
...
+    class func possibleBrands(forNumber cardNumber: String,
+                              completion: @escaping (Result<Set<STPCardBrand>, Error>) -> Void) {
+        cbcBinController.retrieveBINRanges(forPrefix: cardNumber, recordErrorsAsSuccess: false, onlyFetchForVariableLengthBINs: false) { result in
+            switch result {
+            case .failure(let error):
+                completion(.failure(error))
+            case .success:
+                let binRanges = cbcBinController.binRanges(forNumber: cardNumber)
+                var brands = binRanges.map { $0.brand }
+                brands.removeAll { $0 == .unknown }
```
sfriedman-stripe, line 448: This is maybe just stylistic so take it or leave it, but could also just chain this into a single let instead of mutating anything, i.e.

```suggestion
                let brands = binRanges.map { $0.brand }
                                      .filter { $0 != .unknown }
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/RotatingCardBrandsView.swift
```diff
@@ -50,6 +50,8 @@ class RotatingCardBrandsView: UIView {
                     return 5
                 case .unionPay:
                     return 6
+                case .cartesBancaires:
+                    return nil
```
sfriedman-stripe, line 54: Why nil?

davidme-stripe: We don't want CB to show up in the RotatingCardBrandsView, matching our behavior on the web. I'll add a comment though, I had to dig a bit to figure this out.

StripePayments/StripePayments/Source/Helpers/STPCardValidator.swift
davidme-stripe: Good question: If you pass an empty string, the answer is `.allCases`, as any card brand is possible. A nil string should probably match that behavior. The existing `brandForNumber` API returns `.unknown` if multiple brands match.

It'd be messy ergonomically to make this an error: The user isn't creating the `STPPaymentMethodCardParams` on their own, they're receiving it from one of our UI fields, so an empty card field would return an empty/nil string. The user would then need to add a special case to access `cardParams.number` (which we discourage) to check for a nil or empty string before calling `possibleBrands()`, or they would need to handle "empty card number" errors for any empty card field, which they'd probably want to treat separately from network errors.

porter-stripe (approved):

porter-stripe (approved):

StripePayments/StripePayments/Source/Helpers/STPCardValidator.swift
sfriedman-stripe: I feel like a comment here is probably worth including

davidme-stripe: I added one before merge but lmk if you think it's unclear, I can always go back and add more context: https://github.com/stripe/stripe-ios/pull/2373/files#diff-974d2c2ad6918ec7bf57dd1c8bc19f370013acf59c601b8b7c6941be2fd5cfbeR433

sfriedman-stripe: Oh nice! Didn't realize you added already

PR #2376, opened 2023-03-13
Increase timeout for success text in deferred UI tests

porter-stripe:
## Summary
- Tests were failing on `XCTAssertTrue(successText.waitForExistence(timeout: 10.0))` on CI and I was able to see it fail once due to the time out not being long enough

## Motivation
Fix CI nightly builds

## Testing
Manual

## Changelog
N/A

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
+++ b/Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
@@ -139,7 +139,7 @@ class PaymentSheetUITest: XCTestCase {
         app.buttons["Continue"].tap()
         app.buttons["Checkout (Custom)"].tap()
         var successText = app.alerts.staticTexts["Success!"]
-        XCTAssertTrue(successText.waitForExistence(timeout: 20.0))
+        XCTAssertTrue(successText.waitForExistence(timeout: 15.0))
         app.alerts.scrollViews.otherElements.buttons["OK"].tap()
 
         // Reload w/ same customer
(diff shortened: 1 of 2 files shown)


Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
```diff
@@ -139,7 +139,7 @@ class PaymentSheetUITest: XCTestCase {
         app.buttons["Continue"].tap()
         app.buttons["Checkout (Custom)"].tap()
         var successText = app.alerts.staticTexts["Success!"]
-        XCTAssertTrue(successText.waitForExistence(timeout: 10.0))
+        XCTAssertTrue(successText.waitForExistence(timeout: 20.0))
```
yuki-stripe, line 142: This test didn't fail, do we need to increase the timeout?

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
```diff
@@ -628,7 +628,7 @@ extension PaymentSheetUITest {
         app.buttons["Pay $50.99"].tap()
 
         let successText = app.staticTexts["Success!"]
-        XCTAssertTrue(successText.waitForExistence(timeout: 10.0))
+        XCTAssertTrue(successText.waitForExistence(timeout: 20.0))
```
yuki-stripe, line 631: Can we remove the +2 second delay from https://github.com/stripe/stripe-ios/pull/2364/files#diff-2e5b1fdaf1439a692709b9d984e5475e27a0914bb98d482d217772eacb4c8eabR699

Why are these taking > 10 seconds?

PR #2382, opened 2023-03-13
Add BDCC to analytics payload

eurias-stripe:
## Summary
Add analytics for BDCC.
This will add fields to the `payment_sheet_configuration` column.

## Motivation
Part of BDCC

## Testing
Checked splunk for the event


sfriedman-stripe (commented):
How does this manifest, as another column in the analytics table or as part of a JSON payload? I think @yuki-stripe was asking @randyshort-stripe about this in the metrics channel. Would be good to have details in the PR description about how you're approaching this

eurias-stripe:
> How does this manifest, as another column in the analytics table or as part of a JSON payload?

It will appear as part of the JSON payload in the `payment_sheet_configuration` column ([example query](https://hubble.corp.stripe.com/queries/eurias/f438fd5f))

porter-stripe (approved):
Adding it to the payment sheet configuration makes sense here to me.

sfriedman-stripe:
> > How does this manifest, as another column in the analytics table or as part of a JSON payload?
> 
> It will appear as part of the JSON payload in the `payment_sheet_configuration` column ([example query](https://hubble.corp.stripe.com/queries/eurias/f438fd5f)).
> 
> I don't think it would make sense to add another column for this, it should just be logged along with the rest of the configuration.

Chatted about this with @randyshort-stripe in Slack and agreed to just add it to the JSON config

sfriedman-stripe (approved):

PR #2384, opened 2023-03-14
Invoke merchant callback on main thread

porter-stripe:
## Summary
- Our SSC tests were failing on CI due to UI modifications happening in the merchant app (playground app) occurring off the main thread.

```
Main Thread Checker: UI API called on a background thread: -[UISegmentedControl selectedSegmentIndex]
PID: 7251, TID: 52484, Thread name: (none), Queue name: com.apple.root.user-initiated-qos.cooperative, QoS: 25
Backtrace:
4   PaymentSheetExample                 0x000000010433e2e4 $s19PaymentSheetExample0aB14TestPlaygroundC19merchantCountryCodeAC08MerchantgH0Ovg + 244
5   PaymentSheetExample                 0x00000001043412d4 $s19PaymentSheetExample0aB14TestPlaygroundC39confirmHandlerForServerSideConfirmationyySS_Sbys6ResultOySSs5Error_pGctF + 952
6   PaymentSheetExample                 0x0000000104340f0c $s19PaymentSheetExample0aB14TestPlaygroundC12intentConfig06StripeaB00aB0C19IntentConfigurationVvgySS_Sbys6ResultOySSs5Error_pGctcACcfu5_ySS_SbyAMctcfu6_ + 100
7   StripePaymentSheet                  0x0000000106501140 $s18StripePaymentSheet0bC0C35fetchIntentClientSecretFromMerchant12intentConfig15paymentMethodID010shouldSavebM0SSAC0E13ConfigurationV_SSSbtYaKFZyScCySSs5Error_pGXEfU_ + 460
8   libswift_Concurrency.dylib          0x00000001af04d67c $ss31withCheckedThrowingContinuation8function_xSS_yScCyxs5Error_pGXEtYaKlFySccyxsAC_pGXEfU_TA + 156
9   libswift_Concurrency.dylib          0x00000001af04d714 $ss30withUnsafeThrowingContinuationyxySccyxs5Error_pGXEYaKlF + 120
10  libswift_Concurrency.dylib          0x00000001af0754f8 _ZN5swift34runJobInEstablishedExecutorContextEPNS_3JobE + 336
11  libswift_Concurrency.dylib          0x00000001af076170 _ZL17swift_job_runImplPN5swift3JobENS_11ExecutorRefE + 80
12  libdispatch.dylib                   0x0000000180138e8c _dispatch_continuation_pop + 152
13  libdispatch.dylib                   0x00000001801382e4 _dispatch_async_redirect_invoke + 872
14  libdispatch.dylib                   0x0000000180148a7c _dispatch_root_queue_drain + 440
15  libdispatch.dylib                   0x00000001801495b8 _dispatch_worker_thread2 + 224
16  libsystem_pthread.dylib             0x00000001af2568c0 _pthread_wqthread + 224
17  libsystem_pthread.dylib             0x00000001af2556c0 start_wqthread + 8
```

These were also manifesting as purple memory errors in Xcode but they were not causing failures:
CleanShot 2023-03-14 at 05 42 19@2x

## Motivation
Fix tests

## Testing

https://app.bitrise.io/build/e5b00c14-b957-4a5a-8e04-9b4f1c2c6970
https://app.bitrise.io/build/cad1f580-0b1c-4cf9-a95b-a8b218810114

## Changelog
N/A

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
@@ -695,10 +695,7 @@ extension PaymentSheetTestPlayground {
     // Client-side confirmation handler
     func confirmHandler(_ paymentMethodID: String,
                         _ intentCreationCallback: @escaping (Result<String, Error>) -> Void) {
-        // Client-side confirmation, simulate a delay like we are doing validation on our backend
-        DispatchQueue.main.asyncAfter(deadline: .now() + 2) {
-            intentCreationCallback(.success(self.clientSecret!))
-        }
+        intentCreationCallback(.success(self.clientSecret!))
     }
 
     // Server-side confirmation handler
--- a/Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
@@ -695,7 +695,9 @@ extension PaymentSheetTestPlayground {
     // Client-side confirmation handler
     func confirmHandler(_ paymentMethodID: String,
                         _ intentCreationCallback: @escaping (Result<String, Error>) -> Void) {
-        intentCreationCallback(.success(self.clientSecret!))
+        DispatchQueue.global(qos: .background).async {
+            intentCreationCallback(.success(self.clientSecret!))
+        }
     }
 
     // Server-side confirmation handler
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+DeferredAPI.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+DeferredAPI.swift
@@ -97,13 +97,17 @@ extension PaymentSheet {
       try await withCheckedThrowingContinuation { continuation in
 
           if let confirmHandlerForServerSideConfirmation = intentConfig.confirmHandlerForServerSideConfirmation {
-              confirmHandlerForServerSideConfirmation(paymentMethodID, shouldSavePaymentMethod, { result in
-                  continuation.resume(with: result)
-              })
+              DispatchQueue.main.async {
+                  confirmHandlerForServerSideConfirmation(paymentMethodID, shouldSavePaymentMethod, { result in
+                      continuation.resume(with: result)
+                  })
+              }
           } else if let confirmHandler = intentConfig.confirmHandler {
-              confirmHandler(paymentMethodID, { result in
-                  continuation.resume(with: result)
-              })
+              DispatchQueue.main.async {
+                  confirmHandler(paymentMethodID, { result in
+                      continuation.resume(with: result)
+                  })
+              }
           }
       }
     }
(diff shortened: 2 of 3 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+DeferredAPI.swift
```diff
@@ -97,13 +97,17 @@ extension PaymentSheet {
...
+              DispatchQueue.main.async {
+                  confirmHandlerForServerSideConfirmation(paymentMethodID, shouldSavePaymentMethod, { result in
+                      continuation.resume(with: result)
+                  })
+              }
           } else if let confirmHandler = intentConfig.confirmHandler {
-              confirmHandler(paymentMethodID, { result in
-                  continuation.resume(with: result)
-              })
+              DispatchQueue.main.async {
```
porter-stripe, line 106: In the CSC case we didn't see these errors since we were already running on the main thread and also not interacting with any UI in the "merchant app".

Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
```diff
@@ -695,10 +695,7 @@ extension PaymentSheetTestPlayground {
     // Client-side confirmation handler
     func confirmHandler(_ paymentMethodID: String,
                         _ intentCreationCallback: @escaping (Result<String, Error>) -> Void) {
-        // Client-side confirmation, simulate a delay like we are doing validation on our backend
-        DispatchQueue.main.asyncAfter(deadline: .now() + 2) {
-            intentCreationCallback(.success(self.clientSecret!))
-        }
+        intentCreationCallback(.success(self.clientSecret!))
```
yuki-stripe, line 698: nit: I feel like there's a significant difference between calling the callback synchronously vs asynchronously, so we should keep the asyncAfter.  Maybe even put it off the main thread to make it more realistic?  Just don't think waiting 2 seconds buys us anything vs the built-in async delay.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+DeferredAPI.swift
yuki-stripe: Huh. This is a good change, we should have been calling merchant callbacks on the main thread, but why didn't it fail locally?

porter-stripe: I wonder if it has to do with a setting we are enabling for tests on CI that cause these warnings to be treated like errors.

yuki-stripe: Link to that setting?

eurias-stripe: GCC_TREAT_WARNINGS_AS_ERRORS is set to YES on the config file, it should be the same locally and in CI

porter-stripe: I don't know the setting (or if there is one). But I've seen Xcode crash or not crash on these types of memory warnings before. It's not clear to me why it crashes vs when it doesn't.

eurias-stripe (approved):

Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
porter-stripe: Changed this to call on the background thread

PR #2385, opened 2023-03-14
BDCC bug fixes

eurias-stripe:
## Summary
Fixed:
- Phone number wasn't validated, allowing to submit with an invalid number.
- Changing phone country when selecting a different country in address.

## Motivation
Bug fixes

## Testing
- Verified that pay button is disabled when phone number is invalid
- Verified that phone country changes as expected

Simulator Screen Recording - iPhone 12 mini - 2023-03-14 at 12 36 52

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
@@ -153,6 +153,7 @@ extension PaymentSheetFormFactory {
             defaultPhoneNumber: configuration.defaultBillingDetails.phone,
             theme: theme)
         return PaymentMethodElementWrapper(element) { phoneField, params in
+            guard case .valid = phoneField.validationState else { return nil }
             params.paymentMethodParams.nonnil_billingDetails.phone = phoneField.phoneNumber?.string(as: .e164)
             return params
         }
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
@@ -503,7 +503,7 @@ extension PaymentSheetFormFactory {
                 return
             }
 
-            phoneElement.selectedCountryCode = countryCode
+            phoneElement.setSelectedCountryCode(countryCode, shouldUpdateDefaultNumber: true)
         }
 
         if let countryElement = countryElement {
--- a/StripeUICore/StripeUICore/Source/Elements/PhoneNumber/PhoneNumberElement.swift
+++ b/StripeUICore/StripeUICore/Source/Elements/PhoneNumber/PhoneNumberElement.swift
@@ -94,6 +94,13 @@ import UIKit
         self.textFieldElement.delegate = self
     }
 
+    public func setSelectedCountryCode(_ countryCode: String, shouldUpdateDefaultNumber: Bool = false) {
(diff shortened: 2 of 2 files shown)


StripeUICore/StripeUICore/Source/Elements/PhoneNumber/PhoneNumberElement.swift
```diff
@@ -94,6 +94,13 @@ import UIKit
         self.textFieldElement.delegate = self
     }
 
+    public func setSelectedCountryCode(_ countryCode: String, shouldUpdateDefaultNumber: Bool = false) {
```
sfriedman-stripe, line 97: Why is the default value for `shouldUpdateDefaultNumber` both here and in `selectCountry(index:shouldUpdateDefaultNumber:)` false when it's only ever called with true except for in one test?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
```diff
@@ -502,7 +503,7 @@ extension PaymentSheetFormFactory {
                 return
             }
 
-            phoneElement.selectedCountryCode = countryCode
+            phoneElement.setSelectedCountryCode(countryCode, shouldUpdateDefaultNumber: true)
```
sfriedman-stripe, line 506: I'm not seeing anywhere else we set this property, so shouldn't we clean up the custom setter now that we're using a function instead?

StripeUICore/StripeUICore/Source/Elements/PhoneNumber/PhoneNumberElement.swift
eurias-stripe: I don't know why the original was done that way, but IMO it makes semantical sense for the default to be false, if you call `setSelectedCountryCode("US")`, I wouldn't expect it to have additional side effects, changing the default number in this case, so it makes sense that you have to explicitly signal that you want the side effect.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
eurias-stripe: Whenever I create a function to set a property with additional parameters, I like having the setter to perform the default flow so it can still be assigned using `=`. That said, since this is an internal api (`@_spi` protection for the class) and we are not using it, it makes sense to remove it, if it was a public API I would keep it.

sfriedman-stripe: I kinda question whether we should have so much logic in a setter though? A function flags more to the reader that more could be happening under the hood, whereas if I'm setting something, I'm not usually expecting it to have any custom validation (and therefore failure modes)

StripeUICore/StripeUICore/Source/Elements/PhoneNumber/PhoneNumberElement.swift
sfriedman-stripe: I guess my question is more: Why have a default value at all if we're not actually using it?

eurias-stripe: I don't think there is a compelling reason to have it or not have it, but in general I prefer to add defaults when there is a sensible one, even if is not used at the moment it could be useful later, and it signals what's the "default" way to use an API.

sfriedman-stripe: If we can think of a scenario where we'd want to update the country code without changing the default number in the future, and that that would be the common/default way to call the API, I'm fine keeping it. It just doesn't seem like we have a use case for that, which implies just making it a regular parameter without a default

sfriedman-stripe (approved):

PR #2387, opened 2023-03-15
update readme for Identity

ccen-stripe:
## Summary

Remove the beta tag and update the min version of the SDK

## Motivation

Update doc for GA


StripeIdentity/README.md
```diff
@@ -32,7 +32,7 @@ The Stripe Identity iOS SDK makes it quick and easy to verify your user's identi
 
 ## Requirements
 
-The Stripe Identity iOS SDK is compatible with apps targeting iOS 14.3 or above.
```
mludowise-stripe, line 35: I think you can remove this line all together since the root README says that the iOS SDK is compatible with 13.0 and higher, which should apply to all the modules in the repo. It means it's one less thing to keep in sync if we end up bumping the version again.

But up to you :-)

mludowise-stripe (approved):

PR #2391, opened 2023-03-15
Fix COP formatting

eurias-stripe:
## Summary
The COP currency is a weird case, it is officially a 2 decimal currency and that's how Stripe treats it, but in practice the decimal units have been discontinued since the 80's, as a result some formatters treat it as a 0 decimal currency, like swift's `NumberFormatter`, this causes a discrepancy where an amount of, for example, 10,000 COP, gets formatted as 1,000,000, because the Stripe amount has two trailing zeroes representing the cents.
To fix this, I added a "special cases" list with the number of decimals used by Stripe, right now that list only includes COP.

## Motivation
https://jira.corp.stripe.com/browse/RUN_MOBILESDK-2116

## Testing
Verified visually and added unit test

## Changelog
[Fixed] Fixed amounts in COP being formatted incorrectly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## x.x.x yyyy-mm-dd
+## Payments
+* [Fixed] Fixed amounts in COP being formatted incorrectly.
+
 ## 23.5.0 2023-03-13
 ### Payments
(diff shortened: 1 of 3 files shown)


yuki-stripe (approved):
Make sure Android makes this change too

eurias-stripe:
> Make sure Android makes this change too

I checked Android, it is already handled correctly

PR #2392, opened 2023-03-15
Generate inferred mandate data for paypal

eurias-stripe:
## Summary
Mandate data was not being added to setup intents, which requires it, adding it to the list of PMs that should infer mandate data from the client.

## Motivation
Issue reported on Slack.

## Testing
Added to unit test.

## Changelog
[Fixed] Including inferred mandate data for Paypal.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## x.x.x yyyy-mm-dd
+### Payments
+* [Fixed] Including inferred mandate data for Paypal.
+
 ## 23.5.0 2023-03-13
 ### Payments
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## x.x.x yyyy-mm-dd
+### Payments
+* [Fixed] Including inferred mandate data for Paypal.
```
yuki-stripe, line 3: I don't think it's a fix so much as a new feature - enabling SI and PI+SFU support for Paypal in the API bindings.

yuki-stripe: Or actually this PR is just adding SI support, maybe - but we should test the flow to make sure.

eurias-stripe: @charliecruzan-stripe you are currently trying to do this, correct? And making this change made it work for you?

Issue #2398, opened 2023-03-20
[Feature] `UIDevice.currentDevice.name` access isn't allowed on iOS 16 from third party SDKs, but Stripe is using it.


amayers:
## Is your feature request related to a problem? Please describe.
The Stripe SDK [accesses](https://github.com/stripe/stripe-ios/blob/94583773331718db238b78f261ef2e5822e46450/Stripe3DS2/Stripe3DS2/STDSDeviceInformationParameter.m#L227) `UIDevice.currentDevice.name`. [As of iOS 16](https://developer.apple.com/documentation/uikit/uidevice/1620015-name) that device name is no longer the user set device name. Instead it's a generic name like `iPhone` so to not leak PII. Apps can request access to an [entitlement](https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_developer_device-information_user-assigned-device-name) to gain access to the iOS <= 15 behavior. However this comes with some conditions. The relevant one for this discussion is: 
```
Your app doesn’t share the user-assigned device name with any service providers or third parties other than cloud-hosting service providers
```

Stripe is a third party to anybody's app, so per Apple's rules we can't get this entitlement while Stripe is accessing `UIDevice.currentDevice.name` in its SDK.

 
## Describe the solution you'd like
Delete `deviceName` from [STDSDeviceInformationParameter](https://github.com/stripe/stripe-ios/blob/94583773331718db238b78f261ef2e5822e46450/Stripe3DS2/Stripe3DS2/STDSDeviceInformationParameter.m#L227), or if you want to keep it, change it to use something else like `UIDevice.currentDevice.model` that won't contain PII.

## Describe alternatives you've considered
Not using Stripe's SDK.

## Implementation
I've opened a PR https://github.com/stripe/stripe-ios/pull/2400 that removes the `deviceName` param entirely.

porter-stripe:
Hi @amayers, this has been resolved in version 23.5.1!

PR #2400, opened 2023-03-20
Remove usage of `UIDevice.currentDevice.name`

amayers:
## Summary
Remove usage of `UIDevice.currentDevice.name`.

## Motivation
As detailed in [this issue](https://github.com/stripe/stripe-ios/issues/2398), iOS 16 no longer returns a user specified device name like `Andrew’s iPhone 14 Pro`. Instead it just returns `iPhone`. That makes this parameter not very useful for Stripe. However apps can request an entitlement to get the full name again. But those apps are specifically prohibited from allowing that device name to be shared with third parties (Stripe). So by removing this access, app creators can now request this entitlement from Apple, and not be in violation of the rules around that entitlement.

## Testing
There are existing tests that confirmed that this `deviceName` was being included in the params. They have been updated to no longer expect it.


amayers:
Closing in favor of #2404.

PR #2401, opened 2023-03-20
Read error message from server during SSC

porter-stripe:
## Summary
- We now pass the server error to PaymentSheet when we fail to confirm server side
- Disable Link UI tests

## Motivation
Easier to test

## Testing
- Manual
- Updated tests

## Changelog
N/A


Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
```diff
@@ -704,9 +704,21 @@ extension PaymentSheetTestPlayground {
     func confirmHandlerForServerSideConfirmation(_ paymentMethodID: String,
                                                  _ shouldSavePaymentMethod: Bool,
                                                  _ intentCreationCallback: @escaping (Result<String, Error>) -> Void) {
-        enum ServerSideConfirmationError: Error {
+        enum ServerSideConfirmationError: Error, LocalizedError {
```
porter-stripe, line 707: This needs to be a `LocalizedError` and needs to implement `errorDescription: String?` for us to be able to read the [localizedDescription](https://developer.apple.com/documentation/foundation/nserror/1414418-localizeddescription) in PaymentSheet. 

Should we update the intent creation callback to be a `LocalizedError` instead of just an `Error`?

porter-stripe: We read localizedDescription here: https://github.com/stripe/stripe-ios/blob/7791e585d7f08cbdcec029496536c8d97c99d2ab/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Error%2BPaymentSheet.swift#L21

porter-stripe: Hmm, looks like we can't use `LocalizedError` with `Result`
https://developer.apple.com/documentation/swift/result

CleanShot 2023-03-20 at 11 34 48

porter-stripe: I think we either introduce our own new type for errors so they are displayed nicely or find a way to document this.

yuki-stripe: I don't think we need to **require** merchants use a LocalizedError. Also, NSError doesn't conform to it (confusingly). https://figure.ink/2021/07/18/practical-localized-error.html

Adding documentation sounds good.

Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
```diff
@@ -718,13 +730,20 @@ extension PaymentSheetTestPlayground {
...
         ] as [String: Any]
 
-        makeRequest(with: PaymentSheetTestPlayground.confirmEndpoint, body: body, completionHandler: { data, _, error in
+        makeRequest(with: PaymentSheetTestPlayground.confirmEndpoint, body: body, completionHandler: { data, response, error in
             guard
                 error == nil,
                 let data = data,
                 let json = try? JSONSerialization.jsonObject(with: data, options: []) as? [String: Any]
             else {
-                intentCreationCallback(.failure(error ?? ServerSideConfirmationError.unknown))
```
yuki-stripe, line 727: Why did we end up showing such a weird error message?  Can we show something a less bad if a merchant doesn't provide a `LocalizedError`? 

maybe we should add some unit tests for https://github.com/stripe/stripe-ios/blob/7791e585d7f08cbdcec029496536c8d97c99d2ab/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Error%2BPaymentSheet.swift#L21

porter-stripe: I'll work on adding tests, but I want to avoid testing the behavior of `localizedDescription` and rather our computed variable `nonGenericDescription`.

PR #2403, opened 2023-03-20
PS.FC update part 1 

yuki-stripe:
## Summary
- Rename ChoosePaymentOptionViewController -> PaymentSheetFlowControllerViewController
- DRY the code that checks theres >1 pm type available
- Add `update` to PaymentSheet.FlowController w/ TODOs

## Testing
Basic unit tests, UI tests to follow

## Changelog
No user-facing changes

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
@@ -169,15 +169,6 @@ extension PaymentSheet {
             ) { result in
                 switch result {
                 case .success(let intent, let paymentMethods, let isLinkEnabled):
-                    // Verify that there are payment method types available for the intent and configuration.
-                    let paymentMethodTypes = PaymentMethodType.filteredPaymentMethodTypes(
-                        from: intent,
-                        configuration: configuration)
-                    guard !paymentMethodTypes.isEmpty else {
-                        completion(.failure(PaymentSheetError.noPaymentMethodTypesAvailable(intentPaymentMethods: intent.recommendedPaymentMethodTypes)))
-                        return
-                    }
-
                     let flowController = FlowController(
                         intent: intent,
                         savedPaymentMethods: paymentMethods,
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
@@ -287,10 +287,38 @@ extension PaymentSheet {
         /// - Parameter intentConfiguration: An updated IntentConfiguration
         /// - Parameter completion: Called when the update completes with an optional error. Your implementation should get the customer's updated payment option by using the `paymentOption` property and update your UI. If an error occurred, retry. TODO(Update): Tell the merchant they need to disable the buy button.
         /// Don’t call this method while PaymentSheet is being presented.
-        func update(intentConfiguration: IntentConfiguration, completion: (Error?) -> ()) {
-            // 1. Load
-            // 2. Re-initialize PaymentSheetFlowControllerViewController
-            // 3. Update paymentOption - nil it out if its payment method type is no longer valid or we went from not showing a mandate to showing a mandate
+        func update(intentConfiguration: IntentConfiguration, completion: @escaping (Error?) -> ()) {
+            // 1. Load the intent, payment methods, and link data from the Stripe API
+            PaymentSheet.load(
+                mode: .deferredIntent(intentConfiguration),
+                configuration: configuration
+            ) { [weak self] loadResult in
+                guard let self = self else {
+                    assertionFailure("The PaymentSheet.FlowController instance was destroyed during a call to `update(intentConfiguration:completion:)`")
+                    return
+                }
+                switch loadResult {
+                case .success(let intent, let paymentMethods, let isLinkEnabled):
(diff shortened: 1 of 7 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
```diff
@@ -308,6 +280,43 @@ extension PaymentSheet {
...
+                switch loadResult {
+                case .success(let intent, let paymentMethods, let isLinkEnabled):
+                    // 2. Re-initialize PaymentSheetFlowControllerViewController to update the UI to match the newly loaded data e.g. payment method types may have changed.
+                    // TODO(Update:) Preserve the customer's previous inputs.
+                    self.viewController = Self.makeViewController(intent: intent, savedPaymentMethods: paymentMethods, isLinkEnabled: isLinkEnabled, configuration: self.configuration)
+
+                    // ❓
+                    // TODO for PR: What is this? I copied it from the other load call, but I can't figure out what it does.
+                    // Synchronously pre-load image into cache
+                    if let paymentOption = self.paymentOption {
```
porter-stripe, line 308: yeah this is... strange

porter-stripe (dismissed):
looks great!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
yuki-stripe: cc @wooj-stripe - do you know what this code does and if we need it?

wooj-stripe: In the original flow controller, we expose an image to customers so they can embed an image into their checkout page to display which payment method has been selected.  These image URLs can be specified in the payment method spec, and take some time to download, so we decided to prefetch these images prior to calling the completion handler of load(), so that customers wanting to render the image would have instant access to the downloaded image.  We did this to avoid having to change the signature of the load call of the flow controller.

In this context, I haven't looked at the code to figure out if we need to pre-load the image...  Do we expect customers to access the `paymentOption.image` (`UIImage`) immediately such they can render it in their view without having to hit the network?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
```diff
@@ -308,6 +280,43 @@ extension PaymentSheet {
...
 
+        /// 🚧 Under construction
+        /// Call this method when the IntentConfiguration values you used to initialize PaymentSheet.FlowController (amount, currency, etc.) change.
+        /// This ensures the appropriate payment methods are displayed, etc.
+        /// - Parameter intentConfiguration: An updated IntentConfiguration
+        /// - Parameter completion: Called when the update completes with an optional error. Your implementation should get the customer's updated payment option by using the `paymentOption` property and update your UI. If an error occurred, retry. TODO(Update): Tell the merchant they need to disable the buy button.
+        /// Don’t call this method while PaymentSheet is being presented.
+        func update(intentConfiguration: IntentConfiguration, completion: @escaping (Error?) -> Void) {
+            // 1. Load the intent, payment methods, and link data from the Stripe API
+            PaymentSheet.load(
```
porter-stripe, line 291: I wonder if we can make this async/await so it would be easy to in-order updates without canceling or anything like that. I can explore that in a following PR but just a thought.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
yuki-stripe: Thanks, I'll add a comment with all this.  I am more confused about the line below - I think it does nothing?  
```
                    // Synchronously pre-load image into cache
                    if let paymentOption = flowController.paymentOption {
                        _ = paymentOption.image // ? image is a let property, it has no side effects, is this a no-op?
                    }
```

wooj-stripe: AH-- we wanted `manualFlow.paymentOption` to execute to synchronously download the image, but in order for the project to compile, we need to access the newly created `paymentOption` variable based on the if/let.  therefore in order to hush the compiler, we added:
```
_ = paymentOption.image
```

yuki-stripe: 👍 I believe `let _ = self.paymentOption` achieves the same thing, then, so I'll rewrite.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
```diff
@@ -175,10 +175,9 @@ extension PaymentSheet {
                         isLinkEnabled: isLinkEnabled,
                         configuration: configuration)
 
-                    // Synchronously pre-load image into cache
-                    if let paymentOption = flowController.paymentOption {
-                        _ = paymentOption.image
-                    }
+                    // Synchronously pre-load image into cache.
+                    // Accessing flowController.paymentOption has the side-effect of ensuring its `image` property is loaded (e.g. from the internet instead of disk) before we call the completion handler.
+                    let _ = flowController.paymentOption
```
yuki-stripe: cc @wooj!

porter-stripe (approved):

PR #2404, opened 2023-03-20
Use localizedModel instead of name 

porter-stripe:
## Summary
Instead of using name we are moving to localizedModel as to not block merchants from obtaining the entitlement to read `UIDevice.currentDevice.name`.

## Motivation
https://github.com/stripe/stripe-ios/issues/2398

## Testing
Existing tests

## Changelog
See diff

### DIFF
--- a/Stripe3DS2/Stripe3DS2/STDSDeviceInformationParameter.m
+++ b/Stripe3DS2/Stripe3DS2/STDSDeviceInformationParameter.m
@@ -224,7 +224,7 @@ static const NSString * const kParameterNilCode = @"RE04";
     return [[STDSDeviceInformationParameter alloc] initWithIdentifier:@"C009"
                                                       permissionCheck:nil
                                                            valueCheck:^id _Nullable{
-                                                               return [UIDevice currentDevice].name;
+                                                               return [UIDevice currentDevice].localizedModel;
                                                            }];
 }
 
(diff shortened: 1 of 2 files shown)


Stripe3DS2/Stripe3DS2/STDSDeviceInformationParameter.m
```diff
@@ -224,7 +224,7 @@ + (instancetype)deviceName
     return [[STDSDeviceInformationParameter alloc] initWithIdentifier:@"C009"
                                                       permissionCheck:nil
                                                            valueCheck:^id _Nullable{
-                                                               return [UIDevice currentDevice].name;
+                                                               return [UIDevice currentDevice].localizedModel;
```
vardges-stripe, line 227: @porter-stripe why localized instead of model?

porter-stripe: Not a super strong reason besides I _think_ it should be closer to what `name` was since I assume it was localized to some degree based on what the user named their device.

I'm open to either model or localizedModel

vardges-stripe: ah yeah, makes sense.

vardges-stripe (approved):

PR #2419, opened 2023-03-27
Update custom example to use deferred workflow

porter-stripe:
## Summary
- Updates the custom example checkout to use deferred intents and a new server

Simulator Screen Shot - iPhone 13 - 2023-03-27 at 07 17 20

## Motivation
- Testing
- Better example

## Testing
- Manual
- Updated existing test

## Changelog
N/A


Example/PaymentSheet Example/PaymentSheet Example/ExampleCustomCheckoutViewController.swift
```diff
@@ -7,15 +7,62 @@
...
+    @IBOutlet weak var saladQuantityLabel: UILabel!
+    @IBOutlet weak var hotDogStepper: UIStepper!
+    @IBOutlet weak var saladStepper: UIStepper!
+    @IBOutlet weak var subtotalLabel: UILabel!
+    @IBOutlet weak var salesTaxLabel: UILabel!
+    @IBOutlet weak var totalLabel: UILabel!
+    @IBOutlet weak var subscribeSwitch: UISwitch!
+
+    private var paymentSheetFlowController: PaymentSheet.FlowController!
+    private let baseUrl = "https://stripe-mobile-payment-sheet-custom.glitch.me"
```
porter-stripe, line 31: Should we pay for this server so it doesn't sleep? I'm worried since the server goes to sleep it may cause tests to fail.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
```diff
@@ -300,6 +300,7 @@ extension PaymentSheet {
                     // 2. Re-initialize PaymentSheetFlowControllerViewController to update the UI to match the newly loaded data e.g. payment method types may have changed.
                     // TODO(Update:) Preserve the customer's previous inputs.
                     self.viewController = Self.makeViewController(intent: intent, savedPaymentMethods: paymentMethods, isLinkEnabled: isLinkEnabled, configuration: self.configuration)
+                    self.viewController.delegate = self
```
porter-stripe, line 303: Not setting the delegate was causing some parts of the checkout flow to become unresponsive after updating.

Example/PaymentSheet Example/PaymentSheet Example/ExampleCustomCheckoutViewController.swift
```diff
@@ -128,4 +234,66 @@ class ExampleCustomCheckoutViewController: UIViewController {
...
+        })
+
+        task.resume()
+    }
+
+    func createIntentRequestBody(paymentMethodID: String) -> Data {
+        var body: [String: Any?] = [
+            "payment_method_id": paymentMethodID,
+            "currency": "USD",
+            "amount": Int(total * 100),
```
porter-stripe, line 284: Any suggestions on a better example way to send amount?

Example/PaymentSheet Example/PaymentSheet Example/ExampleCustomCheckoutViewController.swift
```diff
@@ -7,15 +7,59 @@
...
+@_spi (STP) import StripePaymentSheet
 import UIKit
 
 class ExampleCustomCheckoutViewController: UIViewController {
     @IBOutlet weak var buyButton: UIButton!
     @IBOutlet weak var paymentMethodButton: UIButton!
     @IBOutlet weak var paymentMethodImage: UIImageView!
-    var paymentSheetFlowController: PaymentSheet.FlowController!
-    let backendCheckoutUrl = URL(string: "https://stripe-mobile-payment-sheet.glitch.me/checkout")!  // An example backend endpoint
+
```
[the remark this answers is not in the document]
porter-stripe: Hm, my understanding was that this workflow would eventually be the default suggested workflow which I why I modified the existing one. But the link would still be confusing until we GA regardless.

Example/PaymentSheet Example/PaymentSheet Example.xcodeproj/project.pbxproj
```diff
@@ -593,6 +596,7 @@
 				84CFB1ACA4915EE5B2E25341 /* EndpointSelectorViewController.swift in Sources */,
 				EB3C76C2A5704F94F0414BEF /* ExampleCheckoutViewController.swift in Sources */,
 				D235E82A7173E8051BC5A261 /* ExampleCustomCheckoutViewController.swift in Sources */,
+				615739F129D218C700111D23 /* ExampleCustomDeferredCheckoutViewController.swift in Sources */,
```
porter-stripe, line 599: @eurias-stripe I'm confused here, I thought we removed project files?

porter-stripe: Is this because we just did a release and I picked up these before they could be removed?

Example/PaymentSheet Example/PaymentSheet Example/ExampleCustomDeferredCheckoutViewController.swift
```diff
@@ -0,0 +1,334 @@
...
+    private let baseUrl = "https://stripe-mobile-payment-sheet-custom-deferred.glitch.me"
+    private var backendCheckoutUrl: URL {
+        return URL(string: baseUrl + "/checkout")!
+    }
+    private var createIntentUrl: URL {
+        return URL(string: baseUrl + "/create_intent")!
+    }
+    private var computeTotalsUrl: URL {
+        return URL(string: baseUrl + "/compute_totals")!
+    }
```
yuki-stripe, line 39: nit: I think these should be `let`s

Example/PaymentSheet Example/PaymentSheet Example/ExampleCustomDeferredCheckoutViewController.swift
```diff
@@ -0,0 +1,334 @@
...
+            }
+        }
+        alertController.addAction(OKAction)
+        present(alertController, animated: true, completion: nil)
+    }
+
+    // MARK: Confirm handler
+
+    func confirmHandler(_ paymentMethodID: String,
+                        _ intentCreationCallback: @escaping (Result<String, Error>) -> Void) {
```
yuki-stripe, line 182: let's show the server-side confirmation example

Example/PaymentSheet Example/PaymentSheet Example/ExampleCustomDeferredCheckoutViewController.swift
```diff
@@ -0,0 +1,334 @@
...
+            "payment_method_id": paymentMethodID,
+            "currency": "USD",
+            "hot_dog_count": hotDogStepper.value,
+            "salad_count": saladStepper.value,
+            "is_subscribing": subscribeSwitch.isOn,
+        ]
+
+        if subscribeSwitch.isOn {
+            body["setup_future_usage"] = "off_session"
+        }
```
yuki-stripe, line 311: do we need this? isn't it already handled by is_subscribing?

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
```diff
@@ -99,6 +99,49 @@ class PaymentSheetUITest: XCTestCase {
         okButton.tap()
     }
 
+    func testPaymentSheetCustomDeferred() throws {
```
yuki-stripe, line 102: how is this different from the other existing `testPaymentSheetCustomSaveAndRemoveCard_DeferredIntent_ServerSideConfirmation()`?  I think it's the fact that it updates, is there anything else?  We should rename this test to indicate what it's testing.  `testPaymentSheetCustomDeferred_update`?

Could we test the preservation of customer input, since we're testing the update flow?  Like, enter card details -> change the quantity -> expect card details to be preserved.

I'm not even sure we need to pay, if that part of the flow is covered by another test.

Example/PaymentSheet Example/PaymentSheet Example/ExampleCustomDeferredCheckoutViewController.swift
```diff
@@ -0,0 +1,334 @@
...
+            case .canceled:
+                print("Canceled!")
+            case .failed(let error):
+                print(error)
+                self.displayAlert("Payment failed: \n\(error.localizedDescription)")
+            }
+        }
+    }
+
+    @IBAction func hotDogStepperDidChange() {
```
yuki-stripe, line 106: I'm not sure if this is possible but can we point these UIs directly to the updateUI function instead of having these middlemen funcs?

Example/PaymentSheet Example/PaymentSheet Example/ExampleCustomDeferredCheckoutViewController.swift
porter-stripe: Computed var's can't be let's.

Example/PaymentSheet Example/PaymentSheet Example/ExampleCustomDeferredCheckoutViewController.swift
porter-stripe: Yeah I tried to consolidate but interface builder seems to not support this.

Example/PaymentSheet Example/PaymentSheet Example/ExampleCustomDeferredCheckoutViewController.swift
porter-stripe: Yep, after moving the amount calculation to the backend I didn't make this connection.

Example/PaymentSheet Example/PaymentSheet Example/ExampleCustomDeferredCheckoutViewController.swift
porter-stripe: Switched!

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
porter-stripe: Added a part of the test which tests input preservation, it passes since it doesn't use a leading zero expiry month.

Example/PaymentSheet Example/PaymentSheet Example/ExampleCustomDeferredCheckoutViewController.swift
yuki-stripe: Right, but why do they need to be computed vars?  They shouldn't change.

Example/PaymentSheet Example/PaymentSheet Example/ExampleCustomDeferredCheckoutViewController.swift
```diff
@@ -0,0 +1,332 @@
...
+    @IBOutlet weak var saladQuantityLabel: UILabel!
+    @IBOutlet weak var hotDogStepper: UIStepper!
+    @IBOutlet weak var saladStepper: UIStepper!
+    @IBOutlet weak var subtotalLabel: UILabel!
+    @IBOutlet weak var salesTaxLabel: UILabel!
+    @IBOutlet weak var totalLabel: UILabel!
+    @IBOutlet weak var subscribeSwitch: UISwitch!
+
+    private var paymentSheetFlowController: PaymentSheet.FlowController!
+    private let baseUrl = "https://stripe-mobile-payment-sheet-custom-deferred.glitch.me"
```
yuki-stripe, line 30: Let's add a comment like 
// View the backend code here: https://glitch.com/edit/#!/stripe-mobile-payment-sheet-custom-deferred

Example/PaymentSheet Example/PaymentSheet Example/ExampleCustomDeferredCheckoutViewController.swift
porter-stripe: Can't be let's since the variables use `baseUrl` which is an instance variable. We have a few options.

1. Make `baseUrl` static
2. Define like `var computeTotalsUrl = URL(string: "https://stripe-mobile-payment-sheet-custom-deferred.glitch.me/compute_totals)`
3. Leave as-is

Example/PaymentSheet Example/PaymentSheet Example/ExampleCustomDeferredCheckoutViewController.swift
```diff
@@ -0,0 +1,332 @@
...
+        present(alertController, animated: true, completion: nil)
+    }
+
+    // MARK: Server-side confirm handler
+
+    func serverSideConfirmHandler(_ paymentMethodID: String,
+                                  _ shouldSavePaymentMethod: Bool,
+                                  _ intentCreationCallback: @escaping (Result<String, Error>) -> Void) {
+        // Create an intent on your server and invoke `intentCreationCallback` with the client secret
+        createIntent(paymentMethodID: paymentMethodID) { result in
```
yuki-stripe, line 185: We need to pass `shouldSavePaymentMethod` to the server too and have it set `setup_future_usage = off_session` if it's `true`.  I've realized this isn't in the guide either, I'll add it there.

Example/PaymentSheet Example/PaymentSheet Example/ExampleCustomDeferredCheckoutViewController.swift
yuki-stripe: 1 sounds right to me.

Example/PaymentSheet Example/PaymentSheet Example/ExampleCustomDeferredCheckoutViewController.swift
porter-stripe: how does `shouldSavePaymentMethod` relate with the subscribe switch? Should we send the `setup_future_usage = off_session` if switch is on or if `shouldSavePaymentMethod` is true?

yuki-stripe: Yep that sounds right.

yuki-stripe (approved):

PR #2423, opened 2023-03-27
Remove generated project files for v23.6.0

eurias-stripe:


kgaidis-stripe (approved):
@eurias-stripe 

Is this something we have to do after every deploy or was it a mistake? Just curious to learn from it, and whether we have this documented anywhere or whether we need to add something to .gitignore etc.

eurias-stripe:
> @eurias-stripe
> 
> Is this something we have to do after every deploy or was it a mistake? Just curious to learn from it, and whether we have this documented anywhere or whether we need to add something to .gitignore etc.

It is documented in the [proposal](https://paper.dropbox.com/doc/Eliminate-Xcode-Project-Files--B1Qu9I5_9ZxPCEBlqoZucoYyAg-2QD9baK8ePqcQfTJUtvSw) to remove xcode project files.
Basically, when we release we need to include the xcode project files to support Carthage, but we don't want to have them in the repo normally, so we only generate them for the release and then we delete them, this is what this PR is doing.

PR #2424, opened 2023-03-27
Adds SimpleMandateElement, require it to be displayed 

yuki-stripe:
## Summary
1. Creates a `SimpleMandateElement` that requires its view to have been displayed before producing a valid payment option.
2. Preserves "view was displayed" state in IntentConfirmParams.

## Motivation
Make `PaymentSheet.FlowController.update` flow preserve customer input when the customer is using a payment method that can require a mandate, like Paypal.  Specifically, handle edge cases around the payment method being updated to include or not include a mandate e.g. going from payment -> setup mode.  See the unit tests for details.

## Testing
See unit test

## Changelog
No user facing changes

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/SimpleMandateElement.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/SimpleMandateElement.swift
@@ -20,14 +20,14 @@ class SimpleMandateElement: PaymentMethodElement {
             return nil
         }
     }
-    
+
     var delegate: StripeUICore.ElementDelegate?
     var view: UIView {
         return mandateTextView
     }
     let mandateTextView: SimpleMandateTextView
     let customerAlreadySawMandate: Bool
-    
+
     init(mandateText: String, customerAlreadySawMandate: Bool = false, theme: ElementsUITheme = .default) {
         mandateTextView = SimpleMandateTextView(mandateText: mandateText, theme: theme)
         self.customerAlreadySawMandate = customerAlreadySawMandate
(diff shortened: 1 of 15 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/SimpleMandateElement.swift
```diff
@@ -0,0 +1,35 @@
...
+        if mandateTextView.viewDidAppear {
+            params.didDisplayMandate = true
+        }
+        if customerAlreadySawMandate || mandateTextView.viewDidAppear {
+            // the customer must have seen the mandate for this to be valid
+            return params
+        } else {
+            return nil
+        }
+    }
```
yuki-stripe, line 22: This is the core logic fyi @porter-stripe @tillh-stripe @jameswoo-stripe

porter-stripe (dismissed):
cool solution!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/SimpleMandateElement.swift
```diff
@@ -0,0 +1,35 @@
...
+//  StripePaymentSheet
+//
+//  Created by Yuki Tokuhiro on 3/26/23.
+//
+
+@_spi(STP) import StripeUICore
+import UIKit
+
+class SimpleMandateElement: PaymentMethodElement {
+    func updateParams(params: IntentConfirmParams) -> IntentConfirmParams? {
```
porter-stripe, line 12: Nit:  Could we maybe add a little more comments around this logic and what it means for these params to be nil.

yuki-stripe: Thanks for the suggestion.  As I wrote it up, I realized we should set `didDisplayMandate = true` in the case where the customer already saw the mandate too.

porter-stripe (approved):

PR #2427, opened 2023-03-28
Support preserving previous customer input for all PaymentSheet fields

yuki-stripe:
## Summary
follow-on to https://github.com/stripe/stripe-ios/pull/2415

This PR adds support for preserving previous customer input for all payment method forms possible today.

## Testing
I've added tests for each field.  

Unfortunately, I can't figure out a way to ensure that we've covered *all* possible fields that can be displayed by the SDK.  Ideally, such a test would break when we **add** new fields, too, until the developer adds support for preserving previous customer input. 

## Changelog
no user facing changes


porter-stripe (approved):
Looks like all the elements are covered to me, tough to ensure future ones will preserve input :/

PR #2429, opened 2023-03-29
Update flow bugfixes

yuki-stripe:
## Summary
Fixed bugs in the update flow.  See commit log for each one.

## Testing
See unit test

## Changelog
no user facing changes yet


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/IntentConfirmParams.swift
```diff
@@ -0,0 +1,183 @@
+//
```
porter-stripe, line 1: What's new here vs copied from Intent.swift?

yuki-stripe: Ah look at the commits after the one where I moved it over - https://github.com/stripe/stripe-ios/pull/2429/files/422c5e023643788944105f9a2cbad383eb0f8eb0..fd117d94b89a860c39bd457cd31fba44f5b60c6c#diff-305ea2bdb8f8647eb2bfc2a7f506c6cecba3250c1884cbd970556c9e08379dce

porter-stripe (approved):
good fixes!

PR #2433, opened 2023-03-30
FC Networking: in warmup pane, we now use manifest to go to next pane when skipping

kgaidis-stripe:
## Summary

In Networking Warm Up Pane, recently Stripe.js changed logic to use the manifest, instead of hard-coding the pane. This PR updates the code to also do that.

## Testing

Here we can see that skipping in the warmup pane still leads to the institution picker (what was previously hard-coded).

https://user-images.githubusercontent.com/105514761/228983417-30515d9a-7470-4d56-91de-c81840e38666.mov

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/Native/NativeFlowController.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/Native/NativeFlowController.swift
@@ -114,18 +114,30 @@ extension NativeFlowController {
         navigationController.setViewControllers(viewControllers, animated: animated)
     }
 
-    private func pushPane(_ pane: FinancialConnectionsSessionManifest.NextPane, animated: Bool) {
+    private func pushPane(
+        _ pane: FinancialConnectionsSessionManifest.NextPane,
+        animated: Bool,
+        // useful for cases where we want to prevent the user from navigating back
+        //
+        // keeping this logic in `pushPane` is helpful because we want to
+        // reuse `skipSuccessPane` and `manualEntryMode == .custom` logic
+        clearNavigationStack: Bool = false
+    ) {
         if pane == .success && dataManager.manifest.skipSuccessPane == true {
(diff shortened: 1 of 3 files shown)


StripeFinancialConnections/StripeFinancialConnections/Source/Native/NativeFlowController.swift
```diff
@@ -114,18 +114,30 @@ extension NativeFlowController {
...
             closeAuthFlow(showConfirmationAlert: false, customManualEntry: true)
         } else {
-            let manualEntryViewController = CreatePaneViewController(
+            let paneViewController = CreatePaneViewController(
                 pane: pane,
                 nativeFlowController: self,
                 dataManager: dataManager
             )
-            pushViewController(manualEntryViewController, animated: animated)
+            if clearNavigationStack, let paneViewController = paneViewController {
```
kgaidis-stripe, line 136: because `clearNavigationStack` default value is set to `false`, this should mean that the changes to this function are a no-op for old callers

StripeFinancialConnections/StripeFinancialConnections/Source/Native/NativeFlowController.swift
```diff
@@ -581,10 +593,19 @@ extension NativeFlowController: NetworkingLinkLoginWarmupViewControllerDelegate
...
+        _ viewController: NetworkingLinkLoginWarmupViewController,
+        didSelectSkipWithManifest manifest: FinancialConnectionsSessionManifest
     ) {
-        pushPane(.institutionPicker, animated: true)
+        dataManager.manifest = manifest
+        pushPane(
+            manifest.nextPane,
+            animated: true,
+            // skipping disables networking, which means
+            // we don't want the user to navigate back to
```
kgaidis-stripe, line 605: https://stripe.slack.com/archives/C02CCTGEZPD/p1680205025687519

vardges-stripe (approved):

PR #2448, opened 2023-04-05
Decoupled metrics

porter-stripe:
## Summary
- Adds flags for isDecoupled and isServerSideConfirmation
- Appends `; deferred` to user agent when confirming a client side intent through the decoupled flow
- Example user agent `"stripe-ios/23.6.0; variant.paymentsheet; PaymentSheet; deferred"`

## Motivation
https://docs.google.com/document/d/1bjZvn5CFLr-l_n9TI0jYuJ5MHjxuSHV17Hw92FZMzeo/edit#

## Testing
Manual

## Changelog
N/A


porter-stripe:
@davidme-stripe any suggestions on a cleaner approach for modifying the user agent? Seems kinda messy but not sure of a cleaner approach.

porter-stripe:
cc @tillh-stripe @jameswoo-stripe

davidme-stripe:
The linked doc mentions "deferred-intent" as the value on Stripe.js, did they change it to "deferred"?

Also, if I'm reading this right, initializing a deferred intent will attribute all future requests as deferred — is that correct? I guess it wouldn't make sense for an app to use both a deferred and non-deferred PaymentSheet...

porter-stripe:
re: deferred vs deferred-intent, you're right good catch! Our doc wasn't updated but the stripe.js doc was. 

Yeah it wouldn't make sense for a merchant to be using a deferred integration and a non-deferred integration so you read that correctly.

Metin6201 (commented):
Danke

davidme-stripe (approved):

Issue #2446, opened 2023-04-05
[BUG] Failed to build the workspace with Tuist/ Xcode 14.3 


hassan-ioki:
## Summary
I am unable to build the workspace with Tuist/ Xcode 14.3 

## iOS version
14+

## Installation method
I install with Tuist but it fails with error

## SDK version
23.6.0

## Other information
The error is: 
`Task failed with exit code 65:
	/usr/bin/xcrun xcodebuild -workspace /Users/muhammad/../Carthage/Checkouts/stripe-ios/Stripe.xcworkspace -scheme StripeCameraCoreTestUtils -configuration Release -derivedDataPath /Users/muhammad/Library/Caches/org.carthage.CarthageKit/DerivedData/14.3_14E222b/stripe-ios/23.6.0 -sdk iphoneos ONLY_ACTIVE_ARCH=NO CODE_SIGNING_REQUIRED=NO CODE_SIGN_IDENTITY= CARTHAGE=YES archive VALIDATE_WORKSPACE=NO -archivePath /var/folders/xn/sqdyt50x5_d_0ffq6vjn_74c0000gn/T/stripe-ios SKIP_INSTALL=YES GCC_INSTRUMENT_PROGRAM_FLOW_ARCS=NO CLANG_ENABLE_CODE_COVERAGE=NO STRIP_INSTALLED_PRODUCT=NO (launched in /Users/muhammad/../Tuist/Dependencies/Carthage/Checkouts/stripe-ios)
`

The build error is: 
`module 'StripeCameraCore' was not compiled for testing
@_spi(STP)@testable import StripeCameraCore
`

davidme-stripe:
That's strange @hassan-ioki, thanks for filing this. It looks like something is picking up on and building the StripeCameraCoreTestUtils, which is unexpected. Can you share the Tuist command you're calling, and a snippet from your Tuist config to show how you're integrating Stripe?

hassan-ioki:
We use Carthage through Tuist, 
`let carthage = CarthageDependencies([ 
.
.
.github(path: "stripe/stripe-ios", requirement: .exact("22.5.0")),
])`

So for discussion, I am on `22.5.0` but I checked with the latest and it fails there. So for us, our old version is working well. 
Then I have stripe add as framework:- ` .xcframework(path: .relativeToRoot("Stripe.xcframework"))`

trevor-holliday-instacart:
I am seeing this issue as well with a carthage update command and no Tuist overhead.

[...]

PR #2453, opened 2023-04-07
Don't export TestUtils schemes

davidme-stripe:
## Summary
TestUtils schemes are being picked up by Carthage, which slows down builds and causes issues on Xcode 14.3. Don't export schemes for these — there's no reason to build them separately, as they're already implicitly built when running tests.

I tried setting `shared: false`, but it doesn't seem to be processed properly by Tuist. Left a comment with more explanation, maybe we can investigate more in the future.

Also fix the Carthage tests to build all configurations — the linked issue was only happening when building for Release.

## Motivation
https://github.com/stripe/stripe-ios/issues/2446

## Testing
CI, manually installing with Carthage on Xcode 14.3.

### DIFF
--- a/Tests/installation_tests/carthage/setup.sh
+++ b/Tests/installation_tests/carthage/setup.sh
@@ -40,7 +40,7 @@ info "Executing carthage bootstrap..."
 
 cd "${script_dir}" || die "Executing \`cd\` failed"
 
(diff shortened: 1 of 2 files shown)


porter-stripe (approved):

sfriedman-stripe:
Should we open an issue with Tuist?

PR #2460, opened 2023-04-12
Fix StripeIdentity install tests on Carthage

davidme-stripe:
## Summary
The Carthage install tests now build the framework for Release instead of Debug, which better matches the end user experience. But this means we can't use `@testable`, so these tests broke.

## Testing
CI, kicked off manual build: https://app.bitrise.io/build/d5980181-a21a-4d0a-a3b8-2abf65c8e306

### DIFF
--- a/Tests/installation_tests/shared_unit_tests/StripeIdentity/StripeIdentityAssetTests.swift
+++ b/Tests/installation_tests/shared_unit_tests/StripeIdentity/StripeIdentityAssetTests.swift
@@ -7,7 +7,7 @@
 import UIKit
 import XCTest
 
-@testable import StripeIdentity
+@_spi(STP) import StripeIdentity
 
 class StripeIdentityAssetTests: XCTestCase {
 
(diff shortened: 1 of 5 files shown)


vardges-stripe (dismissed):

Tests/installation_tests/shared_unit_tests/StripeIdentity/StripeIdentityAssetTests.swift
```diff
@@ -7,7 +7,7 @@
 import UIKit
 import XCTest
 
-@testable import StripeIdentity
```
ccen-stripe, line 10: it seems Tests/installation_tests/shared_unit_tests/StripeCore is also using `@test`, should we update those as well?

ccen-stripe (changes requested):

Tests/installation_tests/shared_unit_tests/StripeIdentity/StripeIdentityAssetTests.swift
davidme-stripe: Those tests appear to be unused, good catch. We should probably include them in the install tests.

ccen-stripe (approved):

Issue #2461, opened 2023-03-21
Visual bug on Payment Sheet saved cards


raphaelssampaio:

porter-stripe:
Thanks @raphaelssampaio for reporting this! I have submitted a fix in https://github.com/stripe/stripe-ios/pull/2462 and will be included in our next release. I will be sure to update this issue once it is live.

PR #2462, opened 2023-04-12
Reload all cells when removing payment methods

porter-stripe:
## Summary
- Fixes a bug where the delete icon was not being shown or removed in certain cases since we were not updating all cells only those visible.

Before:
Simulator Screen Recording - iPhone 12 Pro - 2023-04-12 at 09 58 04

After:
Simulator Screen Recording - iPhone 12 Pro - 2023-04-12 at 09 57 26

## Motivation
https://github.com/stripe/stripe-ios/issues/2461

## Testing
Manual

## Changelog
See diff

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentOptionsViewController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentOptionsViewController.swift
@@ -80,7 +80,7 @@ class SavedPaymentOptionsViewController: UIViewController {
         }
         set {
             collectionView.isRemovingPaymentMethods = newValue
-            collectionView.reloadItems(at: collectionView.indexPathsForVisibleItems)
+            collectionView.reloadData()
             if !collectionView.isRemovingPaymentMethods {
                 // re-select
                 collectionView.selectItem(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## X.X.X
 ### All
 * Xcode 13 is [no longer supported by Apple](https://developer.apple.com/news/upcoming-requirements/). Please upgrade to Xcode 14.1 or later.
+### PaymentSheet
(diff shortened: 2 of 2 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentOptionsViewController.swift
```diff
@@ -80,7 +80,7 @@ class SavedPaymentOptionsViewController: UIViewController {
         }
         set {
             collectionView.isRemovingPaymentMethods = newValue
-            collectionView.reloadItems(at: collectionView.indexPathsForVisibleItems)
+            collectionView.reloadData()
```
eurias-stripe, line 83: Using `reloadData` seems to be removing the animation, you can try using `reloadSections(0)`

porter-stripe: Ah good catch!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentOptionsViewController.swift
```diff
@@ -80,7 +80,7 @@ class SavedPaymentOptionsViewController: UIViewController {
         }
         set {
             collectionView.isRemovingPaymentMethods = newValue
-            collectionView.reloadItems(at: collectionView.indexPathsForVisibleItems)
+            collectionView.reloadSections(.init(integer: 0))
```
eurias-stripe, line 83: Can you try using the arrayLiteral initializer for `IndexSet`? I think `reloadSections([0])` should work.

porter-stripe: Yeah that also works

eurias-stripe (approved):

PR #2464, opened 2023-04-12
Merge master into SavedPaymentMethodBranch & redo storyboard v2

wooj-stripe:
Original branch:
walletModePrototypev4-customercontext

## Summary
Did a merge from master into `walletModePrototypev4-customercontext`, then handled merge and re-did storyboard

## Motivation
Get SavedPaymentMethodsSheet into `master` so others can work on it.

## Testing
Basic smoke tests, basic UI test in place


Example/UI Examples/UI Examples/Source/BrowseViewController.swift
```diff
@@ -11,6 +11,7 @@ import SwiftUI
 import UIKit
 
 @testable import Stripe
+@_spi(STP) import StripePaymentsUI
```
wooj-stripe, line 14: I think we can/should get rid of the @_spi(STP) here.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/SavedPaymentMethods/SavedPaymentMethodsCollectionViewController.swift
```diff
@@ -0,0 +1,416 @@
...
+        )
+    }
+
+    var isRemovingPaymentMethods: Bool {
+        get {
+            return collectionView.isRemovingPaymentMethods
+        }
+        set {
+            collectionView.isRemovingPaymentMethods = newValue
+            collectionView.reloadItems(at: collectionView.indexPathsForVisibleItems)
```
wooj-stripe, line 91: Should be:
```
collectionView.reloadSections([0])
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/SavedPaymentMethods/SavedPaymentMethodsSheetError.swift
```diff
@@ -0,0 +1,37 @@
+//
+//  SavedPaymentMethodsSheetError.swift
+//  StripePaymentSheet
+//
+
+import Foundation
+@_spi(STP) import StripeCore
+import StripePayments
+
+public enum SavedPaymentMethodsSheetError: Error {
```
wooj-stripe, line 10: todo: add `@_spi(PrivateBetaSavedPaymentMethodsSheet)`

Stripe/StripeiOSTests/STPAnalyticsClientPaymentsTest.swift
```diff
@@ -13,6 +13,8 @@ import XCTest
 @testable@_spi(STP) import StripeCore
 @testable@_spi(STP) import StripePayments
 @testable@_spi(STP) import StripePaymentSheet
+@testable@_spi(STP) import StripePaymentsUI
+@testable@_spi(STP) import StripePaymentsUI
```
davidme-stripe, line 17: Duplicate

Stripe/StripeiOSTests/STPAPIClientTest.swift
```diff
@@ -63,7 +64,7 @@ class STPAPIClientTest: XCTestCase {
     func testEphemeralKeyOverwritesHeader() {
         let sut = STPAPIClient(publishableKey: "pk_foo")
         let ephemeralKey = STPFixtures.ephemeralKey()
-        let additionalHeaders = sut.authorizationHeader(using: ephemeralKey)
+        let additionalHeaders = sut.authorizationHeader(using: ephemeralKey.secret)
```
davidme-stripe, line 67: wrong, should use `usingEphemeralKey:` instead of getting the secret directly

StripePaymentSheet/StripePaymentSheet/Source/Helpers/DefaultPaymentMethodStore.swift
```diff
@@ -45,8 +46,20 @@ final class DefaultPaymentMethodStore {
         var customerToDefaultPaymentMethodID = UserDefaults.standard.customerToLastSelectedPaymentMethod ?? [:]
 
         let key = customerID ?? ""
+        // Set Default to legacy behavior
```
davidme-stripe, line 49: I'm not sure what this is trying to do, as it's setting the field twice. Are we trying to migrate the default payment method option from the older style to this new JSON-encoded version? If so, I think it would make more sense to handle this on read:
* In `defaultPaymentMethod(for:)`, you could check for a modern value from `UserDefaults.standard.customerToLastSelectedPaymentMethod` (which would point to a new key for JSON-encoded selections).
* If one doesn't exist, you check `UserDefaults.standard.legacyCustomerToLastSelectedPaymentMethod` (which points to the old key) for a legacy value and convert it to the modern value.
* Otherwise, return nil.

Then in the `setDefaultPaymentMethod` code, you don't need to handle legacy values at all — just store the latest modern value.

Might be misunderstanding something though!

Stripe/StripeiOS/Source/STPPinManagementService.swift
```diff
@@ -9,6 +9,7 @@
 import Foundation
 @_spi(STP) import StripeCore
 @_spi(STP) import StripePayments
+@_spi(STP) import StripePaymentsUI
```
wooj-stripe, line 12: Not needed.

Stripe/StripeiOS/Source/STPPushProvisioningContext.swift
```diff
@@ -9,6 +9,7 @@
 import Foundation
 import PassKit
 @_spi(STP) import StripeCore
+@_spi(STP) import StripePaymentsUI
```
wooj-stripe, line 12: Not needed.

StripePaymentSheet/StripePaymentSheet/Source/Helpers/StaticEphemeralKeyProvider.swift
```diff
@@ -0,0 +1,46 @@
...
+//  StaticEphemeralKeyProvider.swift
+//  StripePaymentSheet
+//
+//  Created by David Estes on 3/13/23.
+//
+
+import Foundation
+@_spi(PrivateBetaSavedPaymentMethodsSheet) import StripePaymentsUI
+
+@_spi(PrivateBetaSavedPaymentMethodsSheet) public extension _stpspmsbeta_STPCustomerContext {
```
wooj-stripe, line 11: Hmm If we really don't want customers to use this --  should this be part of our playground package instead?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/SavedPaymentMethods/SavedPaymentMethodsCollectionViewController.swift
```diff
@@ -0,0 +1,420 @@
...
+  - Removes link supportPersistablePaymentMethodOption
+  - Does not save the selected payment method to the local device settings
+  - Fetches customerId using the underlying backing STPCustomerContext
+ */
+
+/// For internal SDK use only
+@objc(STP_Internal_SavedPaymentMethodsCollectionViewController)
+class SavedPaymentMethodsCollectionViewController: UIViewController {
+    // MARK: - Types
+    // TODO (cleanup) Replace this with didSelectX delegate methods. Turn this into a private ViewModel class
```
wooj-stripe, line 33: remove.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/SavedPaymentMethods/SavedPaymentMethodsFormFactory.swift
```diff
@@ -0,0 +1,67 @@
...
+    func make() -> PaymentMethodElement {
+        if paymentMethod == .card {
+            return makeCard(theme: theme)
+        }
+        assert(false, "Currently only support cards")
+    }
+
+    func makeCard(theme: ElementsUITheme = .default) -> PaymentMethodElement {
+        // TODO: Figure out if we need to add checkbox -- probably not, basic integration add payment method does not.
+//        let saveCheckbox = makeSaveCheckbox(
```
davidme-stripe, line 45: I think we can remove this, afaik it's implicit in the same way a SetupIntent is implicitly saving the card

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/SavedPaymentMethods/SavedPaymentMethodsCollectionViewController.swift
```diff
@@ -0,0 +1,420 @@
...
+        }
+        return true
+    }
+
+    func collectionView(_ collectionView: UICollectionView, didSelectItemAt indexPath: IndexPath) {
+        selectedViewModelIndex = indexPath.item
+        let viewModel = viewModels[indexPath.item]
+
+        // For wallet mode, I don't think we need to be setting the default --
+        // we can call the delegate, and just return the payment method instead
```
wooj-stripe, line 359: Remove!!!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/SavedPaymentMethods/SavedPaymentMethodsFormFactory.swift
```diff
@@ -0,0 +1,67 @@
...
+//            label: String.Localized.save_this_card_for_future_$merchant_payments(
+//                merchantDisplayName: configuration.merchantDisplayName
+//            )
+//        )
+//        let shouldDisplaySaveCheckbox: Bool = saveMode == .userSelectable && !canSaveToLink
+        let cardFormElement = FormElement(elements: [
+            CardSection(theme: theme),
+            // shouldDisplaySaveCheckbox ? saveCheckbox : nil,
+        ], theme: theme)
+//        if isLinkEnabled {
```
davidme-stripe, line 55: We can delete this

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/SavedPaymentMethods/SavedPaymentMethodsFormFactory.swift
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    func makeCard(theme: ElementsUITheme = .default) -> PaymentMethodElement {
+        // TODO: Figure out if we need to add checkbox -- probably not, basic integration add payment method does not.
+//        let saveCheckbox = makeSaveCheckbox(
+//            label: String.Localized.save_this_card_for_future_$merchant_payments(
+//                merchantDisplayName: configuration.merchantDisplayName
+//            )
+//        )
+//        let shouldDisplaySaveCheckbox: Bool = saveMode == .userSelectable && !canSaveToLink
```
wooj-stripe, line 50: remove.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/SavedPaymentMethods/SavedPaymentMethodsSheet.swift
```diff
@@ -0,0 +1,233 @@
...
+//  Doing so exposes internal functionality which may cause unexpected behavior if used directly.
+//
+import Foundation
+@_spi(STP) import StripeCore
+@_spi(STP) import StripePayments
+@_spi(STP) @_spi(PrivateBetaSavedPaymentMethodsSheet) import StripePaymentsUI
+@_spi(STP) import StripeUICore
+import UIKit
+
+@frozen public enum SavedPaymentMethodsSheetResult {
```
davidme-stripe, line 17: Add @_spi(PrivateBetaSavedPaymentMethodsSheet)

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/SavedPaymentMethods/SavedPaymentMethodsSheet.swift
```diff
@@ -0,0 +1,233 @@
...
+extension SavedPaymentMethodsSheet {
+    func loadPaymentMethods(completion: @escaping (Result<[STPPaymentMethod], SavedPaymentMethodsSheetError>) -> Void) {
+        configuration.customerContext.listPaymentMethodsForCustomer {
+            paymentMethods, error in
+            guard let paymentMethods = paymentMethods, error == nil else {
+                // TODO: Pass errors from the customerContext
+                let error = PaymentSheetError.unknown(debugDescription: "Failed to retrieve PaymentMethods for the customer")
+//                let error = error ?? PaymentSheetError.unknown(
+//                    debugDescription: "Failed to retrieve PaymentMethods for the customer"
+//                )
```
wooj-stripe, line 150: remove.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/SavedPaymentMethods/SavedPaymentMethodsSheet.swift
```diff
@@ -0,0 +1,233 @@
...
+import Foundation
+@_spi(STP) import StripeCore
+@_spi(STP) import StripePayments
+@_spi(STP) @_spi(PrivateBetaSavedPaymentMethodsSheet) import StripePaymentsUI
+@_spi(STP) import StripeUICore
+import UIKit
+
+@frozen public enum SavedPaymentMethodsSheetResult {
+    /// The customer completed the payment or setup
+    /// - Note: The payment may still be processing at this point; don't assume money has successfully moved.
```
davidme-stripe, line 19: Does this comment still make sense for SPMS? We're not completing a payment.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/SavedPaymentMethods/SavedPaymentMethodsSheet.swift
```diff
@@ -0,0 +1,233 @@
...
+}
+
+extension SavedPaymentMethodsSheet {
+    func loadPaymentMethods(completion: @escaping (Result<[STPPaymentMethod], SavedPaymentMethodsSheetError>) -> Void) {
+        configuration.customerContext.listPaymentMethodsForCustomer {
+            paymentMethods, error in
+            guard let paymentMethods = paymentMethods, error == nil else {
+                // TODO: Pass errors from the customerContext
+                let error = PaymentSheetError.unknown(debugDescription: "Failed to retrieve PaymentMethods for the customer")
+//                let error = error ?? PaymentSheetError.unknown(
```
davidme-stripe, line 148: Delete?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/SavedPaymentMethods/SavedPaymentMethodsSheet.swift
```diff
@@ -0,0 +1,233 @@
...
+
+@_spi(PrivateBetaSavedPaymentMethodsSheet) extension _stpspmsbeta_STPCustomerContext {
+    /// Returns the selected Payment Option for this customer context.
+    /// You can use this to obtain the selected payment method without loading the SavedPaymentMethodsSheet.
+    public func retrievePaymentOptionSelection(
+        completion: @escaping (SavedPaymentMethodsSheet.PaymentOptionSelection?, Error?) -> Void
+    ) {
+        self.listPaymentMethodsForCustomer { paymentMethods, error in
+            guard let paymentMethods = paymentMethods, error == nil else {
+                // TODO: Pass errors from the customerContext
```
wooj-stripe, line 206: Can't remember who wrote this `TODO`, but is this still needed? seems like calling the completion block is sufficient.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/SavedPaymentMethods/SavedPaymentMethodsSheetConfiguration.swift
```diff
@@ -0,0 +1,108 @@
...
+extension SavedPaymentMethodsSheet {
+
+    public struct Configuration {
+        public typealias CreateSetupIntentHandlerCallback = ((@escaping (String?) -> Void) -> Void)
+
+        private var styleRawValue: Int = 0  // SheetStyle.automatic.rawValue
+        /// The color styling to use for PaymentSheet UI
+        /// Default value is SheetStyle.automatic
+        /// @see SheetStyle
+        @available(iOS 13.0, *)
```
davidme-stripe, line 21: We can remove `@available` (and the below hacky comment and private var), as we require iOS 13

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/SavedPaymentMethods/SavedPaymentMethodsSheetError.swift
```diff
@@ -0,0 +1,37 @@
...
+
+import Foundation
+@_spi(STP) import StripeCore
+import StripePayments
+
+@_spi(PrivateBetaSavedPaymentMethodsSheet) public enum SavedPaymentMethodsSheetError: Error {
+
+    case errorFetchingSavedPaymentMethods(Error)
+
+    /// setupIntent is invalid
```
wooj-stripe, line 14: Update comments for first two cases.

StripePaymentsUI/StripePaymentsUI/Source/Helpers/STPAPIClient+CustomerContext.swift
```diff
@@ -0,0 +1,123 @@
...
+extension STPAPIClient {
+
+    /// Update a customer with parameters
+    /// - seealso: https://stripe.com/docs/api#update_customer
+    func updateCustomer(
+        withParameters parameters: [String: Any],
+        using ephemeralKey: _stpspmsbeta_STPEphemeralKey,
+        completion: @escaping STPCustomerCompletionBlock
+    ) {
+        let endpoint = "\(APIEndpointCustomers)/\(ephemeralKey.customerID ?? "")"
```
wooj-stripe, line 21: Seems like we should `guard let customerID = ephemeralKey.customerID else { completion(...) }` ?

StripePaymentsUI/StripePaymentsUI/Source/Helpers/STPAPIClient+CustomerContext.swift
```diff
@@ -0,0 +1,123 @@
...
+
+    /// Retrieves a list of Payment Methods attached to a customer.
+    /// @note This only fetches card type Payment Methods
+    func listPaymentMethodsForCustomer(
+        using ephemeralKey: _stpspmsbeta_STPEphemeralKey,
+        completion: @escaping STPPaymentMethodsCompletionBlock
+    ) {
+        let header = authorizationHeader(using: ephemeralKey.secret)
+        let params: [String: Any] = [
+            "customer": ephemeralKey.customerID ?? "",
```
wooj-stripe, line 83: Seems like we should guard/let w/ `ephemeralKey.customerID`

StripePaymentsUI/StripePaymentsUI/Source/Helpers/STPAPIClient+CustomerContext.swift
```diff
@@ -0,0 +1,123 @@
...
+        }
+    }
+
+    /// Retrieve a customer
+    /// - seealso: https://stripe.com/docs/api#retrieve_customer
+    func retrieveCustomer(
+        using ephemeralKey: _stpspmsbeta_STPEphemeralKey,
+        completion: @escaping STPCustomerCompletionBlock
+    ) {
+        let endpoint = "\(APIEndpointCustomers)/\(ephemeralKey.customerID ?? "")"
```
wooj-stripe, line 106: guard let with `ephemeralKey.customerID` ?

StripePaymentsUI/StripePaymentsUI/Source/UI Components/CustomerContext/STPEphemeralKey.swift
```diff
@@ -33,9 +33,9 @@ class STPEphemeralKey: NSObject, STPAPIResponseDecodable {
         super.init()
     }
 
-    private(set) var allResponseFields: [AnyHashable: Any] = [:]
+    @_spi(STP) public var allResponseFields: [AnyHashable: Any] = [:]
```
wooj-stripe: This should probably be `@_spi(STP) public  public private(set)`

StripePaymentsUI/StripePaymentsUI/Source/UI Components/CustomerContext/STPEphemeralKeyManager.swift
```diff
@@ -65,7 +65,7 @@ class STPEphemeralKeyManager: NSObject, STPEphemeralKeyManagerProtocol {
         )
     }
 
-    @objc dynamic func getOrCreateKey(_ completion: @escaping STPEphemeralKeyCompletionBlock) {
+    @objc @_spi(STP) public dynamic func getOrCreateKey(_ completion: @escaping STPEphemeralKeyCompletionBlock) {
```
wooj-stripe: Curious as to why this needs to be @_spi(STP) public -- the caller is `_stpspmsbeta_STPCustomerContext` which also lives in the same module -- i guess i can see it being 'public' in the even that users implement their own STPCustomerContext, but not sure i fi understand why `@_spi(STP) public` is needed.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/SavedPaymentMethods/SavedPaymentMethodsSheet.swift
wooj-stripe: resolving-- todo is still there, but i have a comment on it below.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/SavedPaymentMethods/SavedPaymentMethodsSheet.swift
wooj-stripe: Comment doesn't make sense, and actually, we don't need to make this public -- it is purely an object we use for confirming an intent from savedpaymentmethodssheet.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/SavedPaymentMethods/SavedPaymentMethodsSheet.swift
wooj-stripe: Removed public instead :D

StripePaymentsUI/StripePaymentsUI/Source/UI Components/CustomerContext/STPEphemeralKeyManager.swift
davidme-stripe: Stripe.framework still uses it for Issuing, which we haven't split into a separate SDK yet.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/SavedPaymentMethods/SavedPaymentMethodsSheet.swift
davidme-stripe: I think the current implementation is fine, though it seems like we're strangely nesting a PaymentSheet error within the SPMS error:
```
                let error = PaymentSheetError.unknown(debugDescription: "Failed to retrieve PaymentMethods for the customer")
                completion(.failure(.errorFetchingSavedPaymentMethods(error)))
```

StripePaymentSheet/StripePaymentSheet/Source/Helpers/StaticEphemeralKeyProvider.swift
davidme-stripe: I think the convenience init is useful to users, it makes it easy to create an STPCustomerContext with the same values you'd use for a PaymentSheet Customer Settings struct

Stripe/StripeiOS/Source/STPPushProvisioningContext.swift
davidme-stripe: It's not needed today, but it will be by beta. We can take it out for now though

Stripe/StripeiOS/Source/STPPinManagementService.swift
davidme-stripe: Not needed today, but it will be once we de-dupe the CustomerContext.

Stripe/StripeiOS/Source/STPAPIClient+PushProvisioning.swift
```diff
@@ -7,7 +7,9 @@
 //
 
 import Foundation
+@_spi(STP) import StripeCore
 @_spi(STP) import StripePayments
+@_spi(STP) import StripePaymentsUI
```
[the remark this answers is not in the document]
davidme-stripe: Not needed today, but it will be once we de-dupe the CustomerContext.

StripePaymentSheet/StripePaymentSheet/Source/Helpers/DefaultPaymentMethodStore.swift
wooj-stripe: i think this is just a misunderstanding -- I was thinking it was possible that
* `if let paymentMethodOption = PersistablePaymentMethodOption(legacyValue: identifier.value) {` - could fail, but unlikely, in which case, we wouldn't be writing anything to user defaults
* `try encoder.encode(...)` - could also fail, in which case, we are also not writing anything to user defaults.

Both of these cases are unlikely, but it was my attempt at writing "at least something" to user defaults. -- maybe we should just assume that both of those cases will succeed -- and if they fail, write nothing to user defaults?

wooj-stripe: FWIW - i think `defaultPaymentMethod(for:)` is doing what you are outlining -- attempting to decode to the new version first, and if decoding fails, we call `return PaymentMethodIdentifier(value: value)` on line 92

StripePaymentSheet/StripePaymentSheet/Source/Helpers/StaticEphemeralKeyProvider.swift
wooj-stripe: I see- then we should update the comment to:
    /// This is not recommended, use initWithKeyProvider and an STPCustomerEphemeralKeyProvider instead.
    /// The key will expire after 60 minutes.

StripePaymentSheet/StripePaymentSheet/Source/Helpers/DefaultPaymentMethodStore.swift
davidme-stripe: `encoder.encode` shouldn't fail, as we control the entire (very simple) struct — that's a programming error for us if it does. Seems okay to write nothing in that situation.

I'm confused as to why we're using `PersistablePaymentMethodOption(legacyValue:)` to construct this — why are we receiving a legacy value from a different part of our own SDK? Why wouldn't we initialize the `PersistablePaymentMethodOption` directly from the `PaymentMethodIdentifier`, instead of converting it to/from a raw value?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/SavedPaymentMethods/SavedPaymentMethodsCollectionViewController.swift
wooj-stripe: was a copy and paste from `SavedPaymentOptionsViewController`

StripePaymentsUI/StripePaymentsUI/Source/UI Components/CustomerContext/STPEphemeralKeyManager.swift
wooj-stripe: Chatted offline about this  -- was looking at some outdated code. my bad!

StripePaymentSheet/StripePaymentSheet/Source/Helpers/StaticEphemeralKeyProvider.swift
davidme-stripe: Makes sense!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/SavedPaymentMethods/SavedPaymentMethodsSheet.swift
wooj-stripe: Hmm Changed this to just:
```
                completion(nil, error)
```
I think this is what we want.

StripePaymentSheet/StripePaymentSheet/Source/Helpers/DefaultPaymentMethodStore.swift
wooj-stripe: I ran into an issue where `PersistablePaymentMethodOption` (which is a part of StripePaymentsUI) can't have an init which accepts a `PaymentMethodIdentifier` (which is a part of StripePaymentMethodSheet) .. but .. hmmm we could add an extension in `PaymentMethodIdentifier` to do this

wooj-stripe: sorry for the churn here, ready for review!

davidme-stripe (approved):

wooj-stripe (commented):
github won't let me ship, so here is my unofficial🚢

PR #2465, opened 2023-04-12
Use payment method we create when confirming client side 

porter-stripe:
## Summary
- In the deferred flow we were not using the payment method we created when confirming an intent client side. This was resulting in a new payment method being created when we confirmed the intent rather than using the one we vend back to the merchant.
- When you would supply create payment method params and a payment method on the intent it would default to creating a new payment method with the create params overriding any assigned payment method to the intent at time of creation.
- Now when confirming an intent client side we attach the created payment method id to the confirm params to ensure it uses the same payment method as the merchant expects.

cc @yuki-stripe

## Motivation
Deferred private beta

## Testing
- Manual through inspecting the dashboard
- Difficult to test all client side since the backend contains some logic

## Changelog
N/A

### DIFF
--- a/Stripe/StripeiOSTests/PaymentSheet+APITest.swift
+++ b/Stripe/StripeiOSTests/PaymentSheet+APITest.swift
@@ -46,16 +46,14 @@ class PaymentSheetAPITest: XCTestCase {
         cardParams.cvc = "123"
         cardParams.expYear = 32
         cardParams.expMonth = 12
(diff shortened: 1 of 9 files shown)


porter-stripe:
cc @tillh-stripe @jameswoo-stripe

porter-stripe:
@jameswoo-stripe  in favor of https://github.com/stripe/stripe-ios/pull/2466

PR #2466, opened 2023-04-13
Use payment method we create when confirming client side

porter-stripe:
## Summary
- In the deferred flow we were not using the payment method we created when confirming an intent client side. This was resulting in a new payment method being created when we confirmed the intent rather than using the one we vend back to the merchant.
- We were not setting the payment method id on the confirm params resulting in a new payment method being created
- Now when confirming an intent client side we attach the created payment method id to the confirm params to ensure it uses the same payment method as the merchant expects.

cc @yuki-stripe

## Motivation
Deferred private beta

## Testing
- Manual through inspecting the dashboard
- Difficult to test all client side since the backend contains some logic

## Changelog
N/A


wooj-stripe (approved):
lgtm!

PR #2474, opened 2023-04-15
[Identity] Support reusing a verification session

ccen-stripe:
## Summary

Add support to start a verification session halfway.

Introduce `IdentityTopLevelDestination` to abstract different screens.
Instead of navigating to `ConsentDestiantion` or `IndividualWelcomeDestiantion` on `InitialiLoadingScreen`, check the requirement and navigate from where the session is left off, the logic will also be used in test model's preview.

Similar Android change: https://github.com/stripe/stripe-android/pull/6530

## Motivation

* Support start a verification half way
* Support test mode

## Testing

- [x] Added tests
- [x] Modified tests
- [x] Manually verified

# Recordings
| Before  | After |
| reuseBefore  | reuseAfter |

### DIFF
--- a/StripeIdentity/StripeIdentity/Source/NativeComponents/Coordinators/VerificationSheetController.swift
+++ b/StripeIdentity/StripeIdentity/Source/NativeComponents/Coordinators/VerificationSheetController.swift
@@ -143,10 +143,11 @@ final class VerificationSheetController: VerificationSheetControllerProtocol {
         let returnedPromise = Promise<StripeAPI.VerificationPage>()
         // Only update `verificationPageResponse` on main
         apiClient.getIdentityVerificationPage().observe(on: .main) { [weak self] result in
-            self?.verificationPageResponse = result
+            guard let self = self else { return }
+            self.verificationPageResponse = result
             if case .success(let verificationPage) = result {
-                self?.startLoadingMLModels(from: verificationPage)
-                self?.isVerificationPageSubmitted = verificationPage.submitted
+                self.startLoadingMLModels(from: verificationPage)
+                self.isVerificationPageSubmitted = verificationPage.submitted
                 // if result success and requires address, load address spec before continue
                 if verificationPage.requirements.missing.contains(.address) {
                     AddressSpecProvider.shared.loadAddressSpecs {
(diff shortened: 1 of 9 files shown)


ccen-stripe:

StripeIdentity/StripeIdentity/Source/NativeComponents/Coordinators/VerificationSheetController.swift
```diff
@@ -143,6 +146,7 @@ final class VerificationSheetController: VerificationSheetControllerProtocol {
             self?.verificationPageResponse = result
             if case .success(let verificationPage) = result {
                 self?.startLoadingMLModels(from: verificationPage)
+                self?.isVerificationPageSubmitted = verificationPage.submitted
```
vardges-stripe, line 149: nit: usually referencing self weakly more than once is not recommended. As technically self can get deallocated in-between first and second calls. You can add `guard let self = self else { return }` and then use self for both lines

StripeIdentity/StripeIdentity/Source/NativeComponents/Coordinators/VerificationSheetController.swift
```diff
@@ -391,33 +417,34 @@ final class VerificationSheetController: VerificationSheetControllerProtocol {
...
         dataToBeCollected: StripeAPI.VerificationPageCollectedData
     ) -> StripeAPI.VerificationPageClearData {
-        return .init(
-            clearFields: Set(StripeAPI.VerificationPageFieldType.allCases).subtracting(
+
+        let initialMissings: Set<StripeAPI.VerificationPageFieldType>
+        do {
+            initialMissings = try verificationPageResponse?.get().requirements.missing ?? Set()
+        } catch {
+            assertionFailure("verificationPageResponse is nil, using StripeAPI.VerificationPageFieldType.allCases as initialMissings")
```
vardges-stripe, line 439: this will crash in dev builds and will be a no-op in release builds. Is that intentional?

vardges-stripe (commented):
Makes sense, just some small questions

vardges-stripe:
Thanks @ccen-stripe would also be be great to take a screen grab of your manual testing and attach to this PR. It's easier to understand context when there are screenshots/videos.

StripeIdentity/StripeIdentity/Source/NativeComponents/Coordinators/VerificationSheetController.swift
ccen-stripe: Yes, this condition is never supposed to happen in prod, as backend always sets `verificationPageResponse?.get().requirements.missing`, just added a log for sanity check as we do similar logs in other places.

ccen-stripe:
> Thanks @ccen-stripe would also be be great to take a screen grab of your manual testing and attach to this PR. It's easier to understand context when there are screenshots/videos.

Updated with recordings!

vardges-stripe (approved):
Perfect thanks @ccen-stripe

PR #2476, opened 2023-04-17
FC Networking: fixed a janky keyboard animation in signup pane.

kgaidis-stripe:
## Summary

See feedback ("The transition into the Save to Link pane is a little jarring") from [this document](https://docs.google.com/document/d/1TIAPlrpCNQWPo3uBOTs77qPPBdnTGNaHlM9XnfSz684/edit#).

They keyboard was starting to animate as the transition was happening, so here we "delay" the logic until the view appears. This is a simple solution, so its good because its unlikely to cause any unexpected issues.

## Testing

See fixed keyboard animation + also read code (not much changed):

https://user-images.githubusercontent.com/105514761/232534494-40a847a6-910c-4ead-9b6a-7e0615e1ff7e.mp4


vardges-stripe (approved):
the old run once at view did appear hack :)

PR #2489, opened 2023-04-19
Fix constraint error in HTMLViewWithIconLabels

nickdnk:
## Summary
Fixes a constraint error. Edit: Explained in email to Chen Cen, but recap:

The `UIStackView` used has a spacing of 16, but a manually added constraint of 24 points between the divider and the text below - inside the stack view - caused a constraint error, as both are priority 1000. This fix uses the built-in custom spacing for the divider and removes the manually added constraint.

## Motivation
Bug, self-explanatory.

## Testing
I ran it. It no longer produces constraint errors and renders as expected (with the correct 24 point spacing).

## Changelog
n/a


nickdnk:
Fixed with https://github.com/stripe/stripe-ios/pull/2494.

PR #2490, opened 2023-04-19
Decoupled payment user agent

yuki-stripe:
## Summary
Follow-up to https://github.com/stripe/stripe-ios/pull/2448

Instead of adding `IntentConfiguration` as its own product, we add either "PaymentSheet" or "PaymentSheet; deferred-intent" to product usage.

## Motivation
- Adding `IntentConfiguration` to product usage breaks the pattern/intended usage of "product usage", since IntentConfiguration is not a product.
- We should the "deferred-intent" alongside "PaymentSheet" so that its appended to all of PaymentSheet's API calls.

## Testing
See unit tests

## Changelog
not user facing


davidme-stripe (dismissed):

porter-stripe (approved):
I'm struggling with the motivation for this change, I feel it is roughly equivalent to what we currently have. Are there endpoints/requests not having deferred intent appended that are outlined in this [doc](https://docs.google.com/document/d/1bjZvn5CFLr-l_n9TI0jYuJ5MHjxuSHV17Hw92FZMzeo/edit#heading=h.cz1xkpga7giy)? While I agree `IntentConfiguraiton` is not a product, it's a good proxy for deferred intents which could be argued is a product or just a feature? The reason deferred intents was added to the product usage was out of convenience, given how we construct the payment user agent and no perceived downside besides potentially violating a vague definition of a product. I figure this is why we're still adding it to the product usage here even though it's not outlined in the metrics [doc](https://docs.google.com/document/d/1TR9CDlgreEAma1jFW5APVlto-kHCiv0K2GW4hEKnfw4/edit).

I don't have any issues with this change, but it feels unnecessary unless there are endpoints/requests where deferred intent isn't being appended that I overlooked or if we have a specific format we want the product usage of deferred intents to take. We should also sync with Android if we are going to want deferred intent to look a certain way in the product usage, as I'm not sure it's being sent in the product usage at all since in iOS, it was just being sent as a side effect of implementation convenience.

yuki-stripe:
> I'm struggling with the motivation for this change, I feel it is roughly equivalent to what we currently have. Are there endpoints/requests not having deferred intent appended that are outlined in this [doc](https://docs.google.com/document/d/1bjZvn5CFLr-l_n9TI0jYuJ5MHjxuSHV17Hw92FZMzeo/edit#heading=h.cz1xkpga7giy)? While I agree `IntentConfiguraiton` is not a product, it's a good proxy for deferred intents which could be argued is a product or just a feature? The reason deferred intents was added to the product usage was out of convenience, given how we construct the payment user agent and no perceived downside besides potentially violating a vague definition of a product. I figure this is why we're still adding it to the product usage here even though it's not outlined in the metrics [doc](https://docs.google.com/document/d/1TR9CDlgreEAma1jFW5APVlto-kHCiv0K2GW4hEKnfw4/edit).
> 
> I don't have any issues with this change, but it feels unnecessary unless there are endpoints/requests where deferred intent isn't being appended that I overlooked or if we have a specific format we want the product usage of deferred intents to take. We should also sync with Android if we are going to want deferred intent to look a certain way in the product usage, as I'm not sure it's being sent in the product usage at all since in iOS, it was just being sent as a side effect of implementation convenience.

Oops, I mistakenly thought that payment user agent was a header sent in all API requests, like the user agent header.  

I think we should either
1. close this PR and add a comment indicating `IntentConfiguration` is only added to product usage as a hack to get it in the payment user agent
2. keep this PR and properly make it a part of the product_usage string.  The prior way of adding it in after the product is used is inconsistent, not only b/c IntentConfiguration isn't like all the other things in product_usage, but also because it means some analytics requests from the deferred PaymentSheet flow will be "missing" the deferred_intent string.

Since we don't actually want to modify product_usage in the analytics payloads at all, we should probably do #1.

Issue #2485, opened 2023-04-18
[Feature] Provide the option to customize the loading state of the text fields (opacity)


nicklbaert:
## Is your feature request related to a problem? Please describe.
During loading after entering payment information, the input field content changes color (or opacity, not sure) and becomes almost invisible, which does not look very good.

## Describe the solution you'd like
An option to customize the look of the input field text during loading would be awesome!

## Describe alternatives you've considered
-

## Additional context
Screenshot of the state I am talking about:
<img width="567" alt="Screenshot 2023-04-18 at 18 12 30" src="https://user-images.githubusercontent.com/33283708/232839271-358bd09e-1246-43de-8deb-9f5783cf86b1.png">

Thank you guys <3

eurias-stripe:
Hi @nicklbaert, have you made any changes using the appearance API? Looking at this state using the defaults the text is visible while loading:
![Simulator Screen Shot - iPhone 14 Pro - 2023-04-18 at 09 26 30](https://user-images.githubusercontent.com/114543737/232843487-2c9d1706-96f0-4192-8bc5-4d93c5d15edb.png)

nicklbaert:
> Hi @nicklbaert, have you made any changes using the appearance API? Looking at this state using the defaults the text is visible while loading: ![Simulator Screen Shot - iPhone 14 Pro - 2023-04-18 at 09 26 30](https://user-images.githubusercontent.com/114543737/232843487-2c9d1706-96f0-4192-8bc5-4d93c5d15edb.png)

Hi, thanks for the quick reply!

We use these colors:

appearance.colors.background = .white
appearance.colors.primary = UIColor(red: 20/255, green: 20/255, blue: 20/255, alpha: 1)
appearance.colors.textSecondary = UIColor(red: 20/255, green: 20/255, blue: 20/255, alpha: 0.3)
appearance.colors.componentPlaceholderText = UIColor(red: 20/255, green: 20/255, blue: 20/255, alpha: 0.5)
appearance.colors.componentText = UIColor(red: 20/255, green: 20/255, blue: 20/255, alpha: 0.5)
appearance.colors.componentBorder = UIColor(red: 240/255, green: 240/255, blue: 240/255, alpha: 1)
appearance.colors.componentDivider = UIColor(red: 240/255, green: 240/255, blue: 240/255, alpha: 1)
appearance.colors.componentBackground = UIColor(red: 243/255, green: 248/255, blue: 250/247, alpha: 1)

[...]

PR #2492, opened 2023-04-19
Use a calculated disabled color instead of .tertiaryLabel

eurias-stripe:
## Summary
We were hardcoding `.tertiaryLabel` when a field is disabled, this caused issues in some integrations depending on their appearance settings.
Changing to use a version of the color with 60% of the opacity.

## Motivation
https://github.com/stripe/stripe-ios/issues/2485

## Testing
- Visually tested
- Added test

| Enabled | Disabled |
| --- | --- |
| Simulator Screen Shot - iPhone 12 mini - 2023-04-21 at 11 51 51 | Simulator Screen Shot - iPhone 12 mini - 2023-04-21 at 11 51 53 |


yuki-stripe (approved):

Tests/ReferenceImages_64/StripeiOS_Tests.LinkCardEditElementSnapshotTests/testDefault@3x.png
yuki-stripe, line 1: Can we match the old styling more?

eurias-stripe: I could change the opacity modifier to lower than 60%, but I used that because is the modifier we were already using.
One thing to note is that this is for Link, so not super relevant at the moment.

yuki-stripe: Oh, as in this PR doesn't change the 'loading' input text color, it just changes the text color when you edit a card in Link?  

The "Deleted" version here looks much better to me, the "Added" version's color is too similar to the textfield subtitle (e.g. "Card number") cc @jarvis-stripe

eurias-stripe: I changes the color of the text in a disabled state. This particular test is for link edit card, which shows the card in disabled state. 
For the general case, I can just change the opacity to something like 40% of the original, instead of 60%

eurias-stripe: Switched it to 40% of the original color, I think it looks good.

yuki-stripe (approved):

yuki-stripe (approved):

PR #2497, opened 2023-04-20
Pass return URL to STPApplePayContext

eurias-stripe:
## Summary
SetupIntents with auto PMs require a return URL, but we were not passing it for Apple Pay, causing the confirmation to fail.
This was already mitigated on the backend by removing this requirement in https://git.corp.stripe.com/stripe-internal/pay-server/pull/606610, but we still need to pass it on our side.

## Motivation
https://jira.corp.stripe.com/browse/RUN_MOBILESDK-2268

## Testing
Verified that return URL was included in the confirm request.

### DIFF
--- a/StripeApplePay/StripeApplePay/Source/PaymentsCore/API/Models/SetupIntentParams.swift
+++ b/StripeApplePay/StripeApplePay/Source/PaymentsCore/API/Models/SetupIntentParams.swift
@@ -16,14 +16,14 @@ extension StripeAPI {
             clientSecret: String,
             paymentMethodData: StripeAPI.PaymentMethodParams? = nil,
             paymentMethod: String? = nil,
-            returnURL: String? = nil,
+            returnUrl: String? = nil,
             useStripeSdk: Bool? = nil,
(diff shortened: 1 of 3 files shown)


StripeApplePay/StripeApplePay/Source/PaymentsCore/API/Models/SetupIntentParams.swift
```diff
@@ -40,7 +40,7 @@ extension StripeAPI {
         /// The URL to redirect your customer back to after they authenticate or cancel
         /// their payment on the payment method’s app or site.
         /// This should probably be a URL that opens your iOS app.
-        @_spi(STP) public var returnURL: String?
+        @_spi(STP) public var returnUrl: String?
```
eurias-stripe, line 43: I had to rename this because the JSON serializer was converting it to `return_u_r_l` because of the capitalization. It is a private API so it should be safe to change, it wasn't actually used before, otherwise requests would've failed.

porter-stripe (approved):

PR #2512, opened 2023-04-25
Add confirmation method to playground and tests

porter-stripe:
## Summary
- Adds confirmation method to playground so we can test manual confirmation
- Backend was updated to support manual confirmation here https://git.corp.stripe.com/stripe-internal/sandbox-apps/pull/447
- We don't test 3DS2 cards here since we cannot automate tapping through the web view in test mode
- We also don't send the final confirm like we would instruct merchants to do since that is mainly testing backend code at that point

## Motivation
Manual confirmation

## Testing
- New unit tests
- Manually

## Changelog
N/A


yuki-stripe:
> We don't test 3DS2 cards here since we cannot automate tapping through the web view in test mode

Hm have you tried doing something like https://stackoverflow.com/questions/35442972/xcode-ui-test-how-to-find-and-tap-an-element-inside-a-webview

yuki-stripe (approved):

porter-stripe:
I'm not seeing any of the accessibility identifiers set using the accessibility inspector :/

PR #2517, opened 2023-04-27
Add validation to catch blank ephemeral keys in CustomerConfiguration

sfriedman-stripe:
## Summary
Added an assertion to catch blank ephemeral keys when a user instantiates a `CustomerConfiguration`

## Motivation

We currently don't validate that an ephemeral key is non-empty when a
user includes it in the `CustomerConfiguration`. This ends up giving
users a pretty obscure error in the API response, and is difficult to
debug today.

Including this validation will catch this issue early and provide users
a better debugging experience.

As part of this change, I also added a dependency on [CwlPreconditionTesting](https://github.com/mattgallagher/CwlPreconditionTesting).

CwlPreconditionTesting allows us to unit test traps (i.e. fatalError,
assert, precondition). We currently don't have any way to test these in
the repo, so engineers simply don't test our assertions. This will
improve the overall state of our unit tests by encouraging testing of
more code paths.

## Testing

* Added a unit test to test the assertion
* Manually set the ephemeral key to empty in the PaymentSheet playground, and verified that the assertion triggered

## Changelog
- [Added] Improved validation for ephemeral keys


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
```diff
@@ -172,11 +172,13 @@ extension PaymentSheet {
         /// See https://stripe.com/docs/api/customers/object#customer_object-id
         public let id: String
 
-        /// A short-lived token that allows the SDK to access a Customer's payment methods
+        /// A short-lived token that allows the SDK to access a Customer's payment methods. Cannot be empty
```
yuki-stripe, line 175:
```suggestion
        /// A short-lived token that allows the SDK to access a Customer's payment methods. Cannot be empty.
```

StripePaymentSheet/StripePaymentSheetTests/PaymentSheetConfigurationTests.swift
```diff
@@ -0,0 +1,21 @@
...
+import XCTest
+
+final class PaymentSheetConfigurationTests: XCTestCase {
+
+    func test_customerConfigurationInit_assertsWhenEphemeralKeyIsBlank() throws {
+        let exception = catchBadInstruction {
+            _ = PaymentSheet.CustomerConfiguration(id: "foo", ephemeralKeySecret: "")
+        }
+
+        XCTAssertNotNil(exception)
```
yuki-stripe, line 19: 🤩

yuki-stripe (approved):
LG2M, @eurias-stripe can you review the Tuist changes?

sfriedman-stripe:
@yuki-stripe Had to fix a couple tests that I broke. Could you PTAL and make sure I'm not breaking the expected behavior for those tests?

yuki-stripe (approved):

Tuist/ProjectDescriptionHelpers/Tuist+Stripe.swift
```diff
@@ -28,6 +29,8 @@ extension Project {
         ///     If `true`, the OCMock package will be linked.
         ///   - usesStubs: Whether the tests in this target use stubs.
         ///     If `true`, the OHHTTPStubs package will be linked.
+        ///   - usesPreconditionTesting: Whether the tests in this target uses precondition, assert, and/or fatal error testing.
```
eurias-stripe, line 32: Nit:
```suggestion
        ///   - usesPreconditionTesting: Whether the tests in this target use precondition, assert, and/or fatal error testing.
```

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 ### Payments
 * [Fixed] STPPaymentHandler.handleNextAction allows payment methods that are delayed or require further customer action like like SEPA Debit or OXXO.
 
+### PaymentSheet
+* [Added] Improved validation for ephemeral keys
```
eurias-stripe, line 6: Does this need a changelog entry? This is not a user facing change.

In case we do need the entry, tiny nit:
```suggestion
* [Added] Improved validation for ephemeral keys.
```

eurias-stripe (approved):
Finally a way to test assertions!

Tuist/ProjectDescriptionHelpers/Tuist+Stripe.swift
sfriedman-stripe: Oops ty

CHANGELOG.md
sfriedman-stripe: I'm not sure what we normally do/do not include in the changelog. It does improve user-facing error messages that users have complained about in the past, so sorta-kinda user facing?

eurias-stripe: I can go either way with this one. If users have specifically mentioned this then it is worth calling it out.

eurias-stripe (approved):

Issue #2468, opened 2023-04-13
[Feature] Metadata support with Apple Pay


magouyaware:
## Is your feature request related to a problem? Please describe.
We are writing our own SDK (Olo Pay SDK) built on top of the Stripe SDK for processing payments. At Stripe's recommendation we are using this [legacy documentation](https://web.archive.org/web/20190919134715/https://stripe.com/docs/apple-pay) to write our own Apple Pay implementation.

We need to insert metadata when creating payment methods for tracking various things like what version of our Olo Pay SDK was used to create a payment method.

We are unable to do this because this [createPaymentMethod()](https://github.com/stripe/stripe-ios/blob/3655def04f7494d3132ea841da0405a9e89001c4/StripePayments/StripePayments/Source/API%20Bindings/STPAPIClient%2BApplePay.swift#L57-L77) implementation doesn't take metadata as a parameter.

## Describe the solution you'd like
The easiest solution for our specific problem is to add an optional `metadata` parameter to this version of [createPaymentMethod()](https://github.com/stripe/stripe-ios/blob/3655def04f7494d3132ea841da0405a9e89001c4/StripePayments/StripePayments/Source/API%20Bindings/STPAPIClient%2BApplePay.swift#L57-L77). This would change the existing method to look like this, and amounts to changing only 3 lines of code, which I've marked with comments below:

```
@objc(createPaymentMethodWithPayment:metadata:completion:) // CHANGE: Update obj-c signature
public func createPaymentMethod(
    with payment: PKPayment,
    metadata: [String: String]? = nil, // CHANGE: Add optional metadata parameter in a way that doesn't break existing code
    completion: @escaping STPPaymentMethodCompletionBlock
) {
    createToken(with: payment) { token, error in
        if token?.tokenId == nil || error != nil {
            completion(nil, error ?? NSError.stp_genericConnectionError())
        } else {
            let cardParams = STPPaymentMethodCardParams()
            cardParams.token = token?.tokenId
            let billingDetails = STPAPIClient.billingDetails(from: payment)
            let paymentMethodParams = STPPaymentMethodParams(
                card: cardParams,
                billingDetails: billingDetails,
                metadata: metadata // CHANGE: Set metadata using passed in parameter, instead of nil
            )
            self.createPaymentMethod(with: paymentMethodParams, completion: completion)
        }
    }
}
```

## Describe alternatives you've considered
[...]

PR #2518, opened 2023-04-27
Add support for providing metadata with Apple Pay

magouyaware:
## Summary
Added support for providing metadata when creating a payment method with Apple Pay

## Motivation
This [feature request](https://github.com/stripe/stripe-ios/issues/2468)

## Testing
I was not actually able to test the code change. I don't have iOS 16.1 simulators installed for some reason, so I was not able to run the tests. However, the changes are extremely simple and minor. 

NOTE: I'm not saying the code doesn't need to be tested, but just saying the likelihood of tests not passing are, in all likelihood, very minimal, and it will be easy for anyone at Stripe to quickly run the tests and make sure they pass.

## Changelog
I would not consider this a _**notable**_ change. The use case for developers other than myself is likely very small as most other developers would be using STPApplePayContext for Apple Pay payments rather than rolling their own implementation like we are planning on doing.

### DIFF
--- a/StripePayments/StripePayments/Source/API Bindings/STPAPIClient+ApplePay.swift
+++ b/StripePayments/StripePayments/Source/API Bindings/STPAPIClient+ApplePay.swift
@@ -58,6 +58,20 @@ extension STPAPIClient {
         with payment: PKPayment,
         completion: @escaping STPPaymentMethodCompletionBlock
     ) {
+        createPaymentMethod(with: payment, metadata: [:], completion: completion)
+    }
+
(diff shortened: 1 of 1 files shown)


magouyaware:
@porter-stripe I'm not sure why the Bitrise check failed... When I clicked on the details link it didn't give me a whole lot of info to go off of

porter-stripe (approved):
Changes look good, failures on this PR are coming from a failing service we depend on. I'll come back through and re-run CI once it's fixed.

PR #2529, opened 2023-05-01
Refactor Wallet Mode with Swift-y API

davidme-stripe:
## Summary
Replaces the Wallet Mode API with a Swift-first version.

## Motivation
Feedback from API review, continued iteration.

## Testing
CI. Added more Wallet Mode tests.

## Changelog
No, this feature is not yet available.


StripePayments/StripePayments/Source/API Bindings/STPAPIClient+Payments.swift
```diff
@@ -941,6 +941,103 @@ extension STPAPIClient {
...
 
+extension STPAPIClient {
+    @_spi(STP) public func listPaymentMethods(
+        forCustomer customerID: String,
+        using ephemeralKeySecret: String,
+        types: [STPPaymentMethodType] = [.card],
+        completion: @escaping STPPaymentMethodsCompletionBlock
+    ) {
+        let header = authorizationHeader(using: ephemeralKeySecret)
+        // Unfortunately, this API only supports fetching saved pms for one type at a time
```
wooj-stripe, line 952: I think this comment was copy and paste, but I believe that type is used as an optional filter:
https://stripe.com/docs/api/payment_methods/customer_list

`type` seems to be an optional parameter - Seems like at the very least, we should update this comment with a TODO: comment to update our code to fetch all payment methods, and then filter payment methods locally to avoid multiple round trips.

davidme-stripe: Ooh, yeah, the comment made me think this was non-optional. We should probably fetch all PMs and filter them on the client side in PaymentSheet too.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerAdapter/CustomerAdapter.swift
```diff
@@ -0,0 +1,202 @@
...
+    /// Retrieve the last selected payment method for the customer.
+    /// If you are implementing your own <CustomerAdapter>:
+    /// Return a PersistablePaymentMethodOption for the customer's default selected payment method.
+    /// If no default payment method is selected, return nil.
+    func fetchSelectedPaymentMethodOption() async throws -> PersistablePaymentMethodOption?
+
+    /// Returns the a SetupIntent client secret configured to attach a new payment method to a customer.
+    /// If you are implementing your own <CustomerAdapter>:
+    /// Create a SetupIntent on your backend, then return the client secret.
+    func setupIntentClientSecretForCustomerAttach() async throws -> String
```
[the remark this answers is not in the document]
wooj-stripe: Ah! From the standpoint that the provided setupIntent should is used to attach the PM to the customer it does make sense -- if there are any other components that want to re-use the CustomerAdapter, then those components will also need to conform to this behavior.  To your point, I also like that it simplifies the init of SavedPyamentMethodSheet.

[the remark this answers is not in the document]
davidme-stripe: Yep, I'll clean up the comment to make this clearer

Example/PaymentSheet Example/PaymentSheet Example/SavedPaymentMethodSheetTestPlayground.swift
```diff
@@ -141,28 +160,30 @@ class SavedPaymentMethodSheetTestPlayground: UIViewController {
...
-        self.customerContext = customerContext
-        var configuration = SavedPaymentMethodsSheet.Configuration(customerContext: customerContext,
-                                                                   applePayEnabled: applePayEnabled())
-        configuration.createSetupIntentHandler = setupIntentHandler(customerId: customerId)
+        let customerAdapter: StripeCustomerAdapter
+        switch paymentMethodMode {
+        case .setupIntent:
+            customerAdapter = StripeCustomerAdapter(customerEphemeralKeyProvider: {
+                // This should be a block that fetches this from your server
+                .init(customerId: customerId, ephemeralKeySecret: ephemeralKey)
```
wooj-stripe, line 168: To ensure nobody removes `async` from the signature of StripeCustomerAdapter's `customerEphemeralKeyProvider` parameter, this line should probably be:

```
return await self.someFunc(customerId: customerId, ephemeralKey: ephemeralKey)

.... then....

    func someFunction(customerId: String, ephemeralKey: String) async -> CustomerEphemeralKey {
         // This should be a block that fetches this from your server
        return .init(customerId: customerId, ephemeralKeySecret: ephemeralKey)
    }
```
This way, if anyone removes the `async` keyword from the parameter, then this will fail to compile.

Example/PaymentSheet Example/PaymentSheet Example/SavedPaymentMethodSheetTestPlayground.swift
```diff
@@ -200,60 +222,20 @@ extension SavedPaymentMethodSheetTestPlayground {
             self.customerId = customerId
             StripeAPI.defaultPublishableKey = publishableKey
 
-            DispatchQueue.main.async {
+            Task {
                 let configuration = self.savedPaymentMethodSheetConfiguration(customerId: customerId, ephemeralKey: ephemeralKey)
-                self.savedPaymentMethodsSheet = SavedPaymentMethodsSheet(configuration: configuration, delegate: self)
+                self.savedPaymentMethodsSheet = SavedPaymentMethodsSheet(configuration: configuration, customer: self.customerAdapter!)
```
wooj-stripe, line 227: We can get rid of the force unwrap in `self.customerAdapter!` by creating a separate function like:
```
    func customerAdapter(customerId: String, ephemeralKey: String) -> StripeCustomerAdapter? {
        let customerAdapter: StripeCustomerAdapter?
        switch paymentMethodMode {
        case .setupIntent:
            customerAdapter = StripeCustomerAdapter(customerEphemeralKeyProvider: {
                // This should be a block that fetches this from your server
                //.init(customerId: customerId, ephemeralKeySecret: ephemeralKey)
                return await self.someFunction(customerId: customerId, ephemeralKey: ephemeralKey)
            }, setupIntentClientSecretProvider: {
                return try await self.backend.createSetupIntent(customerId: customerId)
            })
        case .createAndAttach:
            customerAdapter = StripeCustomerAdapter(customerEphemeralKeyProvider: {
                // This should be a block that fetches this from your server
                .init(customerId: customerId, ephemeralKeySecret: ephemeralKey)
            })
        }
        return customerAdapter
    }
```
Then....

```
                guard let customerAdapter = self.customerAdapter(customerId: customerId, ephemeralKey:ephemeralKey) else {
                    return
                }
                let configuration = self.savedPaymentMethodSheetConfiguration()
...
```

Example/PaymentSheet Example/PaymentSheet Example/SavedPaymentMethodSheetTestPlayground.swift
```diff
@@ -359,10 +341,9 @@ class SavedPaymentMethodsBackend {
         task.resume()
     }
 
-    func createSetupIntent(customerId: String, completion: @escaping (String?) -> Void) {
-        guard clientSecret == nil else {
-            completion(clientSecret)
-            return
+    func createSetupIntent(customerId: String) async throws -> String {
+        if let clientSecret = clientSecret {
+            return clientSecret
```
wooj-stripe, line 346: Did we change this for testing purposes?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/SavedPaymentMethods/SavedPaymentMethodsSheetError.swift
```diff
@@ -8,31 +8,8 @@ import Foundation
 import StripePayments
 
 @_spi(PrivateBetaSavedPaymentMethodsSheet) public enum SavedPaymentMethodsSheetError: Error {
-
     /// Error while fetching saved payment methods attached to a customer
     case errorFetchingSavedPaymentMethods(Error)
```
wooj-stripe, line 12: Just checking my understanding -- these are errors for terminal states of the sheet closing -- so, there's really only one case -- where we would fail to fetch the saved Payment methods, and then close the sheet?

davidme-stripe: That's the only terminal error state I can think of atm. I need to try to brainstorm a few more.

Example/PaymentSheet Example/PaymentSheet Example/SavedPaymentMethodSheetTestPlayground.swift
davidme-stripe: I think it's okay if a user doesn't treat this as async: They just won't get the benefits of customer ephemeral key auto-refreshing. Our example backend isn't currently set up for it, as it creates a new customer ephemeral key each time and doesn't support fetching an ephemeral key for an arbitrary customer. 🙈

Example/PaymentSheet Example/PaymentSheet Example/SavedPaymentMethodSheetTestPlayground.swift
davidme-stripe: Weird, removed!

Example/PaymentSheet Example/PaymentSheet Example/SavedPaymentMethodSheetTestPlayground.swift
wooj-stripe: I think it's also okay if the user doesn't treat it as async -- i guess my point is more of a nit that if someone accidentally removes the `async` keyword on the init of CustomerAdapter -- e.g.:
from:
```
    public init(customerEphemeralKeyProvider: @escaping () async throws -> CustomerEphemeralKey,
```
to:
```
    public init(customerEphemeralKeyProvider: @escaping () throws -> CustomerEphemeralKey,
```
Our code will continue to compile, but I think that compilation should fail, because there will be some customers who do implement an async keyProvider.  By changing our code to pass in fake "async" key provider, we will have a compile time check that nobody removes the `async` keyword.

StripePayments/StripePayments/Source/API Bindings/Models/PaymentMethods/STPPaymentMethod.swift
```diff
@@ -210,7 +210,7 @@ public class STPPaymentMethod: NSObject, STPAPIResponseDecodable {
 
     // MARK: - STPAPIResponseDecodable
     /// :nodoc:
-    @objc required init(
+    @objc @_spi(STP) public required init(
```
wooj-stripe, line 213: I was trying to figure out why this was needed.  I removed `@_spi(STP) public ` and PaymentSheet Example still compiles -- revert this change?

davidme-stripe: It was for a now-deleted bridge between STPCustomerContext and StripeCustomerAdapter. (STPCustomerContext requires an entire STPPaymentMethod for attach/detach, even though it only uses the ID.) Will delete.

wooj-stripe (approved):
lgtm! thank you so much for doing this!

PR #2530, opened 2023-05-01
Add support for automatic async capture method

porter-stripe:
## Summary
- Adds support for the new capture method - `automatic_async`

## Motivation
Decoupled workflow

## Testing
- Updated unit test

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetIntentConfiguration.swift
```diff
@@ -99,10 +99,14 @@ import Foundation
...
             /// (Default) Stripe automatically captures funds when the customer authorizes the payment.
-            case automatic
+            case automatic = "automatic"
 
             /// Place a hold on the funds when the customer authorizes the payment, but don’t capture the funds until later. (Not all payment methods support this.)
-            case manual
+            case manual = "manual"
+
+            /// Asynchronously capture funds when the customer authorizes the payment.
+            /// - Note: Recommended over `CaptureMethod.automatic` due to improved latency, but may require additional integration changes.
```
yuki-stripe, line 108: I assume you just copied this from somewhere, but this is a frustratingly vague docstring..what additional integration changes??

yuki-stripe (commented):
Looks good but I guess we should hold off on merging this change until it's approved in API Review.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetIntentConfiguration.swift
porter-stripe: Yeah its from here: https://stripe.com/docs/api/payment_intents/object#payment_intent_object-capture_method

They have a link, maybe we can add a seealso and link to: https://stripe.com/docs/payments/payment-intents/asynchronous-capture-automatic-async

PR #2536, opened 2023-05-03
Financial Connections: introduced FinancialConnectionsFont and ClickableLabelNew to support line spacing.

kgaidis-stripe:
## Summary

This is the first, of many PR's, where we want to change the typography in all panes. 

This PR only modifies the Consent Pane (see test plan of all that is modified).

Two new concepts:
- `FinancialConnectionsFont` (combines UIFont and line height)
- `ClickableLabelNew` (a temporary copy-paste of `ClickableLabel` to support the new font type)

Font references:
- https://sail.stripe.me/foundations/typography/
- https://www.figma.com/file/WAFvoybps5uk3FL4tftaee/Sail?type=design&node-id=26-370&mode=design&t=Hd2oiFSiURfCMlmn-0

## Testing

The design/spacing has also been reviewed by design.

[This is what we want to achieve](https://www.figma.com/file/ikSIQS9Lw0qIr3PSeyyrPP/Connections-V2.5?type=design&node-id=1636-89037&mode=design&t=5PkZrVUywIxjEo01-0).

Simulator Screen Shot - iPhone 14 Pro - 2023-05-04 at 16 57 05

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/Helpers/FinancialConnectionsFont.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/Helpers/FinancialConnectionsFont.swift
@@ -1,0 +1,110 @@
+//
+//  FinancialConnectionsFont.swift
+//  StripeFinancialConnections
+//
+//  Created by Krisjanis Gaidis on 5/2/23.
+//
+
+import Foundation
+import UIKit
+
+// A wrapper around `UIFont` that allows us to specify a `lineHeight`.
+// `UIFont` does not support modifying `lineHeight` so this struct
+// helps us to easily pass around font + line height.
+struct FinancialConnectionsFont {
+
+    let uiFont: UIFont
+    let lineHeight: CGFloat
+
+    // An estimated "top padding of the font character"
+    var topPadding: CGFloat {
+        return max(0, ((lineHeight - uiFont.lineHeight) / 2)) + (uiFont.ascender - uiFont.capHeight)
+    }
+
+    enum HeadingToken {
+        case large
+    }
+    static func heading(_ token: HeadingToken) -> FinancialConnectionsFont {
+        let font: UIFont
+        let lineHeight: CGFloat
+        let appleTextStyle: UIFont.TextStyle
+        switch token {
+        case .large:
+            // 24 size / 32 line height / 600 weight
+            font = UIFont.systemFont(ofSize: 24, weight: .bold)
+            lineHeight = 32
+            appleTextStyle = .title2
+        }
+        return .create(font: font, lineHeight: lineHeight, appleTextStyle: appleTextStyle)
+    }
+
+    enum BodyToken {
+        case small
+        case smallEmphasized
+        case medium
+        case mediumEmphasized
+    }
+    static func body(_ token: BodyToken) -> FinancialConnectionsFont {
+        let font: UIFont
+        let lineHeight: CGFloat
+        let appleTextStyle: UIFont.TextStyle
+        switch token {
+        case .small:
+            // 14 size / 20 line height / 400 weight
+            font = UIFont.systemFont(ofSize: 14, weight: .regular)
+            lineHeight = 20
+            appleTextStyle = .footnote
+        case .smallEmphasized:
+            // 14 size / 20 line height / 600 weight
+            font = UIFont.systemFont(ofSize: 14, weight: .semibold)
+            lineHeight = 20
+            appleTextStyle = .footnote
+        case .medium:
+            // 16 size / 24 line height / 400 weight
+            font = UIFont.systemFont(ofSize: 16, weight: .regular)
+            lineHeight = 24
+            appleTextStyle = .body
+        case .mediumEmphasized:
+            // 16 size / 24 line height / 600 weight
+            font = UIFont.systemFont(ofSize: 16, weight: .semibold)
+            lineHeight = 24
+            appleTextStyle = .body
+        }
+        return .create(font: font, lineHeight: lineHeight, appleTextStyle: appleTextStyle)
+    }
+
+    enum LabelToken {
+        case largeEmphasized
+    }
+    static func label(_ token: LabelToken) -> FinancialConnectionsFont {
+        let font: UIFont
+        let lineHeight: CGFloat
+        let appleTextStyle: UIFont.TextStyle
+        switch token {
+        case .largeEmphasized:
+            // 16 size / 24 line height / 600 weight
+            font = UIFont.systemFont(ofSize: 16, weight: .semibold)
+            lineHeight = 24
+            appleTextStyle = .body
+        }
+        return .create(font: font, lineHeight: lineHeight, appleTextStyle: appleTextStyle)
+    }
+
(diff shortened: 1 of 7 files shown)


StripeFinancialConnections/StripeFinancialConnections/Source/Helpers/FinancialConnectionsFont.swift
```diff
@@ -0,0 +1,106 @@
...
+import Foundation
+import UIKit
+
+struct FinancialConnectionsFont {
+
+    let uiFont: UIFont
+    let lineHeight: CGFloat
+
+    var topPadding: CGFloat {
+        return ((lineHeight - uiFont.lineHeight) / 2) + (uiFont.ascender - uiFont.capHeight)
```
vardges-stripe, line 17: Will this be still valid lineHeight is smaller than uiFont's lineHeight? Padding might end up being negative? Maybe we do some abs()?

StripeFinancialConnections/StripeFinancialConnections/Source/Helpers/FinancialConnectionsFont.swift
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    private static func scaleFont(_ font: UIFont, appleTextStyle: UIFont.TextStyle) -> UIFont {
+        let metrics = UIFontMetrics(forTextStyle: appleTextStyle)
+        let scaledFont = metrics.scaledFont(for: font)
+        return scaledFont
+    }
+
+    private static func scaleLineHeight(_ lineHeight: CGFloat, font: UIFont, scaledFont: UIFont) -> CGFloat {
+        return lineHeight * (scaledFont.pointSize / font.pointSize)
```
vardges-stripe, line 104: Are we guaranteed to denominator will never be 0 here?

StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/ClickableLabelNew.swift
```diff
@@ -0,0 +1,174 @@
...
+
+    private func setText(
+        _ text: String,
+        links: [LinkDescriptor]
+    ) {
+        let paragraphStyle = NSMutableParagraphStyle()
+        if alignCenter {
+            paragraphStyle.alignment = .center
+        }
+        paragraphStyle.minimumLineHeight = font.lineHeight
```
vardges-stripe, line 109: Do we need both min and max?

vardges-stripe (approved):
Looks nice! Just some questions

StripeFinancialConnections/StripeFinancialConnections/Source/Helpers/FinancialConnectionsFont.swift
kgaidis-stripe: good catch, thank you, I will max the `(lineHeight - uiFont.lineHeight) / 2)` to 0

StripeFinancialConnections/StripeFinancialConnections/Source/Helpers/FinancialConnectionsFont.swift
kgaidis-stripe: Probably unlikely but worth adding extra safety, thank you!

StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/ClickableLabelNew.swift
kgaidis-stripe: @vardges-stripe Not sure, but the way I look at it is that I only want to set line height. I wish min/max didn't exist. We, so far, plan to scale the line height ourselves. As a result, the thought process is that setting both min/max _is_ like setting `lineHeight` (which is what we want). In other words, this code is trying to turn `minimumLineHeight` + `maximumLineHeight` into a unified `lineHeight` property.

StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/ClickableLabelNew.swift
```diff
@@ -0,0 +1,174 @@
...
+//  Created by Krisjanis Gaidis on 5/2/23.
+//
+
+import Foundation
+import SafariServices
+@_spi(STP) import StripeCore
+@_spi(STP) import StripeUICore
+import UIKit
+
+final class ClickableLabelNew: HitTestView {
```
kgaidis-stripe, line 14: This should be nearly equivalent of `ClickableLabel` (pure copy paste) except for the change from `UIFont` to `FinancialConnectionsFont`

StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/ClickableLabelNew.swift
```diff
@@ -0,0 +1,174 @@
...
+        if alignCenter {
+            paragraphStyle.alignment = .center
+        }
+        paragraphStyle.minimumLineHeight = font.lineHeight
+        paragraphStyle.maximumLineHeight = font.lineHeight
+        let string = NSMutableAttributedString(
+            string: text,
+            attributes: [
+                .paragraphStyle: paragraphStyle,
+                .font: font.uiFont,
```
kgaidis-stripe, line 115: Key danger to look out in this code that we are unwrapping `uiFont`. 

Compiler will happily (but wrongly) accept `FinancialConnectionsFont` instead of what it should be (`UIFont`)

StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/ClickableLabelNew.swift
kgaidis-stripe: The goal of the copy-paste is to create the ability to do a piece-by-piece refactor.

All of these changes will be merged into a feature branch, and when done, we should be able to delete `ClickableLabel`

StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/Button+Extensions.swift
```diff
@@ -11,7 +11,7 @@ import Foundation
 extension Button.Configuration {
     static var financialConnectionsPrimary: Button.Configuration {
         var primaryButtonConfiguration = Button.Configuration.primary()
-        primaryButtonConfiguration.font = .stripeFont(forTextStyle: .bodyEmphasized)
+        primaryButtonConfiguration.font = FinancialConnectionsFont.label(.largeEmphasized).uiFont
```
kgaidis-stripe, line 14: There are cases where we won't be able to set the line height. For example, this Stripe component only supports `UIFont`. This _should_ mostly be fine as these "extra" cases will hopefully only span 1 line (line height for 1 line doesn't matter in some cases...or at least cases where we already define how big the component will be [which we do for buttons]).

kgaidis-stripe:
@vardges-stripe OK I opened up the PR for review (previously it was draft); _insert Kris Norris quote of my PR's get approved before they can be approved_

vardges-stripe (approved):

PR #2550, opened 2023-05-08
[Identity] Improve the DocumentScanner's timeout algorithm

ccen-stripe:
## Summary

Introduce an improved timeout logic in `DocumentCaptureViewController` that when the state transfers from found to non-found, reset the timer.
So that user has more time to complete the flow from scratch again, this logic is similar to the web flow logic.
Note: iOS's `SelfieCaptureViewController` already has a similar timeout reset logic, so the same change is not applied.

Similar Android change: https://github.com/stripe/stripe-android/pull/6688, which applies similar change to both document and selfie

## Motivation

Improve scan success rate
## Testing

- [x] Added tests
- [x] Manually verified


ccen-stripe:

eurias-stripe:
Is there anyone with more context on Identity that can review this?

eurias-stripe (approved):
LGTM but I don't have context on what the logic for this should be.

PR #2565, opened 2023-05-12
Add more snapshots to Wallet Mode, fix presentation issue with Apple Pay

davidme-stripe:
## Summary
* Add more snapshot tests
* Fix an issue where the sheet would not show Apple Pay if Apple Pay was the only available payment method.

## Testing
Adds tests for CI


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/SavedPaymentMethods/SavedPaymentMethodsViewController.swift
```diff
@@ -120,7 +120,7 @@ class SavedPaymentMethodsViewController: UIViewController {
         self.isApplePayEnabled = isApplePayEnabled
         self.spmsCompletion = spmsCompletion
         self.delegate = delegate
-        if savedPaymentMethods.isEmpty {
+        if savedPaymentMethods.isEmpty && !isApplePayEnabled {
```
wooj-stripe, line 123: Great catch, but this means that we need to change line 288:

```
                    return savedPaymentMethods.isEmpty && !isApplePayEnabled ? .close(showAdditionalButton: false) : .back
```

davidme-stripe: Done!

wooj-stripe (approved):
lgtm!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/SavedPaymentMethods/SavedPaymentMethodsViewController.swift
```diff
@@ -120,14 +120,14 @@ class SavedPaymentMethodsViewController: UIViewController {
         self.isApplePayEnabled = isApplePayEnabled
         self.spmsCompletion = spmsCompletion
         self.delegate = delegate
-        if savedPaymentMethods.isEmpty {
+        if !savedPaymentMethods.isEmpty || isApplePayEnabled {
```
wooj-stripe, line 123: We should probably use `shouldShowPaymentMethodCarousel` here?

davidme-stripe: Good point, refactored a bit. We can't call it on `self` before `super.init`, but we can at least share the logic.

wooj-stripe (approved):

wooj-stripe (dismissed):

wooj-stripe (approved):

PR #2567, opened 2023-05-12
Making loading height configurable via init

wooj-stripe:
## Summary
Allow for adjustable height of LoadingViewController to account for shorter height of SavedPaymentMethodsSheet

## Motivation
See summary

## Testing
Manual testing

Before:
screenadjusting

After:
fixed


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
```diff
@@ -247,7 +247,8 @@ public class PaymentSheet {
         let loadingViewController = LoadingViewController(
             delegate: self,
             appearance: configuration.appearance,
-            isTestMode: isTestMode
+            isTestMode: isTestMode,
+            loadingViewHeight: 244
```
sfriedman-stripe, line 251: I don't love that we're hardcoding these numbers. I think this is fine to get the bug fixed because it's not dissimilar from what we're already doing, but can we open a Jira to improve this?

sfriedman-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
davidme-stripe: ++, I also think this number might be different depending on the size class of the phone. Where did the initial value of 244 come from?

wooj-stripe: The value of 244 comes from the constant that was in LoadingVIewController, which was (probably) originally derived from the sum of the height of the views that are expected to show when presenting payment sheet's complete flow:
`navigationBar`
`headerLabel`
`paymentContainerView` (the carousel)
`payButton`

In wallet mode and payment sheet flow controller, we don't show the button at the bottom. This is not a problem for flow controller, because there is no need to show a loadingViewController, because of the nature of the ` PaymentSheet.FlowController.create`, then `confirm()` call with flow controller, but is a problem for wallet mode, because the api is simply `present`.

If we were to open a JIRA and "improve this" is the suggestion to not use a constant?  I'm not sure how we would get around this unless we changed the API.

davidme-stripe: That makes sense! I think ideally the initial height would somehow be calculated from the autolayout constraints of all the subviews, but this seems fine for now.

davidme-stripe (approved):

PR #2569, opened 2023-05-15
Refactor PaymentSheet deferred confirm code 

yuki-stripe:
## Summary
- DRY's all the code that creates confirm parameters for intents into `makePaymentIntent`
- Makes `handleDeferredIntent` use ^ to confirm, rather than calling back to `PaymentSheet.confirm`
- Adds a bunch of tests for this deferred confirm flow

## Motivation
The previous codepath went like `confirm` -> `handleDeferredIntent...` -> `confirm` when confirming client-side.  This let us easily reuse code in the `confirm` method but the recursion was hard to follow and was making it hard to add analytics around server-side/client-side confirm. 

## Testing
- Added a lot of success & failure tests for the deferred confirm flow 
- Added unit tests for `makeIntentParams` helper method.

## Changelog
No user facing changes

### DIFF
--- a/Stripe/StripeiOSTests/PaymentSheet+APITest.swift
+++ b/Stripe/StripeiOSTests/PaymentSheet+APITest.swift
@@ -15,7 +15,7 @@ import XCTest
 @testable@_spi(STP) @_spi(ExperimentalPaymentSheetDecouplingAPI) import StripePaymentSheet
 
 class PaymentSheetAPITest: XCTestCase {
-
+    
     let apiClient = STPAPIClient(publishableKey: STPTestingDefaultPublishableKey)
     lazy var paymentHandler: STPPaymentHandler = {
         return STPPaymentHandler(
@@ -39,7 +39,7 @@ class PaymentSheetAPITest: XCTestCase {
         }
         return config
     }()
-
+    
     lazy var newCardPaymentOption: PaymentSheet.PaymentOption = {
         let cardParams = STPPaymentMethodCardParams()
         cardParams.number = "4242424242424242"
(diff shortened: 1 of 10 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+DeferredAPI.swift
```diff
@@ -12,98 +12,105 @@ import Foundation
...
+        configuration: PaymentSheet.Configuration,
+        intentConfig: PaymentSheet.IntentConfiguration,
+        authenticationContext: STPAuthenticationContext,
+        paymentHandler: STPPaymentHandler,
+        completion: @escaping (PaymentSheetResult) -> Void
+    ) {
         // Hack: Add deferred to analytics product usage as a hack to get it into the payment_user_agent string in the request to create a PaymentMethod
         STPAnalyticsClient.sharedClient.addClass(toProductUsageIfNecessary: IntentConfiguration.self)
-        Task {
+        Task { @MainActor in
```
porter-stripe, line 25: Was this not running on the main thread?

porter-stripe (approved):
this is great!!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+DeferredAPI.swift
yuki-stripe: No, not necessarily.  I don't fully understand what the behavior is - Tasks run on the current actor by default, but I'm not sure what that means if you're not explicitly using `@MainActor` or other actors.  In practice, I observed this Task being called on various non-main queues.  Adding `@MainActor` here ensures the closure executes on the main thread.

Stripe/StripeiOSTests/PaymentSheet+APITest.swift
```diff
@@ -394,6 +394,331 @@ class PaymentSheetAPITest: XCTestCase {
...
+        _testDeferredConfirm(
+            inputPaymentOption: .saved(paymentMethod: insufficient_funds_saved_PM),
+            expectedShouldSavePaymentMethod: false,
+            expectedResult: .failed(error: ExpectedError(errorDescription: "Your card has insufficient funds.")),
+            isPaymentIntent: false, // SetupIntent
+            isServerSideConfirm: true // Server-side confirmation
+        )
+    }
+
+    func _testDeferredConfirm(
```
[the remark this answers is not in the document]
yuki-stripe: 🤷 IDK, it somehow seemed clearer to me to distinguish between the two `merchant_clientSecret` and `clientSecret` vars.

PR #2570, opened 2023-05-16
Add SPMS client side metrics

wooj-stripe:
## Summary
Add client side metrics to spms

## Motivation
Client side metrics to monitor health of spms

## Testing
Added breakpoints to new code and exercised code manually


eurias-stripe (dismissed):
I personally don't like adding helper methods for analytics stuff for each event, but it looks like it is what we have been doing in the past.

eurias-stripe (approved):

PR #2574, opened 2023-05-16
Fix analytics for confirming with setup intent on spms

wooj-stripe:
## Summary
Fixing criteria for success for adding a confirmed payment method in SPMS

## Motivation
The original implementation was wrong.

## Testing
Unfortunately, this was implemented by looking at the code.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/SavedPaymentMethods/SavedPaymentMethodsViewController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/SavedPaymentMethods/SavedPaymentMethodsViewController.swift
@@ -393,14 +393,13 @@ class SavedPaymentMethodsViewController: UIViewController {
                 return
             }
 
-            STPAnalyticsClient.sharedClient.logSPMSAddPaymentMethodViaSetupIntentSuccess()
-
             self.fetchSetupIntent(clientSecret: clientSecret) { result in
                 switch result {
                 case .success(let stpSetupIntent):
                     let setupIntent = Intent.setupIntent(stpSetupIntent)
                     self.confirm(intent: setupIntent, paymentOption: paymentOption)
                 case .failure(let error):
+                    STPAnalyticsClient.sharedClient.logSPMSAddPaymentMethodViaSetupIntentFailure()
                     self.processingInFlight = false
                     self.error = error
                     self.updateUI()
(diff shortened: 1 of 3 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/SavedPaymentMethods/SavedPaymentMethodsViewController.swift
```diff
@@ -414,13 +413,16 @@ class SavedPaymentMethodsViewController: UIViewController {
             self.processingInFlight = false
             switch result {
             case .canceled:
+                STPAnalyticsClient.sharedClient.logSPMSAddPaymentMethodViaSetupIntentFailure()
```
davidme-stripe, line 416: When is the state `.canceled`? If it's just when the user closes the add new payment view, I don't think we should count it as a failure.

wooj-stripe: Taking a step back, and making sure we're on the same page, the entire process contains:
``
customerAdapter.setupIntentClientSecretForCustomerAttach()
self.fetchSetupIntent()
confirmPaymentIntent
  case .cancel:
     //proposed logging of failing `logSPMSAddPaymentMethodViaSetupIntentFailure`
  case  .failed:
    ...
  case .succeeded
    ...
```

So I believe this `canceled` state corresponds to state of the setup intent which is happening in  `STPPaymentHandler`.  How can this happen?  Well, it's unlikely if everything is coded up correctly, but it's possible that when we go to confirm a setup intent, and attempt the handle the next action, the status of the SI could be 'canceled'.  As such, `STPPaymentHandler` complete the action with `canceled`. If we happen to be in this rare state, I would consider this to be a failure because we were unable to attach this paymentMethod to the customer with a (canceled) setup intent.

davidme-stripe: Ah, got it, I misread this as being if the user cancelled the sheet before attempting confirmation. I think it's still a non-error state (e.g. the user got a 3DS2 prompt or redirected to a banking app and decided to not complete the confirmation), but I'm peaceful with either option.

davidme-stripe (dismissed):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/SavedPaymentMethods/SavedPaymentMethodsViewController.swift
wooj-stripe: Ah -- in that case.. hmm maybe failure isn't quite the right thing we want to capture -- I guess we could introduce a canceled metric -- or just not log failure or success.  I'm leaning towards introducing a canceled metric.  Thoughts?

wooj-stripe: chatted offline -- adding canceled metric

davidme-stripe (approved):

PR #2584, opened 2023-05-18
Make currency have a default value of nil in decoupled setup mode

yuki-stripe:
## Summary
Make currency have a default value of nil in decoupled setup mode

## Motivation
It's already optional. No need to make merchants explicitly specify nil.

## Testing
Added a unit test

## Changelog
not breaking


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetIntentConfiguration.swift
```diff
@@ -108,7 +108,7 @@ import Foundation
             /// Use this if your integration creates a SetupIntent
             case setup(
                 /// Three-letter ISO currency code. Optional - setting this ensures only valid payment methods are displayed.
-                currency: String?,
+                currency: String? = nil,
```
yuki-stripe, line 111: cc @tillh-stripe

porter-stripe (approved):

PR #2597, opened 2023-05-24
Fetch payment methods and selectedPaymentMethod before presenting sheet

wooj-stripe:
## Summary
Get paymentMethods and selected payment method at the same time

## Motivation
Awkward behavior of calling the selected payment method multiple times in the collection view

## Testing
A lot of manual testing :( .. need to figure out a way to unit test this.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSheet.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSheet.swift
@@ -154,10 +154,6 @@ internal enum CustomerSheetResult {
     var userCompletion: ((SheetResult) -> Void)?
 }
 
-enum TaskResult<T> {
-    case success(T)
-    case failure(Error)
-}
 extension CustomerSheet {
     func loadPaymentMethodInfo(completion: @escaping (Result<([STPPaymentMethod], PersistablePaymentMethodOption?), Error>) -> Void) {
         Task {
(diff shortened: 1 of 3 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSheet.swift
```diff
@@ -150,15 +154,58 @@ internal enum CustomerSheetResult {
...
-    func loadPaymentMethods(completion: @escaping (Result<[STPPaymentMethod], Error>) -> Void) {
+    func loadPaymentMethodInfo(completion: @escaping (Result<([STPPaymentMethod], PersistablePaymentMethodOption?), Error>) -> Void) {
         Task {
-            do {
-                let paymentMethods = try await customerAdapter.fetchPaymentMethods()
-                let filteredPaymentMethods = paymentMethods.filter { $0.type == .card }
-                completion(.success(filteredPaymentMethods))
-            } catch {
-                completion(.failure(error))
+            await withTaskGroup(of: TaskResult<Any?>.self) { group in
```
davidme-stripe, line 164: Instead of using task groups, could we reduce boilerplate by doing something like:
```
async let paymentMethods = try customerAdapter.fetchPaymentMethods()
async let selectedPaymentMethod = try self.customerAdapter.fetchSelectedPaymentMethodOption()
let (paymentMethods, selectedPaymentMethod) = try await (paymentMethods, selectedPaymentMethod)
// Do all the client-side processing and filtering
```
But I haven't used task groups before, so maybe I'm missing some benefits!

wooj-stripe: Wow, this is really elegant. I think the use of TaskGroups is way overkill.  thanks!

davidme-stripe (approved):

PR #2631, opened 2023-06-05
Fix issue with removing payment method in customer sheet

wooj-stripe:
## Summary
Fixing issue with removing payment methods with customer sheet

## Motivation
Crashing bug

## Testing
Manually tested
Add two cards, then remove the first one, then attempt to remove the second one.

without fix: 
withBug

With Fix:
fixForAssertCrash


davidme-stripe (approved):
Might be a good UI test if you get a chance

PR #2634, opened 2023-06-07
Financial Connections: created a new custom sheet view controller for presenting sheets and integrated it into close sheet

kgaidis-stripe:
## Summary

This PR converts the old `UIAlertView` close pop-up to a "sheet."

In order to do this, we had to start adding a bunch of infra (sorry for the mountain of code in one PR).

⚠️  IMPORTANT NOTE: Note that this is merging into a feature branch. A lot of this code is in-flux and the TODO's are expected. Lots of things will be modified later, but wanted to avoid scope-creep and making the PR even bigger.

## Testing

Somewhat based on [designs](https://www.figma.com/file/wXQ8NJApqSJiLmxxANDRTs/%E2%9C%A8-Connections-v3.0?type=design&node-id=2524-217093&mode=design&t=nHAFOBcDVpMEQUoo-0).

Note: this does not necessarily work well on iPad. Will figure out that later too by checking in with design.

| . | . | . |
| --- | --- | --- |
| Simulator Screen Shot - iPhone 14 Pro - 2023-06-07 at 16 23 56 | Simulator Screen Shot - iPhone 14 Pro - 2023-06-07 at 16 30 55 |  |

https://github.com/stripe/stripe-ios/assets/105514761/b97b8e0f-6fb2-4c3a-bc31-34cd98d6ffc0


StripeFinancialConnections/StripeFinancialConnections/Source/Native/NativeFlowController.swift
```diff
@@ -314,10 +314,8 @@ extension NativeFlowController {
         }
 
         if showConfirmationAlert {
-            CloseConfirmationAlertHandler.present(
-                businessName: dataManager.manifest.businessName,
-                showNetworkingLanguageInConfirmationAlert: showNetworkingLanguageInConfirmationAlert,
-                didSelectOK: {
+            CloseConfirmationSheetViewController.present(
+                didSelectClose: {
```
kgaidis-stripe, line 318: this code shows how we were able to preserve a very simple interface in terms of being able to present a sheet from the core code

StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/CustomSheet/CustomSheetLayoutView.swift
```diff
@@ -0,0 +1,102 @@
...
+@_spi(STP) import StripeUICore
+import UIKit
+
+/// Reusable view that separates panes into two parts:
+/// 1. A scroll view for content
+/// 2. A footer that is "locked" and does not get affected by scroll view
+///
+/// Purposefully NOT a `UIView` subclass because it should only be used via
+/// `addToView` helper function.
+final class CustomSheetLayoutView {
```
kgaidis-stripe, line 18: This is an "infra" component for sheets. 

It allows to easily set a `contentView` and `footerView`, and it embeds this into an automatically-resizable "sheet" with a "handle" on top

StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/CustomSheet/CustomSheetLayoutView.swift
```diff
@@ -0,0 +1,102 @@
...
+    func addTo(view: UIView) {
+        // this function encapsulates an error-prone sequence where we
+        // must add `paneLayoutView` (and all it's subviews) to the `view`
+        // BEFORE we can add a constraint for `UIScrollView` content
+        view.addAndPinSubviewToSafeArea(paneLayoutView)
+        scrollViewContentView?.widthAnchor.constraint(equalTo: view.safeAreaLayoutGuide.widthAnchor).isActive = true
+    }
+}
+
+private final class ResizableScrollView: UIScrollView {
```
kgaidis-stripe, line 56: This helps to creates a `UIScrollView` that will only start scrolling if the content height is greater than available parent height

The implementation is surprisingly simple. Arrived at it from iteration. `UIStackView` auto-magically handles this across multiple iOS versions.

StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/CustomSheet/CustomSheetViewController.swift
```diff
@@ -0,0 +1,206 @@
...
+//  CustomSheetViewController.swift
+//  StripeFinancialConnections
+//
+//  Created by Krisjanis Gaidis on 6/2/23.
+//
+
+import Foundation
+import UIKit
+
+final class CustomSheetCommunicationHelper {
```
kgaidis-stripe, line 11: This is a simple object that facilitates communication between the sheet infra and the view/viewcontroller that manages the sheet. 

It provides a `dismissSheet` function so the "sheet view" can decide to dismiss the sheet programatically.

---

Why design it like this? 

Usually the way the sheet setup works is:
0. ???
1. Create a view for the sheet
2. Present the view as a sheet

The problem is that (1) somehow needs to know about (2) [so (1) can dismiss (2)]. So we create `CustomSheetCommunicationHelper` as part of step (0), and then pass it to both (1) and (2). 

---

Note that I may be able to delete this, but am still debating it (will save for a different PR). Technically, if every sheet is presented from a UIViewController, then calling `self?.dismiss(animated: true)` from the SheetViewController would do the same thing. The problem is may not necessarily be desirable for every sheet to be a UIViewController, but it could at least simplify this + the interface so maybe its worth it.

StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/CustomSheet/CustomSheetViewController.swift
```diff
@@ -0,0 +1,206 @@
...
+    }
+    sheetCommunicationHelper.dismissSheetClosure = { [weak customSheetViewController] in
+        customSheetViewController?.dismiss(animated: true)
+    }
+    invisibleViewController.present(customSheetViewController, animated: true)
+}
+
+// TODO(kgaidis): for iOS 13, we may need to improve sheet background
+@available(iOSApplicationExtension, unavailable)
+private final class CustomSheetViewController: UIViewController, UIGestureRecognizerDelegate {
```
kgaidis-stripe, line 92: I may have to re-write this later because if the sheet is smaller than the size of the window, there is a little bit of a weird "shadow" that is noticeable when presenting or slowly dismissing the sheet. That being said, I should be able to swap it out relatively easily later.

kgaidis-stripe:
closing because there's been change in some plans

PR #2636, opened 2023-06-07
Refactor PaymentSheetFormFactory to reduce duplication w/ customerSheet

wooj-stripe:
## Summary
Refactor PaymentSheetFormFactory to reduce duplication w/ customerSheet

## Motivation
Reduce code duplication and unlock the ability to do billing address collection on customer sheet.

## Testing
Manually smoke tested, but largely relying on existing tests

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactoryConfig.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactoryConfig.swift
@@ -1,0 +1,78 @@
+//
+//  PaymentSheetFormFactoryConfig.swift
+//  StripePaymentSheet
+//
+
+import Foundation
+
+@_spi(STP) import StripePayments
+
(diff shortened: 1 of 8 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactoryConfig.swift
```diff
@@ -0,0 +1,78 @@
...
+        case .customerSheet:
+            return nil // TODO: Pass in adapter.
+        }
+    }
+    var merchantDisplayName: String {
+        switch self {
+        case .paymentSheet(let config):
+            return config.merchantDisplayName
+        case .customerSheet:
+            return ""
```
yuki-stripe, line 27: should this assert too?

wooj-stripe: Sure, it doesn't hurt to add it right now, as a reminder that we'll need to add it when we add support for achv2.

yuki-stripe (approved):

PR #2643, opened 2023-06-09
Refactor Link to use ASWebAuthenticationSession

davidme-stripe:
## Summary
* Delete unused Link code
* Add a Link popup URL generator (not used yet, need backend support)
* Add basic support for opening an ASWebAuthenticationSession from the Link button

## Motivation
Move Link to use ASWebAuthenticationSession.

## Testing
Tested manually, will re-enable CI tests in future patches once backend is working.


StripePaymentSheet/StripePaymentSheet/Source/Helpers/PaymentSheetLinkAccount.swift
```diff
@@ -553,3 +553,20 @@ private extension LinkSettings.FundingSource {
         }
     }
 }
+
+// MARK: UpdatePaymentDetailsParams
+
+struct UpdatePaymentDetailsParams {
```
yuki-stripe, line 559: this seems to be unused, is that expected?

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController.swift
```diff
@@ -1,5 +1,5 @@
 //
-//  PayWithLinkViewController.swift
+//  PayWithLinkWebController.swift
```
yuki-stripe, line 2: Let's rename this file to match

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController.swift
```diff
@@ -134,177 +151,61 @@ final class PayWithLinkViewController: UINavigationController {
...
-
-    private func updateUI() {
-        guard let linkAccount = linkAccount else {
-            if !(rootViewController is SignUpViewController) {
-                setRootViewController(
-                    SignUpViewController(linkAccount: nil, context: context)
-                )
+    var aswas: ASWebAuthenticationSession?
+    /// Defaults to the app's key window
+    func present(over viewController: UIViewController? = nil) {
```
yuki-stripe, line 163: just checking, this needs to be nullable?  Shouldn't we always have a presenting view controller?

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController.swift
```diff
@@ -134,177 +151,61 @@ final class PayWithLinkViewController: UINavigationController {
...
-
-        switch linkAccount.sessionState {
-        case .requiresSignUp:
-            if !(rootViewController is SignUpViewController) {
-                setRootViewController(
-                    SignUpViewController(linkAccount: linkAccount, context: context)
-                )
+            guard let url = url else {
+//              TODO: Log unknown session error, this should never happen
+                return
```
yuki-stripe, line 173: you could combine this and the above error check
```
guard let url = url, error == nil else {
  log(error ?? backup error)
  return
}
```

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController.swift
```diff
@@ -134,177 +151,61 @@ final class PayWithLinkViewController: UINavigationController {
...
-        super.pushViewController(viewController, animated: animated)
-    }
-
-    private func updateUI() {
-        guard let linkAccount = linkAccount else {
-            if !(rootViewController is SignUpViewController) {
-                setRootViewController(
-                    SignUpViewController(linkAccount: nil, context: context)
-                )
+    var aswas: ASWebAuthenticationSession?
```
yuki-stripe, line 161: nit can we give this a full name?

StripePaymentSheet/StripePaymentSheet/Source/Helpers/PaymentSheetLinkAccount.swift
davidme-stripe: This was moved from a deleted file, as it's still used by the API. I *think* it might be used for inline Link enrollment, but we're not sure yet. Planning to do another code deletion pass for unused API bits at the end.

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController.swift
davidme-stripe: Thought I did, good catch

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController.swift
davidme-stripe: Yep, will fix, we don't use it without a VC anywhere.

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkWebController.swift
```diff
@@ -0,0 +1,240 @@
...
+    var bestWindowForPresentation: UIWindow {
+        if let window = presentationVC?.view.window {
+            return window
+        }
+        if let window = UIApplication.shared.windows.first(where: { $0.isKeyWindow }) {
+            return window
+        }
+        return UIApplication.shared.windows.first!
+    }
+    var presentationVC: UIViewController?
```
yuki-stripe, line 84: There's probably a reason we can't do this, but it'd be nice if we could require this in the initializer and make it non-nil

yuki-stripe (approved):

PR #2653, opened 2023-06-13
Lower max number of parallel simulators

wooj-stripe:
## Summary
Lower max number of parallel simulators

## Motivation
Attempting to make builds more stable. We really ought to have bitrise figure out limitation with xcode, but this is a quick band-aid

## Testing
None, other than running CI.


davidme-stripe (approved):

davidme-stripe:
15 minutes isn't great, but I think the real solution here will be to retry when the simulator crashes. I'll ask Bitrise.

PR #2655, opened 2023-06-14
Send `deferred_intent_confirmation_type` in payment event

tillh-stripe:
## Summary

This pull request adds tracking for the `deferred_intent_confirmation_type` similar to https://github.com/stripe/stripe-android/pull/6871 on Android.

## Motivation

Decoupling analytics.

### DIFF
--- a/StripeApplePay/StripeApplePay/Source/ApplePayContext/STPApplePayContext.swift
+++ b/StripeApplePay/StripeApplePay/Source/ApplePayContext/STPApplePayContext.swift
@@ -508,6 +508,7 @@ public class STPApplePayContext: NSObject, PKPaymentAuthorizationControllerDeleg
                 }
 
                 guard clientSecret != STPApplePayContext.COMPLETE_WITHOUT_CONFIRMING_INTENT else {
+                    self.confirmType = STPApplePayContext.ConfirmType.none
                     handleFinalState(.success, nil)
                     return
                 }
--- a/StripeApplePay/StripeApplePay/Source/ApplePayContext/STPApplePayContext.swift
+++ b/StripeApplePay/StripeApplePay/Source/ApplePayContext/STPApplePayContext.swift
@@ -207,7 +207,7 @@ public class STPApplePayContext: NSObject, PKPaymentAuthorizationControllerDeleg
     private weak var delegate: _stpinternal_STPApplePayContextDelegateBase?
     @objc var authorizationController: PKPaymentAuthorizationController?
     @_spi(STP) public var returnUrl: String?
-    
+
     @_spi(STP) @frozen public enum ConfirmType {
         case client
         case server
--- a/StripeApplePay/StripeApplePay/Source/ApplePayContext/STPApplePayContext.swift
+++ b/StripeApplePay/StripeApplePay/Source/ApplePayContext/STPApplePayContext.swift
@@ -214,7 +214,7 @@ public class STPApplePayContext: NSObject, PKPaymentAuthorizationControllerDeleg
         /// The merchant backend used the special string instead of a intent client secret, so we completed the payment without confirming an intent.
         case none
     }
-    ///
+    /// Tracks where the call to confirm the PaymentIntent or SetupIntent happened.
     @_spi(STP) public var confirmType: ConfirmType?
     // Internal state
     private var paymentState: PaymentState = .notStarted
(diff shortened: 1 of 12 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+DeferredAPI.swift
```diff
@@ -20,10 +20,11 @@ extension PaymentSheet {
         paymentHandler: STPPaymentHandler,
         isFlowController: Bool,
         mandateData: STPMandateDataParams? = nil,
-        completion: @escaping (PaymentSheetResult) -> Void
+        completion: @escaping (PaymentSheetResult, STPAnalyticsClient.DeferredIntentConfirmationType?) -> Void
```
tillh-stripe, line 23: This is the main change that has to be propagated through the codebase.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
```diff
@@ -291,7 +309,12 @@ extension PaymentSheet {
             switch confirmOption {
             case .wallet:
                 let linkController = PayWithLinkController(intent: intent, configuration: configuration)
-                linkController.present(completion: completion)
+                linkController.present(
+                    completion: { result in
+                        // TODO(Link): Set intent confirmation type
+                        completion(result, nil)
+                    }
+                )
```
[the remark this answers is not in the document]
davidme-stripe: I think the intent will be confirmed via the usual path using the Link-created payment method.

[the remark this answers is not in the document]
tillh-stripe: Hmm, then I don’t understand what this `completion` block is for. I don’t see it being called in a typical flow. Is that possible?

StripeApplePay/StripeApplePay/Source/ApplePayContext/STPApplePayContext.swift
```diff
@@ -207,6 +207,15 @@ public class STPApplePayContext: NSObject, PKPaymentAuthorizationControllerDeleg
...
     @objc var authorizationController: PKPaymentAuthorizationController?
     @_spi(STP) public var returnUrl: String?
+
+    @_spi(STP) @frozen public enum ConfirmType {
+        case client
+        case server
+        /// The merchant backend used the special string instead of a intent client secret, so we completed the payment without confirming an intent.
+        case none
+    }
+    ///
```
yuki-stripe, line 217:
```suggestion
    /// Tracks where the call to confirm the PaymentIntent or SetupIntent happened.
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
```diff
@@ -26,11 +26,11 @@ extension PaymentSheet {
         paymentHandler: STPPaymentHandler,
         isFlowController: Bool = false,
         paymentMethodID: String? = nil,
-        completion: @escaping (PaymentSheetResult) -> Void
+        completion: @escaping (PaymentSheetResult, STPAnalyticsClient.DeferredIntentConfirmationType?) -> Void
     ) {
         // Translates a STPPaymentHandler result to a PaymentResult
-        let paymentHandlerCompletion: (STPPaymentHandlerActionStatus, NSObject?, NSError?) -> Void = { status, _, error in
-            completion(makePaymentSheetResult(for: status, error: error))
+        let paymentHandlerCompletion: (STPPaymentHandlerActionStatus, STPAnalyticsClient.DeferredIntentConfirmationType?, NSObject?, NSError?) -> Void = { status, deferredIntentConfirmationType, _, error in
```
yuki-stripe, line 32: I think we can delete `NSObject?` from the arguments

yuki-stripe: D we ever pass a non-nil value for the confirmation type?  Seems like we can remove this argument (in which case you can probably revert all the paymentHandlerCompletion changes in this PR, and just pass nil in the line below this one.

yuki-stripe (approved):

PR #2656, opened 2023-06-14
Fix a few AddressElement bugs

porter-stripe:
## Summary
- Always return nil when the user closes the address element
- Fixes a race condition when setting the auto complete tap target. The auto complete line was being init before we set the tap target result in us init'ing the auto complete line with an empty closure.

## Motivation
- https://jira.corp.stripe.com/browse/RUN_MOBILESDK-1940
- https://github.com/stripe/stripe-ios/issues/2658

## Testing
- Manual

## Changelog
See diff

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## x.x.x x-x-x
+### AddressElement
+* [Fixed] A bug that was causing `addressViewControllerDidFinish` to return a non-nil `AddressDetails` when the user closes the AddressElement when default values are provided.
+
 ## 23.9.1 2023-06-12
 ### PaymentSheet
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## x.x.x x-x-x
+### AddressElement
+* [Fixed] A bug that was causing `addressViewControllerDidFinish` to return a non-nil `AddressDetails` when the user closes the AddressElement when default values are provided.
```
```suggestion
* [Fixed] A bug that was causing `addressViewControllerDidFinish` to return a non-nil `AddressDetails` when the user cancels out of the AddressElement when default values are provided.
```

yuki-stripe (approved):

PR #2657, opened 2023-06-14
Create Link launch URLs and ingest Link return URLs

davidme-stripe:
## Summary
* Create Link launch URLs using info available from the PaymentSheet Config and Intent.
* Handle Link return URLs by parsing the returned PaymentMethod.
* Add basic scaffolding to confirm PaymentSheet using the returned PaymentMethod. (Not yet implemented)

## Motivation
Launching the Link popup and handling the return URL.

## Testing
Added tests for the URL generator and parser. Tested manually in the Link test app.


StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkWebController.swift
```diff
@@ -159,21 +159,42 @@ final class PayWithLinkWebController: NSObject, ASWebAuthenticationPresentationC
...
-        let webAuthSession = ASWebAuthenticationSession(url: LinkURLGenerator.url(), callbackURLScheme: "stripesdk") { url, _ in
-            guard let url = url else {
-                //                  TODO: Get analytics logs here from error: Did the user start a session and then reject the non-ephemeralSession dialog? ASWebAuthenticationSession.cancelledLogin error
-                return
+    func present(over viewController: UIViewController? = nil) {
+        Task { @MainActor in
+            // TODO: Log attempt to show web session
+            do {
+                let url = try await LinkURLGenerator.url(configuration: self.context.configuration, intent: self.context.intent)
+                let webAuthSession = ASWebAuthenticationSession(url: url, callbackURLScheme: "link-popup") { [self] url, error in
```
wooj-stripe: It's a bit confusing to use the same name of `url` for both the linkURLGenerator and the url that is being returned in the completion block of ASWebAuthenticationSession.

maybe something like:
```
let generatedURL = try await LinkURLGenerator....
let webAuthSession = ASWebAuthenticationSession(url: generatedUrl,...)
```

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Utils/LinkPopupURLParser.swift
```diff
@@ -0,0 +1,47 @@
...
+//  LinkPopupURLParser.swift
+//  StripePaymentSheet
+//
+
+import Foundation
+import StripePayments
+
+struct LinkResult {
+    enum LinkStatus: String {
+        case complete
```
wooj-stripe, line 11: Feels awkward to have a single case for an enum -- feels like something is missing here?

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Utils/LinkURLGenerator.swift
```diff
@@ -4,36 +4,107 @@
...
+        var currency: String
+        var amount: Int
+    }
+    enum PaymentObjectMode: String, Encodable {
+        case link_payment_method
+        case card_payment_method
+    }
+    var path = "mobile_pay"
+    var integrationType = "mobile"
+    var paymentObject: PaymentObjectMode
```
wooj-stripe, line 29: Just confirming -- in addition to changing everything from var to let we added:
* paymentObject
* stripeAccount
* paymentUserAgent

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Utils/LinkURLGenerator.swift
```diff
@@ -4,36 +4,107 @@
...
-        return URL(string: "https://checkout.link.com/#")!
+    static func linkParams(configuration: PaymentSheet.Configuration, intent: Intent) async throws -> LinkURLParams {
+        guard let publishableKey = configuration.apiClient.publishableKey ?? STPAPIClient.shared.publishableKey else {
+            throw LinkURLGeneratorError.noPublishableKey
+        }
+        guard let merchantCountryCode = intent.countryCode else {
+            throw LinkURLGeneratorError.noCountryCode
+        }
+
+        // US isn't the best default, but we only expect regionCode to be nil in the nil region on a simulator, and we don't want to crash.
```
wooj-stripe, line 51: I think the wording is off:
"we only expect regionCode to be nil in the nil region on a simulator" ?

Also, is there some invalid country code that we could use instead of US?  I feel like if there are errors, then we'll never know about it?  Maybe "--" or worse case like "XX" (hopefully that won't make anything throw on the backend?)

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Utils/LinkURLGenerator.swift
```diff
@@ -4,36 +4,107 @@
...
+        var customerEmail = configuration.defaultBillingDetails.email
+        if customerEmail == nil,
+           let customerID = configuration.customer?.id,
+           let ephemeralKey = configuration.customer?.ephemeralKeySecret,
+           let customer = try? await configuration.apiClient.retrieveCustomer(customerID, using: ephemeralKey)
+        {
+            customerEmail = customer.email
+        }
+
+        let merchantInfo = LinkURLParams.MerchantInfo(businessName: configuration.merchantDisplayName, country: merchantCountryCode)
```
wooj-stripe, line 64: These args will all be base64 encoded, right? So we don't have to worry about non-encodable characters in our URL?

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Utils/LinkURLGenerator.swift
```diff
@@ -4,36 +4,107 @@
...
+        let customerInfo = LinkURLParams.CustomerInfo(country: customerCountryCode, email: customerEmail)
+
+        let paymentInfo: LinkURLParams.PaymentInfo? = {
+            if let currency = intent.currency, let amount = intent.amount {
+                return LinkURLParams.PaymentInfo(currency: currency, amount: amount)
+            }
+            return nil
+        }()
```
wooj-stripe, line 72: Curious -- why not just put this into a private function?

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Utils/LinkURLGenerator.swift
```diff
@@ -4,36 +4,107 @@
...
+    func toURLEncodedBase64() throws -> String {
+        let encodedData = try JSONEncoder().encode(self)
+        return encodedData.base64EncodedString()
     }
 }
+
+enum LinkURLGeneratorError: Error {
+    case urlCreationFailed
+    case noPublishableKey
+    case noCountryCode
```
wooj-stripe, line 109: Maybe change this to noMerchantCountryCode?

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Link/LinkURLGeneratorTests.swift
```diff
@@ -0,0 +1,96 @@
...
+                                   customerInfo: LinkURLParams.CustomerInfo(country: "US", email: "test@example.com"),
+                                   paymentInfo: LinkURLParams.PaymentInfo(currency: "USD", amount: 100),
+                                   experiments: [:],
+                                   flags: [:],
+                                   loggerMetadata: [:],
+                                   locale: Locale.init(identifier: "en_US").toLanguageTag())
+
+    func testURLCreation() {
+        let url = try! LinkURLGenerator.url(params: testParams)
+        XCTAssertEqual(url.absoluteString, "https://checkout.link.com/#eyJsb2dnZXJNZXRhZGF0YSI6e30sImN1c3RvbWVySW5mbyI6eyJjb3VudHJ5IjoiVVMiLCJlbWFpbCI6InRlc3RAZXhhbXBsZS5jb20ifSwicGF5bWVudE9iamVjdCI6ImxpbmtfcGF5bWVudF9tZXRob2QiLCJleHBlcmltZW50cyI6e30sInBheW1lbnRVc2VyQWdlbnQiOiJ0ZXN0IiwibG9jYWxlIjoiZW4tVVMiLCJwYXRoIjoibW9iaWxlX3BheSIsInBheW1lbnRJbmZvIjp7ImN1cnJlbmN5IjoiVVNEIiwiYW1vdW50IjoxMDB9LCJtZXJjaGFudEluZm8iOnsiY291bnRyeSI6IlVTIiwiYnVzaW5lc3NOYW1lIjoiVGVzdCB0ZXN0In0sInB1Ymxpc2hhYmxlS2V5IjoicGtfdGVzdF8xMjMiLCJmbGFncyI6e30sImludGVncmF0aW9uVHlwZSI6Im1vYmlsZSJ9")
```
wooj-stripe, line 31: I'm assuming you converted this back to json and it checks out? :)

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Link/LinkURLGeneratorTests.swift
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    func testURLCreationRegularUnicode() {
+        var params = testParams
+        params.customerInfo.email = "유니코드"
+        _ = try! LinkURLGenerator.url(params: params)
+        // Just make sure it doesn't fail
```
wooj-stripe, line 38: Rather than failing, it's probably safer to hardcode the base64 encoding in this test.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Link/LinkURLGeneratorTests.swift
```diff
@@ -0,0 +1,96 @@
...
+
+        XCTAssertEqual(params, expectedParams)
+    }
+}
+
+extension STPElementsSession {
+    // Should use the one from StripeiOSTests, but we don't have good infrastructure to share these
+    // and we're not using any details from it.
+    static var emptyElementsSession: STPElementsSession {
+        let apiResponse: [String: Any] = ["payment_method_preference": ["ordered_payment_method_types": ["123"],
```
wooj-stripe, line 80: nit, we could use: `"ordered_payment_method_types": []`

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Link/LinkURLGeneratorTests.swift
```diff
@@ -0,0 +1,96 @@
...
+                                                                        "country_code": "US", ] as [String: Any],
```
wooj-stripe, line 81: I changed the country_code to "FR", thinking that currency would come up as "EUR", but it failed ... I don't fully understand why.  Is this expected?

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Utils/LinkURLGenerator.swift
wooj-stripe: yes

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Link/LinkURLGeneratorTests.swift
davidme-stripe: IIRC without that the compiler warns that the type is ambiguous. Easy to work around but "123" works too

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Link/LinkURLGeneratorTests.swift
davidme-stripe: I tried that but it was annoying to update two URLs every time we change parameters, which will be pretty often during development. I kinda trust that the Unicode pipeline works, I don't know how necessary this test is tbh

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Link/LinkURLGeneratorTests.swift
davidme-stripe: Yep, the Link popup can successfully decode it

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Utils/LinkURLGenerator.swift
davidme-stripe: The simulator region is sometimes nil, which causes regionCode to be nil. That's the only situation in which I'd expect a nil region. (Unless someone blew up their `.GlobalPreferences.plist`.)
XX makes sense, though I do think it would cause a backend error. Still better than crashing.

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Utils/LinkURLGenerator.swift
davidme-stripe: Yep. Still TBD, we're still working out the exact shape of this request.

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Utils/LinkPopupURLParser.swift
davidme-stripe: Yeah, we haven't added the other cases yet.

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Utils/LinkURLGenerator.swift
davidme-stripe: Mostly to make it easier to read the construction of this params object.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Link/LinkURLGeneratorTests.swift
davidme-stripe: There's no logic here, it's just an empty STPElementsSession to include some basic values when testing `LinkURLGenerator.linkParams()` above. Changing one value shouldn't modify the other passed-in values.

wooj-stripe (approved):

PR #2669, opened 2023-06-16
[Identity] Support OTP

ccen-stripe:
# Summary

Support OTP screen and additional API endpoints.

See [this](https://docs.google.com/document/d/1MiThhOl-nntB3LFOPlTwUb1kAhqvbO9zPCD1s_gvMZk/edit#heading=h.jcckmejkh2y) internal doc for more details.

Android change: https://github.com/stripe/stripe-android/pull/6815

# Motivation

Support phone verification

## Testing

- [x] Unit
- [x] Snapshot

## Recording
iosRecording


ccen-stripe:

StripeIdentity/StripeIdentity/Source/NativeComponents/Views/PhoneOtpView.swift
```diff
@@ -0,0 +1,162 @@
...
+        stackView.addArrangedSubview(otpBodyLabel)
+        stackView.addArrangedSubview(otpTextField)
+        stackView.addArrangedSubview(otpErrorField)
+        stackView.addArrangedSubview(submittingIndicator)
+
+        NSLayoutConstraint.activate([
+            warningIconView.widthAnchor.constraint(equalToConstant: 16),
+            warningIconView.heightAnchor.constraint(equalToConstant: 16),
+            otpTextField.leadingAnchor.constraint(equalTo: stackView.leadingAnchor),
+            otpTextField.trailingAnchor.constraint(equalTo: stackView.trailingAnchor),
```
kgaidis-stripe, line 94: double-checking: this is a little unexpected for me where `otpTextField` is added _within_ `UIStackView` so usually UIStackView would manage where its placed

what's the reasoning?

StripeIdentity/StripeIdentity/Source/API Bindings/Models/VerificationPage/VerificationPageStaticContentPhoneOtpPage.swift
```diff
@@ -0,0 +1,23 @@
...
+//
+//  Created by Chen Cen on 6/14/23.
+//
+
+import Foundation
+@_spi(STP) import StripeCore
+
+extension StripeAPI {
+
+    struct VerificationPageStaticContentPhoneOtpPage: Decodable, Equatable {
```
kgaidis-stripe, line 13: if these come from the API, I'd double check the nullability (compare it vs. backend) because if something is marked as non-null when it could be null, there would be a decoding failure

StripeIdentity/StripeIdentity/Source/NativeComponents/ViewControllers/PhoneOtpViewController.swift
```diff
@@ -0,0 +1,150 @@
...
+                            return .enabled
+                        case .RequestingOTP:
+                            return .loading
+                        case .RequestingCannotVerify:
+                            return .disabled
+                        }
+                    }(),
+                    isPrimary: false,
+                    didTap: {
+                        self.generateOtp()
```
kgaidis-stripe, line 43: I'd double-check if this is not a retain cycle

So `PhoneOtpViewController retains `IdentityFlowView.ViewModel` and `IdentityFlowView.ViewModel` retains `PhoneOtpViewController`

StripeIdentity/StripeIdentity/Source/NativeComponents/ViewControllers/PhoneOtpViewController.swift
```diff
@@ -0,0 +1,150 @@
...
+                            return .enabled
+                        case .RequestingOTP:
+                            return .disabled
+                        case .RequestingCannotVerify:
+                            return .loading
+                        }
+                    }(),
+                    isPrimary: false,
+                    didTap: {
+                        self.phoneOtpView.configure(with: .RequestingCannotVerify)
```
kgaidis-stripe, line 64: same here with _maybe_ retain cycle:
- https://github.com/stripe/stripe-ios/pull/2669/files#r1244117134

StripeIdentity/StripeIdentity/Source/NativeComponents/ViewControllers/PhoneOtpViewController.swift
```diff
@@ -0,0 +1,150 @@
...
+        phoneOtpContent: StripeAPI.VerificationPageStaticContentPhoneOtpPage,
+        sheetController: VerificationSheetControllerProtocol
+    ) {
+        self.phoneOtpContent = phoneOtpContent
+
+        let bodyText = {
+            if let localPhoneNumber = sheetController.collectedData.phone {
+                return phoneOtpContent.body.replacingOccurrences(of: "&phone_number&", with: localPhoneNumber.phoneNumber!.suffix(4))
+            } else {
+                return phoneOtpContent.body.replacingOccurrences(of: "&phone_number&", with: phoneOtpContent.redactedPhoneNumber!)
```
kgaidis-stripe, line 84: `phoneOtpContent.redactedPhoneNumber!` 

I'd always be careful with using `!` because it means it will crash if its null

just making a note to think about it if OK

StripeIdentity/StripeIdentity/Source/NativeComponents/ViewControllers/PhoneOtpViewController.swift
```diff
@@ -0,0 +1,150 @@
...
+
+    init(
+        phoneOtpContent: StripeAPI.VerificationPageStaticContentPhoneOtpPage,
+        sheetController: VerificationSheetControllerProtocol
+    ) {
+        self.phoneOtpContent = phoneOtpContent
+
+        let bodyText = {
+            if let localPhoneNumber = sheetController.collectedData.phone {
+                return phoneOtpContent.body.replacingOccurrences(of: "&phone_number&", with: localPhoneNumber.phoneNumber!.suffix(4))
```
kgaidis-stripe, line 82: `localPhoneNumber.phoneNumber!` 

I'd always be careful with using `!` because it means it will crash if its null

just making a note to think about it if OK

StripeIdentity/StripeIdentity/Source/NativeComponents/ViewControllers/PhoneOtpViewController.swift
```diff
@@ -0,0 +1,150 @@
...
+                "Phone Verification",
+                "Back button title for returning to the phone verification page"
+            ),
+            viewModel: flowViewModel
+        )
+    }
+}
+
+extension PhoneOtpViewController: PhoneOtpViewDelegate {
+    ///
```
kgaidis-stripe, line 119: leftover comment

StripeIdentity/StripeIdentityTests/Snapshot/PhoneOtpViewControllerSnapshotTest.swift
```diff
@@ -0,0 +1,54 @@
...
+import iOSSnapshotTestCase
+
+@testable import StripeIdentity
+
+final class PhoneOtpViewControllerSnapshotTest: FBSnapshotTestCase {
+
+    override func setUp() {
+        super.setUp()
+
+        //        recordMode = true
```
kgaidis-stripe, line 18: leftover

StripeIdentity/StripeIdentity/Source/NativeComponents/ViewControllers/PhoneOtpViewController.swift
ccen-stripe: updated to `weak self`

StripeIdentity/StripeIdentity/Source/NativeComponents/ViewControllers/PhoneOtpViewController.swift
ccen-stripe: updated to `weak self`

StripeIdentity/StripeIdentity/Source/NativeComponents/ViewControllers/PhoneOtpViewController.swift
ccen-stripe: removed

StripeIdentity/StripeIdentityTests/Snapshot/PhoneOtpViewControllerSnapshotTest.swift
ccen-stripe: this is a common pattern for our snapshot test, please see [this](https://github.com/search?q=repo%3Astripe%2Fstripe-ios%20recordMode%20%3D%20true&type=code) for example

StripeIdentity/StripeIdentity/Source/NativeComponents/Views/PhoneOtpView.swift
kgaidis-stripe: @ccen-stripe 

yea sometimes layout like this gets confusing and easiest is to try a bunch of things

another way maybe is to put the error view in another stack view (that's centered)

but if this works go for it! I'm just not 100% sure on whether this could cause some layout issues (...because UIStackView is trying to do its own layout, yet we add additional constraints on top, which maybe conflict); worth also checking different iOS versions just-in-case

StripeIdentity/StripeIdentity/Source/NativeComponents/ViewControllers/PhoneOtpViewController.swift
ccen-stripe: changed to `??` and added comments and provided a default value `""` - this should never happen but would prevent it from crashing

StripeIdentity/StripeIdentity/Source/NativeComponents/ViewControllers/PhoneOtpViewController.swift
ccen-stripe: changed to `??` and added comments and provided a default value `""` - this should never happen but would prevent it from crashing

StripeIdentity/StripeIdentity/Source/API Bindings/Models/VerificationPage/VerificationPageStaticContentPhoneOtpPage.swift
ccen-stripe: Just made [this](https://github.com/stripe/stripe-ios/compare/093440ede3d27c146687811431ad1626d0b51f5e..3f9f13b99cb843a2db640e037c5aed71bee44b23#diff-98c82b64f3dd675cafde9c2bf4791a5ba71983aa93be0e5e5d152455465e018c) change to guard the field to be non nil, and report an error otherwise

StripeIdentity/StripeIdentity/Source/NativeComponents/Views/PhoneOtpView.swift
ccen-stripe: `otpErrorField` is already a horizontal stackView 😄 , there might be some config to make it work as expected, but will stick to this one as it's tested thoroughly

kgaidis-stripe (approved):
👍

PR #2682, opened 2023-06-22
Fix luxe specs not being loaded in the decoupled flow

yuki-stripe:
## Summary
- Fix luxe specs not being loaded in the decoupled flow
- Also fixed not reporting an error if the luxe specs are not present

## Impact
There are only 3 LUXE PMs: 
- revolut_pay
- mobilepay
- zip

These all have the form specs hardcoded in the SDK already, they're just missing the icons.  So the net effect of this bug is just that the icons are blank.

## Motivation
I just happened to come across this bug while looking at something else.

## Testing
Manually tested. We need to overhaul the loader and the "ElementsSession response hidden inside a PaymentIntent" hack before this can be reasonably unit tested. 

## Changelog
I highly doubt there are any users affected.


porter-stripe (dismissed):
Definitely shows the usefulness of enums and the exhaustive switches as this would have been caught by the complier when we added `deferredIntent`.

porter-stripe (approved):

PR #2684, opened 2023-06-23
Repurposes LinkPaymentViewController for Instant Debits

vardges-stripe:
## Summary

Refactors LinkPaymentViewController to be used for Instant Debits Only flow.
## Motivation

We want this dedicated LinkOnly entry point to be used for instant debits only flow on mobile. The flow calls into a hosted webflow link generation API, gets the link, opens in a secure webview. Upon completion parses URL params. Uses the returned payment method ID to confirm the payment.
## Testing

https://github.com/stripe/stripe-ios/assets/92807969/18fafb4b-efe4-4de5-a226-4fef5df302d5

Tested against my devbox. Will follow up with the playground entry point to test this.

### DIFF
--- a/Stripe/StripeiOSTests/STPAPIClient+LinkAccountSessionTest.swift
+++ b/Stripe/StripeiOSTests/STPAPIClient+LinkAccountSessionTest.swift
@@ -5,7 +5,7 @@
 //  Created by Yuki Tokuhiro on 4/26/23.
 //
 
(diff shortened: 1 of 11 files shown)


davidme-stripe (approved):
I'm not super familiar with this code, but this all looks fine to me!

davidme-stripe (approved):

PR #2686, opened 2023-06-23
Remove shipping address requirement for affirm

yuki-stripe:
## Summary
Remove shipping address requirement for affirm

## Motivation
Shipping is not actually required:
> We recommend passing [shipping](https://stripe.com/docs/api/payment_intents/create#create_payment_intent-shipping) and [billing](https://stripe.com/docs/api/payment_methods/create#create_payment_method-billing_details) details to improve conversion rates, although these are not required.
https://stripe.com/docs/payments/affirm/accept-a-payment?platform=web

https://jira.corp.stripe.com/browse/RUN_BNPL-2333

## Testing
Updated E2E affirm test

### PaymentSheet
* [Fixed] Affirm no longer requires shipping details.

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
@@ -386,6 +386,11 @@ extension PlaygroundController {
             self.clientSecret = json["intentClientSecret"]
             self.ephemeralKey = json["customerEphemeralKeySecret"]
             self.customerID = json["customerId"]
+            print(self.clientSecret)
+            print(self.customerID)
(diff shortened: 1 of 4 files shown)


Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
```diff
@@ -386,6 +386,11 @@ extension PlaygroundController {
             self.clientSecret = json["intentClientSecret"]
             self.ephemeralKey = json["customerEphemeralKeySecret"]
             self.customerID = json["customerId"]
+            print(self.clientSecret)
```
porter-stripe, line 389: Lets remove the prints

porter-stripe (changes requested):

Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
yuki-stripe: oh whoops! this was some testing for an unrelated run task

PR #2687, opened 2023-06-23
Unpublicize PaymentSheet.PaymentMethodType - making it public was a typo

yuki-stripe:
## Motivation
This type is not useful to users and was mistakenly made public.

## Changelog
I'm opting not to add anything to the CHANGELOG because I'm guessing this has zero usage.  This type is not useful to users and was mistakenly made public.


wooj-stripe (approved):
lgtm!

PR #2705, opened 2023-07-04
added option to select PLN currency

fionnbarrett-stripe:
## Summary

Added option PLN to payment sheet example
## Motivation

Adding support for PLN based payment methods like Blik to Paysheet.


egor-stripe (changes requested):
AFAI the backend example app that PaymentSheet playground connects to doesn't yet support `PLN` currency. Can you please double check if this works without the backend change?

fionnbarrett-stripe:
the backend example app now supports `PLN` currency. payment methods using PLN appear to be fully working now in the example app.

fionnbarrett-stripe (commented):
the backend example app now supports `PLN` currency. payment methods using PLN appear to be fully working now in the example app.

egor-stripe (approved):

porter-stripe (approved):

PR #2711, opened 2023-07-05
Mock out sources test for wechat pay

wooj-stripe:
## Summary
Mock out sources test for wechat pay

## Motivation
Wechat is being a bit more strict with testing in production

## Testing
Executed on local machine

### DIFF
--- a/Stripe/StripeiOSTests/STPSourceFunctionalTest.m
+++ b/Stripe/StripeiOSTests/STPSourceFunctionalTest.m
@@ -625,11 +625,37 @@
     [self waitForExpectationsWithTimeout:TestConstants.STPTestingNetworkRequestTimeout handler:nil];
 }
 
-// 7/5/2023: Unfortunately we have to mock this test out because Wechat is strict about the usage of test merchants
-// in production.  We were previously using:
-// STPAPIClient *client = [[STPAPIClient alloc] initWithPublishableKey:@"pk_live_L4KL0pF017Jgv9hBaWzk4xoB"];
-// However, unable to continue with running this test live in production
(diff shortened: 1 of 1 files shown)


yuki-stripe:
Can we use a test publishable key instead of a live one?

Stripe/StripeiOSTests/STPSourceFunctionalTest.m
```diff
@@ -623,18 +624,28 @@ - (void)testCreateSource_klarna {
...
 
-    STPAPIClient *client = [[STPAPIClient alloc] initWithPublishableKey:@"pk_live_L4KL0pF017Jgv9hBaWzk4xoB"];
+    id mockClient = OCMClassMock([STPAPIClient class]);
+    OCMStub([mockClient createSourceWithParams:[OCMArg any]
+                                    completion:[OCMArg any]])
+    .andDo(^(NSInvocation *invocation) {
+        void (^completion)(STPSource *data, NSError *error);
+        [invocation getArgument:&completion atIndex:3];
+        completion(STPFixtures.weChatPaySource, nil);
+    });
```
yuki-stripe, line 645: I ask about testmode b/c if we mock this out I'm not sure if this ends up testing enough of our code.  Maybe we can run this in testmode and comment out the failing assert(s) to unblock for now.

wooj-stripe: Hmmm -- How about we have both:  one test which has the mocked response which tests that we can properly deserialize `weChatAppURL`, and another non-mocked test using the publishable key?

yuki-stripe (approved):

PR #2718, opened 2023-07-10
Adding new PaymentSheet FormElement

fionnbarrett-stripe:
## Summary
Adding new PaymentSheet FormElement for Blik payment method. I have not added Blik to [supportedPaymentMethods](https://github.com/stripe/stripe-ios/blob/0355a45caaf0ce9dd03c6144efda17ea38d6a0be/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet%2BPaymentMethodAvailability.swift#L21) yet.

## Motivation
Adding payment method Blik to mobile PaymentSheet.

## Testing

- [x] Form Validation Tests

PR #2719, opened 2023-07-10
Exposing STPConfirmPaymentMethodOption on IntentConfirmParams

fionnbarrett-stripe:
## Summary

Exposing `STPConfirmPaymentMethodOption` on `IntentConfirmParams`  for use in `PaymentMethodElementWrapper`.

## Motivation
Laying the ground work for #2718 as the stripe api expects `payment_method_options[blik][code]` to contain the one-time Blik code

## Testing

Old tests still work I believe


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/IntentConfirmParams.swift
```diff
@@ -26,6 +26,7 @@ class IntentConfirmParams {
 
     let paymentMethodParams: STPPaymentMethodParams
     let paymentMethodType: PaymentSheet.PaymentMethodType
+    let paymentMethodOptions: STPConfirmPaymentMethodOptions
```
egor-stripe, line 29: Nit: Let's maybe rename to `confirmPaymentMethodOptions` for naming consistency with other two?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
```diff
@@ -390,7 +395,7 @@ extension PaymentSheet {
             }
         }
 
-        let options = STPConfirmPaymentMethodOptions()
+        let options = params.paymentMethodOptions ?? STPConfirmPaymentMethodOptions()
```
egor-stripe, line 398: Given that we initialise this variable in constructor do we need nil-coalescing here?

fionnbarrett-stripe: I think this is necessary as `paymentMethodOptions` [is defined as optional](https://github.com/stripe/stripe-ios/blob/c2bb1fb7c8795f3e3754b81c3258d7791a204c02/StripePayments/StripePayments/Source/API%20Bindings/Models/PaymentIntents/STPPaymentIntentParams.swift#L148C75-L149C1) . Also I believe it could be nil if the `.saved` case in the switch block above it is used.

egor-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/IntentConfirmParams.swift
```diff
@@ -26,6 +26,7 @@ class IntentConfirmParams {
 
     let paymentMethodParams: STPPaymentMethodParams
     let paymentMethodType: PaymentSheet.PaymentMethodType
+    let ConfirmPaymentMethodOptions: STPConfirmPaymentMethodOptions
```
porter-stripe, line 29: Rename this to use camel case. `confirmPaymentMethodOptions`

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
```diff
@@ -365,13 +369,12 @@ extension PaymentSheet {
     enum ConfirmPaymentMethodType {
         case saved(STPPaymentMethod)
         /// - paymentMethod: Pass this if you created a PaymentMethod already (e.g. for the deferred flow).
-        case new(params: STPPaymentMethodParams, paymentMethod: STPPaymentMethod? = nil, shouldSave: Bool)
-
+        case new(params: STPPaymentMethodParams, options: STPConfirmPaymentMethodOptions, paymentMethod: STPPaymentMethod? = nil, shouldSave: Bool)
```
porter-stripe, line 372: Nit: can we rename options to `paymentOptions` for a little bit more clarity, and update all uses of just `options` to `paymentOptions`.

porter-stripe (approved):

PR #2721, opened 2023-07-10
Remove checks that prevents ApplePay to onboard new users

Buratti:
## Summary
Changed a simple check in the conditional initialiser of STPApplePayContext.

## Motivation
Whenever the user does not have a card registered for ApplePay, the conditional initialiser fails, causing no ApplePay sheet to be presented. ApplePay sheet does support a built-in onboarding flow to invite the user to add a new card. 

## Testing
I have removed all the cards saved in my phone and made a payment. The sheet appears as expected, prompting the user to add a new card, once the card has been added, the user can confirm the payment without ever leaving the sheet. Once the payment has been confirmed, the flow resumes as normal.

## Changelog
    - [Fixed] ApplePay sheet prompts the user to add a card if none is registered.

### DIFF
--- a/StripeApplePay/StripeApplePay/Source/ApplePayContext/STPApplePayContext.swift
+++ b/StripeApplePay/StripeApplePay/Source/ApplePayContext/STPApplePayContext.swift
@@ -107,7 +107,11 @@ public class STPApplePayContext: NSObject, PKPaymentAuthorizationControllerDeleg
         delegate: _stpinternal_STPApplePayContextDelegateBase?
     ) {
         STPAnalyticsClient.sharedClient.addClass(toProductUsageIfNecessary: STPApplePayContext.self)
-        if !StripeAPI.canSubmitPaymentRequest(paymentRequest) {
+        if paymentRequest.merchantIdentifier.isEmpty {
+            return nil
+        }
+        // "In versions of iOS prior to version 12.0 and watchOS prior to version 5.0, the amount of the grand total must be greater than zero."
+        if !(paymentRequest.paymentSummaryItems.last?.amount.floatValue ?? 0.0 >= 0) {
             return nil
         }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
 ## X.Y.Z 2023-XX-YY
+### Apple Pay
+* [Fixed] ApplePay sheet prompts the user to add a card if none is registered.
 ### Payments
 * [Fixed] A bug where `mandate_data` was not being properly attached to PayPal SetupIntent's.
 ### PaymentSheet


StripeApplePay/StripeApplePay/Source/ApplePayContext/STPApplePayContext.swift
```diff
@@ -107,7 +107,11 @@ public class STPApplePayContext: NSObject, PKPaymentAuthorizationControllerDeleg
         delegate: _stpinternal_STPApplePayContextDelegateBase?
     ) {
         STPAnalyticsClient.sharedClient.addClass(toProductUsageIfNecessary: STPApplePayContext.self)
-        if !StripeAPI.canSubmitPaymentRequest(paymentRequest) {
+        if paymentRequest.merchantIdentifier.isEmpty {
+            return nil
+        }
+        // "In versions of iOS prior to version 12.0 and watchOS prior to version 5.0, the amount of the grand total must be greater than zero."
+        if !(paymentRequest.paymentSummaryItems.last?.amount.floatValue ?? 0.0 >= 0) {
```
sfriedman-stripe, line 114: We only support iOS 13 and above, so I don't think we need this check

sfriedman-stripe (commented):
Will let @yuki-stripe comment on the pull request overall, but wanted to leave one comment. And thank you for bringing this up!

StripeApplePay/StripeApplePay/Source/ApplePayContext/STPApplePayContext.swift
Buratti: Hi @sfriedman-stripe. Thank you for the feedback.
The comment you indicated comes from the method which was previously invoked, which you can find here: https://github.com/stripe/stripe-ios/blob/0355a45caaf0ce9dd03c6144efda17ea38d6a0be/StripeCore/StripeCore/Source/API%20Bindings/StripeAPI.swift#L86

I can remove it if you'd like, no problem!

Buratti:
After double-checking, it looks like the behaviour I mentioned is only available for devices running iOS15 and above. [WWDC21](https://developer.apple.com/wwdc21/10092?time=326).

I also noticed that the method `canSubmitPaymentRequest(_:)` was only ever referenced by the STPApplePayContext, so I'll revert the deletion in my first commit. Perhaps a good approach would be to add a conditional check based on the iOS version, allowing the ApplePay sheet to handle user onboarding for iOS 15 and above, while falling back to the existing behavior for iOS 14 and 13?

yuki-stripe:
Thank you for the PR @Buratti and sorry for the delayed response!  Agree it would be useful to support this scenario. I think there's a few things we'll need to think about before making this change.  By removing the call to `PKPaymentAuthorizationController.canMakePayments(usingNetworks:`, we lose some checks:
1. if the user already has saved cards that can make payments with the provided networks.
2. if the hardware supports Apple Pay
3. if parental controls allow Apple Pay. 

It think we'd need to see what the Apple Pay experience if (2) or (3) is not true, and decide if we are okay to present Apple Pay in these cases. Removing these checks in the STPApplePayContext initializer might also be an unexpected change for users relying on those checks, so we might want to make this opt-in somehow.

yuki-stripe:
> Thank you for the PR @Buratti and sorry for the delayed response! Agree it would be useful to support this scenario. I think there's a few things we'll need to think about before making this change. By removing the call to `PKPaymentAuthorizationController.canMakePayments(usingNetworks:`, we lose some checks:
> 
> 1. if the user already has saved cards that can make payments with the provided networks.
> 2. if the hardware supports Apple Pay
> 3. if parental controls allow Apple Pay.
> 
> It think we'd need to see what the Apple Pay experience if (2) or (3) is not true, and decide if we are okay to present Apple Pay in these cases. Removing these checks in the STPApplePayContext initializer might also be an unexpected change for users relying on those checks, so we might want to make this opt-in somehow.

I've since realized `PKPaymentAuthorizationController.canMakePayments()` does what we want - it's the same as `canMakePayments(usingNetworks:)` except that it doesn't require the user to have saved cards.  I've updated `STPApplePayContext` accordingly so it now allows Apple Pay if a user can onboard in the sheet like you've suggested here.  It will go out in the next release - probably next Monday.

Thank you again for bringing this up!

PR #2725, opened 2023-07-10
Link: Fix returning user behavior

davidme-stripe:
## Summary
* If a user has already signed up for Link and signs up again, no-op and log an analytic event. This is the behavior we've gaveled on for private beta.
* When opening the popup, use the email address from the Link button if available. This will include the "lastLinkEmail" from the keychain.
* Don't persist the ConsumerSession auth token on successful signup. Only persist the email address.
* Clean up unused 2FA-related analytics events.

## Motivation
Getting Link ready for private beta.

## Testing
In PaymentSheet Example, will re-enable tests once the service is ready.


StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Extensions/STPAnalyticsClient+Link.swift
```diff
@@ -36,28 +36,6 @@ extension STPAnalyticsClient {
...
-        self.logPaymentSheetEvent(event: .link2FAComplete)
-    }
-
-    func logLink2FAFailure() {
-        self.logPaymentSheetEvent(event: .link2FAFailure)
-    }
-
-    func logLink2FACancel() {
-        self.logPaymentSheetEvent(event: .link2FACancel)
-    }
```
wooj-stripe, line 59: Removing these means that.. i guess we can go into STPAnalyticEvent.swift and remove the following:
```
    // MARK: - Link 2FA
    case link2FAStart = "link.2fa.start"
    case link2FAStartFailure = "link.2fa.start_failure"
    case link2FAComplete = "link.2fa.complete"
    case link2FACancel = "link.2fa.cancel"
    case link2FAFailure = "link.2fa.failure"
```

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Utils/LinkURLGenerator.swift
```diff
@@ -51,8 +51,14 @@ class LinkURLGenerator {
...
-        var customerEmail = configuration.defaultBillingDetails.email
+        // Get email from the previously fetched account in the Link button, or the billing details, or the Customer object
+        var customerEmail = LinkAccountContext.shared.account?.email
+
+        if customerEmail == nil,
+           let defaultBillingEmail = configuration.defaultBillingDetails.email {
+            customerEmail = defaultBillingEmail
+        }
+
         if customerEmail == nil,
```
wooj-stripe, line 62: nit: it would be nice to encapsulate this failover logic into a single function, but i'm not going to block a ship it for this

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift
```diff
@@ -118,9 +118,7 @@ final class PaymentSheetLoader {
             }
         }
 
-        if linkAccountService.hasSessionCookie {
-            return try await lookUpConsumerSession(email: nil)
-        } else if let email = linkAccountService.getLastSignUpEmail() {
+        if let email = linkAccountService.getLastSignUpEmail() {
```
wooj-stripe, line 121: I'm probably overthinking this, but do we need to be concerned with any cases where lastSignUpEmail's value is different than the email on the defaultBillingDetails.email.  The code is clear in that the lastSignUpEmail is prioritized, but I'm guessing that "could" fail technically, and we may want to retry with the defaultBillingDetails.email (if it happens to be a different value) ?

davidme-stripe: Possibly, yeah. Not sure if I want to keep making network requests in that case, as these delay PaymentSheet's load and the failure case here is showing a generic Link button.

Though I *just* found out that we're about to change this behavior again! I'll draft this for now.

davidme-stripe:
@wooj-stripe Okay, ready again.

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkWebController.swift
```diff
@@ -199,19 +202,25 @@ final class PayWithLinkWebController: NSObject, ASWebAuthenticationPresentationC
...
+
                 STPAnalyticsClient.sharedClient.logLinkPopupSuccess(sessionType: self.context.intent.linkPopupWebviewOption)
                 self.payWithLinkDelegate?.payWithLinkWebControllerDidComplete(self, intent: self.context.intent, with: paymentOption)
             case .logout:
-                // We don't store any Link session information locally at the moment. Treat this as a cancel.
-                self.canceledWithoutError()
+                // Delete the account information
+                LinkAccountService.defaultCookieStore.clear()
+                STPAnalyticsClient.sharedClient.logLinkPopupLogout(sessionType: self.context.intent.linkPopupWebviewOption)
+                self.payWithLinkDelegate?.payWithLinkWebControllerDidCancel(self)
```
wooj-stripe, line 223: Nit: Not sure if we care, but in this case of logging out we are calling:
```
clear()
loganalytic
didCancel()
```
but in canceledWithoutError():
```
analytic
clear()
didCancel()
```

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Services/LinkAccountService.swift
```diff
@@ -113,4 +113,47 @@ final class LinkAccountService: LinkAccountServiceProtocol {
         return cookieStore.read(key: .lastSignupEmail)
     }
 
+    func getLastPMDetails() -> LinkPMDisplayDetails? {
```
wooj-stripe, line 116: nit: all of these could be guard let's.

wooj-stripe (approved):

PR #2726, opened 2023-07-11
Add achv2 support to CustomerSheet

wooj-stripe:
## Summary
* API for adding ACHv2 support to Customer Sheet (`paymentMethodTypes`)
* Adds scaffolding for achv2, but leaving it disabled as it is pending legal review
* Reviewing the diff for `handleOverrideAction` `addPaymentOption` will be difficult by looking purely at the diff.  I recommend checking out the code and viewing it without the diff.

## Motivation
achv2 support

## Testing
Added tests


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerAddPaymentMethodViewController.swift
```diff
@@ -31,8 +37,50 @@ class CustomerAddPaymentMethodViewController: UIViewController {
...
+        let paymentMethodElement = makeElement(for: selectedPaymentMethodType)
+        if let usBankAccountPaymentMethodElement = paymentMethodElement as? USBankAccountPaymentMethodElement {
+            usBankAccountPaymentMethodElement.presentingViewControllerDelegate = self
+        } else {
+            assertionFailure("Wrong type for usBankAccountFormElement")
+        }
+        return paymentMethodElement as? USBankAccountPaymentMethodElement
+    }()
+
+    var overrideActionButtonBehavior: OverrideableBuyButtonBehavior? {
```
davidme-stripe, line 52: nit: Why is the `.LinkUSBankAccount` state capitalized?

I also don't understand why there's only one state for this enum. Not sure if I totally understand the flow, are we overriding the custom "Confirm" button to just say "Continue" and open the bank account linking controller? Do we expect future other states for this button?

davidme-stripe (approved):
Looks good overall, one comment on the CTA override

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerAddPaymentMethodViewController.swift
wooj-stripe: Why is it capitalized? Copy and pasted from AddPaymentMethodViewController.

Why is there only one state.  From what I remember, because when we implemented USBankAccount for paymentSheet, we thought there would be other use cases for overriding behavior.  Here we are ~1.5 years later and there's still only one use case for overriding the button. I prefer consistency at this point, and refactoring both to remove the enum seems more trouble than it's worth.

davidme-stripe (approved):

PR #2728, opened 2023-07-11
Add analytic events for PaymentSheet initialization health

porter-stripe:
## Summary
- Adds in 4 new payment sheet analytic events for measuring the health of PaymentSheet initialization
- Sends errors in two new columns using the new domains and error codes: https://github.com/stripe/stripe-ios/pull/2729

## Motivation

## Testing
- Manual, by verifying logs and payloads

## Changelog
N/A


StripeCore/StripeCore/Source/Analytics/STPAnalyticEvent.swift
```diff
@@ -166,4 +166,10 @@ import Foundation
 
     // MARK: - PaymentSheet Force Success
     case paymentSheetForceSuccess = "mc_force_success"
+
+    // MARK: - PaymentSheet initialization
+    case paymentSheetStartedLoading = "mc_started_loading"
+    case paymentSheetDismissed = "mc_dismiss"
+    case paymentSheetLoaded = "mc_load_success"
+    case paymentSheetLoadFailed = "mc_load_failed"
```
yuki-stripe, line 174: Naming nits:
- Can we use the same grammar consistently for these?  e.g. "mc_load_started", "mc_load_succeeded", "mc_load_failed"
- Can we match the case names with the event names?  e.g. Instead of `paymentSheetLoaded = "mc_load_success"`, do `paymentSheetLoadSuccess = "mc_load_success"`

StripeCore/StripeCore/Source/Analytics/STPAnalyticEvent.swift
```diff
@@ -166,4 +166,10 @@ import Foundation
 
     // MARK: - PaymentSheet Force Success
     case paymentSheetForceSuccess = "mc_force_success"
+
+    // MARK: - PaymentSheet initialization
+    case paymentSheetStartedLoading = "mc_started_loading"
+    case paymentSheetDismissed = "mc_dismiss"
```
yuki-stripe, line 172: Why is dismiss under `// MARK: - PaymentSheet initialization`?  It's not related to initialization, is it?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift
```diff
@@ -65,6 +69,9 @@ final class PaymentSheetLoader {
                             intent: intent
                         )
                     }
+
+                let loadingEndTime = CFAbsoluteTimeGetCurrent() - loadingStartTime
```
yuki-stripe, line 73: nit: This should be called duration, not loadingEndTime.  

ooc, why `CFAbsoluteTimeGetCurrent` instead of `Date`?

porter-stripe: When looking for referencing material on how to measure execution speed in Swift the most common suggestion is `CFAbsoluteTimeGetCurrent` from what I've seen. [1,](https://www.hackingwithswift.com/example-code/system/measuring-execution-speed-using-cfabsolutetimegetcurrent) [2](https://stackoverflow.com/questions/25006235/how-to-benchmark-swift-code-execution), but there is a bit of [debate](https://news.ycombinator.com/item?id=22824496).

When looking at [Date.swift](https://github.com/apple/swift-corelibs-foundation/blob/main/Sources/Foundation/Date.swift#L27C16-L27C40) it's just using `CFAbsoluteTimeGetCurrent` under the hood. `Date` is probably a bit more idiomatic so I'll swap here as it's equivalent.

StripeCore/StripeCore/Source/Analytics/STPAnalyticEvent.swift
porter-stripe: A dismiss event can happen during initialization, but can obviously happen after init. I moved it to under it's own `MARK:`.

yuki-stripe (approved):

PR #2732, opened 2023-07-13
Refactor PollingViewController to change hardcoded deadline and CTA prompt based on payment method

fionnbarrett-stripe:
## Summary

ATM PollingViewController has a hardcoded deadline of **5 minutes** for UPI and a hardcoded CTA prompt that is UPI specific. Instead I want make these configurable.

## Motivation

Implementing Blik into the PaymentSheet requires a **60 second** count down and different CTA


StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
```diff
@@ -1310,7 +1310,8 @@ public class STPPaymentHandler: NSObject {
             else {
                 return
             }
-            presentingVC.presentPollingVCForAction(currentAction)
+
+            presentingVC.presentPollingVCForAction(action: currentAction, type: .UPI)
```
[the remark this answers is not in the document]
fionnbarrett-stripe: Thanks for the review! The Localised strings used in the `PollingViewController` are only available inside `StripePaymentSheet` package. If I have a class `UPIPollingViewControllerDataSourceProvider`, I would need to select the `dataSourceProvider` based on the payment method which would ideally be done here in `_handleAuthenticationForCurrentAction` but  it doesn't have access to any `StripePaymentSheet` types?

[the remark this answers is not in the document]
fionnbarrett-stripe: how is [`presentPollingVCForAction` in PaymentSheetAuthenticationContext](https://github.com/stripe/stripe-ios/blob/81b560825cf02e4f3d92f4003344584cbdea23ba/StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift#L2313C10-L2313C35) able to call [`presentPollingVCForAction` in BottomViewController](https://github.com/stripe/stripe-ios/blob/81b560825cf02e4f3d92f4003344584cbdea23ba/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/BottomSheetViewController.swift#L284)? They are in different packages and StripePayment doesn't import StripePaymentSheet from what I can see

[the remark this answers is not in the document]
fionnbarrett-stripe: https://github.com/stripe/stripe-ios/compare/master...fionnbarrett/refactor-polling-controller-provider
I've created a new branch with my implementation of what was suggested. however I'm stuck with passing `UPIPollingViewControllerDataSourceProvider` to  [`presentPollingVCForAction`](https://github.com/stripe/stripe-ios/blob/8367bc65578689c8b87c0faa62eddb2b8e0d7b60/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/BottomSheetViewController.swift#L284).

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/BottomSheetViewController.swift
```diff
@@ -281,8 +281,8 @@ extension BottomSheetViewController: PaymentSheetAuthenticationContext {
         completion()
     }
 
-    func presentPollingVCForAction(_ action: STPPaymentHandlerActionParams) {
-        let pollingVC = PollingViewController(currentAction: action,
+    func presentPollingVCForAction(action: STPPaymentHandlerActionParams, type: STPPaymentMethodType) {
+        let pollingVC = PollingViewController(currentAction: action, type: type,
```
porter-stripe, line 285: Based on our zoom call lets create a new view model for the PollingViewController which computes the CTA and deadline, then inject it into the PollingViewController here, and also write tests for the new view controller.

The view model should compute the CTA and deadline based on the payment method type, e.g. inject the payment method type into the view model.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PollingViewModel.swift
```diff
@@ -0,0 +1,46 @@
...
+        }
+    }
+
+    private static func computeDeadline(paymentMethodType: STPPaymentMethodType) -> Date {
+        switch paymentMethodType {
+        case .UPI:
+            return Date().addingTimeInterval(60 * 5) // 5 minutes
+        default:
+            // some arbitrarily long deadline
+            return Date().addingTimeInterval(60 * 5 * 3) // 15 minutes
```
porter-stripe, line 43: For these cases lets throw a fatal error with a message.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PollingViewModel.swift
```diff
@@ -0,0 +1,46 @@
...
+        self.deadline = PollingViewModel.computeDeadline(paymentMethodType: paymentMethodType)
+    }
+
+    private static func computeCallToAction(paymentMethodType: STPPaymentMethodType) -> String {
+        switch paymentMethodType {
+        case .UPI:
+            return .Localized.open_upi_app
+        default:
+            // just show polling countdown
+            return "%@"
```
porter-stripe, line 33: Same here for the default case.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PollingViewModel.swift
```diff
@@ -0,0 +1,46 @@
...
+@_spi(STP) import StripePayments
+@_spi(STP) import StripePaymentsUI
+@_spi(STP) import StripeUICore
+import UIKit
+
+class PollingViewModel {
+
+    let paymentMethodType: STPPaymentMethodType
+    var CTA: String
+    var deadline: Date
```
porter-stripe, line 19: Let's change these to computed lazy variables and remove the compute functions.

E.g.

```swift
lazy var CTA: String {
    switch paymentMethodType {
        case .UPI:
            return .Localized.open_upi_app
        default:
            // throw fatal error
   }
}
```

Can do the same for deadline.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PollingViewModel.swift
```diff
@@ -0,0 +1,46 @@
...
+
+    init(paymentMethodType: STPPaymentMethodType) {
+        self.paymentMethodType = paymentMethodType
```
porter-stripe, line 22: Can we assert/throw a fatal error if a non-supported payment type is injected into `PollingViewModel`?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PollingViewController.swift
```diff
@@ -162,9 +166,18 @@ class PollingViewController: UIViewController {
...
+    init(currentAction: STPPaymentHandlerActionParams, type: STPPaymentMethodType, appearance: PaymentSheet.Appearance) {
         self.currentAction = currentAction
+        self.paymentMethodType = type
         self.appearance = appearance
+
+        switch paymentMethodType {
+        case .UPI:
+            deadline = Date().addingTimeInterval(60 * 5)
+        default:
+            fatalError("No Deadline for PaymentMethodType")
```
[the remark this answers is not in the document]
porter-stripe: IMO fatal error makes sense, we should't allow us to ship/use anything related to polling that doesn't support it.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PollingViewModel.swift
```diff
@@ -0,0 +1,38 @@
...
+        switch paymentMethodType {
+        case .UPI:
+            return Date().addingTimeInterval(60 * 5) // 5 minutes
+        default:
+            fatalError("Polling deadline has not been implemented for \(paymentMethodType)")
+        }
+    }()
+
+    init(paymentMethodType: STPPaymentMethodType) {
+        self.paymentMethodType = paymentMethodType
```
porter-stripe, line 36: Everything looks good, one nit that I mentioned earlier, lets fatal error if we init the view model with an unsupported payment method type. I don't think it should be possible to have an instance of the view model with a payment method type that doesn't require/support polling.

porter-stripe (approved):

PR #2745, opened 2023-07-17
Don't show "Pay using" header if we already have the "Pay using" text below.

davidme-stripe:
## Summary
Don't show "Pay using" twice.

## Motivation
Design feedback.

## Testing
PaymentSheet Example app


wooj-stripe (approved):
i will ignore the screenshots are they are not an accurate reflection of the change.

wooj-stripe (approved):

PR #2752, opened 2023-07-18
Fixes libpng warning for `stp_icon_add@3x.png`

jszumski:
## Summary
Fixes https://github.com/stripe/stripe-ios/issues/2667 by converting `stp_icon_add@3x.png` to 8 bit depth

Before:
```
$ file Stripe/StripeiOS/Resources/Images/stp_icon_add@3x.png
Stripe/StripeiOS/Resources/Images/stp_icon_add@3x.png: PNG image data, 51 x 51, 1-bit colormap, non-interlaced
```

After:
```
$ file Stripe/StripeiOS/Resources/Images/stp_icon_add@3x.png
Stripe/StripeiOS/Resources/Images/stp_icon_add@3x.png: PNG image data, 51 x 51, 8-bit/color RGBA, non-interlaced
```

Related: https://github.com/stripe/stripe-ios/pull/286

## Motivation
Fixes https://github.com/stripe/stripe-ios/issues/2667

## Testing
The libpng warning no longer appears when rebuilding with the updated file

## Changelog
n/a

PR #2758, opened 2023-07-18
Fix libping warning for stp_icon_add@3x.png

### LINKED ISSUE #2667: [BUG] PNG 8 bit warnings
## Summary

When building with Bazel we are seeing this warning:

```
From CopyPng Pods/Stripe/Stripe-intermediates/stp_icon_add@3x.png:
libpng warning: Warning: Input PNG does not have an 8 bit input depth.  Please convert your PNG to 8-bit for optimal performance on iPhone OS.
```

## iOS version

Min: iOS 14.0

## Installation method

Consumed via the CocoaPods podspec.

## SDK version

23.6.0

## Other information

Seems like the same as this previously resolved issue: https://github.com/stripe/stripe-ios/issues/278

wooj-stripe:
## Summary
Attempt to fix #2667, by converting image to 8bit RGBA.

## Motivation
#2667

## Testing
Relying on screenshot tests

Ran `file` on image:
```
$ file ./Stripe/StripeiOS/Resources/Images/stp_icon_add\@3x.png
./Stripe/StripeiOS/Resources/Images/stp_icon_add@3x.png: PNG image data, 51 x 51, 8-bit/color RGBA, non-interlaced
```


davidme-stripe (approved):

davidme-stripe:
Can we pngcrush this while still maintaining 8 bit color depth? Doesn't seem like it should be 4kb when the current one is 114 bytes.

davidme-stripe (changes requested):
Can we pngcrush this while still maintaining 8 bit color depth? Doesn't seem like it should be 4kb when the current one is 114 bytes.

wooj-stripe:
> FWIW in #2752 I just opened the old one in Preview and exported it with the depth dropdown set to "8 bits/channel", which ended up at a reasonable 709 bytes.

@jszumski - Thanks, although I got it down to 210 bytes using Gimp. The original export had XMP data and (i think) Exif Data enabled, which seems to be the cause of the image being ~4KB.

I also tried running the image through ImageOptim, but it was setting the image down to 1 bits/channel (which is probably how the image got to that in the first place)

davidme-stripe (approved):

jszumski:
Thanks for the fix!

PR #2760, opened 2023-07-19
Add script to check for 8-bit png images

wooj-stripe:
## Summary
Add script to check for 8-bit png images
Integrate into CI to run on each commit.
Update image 4bit -> 8-bit

## Motivation
To avoid running into running images through ImageOptim and coming out with images converted to something other than 8-bit png.

## Testing
Tested script with one image that was set to 4-bit.


ci_scripts/check_image_format.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
```
sfriedman-stripe, line 1: `shellcheck` spits out a ton of warnings and suggestions for this script. Would be good to go in and clean them up

wooj-stripe: Hmm I wasn't aware of shellcheck. I'll give it a shot. thanks!

davidme-stripe (approved):

PR #2763, opened 2023-07-20
Fix loading time calculation

porter-stripe:
## Summary
- The existing calculation for loading time was returning a negative value rather than positive (wrong sign), to make our queries easier we can flip it.

## Motivation
Make Hubble queries easier

## Testing
Manual

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift
@@ -71,7 +71,7 @@ final class PaymentSheetLoader {
                     }
 
                 STPAnalyticsClient.sharedClient.logPaymentSheetEvent(event: .paymentSheetLoadSucceeded,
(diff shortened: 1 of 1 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift
```diff
@@ -71,7 +71,7 @@ final class PaymentSheetLoader {
                     }
 
                 STPAnalyticsClient.sharedClient.logPaymentSheetEvent(event: .paymentSheetLoadSucceeded,
-                                                                     duration: loadingStartDate.timeIntervalSinceNow)
+                                                                     duration: Date().timeIntervalSince(loadingStartDate))
```
wooj-stripe, line 74: If we change this from negative to positive, i'm guessing there's some way to unify the negative times with the positive ones we will be sending?

wooj-stripe (approved):
code lgtm, but unclear about how we deal with this on the backend.

PR #2774, opened 2023-07-24
Construct a better loggable string for errors

porter-stripe:
## Summary
- Makes our errors better identifiable with a shorter string for better network performance
- 100% guarantee no PII

## Motivation
Efficiency and safety of logging errors

## Testing
- Manual

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetError.swift
```diff
@@ -69,60 +69,60 @@ extension PaymentSheetError: CustomDebugStringConvertible {
     /// A string that can safley be logged to our analytics service that does not contain any PII
     public var safeLoggingString: String {
```
porter-stripe, line 70: We have to construct these case names manually since we override `debugDescription` via `CustomDebugStringConvertible`. 🙃

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetError.swift
```diff
@@ -69,73 +69,128 @@ extension PaymentSheetError: CustomDebugStringConvertible {
...
         switch self {
-        case .unknown(debugDescription: let message):
-            return "An unknown error occurred in PaymentSheet. " + message
-        case .linkLookupNotFound(serverErrorMessage: let message):
-            return "An error occurred in PaymentSheet. " + message
+        case .unknown:
+            return "An unknown error occurred in PaymentSheet. " + errorMsesageSuffix
         default:
-            return "An error occurred in PaymentSheet. " + safeLoggingString
+            return "An error occurred in PaymentSheet. " + errorMsesageSuffix
```
wooj-stripe, line 193: change errorMsesageSuffix to errorMessageSuffix

porter-stripe: good catch, fixed!

wooj-stripe (approved):

PR #2775, opened 2023-07-24
Update Tuist instructions

davidme-stripe:
I'd prefer that we point users to Tuist's manual instead of including instructions to pipe a script to bash.


porter-stripe (approved):
agree!

PR #2782, opened 2023-07-25
Log some new checkout events for PaymentSheet

porter-stripe:
## Summary
- Adds two new events to track the health of PaymentSheet checkout.

## Motivation

## Testing
- Manual

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/PaymentMethodTypeCollectionView.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/PaymentMethodTypeCollectionView.swift
@@ -114,7 +114,7 @@ extension PaymentMethodTypeCollectionView: UICollectionViewDataSource, UICollect
     func collectionView(_ collectionView: UICollectionView, didSelectItemAt indexPath: IndexPath) {
         collectionView.scrollToItem(at: indexPath, at: .centeredHorizontally, animated: true)
         selected = paymentMethodTypes[indexPath.item]
-        STPAnalyticsClient.sharedClient.logPaymentSheetEvent(event: .paymentSheetPaymentMethodTapped,
+        STPAnalyticsClient.sharedClient.logPaymentSheetEvent(event: .paymentSheetCarouselPaymentMethodTapped,
                                                              params: ["selected_lpm": paymentMethodTypes[indexPath.row].identifier])
     }
 
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/PaymentMethodTypeCollectionView.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/PaymentMethodTypeCollectionView.swift
@@ -34,12 +34,14 @@ class PaymentMethodTypeCollectionView: UICollectionView {
     }
     let paymentMethodTypes: [PaymentSheet.PaymentMethodType]
     let appearance: PaymentSheet.Appearance
+    let isPaymentSheet: Bool
     weak var _delegate: PaymentMethodTypeCollectionViewDelegate?
 
     init(
         paymentMethodTypes: [PaymentSheet.PaymentMethodType],
         initialPaymentMethodType: PaymentSheet.PaymentMethodType? = nil,
         appearance: PaymentSheet.Appearance,
+        isPaymentSheet: Bool = false,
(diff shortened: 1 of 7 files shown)


StripePayments/StripePayments/Source/API Bindings/Models/PaymentMethods/STPPaymentMethodEnums.swift
```diff
@@ -139,4 +139,68 @@ import Foundation
...
+        case .weChatPay:
+            return "wechat_pay"
+        case .boleto:
+            return "boleto"
+        case .link:
+            return "link"
+        case .klarna:
+            return "klarna"
+        case .linkInstantDebit:
+            return "TODO" // TODO(porter) figure out the correct id here
```
porter-stripe, line 195: @davidme-stripe @jaynewstrom-stripe any idea how link instant debit shows up on an intent? E.g. for Afterpay it's `afterpay_clearpay`. Can't find anything in the codebase.

vardges-stripe: Is this enum value actually used? This might be leftover from an old instant debits implementation

porter-stripe: @vardges-stripe I see 31 references to it in our codebase, I don't know enough about it to feel comfortable removing it. I may just put something in here like "link_instant_debits", for now this string will only be used for analytics/metrics. Thoughts?

porter-stripe: @vardges-stripe any FUD with the string I added [here](https://github.com/stripe/stripe-ios/pull/2782/commits/2286b632c4010a6b9fe39e52c1d445056e38d2b0)?

vardges-stripe: nope that's good @porter-stripe please proceed, sorry missed the earlier message.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/PaymentMethodTypeCollectionView.swift
```diff
@@ -114,6 +114,8 @@ extension PaymentMethodTypeCollectionView: UICollectionViewDataSource, UICollect
     func collectionView(_ collectionView: UICollectionView, didSelectItemAt indexPath: IndexPath) {
         collectionView.scrollToItem(at: indexPath, at: .centeredHorizontally, animated: true)
         selected = paymentMethodTypes[indexPath.item]
+        STPAnalyticsClient.sharedClient.logPaymentSheetEvent(event: .paymentSheetCarouselPaymentMethodTapped,
```
wooj-stripe, line 117: This view is also being used in customerSheet, so calling it `paymentSheet` may not be accurate.

porter-stripe: Oh interesting, I'll rename this in that case.

wooj-stripe: Looks like we went with a more generic name.  Seems that we loose some insight into paymentSheet only behavior.  That seems problematic, right?  I think we need to pass in a parameter into this class so that we don't attribute CustomerSheet taps into PaymentSheet taps, and vice versa,  what do you think?

porter-stripe: The string is fairly generic that we will see in Hubble `"mc_carousel_payment_method_tapped"`. We should be able to differentiate between CustomerSheet and PaymentSheet one of two ways:

1. Looking at `product_usage`, this might not work well if someone is using CustomerSheet and PaymentSheet.
2. Looking at the session ID, and tying other PaymentSheet specific events together

However this does make querying more difficult, I'm going to update this so it only sends when being used with PaymentSheet as that is the true intention, and I'm not sure how this event ties into the CustomerSheet health story. If we want to send this event for CustomerSheet it should have a different name such as `"cs_carousel_payment_method_tapped" `or something.

wooj-stripe (approved):

PR #2786, opened 2023-07-26
Enable iDEAL for SetupIntents and PaymentIntents+setup_future_usage

yuki-stripe:
## Summary
- Un-filters iDEAL for SI, PI+sfu and adds form.
- Updates PaymentSheet_LPM_ConfirmFlowTests with some more explanation and a test for iDEAL

## Motivation
https://docs.google.com/document/d/19RVShmZqOQDfShWY3731UBTTRP7INqQXqlAcmy4BY30/edit

## Testing
Manually tested in addition to the LPMConfirmFlow tests

## Changelog
See CHANGELOG


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
```diff
@@ -155,10 +155,15 @@ class PaymentSheetFormFactory {
             additionalElements = [makePaypalMandate()]
         }
 
-        // 2. Element-based forms defined in JSON
         guard let spec = specFromJSONProvider() else {
-            fatalError()
+            assertionFailure("Failed to get form spec!")
+            return FormElement(elements: [], theme: theme)
+        }
+        if paymentMethod.stpPaymentMethodType == .iDEAL && saveMode == .merchantRequired {
```
yuki-stripe, line 162: @wooj-stripe Do you have any opinion on keeping or removing the `saveMode == .merchantRequired` condition here?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
```diff
@@ -278,10 +278,12 @@ extension PaymentSheet {
                         return [.returnURL]
                     case .USBankAccount:
                         return [.userSupportsDelayedPaymentMethods]
-                    case .iDEAL, .bancontact, .sofort:
-                        // SEPA-family PMs are disallowed until we can reuse them for PI+sfu and SI.
```
wooj-stripe, line 282: Do we want to leave some comment about `SEPA-family PMs are disallowed until we can reuse them for PI+sfu and SI.`, or was it intentional to delete?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
wooj-stripe: 1. `saveMode` is not something i have a lot of context with, but it seems that if saveMode is set to `.merchantRequired` we add a mandate. 
2.  I'm not sure if i understand why the logic for checking `paymentMethod.stpPaymentMethodType == .iDEAL` is after the `specFromJSONProvider` call.  I thought that the idea was that we check to see if there are any client side overrides first, and if not rely specs, but by adding this logic after checking the specs, it's like we have an override of the specs?

yuki-stripe: 1. saveMode is `.merchantRequired` if the merchant is using a PI+SFU or SI.

I'm wondering if you have an opinion on continuing to use LUXE to generate the form or just using `makeIdeal()` in code for vanilla PaymentIntents (PIs without setup_future_usage).  We could go either way.

2. It's b/c I wrote `makeIdeal` to depend on the spec https://github.com/stripe/stripe-ios/pull/2786/files/0a4f2a81f1c17034167fd5dd6aa2b91f877eafe1#diff-019a44bb1b90571f9e0662e1c955dc6683079ef3152b45a6b22d6b3399288c32R392-R393

wooj-stripe: Ah, if i understand correctly... we should probably continue to use the spec for PI w/out SFU, considering ElementsWeb, Checkout and Android should be doing that -- if we update the spec, all surfaces should be updated.

yuki-stripe: WDYM if we update the spec?  Also, Web PE doesn't use LUXE for iDEAL afaict based on the link at the bottom of https://trailhead.corp.stripe.com/docs/luxe/platform/architecture.

wooj-stripe: I took at look at this again --- i think we should just get rid of savemode ==.merchantRequired and add `paymentMethod.stpPaymentMethodType == .iDEAL` to the list of `1. Custom, one-off forms` effectively removing it from LUXE.

wooj-stripe (approved):

wooj-stripe (approved):

PR #2792, opened 2023-07-28
[Camera-core] Try search for `builtInTripleCamera`

ccen-stripe:
## Summary

We're currently not searching for `builtInTripleCamera` when trying to take video from back camera, this results in iPhone 14 pro/max choosing `builtInDualCamera`, which fails to focus when getting too close to object.

## Motivation

Fix iPhone 14 pro camera focus issue

## Testing

Manually verified, see recording

## Recording(on iPhone 14 pro)
| Before  | After |
| before  | after |

### DIFF
--- a/StripeCameraCore/StripeCameraCore/Source/Coordinators/CameraSession.swift
+++ b/StripeCameraCore/StripeCameraCore/Source/Coordinators/CameraSession.swift
@@ -441,6 +441,7 @@ extension CameraSession {
         guard let captureDevice = captureDevices.devices.first else {
             throw CameraSessionError.captureDeviceNotFound
         }
+        print("BGLM - captureDevice: \(captureDevice)")
 
         return try AVCaptureDeviceInput(device: captureDevice)
     }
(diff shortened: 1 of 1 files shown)


StripeCameraCore/StripeCameraCore/Source/Coordinators/CameraSession.swift
```diff
@@ -441,6 +441,7 @@ extension CameraSession {
         guard let captureDevice = captureDevices.devices.first else {
             throw CameraSessionError.captureDeviceNotFound
         }
+        print("BGLM - captureDevice: \(captureDevice)")
```
yuki-stripe, line 444: did you mean to include this print statement?

yuki-stripe (dismissed):

StripeCameraCore/StripeCameraCore/Source/Coordinators/CameraSession.swift
ccen-stripe: removed!

yuki-stripe (approved):

PR #2804, opened 2023-08-01
Dismiss sheet after adding account w/ microdeposits

wooj-stripe:
## Summary
After adding a bank account manually, we need to dismiss the sheet, rather than allowing it to be a selectable payment method option.

## Motivation
Bank account isn't confirmed yet, so it can not be selected.

## Testing
- added ui test
- Manually tested w/ and without microdeposits

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/USBankAccount/USBankAccountPaymentMethodElement.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/USBankAccount/USBankAccountPaymentMethodElement.swift
@@ -31,7 +31,7 @@ final class USBankAccountPaymentMethodElement: Element {
     private let theme: ElementsUITheme
     private var linkedBank: LinkedBank? {
         didSet {
-            self.mandateString = Self.attributedMandateText(for: linkedBank, merchantName: merchantName, isSaving: savingAccount.value, theme: theme)
+            self.mandateString = Self.attributedMandateText(for: linkedBank, merchantName: merchantName, isSaving: savingAccount.value, configuration: configuration, theme: theme)
         }
     }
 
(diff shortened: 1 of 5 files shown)


davidme-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/USBankAccount/USBankAccountPaymentMethodElement.swift
```diff
@@ -50,7 +50,7 @@ final class USBankAccountPaymentMethodElement: Element {
...
     static let MicrodepositCopy: String = STPLocalizedString("Stripe will deposit $0.01 to your account in 1-2 business days. Then you’ll get an email with instructions to complete payment to %@.", "Prompt for microdeposit verification before completing purchase with merchant. %@ will be replaced by merchant business name")
-    static let MicrodepositCopy_CustomerSheet: String = STPLocalizedString("Stripe will deposit $0.01 to your account in 1-2 business days. Then you’ll get an email with instructions to complete before confirming your bank account with %@.", "Prompt for microdeposit verification before completing purchase with merchant. %@ will be replaced by merchant business name")
+    static let MicrodepositCopy_CustomerSheet: String = STPLocalizedString("Stripe will deposit $0.01 to your account in 1-2 business days. Then you’ll get an email with instructions before saving your bank account with %@.", "Prompt for microdeposit verification before completing saving payment method with merchant. %@ will be replaced by merchant business name")
```
davidme-stripe, line 53: "Then you'll get an email with instructions before saving your bank account with %@" doesn't make sense, maybe something like "You’ll get an email with instructions to finish saving your bank account with %@"?

davidme-stripe (changes requested):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/USBankAccount/USBankAccountPaymentMethodElement.swift
wooj-stripe: I was trying to get the prompt to explain they needed to verify their account and then select the payment method in-app.. but..  it's a bit clunky.

We could say something like "Then you'll get an email with instructions to verify and save your account before selecting your bank account with %@" -- but that's a mouthful.

Lets go with your suggestion.

davidme-stripe (approved):

PR #2806, opened 2023-08-02
Send "selected_lpm" on success and failure to confirm analytics events

porter-stripe:
## Summary
- Sends `selected_lpm` payload on success and failure events
- This will allow us to see success rate and checkout times for given LPMs

## Motivation

## Testing
Manual

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
@@ -22,6 +22,19 @@ extension PaymentSheet {
         case saved(paymentMethod: STPPaymentMethod)
         case new(confirmParams: IntentConfirmParams)
         case link(option: LinkConfirmOption)
+        
+        var paymentMethodTypeAnalyticsValue: String? {
+            switch self {
+            case .applePay:
+                return PaymentSheet.PaymentMethodType.dynamic("apple_pay").identifier
+            case .saved(paymentMethod: let paymentMethod):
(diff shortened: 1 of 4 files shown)


porter-stripe:
cc @tillh-stripe

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
```diff
@@ -22,6 +22,19 @@ extension PaymentSheet {
         case saved(paymentMethod: STPPaymentMethod)
         case new(confirmParams: IntentConfirmParams)
         case link(option: LinkConfirmOption)
+
+        var paymentMethodTypeAnalyticsValue: String? {
```
wooj-stripe, line 26: Did we consider using `PaymentSheet.PaymentMethodType` instead of exposing a String?
Then for the saved case, we can have `return paymentMethod.paymentSheetPaymentMethodType()`

porter-stripe: Yes this was explored, 2 reasons why I didn't go this route:

1. There is no representation for Apple Pay in `PaymentSheet.PaymentMethodType`. I don't think we should add one just for analytics. I thought about using `.dynamic("apple_pay")` to fill this void... but that creates an invalid/non-existent payment method type that someone could misuse in the future. 
2. Based on the fear of this being misused, I wanted this string to be scoped strictly for analytics.

wooj-stripe (approved):

PR #2808, opened 2023-08-02
Update Link brand string

davidme-stripe:
## Summary
This string was translated incorrectly in some languages — add more context and send it off for re-translation.

## Testing
Existing snapshot tests

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/PayWithLinkButton.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/PayWithLinkButton.swift
@@ -109,7 +109,7 @@ final class PayWithLinkButton: UIControl {
         linkView.font = UIFont.systemFont(ofSize: 15, weight: .medium)
             .scaled(withTextStyle: .callout, maximumPointSize: 16)
 
-        let payWithLinkString = NSMutableAttributedString(string: String.Localized.pay_with_payment_method)
+        let payWithLinkString = NSMutableAttributedString(string: String.Localized.pay_with_link)
 
         // Create the Link logo attachment
         let linkImage = Image.link_logo.makeImage(template: true)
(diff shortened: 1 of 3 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/PayWithLinkButton.swift
```diff
@@ -121,12 +121,12 @@ final class PayWithLinkButton: UIControl {
         linkAttachment.bounds = CGRect(x: 0, y: 0, width: linkLogoHeight * linkLogoRatio, height: linkLogoHeight)
 
         // Add a spacer before the Link logo and after the Link logo
-        let range = payWithLinkString.mutableString.range(of: "%@")
+        let range = payWithLinkString.mutableString.range(of: "Link")
```
wooj-stripe, line 124: This is potentially dangerous if "Link" gets translated to something that isn't exactly "Link".  Which... Maybe could happen for maybe.. non-roman-type languages (chinese, korean, etc)...

For example: If you change "pay_with_payment_method" to "update_card", and then you update the code to replace "Link" with "card", it works great in english, but once you change the language settings to Spanish, then we end up crashing because "card" is actually translated to "tarjeta"

Maybe we can update this to search for Link, and if it does exist, then change it? or... maybe find a better way to ensure the "Link" portion of link is replaced with the image?

wooj-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/PayWithLinkButton.swift
davidme-stripe: Followed up offline, we're going to solve this by communicating with localization team (we want them to translate the string including the word Link)

PR #2811, opened 2023-08-03
Convert a few tests to Swift

davidme-stripe:
## Summary
Try auto-converting a few tests to Swift

## Testing
CI


Stripe/StripeiOSTests/STPAddressTests.swift
```diff
@@ -0,0 +1,532 @@
+//  Converted to Swift 5.8.1 by Swiftify v5.8.28463 - https://swiftify.com/
```
yuki-stripe, line 1: should we delete these lines?

yuki-stripe (approved):

Stripe/StripeiOSTests/STPAddressTests.swift
davidme-stripe: I'm in favor of keeping them, considering these are just tests and it helps explain why the code is so weird

PR #2831, opened 2023-08-05
Fix macOS Catalyst build

davidme-stripe:
## Summary
Went a little too far with the `@available` removal: Camera support was added for macOS Catalyst in the iOS 14 SDK, so we need to hide it for Catalyst in the iOS 13 SDK. Add some availability checks.

## Motivation
Fixing the release

## Testing
Locally, will also run in Xcode Cloud

### DIFF
--- a/Stripe/StripeiOS/Source/STPAddCardViewController.swift
+++ b/Stripe/StripeiOS/Source/STPAddCardViewController.swift
@@ -138,6 +138,7 @@ public class STPAddCardViewController: STPCoreTableViewController, STPAddressVie
     private var doneItem: UIBarButtonItem?
     private var cardHeaderView: STPSectionHeaderView?
 
+    @available(macCatalyst 14.0, *)
     private var cardScanner: STPCardScanner? {
         get {
             _cardScanner as? STPCardScanner
@@ -150,6 +151,7 @@ public class STPAddCardViewController: STPCoreTableViewController, STPAddressVie
     /// Storage for `cardScanner`.
     private var _cardScanner: NSObject?
 
+    @available(macCatalyst 14.0, *)
     private var scannerCell: STPCardScannerTableViewCell? {
         get {
             _scannerCell as? STPCardScannerTableViewCell
@@ -254,7 +256,9 @@ public class STPAddCardViewController: STPCoreTableViewController, STPAddressVie
         addressViewModel.delegate = self
         title = STPLocalizedString("Add a Card", "Title for Add a Card view")
 
-        cardScanner = STPCardScanner()
+        if #available(macCatalyst 14.0, *) {
+            cardScanner = STPCardScanner()
+        }
     }
 
     /// :nodoc:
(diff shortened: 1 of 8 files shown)


Stripe/StripeiOS/Source/STPAddCardViewController.swift
```diff
@@ -687,7 +697,11 @@ public class STPAddCardViewController: STPCoreTableViewController, STPAddressVie
         case STPPaymentCardSection.stpPaymentCardNumberSection.rawValue:
             cell = paymentCell
         case STPPaymentCardSection.stpPaymentCardScannerSection.rawValue:
-            cell = scannerCell
+            if #available(macCatalyst 14.0, *) {
+                cell = scannerCell
+            } else {
+                return UITableViewCell()
```
sfriedman-stripe, line 703: What's the behavior in this case?

Stripe/StripeiOS/Source/STPAddCardViewController.swift
```diff
@@ -391,33 +395,37 @@ public class STPAddCardViewController: STPCoreTableViewController, STPAddressVie
...
-            for: .touchUpInside
-        )
-        cardHeaderView?.setNeedsLayout()
     }
 
     @objc func scanCard() {
-        view.endEditing(true)
-        isScanning = true
-        cardScanner?.start()
+        if #available(macCatalyst 14.0, *) {
```
sfriedman-stripe, line 424: Should we add this to the method signature instead of the impl? From a cursory look at the PR, it looks like it's only called from within an availability block anyway. Maybe that's incompatible with `addTarget(_:action:for:)`?

Stripe/StripeiOS/Source/STPAddCardViewController.swift
```diff
@@ -810,7 +824,9 @@ public class STPAddCardViewController: STPCoreTableViewController, STPAddressVie
         super.viewWillTransition(to: size, with: coordinator)
         let orientation = UIDevice.current.orientation
         if orientation.isPortrait || orientation.isLandscape {
-            cardScanner?.deviceOrientation = orientation
+            if #available(macCatalyst 14.0, *) {
```
sfriedman-stripe, line 827: Should we wrap the whole method impl with this instead? I don't think any of it is relevant if scanning isn't available right?

sfriedman-stripe (approved):
LGTM overall. A few questions/comments but nothing blocking

Stripe/StripeiOS/Source/STPAddCardViewController.swift
davidme-stripe: Should never happen — the "Scan card" button won't appear, so `isScanning` will never be set, so `numberOfRowsInSection` for the scanner section will be 0.

Stripe/StripeiOS/Source/STPAddCardViewController.swift
davidme-stripe: Yep! Just clicked the wrong autofix option.

wooj-stripe (approved):

PR #2845, opened 2023-08-10
Support BLIK Paymentsheet

fionnbarrett-stripe:


porter-stripe:
Looks like this PR has changes that were in previous PRs?

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetBillingCollectionUITests.swift
```diff
@@ -874,4 +874,212 @@ class PaymentSheetBillingCollectionLPMUITests: PaymentSheetBillingCollectionUITe
...
+        XCTAssertFalse(zipField.exists)
+
+        XCTAssertFalse(payButton.isEnabled)
+        let blik_code = app.textFields["BLIK code"]
+        blik_code.tap()
+        blik_code.typeText("123456")
+        blik_code.typeText(XCUIKeyboardKey.return.rawValue)
+
+        payButton.tap()
+//        XCTAssertTrue(successText.waitForExistence(timeout: 10.0))
```
porter-stripe, line 918: Why is this commented out?

porter-stripe: Same with the rest of them.

fionnbarrett-stripe:
> Looks like this PR has changes that were in previous PRs?

Yeah, I needed to merge master.

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetBillingCollectionUITests.swift
fionnbarrett-stripe: The test will poll now but it gets approved automatically in 10 seconds. is this okay?

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
```diff
@@ -1271,12 +1271,12 @@ public class STPPaymentHandler: NSObject {
...
-                let currentAction = self.currentAction
-                    as? STPPaymentHandlerPaymentIntentActionParams
+                let presentingVC = currentAction.authenticationContext
+                    as? PaymentSheetAuthenticationContext
             else {
-                fatalError()
+                return
             }
-            currentAction.complete(with: .succeeded, error: nil)
+            presentingVC.presentPollingVCForAction(action: currentAction, type: .blik)
```
porter-stripe, line 1280: This is a little tricky! If we are using PaymentSheet we want to use the PollingViewController, otherwise we should use the existing logic for API bindings users. I would recommend doing something like by checking if we have a payment sheet context like we do for UPI but add the existing logic into the body of the guard

```swift
guard
    let presentingVC = currentAction.authenticationContext
        as? PaymentSheetAuthenticationContext
else {
   // Do existing bindings logic
}

presentingVC.presentPollingVCForAction(action: currentAction, type: .blik)
```


Read the last paragraph in this docs section for context: https://stripe.com/docs/payments/blik/accept-a-payment?platform=mobile&ui=ios#ios-submit-payment

Add some comments here so future devs know why we're doing this.

porter-stripe:
Also lets add a Blik test in `PaymentSheetUITest.swift`, should be just like Zip.

Another spot we should add a test is in `PaymentSheetLPMConfirmFlowTests.swift`. This will be a little tricker, we need to make some changes so we can pass in the merchant country and use the correct publishable key for a Belgium merchant.

I've already made most of the changes you'll need in https://github.com/stripe/stripe-ios/pull/2848 but I won't be in the office this week, so they may not be changed. In my changes I'm using a Singapore merchant but in your tests you'll need to use the Belgium merchant (`be`).

fionnbarrett-stripe:
> Also lets add a Blik test in `PaymentSheetUITest.swift`, should be just like Zip.
> 
> Another spot we should add a test is in `PaymentSheetLPMConfirmFlowTests.swift`. This will be a little tricker, we need to make some changes so we can pass in the merchant country and use the correct publishable key for a Belgium merchant.
> 
> I've already made most of the changes you'll need in #2848 but I won't be in the office this week, so they may not be changed. In my changes I'm using a Singapore merchant but in your tests you'll need to use the Belgium merchant (`be`).

I'm encountering issues with the `Deferred PaymentIntent - server side confirmation` intent failing on BLIK. it appears to not send the code  with the error `The parameter `payment_method_options[blik][code]` is required.` . I also noticed that `confirm: true` on the payment intent so my understanding is this means it instantly confirms the PaymentIntent but that is where BLIK sends its code.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetLPMConfirmFlowTests.swift
```diff
@@ -108,19 +111,25 @@ final class PaymentSheet_LPM_ConfirmFlowTests: XCTestCase {
...
+            form.getTextFieldElement("BLIK code")?.setText("123456")
+        }
+    }
 }
 
 // MARK: - Helper methods
 extension PaymentSheet_LPM_ConfirmFlowTests {
     enum IntentKind: CaseIterable {
         case paymentIntent
+        case deferredPaymentIntent
```
yuki-stripe, line 132: AFAIU Blik is only unsupported for deferred *server-side confirmation* but is supported for deferred *client-side confirmation*.

Instead of adding a case here, how about we remove the `deferredSSC` intent just for blik, around line 250?

fionnbarrett-stripe: BLIK doesn't show up in the payment sheet anyway when deferredCSC is used.

https://stripe.slack.com/archives/C04HHL5CN1G/p1692125154846659?thread_ts=1690990183.768019&cid=C04HHL5CN1G

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetBillingCollectionUITests.swift
```diff
@@ -874,4 +874,220 @@ class PaymentSheetBillingCollectionLPMUITests: PaymentSheetBillingCollectionUITe
...
+        XCTAssertFalse(payButton.isEnabled)
+        let blik_code = app.textFields["BLIK code"]
+        blik_code.tap()
+        blik_code.typeText("123456")
+        blik_code.typeText(XCUIKeyboardKey.return.rawValue)
+
+        payButton.tap()
+
+        // in test mode polling is auto approved in 10 seconds
+        XCTAssertTrue(successText.waitForExistence(timeout: 25.0))
```
porter-stripe, line 920: These are just too long, I don't think we need to test with so many variations here, maybe just one is fine.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetLPMConfirmFlowTests.swift
```diff
@@ -53,7 +56,7 @@ final class PaymentSheet_LPM_ConfirmFlowTests: XCTestCase {
     }
 
     func testBancontactConfirmFlows() async throws {
-        try await _testConfirm(intentKinds: [.paymentIntent], currency: "EUR", paymentMethodType: .dynamic("bancontact")) { form in
+        try await _testConfirm(intentKinds: [.paymentIntent, .deferredPaymentIntent], currency: "EUR", paymentMethodType: .dynamic("bancontact")) { form in
```
porter-stripe, line 59: +1 to @yuki-stripe suggestion's around not adding deferred intent as it's own type, but rather just ignore it for only Blik in the code below.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetLPMConfirmFlowTests.swift
porter-stripe: From a technical standpoint I don't see why we can't support Blik in CSC for deferred... but elements/session doesn't even send it back when the intent type is deferred.

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
```diff
@@ -573,6 +609,38 @@ class PaymentSheetStandardLPMUITests: PaymentSheetUITestCase {
         webviewCloseButton.tap()
     }
 
+    func testBLIKPaymentMethod() throws {
```
porter-stripe, line 612: Do we need this if we have `testBLIKPaymentMethodPolling` what is it testing that `testBLIKPaymentMethodPolling` isn't?

porter-stripe:
Let's also add an update to the change log, you can see an example [here](https://github.com/stripe/stripe-ios/pull/2848/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR3).

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 ### All
 * Improved redirect UX when using Cash App Pay.
 
+### PaymentSheet
```
porter-stripe, line 9: Move this up under the "## x.x.x x-x-x" header.

porter-stripe (commented):
Everything is looking really good, just need a minor change to the change log and I have one question.

Why does web collect email but we don't?
CleanShot 2023-08-22 at 09 45 31@2x

fionnbarrett-stripe:
> Everything is looking really good, just need a minor change to the change log and I have one question.
> 
> Why does web collect email but we don't? CleanShot 2023-08-22 at 09 45 31@2x

I talked with @khuevu-stripe who implemented BLIK into payment element https://git.corp.stripe.com/stripe-internal/stripe-js-v3/pull/10519 and said this is not necessary to be part of PE.
It was added to simulate failures in test mode by using email pattern.

porter-stripe (approved):

PR #2849, opened 2023-08-10
Enable testmode for Stripe Dashboard app

mludowise-stripe:
## Summary

Allows the Stripe Dashboard app to internally test the payment sheet in test mode.

## Motivation

The Stripe Dashboard app has an internal debug menu that allows us to toggle live mode vs. test mode during development, however the SDK was setting the `Stripe-Livemode` header needed for Dashboard keys based on a run scheme environment variable instead of the debug menu. The env var could get out of sync if the debug menu was changed and would only work when running the app attached to Xcode.

This fix enables the Dashboard app to directly toggle livemode in the SDK when importing the module with `@_spi(DashboardOnly)`.

## Testing

See corresponding Dashboard app change.

## Changelog

n/a

### DIFF
--- a/StripeCore/StripeCore/Source/API Bindings/STPAPIClient.swift
+++ b/StripeCore/StripeCore/Source/API Bindings/STPAPIClient.swift
@@ -87,6 +87,10 @@ import UIKit
         return publishableKey?.hasPrefix("uk_") ?? false
     }
 
+    /// Determines the `Stripe-Livemode` header value when the publishable key is a user key
+    /// :nodoc:
+    @_spi(StripeAppsOnly) public var userKeyLiveMode = true
+
     // MARK: Initializers
     override public init() {
         sourcePollers = [:]
@@ -214,8 +218,7 @@ import UIKit
         var headers = ["Authorization": "Bearer " + authorizationBearer]
 
         if publishableKeyIsUserKey {
-            let liveMode = ProcessInfo.processInfo.environment["Stripe-Livemode"] != "false"
-            headers["Stripe-Livemode"] = liveMode ? "true" : "false"
+            headers["Stripe-Livemode"] = userKeyLiveMode ? "true" : "false"
         }
         return headers
     }
@@ -224,7 +227,7 @@ import UIKit
         guard let publishableKey = publishableKey, !publishableKey.isEmpty else {
             return false
         }
-        return publishableKey.lowercased().hasPrefix("pk_test")
+        return publishableKey.lowercased().hasPrefix("pk_test") || (publishableKeyIsUserKey && !userKeyLiveMode)
     }
 }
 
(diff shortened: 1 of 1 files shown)


StripeCore/StripeCore/Source/API Bindings/STPAPIClient.swift
```diff
@@ -87,6 +87,10 @@ import UIKit
         return publishableKey?.hasPrefix("uk_") ?? false
     }
 
+    /// Determines the `Stripe-Livemode` header value when the publishable key is a user key
+    /// :nodoc:
```
mludowise-stripe, line 91: Do we still need `:nodoc:`? I remember that being necessary for spi-public stuff earlier to let Jazzy know to ignore it.

StripeCore/StripeCore/Source/API Bindings/STPAPIClient.swift
```diff
@@ -87,6 +87,10 @@ import UIKit
...
+    @_spi(StripeAppsOnly) public var userKeyLiveMode = true
```
mludowise-stripe, line 92: I thought about calling this `DashboardOnly` inline with the `LinkOnly` notation, but I think this might be used by more than just dashboard... I'm not totally sure TBH

StripeCore/StripeCore/Source/API Bindings/STPAPIClient.swift
```diff
@@ -224,7 +227,7 @@ import UIKit
         guard let publishableKey = publishableKey, !publishableKey.isEmpty else {
             return false
         }
-        return publishableKey.lowercased().hasPrefix("pk_test")
+        return publishableKey.lowercased().hasPrefix("pk_test") || (publishableKeyIsUserKey && !userKeyLiveMode)
```
mludowise-stripe, line 230: This shows the "Test mode" badge when we're using testmode from the dashboard now

StripeCore/StripeCore/Source/API Bindings/STPAPIClient.swift
yuki-stripe: Seems fine. I think you could also call this @_spi(STP) too.

StripeCore/StripeCore/Source/API Bindings/STPAPIClient.swift
yuki-stripe: nope, can remove, we use DocC now

StripeCore/StripeCore/Source/API Bindings/STPAPIClient.swift
yuki-stripe: Nice

yuki-stripe (approved):

StripeCore/StripeCore/Source/API Bindings/STPAPIClient.swift
mludowise-stripe: I considered using `STP` but I think that anything marked as `STP` is implicitly internal and could change without warning. Whereas marking it as `StripeAppsOnly` or `DashboardOnly` implies that someone else is using this and to be wary of making breaking changes.

I am leaning towards `DashboardOnly` now since I think this is consistent with internal naming for API fields that use the `uk_` key.

yuki-stripe (approved):

PR #2871, opened 2023-08-23
Add FPX to PaymentSheet

porter-stripe:
## Summary
- Adds FPX to PaymentSheet via LUXE and moves GrabPay to LUXE.

Related PRs:
https://git.corp.stripe.com/stripe-internal/pay-server/pull/677050
https://git.corp.stripe.com/stripe-internal/stripe-js-v3/pull/17567

## Motivation
- New LPM
-  Move GrabPay to LUXE

## Testing
- Manual and new unit tests

https://github.com/stripe/stripe-ios/assets/88012362/9fe86352-f078-44e3-99f2-f5ccf5bddb9e

## Changelog
See diff


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 ### PaymentSheet
 * [Added] Support for BLIK with PaymentIntents.
 * [Added] Support for AmazonPay (private beta) with PaymentIntents.
+* [Added] Support for FPX with PaymentIntents.
 * [Fixed] A bug where payment amounts were not displayed correctly for LAK currency.
```
porter-stripe, line 6: If we can get this merged in before the next deploy maybe we combine this into a single row.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
```diff
@@ -25,6 +25,10 @@ extension PaymentSheet {
                 return [.returnURL]
             case .dynamic("zip"):
                 return [.returnURL]
+            case .dynamic("grabpay"):
+                return [.returnURL]
+            case .dynamic("fpx"):
+                return [.returnURL]
```
yuki-stripe, line 31: Should we put grabpay or fpx under 'dynamic' when we have bindings for them?  cc @wooj-stripe

porter-stripe: Oh good catch, I didn't realize this would work without it, but I confirmed they still do.

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
```diff
@@ -846,6 +846,38 @@ class PaymentSheetStandardLPMUITests: PaymentSheetUITestCase {
         let successText = app.staticTexts["Success!"]
         XCTAssertTrue(successText.waitForExistence(timeout: 10.0))
     }
+
+    func testFPXPaymentMethodHasDropdown() throws {
```
yuki-stripe, line 850: Hrm I thought I commented on this already, but I don't see it. Is this test useful?  I think FPX is using the same LUXE codepath as other dropdowns and there's nothing unique about it to test here.

yuki-stripe (dismissed):

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
porter-stripe: True, makes sense!

yuki-stripe (approved):

PR #2873, opened 2023-08-23
[Identity] Add blur detector for document scanner

ccen-stripe:
## Summary

Implement a laplacian blurriness detector for document scan. 

Note: ATM we will not really block any image, but upload the blurscore(laplaican variance) to server(using 0 as the blur threshod). Later we analyze historical blur data and experience different blur threshold from server.

## Motivation

The detector would check the blurriness of captured document image on client side and reduce the high blur error rate on server side.

## Testing

Manually tested on iPhone X, iPhone 14pro and iPhone SE(2022)


StripeIdentity/StripeIdentity/Source/API Bindings/DocumentScanner+API.swift
```diff
@@ -11,6 +11,9 @@ import Foundation
 import Vision
 
 extension DocumentScanner.Configuration {
+    // TODO: collect historical data and update the threshold from server.
+    static let defaultBlurThreshold: Decimal = 0.0
```
ccen-stripe, line 15: currently the default value is set to 0, so any number will pass. Different threshold will be sent from server through experiments later

StripeIdentity/StripeIdentity/Source/NativeComponents/Detectors/LaplacianBlurDetector.swift
```diff
@@ -0,0 +1,93 @@
...
+import Accelerate
+import Foundation
+import Metal
+import MetalKit
+import MetalPerformanceShaders
+@_spi(STP) import StripeCameraCore
+import Vision
+
+/// Detector to determine if an image is blurry based on variance of the  laplacian method implemented with Metal.
+final class LaplacianBlurDetector {
```
ccen-stripe, line 17: source: https://betterprogramming.pub/blur-detection-via-metal-on-ios-16dd02cb1558

StripeIdentity/StripeIdentity/Source/API Bindings/Models/VerificationPage/VerificationPageStaticContentDocumentCapturePage.swift
```diff
@@ -25,6 +25,7 @@ extension StripeAPI {
         let motionBlurMinDuration: Int
         let motionBlurMinIou: Decimal
         let requireLiveCapture: Bool
+        let blurThreshold: Decimal?
```
ccen-stripe, line 28: Currently server doesn't send this filed, will be populated later

StripeIdentity/StripeIdentity/Source/NativeComponents/Detectors/LaplacianBlurDetector.swift
```diff
@@ -0,0 +1,93 @@
...
+        self.blurThreshold = blurThreshold
+    }
+
+    func calculateBlurOutput(inputImage: CGImage) -> Output {
+        guard let mtlCommandQueue = self.mtlCommandQueue, let mtlDevice = self.mtlDevice
+        else{
+            return Output(isBlurry: false, variance: 0)
+        }
+        // Create a command buffer for the transformation pipeline
+        let commandBuffer = mtlCommandQueue.makeCommandBuffer()!
```
kgaidis-stripe, line 40: Worth double-checking whether force cast "!" is the correct choice:
1. Is there a 100% certainty that this will never fail?
2. If (1) is uncertain, is the preferred way that a failure is handled is an app crash? 

If not, maybe there's better ways to fail, but I leave that up to the you!

StripeIdentity/StripeIdentity/Source/NativeComponents/Detectors/LaplacianBlurDetector.swift
```diff
@@ -0,0 +1,93 @@
...
+        // These are the two built-in shaders we will use
+        let laplacian = MPSImageLaplacian(device: mtlDevice)
+        let meanAndVariance = MPSImageStatisticsMeanAndVariance(device: mtlDevice)
+        // Load the captured pixel buffer as a texture
+        let textureLoader = MTKTextureLoader(device: mtlDevice)
+        let sourceTexture = try! textureLoader.newTexture(cgImage: inputImage, options: nil)
+        // Create the destination texture for the laplacian transformation
+        let lapDesc = MTLTextureDescriptor.texture2DDescriptor(pixelFormat: sourceTexture.pixelFormat, width: sourceTexture.width, height: sourceTexture.height, mipmapped: false)
+        lapDesc.usage = [.shaderWrite, .shaderRead]
+        let lapTex = mtlDevice.makeTexture(descriptor: lapDesc)!
```
kgaidis-stripe, line 50: Worth double-checking whether force cast "!" is the correct choice:
1. Is there a 100% certainty that this will never fail?
2. If (1) is uncertain, is the preferred way that a failure is handled is an app crash? 

If not, maybe there's better ways to fail, but I leave that up to the you!

StripeIdentity/StripeIdentity/Source/NativeComponents/Detectors/LaplacianBlurDetector.swift
```diff
@@ -0,0 +1,93 @@
...
+        // Encode this as the first transformation to perform
+        laplacian.encode(commandBuffer: commandBuffer, sourceTexture: sourceTexture, destinationTexture: lapTex)
+        // Create the destination texture for storing the variance.
+        let varianceTextureDescriptor = MTLTextureDescriptor.texture2DDescriptor(pixelFormat: sourceTexture.pixelFormat, width: 2, height: 1, mipmapped: false)
+        varianceTextureDescriptor.usage = [.shaderWrite, .shaderRead]
+        let varianceTexture = mtlDevice.makeTexture(descriptor: varianceTextureDescriptor)!
```
kgaidis-stripe, line 56: Worth double-checking whether force cast "!" is the correct choice:
1. Is there a 100% certainty that this will never fail?
2. If (1) is uncertain, is the preferred way that a failure is handled is an app crash? 

If not, maybe there's better ways to fail, but I leave that up to the you!

StripeIdentity/StripeIdentity/Source/NativeComponents/Detectors/LaplacianBlurDetector.swift
```diff
@@ -0,0 +1,93 @@
...
+
+        // Output would is two arrays of 4 values for the mean and variant of [r, g, b, a] channel
+        // return the mean of [r, g, b] as a indicator of blurriness
+        var result = [SIMD4<UInt8>](repeatElement(SIMD4<UInt8>(repeating: 0), count: 2))
+        let region = MTLRegionMake2D(0, 0, 2, 1)
+        varianceTexture.getBytes(&result, bytesPerRow: 1 * 2 * 4, from: region, mipmapLevel: 0)
+
+        let variants = result[0]
+
+        let rgbVariantsSum = Float(variants[0]) + Float(variants[1]) + Float(variants[2])
```
kgaidis-stripe, line 71: Worth double checking whether the number of variants is always 3+ (the index is accessed directly)

kgaidis-stripe (approved):

StripeIdentity/StripeIdentity/Source/NativeComponents/Detectors/LaplacianBlurDetector.swift
ccen-stripe: Added guard check and default value to return

StripeIdentity/StripeIdentity/Source/NativeComponents/Detectors/LaplacianBlurDetector.swift
ccen-stripe: Added guard check and default value to return

StripeIdentity/StripeIdentity/Source/NativeComponents/Detectors/LaplacianBlurDetector.swift
ccen-stripe: Added guard check and default value to return

StripeIdentity/StripeIdentity/Source/NativeComponents/Detectors/LaplacianBlurDetector.swift
ccen-stripe: Since `variants` is of type `SIMD4<UInt8>`, we can be sure that it always has 4 members(hence no out of index access). For my testing on different phones, the CG image always has values in r,g,b channel, with the last a channel being 0. I believe this should be consistant across all use cases, as all Identity images are captured with type . 


Regarding ensuring the variants is always 3+ - most cases the 3 values are the same(e.g `5, 5, 5`), with some cases one channel is slight different from the other two (e.g `5, 5, 4`) - so technically we can't say if a channel has 0 then the triplet is invalid. Therefore I don't think there's any conclusion we could made just by looking at the absolute value of the channels.

kgaidis-stripe (approved):

PR #2896, opened 2023-09-01
Improve PaymentSheetResult.failed docstring, fix accidental force unwrap

yuki-stripe:
## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-1148
https://jira.corp.stripe.com/browse/MOBILESDK-1208

### DIFF
--- a/StripePayments/StripePayments/Source/API Bindings/Models/STPConnectAccountCompanyParams.swift
+++ b/StripePayments/StripePayments/Source/API Bindings/Models/STPConnectAccountCompanyParams.swift
@@ -13,7 +13,7 @@ import Foundation
 public class STPConnectAccountCompanyParams: NSObject {
 
     /// The company’s primary address.
-    @objc public var address: STPConnectAccountAddress!
+    @objc public var address: STPConnectAccountAddress?
 
     /// The Kana variation of the company’s primary address (Japan only).
     @objc public var kanaAddress: STPConnectAccountAddress?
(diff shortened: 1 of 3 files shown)


StripePayments/StripePayments/Source/API Bindings/Models/STPConnectAccountCompanyParams.swift
```diff
@@ -13,7 +13,7 @@ import Foundation
 public class STPConnectAccountCompanyParams: NSObject {
 
     /// The company’s primary address.
-    @objc public var address: STPConnectAccountAddress!
+    @objc public var address: STPConnectAccountAddress?
```
porter-stripe, line 16: Seems like a breaking change?

yuki-stripe: Simply writing to `address` won't break (e.g. `connectAccountCompanyParams.address = ...` will still compile), but other usages could break. I think it's okay to ship without bumping the major version since so few users will be affected and Xcode gives helpful fix-its any users that do get broken by this change. 

I can update the changelog w/ this change though.

porter-stripe: sorry my comment didn't post from last week... +1 to updating the changelog.

porter-stripe (approved):

PR #2906, opened 2023-09-05
EPMs private beta

yuki-stripe:
## Motivation
- https://docs.google.com/document/d/1TYrUZmd-gz1Cu4blWmRQ7NUZwgIi5n16nc7vLvHtSXk/edit#heading=h.cz1xkpga7giy

## Testing
See unit tests, UI tests. Manually tested too.

## Changelog
No changes. We don't want to advertise this to users yet.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
@@ -44,7 +44,7 @@ extension PaymentSheet {
         case dynamic(String)
         case UPI
         case cashApp
-
+        case externalPayPal // TODO(yuki): Replace this when we support more EPMs
         static var analyticLogForIcon: Set<PaymentMethodType> = []
         static let analyticLogForIconSemaphore = DispatchSemaphore(value: 1)
 
@@ -60,11 +60,14 @@ extension PaymentSheet {
                 self = .UPI
             case STPPaymentMethod.string(from: .cashApp):
                 self = .cashApp
+            case "external_paypal":
+                self = .externalPayPal
             default:
                 self = .dynamic(str)
             }
         }
 
+        // I think this returns the Stripe PaymentMethod object type name i.e. a value in https://stripe.com/docs/api/payment_methods/object#payment_method_object-type
         static func string(from type: PaymentMethodType) -> String? {
             switch type {
             case .card:
@@ -81,6 +84,8 @@ extension PaymentSheet {
                 return STPPaymentMethod.string(from: .cashApp)
             case .dynamic(let str):
                 return str
+            case .externalPayPal:
+                return nil
             }
         }
         var displayName: String {
@@ -97,17 +102,21 @@ extension PaymentSheet {
             } else if case .dynamic(let name) = self {
                 // TODO: We should introduce a display name in our model rather than presenting the payment method type
                 return name
+            } else if case .externalPayPal = self {
+               return STPLocalizedString("PayPal", "Payment Method type brand name")
             }
             assertionFailure()
             return ""
         }
 
         /// The identifier for the payment method type as it is represented on an intent
-        var identifier: String {
-            if let stpPaymentMethodType = stpPaymentMethodType {
+        var identifier: String { // this should be renamed to analyticsIdentifier
+            if let stpPaymentMethodType {
                 return stpPaymentMethodType.identifier
             } else if case .dynamic(let name) = self {
                 return name
+            } else if case .externalPayPal = self {
+                return "external_paypal"
             }
 
             assertionFailure()
@@ -148,6 +157,9 @@ extension PaymentSheet {
                     STPAnalyticsClient.sharedClient.logImageSelectorIconFromBundleIfNeeded(paymentMethod: self)
                 }
                 return stpPaymentMethodType.makeImage(forDarkBackground: forDarkBackground)
+            }
+            if case .externalPayPal = self {
+                return STPPaymentMethodType.payPal.makeImage(forDarkBackground: forDarkBackground)
             }
             if PaymentSheet.PaymentMethodType.shouldLogAnalytic(paymentMethod: self) {
                 STPAnalyticsClient.sharedClient.logImageSelectorIconNotFoundIfNeeded(paymentMethod: self)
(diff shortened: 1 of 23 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
```diff
@@ -44,7 +44,7 @@ extension PaymentSheet {
         case dynamic(String)
         case UPI
         case cashApp
-
+        case externalPayPal // TODO(yuki): Replace this when we support more EPMs
```
eurias-stripe, line 47: Would it be better to have a `.external(String)` case? I assume something like that would be the final shape of this, and I would prefer not to add a case that will be replaced later.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
```diff
@@ -245,6 +272,27 @@ extension PaymentSheet {
 
                 return availabilityStatus == .supported
             }
+
+            if let paymentMethodOrder = configuration.paymentMethodOrder?.map({ $0.lowercased() }) {
```
eurias-stripe, line 276: Does this need to be done in the client? Does the endpoint not take the `paymentMethodOrder` param and return them already ordered?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
```diff
@@ -502,6 +504,12 @@ extension PaymentSheetFormFactory {
         )
     }
 
+    func makeExternalPayPal() -> PaymentMethodElement {
+        let contactInfoSection = makeContactInformationSection(nameRequiredByPaymentMethod: false, emailRequiredByPaymentMethod: false, phoneRequiredByPaymentMethod: false)
+        let billingDetails = makeBillingAddressSectionIfNecessary(requiredByPaymentMethod: false)
```
eurias-stripe, line 509: You need to call `connectBillingDetailsFields` so the country and phone country elements change together.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetLPMConfirmFlowTests.swift
```diff
@@ -46,6 +47,30 @@ final class PaymentSheet_LPM_ConfirmFlowTests: XCTestCase {
         await PaymentSheetLoader.loadMiscellaneousSingletons()
     }
 
+    /// 👋 👨‍🏫  Look at this test to understand how to write your own tests in this file
+    func testiDEALConfirmFlows() async throws {
+        try await _testConfirm(intentKinds: [.paymentIntent], currency: "EUR", paymentMethodType: .dynamic("ideal")) { form in
```
eurias-stripe, line 52: Side note: Could this make this test work with PMs that use redirects? I tried to use it for Alipay but it failed since it could never load the web view and thus never completed.

eurias-stripe (changes requested):
Overall looks good, couple of comments.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
yuki-stripe: It might be what we we end up with in the final version, but I don't want to preemptively do that now.  If I made that change, I'd have to deal with arbitrary strings wherever I use this type, but this implementation can't handle arbitrary external PMs - it only works for paypal.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
yuki-stripe: It doesn't, although I agree it should be done on the endpoint!  I'm hoping to get that into the server implementation of this feature for GA.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
yuki-stripe: I saw that but didn't think it made sense to keep, at least as it currently works. It only updates the phone field based on the country field.  iirc I made this behavior for Link, where the phone number is collected *after* the country.  

Outside of Link's phone field, though, we collect phone *before* country, so this code almost always has no effect.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
eurias-stripe: Fair

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetLPMConfirmFlowTests.swift
yuki-stripe: Yeah it does! Check out https://github.com/stripe/stripe-ios/pull/2906/files/8887046aa5b6af42adaa9436ff94d3f433f0876e#diff-a026e64b444d61822aedf6769b182760b2b8e028a95358d7ccef4cb7dc83c2d6R305-R310 - if we see the redirect we consider it a success.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
eurias-stripe: Interesting, I only added that _because_ it was already done in the address element, I'm okay not using it anymore.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetLPMConfirmFlowTests.swift
yuki-stripe: Oh you tried it for Alipay already and it didn't work?  I'm guessing the _redirectShim isn't called, then, maybe because LUXE takes a different path?  It's definitely a brittle approach to use that shim.

eurias-stripe: Hmm, then I wonder why it didn't work for me, I'll try it again

eurias-stripe (approved):

eurias-stripe (approved):

PR #2907, opened 2023-09-06
Add support for OXXO

eurias-stripe:
## Summary
Added support for OXXO

## Motivation
LPM Sprint.

## Testing

https://github.com/stripe/stripe-ios/assets/114543737/0905a2d6-61c0-4106-b10d-566e14f8f18f

Added UI Test

## Changelog
[Added] Support for OXXO with PaymentIntents.


porter-stripe:
I believe we should also be adding an OXXO png to the CDN. https://trailhead.corp.stripe.com/docs/luxe/add-luxe-spec#stripe-js--pay-server-add-the-logo-image

Same for Alipay.

cc @wooj-stripe

eurias-stripe:
> I believe we should also be adding an OXXO png to the CDN. https://trailhead.corp.stripe.com/docs/luxe/add-luxe-spec#stripe-js--pay-server-add-the-logo-image
> 
> Same for Alipay.
> 
> EDIT: nvm maybe not OXXO since it's not using LUXE.
> 
> cc @wooj-stripe

Alipay was already there, but it didn't have the right link in the spec, I already fixed that, and yeah, oxxo is not using LUXE so no need, although the svg is there, I used that to make a png

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
```diff
@@ -25,6 +25,8 @@ extension PaymentSheet {
                 return [.returnURL]
             case .dynamic("zip"):
                 return [.returnURL]
+            case .dynamic("alipay"):
```
yuki-stripe, line 28: We already defined this on Line 26.  I think we should try to avoid `.dynamic` if the type already a STPPaymentMethodType

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
```diff
@@ -94,6 +96,10 @@ extension PaymentSheet {
                 return "Zip"
             } else if case .dynamic("amazon_pay") = self {
                 return "Amazon Pay"
+            } else if case .dynamic("alipay") = self {
+                return "Alipay"
+            } else if case .dynamic("oxxo") = self {
+                return "OXXO"
```
yuki-stripe, line 102: aren't these already defined by line 90?

yuki-stripe (commented):
Can we add tests for these in `PaymentSheet_LPM_ConfirmFlowTests` too?

yuki-stripe:
oh also this PR adds Alipay too, is that intentional?

eurias-stripe:
> oh also this PR adds Alipay too, is that intentional?

This PR is on top of the alipay PR, once alipay is merged I will rebase this.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
eurias-stripe: Ah! didn't notice that, I was just following copy pasting

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
eurias-stripe: Oh, I didn't notice this was only for PMs not represented by STPPaymentMethodType, is not clear unless you follow the call

yuki-stripe:
> > oh also this PR adds Alipay too, is that intentional?
> 
> This PR is on top of the alipay PR, once alipay is merged I will rebase this.

I see, can you set the base branch to your alipay branch instead of master?  That way we don't see code from the alipay PR in the diff

eurias-stripe:
> Can we add tests for these in `PaymentSheet_LPM_ConfirmFlowTests` too?

I added the confirm flow test for OXXO and to Alipay while I was at it, I found the issue that caused the test not to work initially and fixed it

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetLPMConfirmFlowTests.swift
```diff
@@ -146,6 +149,26 @@ final class PaymentSheet_LPM_ConfirmFlowTests: XCTestCase {
             XCTAssertEqual(form.getAllSubElements().count, 1)
         }
     }
+
+    func testAlipayConfirmFlows() async throws {
+        try await _testConfirm(intentKinds: [.paymentIntent],
+                               currency: "USD",
+                               paymentMethodType: .dynamic("alipay"),
```
yuki-stripe, line 156: such a code smell that we do `.dynamic("alipay")`. What does it mean that this is dynamic?? I don't think we can change anything in this PR, just pointing it out.

yuki-stripe (approved):

PR #2909, opened 2023-09-08
Add Bacs Direct Debit support to PaymentSheet

davidme-stripe:
## Summary
* Add a view to show the Bacs mandate information before confirm. In Stripe.js, this modal is handled by the SDK instead of the PaymentIntent's next_actions, so we'll do something similar here.
* Enable Bacs in PaymentSheet, including form fields.
* Refactor BSBNumber.swift to handle multiple types of bank account routing numbers.

## Motivation
Adding Bacs Direct Debit support to PaymentSheet

## Testing

- [ ] Add snapshot test for the Mandate view
- [ ] Add PaymentSheet_LPM_ConfirmFlowTests test
- [ ] Livemode PaymentIntent/SetupIntent/SFU tests


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
```diff
@@ -36,6 +36,7 @@ extension PaymentSheet {
         case dynamic(String)
         case UPI
         case cashApp
+        case bacsDebit
```
porter-stripe, line 39: Do we need to do this or can we use the `STPPaymentMethodType` instead?

porter-stripe, line 39: Do we need to do this or can we use the STPPaymentMethodType instead?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/BacsDDMandateView.swift
```diff
@@ -0,0 +1,167 @@
...
+// Reusing these colors from Link:
+private let headerColor = Color(UIColor.linkPrimaryText)
+private let secondaryTextColor = Color(UIColor.linkSecondaryText)
+private let borderColor = Color(UIColor.linkSeparator)
+private let backgroundColor = Color(UIColor.linkSecondaryButtonBackground)
+
+private let shadow1Color = Color(red: 18/255, green: 42/255, blue: 66/255, opacity: 0.04)
+private let shadow2Color = Color(red: 0/255, green: 0/255, blue: 0/255, opacity: 0.04)
+
+struct BacsDDMandateView: View {
```
porter-stripe, line 21: This view should get it's colors from the appearance API imo

davidme-stripe: We could but I'm a little worried about it, maybe unnecessarily. PE on the web doesn't, and they had some pretty strict warnings around the design, which I interpreted as including the specific contrast of the colors and such.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
davidme-stripe: We use this in `performLocalActionsIfNeededAndConfirm` — would it make more sense to check the `confirmParams.paymentMethodParams.type` instead? Not totally familiar with whether that's populated correctly in the deferred flow, so I was trying to follow the pattern I saw in `PaymentMethodType`.

porter-stripe: That's a good callout for the deferred flow, I'm not sure what our stance on `PaymentMethodType` vs `STPPaymentMethodType` is. I know we want to avoid `dynamic` but that's not being used here.

porter-stripe (dismissed):

porter-stripe (approved):

PR #2912, opened 2023-09-11
Add PayNow support to PaymentSheet

porter-stripe:
## Summary
- Adds PayNow support to PaymentSheet for PaymentIntents.
- We present a web view, and behind it we present the PollingViewController which handles the polling.

#### Success
https://github.com/stripe/stripe-ios/assets/88012362/3969cb84-4f98-489e-a708-f13cef467972
#### Expired
https://github.com/stripe/stripe-ios/assets/88012362/a4d3cede-d66e-4e02-a59c-3038124cb980
#### Cancel and pay another way
https://github.com/stripe/stripe-ios/assets/88012362/f8480605-c399-42d8-acf9-d12d3bb85ee8

#### Success
https://github.com/stripe/stripe-ios/assets/88012362/25b4bcc5-d67f-4b9e-86da-792231fca4a3
#### Expired
https://github.com/stripe/stripe-ios/assets/88012362/f3cb0f03-45de-4072-9e3b-c1ea3fa4dc8a

#### Success
https://github.com/stripe/stripe-ios/assets/88012362/2539c0b2-318b-49b1-b673-7a829c74ec9f
#### Expired
https://github.com/stripe/stripe-ios/assets/88012362/ef4585ca-faa6-470b-9647-4d48b5c823f1

## Motivation
- LPM sprint

## Testing
- Manual
- New unit tests

## Changelog
See diff


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PollingViewController.swift
```diff
@@ -275,6 +278,9 @@ class PollingViewController: UIViewController {
             self.navigationBar.setStyle(.back)
             self.intentPoller.suspendPolling()
             self.oneSecondTimer?.invalidate()
+
+            // If the intent is canceled while a web view is presented, we must dismiss it before we can complete the action with .canceled so STPPaymentHandler can properly update it's state
```
yuki-stripe, line 282:
```suggestion
            // If the intent is canceled while a web view is presented, we must dismiss it before we can complete the action with .canceled so STPPaymentHandler can properly update its state

Comment (yuki-stripe) [reply StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PollingViewController.swift:282]: what happens if you don't dismiss?  Should we call `self.currentAction.complete()` in the completion block of the dismiss?

File: StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PollingViewModel.swift [hunk 2]
```diff
@@ -32,6 +34,18 @@ class PollingViewModel {
...
+        default:
+            fatalError("Polling deadline has not been implemented for \(paymentMethodType)")
+        }
+    }()
+    lazy var maxRetries: Int = {
+        switch paymentMethodType {
+        case .UPI, .blik:
+            return 12
+        case .paynow:
+            return 1000 // PayNow has a very long expiration, allow a high number of retries
```
yuki-stripe, line 48: What happens after maxRetries is hit?

yuki-stripe: It looks like we do exponential backoff here, is that right? If so, why? Shouldn't we be polling at regular intervals (probably 1 second, like Checkout)?  Exponential backoff would mean we'd soon be waiting incredibly long before updating the PI.

StripePayments/StripePayments/Source/API Bindings/Models/PaymentMethods/Types/STPPaymentMethodPayNow.swift
```diff
@@ -0,0 +1,40 @@
...
+//  StripePayments
+//
+//  Created by Nick Porter on 9/11/23.
+//
+
+import Foundation
+
+/// A PayNow PaymentMethod
+/// - seealso: https://stripe.com/docs/api/payment_methods/object#payment_method_object-paynow
+public class STPPaymentMethodPayNow: NSObject, STPAPIResponseDecodable {
```
yuki-stripe, line 12: Since this is empty, let's not add this.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PollingViewController.swift
yuki-stripe: Hrm actually the fact that this class dismisses `STPPaymentHandler`'s safariVC and mutates its `currentAction` is a code smell IMO b/c it tightly couples these two classes together.

StripePaymentSheet/StripePaymentSheet/Resources/JSON/form_specs.json
```diff
@@ -957,5 +957,16 @@
...
+    {
+      "type": "paynow",
+      "async": false,
+      "selector_icon": {
+        "light_theme_png": "https://js.stripe.com/v3/fingerprinted/img/payment-methods/icon-pm-paynow@3x-06fb143923e8492b0d3fe379ad6692a9.png",
+        "light_theme_svg": "https://js.stripe.com/v3/fingerprinted/img/payment-methods/icon-pm-paynow-b25acf96aac928b13e6206afb8982444.svg",
+        "dark_theme_png": "https://js.stripe.com/v3/fingerprinted/img/payment-methods/icon-pm-paynow_dark@3x-d49fe128abf7fe78fdfa939c445ca2cc.png",
+        "dark_theme_svg": "https://js.stripe.com/v3/fingerprinted/img/payment-methods/icon-pm-paynow_dark-d700c844535c608fd5fd4048f6a21146.svg"
+      },
+      "fields": []
```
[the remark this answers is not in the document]
porter-stripe: Is there a line you are referring to or the entire spec? I guess for either questions it is a direct copy of the spec as it exists in pay-server for the other surfaces. This is a fallback spec AFAIK, we do this for all luxe LPMs.
https://livegrep.corp.stripe.com/view/stripe-internal/pay-server/lib/lpm_ui_platform/private/specs/paynow.json

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PollingViewModel.swift
porter-stripe: Exponential backoff was the original design decision for polling in mobile, which started with UPI. https://docs.google.com/document/d/1HbbLaf0sJ869BZmBoxsfxYjG31ymBk-Zdrl9g_5s62U/edit

I'm open to simplifying this by
1. Retrying on a short interval like 1 second for all polling LPMs
2. Removing retry count in favor of just using the count down timer. IIRC retry count is needed for the exponential back off formula.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PollingViewController.swift
porter-stripe: We could define a delegate (`PollingViewControllerDelegatge`) that `STPPaymentHandler` implements which mutates the current action and handles the dismissal.

StripePaymentSheet/StripePaymentSheet/Resources/JSON/form_specs.json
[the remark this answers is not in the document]
porter-stripe: cc @wooj-stripe for a better explication on why we have a local copy of all specs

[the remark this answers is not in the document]
yuki-stripe: The entire spec - it's unused by the client, right?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PollingViewModel.swift
yuki-stripe: Making the customer wait exponential amounts of time before they can finish checking out for the sake of battery/network conservation isn't the right tradeoff IMO, especially given that we're only polling when the customer is in the app, staring at the spinner. Retrying every second seems fine, and if we wanted to do some backoff, it should slowly go to a max of like every 10 seconds.


Also the retry code looks broken today, although I think we should rewrite it. 
```
pow(Double(1 + maxRetries - retryCount), Double(2))
```

At maxRetries = 1000, retryCount = 0, this gives a delay of 1001^2 = a million seconds and grows smaller as we retry.

StripePaymentSheet/StripePaymentSheet/Resources/JSON/form_specs.json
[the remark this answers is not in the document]
porter-stripe: I don't think it's unused, I believe we have tests that depend on it and if we fail to fetch one from pay-server we use it. cc @wooj-stripe correct me if I'm wrong.

But this is pretty standard for all the other luxe LPMs, nothing different for PayNow in this PR.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PollingViewModel.swift
porter-stripe: @yuki-stripe I updated the polling logic to:
1. Poll every second
2. Remove the max retry count

I re-ran all the tests for PayNow and the experience is a bit better now, much snappier. Same for BLIK and UPI.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PollingViewController.swift
porter-stripe: thoughts on a `PollingViewControllerDelegate`?

yuki-stripe: I think that makes sense! Maybe tackle in a separate PR?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PollingViewModel.swift
porter-stripe: Would this also mean the retry logic for 3DS2 is broken? It's the same formula: https://github.com/stripe/stripe-ios/blob/master/StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift

yuki-stripe: No, but it's not obvious at all.  `retryCount` is the wrong name in that file.  It starts equal to `maxRetries` and goes _down_.  It should really be called `retriesRemaining`.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PollingViewController.swift
```diff
@@ -290,10 +290,7 @@ class PollingViewController: UIViewController {
         // Do one last force poll after deadline
         DispatchQueue.main.asyncAfter(deadline: .now() + 3) { [weak self] in
             guard let self = self else { return }
-            self.intentPoller.forcePoll()
-            // If we don't get a terminal status back after 20 seconds from the previous force poll, set error state to suspend polling.
-            // This could occur if network connections are unreliable
-            DispatchQueue.main.asyncAfter(deadline: .now() + 20, execute: self.setErrorStateWorkItem)
+            self.intentPoller.fetchStatus()
```
yuki-stripe, line 293: Does this actually stop polling?  I'm not seeing how it could.

StripePaymentSheet/StripePaymentSheet/Resources/JSON/form_specs.json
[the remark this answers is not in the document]
yuki-stripe: 🤷 ok

StripePaymentSheet/StripePaymentSheet/Source/Helpers/IntentStatusPoller.swift
```diff
@@ -15,12 +15,11 @@ protocol IntentStatusPollerDelegate: AnyObject {
 }
 
 class IntentStatusPoller {
```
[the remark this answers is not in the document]
porter-stripe: Agreed, added!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PollingViewController.swift
porter-stripe: Not really, the name of this function is misleading. Previously we depended on `viewDidDisappear` to stop polling, but honestly this function should also stop polling so I've added a call to suspendPolling in this function.

StripePaymentSheet/StripePaymentSheet/Source/Helpers/IntentStatusPoller.swift
```diff
@@ -30,17 +29,17 @@ class IntentStatusPoller {
         didSet {
             // Start polling if we weren't already polling
             if !oldValue && isPolling {
-                forcePoll()
+                fetchStatus()
             } else if !isPolling {
                 nextPollWorkItem?.cancel()
```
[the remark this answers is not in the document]
porter-stripe: Very good catch, our new test file would have caught this now.

porter-stripe:
@yuki-stripe thanks for the reviews, this should be ready now.

StripePaymentSheet/StripePaymentSheet/Source/Helpers/IntentStatusPoller.swift
```diff
@@ -30,72 +35,69 @@ class IntentStatusPoller {
...
     // MARK: Public APIs
 
+    /// Begins continuously polling at a frequency defined by `retryInterval`. The delegate will be
+    /// notified of any status changes to the intent during this poll period.
     public func beginPolling() {
         isPolling = true
     }
 
+    /// Suspends the ongoing polling process. After suspending, the delegate will not receive any status
+    /// updates on the intent until polling is resumed using `beginPolling()`.
```
yuki-stripe, line 60: Love these docstrings.

StripePaymentSheet/StripePaymentSheet/Source/Helpers/IntentStatusPoller.swift
```diff
@@ -30,72 +35,69 @@ class IntentStatusPoller {
...
-    private func retryWithExponentialDelay(retryCount: Int, block: @escaping () -> Void) {
-        // Add some backoff time
-        let delayTime = TimeInterval(
-            pow(Double(1 + maxRetries - retryCount), Double(2))
-        )
-
+    private func retryAfterInterval(block: @escaping () -> Void) {
         nextPollWorkItem = DispatchWorkItem {
             block()
         }
```
yuki-stripe, line 98: nit: Taking a block seems unnecessary; just call `fetchStatus()` directly instead.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/IntentStatusPollerTest.swift
```diff
@@ -0,0 +1,115 @@
...
+        mockIntentRetriever = MockPaymentIntentRetriever()
+        mockDelegate = MockIntentStatusPollerDelegate()
+        sut = IntentStatusPoller(retryInterval: 1.0, intentRetriever: mockIntentRetriever, clientSecret: "test_client_secret")
+        sut.delegate = mockDelegate
+    }
+
+    override func tearDown() {
+        sut = nil
+        mockIntentRetriever = nil
+        mockDelegate = nil
```
yuki-stripe, line 30: A new instance of XCTestCase is used for each test, so this is unnecessary.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/IntentStatusPollerTest.swift
```diff
@@ -0,0 +1,115 @@
...
+    var mockIntentRetriever: MockPaymentIntentRetriever!
+    var mockDelegate: MockIntentStatusPollerDelegate!
+    var intentRetrieverExpectation: XCTestExpectation!
+    var delegateExpectation: XCTestExpectation!
+
+    override func setUp() {
+        super.setUp()
+        mockIntentRetriever = MockPaymentIntentRetriever()
+        mockDelegate = MockIntentStatusPollerDelegate()
+        sut = IntentStatusPoller(retryInterval: 1.0, intentRetriever: mockIntentRetriever, clientSecret: "test_client_secret")
```
yuki-stripe, line 23: nit: Might as well use a smaller interval for this test?  Like 0.1

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
```diff
@@ -1604,11 +1621,19 @@ public class STPPaymentHandler: NSObject {
     /// 1. Redirects to an app using url
     /// 2. Open fallbackURL in a webview if 1) fails
             ///
-    func _handleRedirect(to nativeURL: URL?, fallbackURL: URL?, return returnURL: URL?) {
+    func _handleRedirect(to nativeURL: URL?, fallbackURL: URL?, return returnURL: URL?, completion: ((SFSafariViewController?) -> Void)? = nil) {
```
yuki-stripe, line 1624: Can you add a docstring explaining `completion`?

StripePaymentSheet/StripePaymentSheet/Source/Helpers/IntentStatusPoller.swift
```diff
@@ -30,72 +35,69 @@ class IntentStatusPoller {
...
+    /// updates on the intent until polling is resumed using `beginPolling()`.
     public func suspendPolling() {
         isPolling = false
     }
 
-    public func forcePoll() {
-        fetchStatus(forcePoll: true)
+    /// Triggers a single poll operation to fetch the intent status right away, regardless of the
+    /// retryInterval. Whether or not continuous polling in ongoing, `pollOnce()` immediately updates the
+    /// status and informs the delegate if a change in status takes place.
```
yuki-stripe, line 67: How does the user of this class know when polling is completed? 

Like, when the timer runs out, how does PollingViewController know that the last network request completed and it should dismiss itself?

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/IntentStatusPollerTest.swift
```diff
@@ -0,0 +1,115 @@
...
+        mockIntentRetriever.mockedStatus = .succeeded
+
+        wait(for: [intentRetrieverExpectation, delegateExpectation], timeout: 5.0)
+        // delegate should not have been notified of succeeded since polling was suspended
+        XCTAssertEqual(mockDelegate.latestPaymentIntent?.status, .requiresPaymentMethod)
+
+        // Resume polling
+        setExpectations(apiExpectedCount: 1, delegateExpectedCount: 1)
+        sut.beginPolling()
+        XCTAssertTrue(sut.isPolling)
```
yuki-stripe, line 74: nit: if we have `beginPolling` and `suspendPolling`, do we need to expose `isPolling` publicly?  Seems like an implementation detail.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/IntentStatusPollerTest.swift
```diff
@@ -0,0 +1,115 @@
...
+
+        // Make the API client return succeeded, then the delegate should NOT be notified, and API should NOT be called due to suspended polling
+        setExpectations(apiExpectedCount: 1, delegateExpectedCount: 1)
+        intentRetrieverExpectation.isInverted = true // expectations should not be fufilled since polling is suspended
+        delegateExpectation.isInverted = true
+        mockIntentRetriever.mockedStatus = .succeeded
+
+        wait(for: [intentRetrieverExpectation, delegateExpectation], timeout: 5.0)
+        // delegate should not have been notified of succeeded since polling was suspended
+        XCTAssertEqual(mockDelegate.latestPaymentIntent?.status, .requiresPaymentMethod)
```
yuki-stripe, line 69: Does `wait` here wait 5 seconds for the expectations, or does it just immediately succeed?

b/c we should be waiting like `interval + buffer` seconds before validating that the poller didn't make network requests etc.

If we don't want to fiddle with expectations, I think simply waiting and then validating that `lastPaymentIntent.status` didn't update would work.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/IntentStatusPollerTest.swift
porter-stripe: It should only be `internal`

`var isPolling: Bool = false ...`

StripePaymentSheet/StripePaymentSheet/Source/Helpers/IntentStatusPoller.swift
porter-stripe: The user of this class doesn't know when polling is completed since that is up to the user/timer. The idea here is the IntentStatusPoller just polls until you ask it to stop. So you keep polling until you have received the status you are looking for or want to cancel, in which case you call `suspendPolling`.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/IntentStatusPollerTest.swift
yuki-stripe: I mean it should be private - purely an implementation detail.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/IntentStatusPollerTest.swift
porter-stripe: In a functional sense they are pretty much succeeding right away since now our retry interval is 0.1 seconds. If the expectations aren't being fulfilled it will wait here for the full 5 seconds. But since they are being fulfilled so fast it's not really waiting. I prefer this approach over `wait(1)` or something as it waits the minimum amount of time, and just feels a bit better to me.

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
porter-stripe: Added!

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/IntentStatusPollerTest.swift
porter-stripe: Reduced to 0.1

StripePaymentSheet/StripePaymentSheet/Source/Helpers/IntentStatusPoller.swift
porter-stripe: Good point, updated.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/IntentStatusPollerTest.swift
porter-stripe: `@testable` is allowing this to be accessed.

StripePaymentSheet/StripePaymentSheet/Source/Helpers/IntentStatusPoller.swift
yuki-stripe: Well let's say I'm the PollingVC and I want to poll for 60 seconds and then dismiss and return to PaymentSheet.  

After 60 seconds, how do I know when the final request has completed so I can dismiss myself?  

In fact, looking at the code more, it doesn't appear to dismiss at all when the timer expires. It only dismisses when the PI status changes or you cancel out?

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/IntentStatusPollerTest.swift
porter-stripe: Fair, I made it private here.

yuki-stripe: I'm suggesting `isPolling` should be private (or fileprivate) and not internal.  `internal` makes this property accessible to anything inside the same module.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/IntentStatusPollerTest.swift
yuki-stripe: If we have a bug in our poller where it continues to poll after you call `suspendPolling`, would this catch it?  If I'm understanding, it wouldn't - we _immediately_ check that the poller didn't make a network request or update the PI and succeed, instead of waiting until the next `interval`.

StripePaymentSheet/StripePaymentSheet/Source/Helpers/IntentStatusPoller.swift
porter-stripe: The original thought was that if the timer expires on the client, then on the server the intent would move out of `requires_action` back into `requires_payment_method` which would notify the delegate, in which case we show the error/timed out UI in `didUpdate(paymentIntent: STPPaymentIntent)`. But it probably makes sense to show that UI if the final poll doesn't come back as succeeded as a safe guard. I'm going to add a completion handler to `pollOnce` and add that logic.

https://www.figma.com/file/C6FkqGHZuSzR6S0KoseV0T/UPI-Payment?type=design&node-id=24-2762&mode=design&t=gOIJDDZFqkZ3Z8qX-0

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/IntentStatusPollerTest.swift
porter-stripe: Oop, I was looking at the wrong `wait`. I was referencing the one on line 66.

The `wait` on line 58, for the inverted expectations does wait the full five seconds, ensuring they are not fulfilled. Which ensures we aren't making any API requests or invoking the delegate while suspended.

I added some print statements around it
```
started waiting 1694727167.03133
ended waiting 1694727172.035596
```

yuki-stripe: Ah good - then we should change the timeout to be something like `interval*2` (x2 to allow buffer for the timer).

porter-stripe: Done!

yuki-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/Helpers/IntentStatusPoller.swift
yuki-stripe: I think the change you made fixes the problem I'm describing.

PR #2916, opened 2023-09-12
Make FormSpecProvider skip unparsable specs

eurias-stripe:
## Summary
Currently, if a spec fails to parse, the FormSpecProvider will skip all remaining specs and report an error, this is undesirable as the rest of the specs should still be loaded.
This PR changes the behavior to skip failing specs and report a different error for more accurate observability.

## Motivation
ir-powering-solitary

## Testing
Added a test

### DIFF
--- a/Stripe/StripeiOSTests/FormSpecProviderTest.swift
+++ b/Stripe/StripeiOSTests/FormSpecProviderTest.swift
@@ -434,7 +434,7 @@ class FormSpecProviderTest: XCTestCase {
         let decodedFormSpecs = try decoder.decode([FormSpec].self, from: formSpec)
 
         let sut = FormSpecProvider()
-        XCTAssertFalse(sut.containsUnknownNextActions(formSpecs: decodedFormSpecs))
+        XCTAssertFalse(sut.containsUnknownNextActions(formSpec: decodedFormSpecs[0]))
     }
 
     func testContainsUnknownNextAction_confirm() throws {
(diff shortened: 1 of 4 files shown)


Stripe/StripeiOSTests/FormSpecProviderTest.swift
```diff
@@ -574,4 +574,75 @@ class FormSpecProviderTest: XCTestCase {
...
+
+        // Verify that the affirm spec was not overridden.
+        let affirmSpec = sut.formSpec(for: "affirm")
+        XCTAssertEqual(affirmSpec?.fields.count, 1)
+        XCTAssertEqual(affirmSpec?.fields[0], .affirm_header)
+
+        // Verify that the klarna spec is successfully replaced.
+        let klarnaSpec = sut.formSpec(for: "klarna")
+        XCTAssertEqual(klarnaSpec?.fields.count, 1)
+        XCTAssertEqual(klarnaSpec?.fields[0], .klarna_header)
```
wooj-stripe, line 646: Can we check to make sure that eps does not get overridden either?

wooj-stripe (approved):
lgtm!

davidme-stripe (approved):

porter-stripe (approved):

Stripe/StripeiOSTests/FormSpecProviderTest.swift
```diff
@@ -574,4 +574,92 @@ class FormSpecProviderTest: XCTestCase {
         XCTAssertEqual(redirectToURL.urlPath, "next_action[redirect_to_url][url]")
         XCTAssertEqual(redirectToURL.returnUrlPath, "next_action[redirect_to_url][return_url]")
     }
+
+    func testUnparsableSpecIsSkipped() throws {
+        let e = expectation(description: "Loads form specs file")
+        let sut = FormSpecProvider()
```
sfriedman-stripe, line 580: What does `sut` stand for? Can we use a more descriptive name?

sfriedman-stripe: Looking above, seems like this is common in the rest of the tests, but it's still not a term of art I've ever heard before

eurias-stripe: Means [System Under Test](https://en.wikipedia.org/wiki/System_under_test), it is a common term.

PR #2920, opened 2023-09-13
Replace makeDefaultsApplierWrapper with simpler implementation to app…

yuki-stripe:
…ly defaults when attachDefaultsToPaymentMethod is true.

## Motivation
Simplifies the implementation and makes it apply universally to all PMs instead of having each PM need to call `makeDefaultsApplierWrapper`.

## Testing
Relying on existing tests.

## Changelog
Not user facing.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/IntentConfirmParams.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/IntentConfirmParams.swift
@@ -93,6 +93,36 @@ class IntentConfirmParams {
 
         return params
     }
+
+    /// Applies the values of `Configuration.defaultBillingDetails` to this IntentConfirmParams if `attachDefaultsToPaymentMethod` is true.
+    /// - Note: This overwrites `paymentMethodParams.billingDetails`.
+    func setDefaultBillingDetailsIfNecessary(for configuration: PaymentSheet.Configuration) {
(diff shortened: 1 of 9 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/IntentConfirmParams.swift
```diff
@@ -93,6 +93,36 @@ class IntentConfirmParams {
...
+        _setDefaultBillingDetailsIfNecessary(defaultBillingDetails: configuration.defaultBillingDetails, billingDetailsCollectionConfiguration: configuration.billingDetailsCollectionConfiguration)
+    }
+
+    /// Applies the values of `Configuration.defaultBillingDetails` to this IntentConfirmParams if `attachDefaultsToPaymentMethod` is true.
+    /// - Note: This overwrites `paymentMethodParams.billingDetails`.
+    func setDefaultBillingDetailsIfNecessary(for configuration: CustomerSheet.Configuration) {
+        _setDefaultBillingDetailsIfNecessary(defaultBillingDetails: configuration.defaultBillingDetails, billingDetailsCollectionConfiguration: configuration.billingDetailsCollectionConfiguration)
+    }
+
+    private func _setDefaultBillingDetailsIfNecessary(defaultBillingDetails: PaymentSheet.BillingDetails, billingDetailsCollectionConfiguration: PaymentSheet.BillingDetailsCollectionConfiguration) {
```
eurias-stripe, line 109: nit: there is no need to prefix the method name with _

eurias-stripe (approved):

eurias-stripe (approved):

PR #2926, opened 2023-09-14
Add EPM private beta to changelog

yuki-stripe:
Follow up to https://github.com/stripe/stripe-ios/pull/2906


porter-stripe (approved):
next version will be amazing with all these entries

PR #2929, opened 2023-09-14
PaymentSheet Example: Add URL sharing support

davidme-stripe:
## Summary
* Adds support for opening PaymentSheet Example URLs via a scheme
* Buttons to share URLs via QR codes or the share sheet
* Fixes an issue where we would sometimes fail to update the publishable key, as we were accidentally mixing updating defaultPublishableKey with the STPAPIClient's publishableKey. The former will not change the key if it was already set on the singleton STPAPIClient.

## Motivation
Easier sharing of scenarios and on-device testing

## Testing
Tested locally

## Changelog
Not user-facing

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlaygroundSettings.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlaygroundSettings.swift
@@ -272,9 +272,9 @@ struct PaymentSheetTestPlaygroundSettings: Codable, Equatable {
         let jsonData = try! JSONEncoder().encode(self)
         return jsonData.base64EncodedString()
     }
-    
+
     var base64URL: URL {
-        URL(string: "stp-paymentsheet-playground:/?\(base64Data)")!
+        URL(string: "stp-paymentsheet-playground://?\(base64Data)")!
     }
 
     static func fromBase64<T: Decodable>(base64: String, className: T.Type) -> T? {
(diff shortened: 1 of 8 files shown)


Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlaygroundSettings.swift
```diff
@@ -272,6 +272,10 @@ struct PaymentSheetTestPlaygroundSettings: Codable, Equatable {
         let jsonData = try! JSONEncoder().encode(self)
         return jsonData.base64EncodedString()
     }
+    
+    var base64URL: URL {
+        URL(string: "stp-paymentsheet-playground:/?\(base64Data)")!
```
eurias-stripe, line 277: Are you missing a `/` after the scheme?

eurias-stripe (dismissed):

Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlaygroundSettings.swift
davidme-stripe: I usually drop it for these sort of internal URLs, but I guess the spec requires it

davidme-stripe:
@eurias-stripe Forgot to lint, also merged the share sheet into the QR code button

eurias-stripe (approved):

PR #2950, opened 2023-09-20
Add support for swish

eurias-stripe:
## Summary
Add support for swish

## Motivation
LPM Sprint ... extended!

## Testing
Added confirm flow tests and UI Tests

## Changelog
* [Added] Support for Swish with PaymentIntents.


Example/Non-Card Payment Examples/Non-Card Payment Examples/SwishExampleViewController.swift
```diff
@@ -0,0 +1,107 @@
+//
```
davidme-stripe, line 1: We should remove this step from adding a new PM: I don't think it's necessary for user-facing docs, and it's easier to build them in IntegrationTester for our own testing purposes

davidme-stripe (dismissed):
Swoosh!

davidme-stripe (approved):

davidme-stripe (approved):

PR #2951, opened 2023-09-21
Disable card scanning on visionOS

davidme-stripe:
## Summary
visionOS exposes a null camera in the back and a spatial persona in the front. I don't think we'll get a valid PAN out of either.

## Motivation
Removing a button that won't work on visionOS.

## Testing
In simulator


davidme-stripe:
Never mind, `#if os(visionOS)` doesn't work in iPad compatibility mode, and there doesn't appear to be any other non-hacky way to detect the camera at this time.

PR #2952, opened 2023-09-21
Pass along billing address collection config to Apple Pay

davidme-stripe:
## Summary
We should observe the `billingDetailsCollectionConfiguration` when creating a PKPaymentRequest

## Motivation
Some users don't want to collect the postal code from Apple Pay.

## Testing
In PaymentSheet Example

## Changelog
Added


porter-stripe (approved):

sfriedman-stripe:
Should we add any tests for this?

PR #2956, opened 2023-09-25
Financial Connections: small refactor to simplify code logic

kgaidis-stripe:
## Summary

This is a simple refactor that moves the decision of whether we show an confirmation-alert when trying to close the auth flow to a specific function. There previous code was unnecessarily complex which made future changes difficult.

This is the confirmation alert:

Screenshot 2023-09-25 at 1 55 28 PM

### Closing Auth Flow By Using Close Button Works For Certain Screens

Here we see that the close alert appears when user enters deeper into the flow (by selecting a bank):

https://github.com/stripe/stripe-ios/assets/105514761/d5d14c31-feb9-457d-856c-a9e3b0d05ef1

### Closing Auth Flow From Success Pane Displays No Alert

Here we see that we display no alert for when we do not press the "X" button in the auth flow:

https://github.com/stripe/stripe-ios/assets/105514761/3f7bfab5-2788-46a4-b0a6-c8c9d5823937

### Run End To End Tests

Also ran end to end tests that sometimes test some cases with the alert:

`bitrise run financial-connections-stability-tests`

```
Test Suite FinancialConnectionsUITests.xctest started
FinancialConnectionsNetworkingUITests
    ✓ testNativeNetworkingTestMode (79.052 seconds)
FinancialConnectionsUITests
    ✓ testDataLiveModeOAuthNativeAuthFlow (30.685 seconds)
    ✓ testDataLiveModeOAuthWebAuthFlow (26.101 seconds)
    ✓ testDataTestModeOAuthNativeAuthFlow (22.878 seconds)
    ✓ testPaymentTestModeLegacyNativeAuthFlow (24.204 seconds)
    ✓ testPaymentTestModeManualEntryNativeAuthFlow (27.931 seconds)
    ✓ testSearchInLiveModeNativeAuthFlow (25.722 seconds)

	 Executed 7 tests, with 0 failures (0 unexpected) in 236.573 (236.579) seconds
```

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/Native/NativeFlowController.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/Native/NativeFlowController.swift
@@ -90,11 +90,21 @@ class NativeFlowController {
                 || navigationController.topViewController is NetworkingLinkVerificationViewController
                 || navigationController.topViewController is NetworkingSaveToLinkVerificationViewController
                 || navigationController.topViewController is LinkAccountPickerViewController)
-        closeAuthFlow(
-            showConfirmationAlert: showConfirmationAlert,
-            showNetworkingLanguageInConfirmationAlert: (dataManager.manifest.isNetworkingUserFlow == true && navigationController.topViewController is NetworkingLinkSignupViewController),
-            error: nil
-        )
+
+        let finishClosingAuthFlow = { [weak self] in
+            self?.closeAuthFlow()
+        }
+        if showConfirmationAlert {
+            CloseConfirmationAlertHandler.present(
+                businessName: dataManager.manifest.businessName,
+                showNetworkingLanguageInConfirmationAlert: (dataManager.manifest.isNetworkingUserFlow == true && navigationController.topViewController is NetworkingLinkSignupViewController),
+                didSelectOK: {
+                    finishClosingAuthFlow()
+                }
+            )
+        } else {
+            finishClosingAuthFlow()
+        }
     }
 
     @objc private func applicationWillEnterForeground() {
@@ -149,9 +159,9 @@ extension NativeFlowController {
         clearNavigationStack: Bool = false
     ) {
         if pane == .success && dataManager.manifest.skipSuccessPane == true {
-            closeAuthFlow(showConfirmationAlert: false, error: nil)
+            closeAuthFlow(error: nil)
         } else if pane == .manualEntry && dataManager.manifest.manualEntryMode == .custom {
-            closeAuthFlow(showConfirmationAlert: false, customManualEntry: true)
+            closeAuthFlow(customManualEntry: true)
         } else {
             let paneViewController = CreatePaneViewController(
                 pane: pane,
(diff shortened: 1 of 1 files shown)


StripeFinancialConnections/StripeFinancialConnections/Source/Native/NativeFlowController.swift
```diff
@@ -193,7 +203,7 @@ extension NativeFlowController {
 
     private func didSelectAnotherBank() {
         if dataManager.manifest.disableLinkMoreAccounts {
-            closeAuthFlow(showConfirmationAlert: false, error: nil)
+            closeAuthFlow(error: nil)
```
kgaidis-stripe, line 206: a key thing to notice (+ double check) in this refactor is that all call sites of `closeAuthFlow` have `showConfirmationAlert: false` EXCEPT one case...

this means that `CloseConfirmationAlertHandler` needs to only be used in that ONE case

StripeFinancialConnections/StripeFinancialConnections/Source/Native/NativeFlowController.swift
```diff
@@ -267,86 +275,71 @@ extension NativeFlowController {
...
-                    terminalError: customManualEntry ? "user_initiated_with_custom_manual_entry" : nil
-                )
-                .observe(on: .main) { [weak self] result in
-                    guard let self = self else { return }
-                    switch result {
-                    case .success(let session):
-                        let eventType = "object"
-                        if session.status == .cancelled
-                            && session.statusDetails?.cancelled?.reason == .customManualEntry
+        dataManager
```
kgaidis-stripe, line 278: the key thing to double check in this code is that `showConfirmationAlert = false` should now be what the function executes

StripeFinancialConnections/StripeFinancialConnections/Source/Native/NativeFlowController.swift
```diff
@@ -90,11 +90,21 @@ class NativeFlowController {
...
+
+        let finishClosingAuthFlow = { [weak self] in
+            self?.closeAuthFlow()
+        }
+        if showConfirmationAlert {
+            CloseConfirmationAlertHandler.present(
+                businessName: dataManager.manifest.businessName,
+                showNetworkingLanguageInConfirmationAlert: (dataManager.manifest.isNetworkingUserFlow == true && navigationController.topViewController is NetworkingLinkSignupViewController),
+                didSelectOK: {
+                    finishClosingAuthFlow()
```
ccen-stripe, line 102: my kotlin brain is telling me to inline this in 102 and 106, but not sure if this is a swift custom, so will let you decide :)

ccen-stripe (approved):

StripeFinancialConnections/StripeFinancialConnections/Source/Native/NativeFlowController.swift
kgaidis-stripe: Yea I'm probably 'overthinking' this a little bit but the basic fear/thought is that the intention is that whether we enter `showConfirmationAlert`, or not, the logic that is executed should be the same. If we do not put this in a shared closure/function, then there's a non-zero chance that it gets out of sync because its updated in one place, but not the other.

kgaidis-stripe: In a wild twist of faith: I actually ended up adding extra logic to `finishClosingAuthFlow` in a future-PR-I'm-making so it worked out!

Issue #2483, opened 2023-04-17
[BUG] STPPaymentCardTextFieldDelegate.paymentCardTextFieldDidChange() not called when deleting a character also causes field focus change


magouyaware:
## Summary

[STPPaymentCardTextFieldDelegate.paymentCardTextFieldDidChange](https://github.com/stripe/stripe-ios/blob/82887cccd886be71c6f7dde8ae6cd931ccbe5908/StripePaymentsUI/StripePaymentsUI/Source/UI%20Components/STPPaymentCardTextField.swift#L2228) does not fire when deleting a character also causes the `paymentCardTextFieldDigBeginEditing...` function to get called.

## Code to reproduce
Just hook up an `STPPaymentCardTextFieldDelegate` implementation to `STPPaymentCardTextField` and set breakpoints on all the methods.

## iOS version
All

## Installation method
Carthage

## SDK version
v23.2.0, but I looked at the release notes for newer versions and did not see anything mentioned that looked like it would affect this

## Other information
I have uploaded a recording of an app showing this behavior... The app displays a running log of output anytime one of the methods in STPPaymentCardTextFieldDelegate gets called. Whenever a character is deleted, a line beginning with "CardDetails Changed" is logged. Whenever fields gain and lose focus, appropriate lines are also logged. 

In the video, I enter a valid card number and the focus automatically moves to the expiration field. I then hit the back/delete key and the following happens:
1. Focus moves from the expiration field to the number field
2. Field focus lines are logged
3. The last number of the credit card is removed
4. _**There is no Card Details Changed log corresponding to that deleted character**_
5. I delete a second character and only then do I get a Card Details Changed log event

The video doesn't show it, but this happens for every field

https://user-images.githubusercontent.com/4357444/232624934-758cadb1-0a88-48e6-b4ad-c5ec1fb92a9a.mov

magouyaware:
I am not using SwiftUI. I'll see if I can cobble together a quick sample
app I can share with you. Our use case is a little different than most...
We have an SDK that is written on top of the Stripe SDK.

magouyaware:
Didn't want this to get stale... I've started working on a sample app I can share and hope to have that sometime this week.

magouyaware:
@eurias-stripe Ok, I've attached a little sample project, as well as a video clearly showing that any time the back button is pressed, the `STPPaymentCardTextFieldDelegate.paymentCardTextFieldDidChange()` callback is not called.

[...]

PR #2982, opened 2023-10-10
Fix card element not updating value when deleting on empty field

eurias-stripe:
## Summary
This is a followup to https://github.com/stripe/stripe-ios/pull/2971, it turns out that on top of not calling the delegate, we were not updating the backing view model, this was due to the implementation of `UITextField.deleteBackwards()`, which doesn't update the `text` property directly, causing `didSet` to not be called, which we depend on to update the backing view model. To get around this, I'm manually removing the last character in the text if it isn't empty.

## Motivation
https://github.com/stripe/stripe-ios/issues/2483

## Testing
- Verified visually
- Added test

## Changelog
* [Fixed] An issue with `STPPaymentCardTextField`, where the card params were not updated after deleting an empty sub field.

### DIFF
--- a/StripePaymentsUI/StripePaymentsUI/Source/Internal/UI/Views/STPFormTextField.swift
+++ b/StripePaymentsUI/StripePaymentsUI/Source/Internal/UI/Views/STPFormTextField.swift
@@ -197,17 +197,7 @@ import UIKit
     /// :nodoc:
     @objc
     public override func deleteBackward() {
-        // `UITextField.deleteBackwards` doesn't update the `text` property directly, and we depend on the `didSet`
-        // call on it to update our backing store.
-        // To get around this we manually remove the last character if the text is not empty, instead of forwarding
-        // the `deleteBackwards` call.
-        if var text = text, text.count > 0 {
-            text.removeLast()
-            self.text = text
-        } else {
-            super.deleteBackward()
-        }
-
+        super.deleteBackward()
         if (text?.count ?? 0) == 0 {
             if formDelegate?.responds(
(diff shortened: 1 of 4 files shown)


StripePaymentsUI/StripePaymentsUI/Source/Internal/UI/Views/STPFormTextField.swift
```diff
@@ -197,7 +197,17 @@ import UIKit
...
-        super.deleteBackward()
+        // `UITextField.deleteBackwards` doesn't update the `text` property directly, and we depend on the `didSet`
+        // call on it to update our backing store.
+        // To get around this we manually remove the last character if the text is not empty, instead of forwarding
+        // the `deleteBackwards` call.
+        if var text = text, text.count > 0 {
+            text.removeLast()
+            self.text = text
+        } else {
+            super.deleteBackward()
```
porter-stripe, line 208: Why isn't forwarding this an issue when it's not the last character? Seems like we should always manually remove the character.

eurias-stripe: This method is kinda weird, it only gets called when you delete on an empty field, but we also call it directly [here](https://github.com/stripe/stripe-ios/blob/d8e8b4bbdea990a4c1514672a4e3df4137457c03/StripePaymentsUI/StripePaymentsUI/Source/UI%20Components/STPPaymentCardTextField.swift#L1608), which is where the alternative behavior is triggered.

This actually makes me think that it might be better to have this alternate logic at the caller instead of here

PR #2984, opened 2023-10-11
Refactor and add tests for dashboard params

porter-stripe:
## Summary
- Refactors the Dashboard hack so it is only applied to deferred PaymentIntents
- Adds unit tests

## Motivation
https://jira.corp.stripe.com/browse/RUN_MOBILESDK-2650
https://jira.corp.stripe.com/browse/RUN_MOBILESDK-2653

## Testing
- New unit tests

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+DeferredAPI.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+DeferredAPI.swift
@@ -169,6 +169,7 @@ extension PaymentSheet {
         switch confirmType {
         case .saved:
             // The Dashboard app requires MOTO
+            intentParamsCopy.paymentMethodOptions = intentParamsCopy.paymentMethodOptions == nil ? .init() : intentParamsCopy.paymentMethodOptions
             intentParamsCopy.paymentMethodOptions?.setMoto()
         case .new(_, _, let paymentMethod, let shouldSave):
             // The Dashboard app cannot pass `paymentMethodParams` ie payment_method_data
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+DeferredAPI.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+DeferredAPI.swift
@@ -17,6 +17,8 @@ extension PaymentSheet {
         authenticationContext: STPAuthenticationContext,
         paymentHandler: STPPaymentHandler,
         isFlowController: Bool,
+        isDashboardApp: Bool,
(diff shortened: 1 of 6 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+DeferredAPI.swift
```diff
@@ -17,6 +17,8 @@ extension PaymentSheet {
         authenticationContext: STPAuthenticationContext,
         paymentHandler: STPPaymentHandler,
         isFlowController: Bool,
+        isDashboardApp: Bool,
+        dashboardParamsUpdater: @escaping DashboardParamsUpdater = setParamsForDashboardApp(confirmType:paymentIntentParams:paymentIntent:configuration:),
```
eurias-stripe, line 21: Will this ever be anything other than the default value? If we don't have a case where we need to pass in a different updater here, I would just remove it and call the default function directly

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+DeferredAPI.swift
```diff
@@ -17,6 +17,8 @@ extension PaymentSheet {
         authenticationContext: STPAuthenticationContext,
         paymentHandler: STPPaymentHandler,
         isFlowController: Bool,
+        isDashboardApp: Bool,
```
eurias-stripe, line 20: Do we need this param? Can't we just check the key here directly?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+DeferredAPI.swift
porter-stripe: > Will this ever be anything other than the default value?

Nope! The only reason I did this was for testing so we can test that the `dashboardParamsUpdater` is called under the correct circumstances. https://github.com/stripe/stripe-ios/pull/2984/files#diff-d596d94b7af6c551f33f81a8feceef006d2345d8af7718e26fff50f52e22cbbeR108

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+DeferredAPI.swift
eurias-stripe: I'm not sure how I feel about adding extra params (and thus extra complexity) just for the sake of testing, plus this makes the code more tightly bound to dashboard.
Also, this param could potentially be out of sync with the key (we could use a uk and set param to false, or viceversa).

eurias-stripe (approved):

eurias-stripe (approved):

PR #2996, opened 2023-10-17
Show correct card brand (saved & flow controller)

porter-stripe:
## Summary
- Updates the SavedPayment view to show the correct brand for co-branded cards
- Updates the `paymentOption` to return back the correct card brand image for co-branded cards for FlowController

## Motivation
https://docs.google.com/document/d/1T12LgVTkEv-LaDtQO3lGgNDt3nV_mwrUOJWgfUmTa9U/edit

## Testing
- New UI test
- Manual

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentOption+Images.swift
```diff
@@ -195,3 +195,9 @@ extension STPPaymentMethodType {
         return image.makeImage(overrideUserInterfaceStyle: forDarkBackground ? .dark : .light)
     }
 }
+
+extension String {
+    var toCardBrand: STPCardBrand? {
```
davidme-stripe, line 200: Tiny nit, I'd leave this as `STPCard.brand(from: self)` this convenience extension seems a little unusual for a String. I don't feel strongly about it though

davidme-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentOption+Images.swift
porter-stripe: It's mainly useful for the optional chaining so we can invoke it on a potentially nil string, `STPCard.brand(from: self)` takes in a non-nil `String` and returns a non-nil `STPCardBrand`, if we pass an empty string (in place of a nil string) it will give back `unknown`, when in these cases we'd want to fallback on the original card brand. To use `STPCard.brand(from: self)` directly we'd need to do some additional unwrapping/branching which isn't terrible but a bit cumbersome.

davidme-stripe: Makes sense!

PR #2997, opened 2023-10-18
Run export_builds in a VM

davidme-stripe:
## Summary
During the deploy script, we should build a Ventura VM and use it to run the `export_builds` step.

## Motivation
We need to continue building with Xcode 14.1 to support binary framework users using Xcode 14.

## Testing
Ran locally, stopped before actually deploying the artifact.


eurias-stripe (approved):
🎉

Looks sound to me

PR #3001, opened 2023-10-19
Fix double url encoding URL query params on iOS 17

vbc-stripe:
## Summary

Avoid double url encoding query params when running iOS 17. See https://developer.apple.com/documentation/foundation/url/3126806-init

## Motivation
#ir-after-deluge

## Testing
unit test, manual testing using Push Provisioning Test app

## Changelog

[Fixed] failure to push provision card using `STPAPIClient.retrievePushProvisioningDetails`

### DIFF
--- a/StripeCore/StripeCore/Source/Helpers/URLEncoder.swift
+++ b/StripeCore/StripeCore/Source/Helpers/URLEncoder.swift
@@ -39,6 +39,18 @@ import Foundation
     public class func queryString(from parameters: [String: Any]) -> String {
         return query(parameters)
     }
+
+    // For apps linked on or after iOS 17 and aligned OS versions, `URL` automatically percent- and IDNA-encodes invalid
+    // characters to help create a valid URL. See https://developer.apple.com/documentation/foundation/url/3126806-init
+    @objc(queryStringURLSafeIOS17FromParameters:)
+    public class func queryStringURLSafeIOS17(from params: [String: Any]) -> String {
+        let encodedQuery = URLEncoder.queryString(from: params)
+        return if #available(iOS 17.0, *) {
+            encodedQuery.removingPercentEncoding!
+        } else {
+            encodedQuery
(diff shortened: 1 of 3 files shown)


StripeCore/StripeCore/Source/Helpers/URLEncoder.swift
```diff
@@ -37,7 +37,20 @@ import Foundation
...
     public class func queryString(from parameters: [String: Any]) -> String {
-        return query(parameters)
+        return query(parameters, encoder: escape)
+    }
+
+    // For apps linked on or after iOS 17 and aligned OS versions, `URL` automatically percent- and IDNA-encodes invalid
+    // characters to help create a valid URL. See https://developer.apple.com/documentation/foundation/url/3126806-init
+    @objc(queryStringForURLFromParameters:)
+    public class func queryStringForURL(from params: [String: Any]) -> String {
+        let encoder = if #available(iOS 17.0, *) {
```
davidme-stripe, line 47: This only applies for apps *linked on* and running on iOS 17, so we might need to check the compiler version as well. But IMO we should rewrite this to skip the escaping and use URLComponents instead, that way we'll have consistent behavior without checking the version.

vbc-stripe: Ah ok, this was this `#if compiler(>=5.9)` part right? 

I think you're right about URLComponents so I'll abandon this hacky fix.

davidme-stripe: I can take on the URLComponents fix, but I'll reuse your test! Thanks for writing this!

vbc-stripe: @davidme-stripe awesome, glad there's something helpful for you here, especially after all your help with #ir-after-deluge

PR #3004, opened 2023-10-19
[Identity] Support softblock and continue anyway 

ccen-stripe:
## Summary

Support softblock/continue anyway when user fails to upload correct image more than once*.

Note the behavior is driven by server, please see details in [this](https://docs.google.com/document/d/1s6WwYpvnP_V6QRqjEbDnnVlzv8g5eUlROrSGh3XrK4U/edit#heading=h.nkvu6fer6ghi) internal doc. 

Client side change - When `VerificationPageDataRequirementError.continueButtonText` is populated by server, invoke the VerificationData endpoint with new parameter `force_confirm`

Android: https://github.com/stripe/stripe-android/pull/7430

# Motivation

Don't hardblock users when uploading incorrect images

## Testing

- [x] Added unit tests
- [x] Modified unit tests
- [x] Added screenshot tests
- [x] Manually verified

# Screenshots
| Before(hardblocked)  | After(allow use photo anyway after 2 tries) |
|   before   |   after   |


ccen-stripe:
@stripe/stripe-identity-observers

StripeIdentity/StripeIdentity/Source/NativeComponents/Coordinators/VerificationSheetFlowController.swift
```diff
@@ -588,7 +614,7 @@ extension VerificationSheetFlowController: VerificationSheetFlowControllerProtoc
                 documentType: documentType,
                 requireLiveCapture: staticContent.documentCapture.requireLiveCapture,
                 sheetController: sheetController,
-                documentUploader: documentUploader
+                documentUploader: documentUploader!
```
kgaidis-stripe, line 617: the force unwrap (`!`) seems like it should be safe but worth considering being cautious 

another way one could write this is:

```
        let documentUploader = DocumentUploader(
            imageUploader: IdentityImageUploader(
                configuration: .init(from: staticContent.documentCapture),
                apiClient: sheetController.apiClient,
                analyticsClient: sheetController.analyticsClient,
                idDocumentType: documentType
            )
        )
       self.documentUploader = documentUploader
```

then the local variable remains to be non-null

StripeIdentity/StripeIdentity/Source/NativeComponents/Coordinators/VerificationSheetFlowController.swift
```diff
@@ -597,7 +623,7 @@ extension VerificationSheetFlowController: VerificationSheetFlowControllerProtoc
                 documentType: documentType,
                 sheetController: sheetController,
                 cameraSession: makeDocumentCaptureCameraSession(),
-                documentUploader: documentUploader,
+                documentUploader: documentUploader!,
```
kgaidis-stripe, line 626: the force unwrap (`!`) seems like it should be safe but worth considering being cautious (force unwrap with null is a crash)

another way one could write this is:

```
        let documentUploader = DocumentUploader(
            imageUploader: IdentityImageUploader(
                configuration: .init(from: staticContent.documentCapture),
                apiClient: sheetController.apiClient,
                analyticsClient: sheetController.analyticsClient,
                idDocumentType: documentType
            )
        )
       self.documentUploader = documentUploader
```

then the local variable remains to be non-null

StripeIdentity/StripeIdentity/Source/NativeComponents/ViewControllers/DocumentCaptureViewController.swift
```diff
@@ -367,10 +375,12 @@ final class DocumentCaptureViewController: IdentityFlowViewController {
     private func saveFrontAndDecideBack(
         frontImage: UIImage
     ) {
+        isDecidingBack = true
         sheetController?.saveDocumentFrontAndDecideBack(
             from: analyticsScreenName,
             documentUploader: documentUploader,
             onCompletion: { [weak self] isBackRequired in
+                self?.isDecidingBack = false
```
kgaidis-stripe, line 383: I'm just flagging this just-in-case. I don't know if this is actually any issue.

Here we set `self?.isDecidingBack = false`, saying that BACK is not decided. 

There's code below that would make it sound like it will try to decide to scan back:

```
                if isBackRequired {
                    self?.imageScanningSession.startScanning(
                        expectedClassification: DocumentSide.back
                    )
                    self?.updateUI()
                } 
```

The worry is that this "back" code is wrong considering the `decidingBack = false` state. 

I might be completely wrong on calling it out, just flagging.

kgaidis-stripe (approved):

StripeIdentity/StripeIdentity/Source/NativeComponents/ViewControllers/DocumentCaptureViewController.swift
ccen-stripe: ah actually it doesn't mean "back is not decided", it means "back is not _being_ decided", this boolean would toggle the button to loading state and block any other button click while the logic is running. After back is decided we'll need to reenable the button

kgaidis-stripe (approved):

PR #3022, opened 2023-10-25
Improve snapshot test recording

davidme-stripe:
## Summary
* Added a new `snapshots.rb` script to run and record snapshot tests. 
* Moved "recordMode" to an environment variable instead of a hardcoded flag for each test. Added an "AllStripeFrameworks-RecordMode" scheme to record tests for all frameworks at once.
* Added the ability to run all snapshot tests to `test.rb`.
* Enforce a specific device and OS version when running snapshot tests, to reduce confusion.
* Rename a few snapshot tests that didn't have "Snapshot" in their names, so that we can remove some hacks from the test script.
* Re-recorded all snapshot tests on the current test device (iPhone 12 mini, 16.4)

* This breaks StripeLinkCore, which is an empty framework we don't plan to continue working on. Delete it.

## Motivation
Snapshot tests are annoying to maintain

## Testing
CI, running record script locally

## Changelog
None


eurias-stripe (approved):
Amazing work! 
Can't say I understand the ruby code all that well, but it looks sound.

PR #3026, opened 2023-10-27
Add default initializer for PaymentSheet.BillingDetails

davidme-stripe:
## Summary
Add a default initializer for `PaymentSheet.BillingDetails`.

## Motivation
LinkPaymentController accepts an optional billing details, but it wasn't possible to construct one.

## Testing
Use Billing Details in PaymentSheet Example

## Changelog
Added fixed entry

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
@@ -349,7 +349,12 @@ extension PaymentSheet {
         public var phone: String?
 
         /// Initializes billing details
-        public init() { }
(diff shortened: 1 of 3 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
```diff
@@ -347,6 +347,9 @@ extension PaymentSheet {
 
         /// The customer's phone number without formatting (e.g. 5551234567)
         public var phone: String?
+
+        /// Initializes billing details
+        public init() { }
```
yuki-stripe, line 352: I think here we should generate the default member-wise initializer, wdyt

davidme-stripe: Ooh, forgot that button existed, done

sfriedman-stripe (approved):

PR #3034, opened 2023-11-01
Add CBC edit card screen

porter-stripe:
## Summary
- Adds in a new screen to allow the editing of a co-branded card
- Following PRs will add ability to remove and update the card
simulator_screenshot_F6C09EB7-87A7-40B7-88E7-B4B77DA9E140

## Motivation
https://docs.google.com/document/d/1T12LgVTkEv-LaDtQO3lGgNDt3nV_mwrUOJWgfUmTa9U/edit#heading=h.jc7kakrp60yy

## Testing
- Manual
- Updated existing unit test
- Checked off design with jj (https://stripe.slack.com/archives/C05GTTC4PB5/p1698430632196099)

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdateCardViewController.swift
```diff
@@ -0,0 +1,184 @@
...
+// MARK: ElementDelegate
+extension UpdateCardViewController: ElementDelegate {
+    func continueToNextField(element: Element) {
+        // no-op
+    }
+
+    func didUpdate(element: Element) {
+        let selectedBrand = cardBrandDropDown.selectedItem.rawData
+        let currentCardBrand = STPCard.brand(from: paymentMethod.card?.networks?.preferred ?? "").rawValue
+        updateButton.update(state: selectedBrand != "\(currentCardBrand)" && selectedBrand != "-1" ? .enabled : .disabled)
```
davidme-stripe, line 182: Is this converting the rawValue of the preferred card brand to a String? And why `-1` here?

davidme-stripe (commented):
Mostly LGTM, just one note on how we're comparing Int-derived strings. I think it might be better if, for the logic checks, we converted the rawValue to an STPCardBrand then compared those. Maybe also for the cardBrandDropDown's rawValue to be the card brand ID instead of an int value.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdateCardViewController.swift
porter-stripe: Good feedback! I switch this from string comparison to STPCardBrand comparison.

davidme-stripe (approved):

PR #3038, opened 2023-11-02
Fixed localizedDescription for PaymentSheetError.

nschris-stripe:
## Summary
PaymentSheetError was overriding localizedDescription directly which doesn't work when the PaymentSheetError is upcasted to Error. This PR implements the LocalizedError protocol which will make localizedDescription accurate and implements a test to ensure that.

### Before
Screenshot 2023-11-02 at 10 40 03 AM

### After
Simulator Screenshot - iPhone 15 Pro - 2023-11-02 at 10 38 50

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetError.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetError.swift
@@ -14,7 +14,7 @@ import StripePayments
 ///
 /// Most errors do not originate from PaymentSheet itself; instead, they come from the Stripe API
 /// or other SDK components like STPPaymentHandler, PassKit (Apple Pay), etc.
-public enum PaymentSheetError: Error {
+public enum PaymentSheetError: Error, LocalizedError {
 
     /// An unknown error.
     case unknown(debugDescription: String)
@@ -53,8 +53,7 @@ public enum PaymentSheetError: Error {
     case failedToCreateLinkSession
     case linkNotAuthorized
 
-    /// Localized description of the error
-    public var localizedDescription: String {
+    public var errorDescription: String? {
         return NSError.stp_unexpectedErrorMessage()
     }
 }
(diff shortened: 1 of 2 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetError.swift
```diff
@@ -53,8 +53,7 @@ public enum PaymentSheetError: Error {
     case failedToCreateLinkSession
     case linkNotAuthorized
 
-    /// Localized description of the error
-    public var localizedDescription: String {
```
mludowise-stripe, line 57: 👋 To add context on why this change is backwards-compatible:

If the error type is preserved as a `PaymentSheetError` (e.g. `(error as? PaymentSheetError).localizedDescription`) returns the same value. 

But this fixes getting the localized desc when not casting the error type (e.g. `(error as Error).localizedDescription` will now return NSError.stp_unexpectedErrorMessage() whereas before it did not.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetErrorTest.swift
```diff
@@ -0,0 +1,16 @@
...
+//
+
+@testable@_spi(STP) import StripeCore
+@testable import StripePaymentSheet
+import XCTest
+
+final class PaymentSheetErrorTest: XCTestCase {
+    func testLocalizedDescription() throws {
+        XCTAssertEqual(PaymentSheetError.accountLinkFailure.localizedDescription, NSError.stp_unexpectedErrorMessage())
+    }
```
mludowise-stripe, line 15: I think this test should cast the error to a generic `Error` type. Before this change, `localizedDescription` worked if the error was treated as an `PaymentSheetError`. It only didn't work when cast to a generic Error type.

```suggestion
    func testLocalizedDescription() throws {
        XCTAssertEqual((PaymentSheetError.accountLinkFailure as Error).localizedDescription, NSError.stp_unexpectedErrorMessage())
    }
```

mludowise-stripe: It might also be helpful to add a comment clarifying why we're casting in the tests

mludowise-stripe (approved):

PR #3041, opened 2023-11-03
Hide Apple/Google Pay saved cards in CustomerSheet

yuki-stripe:
## Summary
This is the same work as https://github.com/stripe/stripe-ios/pull/3027 but for CustomerSheet

## Testing
See unit test

## Changelog
* [Fixed] CustomerSheet no longer displays saved cards that originated from Apple Pay or Google Pay.

### DIFF
--- a/Stripe/StripeiOSTests/CustomerAdapterTests.swift
+++ b/Stripe/StripeiOSTests/CustomerAdapterTests.swift
@@ -42,7 +42,6 @@ class CustomerAdapterTests: APIStubbedTestCase {
         key: CustomerEphemeralKey,
         paymentMethodType: String,
         paymentMethodJSONs: [[AnyHashable: Any]],
-        expectedCount: Int,
         apiClient: STPAPIClient
     ) {
         stub { urlRequest in
--- a/Stripe/StripeiOSTests/CustomerAdapterTests.swift
+++ b/Stripe/StripeiOSTests/CustomerAdapterTests.swift
@@ -7,9 +7,11 @@ import Foundation
 import OHHTTPStubs
 import OHHTTPStubsSwift
 @_spi(STP) @testable import StripeCore
-import StripeCoreTestUtils
+@_spi(STP) @testable import StripeCore
+@testable import StripeCoreTestUtils
 @_spi(STP) @testable import StripePayments
 @_spi(STP) @testable import StripePaymentSheet
+@_spi(STP) @testable import StripePaymentsTestUtils
 import XCTest
 
 enum MockEphemeralKeyEndpoint {
@@ -43,8 +45,6 @@ class CustomerAdapterTests: APIStubbedTestCase {
         expectedCount: Int,
         apiClient: STPAPIClient
     ) {
-        let exp = expectation(description: "listPaymentMethod")
-        exp.expectedFulfillmentCount = expectedCount
         stub { urlRequest in
             if urlRequest.url?.absoluteString.contains("/payment_methods") ?? false
                 && urlRequest.url?.absoluteString.contains("type=\(paymentMethodType)") ?? false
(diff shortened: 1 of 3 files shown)


Stripe/StripeiOSTests/CustomerAdapterTests.swift
```diff
@@ -123,7 +119,6 @@ class CustomerAdapterTests: APIStubbedTestCase {
 
         XCTAssertEqual(pms.count, 1)
         XCTAssertEqual(pms[0].stripeId, expectedPaymentMethods[0].stripeId)
-        await waitForExpectations(timeout: 2)
```
yuki-stripe, line 126: These were giving warnings:
```
Instance method 'waitForExpectations' is unavailable from asynchronous contexts; Use await fulfillment(of:timeout:enforceOrder:) instead; this is an error in Swift 6
```

and didn't seem necessary since we already wait for `fetchPaymentMethods` to return, so I removed it.

Stripe/StripeiOSTests/CustomerAdapterTests.swift
```diff
@@ -43,8 +45,6 @@ class CustomerAdapterTests: APIStubbedTestCase {
         expectedCount: Int,
         apiClient: STPAPIClient
     ) {
-        let exp = expectation(description: "listPaymentMethod")
-        exp.expectedFulfillmentCount = expectedCount
```
wooj-stripe, line 47: What was the reason for removing expected count?  If we decide to remove it, then we should remove it from the function parameter.

wooj-stripe: I can't remember why we needed this, but i think you're right in that it isn't needed. Lets remove the `expectedCount` parameter and then happy to ship it.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerAdapter/CustomerAdapter.swift
```diff
@@ -143,11 +143,16 @@ open class StripeCustomerAdapter: CustomerAdapter {
...
                 types: savedPaymentMethodTypes
             ) { paymentMethods, error in
-                guard let paymentMethods = paymentMethods, error == nil else {
+                guard var paymentMethods, error == nil else {
                     let error = error ?? PaymentSheetError.unexpectedResponseFromStripeAPI // TODO: make a better default error
                     continuation.resume(throwing: error)
                     return
                 }
+                // Remove cards that originated from Apple or Google Pay
+                paymentMethods = paymentMethods.filter { paymentMethod in
```
jaynewstrom-stripe, line 152: On Android we pass these into a higher level object/method which actually makes the API call. Should we do the filtering there?

Stripe/StripeiOSTests/CustomerAdapterTests.swift
yuki-stripe: Oops, removed.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerAdapter/CustomerAdapter.swift
yuki-stripe: Talked about it offline - we don't have that higher level object yet, but feels reasonable to do that!  I won't do that in this PR though.

jaynewstrom-stripe: Chatted offline, doesn't make sense for iOS to do right now since they don't have the same abstractions as Android.

wooj-stripe (approved):

PR #3046, opened 2023-11-06
Fix visual bug in PaymentSheet after dismissing sheet

wooj-stripe:
## Summary
1. Present payment sheet (complete)
2. Select a different PM other than Card
3. Tap to dismiss the sheet
4. Represent the sheet

Observed behavior:
Upon re-presenting the sheet, you see the last PM selected w/ the info you filled in, and then it quickly changes back to card.

Expected behavior:
The behavior is due to the sheet being reloaded.  Therefore, we should be showing a loading indicator upon re-presenting the sheet.

## Motivation
Internal bug filed.

## Testing
Manual testing.


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
```diff
@@ -125,6 +125,7 @@ public class PaymentSheet {
             } else {
                 completion(status)
             }
+            self.bottomSheetViewController.contentStack = [self.loadingViewController]
```
yuki-stripe, line 128: Feel free to ship as-is, just a suggestion. I think it'd be more obvious that we're starting every `present` call with a loading screen if we did this immediately before presenting on line 190.

yuki-stripe (dismissed):

CHANGELOG.md
```diff
@@ -1,4 +1,7 @@
 ## X.Y.Z 2023-y-z
+### PaymentSheet
+* [Fixed] Visual bug when re-presenting payment sheet
```
```suggestion
* [Fixed] Visual bug where re-presenting PaymentSheet wouldn't show a spinner while it reloads.
```

feel free to modify as you see fit

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
wooj-stripe: Unfortunately, this makes the sheet not present at all.

yuki-stripe: ! That's surprising, why?

wooj-stripe: I suspect it has something to do with BottomSheetViewController because the viewcontroller is already set to loadingViewController.  I did not investigate it too deeply though.

wooj-stripe: Dug into this a little bit more, and seems that my initial theory around it being the same view is correct.

Option 1:
Ship as-is

Option 2:
In BottomSheetViewController add the following to the contentViewController's didSet:
```
    private var contentViewController: BottomSheetContentViewController {
        didSet(oldContentViewController) {
+            guard self.contentViewController !== oldContentViewController else {
+                return
+            }
``` 
Then as you were suggesting on line 190:
```
        self.bottomSheetViewController.contentStack = [self.loadingViewController]
```

I am kind of leaning towards option 2, it could lead to future footguns in the future.  Thoughts?

yuki-stripe: Ahh that makes perfect sense! Agreed, option 2 is better. Thanks for investigating and making the deeper fix!

yuki-stripe (approved):

PR #3057, opened 2023-11-08
Improve error handling for payment sheet w/out returnURL

wooj-stripe:
## Summary
Turn on availability & update error message

## Motivation
Internal request to improve dev exp

## Testing
manually tested


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetError.swift
```diff
@@ -163,7 +163,7 @@ extension PaymentSheetError: CustomDebugStringConvertible {
             case .linkSignUpNotRequired:
                 return "Don't call sign up if not needed"
             case .noPaymentMethodTypesAvailable(intentPaymentMethods: let intentPaymentMethods):
-                return "None of the payment methods on the PaymentIntent/SetupIntent can be used in PaymentSheet: \(intentPaymentMethods). You may need to set `allowsDelayedPaymentMethods` or `allowsPaymentMethodsRequiringShippingAddress` in your PaymentSheet.Configuration object."
+                return "None of the payment methods on the PaymentIntent/SetupIntent can be used in PaymentSheet: \(intentPaymentMethods). You may need to set `allowsDelayedPaymentMethods` or `allowsPaymentMethodsRequiringShippingAddress` or set `returnURL` in your PaymentSheet.Configuration object."
```
yuki-stripe, line 166: It'd be even better if we could say the specific reason to this error message but this is a perfectly fine incremental improvement.

wooj-stripe: +1 I think the path of the code that is calling this makes it difficult to know the exact reason. Looking for an incremental improvement at this point.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift
```diff
@@ -188,7 +188,7 @@ final class PaymentSheetLoader {
             }
         }
         // Ensure that there's at least 1 payment method type available for the intent and configuration.
-        let paymentMethodTypes = PaymentSheet.PaymentMethodType.filteredPaymentMethodTypes(from: intent, configuration: configuration)
+        let paymentMethodTypes = PaymentSheet.PaymentMethodType.filteredPaymentMethodTypes(from: intent, configuration: configuration, logAvailability: true)
```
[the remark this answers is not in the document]
wooj-stripe: Great callout.  It was being printed out twice for the complete flow.

Taking a step back:
`filteredPaymentMethodTypes` is being called from:
* `AddPaymentMethodViewController`
* `PaymentSheetLoader`

I've changed `AddPaymentMethodViewController` to `logAvailability==false`, and continuing to have `PaymentSheetLoader`  with `logAvailability==true`. This should allow us to only see the message once when paymentsheet is loaded for both complete and flowcontroller use cases.

PR #3059, opened 2023-11-10
Add SEPA debit support to CustomerSheet

wooj-stripe:
## Summary
Adding support for SEPA Debit

## Motivation
Parity of `sepa_debit` of a saved payment method in payment sheet and customer sheet

## Testing
- Added snapshot, ui , and unit tests

## Changelog
Added in to change log


davidme-stripe (approved):

AlfredReisner:
Hi, we want to integrate the Customer Sheet into our apps, we are using PaymentSheet already.
Could it be that there is a different setup for both sheets? We are using the same clientSecret, customerId and ephemeralKey to setup both sheets, but we are not able to disable SEPAdebit for the CustomerSheet. It feels like hardcoded and not respecting the corresponding configuration of the provided intents.

wooj-stripe:
Hi @AlfredReisner ,

Thanks for reaching out.  Just want to confirm the use cases you are requesting

Use Case 1: The ability to disable showing the SEPA payment method when **adding** a saved payment method in CustomerSheet.

Use Case 2: The ability to disable showing **already saved** SEPA payment methods in CustomerSheet.

Questions:
1. For use case 1, this is not currently supported in CustomerSheet. To help us better prioritize, is it possible to share the reasoning behind not wanting to show SEPA, despite SEPA being activated for the account?

2. For this case 2, you could override CustomerAdapter's `fetchPaymentMethods` to hide saved SEPA based payment methods, but will lead to a poor user experience if customers can add SEPA based payment methods.

AlfredReisner:
Hey @wooj-stripe, sorry for the delay. 

Yes we want to disable the ability of SEPA payments for the CustomerSheet only. Our system supports custom payments and invoices so we can't easily deactivate SEPA globally. We would consider that the capabilities of the client_secret distinguish about the available payment methods in the CustomerSheet. We configure this client_secret to only be capable of card payments and setup the CustomerSheet with this clientSecret. Could this be an option you could consider?

wooj-stripe:
Hey @AlfredReisner -
Thanks for explaining your use case as it provides context into how we shape our APIs.

I understand your desire to respect the payment methods that are associated with the SetupIntent's `client_secret`. We will take this into consideration as we look to add configurability with what payment methods are available to add in CustomerSheet.

That being said, if you create a SetupIntent and specify "card" as the only payment method, then attempting to confirm with any other payment method should fail on confirmation. I recognize that showing a Payment method type (e.g. sepa_debit) and failing on confirmation is not ideal, so as stated before, we are looking to add some configurability in this area.  

Unfortunately, I do not have a timeline available at the moment, I'll update on this thread with any progress made in this area.

AlfredReisner:
Alright, thanks for the explanation. 👍

wooj-stripe:
Hey @AlfredReisner

Thanks for your patience. We added [paymentMethodTypes](https://github.com/stripe/stripe-ios/blob/945205413ff7eeb4d13a0a3a162f603c3679165c/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerAdapter/CustomerAdapter.swift#L78) in CustomerAdapter which allows users to control what payment methods are displayed in CustomerSheet.  Therefore, if you only want to allow `card` as a payment method:
```
let customerAdapter = StripeCustomerAdapter(customerEphemeralKeyProvider: {
                        ...
            }, setupIntentClientSecretProvider: {
                        ...
+           }, paymentMethodTypes: ["card"])
```            
https://github.com/stripe/stripe-ios/releases/tag/23.25.0

Hope this helps!

PR #3062, opened 2023-11-13
Make some initial changes to prepare for future generic EPM support

yuki-stripe:
## Summary
Make some initial changes to prepare for future generic EPM support. We still only support "external_paypal", but we represent the type in a generic way.

- Changes PaymentMethodType.externalPayPal to .external(ExternalPaymentMethod)
- Changes PaymentOption.externalPayPal(ConfirmParams) to .external(ExternalPaymentMethod, STPPaymentMethodBillingDetails)

In the next PRs, I'll:
- Make the SDK send the merchant's desired epms to the v1/elements/sessions endpoint, and have its response determine which EPMs we show.
- Add image URLs to `ExternalPaymentMethod` and fetch the icon from them instead of using the hardcoded paypal asset 
- Add analytics

## Testing
Relying on existing tests, since no new behavior was added.

## Changelog
Not user facing


porter-stripe (approved):
💯

PR #3069, opened 2023-11-16
Avoid clashing, allow new returning customer in PaymentSheetTestPlayground

wooj-stripe:
## Summary
To avoid us all using the same returning customer, I propose
A new customer results in creating a new customer and save/

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
@@ -66,6 +66,7 @@ struct PaymentSheetTestPlayground: View {
                         SettingPickerView(setting: $playgroundController.settings.integrationType)
                         SettingView(setting: customerModeBinding)
                         TextField("CustomerId", text: customerIdBinding)
+                            .disabled(true)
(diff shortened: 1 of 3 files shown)


Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
```diff
@@ -113,7 +114,28 @@ struct PaymentSheetTestPlayground: View {
...
+            switch playgroundController.settings.customerMode {
+            case .guest:
+                return ""
+            case .new:
+                return playgroundController.settings.customerId ?? ""
+            case .returning:
+                return playgroundController.settings.customerId ?? ""
+            }
+        } set: { newString in
+            playgroundController.settings.customerId = (newString != "") ? newString : nil
```
porter-stripe, line 137: When we set the customerId manually, should we update the customerMode to something, like new? What if I enter a customerId and I'm in guest mode.

wooj-stripe: There are some issues with manually typing in the customer id.  For now, I've disabled it until I can get back to this.

porter-stripe (approved):

PR #3076, opened 2023-11-20
Don't enable the "save this card" checkbox by default

davidme-stripe:
## Summary
Change the "save this card" checkbox to be unchecked by default.

## Motivation
Latest guidance.

## Testing
CI, updated existing tests.

## Changelog
Added a changelog entry.


Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
```diff
@@ -2647,6 +2638,10 @@ class PaymentSheetDeferredServerSideUITests: PaymentSheetUITestCase {
 
         try! fillCardData(app)
 
+        var saveThisCardToggle = app.switches["Save this card for future Example, Inc. payments"]
+        XCTAssertFalse(saveThisCardToggle.isSelected)
+        saveThisCardToggle.tap()
+
```
wooj-stripe, line 2644: Nit: seems like we should add  (for multiple of these)
```
        saveThisCardToggle.tap()
+       XCTAssertTrue(saveThisCardToggle.isSelected)
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
```diff
@@ -61,8 +61,7 @@ extension PaymentSheet {
         var isSelectedByDefault: Bool {
             switch self {
             case .automatic:
-                // only enable the save checkbox by default for US
-                return Locale.current.stp_regionCode == "US"
+                return false
```
wooj-stripe, line 64: We should probably add this to our list of "things to break" in the next major version

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
davidme-stripe: I think the test enforces this (it checks if the card is saved!) but I can

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
davidme-stripe: I like having .automatic around just in case, maybe someday we can control it server-side

wooj-stripe (approved):

PR #3079, opened 2023-11-20
Add the ability to customize the success color of the primary button

porter-stripe:
## Summary
- Add the ability to customize the success color of the primary button with the appearance API
- Adds a new button to the playground to test this
- [API Review](https://docs.google.com/document/d/1NbuMpp-NdGqRT_4j9jHYXsnbEaZo_WeaatJEzSYiUyA/edit#heading=h.570u4leviueu)

Demo (setting success color to red)

https://github.com/stripe/stripe-ios/assets/88012362/d786a1e8-d5fe-4fea-85cc-ea8dcc5867b8

## Motivation
https://jira.corp.stripe.com/browse/RUN_MOBILESDK-2737

## Testing
- Manual
- New snapshot test

## Changelog
See diff


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/ConfirmButton.swift
```diff
@@ -63,9 +63,6 @@ class ConfirmButton: UIView {
         get {
             return buyButton.succeededBackgroundColor
         }
-        set {
-            buyButton.succeededBackgroundColor = newValue
-        }
```
porter-stripe, line 68: Dead code, was never hit.

Example/PaymentSheet Example/PaymentSheet Example/AppearancePlaygroundView.swift
```diff
@@ -127,6 +127,16 @@ struct AppearancePlaygroundView: View {
             set: { self.appearance.primaryButton.backgroundColor = UIColor($0) }
         )
 
+        let primaryButtonSuccessColorBinding = Binding(
+            get: { Color(self.appearance.primaryButton.successBackgroundColor) },
+            set: { self.appearance.primaryButton.successBackgroundColor = UIColor($0) }
+        )
+
+        let primaryButtonSuccessTextColorBinding = Binding(
+            get: { Color(self.appearance.primaryButton.successTextColor ?? self.appearance.primaryButton.textColor ?? .white) },
```
yuki-stripe, line 136: Just a non-blocking comment - IIUC seems like it'd be more accurate to refactor our code/UI to allow for nil values here instead of assuming what the default color is.  If we ever change the defaulting logic this will be wrong.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetAppearance.swift
```diff
@@ -167,6 +167,15 @@ public extension PaymentSheet {
             /// - Note: If `nil`, defaults to either white or black depending on the color of the button
             public var textColor: UIColor?
 
+            /// The background color of the primary button when in a success state.
+            /// - Note: Only applies to PaymentSheet. The primary button transitions to the success state when payment succeeds.
+            public var successBackgroundColor: UIColor = .systemGreen
+
+            /// The text color of the primary button when in a success state.
+            /// - Note: Only applies to PaymentSheet. The primary button transitions to the success state when payment succeeds.
+            /// - Note: If `nil`, defaults to  `textColor`
```
yuki-stripe, line 176:
```suggestion
            /// - Note: If `nil`, defaults to `textColor`
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/ConfirmButton.swift
yuki-stripe: Can we delete the entire `var succeededBackgroundColor: UIColor`?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/ConfirmButton.swift
```diff
@@ -536,9 +533,9 @@ class ConfirmButton: UIView {
         private func foregroundColor(for status: Status) -> UIColor {
             let background = backgroundColor(for: status)
 
-            if status == .succeeded {
-                // always use hardcoded color for foreground color when in success state
-                return .white
+            // Use successTextColor if in succeded state and provided, otherwise fallback on foreground color
```
yuki-stripe, line 536:
```suggestion
            // Use successTextColor if in succeeded state and provided, otherwise fallback to foreground color
```

yuki-stripe (dismissed):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/ConfirmButton.swift
porter-stripe: Hmm yep can remove that too.

Example/PaymentSheet Example/PaymentSheet Example/AppearancePlaygroundView.swift
porter-stripe: Agree that would be ideal, always resolving to a non-nil color is mainly driven by the fact that `ColorPicker` requires a non-nil color. Maybe a button with text `nil` that when tapped presents the color picker? I looked around at this a bit but programmatically presenting a `ColorPicker` from a different button seems non-trivial: https://github.com/franklynw/PresentableColorPicker

porter-stripe: I guess there is another way, slightly hacky: https://stackoverflow.com/questions/73382095/activate-color-picker-overlay-when-button-pressed-swiftui

yuki-stripe (approved):

PR #3086, opened 2023-11-22
Fix build with tuist removal

ccen-stripe:
## Summary

Manually update the `.xcodeproj` with added/removed files in the release branch


ccen-stripe:
@stripe/stripe-identity-observers

PR #3089, opened 2023-11-27
Fix border when toggling light/dark mode

wooj-stripe:
## Summary
Fix border when toggling light/dark mode

## Motivation
Bug found during dogfooding

## Testing
Manually tested

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/CVCReconfirmationViewController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/CVCReconfirmationViewController.swift
@@ -94,6 +94,10 @@ class CVCReconfirmationViewController: UIViewController {
         ])
         updateUI()
     }
+    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        super.traitCollectionDidChange(previousTraitCollection)
+        self.cvcFormElement.traitCollectionDidChange(previousTraitCollection)
+    }
     func didFinishPresenting() {
         self.cvcFormElement.didFinishPresenting()
     }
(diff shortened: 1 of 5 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/CVCReconfirmationViewController.swift
```diff
@@ -94,6 +94,14 @@ class CVCReconfirmationViewController: UIViewController {
         ])
         updateUI()
     }
+
+    #if !STP_BUILD_FOR_VISION
+    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        super.traitCollectionDidChange(previousTraitCollection)
+        self.cvcFormElement.traitCollectionDidChange(previousTraitCollection)
```
yuki-stripe, line 101: IMO instead of manually doing this, we should make a CVCRecollectionView that subclasses UIView so it can just override `traitCollectionDidChange` like other element views.

wooj-stripe: Chatted offline -- thanks for the help! great suggestion!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/CVCReconfirmationView.swift
```diff
@@ -0,0 +1,77 @@
...
+//
+
+import Foundation
+@_spi(STP) import StripeCore
+@_spi(STP) import StripePayments
+@_spi(STP) import StripePaymentsUI
+@_spi(STP) import StripeUICore
+import UIKit
+
+class CVCReconfirmationView: UIView {
```
yuki-stripe, line 13: supernit you don't need to fix, but did you mean to call this `CVCRecollectionView` since the element is `CVCRecollectionElement`?

yuki-stripe (dismissed):

yuki-stripe (dismissed):

yuki-stripe (approved):

PR #3092, opened 2023-11-28
Updating cvc recollection interface

wooj-stripe:
## Summary
Updating interface for cvc recolleciotn

## Motivation
Dogfooding feedback

## Testing
None

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetIntentConfiguration.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetIntentConfiguration.swift
@@ -34,7 +34,7 @@ public extension PaymentSheet {
             _ intentCreationCallback: @escaping ((Result<String, Error>) -> Void)
         ) -> Void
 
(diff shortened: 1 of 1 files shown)


yuki-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetIntentConfiguration.swift
```diff
@@ -34,7 +34,7 @@ public extension PaymentSheet {
             _ intentCreationCallback: @escaping ((Result<String, Error>) -> Void)
         ) -> Void
 
-        public typealias CVCRecollectionHandler = () -> Bool
+        public typealias CVCRecollectionEnabledCallback = () -> Bool
```
yuki-stripe, line 37: Sorry @wooj I approved, but realized - can we put this under `_spi` too?  And also add a docstring

wooj-stripe: oops, yes. this needs to be under _spi!

PR #3101, opened 2023-11-30
Update maskView bg color of cvc pm view

wooj-stripe:
## Summary
Add mask view to show field is disabled

## Motivation
Internal dogfooding feedback

## Testing
Manually tested and verified with design


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/PaymentMethodInformationView.swift
```diff
@@ -32,22 +32,22 @@ class PaymentMethodInformationView: UIView {
...
     }()
 
+    lazy var transparentMaskView: UIView = {
+        let view = UIView()
+        view.backgroundColor = transparentMaskViewBackgroundColor()
+        view.layer.cornerRadius = appearance.cornerRadius
+        view.layer.maskedCorners = [.layerMinXMinYCorner, .layerMinXMaxYCorner]
+        return view
+    }()
+
```
yuki-stripe, line 42: Some nits unrelated to this line, feel free to take or drop:
- The name PaymentMethodInformationView made me think this was a much more general-purpose view, but since it's actually specific to the CVCRecollectionView I might rename it `CVCPaymentMethodInformationView` or something.
- `@objc(STP_Internal_PaymentMethodInformation)` is also unnecessary afaik

yuki-stripe (approved):

yuki-stripe (approved):

PR #3103, opened 2023-12-01
Set newly saved payment methods as default

yuki-stripe:
## Summary
PaymentSheet now sets newly saved PMs as the default if they can be displayed as a saved PM (if they are cards, us bank account, or SEPA Debit).

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-1035

## Testing
Added tests for both deferred and non-deferred confirm flows using:
- PI (shouldn't set as default)
- PI w/ checkbox set (should set as default)
- PI w/ SFU (should set as default)
- SI (should set as default)

Modified an end-to-end test that saves a card to check that it is set as the default instead of Apple Pay. This test failed before this PR and passes now.

Manually tested server-side confirmation flow, since that's not in the unit tests.

## Changelog
* [Fixed] PaymentSheet sets newly saved payment methods as the default.


Stripe/StripeiOSTests/STPSetupIntentTest.swift
```diff
@@ -68,7 +68,7 @@ class STPSetupIntentTest: XCTestCase {
 
         XCTAssertEqual(url, URL(string: "https://hooks.stripe.com/redirect/authenticate/src_1Cl1AeIl4IdHmuTb1L7x083A?client_secret=src_client_secret_DBNwUe9qHteqJ8qQBwNWiigk"))
         XCTAssertEqual(setupIntent.paymentMethodID, "pm_123456")
-        XCTAssertEqual(setupIntent.status, STPSetupIntentStatus.requiresAction)
+        XCTAssertEqual(setupIntent.status, STPSetupIntentStatus.requiresConfirmation)
```
wooj-stripe, line 71: Probably needs to be reverted

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
```diff
@@ -429,6 +436,52 @@ extension PaymentSheet {
...
+            case .paymentIntent(let paymentIntent):
+                return paymentIntent.paymentMethod
+            case .setupIntent(let setupIntent):
+                return setupIntent.paymentMethod
+            }
+        }
+    }
+
+    /// A helper method that sets the Customer's default payment method if necessary
+    /// - Parameter isSaving: Whether or not we are saving the given `paymentMethod`. If it's not being saved, it can't be reused by the customer, so don't set it as the default.
```
wooj-stripe, line 468: need to update this comment.  `isSaving` doesn't seem to be a parameter.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+DeferredAPI.swift
```diff
@@ -48,6 +48,13 @@ extension PaymentSheet {
                 switch intentConfig.mode {
                 case .payment:
                     let paymentIntent = try await configuration.apiClient.retrievePaymentIntent(clientSecret: clientSecret, expand: ["payment_method"])
+                    let completion = { (status: STPPaymentHandlerActionStatus, paymentIntent: STPPaymentIntent?, error: NSError?, deferredIntentConfirmationType: STPAnalyticsClient.DeferredIntentConfirmationType) in
```
wooj-stripe, line 51: i think we should rename this `completion` to avoid masking the existing `completion` pass in as an argument to the function.

wooj-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+DeferredAPI.swift
wooj-stripe: I think it be clearer to avoid overriding the completion variable and calling it wrappedCompletion, but I'm not going to block on this because the two completion handlers have different arguments.

PR #3115, opened 2023-12-07
Fix selected pm in ps & cs playgrounds

wooj-stripe:
## Summary
Update text color of selected payment method when toggling between light/dark

## Motivation
Feedback during dogfooding

## Testing
Manually tested

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/CustomerSheetTestPlayground.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/CustomerSheetTestPlayground.swift
@@ -174,6 +174,7 @@ struct CustomerSheetButtons: View {
 
 struct CustomerSheetPaymentOptionView: View {
     let paymentOptionDisplayData: CustomerSheet.PaymentOptionSelection?
+    @Environment(\.colorScheme) var colorScheme
 
     var body: some View {
         HStack {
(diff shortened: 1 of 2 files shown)


Example/PaymentSheet Example/PaymentSheet Example/CustomerSheetTestPlayground.swift
```diff
@@ -184,6 +185,8 @@ struct CustomerSheetPaymentOptionView: View {
                 .foregroundColor(.black)
             Text(paymentOptionDisplayData?.displayData().label ?? "None")
                 .accessibility(identifier: "Payment method")
+                .foregroundColor(colorScheme == .dark ? .white : .black)
+
         }.padding()
             .foregroundColor(.black)
```
davidme-stripe, line 191: Could you set this line to `.foregroundColor(.primary)` instead to reflect the system default? Checking the color scheme manually shouldn't be necessary

Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
```diff
@@ -320,6 +321,7 @@ struct PaymentOptionView: View {
                 // Surprisingly, setting the accessibility identifier on the HStack causes the identifier to be
                 // "Payment method-Payment method". We'll set it on a single View instead.
                 .accessibility(identifier: "Payment method")
+                .foregroundColor(colorScheme == .dark ? .white : .black)
         }
         .padding()
         .foregroundColor(.black)
```
davidme-stripe, line 327: Same here

yuki-stripe (approved):

PR #3116, opened 2023-12-07
Fix test playground when switching countries

wooj-stripe:
## Summary
I found that switching countires in test playground results in the backend not sending a response, in which case the client just hangs. 

Proposed changes:
1. Instead of waiting for a response to come back, i suggest that we just cancel the existing job, and kick off a new one.
2. Default to a "guest" when changing countries as to avoid crashing the backend
3. re-order autoload below cvc recollection
4.  nil out the lastPaymentResult if we've successfully loaded

## Motivation
Spinner occurs for a long period of time when switching countries

## Testing
manually tested/emulator

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
@@ -237,6 +237,7 @@ class PlaygroundController: ObservableObject {
     var checkoutEndpoint: String = PaymentSheetTestPlaygroundSettings.defaultCheckoutEndpoint
     var addressViewController: AddressViewController?
     var appearance = PaymentSheet.Appearance.default
+    var currentDataTask: URLSessionDataTask?
 
     func makeAlertController() -> UIAlertController {
         let alertController = UIAlertController(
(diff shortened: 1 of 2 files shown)


Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
```diff
@@ -419,9 +427,12 @@ extension PlaygroundController {
             }
 
             DispatchQueue.main.async {
+                self.lastPaymentResult = nil
                 self.clientSecret = json["intentClientSecret"]
                 self.ephemeralKey = json["customerEphemeralKeySecret"]
-                self.settings.customerId = json["customerId"]
+                if self.settings.customerId != json["customerId"] {
```
davidme-stripe, line 433: Why is this needed? Is there a side effect to setting this again?

davidme-stripe (approved):

Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
davidme-stripe: Ah! I see now, thanks!

PR #3127, opened 2023-12-12
Support visionOS across Xcode 14 and 15

davidme-stripe:
## Summary
We need to support visionOS across Xcode 14 and 15 and all package managers. To do this, we need a solution that doesn't involve compile-time flags, as SPM won't let us provide them.

See https://github.com/stripe/stripe-ios/pull/3127/files#diff-ae2869cf29a9ac2a97e2b2f976d80f6ec5b7e59a004b6bf6bee0aa2e3f281ba5R148 for a full explanation of the hack.

## Motivation
Fix visionOS support in various package managers.

## Testing
CI and manually in Xcode 15.3. Re-adding automated visionOS tests in https://github.com/stripe/stripe-ios/pull/3126.

## Changelog
Added.

### DIFF
--- a/StripeApplePay/StripeApplePay/Source/ApplePayContext/STPApplePayContext.swift
+++ b/StripeApplePay/StripeApplePay/Source/ApplePayContext/STPApplePayContext.swift
@@ -145,7 +145,17 @@ public class STPApplePayContext: NSObject, PKPaymentAuthorizationControllerDeleg
     @available(macCatalystApplicationExtension, unavailable)
     @objc(presentApplePayWithCompletion:)
     public func presentApplePay(completion: STPVoidBlock? = nil) {
-        #if STP_BUILD_FOR_VISION
+        // A note on `canImport(CompositorServices)`, found here and elsewhere
+        // in the codebase: This is a terrible, terrible hack.
+        // We should use #if os(visionOS), but that will fail to compile
+        // on Xcode 14. Because we need to continue supporting both Xcode 14
+        // *and* building on visionOS, we instead check `canImport(CompositerServices)`, as CompositerServices currently
+        // only exists on visionOS. I'm sure it will exist on iOS someday,
+        // but we only need this in place until we drop Xcode 14.
+        //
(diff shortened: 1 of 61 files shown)


StripeApplePay/StripeApplePay/Source/ApplePayContext/STPApplePayContext.swift
```diff
@@ -145,7 +145,17 @@ public class STPApplePayContext: NSObject, PKPaymentAuthorizationControllerDeleg
     @available(macCatalystApplicationExtension, unavailable)
     @objc(presentApplePayWithCompletion:)
     public func presentApplePay(completion: STPVoidBlock? = nil) {
-        #if STP_BUILD_FOR_VISION
+        // A note on `canImport(CompositorServices)`, found here and elsewhere
+        // in the codebase: This is a terrible, terrible hack.
+        // We should use #if os(visionOS), but that will fail to compile
+        // on Xcode 14. Because we need to continue supporting both Xcode 14
+        // *and* building on visionOS, we instead check `canImport(CompositerServices)`, as CompositerServices currently
```
yuki-stripe, line 152: ooc would something like be an alternative?  
```
#if compiler(>=5.9) && os(visionOS) 
```

I kind of did something like this to use an Xcode 15-only type https://github.com/stripe/stripe-ios/blob/2781c735dfd10e7ff17d0c7deba1a98065f85ebb/Stripe/StripeiOS/Source/STPPaymentContext.swift#L334-L335

yuki-stripe (approved):

StripeApplePay/StripeApplePay/Source/ApplePayContext/STPApplePayContext.swift
davidme-stripe: I tried that, but Xcode 14 reads past the `compiler()` directive and fails on `os(visionOS)`. I couldn't figure out a way to structure the second option that didn't involve it being super clunky for `#else` statements.

PR #3128, opened 2023-12-13
Mark STPPinManagementService deprecated & suggest alternative

vbc-stripe:
## Motivation
https://jira.corp.stripe.com/browse/RUN_ISSUING_TED-2233

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### PaymentsUI
 * [Added] Adds support for [card brand choice](https://stripe.com/docs/card-brand-choice) to STPPaymentCardTextField and STPCardFormView. To set a default preferred network for these UI elements, use the new `preferredNetworks` parameter.
 
+* [Deprecated] Mark STPPinManagementService deprecated & suggest alternative.
 
 ## 23.19.0 2023-12-11
 ### Apple Pay
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
 ### PaymentsUI
 * [Added] Adds support for [card brand choice](https://stripe.com/docs/card-brand-choice) to STPPaymentCardTextField and STPCardFormView. To set a default preferred network for these UI elements, use the new `preferredNetworks` parameter.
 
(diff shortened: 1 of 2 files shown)


Stripe/StripeiOS/Source/STPPinManagementService.swift
```diff
@@ -13,6 +13,7 @@ import Foundation
 import UIKit
 
 /// STPAPIClient extensions to manage PIN on Stripe Issuing cards
+@available(*, deprecated, message: "Please use Issuing Elements instead: https://stripe.com/docs/issuing/elements")
```
wooj-stripe, line 16: 1. The link leads to a page which seems to focus on a web integration?
2. I would expect that the deprecation message to tell users what other class/integration point to use?

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 ### PaymentsUI
 * [Added] Adds support for [card brand choice](https://stripe.com/docs/card-brand-choice) to STPPaymentCardTextField and STPCardFormView. To set a default preferred network for these UI elements, use the new `preferredNetworks` parameter.
 
+* [Deprecated] Mark STPPinManagementService deprecated & suggest alternative.
```
wooj-stripe, line 12: 1. Will this require a major version change, or a minor version is acceptable?
2. Do we need to update the MIGRATING.md file? 
3. Seems like we've avoided using [Deprecated], and use [Changed] instead. ([example](https://github.com/stripe/stripe-ios/blob/master/CHANGELOG.md?plain=1#L372))

Stripe/StripeiOS/Source/STPPinManagementService.swift
vbc-stripe: Yeah that's intended, we don't have an native Issuing Elements yet so this is the only option at the moment. 

Please https://jira.corp.stripe.com/browse/RUN_ISSUING_TED-2233 for context

CHANGELOG.md
vbc-stripe: done

wooj-stripe (approved):

PR #3138, opened 2023-12-18
Miscellaneous Link bugfixes

davidme-stripe:
## Summary
* Fix an issue where the Link signup checkbox did not appear even when it was available during SetupIntents/PI+SFU.
* Remove an unnecessary use of `async`, which may have been slowing down Link URL generation

## Testing
In PaymentSheet Example, existing end-to-end tests


wooj-stripe (approved):
lgtm!

PR #3152, opened 2024-01-04
Disable link on usage of billingDetailsCollectionConfig

wooj-stripe:
## Summary
Disable link if user attempts to use BDCC

## Motivation
Configuring BDCC will not work in link popup.  Address info can not be passed via url navigate, etc.  Therefore, disable link in this case for now

## Testing
Manually testing, relying on unit tests


davidme-stripe:
Will this block saving to Link as well? There are a bunch of different Link availability paths unfortunately, so I'm wondering if this catches all of them. The other one is `canSaveToLink` in PaymentSheetFormFactory.swift.

wooj-stripe:
> Will this block saving to Link as well? There are a bunch of different Link availability paths unfortunately, so I'm wondering if this catches all of them. The other one is `canSaveToLink` in PaymentSheetFormFactory.swift.

Thanks for catching this. `canSaveToLink` would have been true, however, we would not have allowed saving to link becasue `isLinkEnabled` was being used in `shouldOfferLinkSignup` delegate.  Nonetheless, i think it makes sense to update the computed variable `canSaveToLink` consistent with isLinkEnabled.

davidme-stripe (approved):

PR #3159, opened 2024-01-05
Link Passthrough V2

davidme-stripe:
## Summary
* Pipe the `link_passthrough_mode_enabled` flag to the Link popup.
* Remove some dead test code.
* Make LinkEnabledPaymentMethodElement a ContainerElement, fixing some
code in PaymentSheet that relies on the ability to traverse elements.
(e.g. filling the address fields based on an AddressViewController)
* Adds support for signing up for Link in passthrough mode via the native flow. In this situation, we `/share` the payment details as a card payment method, then confirm the PI as usual with that PM ID.

## Testing
Manually in PaymentSheet Example, bringing up end-to-end signup test.

Note that passthrough mode will not be offered on the current SDK version because of a server-side gate.

## Changelog
None.


StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Extensions/Intent+Link.swift
```diff
@@ -9,16 +9,16 @@
 @_spi(STP) import StripePayments
 
 extension Intent {
-    var supportsLink: Bool {
-        return recommendedPaymentMethodTypes.contains(.link)
+    func supportsLink(allowV2Features: Bool) -> Bool {
+        return recommendedPaymentMethodTypes.contains(.link) || (linkPassthroughModeEnabled && allowV2Features)
```
wooj-stripe, line 13: This seems odd to me -- i would expect that `.link` would need to be contained within  `recommendedPaymentMethodTypes` ?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
```diff
@@ -369,7 +369,22 @@ extension PaymentSheet {
...
+                            if intent.linkPassthroughModeEnabled {
+                                // If passthrough mode, share payment details
+                                linkAccount.sharePaymentDetails(id: paymentDetails.stripeID) { result in
+                                    switch result {
+                                    case .success(let shareResponse):
+                                        confirmWithPaymentMethod(STPPaymentMethod(stripeId: shareResponse.paymentMethod))
+                                    case .failure(let error):
+                                        assertionFailure("Failed to share payment details \(error)")
+                                        // If this fails, confirm directly
+                                        confirmWithPaymentMethodParams(paymentMethodParams)
```
wooj-stripe, line 381: It feels weird that we have an assertionFailure here and then attempt to confirmWithPaymentMethodParams... if this is really the intended behavior, then maybe we should add some metrics w/ an alarm?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
```diff
@@ -166,6 +166,10 @@ extension PaymentSheet {
         // MARK: Internal
         internal var linkPaymentMethodsOnly: Bool = false
 
+        /// If enabled, PaymentSheet will offer V2 Link features, such as passthrough mode
+        /// 🚧 Under construction, do not use 🚧
+        @_spi(STP) public var allowLinkV2Features: Bool = false
```
wooj-stripe, line 171: Instead of STP, add some cautionary verbage to the _spi, e.g.: `@_spi(STP_UNDER_CONSTRUCTION)` ?

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Extensions/Intent+Link.swift
davidme-stripe: It's also possible that `supportsLink` is true even if Link is not a recommended PaymentMethod type, because in passthrough mode Link acts as a Card PaymentMethod.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
davidme-stripe: I kind of disagree with the assert here too but I'm following the existing code. Planned to clean it up and switch to alarms in MOBILESDK-1459

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
davidme-stripe: This is (and always will be) internal-only, so it should be scoped the same as our other internal imports. I'm deleting this flag at the end. (In MOBILESDK-1468)

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Extensions/Intent+Link.swift
davidme-stripe: Added a comment to clarify

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
davidme-stripe: I'll remove the "under construction" comment to avoid confusion

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
```diff
@@ -2232,419 +2236,77 @@ class PaymentSheetDeferredServerSideUITests: PaymentSheetUITestCase {
...
+        XCTAssert(phoneField.waitForExistence(timeout: 10))
+        phoneField.tap()
+        phoneField.typeText("3105551234")
+
+        // The name field is only required for non-US countries. Only fill it out if it exists.
+        let nameField = app.textFields["Name"]
+        if nameField.exists {
+            nameField.tap()
+            nameField.typeText("Jane Done")
+        }
```
wooj-stripe, line 2272: Seems like we should have two tests here:
* One test without a name field (US country)
* One test with a name field (non-us country)

wooj-stripe (approved):
We should probably update that test, but otherwise lgtm

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
davidme-stripe: MOBILESDK-1476

PR #3161, opened 2024-01-06
Financial Connections: v3 account picker modifications

kgaidis-stripe:
## Summary

This PR:
- Modifies Account Picker design to (mostly) match V3.

Details to understand before reviewing PR:
- This is one PR, out of expected ~50+, to change the design of Financial Connections from a "V2" design to a "V3" design. [Here is a link to the designs](https://www.figma.com/file/wXQ8NJApqSJiLmxxANDRTs/%E2%9C%A8-Connections-3.0?type=design&node-id=2524-217093&mode=design&t=kXaBiAlgVsC7YcH1-0).
- The code here is not the finished version. It's just an iteration towards the final version. This code will be re-visited later. The goal is to go screen-by-screen and make easy changes to learn about all the difficulties in each screen. Later, all the learnings will be combined to do a "final" polish of all screens.
- The above also applies to design. The design here is not the finished version.
- This PR is merged into a feature branch `fc-v3`. It is not merged into `master`.
- There might be TODO's scattered throughout the code. That's expected and they will be fixed later. 

## Testing

Simulator Screenshot - iPhone 15 Pro - 2024-01-10 at 14 01 20

https://github.com/stripe/stripe-ios/assets/105514761/53d5328e-48fa-4d96-ba60-5851c224cd7b


StripeFinancialConnections/StripeFinancialConnections/Source/Native/AccountPicker/AccountPickerViewController.swift
```diff
@@ -94,8 +90,19 @@ final class AccountPickerViewController: UIViewController {
                 )
                 self.didSelectLinkAccounts()
             },
-            didSelectMerchantDataAccessLearnMore: { [weak self] in
+            didSelectMerchantDataAccessLearnMore: { [weak self] url in
                 guard let self = self else { return }
+                SFSafariViewController.present(url: url)
+
+                // TODO(kgaidis): fix/return `dataAccessNotice`
```
kgaidis-stripe, line 97: this will get fixed later (may need backend changes)

StripeFinancialConnections/StripeFinancialConnections/Source/Native/AccountPicker/AccountPickerFooterView.swift
```diff
@@ -74,32 +78,29 @@ final class AccountPickerFooterView: UIView {
...
-        } else {
-            let singleAccountButtonTitle = STPLocalizedString(
-                "Link account",
-                "A button that allows users to confirm the process of saving their bank account for future payments. This button appears in a screen that allows users to select which bank accounts they want to use to pay for something."
-            )
-            let multipleAccountButtonTitle = STPLocalizedString(
-                "Link accounts",
-                "A button that allows users to confirm the process of saving their bank accounts for future payments. This button appears in a screen that allows users to select which bank accounts they want to use to pay for something."
-            )
+        let singleAccountButtonTitle = STPLocalizedString(
```
ccen-stripe, line 81: nit: we started to move newly added strings into`String+Localized.swift` for better readability, seems there's one for FC too, are you planning to add them there?

StripeFinancialConnections/StripeFinancialConnections/Source/Native/AccountPicker/RetrieveAccountsLoadingView.swift
```diff
@@ -0,0 +1,103 @@
...
+                }(),
+                title: STPLocalizedString(
+                    "Retrieving accounts...",
+                    "The title of the loading screen that appears when a user just logged into their bank account, and now is waiting for their bank accounts to load. Once the bank accounts are loaded, user will be able to pick the bank account they want to to use for things like payments."
+                ),
+                subtitle: nil,
+                contentView: {
+                    let verticalStackView = UIStackView(
+                        arrangedSubviews: [
+                            ShimmeringAccountPickerRow(),
```
ccen-stripe, line 34: Is this expected? if ShimmeringAccountPickerRow is only used here can we just pack the 4 dummy rows within it?

StripeFinancialConnections/StripeFinancialConnections/Source/Native/PartnerAuth/PrepaneView.swift
```diff
@@ -30,9 +30,13 @@ final class PrepaneView: UIView {
         let paneLayoutView = PaneLayoutView(
             contentView: PaneLayoutView.createContentView(
                 iconView: {
-                    let institutionIconView = InstitutionIconView()
```
ccen-stripe, line 33: I think the deleted approach is more elegant, is it possible to let the InstitutionIconView to render dummy view when `setImageUrl(nil)` ?

ccen-stripe (approved):

StripeFinancialConnections/StripeFinancialConnections/Source/Native/AccountPicker/AccountPickerFooterView.swift
kgaidis-stripe: FC right now is excluded from localization (part because we sometimes return strings from backend, so it wouldn't make sense to localize client, and then have non-localized strings from backend), so all of this is mostly to avoid future extra work for when we do plan to localize

I think FC was forced to create `String+Localized.swift` when localization WAS turned on because the script Stripe has blocked duplicate strings (even though that should actually not be a requirement - it can be handled automatically). So the file was used to solve that problem: reuse duplicate strings.

The pattern is debatable, but I always err on the side of simplicity: I _try_ to _only_ create abstractions when absolutely necessary (though I am definitely not perfect). So I try to only create extra files, extra abstractions, when I am "forced" to. This minimizes the chance of over-engineering. I can't claim that `String.Localized.linked_accounts` is worse, but I can't confidently claim that its better. It's completely debatable. As a result, the only thing I can rely on is principles: do the simplest thing possible and no more.

StripeFinancialConnections/StripeFinancialConnections/Source/Native/AccountPicker/RetrieveAccountsLoadingView.swift
kgaidis-stripe: Could you expand on what you mean by this? 

The `ShimmeringAccountPickerRow` is created to add some extra separation of code. It's marked as `private` so it can't be referenced outside of this file. Then here we just put 4 rows in a Vertical Stack to make a list of them.

ccen-stripe: I meant if there's always 4 of them added to `verticalStackView` and it's private, can we instead create a single view with the 4 rows hardcoded within

StripeFinancialConnections/StripeFinancialConnections/Source/Native/PartnerAuth/PrepaneView.swift
kgaidis-stripe: I do believe it renders a dummy (a placeholder image), but what the new code is trying to say is that: "if backend does not return an image, do not render it." The idea is that the backend is telling us that "null means don't render the UI element." 

In terms of not rendering _anything_ when `null` is passed, this is quite tricky (or maybe impossible with `UIKit`). Though `SwiftUI` does give options to handle this nicely (but we use UIKit here).

StripeFinancialConnections/StripeFinancialConnections/Source/Native/AccountPicker/AccountPickerFooterView.swift
ccen-stripe: I'm just used to all the strings defined in a file as android forces you to do so, but your concern is valid too

StripeFinancialConnections/StripeFinancialConnections/Source/Native/AccountPicker/RetrieveAccountsLoadingView.swift
kgaidis-stripe: Sorry, you'd maybe have to type out the idea 😅 

I could see someone creating an extra view to encapsulate the logic, but at the moment I don't see easy ways to simplify it. The `UIStackView` is there for layout. `ShimmeringAccountPickerRow` is to put the "4 row code" to be reusable 4 times (also utilizes `ShimmeringView` subclass which likely will be used elsewhere).

```
// just quick psuedocode
final class ShimmeringAccountPickerList: UIView {

    init() {
         super.init(frame: .zero)
                    let verticalStackView = UIStackView(
                        arrangedSubviews: [
                            ShimmeringAccountPickerRow(),
                            ShimmeringAccountPickerRow(),
                            ShimmeringAccountPickerRow(),
                            ShimmeringAccountPickerRow(),
                        ]
                    )
                    verticalStackView.axis = .vertical
                    verticalStackView.spacing = 16
                   addSubview(verticalStackView)
     }
}
```

The `ShimmeringView` will likely eventually be moved out to a separate file as there are other parts that will need some "shimmering," but I kept it here until that happens.

ccen-stripe: if it will be used elsewhere, then it makes sense to do it this way

StripeFinancialConnections/StripeFinancialConnections/Source/Native/AccountPicker/AccountPickerFooterView.swift
kgaidis-stripe: ahh yea thats a good point on Android

iOS localization is a little odd but there's an Apple script that looks through _all_ of the codebase for a specific pattern and then it creates a file to send to localization 

this process is a little magical but I like it from the perspective is that, in an optimal case, localization should be as easy as just typing a string in-line

its not as magic as typing a string in-line, part because creating a comment for localizers is helpful, but it comes a little closer than putting the strings in specific files

PR #3173, opened 2024-01-12
Update link sign up ui test to run in non-US locale

wooj-stripe:
## Summary
Added ability to run link signup test in non-us country

## Motivation
Name field is required in non-US countries

## Testing
Added test

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlaygroundSettings.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlaygroundSettings.swift
@@ -166,7 +166,7 @@ struct PaymentSheetTestPlaygroundSettings: Codable, Equatable {
         case off
     }
 
-    enum LinkOverrideCountry: String, PickerEnum {
+    enum UserOverrideCountry: String, PickerEnum {
         static var enumName: String { "Link override country" }
 
         case off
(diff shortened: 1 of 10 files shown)


StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/STPElementsSession.swift
```diff
@@ -11,6 +11,9 @@ import Foundation
 
 /// The response returned by v1/elements/sessions
 final class STPElementsSession: NSObject {
+    #if DEBUG && targetEnvironment(simulator)
+    public static let countryCodeOverride: String? = nil
```
davidme-stripe, line 15: Do we need to scope this to DEBUG/simulator? I can see pros and cons, but if we do this it won't work in TestFlight builds. Maybe that's okay, but we should point that out in the description in the test playground.

Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlaygroundSettings.swift
```diff
@@ -251,6 +258,7 @@ struct PaymentSheetTestPlaygroundSettings: Codable, Equatable {
     var defaultBillingAddress: DefaultBillingAddress
     var linkEnabled: LinkEnabled
     var linkV2Allowed: LinkV2Allowed
+    var linkOverrideCountry: LinkOverrideCountry
```
davidme-stripe, line 261: Maybe `overrideUserCountry` instead of `linkOverrideCountry`, I think this might have implications beyond Link

davidme-stripe:
Looks great overall, thanks!!!

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/STPElementsSession.swift
wooj-stripe: Hmmm -- I've changed this to DEBUG scope only, and updated the playground to note debug only

davidme-stripe (approved):

PR #3189, opened 2024-01-19
Send CVC for transaction after Link signup

davidme-stripe:
## Summary
Send CVC to the correct signup endpoints for both PM and passthrough mode.

## Motivation
We weren't passing along the CVC during Link signup transactions in PM or passthrough mode.

## Testing
CI, manually in PaymentSheet example.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Helpers/PaymentSheetLinkAccount.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Helpers/PaymentSheetLinkAccount.swift
@@ -227,7 +227,7 @@ class PaymentSheetLinkAccount: PaymentSheetLinkAccountInfoProtocol {
         }
     }
 
-    func sharePaymentDetails(id: String, completion: @escaping (Result<PaymentDetailsShareResponse, Error>) -> Void) {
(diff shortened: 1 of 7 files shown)


StripePaymentSheet/StripePaymentSheet/Source/Helpers/PaymentSheetLinkAccount.swift
```diff
@@ -428,7 +429,7 @@ extension PaymentSheetLinkAccount {
         params.link?.paymentDetailsID = paymentDetails.stripeID
         params.link?.credentials = ["consumer_session_client_secret": currentSession.clientSecret]
 
-        if let cvc = paymentDetails.cvc {
+        if let cvc = cvc {
```
wooj-stripe, line 432: I'm a little confused.  Previously, we were using ConsumerPaymentDetails.Details to pass around a CVC, but now we are passing it in directly?

Should we remove `ConsumerPaymentDetails.cvc`, or additional code in ConsumerPaymentDetails so that we don't use it?

davidme-stripe: Yeah, this is unused code from Link v1. I'll rip it out.

wooj-stripe (dismissed):
lgtm!

wooj-stripe (approved):

PR #3194, opened 2024-01-22
Use email address for accelerator

davidme-stripe:
## Summary
If an email address is passed in, show it in the accelerator.

## Motivation
MOBILESDK-1524

## Testing
CI, updating snapshot tests


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/PayWithLinkButton.swift
```diff
@@ -215,25 +215,23 @@ final class PayWithLinkButton: UIControl {
...
-    //        DispatchQueue.main.async { [weak self] in
-    //            self?.linkAccount = notification.object as? PaymentSheetLinkAccount
-    //        }
-    //    }
-    //    deinit {
-    //        // Stop listening for account changes
-    //        LinkAccountContext.shared.removeObserver(self)
-    //    }
+        // Listen for account changes
+        LinkAccountContext.shared.addObserver(self, selector: #selector(onAccountChange(_:)))
```
wooj-stripe, line 223: I believe that addObserver will have a strong reference to 'self', which means that `deinit{}` will never get called.  I verified w/ memory profiler that an instance of `PayWithLinkButton` stay in memory when reinitializing paymentsheet when the call to `addObserver` is made, and doesn't stay in memory when omitting this call.

I think if we want to use this type of add/remove observer we can add weak references to the observer.

wooj-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/PayWithLinkButton.swift
sfriedman-stripe: That definitely used to be true, but the behavior changed in iOS 9. From the [docs](https://developer.apple.com/documentation/foundation/nsnotificationcenter/1415360-addobserver#discussion):

> If your app targets iOS 9.0 and later or macOS 10.11 and later, you do not need to unregister an observer that you created with this function. If you forget or are unable to remove an observer, the system cleans up the next time it would have posted to it.

Maybe the way you're testing in the profiler isn't giving enough opportunity for the system to clean it up?

sfriedman-stripe: Ah no, I think I'm thinking of the class being observed leaking memory pre-iOS 9 if it's deallocated before `removeObserver` is called. Probably ignore me

PR #3201, opened 2024-01-23
Removed unused fields from consumers endpoint

wooj-stripe:
## Summary
Removing unused response params

## Motivation
Cleanup

## Testing
None, relying on compilation


davidme-stripe (approved):
Nice!

PR #3202, opened 2024-01-23
Link: If phone number and email are present, don't use phone number in inline signup

davidme-stripe:
## Summary
* Show "by providing your phone number" legal text when signing up with a prefilled email address, and show the phone number field by default. Expand it once a phone number is entered.
* If a user passes both an email address and a phone number, we leave the phone number field blank, and continue to treat filling it as the consent action. 
* Add a few PaymentSheet Playground options to more easily test these scenarios.

## Motivation
If a user passes in a default email address and phone number, we don't want to sign the consumer up for Link unless they opt in.

## Testing
Updated snapshot tests


Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
```diff
@@ -121,6 +120,19 @@ class PlaygroundController: ObservableObject {
...
         }
+        switch settings.defaultBillingAddress {
+        case .on:
+            configuration.defaultBillingDetails.email = "foo@bar.com"
+            configuration.defaultBillingDetails.phone = "+13105551234"
+        case .randomEmail:
+            configuration.defaultBillingDetails.email = "test-\(UUID().uuidString)@stripe.com"
+            configuration.defaultBillingDetails.phone = "+13105551234"
+        case .randomEmailNoPhone:
+            configuration.defaultBillingDetails.email = "test-\(UUID().uuidString)@stripe.com"
```
wooj-stripe, line 131: In efforts of trying to reduce the number of options, is it possible to remove `.on`, and just have:
* randomEmailPhone
* randomEmailNoPhone
* off

?

Stripe/StripeiOSTests/LinkInlineSignupElementSnapshotTests.swift
```diff
@@ -135,17 +138,27 @@ extension LinkInlineSignupElementSnapshotTests {
         configuration.defaultBillingDetails.name = preFillName
         configuration.defaultBillingDetails.phone = preFillPhone
 
+        var linkAccount: PaymentSheetLinkAccount?
+
+        if let prefilledEmailAddress = prefilledEmailAddress {
```
wooj-stripe, line 143: nit: `if let prefilledEmailAddress {`

Stripe/StripeiOSTests/LinkInlineSignupElementSnapshotTests.swift
```diff
@@ -135,17 +138,27 @@ extension LinkInlineSignupElementSnapshotTests {
...
             mode: mode,
             accountService: MockAccountService(),
+            linkAccount: linkAccount,
             country: country
         )
 
         viewModel.saveCheckboxChecked = saveCheckboxChecked
-        viewModel.emailAddress = emailAddress
+        // Won't trigger the "email address prefilled" path, because it wasn't there when initialized
+        if let emailAddress = emailAddress {
```
wooj-stripe, line 157: nit: `if let emailAddress {`

Stripe/StripeiOSTests/LinkInlineSignupElementSnapshotTests.swift
```diff
@@ -124,6 +126,7 @@ extension LinkInlineSignupElementSnapshotTests {
 
     func makeSUT(
         saveCheckboxChecked: Bool = false,
+        prefilledEmailAddress: String? = nil,
         emailAddress: String? = nil,
```
wooj-stripe, line 130: I'm having trouble understanding the difference between `emailAddress` and `prefilledEmailAddress` -- maybe it would help to update the name of `emailAddress` to differentiate between the two, or maybe a comment on what each parameter is instrumenting?

Stripe/StripeiOSTests/LinkInlineSignupElementSnapshotTests.swift
```diff
@@ -74,10 +74,12 @@ class LinkInlineSignupElementSnapshotTests: STPSnapshotTestCase {
         verify(sut)
     }
 
+    // WARNING: If this tests fails, see go/link-signup-consent-action-log to determine if a new consent_action is needed.
```
wooj-stripe, line 77: We should open a jira to update the doc w/ the updated screenshot.

Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
davidme-stripe: `.on` is nice because it uses a set email address (which is already registered for Link). I'd also like to keep it named `.on` to maintain compatibility with older QR codes.

Stripe/StripeiOSTests/LinkInlineSignupElementSnapshotTests.swift
davidme-stripe: TIL! 🫠

Stripe/StripeiOSTests/LinkInlineSignupElementSnapshotTests.swift
davidme-stripe: Will rename them for clarity

wooj-stripe (approved):

PR #3208, opened 2024-01-24
Financial Connections: for V3, various small polishes to Institution Picker

kgaidis-stripe:
## Summary

This PR:
- Added Institution Picker loading 
- Modified Institution Picker to have a scrollable header
- Modified sheets to have a different header font

Details to understand before reviewing PR:
- This is one PR, out of expected ~50+, to change the design of Financial Connections from a "V2" design to a "V3" design. [Here is a link to the designs](https://www.figma.com/file/wXQ8NJApqSJiLmxxANDRTs/%E2%9C%A8-Connections-3.0?type=design&node-id=2524-217093&mode=design&t=kXaBiAlgVsC7YcH1-0).
- The code here is not the finished version. It's just an iteration towards the final version. This code will be re-visited later. The goal is to go screen-by-screen and make easy changes to learn about all the difficulties in each screen. Later, all the learnings will be combined to do a "final" polish of all screens.
- The above also applies to design. The design here is not the finished version.
- This PR is merged into a feature branch `fc-v3`. It is not merged into `master`.
- There might be TODO's scattered throughout the code. That's expected and they will be fixed later. 

### Loading State Changed
 
| Before | After |
| --- | --- |
| before_picker_load | after_picker_loader |

### Sheet Header Font Changed
 
| Before | After |
| --- | --- |
| before_sheet | after_sheet |

#### Before

https://github.com/stripe/stripe-ios/assets/105514761/b1200c1b-9f81-4bd2-be68-2428bcebd4f7

#### After

https://github.com/stripe/stripe-ios/assets/105514761/9f5e9175-a820-44f7-8365-94b2daf73667

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/Native/InstitutionPicker/InstitutionTableLoadingView.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/Native/InstitutionPicker/InstitutionTableLoadingView.swift
@@ -6,11 +6,11 @@
 //
 
 import Foundation
-import UIKit
 @_spi(STP) import StripeUICore
+import UIKit
 
 final class InstitutionTableLoadingView: UIView {
-    
+
     init() {
         super.init(frame: UIScreen.main.bounds)
         let verticalStackView = UIStackView(
@@ -30,7 +30,7 @@ final class InstitutionTableLoadingView: UIView {
         ])
         clipsToBounds = true
     }
-    
+
     required init?(coder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
@@ -41,19 +41,11 @@ private class ShimmeringInstitutionRowView: ShimmeringView {
     override init(frame: CGRect) {
         super.init(frame: frame)
         backgroundColor = .clear
-        
-        let iconView = UIView()
-        iconView.backgroundColor = .backgroundOffset
-        iconView.translatesAutoresizingMaskIntoConstraints = false
-        NSLayoutConstraint.activate([
-            iconView.heightAnchor.constraint(equalToConstant: 56),
-            iconView.widthAnchor.constraint(equalToConstant: 56),
-        ])
-        
+
         let horizontalStackView = UIStackView(
             arrangedSubviews: [
                 CreateRowIconView(),
-                CreateRowMultipleLabelView()
+                CreateRowMultipleLabelView(),
             ]
         )
         horizontalStackView.axis = .horizontal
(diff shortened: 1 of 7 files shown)


StripeFinancialConnections/StripeFinancialConnections/Source/Native/InstitutionPicker/InstitutionTableLoadingView.swift
```diff
@@ -0,0 +1,136 @@
...
+        super.init(frame: frame)
+        backgroundColor = .clear
+        
+        let iconView = UIView()
+        iconView.backgroundColor = .backgroundOffset
+        iconView.translatesAutoresizingMaskIntoConstraints = false
+        NSLayoutConstraint.activate([
+            iconView.heightAnchor.constraint(equalToConstant: 56),
+            iconView.widthAnchor.constraint(equalToConstant: 56),
+        ])
```
bric-stripe, line 51: rm? `iconView` is never used (looks like this was moved in to the CreateRowIconView helper and just missed being cleaned up?)

StripeFinancialConnections/StripeFinancialConnections/Source/Native/InstitutionPicker/InstitutionTableView.swift
```diff
@@ -255,6 +242,14 @@ final class InstitutionTableView: UIView {
         }
     }
     
+    func setTableHeaderView(_ tableHeaderView: UIView?) {
+        if let tableHeaderView = tableHeaderView {
```
bric-stripe, line 246: nit: not sure if there's a swift version compatibility concern but can you use the shorthand? `if let tableHeaderView {`

bric-stripe (approved):

StripeFinancialConnections/StripeFinancialConnections/Source/Native/InstitutionPicker/InstitutionTableLoadingView.swift
kgaidis-stripe: yes, great catch, thank you!

StripeFinancialConnections/StripeFinancialConnections/Source/Native/InstitutionPicker/InstitutionTableView.swift
kgaidis-stripe: Changed it in the new commit! 

I think I am just stuck being "old school." The syntax feels weird because it looks like a "floating" 'uninitialized' variable:

```
let variable1: Int

// vs.

if let variable2 {

}
```

@bric-stripe out of curiosity, if you had to convince someone why `if let variable2` is better than `if let variable2 = variable2`, what would you say? The old way definitely looks worse/unnecessary when directly compared that way, but there is a question of cost in terms of readability/'learnability' (its another instance of "magic shorthand syntax" that, seemingly, only makes sense if you know what its a shorthand of)

bric-stripe: > if you had to convince someone why if let variable2 is better than if let variable2 = variable2, what would you say?

I don't feel strongly here but I think its one of the easier to infer bits of syntax so would just go for the fewer words option

PR #3230, opened 2024-01-30
Add default payment method to PaymentSheet loading analytic

yuki-stripe:
## Summary
Sends `selected_lpm` in the `mc_load_succeeded` analytic. This was pretty difficult, since the code that determines the default is inside the SavedPaymentOptionsVC which isn't necessarily loaded until after `mc_load_succeeded` is sent.

To make this work I extracted out the code to determine the SavedPaymentOptionsVC view models and default selection, and call it in load to know the value ahead of time. 

The long term fix might be to continue to create view models in the load path and then *flow that through* to the SavedPaymentOptionsVC - and more generally, generate view models for *all* the UI at load and then flow that through to the views.  

Also renames BI's equivalent analytic's param from `default_payment_method` to `selected_lpm`.

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-1562

## Testing
Unit tests + manual testing for "apple_pay", "link", "card", "us_bank_account", "none"

## Changelog
Not user facing

### DIFF
--- a/Stripe/StripeiOS/Source/STPAnalyticsClient+BasicUI.swift
+++ b/Stripe/StripeiOS/Source/STPAnalyticsClient+BasicUI.swift
@@ -51,7 +51,7 @@ extension STPPaymentContext {
             }()
             let params: [String: Any] = [
                 "duration": duration,
-                "default_payment_method": defaultPaymentMethod,
+                "selected_lpm": defaultPaymentMethod,
             ]
             log(event: event, params: params)
         }
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift
@@ -118,35 +118,6 @@ final class PaymentSheetLoader {
         return !configuration.isUsingBillingAddressCollection()
     }
 
-    /// Returns a list of viewmodels to display in SavedPaymentOptionsViewController aka the "saved PMs" screen.
-    static func makePaymentOptionsListViewModels(
-        savedPaymentMethods: [STPPaymentMethod],
-        configuration: PaymentSheet.Configuration,
-        showApplePay: Bool,
-        showLink: Bool
-    ) -> [SavedPaymentOptionsViewController.Selection] {
-        var savedPaymentMethods = savedPaymentMethods
-        let defaultPaymentMethod = CustomerPaymentOption.defaultPaymentMethod(for: configuration.customer?.id)
-
-        // Move default to front
-        if let defaultPMIndex = savedPaymentMethods.firstIndex(where: {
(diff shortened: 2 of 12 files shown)


Stripe/StripeiOS/Source/STPAnalyticsClient+BasicUI.swift
```diff
@@ -51,7 +51,7 @@ extension STPPaymentContext {
             }()
             let params: [String: Any] = [
                 "duration": duration,
-                "default_payment_method": defaultPaymentMethod,
+                "selected_lpm": defaultPaymentMethod,
```
porter-stripe, line 54: Will this break existing queries that expect default_payment_method to be populated?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift
```diff
@@ -72,14 +74,30 @@ final class PaymentSheetLoader {
...
                 let isLinkEnabled = isLinkEnabled(intent: intent, configuration: configuration)
-                analyticsClient.logPaymentSheetEvent(event: .paymentSheetLoadSucceeded,
-                                                                     duration: Date().timeIntervalSince(loadingStartDate))
+                let isApplePayEnabled = StripeAPI.deviceSupportsApplePay()
+                    && configuration.applePay != nil
+                    && intent.isApplePayEnabled
+
+                // Send load finished analytic
+                // This is hacky; the logic to determine the default selected payment method belongs to the SavedPaymentOptionsViewController. We invoke it here just to report it to analytics before that VC loads.
+                let (defaultSelectedIndex, paymentOptionsViewModels) = SavedPaymentOptionsViewController.makeViewModels(
```
porter-stripe, line 86: Does the default pm need to be sent in the loaded event? Seems like we have access to an initialized `SavedPaymentOptionsViewController` in `PaymentSheetViewController` and `PaymentSheetFlowControllerViewController`, which is where we also log the show event. Can we attach the default to the show event? Would that be less hacky?

Stripe/StripeiOS/Source/STPAnalyticsClient+BasicUI.swift
yuki-stripe: No, `default_payment_method` was added earlier this week.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift
yuki-stripe: I think so. For FlowController it's possible (likely, even, if you have a default PM) to never show the sheet before completing payment.

porter-stripe (approved):

porter-stripe (dismissed):

porter-stripe (approved):

PR #3231, opened 2024-01-30
Update separator text to be more generic for SetupIntents

porter-stripe:
## Summary
- When using a SetupIntent we currently mention "pay" instead we should be more generic only for SetupIntents.
- Confirmed the WalletHeaderView is not used in CustomerSheet.
- Downloaded latest strings from Lokalise.

## Motivation
https://jira.corp.stripe.com/browse/RUN_MOBILESDK-2861

## Testing
- Manual
- New snapshots

## Changelog
See diff


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/WalletHeaderView.swift
```diff
@@ -45,6 +45,7 @@ extension PaymentSheetViewController {
 
             static let applePay = WalletOptions(rawValue: 1 << 0)
             static let link = WalletOptions(rawValue: 1 << 1)
+            static let setupIntent = WalletOptions(rawValue: 1 << 2)
```
yuki-stripe, line 48: Shoot I didn't notice this before but this doesn't seem to make sense here, right? It implies the wallet can show Apple Pay, Link, and SetupIntent.

yuki-stripe (changes requested):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/WalletHeaderView.swift
porter-stripe: That's a fair point! I think it depends on the interpretation of `options`. If it is interpreted as "available payment methods" or "available payment options", then I agree it doesn't make much sense to add `setupIntent`. Although the term "options" is not very common language with respect to payment methods e.g. Apple Pay or Link throughout our codebase. I'm guessing the usage of "options" was mainly influenced by the usage of `OptionSet`. Given the existing definition of only containing `applePay` & `link`, I think it's reasonable to interpret it as "available payment methods", and maybe a supporting piece for this interpretation is the existence of `PaymentOption`, but I didn't make that connection.

My interpretation of it was more along the lines of options of the wallet, more like a configuration, and in this case one of the options/configuration of the wallet is to set up a payment method. Additionally, `options` felt naturally extensible for this given how `options` is used throughout `WalletHeaderView`, to determine what parts of the UI to show. Maybe this is an indication that we should rename `options` to `paymentOptions` or something along those lines to prevent `options` from being extended into a pseudo-configuration type.

I don't feel too strongly, so I'll update this to inject `isPaymentIntent` into `WalletHeaderView`.

yuki-stripe (approved):

PR #3240, opened 2024-02-01
Full screen blur on link webpopup

wooj-stripe:
## Summary
Added a full screen blur with checkmark

## Motivation
dogfooding feedback

## Testing
Manually tested.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
@@ -287,16 +287,17 @@ extension PaymentSheet: PaymentSheetViewControllerDelegate {
         }
     }
     func paymentSheetViewControllerFinishedOnPay(_ paymentSheetViewController: PaymentSheetViewController,
-                                                 completion: @escaping () -> Void) {
+                                                 completion: (() -> Void)? = nil) {
         self.bottomSheetViewController.removeBlurEffect(animated: true, completion: completion)
     }
 
-    func paymentSheetViewControllerCanceledOnPay(_ paymentSheetViewController: PaymentSheetViewController) {
-        self.bottomSheetViewController.removeBlurEffect(animated: true)
+    func paymentSheetViewControllerCanceledOnPay(_ paymentSheetViewController: PaymentSheetViewController,
+                                                 completion: (() -> Void)? = nil) {
+        self.bottomSheetViewController.removeBlurEffect(animated: true, completion: completion)
     }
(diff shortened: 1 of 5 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
```diff
@@ -286,6 +286,22 @@ extension PaymentSheet: PaymentSheetViewControllerDelegate {
             self.completion?(result)
         }
     }
+    func paymentSheetViewControllerFinishedOnPay(_ paymentSheetViewController: PaymentSheetViewController,
+                                                 completion: (() -> Void)? = nil) {
+        self.bottomSheetViewController.transitionSpinnerToComplete(animated: true) {
+            self.bottomSheetViewController.removeBlurEffect(animated: true, completion: completion)
+        }
```
wooj-stripe, line 293: If we don't care about the animation of the button turning from blue to green, we can make this:
```
-            self.bottomSheetViewController.removeBlurEffect(animated: true, completion: completion)
+            if let completion {
+                completion()
+            }
+            self.bottomSheetViewController.removeBlurEffect(animated: true)
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/BottomSheetViewController.swift
```diff
@@ -66,6 +80,73 @@ class BottomSheetViewController: UIViewController, BottomSheetPresentable {
         contentViewController = toVC
         return popped
     }
+    func addBlurEffect(animated: Bool, completion: @escaping () -> Void) {
+        if let containingSuperview = self.view.superview {
```
davidme-stripe, line 84: Not sure if this would always cover the entire screen, depending on where we're presenting the BottomSheetViewController from.

davidme-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/BottomSheetViewController.swift
wooj-stripe: I believe the BottomSheetViewController's view's parent is a uitransition view, which is expected to cover the entire screen, so I think we are good here, but if Apple changes something, then right -- may not always work.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
wooj-stripe: Design is in flux, we can always change it later.

PR #3244, opened 2024-02-02
Test analytics in PaymentSheet UI tests

yuki-stripe:
## Summary
- Make STPAnalyticsClient report to a new delegate when it sends an analytic. 
- Make the test playground a STPAnalyticsClientDelegate
- Add a hidden view to the playground whose accessibility label contains all analytics sent as a base64 encoded string
- Add asserts in our XCUITest against analytics.

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-1575

## Testing
Yes

## Changelog
Not user facing

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
@@ -639,7 +639,9 @@ extension PaymentSheet.IntentConfiguration.Mode {
 
 extension PlaygroundController: STPAnalyticsClientDelegate {
     func analyticsClientDidLog(analyticsClient: StripeCore.STPAnalyticsClient, payload: [String: Any]) {
-        analyticsLog.append(payload)
+        DispatchQueue.main.async {
+            self.analyticsLog.append(payload)
+        }
     }
 }
 
--- a/Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
@@ -12,10 +12,11 @@
 import Combine
 import Contacts
 import PassKit
+@_spi(STP) import StripeCore
 import StripePayments
-@_spi(STP) import StripePaymentSheet
-@_spi(STP) @_spi(PaymentSheetSkipConfirmation) import StripePaymentSheet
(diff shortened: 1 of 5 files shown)


davidme-stripe (approved):

yuki-stripe:
> Wondering if this is just a bad run or if the new changes introduced this 1hr time? Seems like all the commits in this PR have a long runtime for these tests. CleanShot 2024-02-06 at 10 48 31

Looks like 20 link-related test failures @ around 2-2.5 minutes each accounts for most of the test time

porter-stripe:
@yuki-stripe ah yep, looks all related to PaymentSheetUITest.swift:2751. I'll take a look, I think since we re-run each 5 times IIRC it adds up.

Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
```diff
@@ -628,6 +635,14 @@ extension PaymentSheet.IntentConfiguration.Mode {
     }
 }
 
+// MARK: - STPAnalyticsClientDelegate
+
+extension PlaygroundController: STPAnalyticsClientDelegate {
+    func analyticsClientDidLog(analyticsClient: StripeCore.STPAnalyticsClient, payload: [String: Any]) {
+        analyticsLog.append(payload)
```
porter-stripe, line 642: I think it might be legitimate failures, I ran your branch locally and noticed some Xcode warnings (purple) when I ran the failing tests but not on master (also didn't see failures on master).

You may need to wrap access to the `analyticsLog`  in a DispatchQueue.main.async. When I did this the test passed, the "Success!" text wasn't showing up in the failure case.

On a related note we should find a way to shorten the duration on these tests!

CleanShot 2024-02-06 at 13 12 40@2x

porter-stripe: I wonder if this is indicative of some other issue, e.g. it looks like getter on FlowController.paymentOption isn't being called on the main thread or something

porter-stripe (approved):

Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
yuki-stripe: TY!

PR #3247, opened 2024-02-02
Fix default selection and other FlowController behavior for Link

davidme-stripe:
## Summary
* In FlowController, the default PM should be one of Apple Pay, the first available saved payment method, or nothing.
* Don't show the SPM list at all in FlowController if Link is the only available PM. Act as if it isn't a saved PM.

This is unfortunately a hack — we don't want Link to surface the Saved PMs view, so we add it as a wallet button in FlowController. But FlowController doesn't otherwise support wallet buttons, so we have to hack around a lot of things to get it to work.

## Motivation
We're selecting Link as the default PM too aggressively, and we're also showing it in the Saved Payment Methods view even when no other PMs are available.

## Testing
CI, manual testing


Example/PaymentSheet Example/PaymentSheet Example/Resources/Assets.xcassets/AppIcon.solidimagestack/Front.solidimagestacklayer/Content.imageset/Contents.json
```diff
@@ -0,0 +1,13 @@
+{
+  "images" : [
+    {
```
porter-stripe, line 3: Are these supposed to be in here?

davidme-stripe: Oof good catch, committed these by mistake

porter-stripe (approved):

PR #3249, opened 2024-02-05
Fix ScannedCard name incorrectly being set to expiryYear

tom-matches:
## Summary
`ScannedCard`'s name was incorrectly being set to `expiryYear` instead of `name` in its `init`.


porter-stripe (approved):
Thank for this contribution!

PR #3254, opened 2024-02-05
Pass Link flags from elements/session to Link Popup URL

davidme-stripe:
## Summary
Pass the flags from elements/session to the Link popup URL.

## Motivation
Some features are missing when flags are not passed.

## Testing
In PaymentSheet Example, existing CI tests


porter-stripe (approved):
Non-blocking, but a question!

PR #3251, opened 2024-02-05
Experimental allowsRemovalOfLastSavedPaymentMethod

yuki-stripe:
## Summary
See [API Review](https://docs.google.com/document/d/12pqFEANgkDvvwFNWtXOAB_a4vYE0gikWCICj78Su8kQ/edit#heading=h.1kuxwrt88dsj)

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-1605

## Testing
Manually tested the cases [here](https://docs.google.com/document/d/12pqFEANgkDvvwFNWtXOAB_a4vYE0gikWCICj78Su8kQ/edit#heading=h.1kuxwrt88dsj)

Automated tests to follow in another PR

## Changelog
Not public.

PR #3263, opened 2024-02-07
Remove last saved pm PaymentSheet UI tests

yuki-stripe:
## Summary
Follow-up to https://github.com/stripe/stripe-ios/pull/3251

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
+++ b/Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
@@ -288,8 +288,13 @@ class PaymentSheetStandardUITests: PaymentSheetUITestCase {
         XCTAssertTrue(editButton.waitForExistence(timeout: 60.0))
         editButton.tap()
 
-        XCTAssertTrue(app.buttons["Remove"].waitForExistenceAndTap())
-        XCTAssertTrue(app.alerts.buttons["Remove"].waitForExistenceAndTap())
+        let removeButton = app.buttons["Remove"]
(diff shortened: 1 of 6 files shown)


Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
```diff
@@ -2155,6 +2155,117 @@ class PaymentSheetDeferredServerSideUITests: PaymentSheetUITestCase {
...
+        XCTAssertTrue(app.staticTexts["Edit"].waitForExistenceAndTap())
+        XCTAssertTrue(app.staticTexts["Done"].waitForExistence(timeout: 1)) // Sanity check "Done" button is there
+
+        // Remove one saved PM
+        XCTAssertNotNil(scroll(collectionView: app.collectionViews.firstMatch, toFindButtonWithId: "CircularButton.Remove")?.tap())
+        XCTAssertTrue(app.alerts.buttons["Remove"].waitForExistenceAndTap())
+
+        // Should be kicked out of edit mode now that we have one saved PM
+        XCTAssertFalse(app.staticTexts["Done"].waitForExistence(timeout: 1)) // "Done" button is gone - we are not in edit mode
+        XCTAssertFalse(app.staticTexts["Edit"].waitForExistence(timeout: 1)) // "Edit" button is gone - we can't edit
```
wooj-stripe, line 2212: Since we are updating the nav, maybe we should verify that the close button is still there? If you agree with this, then we can also do it for the flow controller case as well.

wooj-stripe (approved):

wooj-stripe (approved):

PR #3265, opened 2024-02-07
Fix autolayout issue on presentation with cvc recollection

wooj-stripe:
## Summary
Fix issue with autolayout with cvc recollection. View was not loaded in memory at the time.

## Motivation
Bug fix

## Testing
manual testing


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentOptionsViewController.swift
```diff
@@ -321,7 +321,9 @@ class SavedPaymentOptionsViewController: UIViewController {
 
     func didFinishPresenting() {
         DispatchQueue.main.asyncAfter(deadline: .now().advanced(by: .milliseconds(750))) {
-            self.updateFormElement()
+            if self.isViewLoaded {
+                self.updateFormElement()
+            }
```
yuki-stripe, line 326: Would be good to have a comment for why we wait 750 milliseconds and then call this.

yuki-stripe (approved):

PR #3266, opened 2024-02-07
Log out of Link after transaction completes

davidme-stripe:
## Summary
Invalidate the client secret by calling `log_out` immediately after the transaction completes.

Remove a bit of related unused code from Link v1.

## Testing
* CI end-to-end tests, manual testing of Link inline signup.
* Bring the ConsumerSessionsTests to test this scenario, ensure the token no longer works after logout.

### DIFF
--- a/Stripe/StripeiOSTests/ConsumerSessionTests.swift
+++ b/Stripe/StripeiOSTests/ConsumerSessionTests.swift
@@ -10,10 +10,10 @@ import XCTest
 
 @testable@_spi(STP) import Stripe
 @testable@_spi(STP) import StripeCore
+import StripeCoreTestUtils
 @testable@_spi(STP) import StripePayments
 @testable@_spi(STP) import StripePaymentSheet
 @testable@_spi(STP) import StripePaymentsUI
-import StripeCoreTestUtils
 
 class ConsumerSessionTests: XCTestCase {
 
@@ -21,7 +21,6 @@ class ConsumerSessionTests: XCTestCase {
         let apiClient = STPAPIClient(publishableKey: STPTestingDefaultPublishableKey)
         return apiClient
     }()
-
 
     func testLookupSession_noParams() {
         let expectation = self.expectation(description: "Lookup ConsumerSession")
@@ -157,7 +156,6 @@ class ConsumerSessionTests: XCTestCase {
             address.postalCode = "55555"
             address.country = "US"
             billingParams.address = address
-            
 
             let paymentMethodParams = STPPaymentMethodParams.paramsWith(
                 card: cardParams,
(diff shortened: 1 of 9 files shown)


Stripe/StripeiOSTests/ConsumerSessionTests.swift
```diff
@@ -139,8 +81,7 @@ class ConsumerSessionTests: XCTestCase {
 
         ConsumerSession.lookupSession(
             for: "mobile-payments-sdk-ci+not-a-consumer@stripe.com",
-            with: apiClient,
-            cookieStore: cookieStore
```
wooj-stripe, line 143: nit: probably should be: `not-a-consumer+\(UUID())`

Stripe/StripeiOSTests/ConsumerSessionTests.swift
```diff
@@ -175,9 +116,8 @@ class ConsumerSessionTests: XCTestCase {
             phoneNumber: "+13105551234",
             legalName: nil,
             countryCode: "US",
-            consentAction: nil,
-            with: apiClient,
-            cookieStore: cookieStore
+            consentAction: "clicked_checkbox_nospm_mobile_v0",
```
wooj-stripe, line 119: use: `PaymentSheetLinkAccount.ConsentAction.checkbox_v0.rawValue` ?
though, this makes me think the function should accept an enum vs a string

Stripe/StripeiOSTests/ConsumerSessionTests.swift
```diff
@@ -223,18 +164,39 @@ class ConsumerSessionTests: XCTestCase {
...
+                                case .failure(let error):
+
+                                    guard let stripeError = error as? StripeError,
+                                        case let .apiError(stripeAPIError) = stripeError else {
+                                        XCTFail("Received unexpected error response")
+                                        return
+                                    }
+                                    XCTAssertEqual(stripeAPIError.code, "consumer_session_credentials_invalid")
+                                }
+                                useDetailsAfterLogoutExpectation.fulfill()
```
wooj-stripe, line 194: Seems like this fulfill should be right under `XCTAssertEqual(stripeAPIError.code`, or we can fulfill prior to asserting.

davidme-stripe: If we fulfill prior to asserting, I think we have a chance of the main thread ending the test as successful before the assertion is checked. I've had to fix that issue in our other tests before.

wooj-stripe: Sorry, i messed up here.. I was thinking something that was just wrong. my bad!

wooj-stripe (approved):

PR #3273, opened 2024-02-09
Fix improper setup `STPPaymentOptionsViewController` in `PaymentContext`.

samer-stripe:
## Summary
Fix improper setup `STPPaymentOptionsViewController` in `PaymentContext`.

If a user chooses to wait for the `PaymentContext` to completely load, loading the payment options can cause its overall setup to not be taken into account because the `loadingPromise` it waits for will already be complete, causing `onSuccess` to be called synchronously before any additional setup is called. See the issue and this [thread](https://stripe.slack.com/archives/C01CY476ESJ/p1707359504378399) for more information.

## Motivation
Resolves [RUN_MOBILESDK-2929](https://jira.corp.stripe.com/browse/RUN_MOBILESDK-2929)

## Testing
To be added. I want to validate my approach first with the team.

### DIFF
--- a/Stripe/StripeiOS/Source/STPPaymentContext.swift
+++ b/Stripe/StripeiOS/Source/STPPaymentContext.swift
@@ -530,8 +530,10 @@ public class STPPaymentContext: NSObject, STPAuthenticationContext,
         if state == STPPaymentContextState.none {
             state = .showingRequestedViewController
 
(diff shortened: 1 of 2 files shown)


samer-stripe:
Already fixed is https://github.com/stripe/stripe-ios/pull/2314

PR #3278, opened 2024-02-09
Assert that a return URL is configured

porter-stripe:
## Summary
- Some merchants skip this step even though it is required. This leads to some LPMs or 3DS not working properly. We should not allow initialization of PaymentSheet without a return URL. Otherwise failures may happen silently when a return url is required.
- This change could lead to some merchants having crashes during development they didn't previously have, but it would only catch bad integrations.

## Motivation
- https://stripe.slack.com/archives/C067GRGBP3R/p1704486362992749
- https://stripe.slack.com/archives/C0523SWAA78/p1706892365064939?thread_ts=1706038686.959179&cid=C0523SWAA78

## Testing
Commented out the return url in an example view controller to see the failure

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
@@ -104,6 +104,9 @@ public class PaymentSheet {
     }
 
     required init(mode: InitializationMode, configuration: Configuration) {
+        assert(configuration.returnURL != nil,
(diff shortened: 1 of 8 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
```diff
@@ -104,6 +104,9 @@ public class PaymentSheet {
     }
 
     required init(mode: InitializationMode, configuration: Configuration) {
+        assert(configuration.returnURL != nil,
```
porter-stripe, line 107: It crossed my mind this might be too aggressive, and maybe we should assert the return URL in PaymentSheetLoader or something, but I figure the sooner in the process to assert the better.

PR #3286, opened 2024-02-12
Add a flag to disable Link signup if needed

davidme-stripe:
## Summary
Add a flag that disables Link signup, but not paying with Link.

## Testing
Added snapshot test and mocked session response.


StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Extensions/Intent+Link.swift
```diff
@@ -60,6 +60,17 @@ extension Intent {
...
     }
 
+    var disableLinkSignup: Bool {
+        switch self {
+        case .paymentIntent(let paymentIntent, _):
+            return paymentIntent.linkSettings?.disableSignup ?? false
+        case .setupIntent(let setupIntent, _):
+            return setupIntent.linkSettings?.disableSignup ?? false
+        case .deferredIntent(let elementsSession, _):
+            return elementsSession.linkSettings?.disableSignup ?? false
```
wooj-stripe, line 70: Maybe not a big deal, but if this flag ever goes away, should we default to `false` or `true` 🤔

wooj-stripe (approved):
Code looks good but, unclear what the correct default behavior should be.

PR #3288, opened 2024-02-13
Financial Connections: for v3, fixed a bug where sheet rotations would sometimes not relayout properly

kgaidis-stripe:
## Summary

This PR fixes a bug with sheet rotations where the sheet expands after rotation (when it should not). See "Before" video.

### Before

https://github.com/stripe/stripe-ios/assets/105514761/682cb565-915a-4f9c-9ad2-200fb9876492

### After

https://github.com/stripe/stripe-ios/assets/105514761/efc19ebf-a3e0-4fba-888b-d45d7568acc1


kgaidis-stripe:
doing some big branching changes so will merge this in quickly as it goes into feature branch (pretty simple stuff)

PR #3291, opened 2024-02-13
Add support for PaymentMethodConfigurations

davidme-stripe:
## Summary
Add support for [Payment Method Configurations](https://stripe.com/docs/payments/multiple-payment-method-configs) in the deferred intent flow.

See also: https://github.com/stripe/stripe-android/pull/7929

## Motivation
RUN_MOBILESDK-2935

## Testing
Added API test, tested manually in PaymentSheet Playground

## Changelog
Added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## X.X.X
+### PaymentSheet
+* [Added] Added support for [payment method configurations](https://docs.stripe.com/payments/multiple-payment-method-configs) when using the deferred intent integration path.
+
 ## 23.22.0 2024-02-12
 ### PaymentSheet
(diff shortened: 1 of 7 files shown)


porter-stripe (commented):
Do we need an API review for this? I don't really have FUD w/ the changes, and since the init has a default value it won't break any existing integrations. Just want to make sure we don't need API review.

davidme-stripe:
> Do we need an API review for this? I don't really have FUD w/ the changes, and since the init has a default value it won't break any existing integrations. Just want to make sure we don't need API review.

Yep, it's done: https://docs.google.com/document/d/1W09KQYdmBx7OvRZi1MauMKiXhrHdGPSeC_1hefh7POQ/edit

porter-stripe (approved):

PR #3315, opened 2024-02-21
Financial Connections: for v3, more various polish/fixes

kgaidis-stripe:
## Summary

- fixed an issue where device sheet orientation layout was buggy.
- fixed institution cell background color.
- increased search bar text field height.
- changed the sheet handle/knob color and radius.
- removed unnecessary prepane padding.

### Prepane Padding

| Before | After |
| --- | --- |
| prepane-padding-before | prepane-padding-after |

### Kob color / radius

| Before | After |
| --- | --- |
| sheet-knob-before | sheet-knob-after |

### Search Bar Height

| Before | After |
| --- | --- |
| before-search-bar-height | after-search-bar-height |

### Landscape Background Colors

| Before | After |
| --- | --- |
| before-background-color | after-background-color |

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/SheetViewController.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/SheetViewController.swift
@@ -132,13 +132,15 @@ class SheetViewController: UIViewController {
         super.viewDidLayoutSubviews()
 
         if panePresentationStyle == .sheet {
-            let isLandscape = UIDevice.current.orientation.isLandscape
+            // Note that using `UIDevice.current.orientation.isLandscape`
+            // performed worse (/ was buggy) when testing on device
+            let isLandscapePhone = UIDevice.current.userInterfaceIdiom == .phone && (UIScreen.main.bounds.size.width > UIScreen.main.bounds.size.height)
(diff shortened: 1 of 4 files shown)


StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/SheetViewController.swift
```diff
@@ -132,13 +132,15 @@ class SheetViewController: UIViewController {
         super.viewDidLayoutSubviews()
 
         if panePresentationStyle == .sheet {
-            let isLandscape = UIDevice.current.orientation.isLandscape
+            // Note that using `UIDevice.current.orientation.isLandscape`
+            // performed worse (/ was buggy) when testing on device
+            let isLandscapePhone = UIDevice.current.userInterfaceIdiom == .phone && (UIScreen.main.bounds.size.width > UIScreen.main.bounds.size.height)
```
kgaidis-stripe, line 137: not sure what logic iOS is using to calculate `isLandscape` but it was performing noticeably worse 

Logically, however, I could see how this new code is better anyway: the layout logic, when boiled down to first-principles, isn't really about "landscape" but rather the fact that height got significantly decreased. The idea of "landscape" is just an abstraction away from what we are trying to do.

bric-stripe (approved):

PR #3320, opened 2024-02-21
Financial Connections: for v3 added primary button shadows

kgaidis-stripe:
## Summary

^ added shadow to primary button

## Testing

subtle, but its there 😅 

| Before | After |
| --- | --- |
| before-cell-shadow | after-cell-shadow |

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/AccountPickerRowView.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/AccountPickerRowView.swift
@@ -26,8 +26,8 @@ final class AccountPickerRowView: UIView {
                     transform: nil
                 )
                 layer.shadowColor = UIColor.black.cgColor
-                layer.shadowRadius = 0.5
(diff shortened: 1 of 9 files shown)


StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/AccountPickerRowView.swift
```diff
@@ -26,8 +26,8 @@ final class AccountPickerRowView: UIView {
                     transform: nil
                 )
                 layer.shadowColor = UIColor.black.cgColor
-                layer.shadowRadius = 0.5
-                layer.shadowOpacity = 0.17
+                layer.shadowRadius = 1.5 / UIScreen.main.nativeScale
```
kgaidis-stripe, line 29: this is an intentional somewhat unrelated change but is still pictured in the test plan (the account rows)

bric-stripe (approved):
> subtle, but its there 😅

I had to open both images full screen and swap between the two to see any difference :) 👍

PR #3321, opened 2024-02-22
Apply fixes to commit scripts for latest tart/packer version

davidme-stripe:
## Summary
* Allow Cocoapods warnings, as Xcode 15 has a non-suppressable warning when building our combined Objective-C/Swift target: `WARN | [StripePayments/Stripe3DS2] xcodebuild: /var/folders/ff/_zpjygb51v7_f4nj9zqdkm640000gn/T/CocoaPods-Lint-20231128-51573-vz9cuq-StripePayments/App/main.m:4:9: warning: a function declaration without a prototype is deprecated in all versions of C [-Wstrict-prototypes]`
* Use the correct `packer init` command
* Set `, config: false` in the SSH gem, as some laptops have a misconfigured .sshconfig and are pointing this request to github.com.
* Lint the scripts

## Motivation
https://jira.corp.stripe.com/browse/RUN_MOBILESDK-2762

## Testing
During last deploy

### DIFF
--- a/ci_scripts/pod_tools.rb
+++ b/ci_scripts/pod_tools.rb
@@ -60,15 +60,13 @@ def capture_retrying(cmd, max_retries: 10)
   retries = 0
   loop do
     stdout, stderr, status = Open3.capture3(cmd)
-    if !status.success? && retries <= max_retries
-      retries += 1
-      delay = [2**retries, 32].min
-      puts "Something went wrong. Trying again in #{delay} seconds..."
-      sleep(delay)
-      puts 'Retrying...'
-    else
-      return stdout, stderr, status
-    end
+    return stdout, stderr, status unless !status.success? && retries <= max_retries
+
+    retries += 1
+    delay = [2**retries, 32].min
+    puts "Something went wrong. Trying again in #{delay} seconds..."
+    sleep(delay)
+    puts 'Retrying...'
   end
 end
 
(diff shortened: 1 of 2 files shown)


ci_scripts/vm_tools.rb
```diff
@@ -26,15 +26,16 @@ def bring_up_vm_and_wait_for_boot
...
-    rescue => exception
+    rescue StandardError => e
+      puts e
       puts 'Not online yet, retrying...'
       sleep(10)
     end
   end
   # Copy files to VM
-  run_command_vm("mkdir -p /tmp/build/stripe-ios")
+  run_command_vm('mkdir -p /tmp/build/stripe-ios')
```
wooj-stripe, line 38: If there's any chance that old files could be here, would be better to use `mktemp -d`

ci_scripts/pod_tools.rb
```diff
@@ -79,7 +77,7 @@ def push_retrying(podspec, expected_pod_version, max_retries: 10)
       puts "No need to push: #{podspec}.  Latest version is already #{expected_pod_version}"
       return true
     end
-    system("pod trunk push #{podspec} --synchronous")
+    system("pod trunk push #{podspec} --synchronous --allow-warnings")
```
wooj-stripe, line 80: thank you

wooj-stripe (approved):

ci_scripts/vm_tools.rb
davidme-stripe: We destroy the VM beforehand, so there shouldn't be

PR #3325, opened 2024-02-22
Fix query param double encoding on iOS 17

yuki-stripe:
## Summary
Uses `CFURLCreateWithString` instead of `URL(string:)` because the latter [changed its implementation ](https://developer.apple.com/documentation/foundation/url/3126806-init)in iOS 17 to automatically encode the string according to RFC 3986, which ends up double encoding certain characters. More details in comments of https://jira.corp.stripe.com/browse/MOBILESDK-1335. 

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-1658
https://jira.corp.stripe.com/browse/MOBILESDK-1335

## Testing
Ran NSMutableURLRequest+StripeTest on iOS 14, 15, 16, 17

## Changelog
Not user facing.

### DIFF
--- a/StripeCore/StripeCore/Source/Categories/NSMutableURLRequest+Stripe.swift
+++ b/StripeCore/StripeCore/Source/Categories/NSMutableURLRequest+Stripe.swift
@@ -10,13 +10,17 @@ import Foundation
 
 extension URLRequest {
     @_spi(STP) public mutating func stp_addParameters(toURL parameters: [String: Any]) {
-        guard let url = url else {
+        guard let url else {
             assertionFailure()
             return
         }
-        let urlString = url.absoluteString
         let query = URLEncoder.queryString(from: parameters)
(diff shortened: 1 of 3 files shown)


davidme-stripe:
But `CFURLCreateWithString` does not!? That's surprising — maybe they migrated this to [Swift Foundation](https://github.com/apple/swift-foundation) in iOS 17, so the behavior is slightly different.

yuki-stripe:
> But `CFURLCreateWithString` does not!? That's surprising — maybe they migrated this to [Swift Foundation](https://github.com/apple/swift-foundation) in iOS 17, so the behavior is slightly different. Great find!

That's what I'm seeing! H/T to SomeRandomiOSDev in https://forums.swift.org/t/url-string-behavior-changed-with-xcode-15-0-beta-5/66570/10

PR #3331, opened 2024-02-23
Don't deserialize luxe specs for backup elementSession instances

wooj-stripe:
## Summary
If the call to elementSessions fails, the SDK creates a backup instance of ElementsSession, which results in attempting to deserialize nil and emits an error.

## Motivation
We should avoid deserializing if we know that we failed the initial call.

## Testing
Manually, unfortunately.  Writing a test seems for this seems very difficult.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift
@@ -54,11 +54,15 @@ final class PaymentSheetLoader {
                 // Overwrite the form specs that were already loaded from disk
                 switch intent {
                 case .paymentIntent(let elementsSession, _):
-                    _ = FormSpecProvider.shared.loadFrom(elementsSession.paymentMethodSpecs as Any)
+                    if !elementsSession.isBackupInstance {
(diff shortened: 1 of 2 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift
```diff
@@ -54,11 +54,15 @@ final class PaymentSheetLoader {
                 // Overwrite the form specs that were already loaded from disk
                 switch intent {
                 case .paymentIntent(let elementsSession, _):
-                    _ = FormSpecProvider.shared.loadFrom(elementsSession.paymentMethodSpecs as Any)
+                    if !elementsSession.isBackupInstance {
+                        _ = FormSpecProvider.shared.loadFrom(elementsSession.paymentMethodSpecs as Any)
+                    }
```
porter-stripe, line 59: Is it possible for the backup session to ever have a valid form spec? Is it always guaranteed to be invalid?

wooj-stripe: Based on my understanding of the code, if the call to elements/sessions fails, then we try hitting the official `v1/*_intents` api, which is guaranteed not to have specs.

porter-stripe (approved):

PR #3338, opened 2024-02-26
Send Link card params correctly

davidme-stripe:
## Summary
The Consumer endpoint has a different format from the public API: Use the "preferred_network" field instead of "networks[preferred]", and move this code out to make the difference more explicit.

## Motivation
Fixing CBC + Link

## Testing
In PS Example, existing tests

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/STPAPIClient+Link.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/STPAPIClient+Link.swift
@@ -110,18 +110,13 @@ extension STPAPIClient {
 
         let billingParams = billingDetails.consumersAPIParams
 
-        var card = STPFormEncoder.dictionary(forObject: cardParams)["card"] as? [AnyHashable: Any]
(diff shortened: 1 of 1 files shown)


StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/STPAPIClient+Link.swift
```diff
@@ -110,18 +110,13 @@ extension STPAPIClient {
 
         let billingParams = billingDetails.consumersAPIParams
 
-        var card = STPFormEncoder.dictionary(forObject: cardParams)["card"] as? [AnyHashable: Any]
-        card?["cvc"] = nil // payment_details doesn't store cvc
```
wooj-stripe, line 114: We were previously niling out this field, but no longer seem to be doing so.

davidme-stripe: Instead, we're using `consumersAPIParams` and not adding the `cvc` field to begin with.

wooj-stripe (approved):

PR #3351, opened 2024-02-29
Financial Connections: for v3, various polish fixes

kgaidis-stripe:
## Summary

This PR contains various other polish fixes:
- fixed an issue where dark mode prepane would flash a dark background for institution  GIF.
- added better scroll behavior for pressing on search for more banks from institution picker.
- increased OTP height by 2 px.
- changed the institution picker loading animations.
- added shadow to consent logos and made primary button shadow stronger.

### Added shadow to logo and more shadow to primary button

| Before | After |
| --- | --- | 
| before_Logo_button_shadow | afteR_logo_button_shadow |

### Fixed an issue where programatically selecting search bar would scroll to the search bar oddly

| Before | After |
| --- | --- | 
| before_pressing_search_more | aftter_pressing_ssearch_more |

### Fixed an issue where a black overlay would temporarily appear while a GIF/image is loading

| Before | After |
| --- | --- | 
| before_gif_dark | after_gif_dark |

### Added different institution loaders (design request)

Not sure if actually better but that was the request 🤷 

#### Before

https://github.com/stripe/stripe-ios/assets/105514761/0ce80a21-886f-4902-9baa-aa8bb66100df

#### After

https://github.com/stripe/stripe-ios/assets/105514761/979f9b0d-2604-4814-bb20-df4344c7a12e

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/Native/InstitutionPicker/InstitutionTableView.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/Native/InstitutionPicker/InstitutionTableView.swift
@@ -85,9 +85,7 @@ final class InstitutionTableView: UIView {
         )
         return manualEntryTableFooterView
     }()
-    private lazy var loadingView: UIView = {
-        return InstitutionTableLoadingView()
-    }()
+    private var loadingView: UIView?
 
     init(frame: CGRect, allowManualEntry: Bool) {
         self.allowManualEntry = allowManualEntry
@@ -124,7 +122,6 @@ final class InstitutionTableView: UIView {
         tableView.register(InstitutionTableViewCell.self, forCellReuseIdentifier: cellIdentifier)
         tableView.delegate = self
         addAndPinSubview(tableView)
-        addAndPinSubviewToSafeArea(loadingView)
         showLoadingView(false)
     }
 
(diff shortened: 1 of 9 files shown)


StripeFinancialConnections/StripeFinancialConnections/Source/Native/InstitutionPicker/InstitutionTableView.swift
```diff
@@ -218,8 +215,16 @@ final class InstitutionTableView: UIView {
...
 
     func showLoadingView(_ show: Bool) {
-        loadingView.isHidden = !show
-        bringSubviewToFront(loadingView)  // defensive programming to avoid loadingView being hiddden
+        if show {
+            if loadingView?.superview == nil {
+                let loadingView = InstitutionTableLoadingView()
+                addAndPinSubviewToSafeArea(loadingView)
+                self.loadingView = loadingView
+            }
```
bric-stripe, line 223: if show is true and the superview != nil any need to still run the `bringSubviewToFront`?

bric-stripe (approved):

StripeFinancialConnections/StripeFinancialConnections/Source/Native/InstitutionPicker/InstitutionTableView.swift
kgaidis-stripe: thank you for the check - the `bringSubviewToFront` previously was defensive programming where _if_ the code changed, the loading view would always go to top

what's unique about code now is that instead of persisting the loading view (ex. it would always be part of view hierarchy), it now gets re-added every single time we show/hide; this _essentially_ acts as the `bringSubviewToFront` because its brought to front each time we change the loading (rather than at `init`). there's certainly still _a chance_ for a 'bug' but its much less and given my current understanding of the code, is impossible atm

PR #3384, opened 2024-03-07
Adjust layout for consumer assets

davidme-stripe:
## Summary
* Added latest consumer assets behind a flag. The flag intentionally defaults to `true`, but it currently set to `false` server-side.
* Updated snapshot tests to use new assets, added legacy tests for older assets
* Added a singleton (I'm sorry) to set this state from the `elements/sessions` endpoint. This made more sense than trying to pass the bool down to each view in the hierarchy and the public-facing PaymentOption. We'll get rid of the singleton once the flag is removed server-side in the next few months.
* Removed some unused Link UI theme, and stopped using the Link UI theme for the OTP UI snapshot test (as it was only used in that test)

## Testing
Existing snapshot tests, CI


StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Views/LinkMoreInfoView.swift
```diff
@@ -15,12 +15,21 @@ import UIKit
 @objc(STP_Internal_LinkMoreInfoView)
 final class LinkMoreInfoView: UIView {
     struct Constants {
-        static let logoWidth: CGFloat = 31
-        static let logoHeight: CGFloat = 14
+        static var logoWidth: CGFloat {
+            LinkUI.useNewBrand ? 42 : 31
+        }
+        static var logoHeight: CGFloat {
+            LinkUI.useNewBrand ? 14 : 14
```
wooj-stripe, line 22: arguably, we can just return 14?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift
```diff
@@ -84,6 +84,9 @@ final class PaymentSheetLoader {
                     && configuration.applePay != nil
                     && intent.isApplePayEnabled
 
+                // If Link is enabled, set the `useNewBrand`
+                LinkUI.useNewBrand = intent.elementsSession.linkSettings?.useRebrand ?? false
```
wooj-stripe, line 88: This is a little bit confusing, because the deserlization logic defaults to true, but this defaults to false.  Ideally, we'd have one place where we have defaulting logic.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/PayWithLinkButton.swift
```diff
@@ -109,13 +109,25 @@ final class PayWithLinkButton: UIControl {
...
         let linkAttachment = NSTextAttachment(image: linkImage)
-        let linkLogoRatio = linkImage.size.width / linkImage.size.height
 
-        // Link logo should be a little taller than the height of a cap in SF Medium
-        let linkLogoHeight = linkView.font.capHeight + (linkView.font.pointSize * 0.1)
-        linkAttachment.bounds = CGRect(x: 0, y: 0, width: linkLogoHeight * linkLogoRatio, height: linkLogoHeight)
+        if LinkUI.useNewBrand {
+            let linkLogoRatio = linkImage.size.width / linkImage.size.height
+
+            let linkTextSpacing = 0.073 // the total top+bottom space outside the Link logo
```
wooj-stripe, line 118: Looks like these two lines are shared between the if/else and can be pulled out before the if statement?

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Extensions/UIColor+Link.swift
```diff
@@ -17,110 +17,40 @@ extension UIColor {
...
-    static let linkNavTint: UIColor = .dynamic(
-        light: UIColor(red: 0.188, green: 0.192, blue: 0.239, alpha: 1.0),
-        dark: UIColor(red: 0.922, green: 0.922, blue: 0.961, alpha: 0.6)
-    )
-
     /// Color for borders and dividers.
     static let linkSeparator: UIColor = .dynamic(
         light: UIColor(red: 0.878, green: 0.902, blue: 0.922, alpha: 1),
         dark: UIColor(red: 0.471, green: 0.471, blue: 0.502, alpha: 0.36)
     )
```
wooj-stripe, line 41: Looks like `linkPrimaryText`, `linkSeparator`, and `linkSecondaryButtonBackground` are only being used by BacsDDMandateview -- maybe we should these over to that view?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift
davidme-stripe: I think the logic in my mind was like:
* If the elements/sessions response doesn't contain the bool, we return `true` (as the bool has probably been intentionally removed)
* If the elements/sessions response doesn't exist, we return `false` (because the `elements/sessions` request failed, so we don't want the new brand to leak to a user?)

But yeah, maybe it doesn't make sense. Ideally linkSettings would be non-optional, but that feels a little dangerous. I'll just simplify them and make it `true`, it's probably fine if it accidentally shows in some cases.

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Extensions/UIColor+Link.swift
davidme-stripe: Good catch!

StripePayments/StripePayments/Source/API Bindings/Models/Shared/LinkSettings.swift
```diff
@@ -61,6 +64,8 @@ import Foundation
         let webviewOption = PopupWebviewOption(rawValue: response["link_popup_webview_option"] as? String ?? "")
         let passthroughModeEnabled = response["link_passthrough_mode_enabled"] as? Bool ?? false
         let disableSignup = response["link_mobile_disable_signup"] as? Bool ?? false
+        // Should default to true in case we remove the server-side flag someday
+        let useRebrand = response["link_2024_rebrand_m1"] as? Bool
```
wooj-stripe, line 68: The comment indicates defaulting to true, but there doesn't seem to be a default value?

davidme-stripe: Fixed, thanks!

wooj-stripe (approved):

PR #3397, opened 2024-03-11
[Identity] Fix paddings of text on top of buttons

ccen-stripe:
## Summary

Fix a padding issue where the consent button might incorrectly kept disabled when scroll to bottom and the Stripe Privacy Policy text overlaps with content

## Motivation

Butter UX

## Testing

- [x] Screenshot test

## Recordings
| Before  | After |
| before  | after |


StripeIdentity/StripeIdentity/Source/NativeComponents/Views/IdentityFlowView.swift
```diff
@@ -114,6 +118,12 @@ class IdentityFlowView: UIView {
         return buttonBackgroundView
     }()
 
+    private let buttonTopBackgroundView: UIView = {
+        let buttonTopBackgroundView = UIView()
+        buttonTopBackgroundView.backgroundColor = .systemBackground
```
kgaidis-stripe, line 123: `systemBackground` _might_ be different in dark/light mode? so just adding a note in case that's important to check

ccen-stripe: yes, this is expected and also verified in darkmode, we also use the same property for `buttonBackgroundView` at line 117

StripeIdentity/StripeIdentity/Source/NativeComponents/Views/IdentityFlowView.swift
```diff
@@ -231,12 +244,17 @@ extension IdentityFlowView {
...
         buttonTopContentView.translatesAutoresizingMaskIntoConstraints = false
 
         NSLayoutConstraint.activate([
             // Constrain buttonTop top of buttons
-            buttonTopContentView.centerXAnchor.constraint(equalTo: centerXAnchor),
-            buttonTopContentView.bottomAnchor.constraint(equalTo: buttonBackgroundView.topAnchor, constant: -Style.buttonTopBottomPadding),
+            buttonTopBackgroundView.leadingAnchor.constraint(equalTo: leadingAnchor),
+            buttonTopBackgroundView.trailingAnchor.constraint(equalTo: trailingAnchor),
+            buttonTopBackgroundView.bottomAnchor.constraint(equalTo: buttonBackgroundView.topAnchor),
+            buttonTopContentView.centerXAnchor.constraint(equalTo: buttonTopBackgroundView.centerXAnchor),
```
kgaidis-stripe, line 257: not sure if I am understanding the layout correctly but adding a note just in case: `centerXAnchor` will center but will likely not handle the layout properly if `buttonTopContentView` width is greater than its parent (it might clip?)

StripeIdentity/StripeIdentity/Source/NativeComponents/Views/IdentityFlowView.swift
```diff
@@ -28,15 +28,19 @@ class IdentityFlowView: UIView {
         )
         static let buttonSpacing: CGFloat = 10
         static let buttonInsets = NSDirectionalEdgeInsets(
-            top: 8,
+            top: 4,
             leading: 16,
             bottom: 8,
             trailing: 16
         )
+        static let buttontopInsets = NSDirectionalEdgeInsets(
```
kgaidis-stripe, line 36: not sure if it applies here, but adding a note of `directionalLayoutMargins` on UIStackView in case it is helpful (I tend to use that a lot even for just padding as its easier than autolayout):

https://github.com/stripe/stripe-ios/blob/89af67e0/StripeFinancialConnections/StripeFinancialConnections/Source/Native/AccountPicker/AccountPickerSelectionRowView.swift#L110-L116

kgaidis-stripe (approved):

StripeIdentity/StripeIdentity/Source/NativeComponents/Views/IdentityFlowView.swift
ccen-stripe: yes, but this field is always `Stripe Privacy Policy`

StripeIdentity/StripeIdentity/Source/NativeComponents/Views/IdentityFlowView.swift
ccen-stripe: hmm it seems it does't work on buttonTopBackgroundView, which is an UIView instead of UIStackView

PR #3402, opened 2024-03-13
Update docs for CartesBancaires support on ApplePay

bernardo-stripe:
## Summary
Update comments/doc on Cartes Bancaires usage for Apple Pay.

## Motivation
Improve documentation for the release of Cartes Bancaires Apple Pay adoption on iOS.

## Testing

Duplicated the test for the similar code for JCB.

## Changelog

* Adds configurable support for Cartes Bancaires (Apple Pay). [#1158](https://github.com/stripe/stripe-ios/pull/3402)


StripeCore/StripeCore/Source/API Bindings/StripeAPI.swift
```diff
@@ -68,9 +68,33 @@ import PassKit
             }
         }
     }
+
+    /// Users that have the Payment Method Cartes Bancaires set to Active
+    /// can enable Cartes Bancaires for Apple Pay by setting this to `YES`.
+    ///
+    /// The default value is NO.
+    @objc public class var cartesBancairesPaymentNetworkSupported: Bool {
```
bernardo-stripe, line 76: A question for the reviewer:
I see on the [PaymentSheet docs](https://stripe.dev/stripe-ios/stripepaymentsheet/documentation/stripepaymentsheet/additionalenabledapplepaynetworks) that to add an additional network to the Payment Sheet Apple Pay we should use `StripeAPI.additionalEnabledApplePayNetworks`, so I'm assuming this would be enough to support CB Apple Pay on Payment Sheet too?

wooj-stripe (commented):
Looks like we can add an additional tests in STPApplePaySwiftTest (similar to testAdditionalPaymentNetwork)?

StripeCore/StripeCore/Source/API Bindings/StripeAPI.swift
wooj-stripe: Seems that way considering PaymentSheet (and CustomerSheet) use `deviceSupportsApplePay`, which hooks into `self.supportedPKPaymentNetworks()`, which includes `additionalEnabledApplePayNetworks`.

cc: @davidme-stripe

StripeCore/StripeCore/Source/API Bindings/StripeAPI.swift
```diff
@@ -68,9 +68,33 @@ import PassKit
...
+        get {
+            return self.additionalEnabledApplePayNetworks.contains(.cartesBancaires)
+        }
+        set(CartesBancairesPaymentNetworkSupported) {
+            if CartesBancairesPaymentNetworkSupported
+                && !self.additionalEnabledApplePayNetworks.contains(.cartesBancaires)
+            {
+                self.additionalEnabledApplePayNetworks = [PKPaymentNetwork.cartesBancaires] +
+                    self.additionalEnabledApplePayNetworks
```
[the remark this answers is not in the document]
bernardo-stripe: Yes, we try to go through CB before other gateways, because it's usually more cost effective.
We currently prepend Cartes Bancaires to the list of supported networks on Apple Pay web. 
The outcome here still isn't ideal, because `additionalEnabledApplePayNetworks` is appended to the "default" networks list [here](https://github.com/stripe/stripe-ios/blob/da89a2cd/StripeCore/StripeCore/Source/API%20Bindings/StripeAPI.swift#L109).
Ideally we would prepend CB to the whole list of supportedPKPaymentNetworks, but this is out of the scope of this PR may be a future optimization.

davidme-stripe (changes requested):
Sorry, but I don't think we want to extend this pattern — the JCB-only option was only added because we didn't support any other card networks at that time. We should add `StripeAPI.additionalEnabledApplePayNetworks = [.cartesBancaires]` to the docs instead.

I'd support a change to move CB to the front of the `additionalEnabledApplePayNetworks` list when available.

bernardo-stripe:
Thank you @davidme-stripe and @wooj-stripe.
I'll follow up with another PR with your suggestion to move CB up the list.

I removed the new code and just left the update in the comments and test.

StripeCore/StripeCore/Source/API Bindings/StripeAPI.swift
```diff
@@ -116,6 +117,8 @@ import PassKit
     /// American Express, Visa, Mastercard, Discover, Maestro.
     /// Japanese users can enable JCB by setting `JCBPaymentNetworkSupported` to YES,
     /// after they have been approved by JCB.
+    /// Users that have the Payment Method Cartes Bancaires set to Active, can enable it
+    /// by setting `CartesBancairesNetworkSupported` to YES.
```
wooj-stripe, line 121: Is this true?  I'm not seeing a variable named `CartesBancairesNetworkSupported` that I can set to YES/True

wooj-stripe (commented):
Is the plan just to update docs?  If so, can we change the commit title to something like "Update docs for CartesBancaires support on ApplePay"

StripeCore/StripeCore/Source/API Bindings/StripeAPI.swift
davidme-stripe: Correct!

davidme-stripe:
@bernardo-stripe This LGTM! But if you want to default to CB, I'd support moving `additionalEnabledApplePayNetworks` to the front of the list (in `@_spi(STP) public class func supportedPKPaymentNetworks() -> [PKPaymentNetwork] `). That way the user's specified networks (CB/JCB) will take precedence over the default networks.

My only question is how that would work for a Discover card: Would it then be processed on JCB? I don't think so, just want to make sure.

davidme-stripe (approved):

PR #3421, opened 2024-03-15
Add option to update add card label.

nschris-stripe:
## Summary
This allows us to customize the header text on the add card bottom sheet. This will allow us to provide more context around what customer the card is being added to.


kgaidis-stripe (dismissed):
LGTM (has spi gating and is harmless even if we want to revert) but might want to check with payments folks just in case (ex. @.davidestes )

porter-stripe:
Updating branch, I think it was running an old CI script causing failures.

porter-stripe (dismissed):

porter-stripe (approved):

PR #3432, opened 2024-03-20
Analytics helper method to extract details about Errors 

yuki-stripe:
## Summary
Added helpers so that you can send analytics for errors like this:
```
struct MyError: Error {
  case foo
  case bar
}

analyticsClient.log(event: ErrorAnalytic(event: .someErrorEvent, error: MyError.foo))
```

Under the hood, it extracts error_type and error_code from the error in a standardized way for different errors. 

A future PR will let allow Errors to declare extra details to be logged or override `error_code` and `error_type`.

## Testing
See unit tests

## Changelog
Not user facing

### DIFF
--- a/StripeCore/StripeCore/Source/Analytics/AnalyticLoggableError.swift
+++ b/StripeCore/StripeCore/Source/Analytics/AnalyticLoggableError.swift
@@ -93,17 +93,17 @@ where Self: RawRepresentable, Self.RawValue == String {
     /// - For NSErrors, the error domain e.g. “NSURLErrorDomain”.
     static func extractErrorType(from error: Error) -> String {
         if type(of: error) is NSError.Type {
+            // Note: checking `error as NSError` always succeeds because Swift errors are bridged - this ensures the error is an instance of NSError.
             let error = error as NSError
             if error.domain == STPError.stripeDomain, let stripeAPIErrorType = error.userInfo[STPError.stripeErrorTypeKey] as? String {
                 // For Stripe API Error, use the error type key's value
                 return stripeAPIErrorType
             } else {
-                // Default behavior for other errors.
-                // Note: For Swift Error enums, `domain` is the type name
-                // e.g. `LinkURLGeneratorError.noPublishableKey` -> "StripePaymentSheet.LinkURLGeneratorError"
+                // For other NSErrors, use the domain
                 return "\(error.domain)"
             }
         } else {
+            // This is a Swift Error, use the qualified type name e.g. "Swift.DecodingError" or "StripePaymentSheet.PaymentSheetError"
             return String(reflecting: type(of: error))
         }
     }
--- a/StripeCore/StripeCore/Source/Analytics/AnalyticLoggableError.swift
+++ b/StripeCore/StripeCore/Source/Analytics/AnalyticLoggableError.swift
@@ -31,6 +31,8 @@ where Self: RawRepresentable, Self.RawValue == String {
 }
 
 @_spi(STP) extension Error {
+    /// Serialize an Error for logging, suitable for the Identity and Financial Connections SDK.
+    // TODO: Rename to serializeForIdentityAndFinancialConnectionsSDKLogging()
     public func serializeForLogging() -> [String: Any] {
         if let loggableError = self as? AnalyticLoggableError {
             return loggableError.analyticLoggableSerializeForLogging()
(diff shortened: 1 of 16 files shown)


StripeCore/StripeCore/Source/Analytics/AnalyticLoggableError.swift
```diff
@@ -65,4 +67,79 @@ where Self: RawRepresentable, Self.RawValue == String {
...
+        } else {
+            return String(reflecting: type(of: error))
+        }
+    }
+    
+    /// Extracts a value suitable for the `"error_code"` analytic parameter
+    /// - For Stripe API errors, the error's code e.g. "invalid_number".
+    /// - For NSErrors, the error code e.g. “-1009”.
+    /// - For Swift errors, the enum case name as a string for Swift errors e.g. “noPublishableKey”.
+    static func extractErrorCode(from error: Error) -> String {
```
yuki-stripe, line 114: The code is pretty complicated :(. Turns out it's really hard to get details about Errors if you have to avoid invoking "description" or "debugDescription"!  And relying on `(error as NSError).code` leads to confusing values if the enum has associated values, which is a common thing - the cases with associated values seem to get ordered first. For example:
```
Enum MyError: Error {
 Case zero
 Case one(foo: String)
}
print((MyError.zero as NSError).code) // 1
print((MyError.one(foo: “”) as NSError).code) // 0
```

I'm really open to ideas to improve it. If it's too gnarly we can just use `(error as NSError).code` and deal with the confusing values.

StripeCore/StripeCore/Source/Analytics/STPAnalyticsClient.swift
```diff
@@ -97,12 +96,7 @@ import UIKit
         payload["additional_info"] = additionalInfo()
         payload["product_usage"] = productUsage.sorted()
 
-        // Attach error information if this is an error analytic
-        if let errorAnalytic = analytic as? ErrorAnalytic {
-            payload["error_dictionary"] = errorAnalytic.error.serializeForLogging()
-        }
```
yuki-stripe, line 103: I confirmed that this error_dictionary is unused.

StripeCore/StripeCore/Source/Analytics/AnalyticLoggableError.swift
```diff
@@ -65,4 +68,69 @@ where Self: RawRepresentable, Self.RawValue == String {
...
+        ]
+    }
+
+    /// Extracts a value suitable for the `"error_type"` analytic parameter
+    /// - For Stripe API errors, the error’s [type](https://docs.stripe.com/api/errors#errors-type) e.g. “invalid_request_error”.
+    /// - For Swift errors, the fully qualified type name e.g. “StripePaymentSheet.LinkURLGeneratorError”.
+    /// - For NSErrors, the error domain e.g. “NSURLErrorDomain”.
+    static func extractErrorType(from error: Error) -> String {
+        if type(of: error) is NSError.Type {
+            let error = error as NSError
```
porter-stripe, line 96: Does this if statement catch anything that `if let error = error as? NSError` wouldn't?

porter-stripe (approved):

StripeCore/StripeCore/Source/Analytics/AnalyticLoggableError.swift
```diff
@@ -65,4 +68,69 @@ where Self: RawRepresentable, Self.RawValue == String {
...
+    /// Serialize an Error for logging to q.stripe.com and the `sdk.analytics_events` table
+    ///
+    /// It sends the following fields:
+    /// - error_type: For Stripe API errors, the error’s [type](https://docs.stripe.com/api/errors#errors-type) e.g. “invalid_request_error”.
+    ///           For Swift errors, the fully qualified type name e.g. “StripePaymentSheet.LinkURLGeneratorError”.
+    ///           For NSErrors, the error domain e.g. “NSURLErrorDomain”.
+    /// - error_code: For Stripe API errors, the error's code e.g. "invalid_number".
+    ///            For NSErrors, the error code e.g. “-1009”.
+    ///            For Swift errors, the enum case name as a string for Swift errors e.g. “noPublishableKey”.
+    public func serializeForV1Analytics() -> [String: Any] {
```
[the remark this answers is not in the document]
yuki-stripe: Only that theoretically someone might want to make their Error send an Int or Bool or whatever as a value. Happy to change to String if you think it makes more sense.

StripeCore/StripeCore/Source/Analytics/AnalyticLoggableError.swift
yuki-stripe: Oh this code's comments need to be improved, thanks for bringing it up.

Casting an Error to NSError always succeeds - Swift automatically bridges them. This line checks that the error is an actual instance of NSError.

porter-stripe (approved):

PR #3445, opened 2024-03-21
Financial Connections: added extra safeguards in Partner Auth to prevent double completing

kgaidis-stripe:
## Summary

I was doing some error investigation and my conclusion was that there's a chance that Partner Auth could complete twice. This code adds a wrapper to ensure this does not complete. It doesn't fix the root cause, _but_ its still good to do because this is not the first time this has happened. [Click here for code on the first time this happened](https://github.com/stripe/stripe-ios/blob/04f73fe7/StripeFinancialConnections/StripeFinancialConnections/Source/Native/PartnerAuth/PartnerAuthViewController.swift#L368-L372). To prevent all future cases of this potential issue, we add this extra safeguard code.

Note that "double completing" likely causes auth flow to complete successfully anyway, but user could see misleading / confusing UX.

## Testing

Run `bitrise run financial-connections-stability-tests`:

```
FinancialConnectionsNetworkingUITests
    ✓ testNativeNetworkingTestMode (79.067 seconds)
FinancialConnectionsUITests
    ✓ testDataLiveModeOAuthNativeAuthFlow (35.001 seconds)
    ✓ testDataLiveModeOAuthWebAuthFlow (24.894 seconds)
    ✓ testDataTestModeOAuthNativeAuthFlow (26.262 seconds)
    ✓ testPaymentTestModeLegacyNativeAuthFlow (25.618 seconds)
    ✓ testPaymentTestModeManualEntryNativeAuthFlow (29.783 seconds)
    ✓ testSearchInLiveModeNativeAuthFlow (32.613 seconds)

Executed 7 tests, with 0 failures (0 unexpected) in 253.237 (253.243) seconds
```

I also tested:
1) on simulator testing OAuth and legacy flows via test mode (cancelling them, and then completing them regardless)
2) live mode app to app on device
3) live mode bank of america

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/Native/PartnerAuth/PartnerAuthViewController.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/Native/PartnerAuth/PartnerAuthViewController.swift
@@ -212,7 +212,7 @@ final class PartnerAuthViewController: SheetViewController {
                 self.authorizeAuthSession(authSession)
             } else {
                 // for legacy flows (non-OAuth), we do not need to fetch OAuth results, or call authorize
-                self.delegate?.partnerAuthViewController(self, didCompleteWithAuthSession: authSession)
+                self.didComplete(withAuthSession: authSession)
             }
         } else if status == "failure" {
             self.dataSource.recordAuthSessionEvent(
@@ -495,7 +495,7 @@ final class PartnerAuthViewController: SheetViewController {
                 guard let self = self else { return }
                 switch result {
                 case .success(let authSession):
-                    self.delegate?.partnerAuthViewController(self, didCompleteWithAuthSession: authSession)
+                    self.didComplete(withAuthSession: authSession)
 
                     // hide the loading view after a delay to prevent
                     // the screen from flashing _while_ the transition
@@ -600,10 +600,7 @@ final class PartnerAuthViewController: SheetViewController {
                             authSessionId: authSession.id
                         )
                         // abstract auth handles calling `authorize`
-                        self.delegate?.partnerAuthViewController(
-                            self,
-                            didCompleteWithAuthSession: authSession
-                        )
+                        self.didComplete(withAuthSession: authSession)
                     } else {
                         completionHandler(false)
                     }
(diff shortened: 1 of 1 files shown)


StripeFinancialConnections/StripeFinancialConnections/Source/Native/PartnerAuth/PartnerAuthViewController.swift
```diff
@@ -637,6 +634,28 @@ final class PartnerAuthViewController: SheetViewController {
...
+    // Defensive programming to avoid completing the same auth session twice.
+    //
+    // There's been a series of odd edge-cases where the same auth session
+    // could complete twice, so this acts as future-proofing that this
+    // never happens again.
+    private var lastCompletedAuthSessionId: String?
+    private func didComplete(withAuthSession authSession: FinancialConnectionsAuthSession) {
+        if lastCompletedAuthSessionId != authSession.id {
+            lastCompletedAuthSessionId = authSession.id
+            delegate?.partnerAuthViewController(self, didCompleteWithAuthSession: authSession)
```
kgaidis-stripe, line 647: this is very important to double check: this code should ensure that `didCompleteWithAuthSession` DOES get called but only once per an `authSession` 

_(a user can have multiple auth sessions per one "link account session" [or a session through the financial connections auth flow]...all they have to do is cancel their current browser session or go back to institution/bank picker screen)_

StripeFinancialConnections/StripeFinancialConnections/Source/Native/PartnerAuth/PartnerAuthViewController.swift
```diff
@@ -637,6 +634,28 @@ final class PartnerAuthViewController: SheetViewController {
                 pane: .partnerAuth
             )
     }
+
+    // Defensive programming to avoid completing the same auth session twice.
+    //
+    // There's been a series of odd edge-cases where the same auth session
+    // could complete twice, so this acts as future-proofing that this
+    // never happens again.
+    private var lastCompletedAuthSessionId: String?
```
kgaidis-stripe, line 643: not proud of "patches" like this but this pane has had a history of mysterious edge-cases so this safe guard is a good idea

StripeFinancialConnections/StripeFinancialConnections/Source/Native/PartnerAuth/PartnerAuthViewController.swift
```diff
@@ -637,6 +634,28 @@ final class PartnerAuthViewController: SheetViewController {
...
+    private func didComplete(withAuthSession authSession: FinancialConnectionsAuthSession) {
+        if lastCompletedAuthSessionId != authSession.id {
+            lastCompletedAuthSessionId = authSession.id
+            delegate?.partnerAuthViewController(self, didCompleteWithAuthSession: authSession)
+        } else {
+            dataSource
+                .analyticsClient.log(
+                    eventName: "ios_double_complete_attempt",
```
kgaidis-stripe, line 651: log will help future investigations into edge-cases like this

bric-stripe (approved):

PR #3460, opened 2024-04-01
Send Unexpected PaymentSheet Error in FormFactory

yuki-stripe:
## Summary
- Sends an error analytic when PaymentSheet.FormFactory fails to create a form spec
- Adds an extremely rudimentary way to test stpasserts

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-1858
https://docs.google.com/document/d/1r_DS7mDLlrwThcvX8CBUqUBRVejeVDy05Ejth8gnBGI/edit?pli=1

## Testing
See unit test

## Changelog
Not user facing

### DIFF
--- a/StripeCore/StripeCore/Source/Helpers/STPAssert.swift
+++ b/StripeCore/StripeCore/Source/Helpers/STPAssert.swift
@@ -7,10 +7,23 @@
 
 import Foundation
 
+/// A very barebones way to test stpasserts in XCTest.
+@_spi(STP) public class STPAssertTestUtil {
+    /// If set to `true` in an XCTest, the next assertion that fires populates `_testExpectAssertMessage` instead of crashing and resets this flag to `false`.
+    public static var shouldSuppressNextSTPAlert: Bool = false
+    /// The message of the assertion that fired when `_testExpectAssert` was `true`.
+    public static var lastAssertMessage: String = ""
+}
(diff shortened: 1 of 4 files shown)


amk-stripe (dismissed):

sfriedman-stripe:
I've had [this PR](https://github.com/stripe/stripe-ios/pull/2517) open for truly eons, but it has a good pattern for testing assertions using the `CwlPreconditionTesting` library if you'd rather go that route!

yuki-stripe:
> I've had [this PR](https://github.com/stripe/stripe-ios/pull/2517) open for truly eons, but it has a good pattern for testing assertions using the `CwlPreconditionTesting` library if you'd rather go that route!

Oh yeah, we should totally do that. I'll follow up in another PR.

StripeCore/StripeCore/Source/Helpers/STPAssert.swift
```diff
@@ -7,10 +7,23 @@
 
 import Foundation
 
+/// A very barebones way to test stpasserts in XCTest.
+@_spi(STP) public class STPAssertTestUtil {
```
davidme-stripe, line 11: Should we only compile this into debug builds?

wooj-stripe (approved):

PR #3465, opened 2024-04-02
AddPaymentMethodViewController assert->events

wooj-stripe:
## Summary
AddPaymentMethodViewController asserts to events

## Motivation
asserts->events

## Testing
none


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/AddPaymentMethodViewController.swift
```diff
@@ -27,6 +27,12 @@ enum OverrideableBuyButtonBehavior {
 /// For internal SDK use only
 @objc(STP_Internal_AddPaymentMethodViewController)
 class AddPaymentMethodViewController: UIViewController {
+    enum AddPaymentMethodViewControllerError: Error {
```
yuki-stripe, line 30: I'm wondering if "AddPaymentMethodViewControllerError" is unnecessary, since this is already nested. Like, the fully qualified type name that gets sent in analytics will be `AddPaymentMethodViewController. AddPaymentMethodViewControllerError`.  IMO it should either be just a nested `enum Error: Swift.Error` or put the error outside the class and keep the long name.

StripeCore/StripeCore/Source/Analytics/STPAnalyticEvent.swift
```diff
@@ -234,6 +234,7 @@ import Foundation
     // MARK: - Unexpected errors
     // These errors should _never happen_ and indicate a problem with the SDK or the Stripe backend.
     case unexpectedPaymentSheetFormFactoryError = "unexpected_error.paymentsheet.formfactory"
+    case unexpectedAddPaymentMethodViewControllerError = "unexpected_error.paymentsheet.addpaymentmethodviewcontroller"
```
yuki-stripe, line 237: I wonder if we should coalesce all these to a single "unexpectedPaymentSheetError" or something for these, since the error names end up containing the same context?  e.g. right now we're sending:
- event: "unexpected_error.paymentsheet.addpaymentmethodviewcontroller"
- error_type: "AddPaymentMethodViewController. AddPaymentMethodViewControllerError"

wooj-stripe: Sounds good. Lets introduce a more generic form for paymentsheet views/viewcontrollers
```
    case unexpectedPaymentSheetError = "unexpected_error.paymentsheet"
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/AddPaymentMethodViewController.swift
```diff
@@ -27,6 +27,12 @@ enum OverrideableBuyButtonBehavior {
 /// For internal SDK use only
 @objc(STP_Internal_AddPaymentMethodViewController)
 class AddPaymentMethodViewController: UIViewController {
+    enum Error: Swift.Error {
+        case paymentMethodTypesEmpty
+        case usBankAccountFormElementWrongElementCreated
+        case usBankAccountCollectParams
```
yuki-stripe, line 33:
```suggestion
        case usBankAccountParamsMissing
```

nit?

yuki-stripe (approved):

yuki-stripe (approved):

PR #3474, opened 2024-04-03
PMTCollection & PaymentSheetUIKitAdditions events

wooj-stripe:
## Summary
Adding events to files

## Motivation
auditing code

## Testing
None

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/PayWithLinkController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/PayWithLinkController.swift
@@ -13,7 +13,9 @@ import UIKit
 
 /// Standalone Link controller
 final class PayWithLinkController {
-
+    enum Error: Swift.Error {
+        case noWindowWithViewController
+    }
     typealias CompletionBlock = ((PaymentSheetResult, STPAnalyticsClient.DeferredIntentConfirmationType?) -> Void)
 
     private let paymentHandler: STPPaymentHandler
@@ -36,7 +38,10 @@ final class PayWithLinkController {
             let keyWindow = UIApplication.shared.stp_hackilyFumbleAroundUntilYouFindAKeyWindow(),
             let presentedViewController = keyWindow.findTopMostPresentedViewController()
         else {
-            assertionFailure("No key window with view controller found")
+            let errorAnalytic = ErrorAnalytic(event: .unexpectedPaymentSheetError,
+                                              error: Error.noWindowWithViewController)
+            STPAnalyticsClient.sharedClient.log(analytic: errorAnalytic)
+            stpAssertionFailure("No key window with view controller found")
             return
         }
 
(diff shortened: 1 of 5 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/PayWithLinkController.swift
```diff
@@ -36,7 +38,10 @@ final class PayWithLinkController {
             let keyWindow = UIApplication.shared.stp_hackilyFumbleAroundUntilYouFindAKeyWindow(),
             let presentedViewController = keyWindow.findTopMostPresentedViewController()
         else {
-            assertionFailure("No key window with view controller found")
+            let errorAnalytic = ErrorAnalytic(event: .unexpectedPaymentSheetError,
+                                              error: Error.noWindowWithViewController)
+            STPAnalyticsClient.sharedClient.log(analytic: errorAnalytic)
+            stpAssertionFailure("No key window with view controller found")
             return
```
yuki-stripe, line 45: It looks like the completion block isn't called here. That seems bad!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/PayWithLinkController.swift
```diff
@@ -36,7 +38,11 @@ final class PayWithLinkController {
             let keyWindow = UIApplication.shared.stp_hackilyFumbleAroundUntilYouFindAKeyWindow(),
             let presentedViewController = keyWindow.findTopMostPresentedViewController()
         else {
-            assertionFailure("No key window with view controller found")
+            let errorAnalytic = ErrorAnalytic(event: .unexpectedPaymentSheetError,
+                                              error: Error.noWindowWithViewController)
+            STPAnalyticsClient.sharedClient.log(analytic: errorAnalytic)
+            stpAssertionFailure("No key window with view controller found")
+            completion(.failed(error: PaymentSheetError.unableToPresent), nil)
             return
```
yuki-stripe, line 46: Why do we use this method?  PaymentSheet always has a view controller to present on.  Is it difficult to refactor PaymentSheet to use `present(from:completion:)` below, instead, and delete this method + the error code? It looks like we can just pass authenticationContext.presentingViewController

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/PaymentSheetUIKitAdditions.swift
```diff
@@ -60,7 +63,18 @@ extension UIViewController {
     func switchContentIfNecessary(
         to toVC: UIViewController, containerView: DynamicHeightContainerView
     ) {
-        assert(children.count <= 1)
+        if children.count > 1 {
+            let from_vc_name = NSStringFromClass(children.first!.classForCoder)
+            let to_vc_name = NSStringFromClass(toVC.classForCoder)
```
yuki-stripe, line 68: just curious, is there a diff. between this and like

yuki-stripe (approved):

PR #3480, opened 2024-04-04
Cleanup usbankaccount construction in addpaymentmethodvc

wooj-stripe:
## Summary
Cleans up logic to avoid having to have another two events

## Motivation
Offline chat, @yuki-stripe pointed out that the code could be accomplished in a cleaner manner

## Testing
Relying on existing tests

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/AddPaymentMethodViewController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/AddPaymentMethodViewController.swift
@@ -114,6 +114,8 @@ class AddPaymentMethodViewController: UIViewController {
     private let intent: Intent
     private let configuration: PaymentSheet.Configuration
     var previousCustomerInput: IntentConfirmParams?
+
+    // We are keeping usBankAccountInfo in memory to preserve state if the user switches payment method types
     private var usBankAccountFormElement: USBankAccountPaymentMethodElement?
     private lazy var paymentMethodFormElement: PaymentMethodElement = {
         if selectedPaymentMethodType == .stripe(.USBankAccount) {
@@ -123,7 +125,7 @@ class AddPaymentMethodViewController: UIViewController {
             } else {
                 // Cache the form
                 let element = makeElement(for: .stripe(.USBankAccount))
-                usBankAccountFormElement = (element as! USBankAccountPaymentMethodElement)
+                usBankAccountFormElement = element as? USBankAccountPaymentMethodElement
                 return element
             }
         }
@@ -307,7 +309,7 @@ class AddPaymentMethodViewController: UIViewController {
             } else {
                 // Cache the form
                 paymentMethodFormElement = makeElement(for: .stripe(.USBankAccount))
-                usBankAccountFormElement = (paymentMethodFormElement as! USBankAccountPaymentMethodElement)
+                usBankAccountFormElement = paymentMethodFormElement as? USBankAccountPaymentMethodElement
             }
         } else {
             paymentMethodFormElement = makeElement(for: selectedPaymentMethodType)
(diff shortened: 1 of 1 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/AddPaymentMethodViewController.swift
```diff
@@ -120,25 +114,18 @@ class AddPaymentMethodViewController: UIViewController {
...
-            usBankAccountPaymentMethodElement.presentingViewControllerDelegate = self
-        } else {
-            let errorAnalytic = ErrorAnalytic(event: .unexpectedPaymentSheetError,
-                                              error: Error.usBankAccountFormElementWrongElementCreated)
-            STPAnalyticsClient.sharedClient.log(analytic: errorAnalytic)
-            stpAssertionFailure("Invalid type of paymentMethodElement")
-        }
-        return paymentMethodElement as? USBankAccountPaymentMethodElement
-    }()
+    private var usBankAccountFormElement: USBankAccountPaymentMethodElement?
```
yuki-stripe, line 117: Let's keep this comment
```
 // We are keeping usBankAccountInfo in memory to preserve state if the user switches payment method types
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/AddPaymentMethodViewController.swift
```diff
@@ -41,11 +39,7 @@ class AddPaymentMethodViewController: UIViewController {
             configuration: configuration,
             logAvailability: false
         )
-        if paymentMethodTypes.isEmpty {
-            let errorAnalytic = ErrorAnalytic(event: .unexpectedPaymentSheetError,
-                                              error: Error.paymentMethodTypesEmpty)
-            STPAnalyticsClient.sharedClient.log(analytic: errorAnalytic)
-        }
+
```
yuki-stripe, line 42: how come we got rid of this?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/AddPaymentMethodViewController.swift
```diff
@@ -313,10 +300,15 @@ class AddPaymentMethodViewController: UIViewController {
...
-        {
-            paymentMethodFormElement = usBankAccountFormElement
+        if selectedPaymentMethodType == .stripe(.USBankAccount) {
+            if let usBankAccountFormElement {
+                // Use the cached form instead of creating a new one
+                paymentMethodFormElement = usBankAccountFormElement
+            } else {
+                // Cache the form
+                paymentMethodFormElement = makeElement(for: .stripe(.USBankAccount))
+                usBankAccountFormElement = (paymentMethodFormElement as! USBankAccountPaymentMethodElement)
```
yuki-stripe, line 310: I'm realizing that theoretically this _could_ crash.  So maybe it should just be `as?` instead, and not send an error here b/c it's FormFactory's responsibility?

wooj-stripe: Changed it to `as?` as proposed. As for the metric, agree we don't need to metric here for the reason you mentioned, and because `handleCollectBankAccount()` will fire it's .unexpectedPaymentSheetError if it is not returning a `USBankAccountPaymentMethodElement`

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/AddPaymentMethodViewController.swift
wooj-stripe: This is being removed because there is an assert on line 44.  I believe this is more or less a critical error, so we should break loudly.

yuki-stripe (approved):

PR #3482, opened 2024-04-05
Load degrade errors

porter-stripe:
## Summary
- Adds analytic events for code paths that can occur but don't block loading
- These can all occur within PaymentSheet or CustomerSheet so I defaulted to using a generic unexpected error for each instance.

## Motivation
Current sprint

## Testing
N/A

## Changelog
N/A


StripeCore/StripeCore/Source/DownloadManager/DownloadManager.swift
```diff
@@ -94,6 +99,10 @@ extension DownloadManager {
                 let data = self.getDataFromURL(tempURL),
                 let image = self.persistToMemory(data, forImageName: imageName)
             else {
+                let errorAnalytic = ErrorAnalytic(event: .unexpectedStripeCoreDownloadManagerError,
+                                                  error: InternalError.downloadSyncFailure,
```
wooj-stripe, line 103: We've been erring on the side creating less events when the name of the error is specific.  So, in this case, we are sending:
```
event: unexpected_error.stripecore.downloadmanager
error: DownloadManager.InternalError.downloadSyncFailure
```

For this case, I think it makes sense to:
1.  Change `InternalError` to `Error` ( e.g.: `private enum Error: Swift.Error`)
2. For the event use `unexpectedPaymentSheetError`

porter-stripe: 1, What's the benefit of using `private enum Error: Swift.Error` over `private enum InternalError: Error`?
2. This specific error lives outside of PaymentSheet so I don't think `unexpectedPaymentSheetError` makes sense. However I can update the others that do to use `unexpectedPaymentSheetError`.

wooj-stripe: 1. I suspect `InternalError` was used in the case you linked, because the cost to change it over to `Error`, and converting every existing place from Error to Swift.Error was too high.  To be honest, this comment was more or less a nit, so seems fine either way.
2. Ah! Right, it is part of stripe core. whoops!  Sounds good.

porter-stripe: I updated to `private enum Error: Swift.Error` for consistency.

PR #3487, opened 2024-04-05
Refactor `DownloadManager` and add analytics

porter-stripe:
### DownloadManager refactor
- Removed conformance to `URLSessionDelegate`, it was never implemented or used.
- Removed various variables relating to storing and keeping track of pending requests. Usage of Swift concurrency takes care of this since the functions "look" like they are executing synchronously. Some of these such as the `downloadQueue` and `downloadOperationQueue` were never used.
- Removed `STPCacheExpirationInterval`, it was never used. Instead the `URLCache` uses the [`Cache-Control`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control) HTTP header to determine when an item will expire.
- Created a common function `downloadImage` which uses Swift async to download images and catch errors in a single location
- Uses `data` instead of `downloadTask` on the `URLSession` as this is a small operation that does not need to occur in the background, and allows us to not worry about converting the response to data which could result in an additional error we need to handle.
- Refactored the in-memory cache to not use GCD and instead use Actor pattern.
- The public API of `DownloadManager` is not modified/refactored to avoid refactoring all usage of the class, kept the refactor internal. We should consider removing the blocking image download and update the usages.
- URL Cache is thread safe: https://developer.apple.com/documentation/foundation/urlcache#2317483

### Analytics changes
- If we get an error back from the network we log that error so that `error_type` = `NSURLErrorDomain` and `error_code` is the error code from the response.
- If we fail to convert the data into an image we send `error_type` = `StripeCore.DownloadManager.Error` and `error_code` = `failedToMakeImageFromData`

## Motivation
Feedback

## Testing
- Manual in PaymentSheet playground
- Existing tests ran 1k times each
- Few new tests

## Changelog
N/A


StripeCore/StripeCore/Source/DownloadManager/DownloadManager.swift
```diff
@@ -93,15 +93,16 @@ extension DownloadManager {
...
                 let data = self.getDataFromURL(tempURL),
                 let image = self.persistToMemory(data, forImageName: imageName)
             else {
-                let errorAnalytic = ErrorAnalytic(event: .unexpectedStripeCoreDownloadManagerError,
-                                                  error: Error.downloadSyncFailure,
-                                                  additionalNonPIIParams: ["url": url.absoluteString])
+                let errorAnalytic = ErrorAnalytic(event: .stripeCoreDownloadManagerError,
+                                                  error: Error.downloadAsyncFailure,
+                                                  additionalNonPIIParams: ["url": url.absoluteString,
+                                                                           "url_error_code": (error as? NSError)?.code ?? "none",])
```
porter-stripe, line 105: We will create an alert which always alerts on a "none" "url_error_code", as that means we got a valid response but failed to create an image from it.

StripeCore/StripeCore/Source/DownloadManager/DownloadManager.swift
```diff
@@ -93,19 +93,34 @@ extension DownloadManager {
...
-        let task = self.session.downloadTask(with: url) { tempURL, response, _ in
-            guard let tempURL = tempURL,
-                let response = response,
-                let data = self.getDataFromURL(tempURL),
-                let image = self.persistToMemory(data, forImageName: imageName)
-            else {
-                let errorAnalytic = ErrorAnalytic(event: .unexpectedStripeCoreDownloadManagerError,
-                                                  error: Error.downloadSyncFailure,
+        let task = self.session.downloadTask(with: url) { tempURL, response, responseError in
+            if let responseError = responseError as? NSError {
```
yuki-stripe, line 97: Why cast `as? NSerror`?

StripeCore/StripeCore/Source/Analytics/STPAnalyticEvent.swift
```diff
@@ -236,9 +236,11 @@ import Foundation
...
     case unexpectedStripeUICoreAddressSpecProvider = "unexpected_error.stripeuicore.addressspecprovider"
     case unexpectedStripeUICoreBSBNumberProvider = "unexpected_error.stripeuicore.bsbnumberprovider"
-    case unexpectedStripeCoreDownloadManagerError = "unexpected_error.stripecore.downloadmanager"
     case unexpectedPaymentSheetError = "unexpected_error.paymentsheet"
     case unexpectedPaymentSheetViewControllerError = "unexpected_error.paymentsheet.paymentsheetviewcontroller"
     case unexpectedFlowControllerViewControllerError = "unexpected_error.paymentsheet.flowcontrollerviewcontroller"
     case unexpectedPaymentHandlerError = "unexpected_error.paymenthandler"
+
+    // MARK: - Misc. errors
+    case stripeCoreDownloadManagerError = "stripecore.downloadmanager"
```
yuki-stripe, line 245:
```suggestion
    case stripeCoreDownloadManagerError = "stripecore.downloadmanager.error"
```

StripeCore/StripeCore/Source/DownloadManager/DownloadManager.swift
```diff
@@ -93,19 +93,34 @@ extension DownloadManager {
...
+                blockingDownloadSemaphore.signal()
+                let errorAnalytic = ErrorAnalytic(event: .stripeCoreDownloadManagerError,
+                                                  error: responseError,
+                                                  additionalNonPIIParams: ["url": url.absoluteString])
+                STPAnalyticsClient.sharedClient.log(analytic: errorAnalytic)
+                return
+            }
+
+            guard let tempURL = tempURL, let response = response, let data = self.getDataFromURL(tempURL) else {
```
yuki-stripe, line 106:
```suggestion
            guard let tempURL, let response, let data = self.getDataFromURL(tempURL) else {
```

yuki-stripe: We should really use the modern `async throw` APIs so that we're guaranteed a tempURL and response if there's no error, among other things.

ccen-stripe (dismissed):

StripeCore/StripeCore/Source/DownloadManager/DownloadManager.swift
```diff
@@ -93,19 +93,34 @@ extension DownloadManager {
...
+                blockingDownloadSemaphore.signal()
+                let errorAnalytic = ErrorAnalytic(event: .stripeCoreDownloadManagerError,
+                                                  error: Error.failedToMakeDataFromResponse,
```
yuki-stripe, line 109: The implementation of `getDataFromURL` hides its error from the caller, ideally we are able to report the error.

StripeCore/StripeCore/Source/DownloadManager/DownloadManager.swift
yuki-stripe: Also why are we using a downloadTask instead of dataTask for this? These images are tiny and we're not downloading them in the background, any reason we shouldn't switch to dataTask and just get the data directly instead of having to read the data from disk? cc @wooj-stripe

StripeCore/StripeCore/Source/DownloadManager/DownloadManager.swift
```diff
@@ -125,18 +140,34 @@ extension DownloadManager {
...
-                let response = response,
-                let data = self.getDataFromURL(tempURL),
-                let image = self.persistToMemory(data, forImageName: imageName)
-            else {
-                self?.pendingRequestsSemaphore.wait()
-                self?.pendingRequests.removeValue(forKey: imageName)
-                self?.pendingRequestsSemaphore.signal()
-                let errorAnalytic = ErrorAnalytic(event: .unexpectedStripeCoreDownloadManagerError,
-                                                  error: Error.downloadAsyncFailure,
+        let task = self.session.downloadTask(with: url) { [weak self] tempURL, response, responseError in
```
yuki-stripe, line 143: It feels like we could (not in this PR) rewrite this class to implement this method to simply call the other one or vice versa.

StripeCore/StripeCore/Source/Analytics/STPAnalyticEvent.swift
worldofnick: Accepted suggestion.

StripeCore/StripeCore/Source/DownloadManager/DownloadManager.swift
worldofnick: Not needed, removed.

porter-stripe: Not needed revmoed.

StripeCore/StripeCore/Source/Analytics/STPAnalyticEvent.swift
porter-stripe: Accepted suggestion.

StripeCore/StripeCore/Source/DownloadManager/DownloadManager.swift
porter-stripe: Good point, after refactoring to use `data` on the URLSession we could remove the function entirely.

StripeCore/StripeCore/Source/DownloadManager/DownloadManager.swift
porter-stripe: Refactored this class to use async functions and updated to use a data task.

StripeCore/StripeCore/Source/DownloadManager/DownloadManager.swift
```diff
@@ -53,177 +41,114 @@ import UIKit
...
 
-        self.pendingRequestsSemaphore.wait()
-        guard self.pendingRequests[imageName] == nil else {
-            addUpdateHandlerWithoutLocking(updateHandler, forImageName: imageName)
-            self.pendingRequestsSemaphore.signal()
+        let urlRequest = URLRequest(url: url, cachePolicy: .useProtocolCachePolicy)
+        do {
+            let (data, response) = try await session.data(for: urlRequest)
+            let image = try persistToMemory(data, forImageName: imageName) // Throws a Error.failedToMakeImageFromData
+            diskCache?.storeCachedResponse(CachedURLResponse(response: response, data: data), for: urlRequest)
```
yuki-stripe, line 90: Why do we manually handle the cache?  And if we do, why set it on urlSession.configuration?  Seems like it should be one or the other.

StripeCore/StripeCore/Source/DownloadManager/DownloadManager.swift
```diff
@@ -9,37 +9,25 @@ import UIKit
...
-    var updateHandlers: [String: [UpdateImageHandler]]
-
-    let imageCacheSemaphore: DispatchSemaphore
-    let pendingRequestsSemaphore: DispatchSemaphore
-
-    let STPCacheExpirationInterval = (60 * 60 * 24 * 7)  // 1 week
-    var urlCache: URLCache?
+    private let session: URLSession
+    private let analyticsClient: STPAnalyticsClient
+    private var diskCache: URLCache?
```
yuki-stripe, line 25: nit: This can be a let

StripeCore/StripeCore/Source/DownloadManager/DownloadManager.swift
```diff
@@ -9,37 +9,25 @@ import UIKit
...
+    enum Error: Swift.Error {
+        case failedToMakeImageFromData
     }
 
     public static let sharedManager = DownloadManager()
 
-    let downloadQueue: DispatchQueue
-    let downloadOperationQueue: OperationQueue
-    let session: URLSession!
+    let imageCache = ImageCache()
```
yuki-stripe, line 21: Why have this cache and URLCache?

StripeCore/StripeCore/Source/DownloadManager/DownloadManager.swift
```diff
@@ -9,37 +9,25 @@ import UIKit
...
-        urlSessionConfiguration: URLSessionConfiguration = .default
+        urlSessionConfiguration: URLSessionConfiguration = .default,
+        analyticsClient: STPAnalyticsClient = .sharedClient
     ) {
-        downloadQueue = DispatchQueue(label: "Stripe Download Cache", attributes: .concurrent)
-        downloadOperationQueue = OperationQueue()
-        downloadOperationQueue.underlyingQueue = downloadQueue
-
         let configuration = urlSessionConfiguration
         if let cachesURL = FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask)
```
yuki-stripe, line 32: Why create a custom URL instead of using the default?

StripeCore/StripeCore/Source/DownloadManager/DownloadManager.swift
```diff
@@ -53,177 +41,114 @@ import UIKit
...
-        if updateHandler == nil {
-            return downloadImageBlocking(placeholder: placeholder, url: url)
-        } else {
-            return downloadImageAsync(url: url, placeholder: placeholder, updateHandler: updateHandler)
-        }
-    }
 
-    func downloadImageBlocking(placeholder: UIImage?, url: URL) -> UIImage {
+    /// Downloads an image from a provided URL, using either a synchronous method or an asynchronous method.
+    /// If no `updateHandler` is provided, this function will block the current thread until the image is downloaded. If an `updateHandler` is provided, the function does not wait for the download to finish and returns a placeholder image immediately instead. When the image finishes downloading, the `updateHandler` will be called with the downloaded image.
```
yuki-stripe, line 57: Why not have separate async and sync methods?  Having a method be async/sync based on parameters feels surprising and complicated.

porter-stripe: Agree this is not optimal, but main goal of this refactor was to not modify the public interface and just refactor the internals. I think we should refactor usage of this to not need a blocking image download, but I'm proposing we keep the public contract of this class unmodified for this PR. Ideally we can just remove the sync usage later. I can log a JIRA if you agree!

StripeCore/StripeCore/Source/DownloadManager/DownloadManager.swift
porter-stripe: I don't see a great reason why we need a custom URL here, but given that is what we already have I think it makes sense to keep it.
1. I can't see any drawbacks to using a custom URL
2. If we change it and a merchant updates the SDK they'd have this existing cache directory, but then we'd be putting stuff in the default URL which seems odd.
3. Merchant loses existing cache on SDK update

StripeCore/StripeCore/Source/DownloadManager/DownloadManager.swift
porter-stripe: Good question! I incorrectly assumed this class previously had a in-memory cache and disk cache, with the `URLCache` being the disk cache. But turns out `URLCache` is also a in-memory cache so the `imageCache` was redundant. This should have been more obvious b/c we are passing in a `memoryCapacity` to the `URLCache`. 

> The URLCache class implements the caching of responses to URL load requests, by mapping [NSURLRequest](https://developer.apple.com/documentation/foundation/nsurlrequest) objects to [CachedURLResponse](https://developer.apple.com/documentation/foundation/cachedurlresponse) objects. It provides a composite in-memory and on-disk cache, and lets you manipulate the sizes of both the in-memory and on-disk portions.

https://developer.apple.com/documentation/foundation/urlcache

I removed `ImageCache` which also means we need to remove our cache specific tests since testing `URLCache` w/ OHHTTPStubs isn't supported, but I think that is ok given that we shouldn't really write tests for`URLCache`. I did add a test to ensure we are configuring the cache properly and attaching it to `URLSessionConfiguration`.

StripeCore/StripeCore/Source/DownloadManager/DownloadManager.swift
porter-stripe: Based on this [comment](https://github.com/stripe/stripe-ios/pull/3487#discussion_r1559955101), this ended up being removed!

StripeCore/StripeCore/Source/DownloadManager/DownloadManager.swift
yuki-stripe: Maybe one drawback is that creating the URL could fail, in which case we'd silently lose caching.  I don't have a sense for how possible that is. I'm fine to keep since it's existing behavior - maybe it makes sense to have it separate from the default cache so we can delete it if we want.

StripeCore/StripeCore/Source/DownloadManager/DownloadManager.swift
yuki-stripe: That's reasonable. This whole interface is very complicated. Like, besides async/sync, the return behavior is totally different based on the presence of `updateHandler` etc. It makes it really hard to tell if this PR keeps the old behavior. If you're confident in the correctness though, I'm okay to ship what we have and ticket up to fix this. I think breaking this out into multiple methods could make this a lot simpler.

StripeCore/StripeCore/Source/DownloadManager/DownloadManager.swift
```diff
@@ -53,193 +38,127 @@ import UIKit
...
-        }
-        if let blocks = self.updateHandlers[imageName] {
-            self.updateHandlers[imageName] = blocks + [handler]
-        } else {
-            self.updateHandlers[imageName] = [handler]
-        }
+    private func downloadImageBlocking(url: URL, placeholder: UIImage) -> UIImage {
+        return _unsafeWait({
+            return await self.downloadImage(url: url, placeholder: placeholder)
+        }) ?? imagePlaceHolder()
```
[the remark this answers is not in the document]
porter-stripe: Good flag, `Box` was a very generic workaround for running Swift async code sync, probably too generic for our use case. I made `Box` less generic and fixed this weirdness!

wooj-stripe (approved):

PR #3490, opened 2024-04-05
Fix size test

davidme-stripe:
## Summary
* Only set LD_GENERATE_MAP_FILE in the app target — if we add it to the xcodebuild command, it will also generate a file at the same path for each framework.
* Set LD_GENERATE_MAP_FILE in the SPMTest.xcodeproj project, and regenerate the patch.
* Use Xcode 15.3
* Always checkout from the `measure_branch`: Because of how Bitrise checks out branches in recent stacks,  `git rev-parse --abbrev-ref HEAD` was just returning `HEAD`, which was causing us to then re-checkout `HEAD` after checking out `master`, which triggered it to use an *old* version of the Xcodeproj patch. This was confusing and caused it to not pick up changes from this PR.

## Motivation
Size_test was broken in Xcode 15

## Testing
CI


davidme-stripe:
^ The above missing base ref is expected, as we haven't sent Emerge Tools a build in a while.

Tests/installation_tests/size_test/size_report.rb
```diff
@@ -107,32 +107,34 @@ def build(dir, target_name = 'SPMTest')
...
+      '-sdk "iphoneos" ' +
+      '-destination "generic/platform=iOS" ' +
+      '-archivePath build/SPMTest.xcarchive ' +
+      'CODE_SIGN_IDENTITY="-" ' +
+      'CODE_SIGNING_REQUIRED="NO" ' +
+      'CODE_SIGN_ENTITLEMENTS="" ' +
+      "LD_MAP_FILE_PATH=\"#{dir}/build/#{target_name}-LinkMap.txt\" " +
+      'CODE_SIGNING_ALLOWED="NO"').to_s
+      command_succeeded = system(xcode_command)
+      puts xcode_command
```
wooj-stripe, line 123: you probably want `puts` before `command_succeeded = system(xcode_command)`

wooj-stripe (approved):

PR #3492, opened 2024-04-05
Add GitHub workflow to detect new public APIs

porter-stripe:
## Summary
- Adds a script to prevent shipping public APIs
- Excludes test files and directories
- Excludes Identity, Connections, and FC.

## Motivation
Prevent shipping public APIs on accident

## Testing
See commit history

## Changelog
N/A


.github/workflows/check-public-apis.yml
```diff
@@ -0,0 +1,76 @@
...
+on:
+  pull_request:
+    types: [opened, labeled, unlabeled, synchronize]
+    paths:
+      - '**/*.swift'
+      - '!StripeFinancialConnections/**'
+      - '!StripeIdentity/**'
+      - '!StripeConnections/**'
+    branches-ignore:
+      - 'releases/*' # ignore release PRs due to them modifying StripeAPIConfiguration.STPSDKVersion
```
wooj-stripe, line 11: Do we need to ignore:
```
docs*
```
?

.github/workflows/check-public-apis.yml
```diff
@@ -0,0 +1,76 @@
...
+      - name: Check for new public APIs
+        id: check-api
+        run: |
+          # Run the Ruby script and make a list of new public APIs 
+          echo "Checking for new Public APIs..."
+          OUTPUT=$(ruby ci_scripts/check_for_new_public_apis.rb ${{ github.workspace }})
+          echo "${OUTPUT}"
+          echo "API_CHECK_OUTPUT<<EOF" >> $GITHUB_ENV
+          echo "${OUTPUT}" >> $GITHUB_ENV
+          echo "EOF" >> $GITHUB_ENV
```
wooj-stripe, line 39: I guess the only way this could break is if ${OUTPUT} contained an EOF marker, which I think is impossible?

ci_scripts/check_for_new_public_apis.rb
```diff
@@ -0,0 +1,50 @@
...
+  previous_line_spi_annotated = false
+
+
+  File.foreach(file) do |line|
+    line = line.strip
+    opening_braces = line.count('{')
+    closing_braces = line.count('}')
+    
+    block_depth += opening_braces - closing_braces
+    if line =~ /@_spi\(\w+\)\s+(public\s+|open\s+)?(class|struct|enum|protocol)\s+\w+/ && opening_braces > 0
```
wooj-stripe, line 26: Is `&& opening_braces > 0` accounting for import statements?  If so, consider explicitly having a pattern on "import", but if it makes it more complicated, ignore this comment.

.github/workflows/check-public-apis.yml
porter-stripe: Good callout! I didn't consider this case actually, but since the docs PR only modifies `.json` files this workflow shouldn't be trigged since it looks for `.swift` file changes!

.github/workflows/check-public-apis.yml
porter-stripe: ++, I think it's impossible too. We can keep an eye on it though.

porter-stripe: ++, I think it's impossible too. We can keep an eye on it though.

ci_scripts/check_for_new_public_apis.rb
porter-stripe: Hmm, I don't quite understand why we should account for import statements. Could you give an example?

.github/workflows/check-public-apis.yml
porter-stripe: If no swift files are modified the workflow just won't be triggered! See on this PR how the latest check it's not ran since no swift files are currently modified. If you tap into the checks on earlier commits you'll see it being ran.

ci_scripts/check_for_new_public_apis.rb
wooj-stripe: Sorry -- I'm in agreement that we shouldn't account for import statements, I was just trying to understand the code.  It seems that `opening_braces > 0` implies that this statement is false for looking at import statements, correct?  If so, are there any other cases that this `opening_braces > 0` is for?

If not, I was suggesting to update the code to something like:
```
 if line =~ /@_spi\(\w+\)\s+(public\ .... && not_import_statement
```

porter-stripe: @wooj-stripe Yeah, we could have nested types or multiple classes in a file. So, the opening_braces > 0 condition is primarily there to ensure we are within the scope of a Swift type definition (i.e., class, struct, enum, protocol, typealias). We need to be within this scope to correctly identify new items that should be annotated with "@_spi" modifier.

wooj-stripe (approved):

PR #3500, opened 2024-04-09
Add load events for CustomerSheet

porter-stripe:
## Summary
- Log 3 new events for CustomerSheet around load time

## Motivation
Visibility sprint

## Testing
Manual

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSheet.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSheet.swift
@@ -128,7 +128,9 @@ public class CustomerSheet {
                 STPAnalyticsClient.sharedClient.logPaymentSheetEvent(event: .customerSheetLoadSucceeded,
                                                                      duration: Date().timeIntervalSince(loadingStartDate))
             case .failure(let error):
-                STPAnalyticsClient.sharedClient.log(analytic: ErrorAnalytic(event: .customerSheetLoadFailed, error: error))
+                STPAnalyticsClient.sharedClient.logPaymentSheetEvent(event: .customerSheetLoadFailed,
+                                                                     duration: Date().timeIntervalSince(loadingStartDate),
+                                                                     error: error)
                 csCompletion(.error(CustomerSheetError.errorFetchingSavedPaymentMethods(error)))
                 DispatchQueue.main.async {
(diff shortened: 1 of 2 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSheet.swift
```diff
@@ -91,6 +91,7 @@ public class CustomerSheet {
     public func present(from presentingViewController: UIViewController,
                         completion csCompletion: @escaping (CustomerSheetResult) -> Void
     ) {
+        STPAnalyticsClient.sharedClient.logPaymentSheetEvent(event: .customerSheetLoadStarted)
```
porter-stripe, line 94: Interesting to note we use `logPaymentSheetEvent`, but this is already the case for all other CustomerSheet events.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSheet.swift
```diff
@@ -123,7 +125,10 @@ public class CustomerSheet {
                              selectedPaymentMethodOption: selectedPaymentMethodOption,
                              merchantSupportedPaymentMethodTypes: merchantSupportedPaymentMethodTypes,
                              cbcEligible: elementsSession.cardBrandChoice?.eligible ?? false)
+                STPAnalyticsClient.sharedClient.logPaymentSheetEvent(event: .customerSheetLoadSucceeded,
+                                                                     duration: Date().timeIntervalSince(loadingStartDate))
             case .failure(let error):
+                STPAnalyticsClient.sharedClient.log(analytic: ErrorAnalytic(event: .customerSheetLoadFailed, error: error))
```
wooj-stripe, line 131: 1. Should we we be passing duration in for the failure as well?
2. Any reason why we chose to use `.logPaymentSheetEvent` for success but  `.log` for the failure case?

porter-stripe: No real reason to use log over logPaymentSheetEvent! I omitted duration here cause we don't really use it on the PaymentSheet event, but I don't think there is any harm in sending it incase we do end up using it and for consistency. 

Updated!

wooj-stripe (approved):

PR #3501, opened 2024-04-09
Send analytics events on simulator but not during tests

porter-stripe:
## Summary
- Updates the `STPAnalyticsClient` to only send analytics if we are not in a test. Now includes sending when running from a simulator.
- Intentionally does not update the V2 analytics client as that is used outside the scope of our team.
- Left in place the `STPAnalyticsClientDelegate` so our playground can still inspect analytics logs easily.

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-1903

## Testing
Manual

## Changelog
N/A

### DIFF
--- a/StripeCore/StripeCore/Source/Analytics/STPAnalyticsClient.swift
+++ b/StripeCore/StripeCore/Source/Analytics/STPAnalyticsClient.swift
@@ -66,18 +66,6 @@ import UIKit
         additionalInfoSet.removeAll()
     }
 
-    public static var isSimulatorOrTest: Bool {
-        #if targetEnvironment(simulator)
-            return true
-        #else
-            return NSClassFromString("XCTest") != nil
-        #endif
-    }
-
-    @objc public class func shouldCollectAnalytics() -> Bool {
-        return !isSimulatorOrTest
-    }
-
     public func additionalInfo() -> [String] {
         return additionalInfoSet.sorted()
     }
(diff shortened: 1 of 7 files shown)


StripeCore/StripeCore/Source/Analytics/STPAnalyticsClient.swift
```diff
@@ -115,11 +115,9 @@ import UIKit
         delegate?.analyticsClientDidLog(analyticsClient: self, payload: payload)
         #endif
 
+        // Don't send the analytic, but add it to `_testLogHistory` if we're in a test.
         guard type(of: self).shouldCollectAnalytics() else {
-            // Don't send the analytic, but add it to `_testLogHistory` if we're in a test.
-            if NSClassFromString("XCTest") != nil {
-                _testLogHistory.append(payload)
-            }
+            _testLogHistory.append(payload)
```
yuki-stripe, line 120: nit: I don't think it's right to assume that `shouldCollectAnalytics` tells us whether or not we're in a test, so we should keep the explicit XCTest check.

yuki-stripe (changes requested):
We need to make sure analytics correctly report that this is testmode / is_development or something, let's chat in slack

StripeCore/StripeCore/Source/Analytics/STPAnalyticsClient.swift
porter-stripe: Chatted in Slack, confirmed that this doesn't change how we populate testmode or is_developement.
1. testmode is derived from the publishable key
2. is_development is derived from STPAnalyticsClient.isSimulatorOrTest

porter-stripe:
@yuki-stripe I moved `is_development` to `commonPayload()` and verified all removed instances of it being set in the payload eventually flow down to `commonPayload()`.

Additionally since we no longer check for `targetEnvironment(simulator)` when determining to collect analytics we need to now check if we are in a UI test explicitly. 

Regarding the nit comment, I updated the name to make more sense, I don't think we need two checks if we are testing so renaming we preferred.

PR #3508, opened 2024-04-10
Log request id in error analytics

yuki-stripe:
## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-1564

## Testing
See tests

## Changelog
Not user facing

### DIFF
--- a/StripeCore/StripeCore/Source/Analytics/AnalyticLoggableError.swift
+++ b/StripeCore/StripeCore/Source/Analytics/AnalyticLoggableError.swift
@@ -87,6 +87,8 @@ extension AnalyticLoggableError where Self: Error {}
             "error_type": errorType,
             "error_code": errorCode,
         ]
+        params["request_id"] = Self.extractStripeAPIRequestID(from: self)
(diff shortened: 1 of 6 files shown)


StripeCore/StripeCore/Source/Analytics/AnalyticLoggableError.swift
```diff
@@ -139,4 +141,13 @@ extension AnalyticLoggableError where Self: Error {}
             return String((error as NSError).code)
         }
     }
+
+    static func extractStripeAPIRequestID(from error: Error) -> String? {
+        let error = error as NSError
```
porter-stripe, line 146: Is there ever a time we would want to log the request ID on something that is not an error?

yuki-stripe: I can't immediately think of a reason, but possibly!  I feel like I'm only looking at request IDs to investigate problems.

porter-stripe (approved):

PR #3527, opened 2024-04-18
Prioritize user specified networks for Apple Pay

bernardo-stripe:
## Summary

Add `additionalEnabledApplePayNetworks` in front of the default network list at `supportedPKPaymentNetworks()` instead of adding it to the end.

## Motivation

Currently `additionalEnabledApplePayNetworks` that are set by the user are appended to the list of supported networks, having a lower priority than the default networks. 
As Apple Pay prioritizes the networks in order, we would like to make the user's specified networks take precedence over the default networks.

## Testing

Automated tests were added to verify the list is in the correct order.

## Changelog

[Changed] Apple Pay additionalEnabledApplePayNetworks are now in front of the supported network list.

### DIFF
--- a/Stripe/StripeiOSTests/STPApplePayTest.swift
+++ b/Stripe/StripeiOSTests/STPApplePayTest.swift
@@ -30,6 +30,13 @@ class STPApplePaySwiftTest: XCTestCase {
         StripeAPI.additionalEnabledApplePayNetworks = []
     }
 
+    func testAdditionalPaymentNetworksGetPrepended() {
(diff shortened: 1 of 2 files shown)


bernardo-stripe:
Hi @davidme-stripe, I created this separate PR with your suggestion to prioritize the user specified networks on Apple Pay.

Concerning your question on my [previous PR](https://github.com/stripe/stripe-ios/pull/3402):
> My only question is how that would work for a Discover card: Would it then be processed on JCB? I don't think so, just want to make sure.

This would not happen because we would still get a card with Discover brand from Apple Pay. The routing to Discover or JCB network will be done at the backend.

davidme-stripe (approved):

PR #3534, opened 2024-04-22
Refactor formFactory not to use saveMode

wooj-stripe:
## Summary
Refactor formFactory not to use saveMode

## Motivation
Will have the ability to control when the save checkbox is saved

## Testing
Will add additional use cases

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
@@ -22,7 +22,7 @@ class PaymentSheetFormFactory {
         case missingFormSpec
         case missingV1FromSelectorSpec
     }
-
+/*
     enum SaveMode {
         /// We can't save the PaymentMethod. e.g., Payment mode without a customer
         case none
@@ -31,7 +31,32 @@ class PaymentSheetFormFactory {
         /// `setup_future_usage` is set on the PaymentIntent or Setup mode
         case merchantRequired
     }
-    let saveMode: SaveMode
+*/
+    struct SaveMetadata {
+        enum IntentType {
+            case payment_intent
+            case payment_intent_sfu
+            case setup_intent
+        }
+        let intentType: IntentType
+        let hasCustomer: Bool
+//        let checkboxBehavior: PaymentSheet.SavePaymentMethodConsentBehavior
+        let paymentMethodType: PaymentSheet.PaymentMethodType
+
+        func isSaving() -> Bool {
+            return intentType == .payment_intent_sfu || intentType == .setup_intent
+        }
+
+        // To be removed... eventually
+        // because.. we should expose something else to show/hide a checkbox.
+        func isUserSelectable() -> Bool {
+            intentType == .payment_intent && hasCustomer && paymentMethodType.supportsSaveForFutureUseCheckbox()
+        }
+
+    }
+
+//    let saveMode: SaveMode
+    let saveMetadata: SaveMetadata
     let paymentMethod: PaymentSheet.PaymentMethodType
     let configuration: PaymentSheetFormFactoryConfig
     let addressSpecProvider: AddressSpecProvider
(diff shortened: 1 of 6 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
```diff
@@ -23,15 +23,27 @@ class PaymentSheetFormFactory {
...
-    let saveMode: SaveMode
+    struct SaveMetadata {
+        enum IntentType {
+            case payment_intent
+            case payment_intent_sfu
+            case setup_intent
+        }
+        let intentType: IntentType
+        let hasCustomer: Bool
+        let paymentMethodType: PaymentSheet.PaymentMethodType
```
yuki-stripe, line 34: Feels like we can get rid of this and instead calculate the two boolean vars - `isSaving` and `savingIsSelectable` store them on the FormFactory in the initializer.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
```diff
@@ -23,15 +23,27 @@ class PaymentSheetFormFactory {
...
+
+        func isSaving() -> Bool {
```
yuki-stripe, line 36: My only qualm with this is that whether or not the payment method is being saved depends on the state of the checkbox.

wooj-stripe: AH! Right -- isSaving is probably not the best name.. maybe something like `isSaveable` would be a better term.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
wooj-stripe: Hmm, we could, but then .deferredIntent logic alongside intentfirst logic.  For example, savingIsSelectable becomes:
```
        if case .paymentIntent(_, let paymentIntent) = intent,
           paymentIntent.setupFutureUsage == .none {
            return hasCustomer && paymentMethod.supportsSaveForFutureUseCheckbox()
        } else if case .deferredIntent(_, let intentConfig) = intent,
                  case .payment(_, _, let setupFutureUsage, _) = intentConfig.mode,
                  setupFutureUsage != .none {
            return hasCustomer && paymentMethod.supportsSaveForFutureUseCheckbox()
        }
```
versus the proposed:
```
            intentType == .payment_intent && hasCustomer && paymentMethodType.supportsSaveForFutureUseCheckbox()
```

I think this struct is an attempt to simplify some of the inputs we use towards showing mandates/showing checkboxes. Eventually it will have a config where merchants can control whether or not they want to show/hide the checkbox as well as whether or not they want the on/off session if not specified. Open to other ideas though.

yuki-stripe: How about this:
```
        var intentType: IntentType
        switch intent {
        case let .paymentIntent(_, paymentIntent):
            saveMode = saveModeFor(merchantRequiresSave: paymentIntent.setupFutureUsage != .none)
            intentType = paymentIntent.setupFutureUsage != .none ? .payment_intent_sfu : .payment_intent
        case .setupIntent:
            saveMode = .merchantRequired
            intentType = .setup_intent
        case .deferredIntent(_, let intentConfig):
            switch intentConfig.mode {
            case .payment(_, _, let setupFutureUsage, _):
                saveMode = saveModeFor(merchantRequiresSave: setupFutureUsage != .none)
                intentType = setupFutureUsage != .none ? .payment_intent_sfu : .payment_intent
            case .setup:
                saveMode = .merchantRequired
                intentType = .setup_intent
            }
        }
  // ... now calculate savingIsSelectable and isSavable
```

wooj-stripe: Found a way to get rid of the metadata struct by using an existing function called isSettingUp on the `Intent` class.

yuki-stripe (approved):

PR #3538, opened 2024-04-25
Fix PaymentHandler reporting certain next action payment errors as unexpected errors

yuki-stripe:
## Summary
The changes in #3497 have a bug where failing to complete the next action incorrectly report an unexpected error if there is no API-provided error, instead of reporting the last payment error. 

## Testing
Manually tested. Doesn't affect 3DS2, but does affect CashApp. 
Before: error string is "The operation couldn't be completed (STPPaymentHandlerErrorDomain error 12.)"
After: error string is "The customer declined this payment."

## Changelog
[Fixed] Fixed an issue where STPPaymentHandler sometimes reported errors using `unexpectedErrorCode` instead of a more specific error when customers fail a PaymentIntent next action.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## x.x.x x-x-x
 ### PaymentSheet
 * [Added] Support for Multibanco with PaymentIntents.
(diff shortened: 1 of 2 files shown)


porter-stripe (dismissed):
Do we need to update the alert to account for the buggy version?

porter-stripe (approved):

PR #3543, opened 2024-04-26
PaymentSheet now always shows Apple Pay as a button above the saved P…

yuki-stripe:
…Ms list, instead of inside the list. Previously, it only showed as a button if link was also shown as a button.

## Summary
Before:
image

After:
CleanShot 2024-04-26 at 16 13 46@2x

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-1539

## Testing
TBD

## Changelog
* [Changed] PaymentSheet displays Apple Pay as a button when there are saved payment methods and Link isn't available instead of within the list of saved payment methods.


StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetSnapshotTests.swift
```diff
@@ -37,6 +37,7 @@ class PaymentSheetSnapshotTests: STPSnapshotTestCase {
     var runAgainstLiveService: Bool = false
     override func setUp() {
         super.setUp()
+//        recordMode = true
```
davidme-stripe, line 40: Note that you can rerun these by choosing the "AllStripeFrameworks-RecordMode" scheme and running the test, you don't need to add this line anymore

davidme-stripe (approved):

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetSnapshotTests.swift
yuki-stripe: Oooh thanks

PR #3544, opened 2024-04-29
Don't send error analytic every time "terms" is clicked

davidme-stripe:
## Summary
This error should only be sent if the URL fails to launch.

## Motivation
Unnecessary alerts

## Testing
Tested in PS Example


yuki-stripe:
https://github.com/stripe/stripe-ios/pull/3542 😂

yuki-stripe (approved):

PR #3556, opened 2024-05-02
Add vertical layout toggle to playground

yuki-stripe:
## Summary
Adds the toggle and a dummy placeholder API.

Future PRs will update the placeholder API to a real (but internal) API and, of course, make the toggle do something :) 

## Motivation
Keep the PRs small


porter-stripe (approved):
I'm guessing the bot is leaving a comment about the public APIs since the branch it out of date w/ master?

yuki-stripe:
> I'm guessing the bot is leaving a comment about the public APIs since the branch it out of date w/ master?

@porter-stripe  Hmm I guess so. Should we tweak the script? O/w I'd think this would trigger often.

porter-stripe:
@yuki-stripe I'll look into modifying the script to look at the first commit from when the branch was started rather than current master. But it gets me thinking, maybe we shouldn't allow merging into master if your branch isn't up to date with master. I could imagine a scenario where something could break, and technically the bitrise build is building an out of date build since it doesn't have the latest changes in master.

porter-stripe (approved):

PR #3557, opened 2024-05-03
Extract common PaymentSheet initialization parameters

yuki-stripe:
## Summary
Extract common initialization parameters from PS, PS.FC for reuse in vertical mode VC

## Testing
Pure refactor, no behavior change, relying on existing tests.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift
@@ -11,15 +11,12 @@ import Foundation
 @_spi(STP) import StripeUICore
 
 final class PaymentSheetLoader {
-    /// `PaymentSheet.load()` result.
-    enum LoadingResult {
-        case success(
-            intent: Intent,
-            savedPaymentMethods: [STPPaymentMethod],
-            isLinkEnabled: Bool,
-            isApplePayEnabled: Bool
-        )
-        case failure(Error)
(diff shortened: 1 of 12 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift
```diff
@@ -11,15 +11,12 @@ import Foundation
...
-    enum LoadingResult {
-        case success(
-            intent: Intent,
-            savedPaymentMethods: [STPPaymentMethod],
-            isLinkEnabled: Bool,
-            isApplePayEnabled: Bool
-        )
-        case failure(Error)
+    /// All the data that PaymentSheetLoader loaded.
+    struct LoadResult {
```
yuki-stripe, line 15: other name ideas:
- PaymentSheetInitializationParameters
- PaymentSheetInitializationData
- PaymentSheetInitialState

porter-stripe: Not any strong feelings but I like `PaymentSheetInitializationParameters`.

porter-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift
yuki-stripe: Ack, will consider renaming in a future PR.

PR #3563, opened 2024-05-06
Show initial blank vertical view controller in vertical mode for FlowController

yuki-stripe:
## Summary
- Make `FlowControllerViewControllerDelegate` which abstracts out FC's view controller. Both horizontal & vertical VCs conform to it.
- Adds `PaymentSheetVerticalViewController`, barebones implementation.

PaymentSheet to follow

## Testing
Manually tested for now.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetVerticalViewController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetVerticalViewController.swift
@@ -15,10 +15,12 @@ class PaymentSheetVerticalViewController: UIViewController, FlowControllerViewCo
     var selectedPaymentOption: PaymentSheet.PaymentOption?
     /// The type of the payment method that's currently selected in the UI, or unknown if no payment method is selected.
     var selectedPaymentMethodType: PaymentSheet.PaymentMethodType = .stripe(.unknown)
-
-    var delegate: (any FlowControllerViewControllerDelegate)?
+    weak var delegate: FlowControllerViewControllerDelegate?
     let loadResult: PaymentSheetLoader.LoadResult
     let configuration: PaymentSheet.Configuration
+    var intent: Intent {
+        return loadResult.intent
+    }
     var error: Error?
 
     // MARK: - UI properties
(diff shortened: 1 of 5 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetVerticalViewController.swift
```diff
@@ -0,0 +1,83 @@
...
+@_spi(STP) import StripeCore
+@_spi(STP) import StripePayments
+@_spi(STP) import StripePaymentsUI
+@_spi(STP) import StripeUICore
+import UIKit
+
+class PaymentSheetVerticalViewController: UIViewController, FlowControllerViewController {
+    var selectedPaymentOption: PaymentSheet.PaymentOption?
+    /// The type of the payment method that's currently selected in the UI, or unknown if no payment method is selected.
+    var selectedPaymentMethodType: PaymentSheet.PaymentMethodType = .stripe(.unknown)
```
porter-stripe, line 17: Why not nil instead of unknown? I notice it's not optional on the `FlowControllerViewController` protocol but curious.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetVerticalViewController.swift
```diff
@@ -0,0 +1,83 @@
...
+
+    var delegate: (any FlowControllerViewControllerDelegate)?
```
porter-stripe, line 19: Shouldn't this be `weak`? Also I think this can just be `weak var delegate: FlowControllerViewControllerDelegate?`

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetVerticalViewController.swift
yuki-stripe: Great question, I agree this should be nil. I tried making it nil but the refactor was a bit too much for the PR. Will add a TODO though.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetVerticalViewController.swift
yuki-stripe: Oh oops ty, this was auto-generated.

PR #3568, opened 2024-05-06
Remove PaymentSheet Link delegate methods called after confirmation

yuki-stripe:
## Summary
The old flow had unnecessary back and forth between PaymentSheet (PS) and PaymentSheetViewController (PSVC). PSVC was invoking these delegate methods to inform PS that confirmation had finished, but PS was the one who finished the confirmation in the first place! It can simply directly handle Link animations after finishing confirmation before passing control to PSVC.

Small improvement - old code used to wait for the Link spinner to show success animation before continuing, even if Link wasn't used. New code only does that for Link.

## Motivation
Reduce code that vertical mode has to handle.

## Testing
Manually tested cancel, failed payment, successful payment.

https://github.com/stripe/stripe-ios/assets/47796191/779e19ff-b116-4c8f-8d2a-0aad68046ae4

https://github.com/stripe/stripe-ios/assets/47796191/687c5af6-e8ec-4a0f-9db4-fa4b9abe93be

https://github.com/stripe/stripe-ios/assets/47796191/7922af9d-6a19-4086-9825-1b30d9e74bbc

## Changelog
No user facing change.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
@@ -255,7 +255,22 @@ extension PaymentSheet: PaymentSheetViewControllerDelegate {
                 if case let .failed(error) = result {
                     self.mostRecentError = error
                 }
-                completion(result, deferredIntentConfirmationType)
+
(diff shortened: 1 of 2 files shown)


davidme-stripe (approved):
Good catch, thanks for fixing this!

PR #3571, opened 2024-05-08
Refactor PaymentSheet didSelectPayWithLink delegate flow

yuki-stripe:
## Motivation
Cleaner if PaymentSheetViewController is responsible for this - it can clear its text fields, hide the keyboard etc. when Link payment starts. Will make it cleaner for Vertical mode, too, since it will do different things (no text fields to clear or keyboard to hide).  

## Testing
Manually tested Link animations

https://github.com/stripe/stripe-ios/assets/47796191/f95586b7-9fdb-4c3c-ac40-8af9ebe9ac1b

## Changelog
Not user facing


yuki-stripe:
Not actually sure this is better, changing to draft

PR #3572, opened 2024-05-08
Make PaymentSheet show PaymentSheetVerticalViewController in vertical mode

yuki-stripe:
## Summary
Make PaymentSheet show PaymentSheetVerticalViewController in vertical mode.

Abstracts out `PaymentSheetViewController` into `PaymentSheetViewController`

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-2009

## Testing
Manual testing for now

## Changelog
Not user facing

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
@@ -153,21 +153,27 @@ public class PaymentSheet {
             switch result {
             case .success(let loadResult):
                 // Set the PaymentSheetViewController as the content of our bottom sheet
-                let presentPaymentSheetVC = {
-                    let paymentSheetVC = PaymentSheetViewController(
-                        configuration: self.configuration,
-                        loadResult: loadResult,
-                        delegate: self
-                    )
-
-                    self.configuration.style.configure(paymentSheetVC)
-
-                    let updateBottomSheet: () -> Void = {
-                        self.bottomSheetViewController.contentStack = [paymentSheetVC]
+                let paymentSheetVC: PaymentSheetViewControllerProtocol = {
+                    switch self.configuration.appearance.layout {
+                    case .horizontal:
+                        return PaymentSheetViewController(
+                            configuration: self.configuration,
+                            loadResult: loadResult,
+                            delegate: self
+                        )
+                    case .vertical:
+                        let verticalVC = PaymentSheetVerticalViewController(
+                            configuration: self.configuration,
+                            loadResult: loadResult,
+                            isFlowController: false
+                        )
+                        verticalVC.paymentSheetDelegate = self
+                        return verticalVC
                     }
-                    updateBottomSheet()
-                }
-                presentPaymentSheetVC()
+                }()
+
+                self.configuration.style.configure(paymentSheetVC)
+                self.bottomSheetViewController.contentStack = [paymentSheetVC]
             case .failure(let error):
                 completion(.failed(error: error))
             }
@@ -238,7 +244,7 @@ public class PaymentSheet {
 extension PaymentSheet: PaymentSheetViewControllerDelegate {
 
     func paymentSheetViewControllerShouldConfirm(
-        _ paymentSheetViewController: PaymentSheetViewController,
+        _ paymentSheetViewController: PaymentSheetViewControllerProtocol,
         with paymentOption: PaymentOption,
         completion: @escaping (PaymentSheetResult, StripeCore.STPAnalyticsClient.DeferredIntentConfirmationType?) -> Void
     ) {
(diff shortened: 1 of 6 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetVerticalViewController.swift
```diff
@@ -11,16 +11,18 @@
...
-    weak var delegate: FlowControllerViewControllerDelegate?
     let loadResult: PaymentSheetLoader.LoadResult
     let configuration: PaymentSheet.Configuration
     var intent: Intent {
         return loadResult.intent
     }
     var error: Error?
+    let isFlowController: Bool
+    weak var flowControllerDelegate: FlowControllerViewControllerDelegate?
+    weak var paymentSheetDelegate: PaymentSheetViewControllerDelegate?
```
yuki-stripe, line 25: This is a bit weird, TBD if/how we'll make this nicer.  I thought about using an enum like
````
enum Delegate {
 case fc(FlowControllerViewControllerDelegate)
 case ps(...)
}
```
but that would make the delegate strong instead of weak.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
```diff
@@ -153,21 +153,25 @@ public class PaymentSheet {
...
+                            isFlowController: false
+                        )
+                        verticalVC.paymentSheetDelegate = self
+                        return verticalVC
                     }
-                    updateBottomSheet()
-                }
-                presentPaymentSheetVC()
+                }()
+                self.bottomSheetViewController.contentStack = [paymentSheetVC]
```
porter-stripe, line 174: Any reason we set the stack explicitly instead of using `pushContentViewController`?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetVerticalViewController.swift
```diff
@@ -11,16 +11,18 @@
...
     var selectedPaymentOption: PaymentSheet.PaymentOption?
     var selectedPaymentMethodType: PaymentSheet.PaymentMethodType?
-    weak var delegate: FlowControllerViewControllerDelegate?
     let loadResult: PaymentSheetLoader.LoadResult
     let configuration: PaymentSheet.Configuration
     var intent: Intent {
         return loadResult.intent
     }
     var error: Error?
+    let isFlowController: Bool
```
porter-stripe, line 23: Does it make sense to attach this to the load result?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetVerticalViewController.swift
```diff
@@ -77,7 +80,11 @@ extension PaymentSheetVerticalViewController: BottomSheetContentViewController {
 
     func didTapOrSwipeToDismiss() {
         // TODO
-        delegate?.flowControllerViewControllerShouldClose(self, didCancel: true)
+        if isFlowController {
+            flowControllerDelegate?.flowControllerViewControllerShouldClose(self, didCancel: true)
+        } else {
+            paymentSheetDelegate?.paymentSheetViewControllerDidCancel(self)
+        }
```
porter-stripe, line 87: Is there any harm in calling both of them? One should be nil so it'd no-op right? Then we could remove `isFlowController` all together.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetVerticalViewController.swift
porter-stripe: I think the current approach seems reasonable, should we assert that we don't have both set at one time, that seems like invalid state?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
yuki-stripe: I think it's so that we can replace the loading screen.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetVerticalViewController.swift
yuki-stripe: I don't think so, it's not something that's loaded.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetVerticalViewController.swift
yuki-stripe: No harm in calling both. We'll need `isFlowController` or something similar later though.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetVerticalViewController.swift
yuki-stripe: Hmm if we are worried about that invalid state I think I'd rather try to find some way to make it impossible vs. asserting. I don't think I am though, like...I don't think anything even breaks if both are set?

porter-stripe (dismissed):

porter-stripe (approved):

PR #3582, opened 2024-05-15
Remove server-side driven next_action parsing and processing

wooj-stripe:
## Summary
Removing next action server-side driven next_action parsing and processing

## Motivation
Change in direction. For reference: [Original commit](https://github.com/stripe-ios/stripe-ios-private/pull/1306/files#diff-ba4b18b5a8005808a22a28c75ed8665bdcfb68e16bbea4013039dbdba2d7fcfc)

## Testing
Relying on existing tests w/ stp payment handler.


StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
```diff
@@ -210,25 +206,16 @@ public class STPPaymentHandler: NSObject {
...
+            if let paymentIntent = paymentIntent, status == .succeeded {
                 let successIntentState =
                     paymentIntent.status == .succeeded || paymentIntent.status == .requiresCapture
                     || (paymentIntent.status == .processing
                         && STPPaymentHandler._isProcessingIntentSuccess(
                             for: paymentIntent.paymentMethod?.type ?? .unknown
                         )
                         || (paymentIntent.status == .requiresAction
-                            && strongSelf.isNextActionSuccessState(
+                            && self.isNextActionSuccessState(
```
wooj-stripe, line 218: We were using `self` in the deleted code which was keeping the self instance around. It seems to have been needed for some of the dashboard tests: 
* testDashboard_DeferredPaymentIntent_new_saveChecked
* testDashboard_DeferredPaymentIntent_new_saveUnchecked
* testDashboard_DeferredPaymentIntent_saved

I ran these tests with  the profiler and ran "leaks" and didn't see any issue.  I addition, I ran the profiler with paymentsheet example (confirmed and intent) and also didn't see any leaks.

yuki-stripe: IMO we shouldn't retain self in this completion block so that it's consistent with the rest of the code here. This is a retain cycle that's removed when the completion block is called and currentAction is released (it retains the completion block as a property).

Those tests are failing without the retain cycle b/c they don't retain their STPPaymentHandler instance.

wooj-stripe: AH! I was searching for how to hold onto STPPaymentHandler, and didn't realize until now it was being passed into the confirm call!  Thanks!!

PR #3583, opened 2024-05-15
Update vertical layout API, stop using import @testable in playground

yuki-stripe:
## Summary
Update API per https://docs.google.com/document/d/1ZVVxyD5Ue6Xmh54JXf5r3rRCUBPxas01vl7if-FaO_M/edit

Also removes @testable import, which breaks testflight and profiler.

## Testing
Existing tests

## Changelog
Not user facing

### DIFF
--- a/Stripe.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/Stripe.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -38,5 +38,5 @@
       }
     }
   ],
-  "version" : 3
+  "version" : 2
 }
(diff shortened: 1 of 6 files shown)


Stripe.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -38,5 +38,5 @@
       }
     }
   ],
-  "version" : 2
+  "version" : 3
```
wooj-stripe, line 41: probably not intentional?

yuki-stripe: Ugh, why does this change?

wooj-stripe (approved):

PR #3586, opened 2024-05-16
Ability to remove payment methods in vertical mode

porter-stripe:
## Summary
- Uses `RowButton` in `PaymentMethodRowButton`
- Made the function in `VerticalSavedPaymentMethodsViewControllerDelegate` more generic to handle passing back the most up to date payment method list
- Added the ability to remove a payment method
- Adds snapshot tests for `PaymentMethodRowButton` to more easily test the different states of the button
- Adds some UI tests for various removal flows
- Adds some unit tests to `VerticalSavedPaymentMethodsViewController` in an attempt to better test the state outside of UI tests

https://github.com/stripe/stripe-ios/assets/88012362/682b55ad-701d-4b6b-84f4-c37c6e2ca553

## Motivation
Vertical mode

## Testing
- Manual
- New snapshots
- New UI tests
- New unit tests

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/Vertical Saved Payment Method Screen/PaymentMethodRowButton.swift
```diff
@@ -14,34 +14,29 @@ import UIKit
...
     }
 
     // MARK: Internal properties
     var state: State = .unselected {
         didSet {
             previousState = oldValue
 
-            selectionTapGesture.isEnabled = !isEditing
-            shadowRoundedRect.isSelected = isSelected
+            // Any gesture recognizers on the rowButton will intercept taps on the update and remove buttons, disable them if in edit mode
```
yuki-stripe, line 34: Can you try putting the gesture recognizer on the shadowed rounded rect in `RowButton`?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/Vertical Saved Payment Method Screen/PaymentMethodRowButton.swift
```diff
@@ -92,78 +94,58 @@ final class PaymentMethodRowButton: UIView {
...
+        let button: RowButton = .makeForSavedPaymentMethod(paymentMethod: paymentMethod, appearance: appearance) { [weak self] _ in
+            guard let self, !isEditing else { return }
+            state = .selected
+            delegate?.didSelectButton(self, with: paymentMethod)
+        }
 
-    private lazy var selectionTapGesture: UITapGestureRecognizer = {
-        return UITapGestureRecognizer(target: self, action: #selector(handleSelectionTap))
+        button.translatesAutoresizingMaskIntoConstraints = false
+        button.addAndPinSubview(stackView)
```
yuki-stripe, line 125: I don't think this is right, this means the button content can overlap the `stackView` here?  `RowButton` has an unimplemented `rightAccessoryView` that I was thinking we would use for this purpose.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/Vertical Saved Payment Method Screen/VerticalSavedPaymentMethodsViewController.swift
```diff
@@ -12,14 +12,14 @@ import Foundation
 import UIKit
 
 protocol VerticalSavedPaymentMethodsViewControllerDelegate: AnyObject {
-    func didSelectPaymentMethod(_ paymentMethod: STPPaymentMethod)
+    func didComplete(with selectedPaymentMethod: STPPaymentMethod?, latestPaymentMethods: [STPPaymentMethod])
```
yuki-stripe, line 15: Let's add a docstring?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/Vertical Saved Payment Method Screen/PaymentMethodRowButton.swift
porter-stripe: Ah, thanks. Didn't see `rightAccessoryView`, updated to use it.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/Vertical Saved Payment Method Screen/PaymentMethodRowButton.swift
porter-stripe: Good idea! Yeah that works, one side-effect is that we need to disable interaction on the labels and image since they sit above the rounded rect, but this seems reasonable to me.

Alternatively we could add those views as subviews on the rounded rect.

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
```diff
@@ -3483,6 +3483,122 @@ extension PaymentSheetUITestCase {
...
+
+        // We should automatically exit edit mode hiding the edit/done button
+        XCTAssertFalse(app.buttons["Edit"].waitForExistence(timeout: 2.0))
+        XCTAssertFalse(app.buttons["Done"].waitForExistence(timeout: 2.0))
+
+        // We should also hide the remove button on the last saved payment method as allowsRemovalOfLastSavedPaymentMethod is false
+        XCTAssertFalse(app.buttons["CircularButton.Remove"].waitForExistence(timeout: 2.0))
+        // We should still be on the manage screen not kicked to the main screen as one saved payment method remains
+        XCTAssertTrue(app.staticTexts["Select card"].waitForExistence(timeout: 5.0))
+    }
```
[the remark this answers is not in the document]
porter-stripe: Yeah good point, we should discuss what functional tests may look like in iOS weekly. I'll put something on the agenda.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/Vertical Saved Payment Method Screen/VerticalSavedPaymentMethodsViewController.swift
```diff
@@ -109,6 +139,32 @@ class VerticalSavedPaymentMethodsViewController: UIViewController {
...
+
+        // If we deleted the last payment method kick back out to the main screen
+        if paymentMethodRows.isEmpty {
+            completeSelection()
+        }
+    }
+
+    private func completeSelection() {
+        _ = self.bottomSheetController?.popContentViewController()
+        self.delegate?.didComplete(with: selectedPaymentMethod, latestPaymentMethods: paymentMethods)
```
yuki-stripe, line 166: General best practice is to let the thing that presented/pushed you also pop/dismiss you. In this case I think we'll want the delegate to control when to pop the sheet so that it can re-render the list of payment methods (possibly removing the saved PM row) before popping.

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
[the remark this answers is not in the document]
yuki-stripe: I think having one UI test for the behavior of kicking the user out is good. 

But then the edge cases for like 
- Only 1 payment method left and allowsRemovalOfLastSavedPaymentMethod is false
- Only 1 payment method left and allowsRemovalOfLastSavedPaymentMethod is false, but we have an updatable card

feel like they shouldn't just duplicate the UI test, they can e.g. just test that `canEdit` returns the right value for the given state.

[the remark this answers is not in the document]
yuki-stripe: There are going to be unfortunately even more states that this thing can be in w/ the upcoming `CustomerSession` property that's like `allowsRemovalOfSavedPaymentMethods` and we wouldn't want UI tests for all the combinations.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButton.swift
```diff
@@ -98,11 +97,22 @@ class RowButton: UIView {
...
         ])
-        addGestureRecognizer(UITapGestureRecognizer(target: self, action: #selector(handleTap)))
+
+        if let rightAccessoryView {
+            rightAccessoryView.translatesAutoresizingMaskIntoConstraints = false
+            addSubview(rightAccessoryView)
+            NSLayoutConstraint.activate([
+                rightAccessoryView.topAnchor.constraint(equalTo: topAnchor),
+                rightAccessoryView.bottomAnchor.constraint(equalTo: bottomAnchor),
+                rightAccessoryView.trailingAnchor.constraint(equalTo: trailingAnchor),
```
yuki-stripe, line 107: This is missing a leading anchor constraint - w/o this can overlap the other content in the button.

porter-stripe: I think the labels are actually missing their trailing anchor, the labels shouldn't overlap with the right accessory view or spill over if it has long text. I updated is so the labels trailing anchor is either based on the containing view or the right accessory view.
Before:
Simulator Screenshot - iPhone 15 Plus - 2024-05-20 at 12 04 17

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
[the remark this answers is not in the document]
porter-stripe: Agree we won't want UI tests for all combinations, and I can see your point about these being heavy, but they definitely give the highest level of confidence out of any tests (and I don't think it's close). I think the UI tests in this PR are reasonable, and really increase my confidence in a newly built feature.

To scale for combinations related to `CustomerSession` or other edge cases I think extending the new tests in `VerticalSavedPaymentMethodsViewControllerTests` would be my recommendation.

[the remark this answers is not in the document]
yuki-stripe: Dangit! I left two comments but the first one disappeared! That makes this second comment misleading.  

My first comment - I think this UI test is good for testing the whole flow including kicking the customer out of edit mode.

For the edge cases, though:
-        // Only 1 payment method left and allowsRemovalOfLastSavedPaymentMethod is false, but we have an updatable card
-        // Only 1 payment method left and allowsRemovalOfLastSavedPaymentMethod is false

What if we just test that the VC's `canEdit` bool returns the right value for the given state?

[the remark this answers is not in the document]
porter-stripe: That makes sense, I've deleted the 2 extra UI tests. Then for the edge cases mentioned we do already have unit tests (VerticalSavedPaymentMethodsViewControllerTests).

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButton.swift
```diff
@@ -118,9 +118,79 @@ extension RowButton {
...
+    static func makeForSavedPaymentMethod(paymentMethod: STPPaymentMethod, appearance: PaymentSheet.Appearance, rightAccessoryView: UIView? = nil, didTap: @escaping (RowButton) -> Void) -> RowButton {
         let imageView = UIImageView(image: paymentMethod.makeSavedPaymentMethodRowImage())
         imageView.contentMode = .scaleAspectFit
-        return RowButton(appearance: appearance, imageView: imageView, text: paymentMethod.paymentSheetLabel, didTap: didTap)
+        let button = RowButton(appearance: appearance, imageView: imageView, text: paymentMethod.paymentSheetLabel, rightAccessoryView: rightAccessoryView, didTap: didTap)
+        button.shadowRoundedRect.accessibilityLabel = paymentMethod.paymentSheetAccessibilityLabel
+        return button
+    }
+
+    init(appearance: PaymentSheet.Appearance, imageView: UIImageView, text: String, subtext: String? = nil, rightAccessoryView: UIView? = nil, didTap: @escaping (RowButton) -> Void) {
```
yuki-stripe, line 129: Hrm is this a bad merge? Looks like a ~dupe of the init earlier in this file

porter-stripe: Hrm yep, the merge didn't pick up any of the diff correctly :/. Should be fixed now.

yuki-stripe (approved):

PR #3602, opened 2024-05-23
Integrate CustomerSession directly into CustomerSheet

wooj-stripe:
## Summary
Integrate CustomerSession directly into CustomerSheet

## Motivation
Change in direction of integration shape

## Testing
Relying on existing UI tests for CustomerSession + Customer Sheet

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSheetDataSource.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSheetDataSource.swift
@@ -35,18 +35,22 @@ class CustomerSheetDataSource {
     func loadPaymentMethodInfo(customerSessionAdapter: CustomerSessionAdapter,
                                completion: @escaping (Result<([STPPaymentMethod], CustomerPaymentOption?, STPElementsSession), Error>) -> Void) {
         Task {
-            async let (elementsSessionResult, customerSessionClientSecret) = try customerSessionAdapter.elementsSessionWithCustomerSessionClientSecret()
+            do {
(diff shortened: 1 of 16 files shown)


oleksandr-vonahi:
This is massive. @gooroodev can you help reviewing it?

admsev:
This is massive. @gooroodev would you mind reviewing it?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSheetDataSource.swift
```diff
@@ -0,0 +1,164 @@
...
+        }
+    }
+
+    func fetchSavedPaymentMethods() async throws -> [STPPaymentMethod]? {
+        switch dataSource {
+        case .customerAdapter(let customerAdapter):
+            return try await customerAdapter.fetchPaymentMethods()
+        case .customerSession(let customerSessionAdapter):
+            let elementsSessionResult = try await customerSessionAdapter.elementsSession()
+            return elementsSessionResult.customer?.paymentMethods ?? []
```
davidme-stripe, line 124: Should this return nil or an empty array?

davidme-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSheetDataSource.swift
wooj-stripe: Oof, great catch.  In the event that we fail to get payment methods, we need to return nil, which corresponds to dismissing the sheet and force users to effectively re-open the sheet.

davidme-stripe (approved):

PR #3609, opened 2024-05-28
Add AccessoryButton

porter-stripe:
## Summary
- Added the view more button with various modes
- In future PRs we will set the correct button type, and handle taps back to the delegate
Simulator Screenshot - iPhone 15 Plus - 2024-05-28 at 12 51 47
- Respects `paymentMethodRemove` when removing a payment method

## Motivation
- Vert mode

## Testing
- Manual
- New snapshots

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/AccessoryButton.swift
```diff
@@ -58,11 +59,14 @@ final class AccessoryButton: UIButton {
         super.init(frame: .zero)
         setTitle(accessoryType.text, for: .normal)
         setTitleColor(appearance.colors.primary, for: .normal) // TODO read secondary action color
-        titleLabel?.font = appearance.scaledFont(for: appearance.font.base.medium, style: .subheadline, maximumPointSize: 20)
+        titleLabel?.font = appearance.scaledFont(for: appearance.font.base.medium, style: .footnote, maximumPointSize: 20)
```
porter-stripe, line 62: This is technically font size 13 when the figma calls for 12. I felt like 12 just looks really small.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/AccessoryButton.swift
```diff
@@ -0,0 +1,85 @@
...
+
+import Foundation
+@_spi(STP) import StripeCore
+@_spi(STP) import StripeUICore
+import UIKit
+
+final class AccessoryButton: UIButton {
+
+    enum AccessoryType {
+        case none
```
yuki-stripe, line 16: This doesn't seem right, we should remove it. The `none` case isn't a type of AccessoryButton - you can't initialize one with this value.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/AccessoryButton.swift
```diff
@@ -0,0 +1,85 @@
...
+            case .viewMore:
+                return UIEdgeInsets(top: 2, left: 4, bottom: 0, right: 0)
+            }
+        }
+    }
+
+    override var intrinsicContentSize: CGSize {
+        let size = super.intrinsicContentSize
+
+        // TODO(porter) Figure out how to handle this for VisionOS, the alternative (`UIButton.Configuration) was introduced in iOS 15 and we currently support below iOS 15.
```
yuki-stripe, line 53: What are we trying to solve by overriding this method?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/AccessoryButton.swift
```diff
@@ -0,0 +1,85 @@
...
+//
+//  Created by Nick Porter on 5/28/24.
+//
+
+import Foundation
+@_spi(STP) import StripeCore
+@_spi(STP) import StripeUICore
+import UIKit
+
+final class AccessoryButton: UIButton {
```
yuki-stripe, line 13: Why override UIButton? I think by default, as a matter of best practice, we should avoid subclassing UIKit stuff and prefer composition.  In this case, this doesn't really act like a subclass of UIButton. e.g. users can call `AccessoryButton.buttonWithType` on it, but what does that mean here?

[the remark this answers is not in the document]
porter-stripe: Makes sense. I've moved this under the `RowButton` namespace and renamed to `RightAccessoryButton`.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/AccessoryButton.swift
porter-stripe: Fair point, removed this.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/AccessoryButton.swift
porter-stripe: Overriding this ensures that auto layout respects the custom padding to accurately size and position the button. Without this, if we move the image to align with the figmas it could draw outside of it's size and that doesn't mesh well with auto layout. So by taking into account the image edge sets we can define the size more accurately. I added a comment.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/AccessoryButton.swift
porter-stripe: Yeah agree with preferring composition. In this case it's mainly so we can override the intrinsic content size so the button is sized correctly w.r.t to the image offset. Open to suggestions for alternative approaches, could maybe change this to sub-class `UIView` or `UIControl` like we do for `CircularButton` but that is still UIKit.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RightAccessoryButton.swift
```diff
@@ -31,49 +31,63 @@ extension RowButton {
...
+            imageView.contentMode = .scaleAspectFit
+            imageView.isAccessibilityElement = false
+            return imageView
+        }
+
+        private var stackView: UIStackView {
+            var views: [UIView] = [label]
+            if let imageView {
+                views.append(imageView)
+            }
```
yuki-stripe, line 63: Can do like
```
let views: [UIView] = [label, imageView].compactMap { $0 }
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RightAccessoryButton.swift
```diff
@@ -31,49 +31,63 @@ extension RowButton {
...
 
-            // TODO(porter) Figure out how to handle this for VisionOS, the alternative (`UIButton.Configuration) was introduced in iOS 15 and we currently support below iOS 15.
-            #if !canImport(CompositorServices)
-            return CGSize(width: size.width + imageEdgeInsets.left + imageEdgeInsets.right,
-                          height: size.height + imageEdgeInsets.top + imageEdgeInsets.bottom)
-            #else
-            return size
-            #endif
+            let stackView = UIStackView(arrangedSubviews: views)
+            stackView.axis = .horizontal
```
yuki-stripe, line 66: can remove, this is the default

yuki-stripe (approved):

porter-stripe:
ACK the comments, will add in a following PR.

PR #3618, opened 2024-05-30
Instant Debits: fixed an issue where banks would sometimes have a + sign instead of a space

kgaidis-stripe:
## Summary

^ see title where banks with spaces would be returned with a space in between:

```
▿ InstantDebitsLinkedBankImplementation
  - paymentMethodId : "...."
  ▿ bankName : Optional<String>
    - some : "Bank+of+America". //// <----- this is a problem
  ▿ last4 : Optional<String>
    - some : "1234"
```

## Testing

Here in the screenshot we can see that the bank with spaces is now 

| Before | After |
| --- | --- |
| Screenshot 2024-05-30 at 3 07 50 PM | IMG_63E00D66962A-1 |

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/Web/FinancialConnectionsWebFlowViewController.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/Web/FinancialConnectionsWebFlowViewController.swift
@@ -292,7 +292,10 @@ extension FinancialConnectionsWebFlowViewController {
         return components
             .queryItems?
             .first(where: { $0.name == key })?
-            .value?.removingPercentEncoding
+            .value?
+            .removingPercentEncoding?
+        // backend can return "+" instead of a more-common encoding of "%20" for spaces
+            .replacingOccurrences(of: "+", with: " ")
     }
 }
 


StripeFinancialConnections/StripeFinancialConnections/Source/Web/FinancialConnectionsWebFlowViewController.swift
```diff
@@ -292,7 +292,10 @@ extension FinancialConnectionsWebFlowViewController {
         return components
             .queryItems?
             .first(where: { $0.name == key })?
-            .value?.removingPercentEncoding
+            .value?
+            .removingPercentEncoding?
+        // backend can return "+" instead of a more-common encoding of "%20" for spaces
+            .replacingOccurrences(of: "+", with: " ")
```
bric-stripe, line 298: non-blocking, but any concern with banks that legitimately have a + as part of their name? seems only a matter of time before we have banks hopping on the "plus" branding :)  "Chase+"

bric-stripe (approved):

StripeFinancialConnections/StripeFinancialConnections/Source/Web/FinancialConnectionsWebFlowViewController.swift
kgaidis-stripe: yea its a good point; I looked at various institutions we have and this should nearly never happen (if at all), and if it does this is used just for display so this seems like an OK decision _for now_ (not perfect! but justifiable)

PR #3621, opened 2024-05-31
Fall back to locale country code in Link createPaymentDetails

yuki-stripe:
## Motivation
https://jira.corp.stripe.com/browse/RUN_MOBILESDK-3241

## Testing
Manually tested using the ConsumerSessionTests - if I don't set address.country, the Link flow fails.  After this change, it succeeds.

## Changelog
Not user facing?

### DIFF
--- a/Stripe/StripeiOSTests/ConsumerSessionTests.swift
+++ b/Stripe/StripeiOSTests/ConsumerSessionTests.swift
@@ -141,72 +141,70 @@ class ConsumerSessionTests: XCTestCase {
 
         wait(for: [expectation], timeout: STPTestingNetworkRequestTimeout)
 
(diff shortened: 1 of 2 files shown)


Stripe/StripeiOSTests/ConsumerSessionTests.swift
```diff
@@ -141,72 +141,70 @@ class ConsumerSessionTests: XCTestCase {
 
         wait(for: [expectation], timeout: STPTestingNetworkRequestTimeout)
 
-        if let consumerSession = sessionWithKey?.consumerSession {
```
yuki-stripe, line 144: Sorry for this weird diff.  All I did was change this line and remove the { }.

davidme-stripe (approved):

PR #3528, opened 2024-04-19
Instant Debits: first iteration of Instant Debits in Payment Sheet

kgaidis-stripe:
## Summary

This PR implements the first iteration of Instant Debits in the Payment Sheet. The PR:
1. Adds a new payment method (instant debits) so we could display a new "Bank Account" payment method in the payment sheet
2. If "Bank Account" (Instant Debits) is selected, the code displays UI, and routes through Payment Sheet and Financial Connections

⚠️  Some important notes for reviewing:
- This is just the first iteration. Despite Android equivalent already being landed, we still have a lot of things to resolve and improve before launching to merchants: handle flags for rollout, design, testing, etc.

## Testing

See two videos:

### PaymentIntent Succeeds

https://github.com/stripe/stripe-ios/assets/105514761/dddd0e22-9a9a-4415-8933-0c3aa38368d9

### SetupIntent Succeeds

https://github.com/stripe/stripe-ios/assets/105514761/4ff6341c-a3d4-4d72-b23a-9957959c4873

### Financial Connections Still Works

https://app.bitrise.io/build/30cd250d-0695-4af1-bd81-5c5de2f882f3

```
FinancialConnectionsNetworkingUITests
    ✓ testNativeNetworkingTestMode (111.850 seconds)
    ✓ testNativeNetworkingTestModeSignUpWithMultiSelectAndPrefilledEmail (30.682 seconds)
FinancialConnectionsUITests
    ✓ testDataLiveModeOAuthNativeAuthFlow (29.338 seconds)
    ✓ testDataLiveModeOAuthWebAuthFlow (18.610 seconds)
    ✓ testDataTestModeOAuthNativeAuthFlow (18.501 seconds)
    ✓ testPaymentSearchInLiveModeNativeAuthFlow (26.620 seconds)
    ✓ testPaymentTestModeLegacyNativeAuthFlow (20.153 seconds)
    ✓ testPaymentTestModeManualEntryNativeAuthFlow (24.347 seconds)
```

PR #3625, opened 2024-05-31
Instant Debits: switched Instant Debits payment method to be more correct in when its enabled

kgaidis-stripe:
## Summary

^

This [Slack thread](https://stripe.slack.com/archives/C02HQBW38JX/p1717178825321049?thread_ts=1717108585.415089&cid=C02HQBW38JX) brought up a need to double check whether the Instant Debits logic is the same as Android, and this PR modifies the logic of whether to add `.instantDebits` just a little bit. 

Note that Instant Debits is still not released to users and is in active development.

## Testing

I ran the UI tests manually myself which check Instant Debits end-to-end (payment and setup intents):

Screenshot 2024-05-31 at 11 48 27 AM

These tests will also be run on this PR by Bitrise!

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/Web/FinancialConnectionsWebFlowViewController.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/Web/FinancialConnectionsWebFlowViewController.swift
@@ -165,7 +165,9 @@ extension FinancialConnectionsWebFlowViewController {
                         {
                             let instantDebitsLinkedBank = InstantDebitsLinkedBankImplementation(
                                 paymentMethodId: paymentMethodId,
-                                bankName: Self.extractValue(from: returnUrl, key: "bank_name"),
+                                bankName: Self.extractValue(from: returnUrl, key: "bank_name")?
+                                // backend can return "+" instead of a more-common encoding of "%20" for spaces
+                                    .replacingOccurrences(of: "+", with: " "),
                                 last4: Self.extractValue(from: returnUrl, key: "last4")
                             )
                             self.notifyDelegateOfSuccess(result: .instantDebits(instantDebitsLinkedBank))
@@ -293,9 +295,7 @@ extension FinancialConnectionsWebFlowViewController {
             .queryItems?
             .first(where: { $0.name == key })?
             .value?
-            .removingPercentEncoding?
-        // backend can return "+" instead of a more-common encoding of "%20" for spaces
-            .replacingOccurrences(of: "+", with: " ")
+            .removingPercentEncoding
     }
 }
 
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
@@ -126,6 +126,19 @@ extension PaymentSheet {
                 }
             }
 
+            if
+                recommendedStripePaymentMethodTypes.contains(.link),
+                !recommendedStripePaymentMethodTypes.contains(.USBankAccount),
+                !intent.isDeferredIntent,
+                intent.linkFundingSources?.contains(.bankAccount) == true
+            {
+                // we must add this BEFORE we do filtering via
+                // `PaymentSheet.PaymentMethodType.supportsAdding`
+                // because we want to filter out instant debits
+                // IF the user has not linked Financial Connections
+                recommendedStripePaymentMethodTypes.append(.instantDebits)
+            }
+
             recommendedStripePaymentMethodTypes = recommendedStripePaymentMethodTypes.filter { paymentMethodType in
                 let availabilityStatus = PaymentSheet.PaymentMethodType.supportsAdding(
                     paymentMethod: paymentMethodType,
@@ -150,15 +163,6 @@ extension PaymentSheet {
                 .subtracting(Set(intent.elementsSession.externalPaymentMethods.map { $0.type }))
             if logAvailability && !missingExternalPaymentMethods.isEmpty {
                 print("[Stripe SDK]: PaymentSheet could not offer these external payment methods: \(missingExternalPaymentMethods). See https://stripe.com/docs/payments/external-payment-methods#available-external-payment-methods")
-            }
-
-            if
-                intent.supportsLink,
-                !recommendedStripePaymentMethodTypes.contains(.USBankAccount),
-                !intent.isDeferredIntent,
-                intent.linkFundingSources?.contains(.bankAccount) == true
-            {
-                recommendedStripePaymentMethodTypes.append(.instantDebits)
             }
 
             // The full ordered list of recommended payment method types:
@@ -221,6 +225,8 @@ extension PaymentSheet {
                         return [.returnURL]
                     case .USBankAccount, .boleto:
                         return [.userSupportsDelayedPaymentMethods]
+                    case .instantDebits:
+                        return [.financialConnectionsSDK]
                     case .sofort, .iDEAL, .bancontact:
                         // n.b. While sofort, iDEAL, and bancontact are themselves not delayed, they turn into SEPA upon save, which IS delayed.
                         return [.returnURL, .userSupportsDelayedPaymentMethods]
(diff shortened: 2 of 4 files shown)


StripeFinancialConnections/StripeFinancialConnections/Source/Web/FinancialConnectionsWebFlowViewController.swift
```diff
@@ -165,7 +165,9 @@ extension FinancialConnectionsWebFlowViewController {
                         {
                             let instantDebitsLinkedBank = InstantDebitsLinkedBankImplementation(
                                 paymentMethodId: paymentMethodId,
-                                bankName: Self.extractValue(from: returnUrl, key: "bank_name"),
+                                bankName: Self.extractValue(from: returnUrl, key: "bank_name")?
+                                // backend can return "+" instead of a more-common encoding of "%20" for spaces
+                                    .replacingOccurrences(of: "+", with: " "),
```
kgaidis-stripe, line 170: this is a small change that mitigates the impact of `.replacingOccurrences(of: "+", with: " ")` to only the `bank_name`; just defensive programming

More context is here: https://github.com/stripe/stripe-ios/pull/3618

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
```diff
@@ -126,6 +126,19 @@ extension PaymentSheet {
...
+                !recommendedStripePaymentMethodTypes.contains(.USBankAccount),
+                !intent.isDeferredIntent,
+                intent.linkFundingSources?.contains(.bankAccount) == true
+            {
+                // we must add this BEFORE we do filtering via
+                // `PaymentSheet.PaymentMethodType.supportsAdding`
+                // because we want to filter out instant debits
+                // IF the user has not linked Financial Connections
+                recommendedStripePaymentMethodTypes.append(.instantDebits)
+            }
```
kgaidis-stripe, line 140: I moved the code up so we could start hitting the `recommendedStripePaymentMethodTypes.filter { paymentMethodType` code

We want to execute that code so we could check whether Financial Connections is linked:

```
                    case .instantDebits:
                        return [.financialConnectionsSDK]
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
```diff
@@ -126,6 +126,19 @@ extension PaymentSheet {
                 }
             }
 
+            if
+                recommendedStripePaymentMethodTypes.contains(.link),
```
kgaidis-stripe, line 130: this was changed from `intent.supportsLink`

which is:

```
extension Intent {
    var supportsLink: Bool {
        // Either Link is an allowed Payment Method in the elements/sessions response, or passthrough mode (Link as a Card PM) is allowed
        return recommendedPaymentMethodTypes.contains(.link) || linkPassthroughModeEnabled
    }
```

so the totality of the change is just removing `linkPassthroughModeEnabled`

which better matches what Android is doing

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+PaymentMethodAvailability.swift
```diff
@@ -23,6 +23,7 @@ extension PaymentSheet {
         .klarna, .afterpayClearpay, .affirm,
         .iDEAL, .bancontact, .sofort, .SEPADebit, .EPS, .giropay, .przelewy24,
         .USBankAccount,
+        .instantDebits,
```
kgaidis-stripe, line 26: This was necessary because we now do extra filtering of `instantDebits`

Code point 1:
- https://github.com/stripe/stripe-ios/blob/ef830fd8/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift#L210

Code point 2:
- https://github.com/stripe/stripe-ios/blob/ef830fd8/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift#L336-L338

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
kgaidis-stripe: This is the Android equivalent (note that thier ".link" check is more upstream so its not immediately obvious in this snippet):

https://github.com/stripe/stripe-android/blob/9785cc59af2773ec3bd059ce7ab168080060ce08/paymentsheet/src/main/java/com/stripe/android/lpmfoundations/paymentmethod/AddPaymentMethodRequirement.kt#L60-L68

kgaidis-stripe:
@yuki-stripe 

Your message on whether Instant Debits should check "link" spurred me to check the logic once more.

It still keeps the ".link" check which maybe doesn't solve what you were trying to get at, but it did land a PR to review as you might be familiar with some of these type of checks in PaymentSheet 😄

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
yuki-stripe: Sweet!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
yuki-stripe: It's surprising to me that `instantDebits` is added despite it not being, apparently, on the payment method types. Like, it's valid to confirm a PI with `instantDebits` payment method type even if the PI.payment_method_types does not contain `instantDebits`?

yuki-stripe: Wow, interesting!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
yuki-stripe: I feel like this should only be applicable if linkPassthroughMode is false?  Like, passthrough mode means merchants see the PM type as the underlying PM (so, US Bank Account here)?

kgaidis-stripe: > I feel like this should only be applicable if linkPassthroughMode is false? Like, passthrough mode means merchants see the PM type as the underlying PM (so, US Bank Account here)?

I reached out to folks on this part - atm I believe Android doesn't contain this logic

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
kgaidis-stripe: No I don't think it is, its just for PaymentSheet client-side code (the switch I mentioned earlier happens client-side)

yuki-stripe: Ah okay. We should remove it from STPPaymentMethodType then, that is a (public) type that corresponds to https://docs.stripe.com/api/payment_methods/object#payment_method_object-type.

kgaidis-stripe: hopefully we can do it as part of a separate PR as this particular decision was approved in https://github.com/stripe/stripe-ios/pull/3528 (happy to refactor [will just need help from payment on how else to structure it])

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
kgaidis-stripe: @yuki-stripe 

Talked to Android folks, and this looks to be fine as Instant Debits logic makes sure its false anyway (note that `intent.linkFundingSources?.contains(.bankAccount) == true` has a lot of logic behind it for it to be true [including some instant debits flags that must be true for it to add the instant debits payment method])

Quotes from @.carlosmuvi :

> passthrough mode cannot be enabled if the link payment method is passed (PM mode and PT mode are mutually exclusive).

> Instant debits is not compatible with passthrough mode (something that will change with Panther)

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
yuki-stripe: Yes please!

yuki-stripe (approved):

PR #3629, opened 2024-06-02
Refactor, DRY AddPaymentMethodViewController

yuki-stripe:
## Summary
`AddPaymentMethodViewController` (APMVC) originally controlled both the form *and* the horizontal list of PMs, all in one. 

This PR **moves** form-related logic from APMVC to `PaymentMethodFormViewController` (PMFVC), so that it can be reused by vertical mode.  It also makes APMVC **use* PMFVC.

Some big bits of logic that was moved:
- Creating paymentOption from form. Link and Instant Debits both have special hacks.
- Default billing <> shipping address code in viewWillAppear

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-2130

## Testing

- Form cache is reset on load
- US Bank and others preserve details when switching between forms
- Shipping details updates form

## Changelog
Not user facing

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/AddPaymentMethodViewController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/AddPaymentMethodViewController.swift
@@ -267,8 +267,7 @@ class AddPaymentMethodViewController: UIViewController {
             paymentMethod: type,
             previousCustomerInput: previousCustomerInput,
             offerSaveToLinkWhenSupported: offerSaveToLinkWhenSupported,
-            linkAccount: LinkAccountContext.shared.account,
-            cardBrandChoiceEligible: intent.cardBrandChoiceEligible
+            linkAccount: LinkAccountContext.shared.account
         ).make()
         formElement.delegate = self
         return formElement
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/AddPaymentMethodViewController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/AddPaymentMethodViewController.swift
@@ -55,33 +55,7 @@ class AddPaymentMethodViewController: UIViewController {
         return paymentMethodTypesView.selected
     }
     var paymentOption: PaymentOption? {
-        if let linkEnabledElement = paymentMethodFormElement as? LinkEnabledPaymentMethodElement {
-            return linkEnabledElement.makePaymentOption()
-        } else if let instantDebitsFormElement = paymentMethodFormElement as? InstantDebitsPaymentMethodElement {
-            // we use `.instantDebits` payment method locally to display a
-            // new button, but the actual payment method is `.link`, so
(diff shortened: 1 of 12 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/AddPaymentMethodViewController.swift
```diff
@@ -134,44 +107,27 @@ class AddPaymentMethodViewController: UIViewController {
 
     private let intent: Intent
     private let configuration: PaymentSheet.Configuration
+    private let isLinkEnabled: Bool
```
porter-stripe, line 110: Maybe naive: But can we read is link enabled from the intent?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentMethodFormViewController.swift
```diff
@@ -5,40 +5,142 @@
...
 
+protocol PaymentMethodFormViewControllerDelegate: AnyObject {
+    func didUpdate(_ viewController: PaymentMethodFormViewController)
+    func updateErrorLabel(for error: Error?)
+}
+
 class PaymentMethodFormViewController: UIViewController {
+    /// This caches forms for payment methods so that customers don't have to re-enter details
+    /// This class expects the formCache to be invalidated (cleared) when we load PaymentSheet; we assume the form generated for a given PM type _does not change_ at any point after load.
+    static var formCache: [PaymentSheet.PaymentMethodType: PaymentMethodElement] = [:]
```
porter-stripe, line 21: Struggling to see why this needs to be static

porter-stripe (approved):
looks good, nothing blocking!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/AddPaymentMethodViewController.swift
yuki-stripe: Great q, a lot of code assumed this incorrectly and I just fixed it! We can't; whether link is enabled on the Intent is not the same as whether it's enabled for us. e.g. we disable Link if the merchant requires billing address.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentMethodFormViewController.swift
yuki-stripe: We create instances of this class ~every time we go to the PM form, and the cache needs to live longer than that.

porter-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/AddPaymentMethodViewController.swift
porter-stripe: Makes sense, thanks!

porter-stripe (approved):

PR #3647, opened 2024-06-07
Detect strings that have overlapping ranges to avoid segfault

wooj-stripe:
## Summary
Detect if there are overlapping ranges, if so, return an undecorated string

## Motivation
Has a potential for causing a segfault

## Testing
Added tests
manually verified


porter-stripe (approved):
🚢

PR #3649, opened 2024-06-10
Remove link instant debits dead code

yuki-stripe:
## Summary
"link_instant_debits" isn't a real payment method type and there's only one way it can end up as a type in our code - when `configuration.linkPaymentMethodsOnly = true`, it's injected into PaymentSheet's payment method types list.

LinkPaymentController does this, but LinkPaymentController never actually invokes PaymentSheet or uses `linkInstantDebits` as a type, so all this code is dead.

## Motivation
Clean up code, reduce surface area for vertical mode.

## Testing
Tested LinkPaymentController manually.

## Changelog
Not user facing


yuki-stripe:
> 
> ### StripePayments
> ```diff
> - case linkInstantDebit
> ```
> 
> If you are adding a new public API consider the following:
> 
> * Do these APIs need to be `public` or can they be protected with `@_spi(STP)`?
> * If these APIs need to be `public`, assess whether they require an API review.
> 
> If you are modifying or removing a public API:
> 
> * Does this require a breaking version change?
> * Do these changes require API review?
> 
> If you confirm these APIs need to be added/updated and have undergone necessary review, add the label `modifies public API` to this PR to acknowledge and bypass this check.

This should never have been added to STPPaymentMethodType, since it's _not_ a payment method type.

wooj-stripe (approved):

PR #3643, opened 2024-06-06
Handle Apple Pay and Link taps in Wallet Header (vert mode)

porter-stripe:
## Summary
- Handles taps on Apple Pay and Link in the wallet header view to trigger the payment flows
- Improves the header view animation by moving the header view into the form
- Handles more edge cases w.r.t to the header view

### Link (complete)
https://github.com/stripe/stripe-ios/assets/88012362/af0c8481-8f56-41c2-b25e-93770e31ca11

### Link (flow controller)
https://github.com/stripe/stripe-ios/assets/88012362/4bb74868-beb4-45a0-8365-3c73414f6569

### Apple Pay (complete)
https://github.com/stripe/stripe-ios/assets/88012362/b2f6f703-d381-446f-af14-d33ad441d922

## Motivation
Vert mode

## Testing
- Manual
- Two new UI tests

## Changelog
N/A

PR #3651, opened 2024-06-10
Add primary button to vertical mode

yuki-stripe:
## Summary
Add primary button to vertical mode

## Testing
Manually tested; automated tests to follow!

## Changelog
Not user facing


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/VerticalPaymentMethodListViewController.swift
```diff
@@ -8,16 +8,8 @@
 @_spi(STP) import StripeUICore
 import UIKit
 
-protocol VerticalPaymentMethodListViewControllerDelegate: AnyObject {
-    /// - Returns: Whether or not the payment method row button should appear selected.
-    func didTapPaymentMethod(_ selection: VerticalPaymentMethodListSelection) -> Bool
-    /// Called when the accessory button on the saved payment method row is tapped
-    func didTapSavedPaymentMethodAccessoryButton()
-}
-
```
porter-stripe, line 17: 👍

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetVerticalViewController.swift
```diff
@@ -40,17 +41,24 @@ class PaymentSheetVerticalViewController: UIViewController, FlowControllerViewCo
                 return nil
             }
             return paymentMethodFormViewController.paymentOption
-
         }
     }
-    var selectedPaymentMethodType: PaymentSheet.PaymentMethodType?
+    var selectedPaymentMethodType: PaymentSheet.PaymentMethodType? {
```
porter-stripe, line 46: I think we need to cover the weird edge case w/ Link in the wallet?

porter-stripe: If so, can be done before/after https://github.com/stripe/stripe-ios/pull/3643 lands.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
```diff
@@ -49,6 +49,19 @@ extension PaymentSheet {
             }
         }
 
+        var paymentMethodType: PaymentMethodType? {
+            switch self {
+            case let .saved(paymentMethod: paymentMethod, _):
+                return .stripe(paymentMethod.type)
+            case let .new(confirmParams: intentConfirmParams):
+                return intentConfirmParams.paymentMethodType
+            case .applePay, .link:
```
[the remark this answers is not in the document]
yuki-stripe: Good question. It highlights the ambiguity of what this property is. If it means the type of the PaymentMethod object - e.g. the type of the PaymentMethod that ultimately gets created in the payment process - `.stripe(.link)` is not necessarily accurate, because Link has the "passthrough" mode where the PM type is the underlying pm (e.g. card) and not Link.

In practice that's not what this property is used for. It's for analytics and to determine if the type == sepa, in order to handle special mandate stuff, so returning `.stripe(.link)` is probably more useful for analytics.  I'll add comments to clarify.

porter-stripe (approved):

PR #3653, opened 2024-06-10
Payment Sheet: for FC and ID, added hosted_surface parameter, and improved parameter passing

kgaidis-stripe:
## Summary

^ this PR:
1) Adds a "hosted_surface" parameter so backend can better identify who is making the call on what surface
2) improves/fixes some parameter passing as part of this

## Testing

CI tests go through these flows, so if CI passes for this PR, we should be in a pretty good standing. Also ran some of these tests locally. 

Here is also an example of Instant Debits  where we had hard-coded logic, and we can see it appear here even though its not hard-coded:

Screenshot 2024-06-13 at 9 53 55 AM

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentMethodFormViewController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentMethodFormViewController.swift
@@ -288,11 +288,15 @@ extension PaymentMethodFormViewController {
                 self.delegate?.updateErrorLabel(for: genericError)
             }
         }
+        let additionalParameters: [String: Any] = [
+            "hosted_surface": "payment_element",
+        ]
         switch intent {
         case .paymentIntent(_, let paymentIntent):
             client.collectBankAccountForPayment(
                 clientSecret: paymentIntent.clientSecret,
                 returnURL: configuration.returnURL,
+                additionalParameters: additionalParameters,
                 onEvent: nil,
                 params: params,
                 from: viewController,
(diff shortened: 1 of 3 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentMethodFormViewController.swift
```diff
@@ -360,11 +366,17 @@ extension PaymentMethodFormViewController {
                 self.delegate?.updateErrorLabel(for: genericError)
             }
         }
+        let additionalParameters: [String: Any] = [
+            "product": "instant_debits",
+            "attach_required": true,
+            "hosted_surface": "payment_element",
+        ]
```
kgaidis-stripe, line 373: moved from here:

https://github.com/stripe/stripe-ios/pull/3653/files#diff-d34783ba39a9570594c629d85f6d940d3abc6532d8984d12611f9365031305d9L109-L114

kgaidis-stripe: this is much better implementation because the previous code was actually doing the wrong thing (but there luckily is no impact)

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentMethodFormViewController.swift
```diff
@@ -288,11 +288,15 @@ extension PaymentMethodFormViewController {
                 self.delegate?.updateErrorLabel(for: genericError)
             }
         }
+        let additionalParameters: [String: Any] = [
+            "hosted_surface": "payment_element",
+        ]
```
kgaidis-stripe, line 293: similar pattern/precedent of passing parameters already exists here:

https://github.com/stripe/stripe-ios/blob/e371135c/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkPaymentController.swift#L135-L138

ccen-stripe (approved):

PR #3655, opened 2024-06-11
Vertical mode errors

yuki-stripe:
## Summary
Adds error display to vertical mode

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-2209

## Testing
See snapshot tests. Also manually tested

https://github.com/stripe/stripe-ios/assets/47796191/0e9f0a51-7ac9-40f2-83f3-a0537315cd81

## Changelog
Not user facing

PR #3658, opened 2024-06-12
Fix lingering usage of `CallToActionType.add`

samer-stripe:
## Summary
Fix lingering usage of `CallToActionType.add`.

## Motivation
Fixes `master` branch.


samer-stripe:
Fixed in https://github.com/stripe/stripe-ios/pull/3655

PR #3660, opened 2024-06-12
Add `payment_method_layout ` to `mpe_config`

porter-stripe:
## Summary
- Adds `ui_mode` so we can track horizontal vs vertical
- Adds special case handling for US bank account header text

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-2207

## Testing
- Manual

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/Categories/String+Localized.swift
```diff
@@ -311,4 +311,11 @@ extension String.Localized {
             "Title shown above a view allowing the customer to save a card."
         )
     }
+
+    static var add_us_bank_account: String {
+        STPLocalizedString(
+            "Add US bank account",
+            "Title shown above a view allowing the customer to save a US bank account."
```
yuki-stripe, line 318:
```suggestion
            "Title shown above a view allowing the customer to add a US bank account."
```
Let's not use save, since the customer isn't necessarily saving their bank account.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/STPAnalyticsClient+PaymentSheet.swift
```diff
@@ -414,6 +414,8 @@ extension PaymentSheet.Configuration {
         payload["appearance"] = appearance.analyticPayload
         payload["billing_details_collection_configuration"] = billingDetailsCollectionConfiguration.analyticPayload
         payload["preferred_networks"] = preferredNetworks?.map({ STPCardBrandUtilities.apiValue(from: $0) }).joined(separator: ", ")
+        payload["ui_mode"] = paymentMethodLayout.description
```
yuki-stripe, line 417: Why UI mode?  I'd expect this to be called "payment_method_layout" mirroring the API

porter-stripe: Got it here, I can change to `payment_method_layout`.
CleanShot 2024-06-13 at 08 39 42

yuki-stripe: Ah I probably wrote that before we knew the API.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/FormHeaderView.swift
```diff
@@ -16,15 +16,23 @@ final class FormHeaderView: UIView {
         let label = PaymentSheetUI.makeHeaderLabel(appearance: appearance)
         if paymentMethodType == .stripe(.card) {
             label.text = hasASavedCard ? String.Localized.add_card : String.Localized.add_new_card
+        } else if paymentMethodType == .stripe(.USBankAccount) {
+            label.text = String.Localized.add_us_bank_account
```
yuki-stripe, line 20: Strange, this text seems to be bigger than `paymentMethodType.displayName`.  I think the icon height is constraining the text height.

yuki-stripe (dismissed):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/FormHeaderView.swift
porter-stripe: Seems to be tied to: 
```swift
    header.adjustsFontSizeToFitWidth = true
    header.adjustsFontForContentSizeCategory = true
```
when I comment these out the label is the correct size.

yuki-stripe (approved):

yuki-stripe (approved):

PR #3662, opened 2024-06-12
Add subtitle text for Afterpay and Link

porter-stripe:
## Summary
- Adds subtitle text for Link and Afterpay

## Motivation
- Vert mode

## Testing
- Manual
- See snapshots

## Changelog
N/A


Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
```diff
@@ -3862,27 +3862,26 @@ extension PaymentSheetUITestCase {
...
     func testPayingWithNoFormPMs_verticalmode() {
-        // We choose Cash App as a representative PM that does not require form details
+        // We choose Alipay as a representative PM that does not require form details
         var settings = PaymentSheetTestPlaygroundSettings.defaultValues()
         settings.mode = .payment
         settings.layout = .vertical
         loadPlayground(app, settings)
 
-        // Try Cash App
+        // Try Alipay
```
porter-stripe, line 3871: Switched to Alipay due to https://github.com/stripe/stripe-ios/pull/3661, otherwise we'd need to wait 15 seconds for polling to timeout.

yuki-stripe: Is that a new change? Cash App canceled immediately for me.

StripePaymentSheet/StripePaymentSheet/Source/Categories/String+Localized.swift
```diff
@@ -311,4 +311,18 @@ extension String.Localized {
             "Title shown above a view allowing the customer to save a card."
         )
     }
+
+    static var after_pay_subtitle_text: String {
+        STPLocalizedString(
+            "%d interest-free payments of %@",
```
yuki-stripe, line 317: Multiple parameters in a localized string need to include the position of the arguments, it should be like "%1$d interest-free payments of %2$d" so that languages can change the order of the parameters in their translation

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
porter-stripe: Yes new, CAP was mistakenly not addd to this list when we disabled refreshing for it.

yuki-stripe (dismissed):

PR #3689, opened 2024-06-21
Add button to autofill manual entry data while in test mode

mats-stripe:
## Summary

This adds a test mode autofill button / banner component, which will be shown on the manual entry and OTP entry screens while in test mode. For now, this adds the banner to the manual entry screen, and the changes to the OTP screen will come in a follow-up PR. 

First commit creates the new banner view, and the second commit integrates it into the Manual Entry view.

## Motivation

This will navigating through the flow much quicker while in test mode. Both web and Android have this implemented already!

## Testing

Verified the banner only appears while test mode is enabled:

https://github.com/stripe/stripe-ios/assets/172562065/04293860-2b05-424e-8bd2-4bac683e324a


StripeFinancialConnections/StripeFinancialConnections/Source/Common/TestModeAutofillBannerView.swift
```diff
@@ -0,0 +1,147 @@
...
+
+struct TestModeAutofillBannerView_Previews: PreviewProvider {
+    static var previews: some View {
+        VStack(spacing: 20) {
+            TestModeAutofillBannerViewRepresentable(bannerContext: .account)
+                .frame(height: 38)
+            
+            TestModeAutofillBannerViewRepresentable(bannerContext: .otp)
+                .frame(height: 38)
+        }
```
mats-stripe, line 143: image

StripeFinancialConnections/StripeFinancialConnections/Source/Native/ManualEntry/ManualEntryFormView.swift
```diff
@@ -82,15 +87,24 @@ final class ManualEntryFormView: UIView {
...
         let contentVerticalStackView = UIStackView(
             arrangedSubviews: [
                 textFieldStackView,
             ]
         )
+
+        if isTestMode {
+            let testModeBanner = TestModeAutofillBannerView(
+                context: .account,
+                didTapAutofill: applyTestModeValues
```
tillh-stripe, line 101: On Android and Web, we also submit the input on `Use test code` clicks instead of just filling it in. Should we do the same on iOS?

mats-stripe: Ouu good catch, yeah we could do that!

StripeFinancialConnections/StripeFinancialConnections/Source/Common/TestModeAutofillBannerView.swift
```diff
@@ -0,0 +1,147 @@
...
+    private let didTapAutofill: () -> Void
+
+    // MARK: - Subviews
+
+    private lazy var messageLabel: UILabel = {
+        // Create icon as a text attachement.
+        let icon = Image.info
+            .makeImage(template: true)
+            .withTintColor(.attention300)
+        let textAttachment = NSTextAttachment(image: icon)
```
kgaidis-stripe, line 44: I haven't done it this way, but if it works well it could be better (because then iOS handles the text wrapping)

The things to consider are:
- does this set the image to the right size?
- does it get us the correct spacing?
- how does it work with different text (ex. localization) and different font sizes (ex. dynamic font)? does the text wrap/scale correctly?

StripeFinancialConnections/StripeFinancialConnections/Source/Common/TestModeAutofillBannerView.swift
```diff
@@ -0,0 +1,147 @@
...
+            x: 0,
+            y: -2,
+            width: icon.size.width,
+            height: icon.size.height
+        )
+
+        let attributedString = NSMutableAttributedString()
+        attributedString.append(NSAttributedString(attachment: textAttachment))
+        attributedString.append(NSAttributedString(string: STPLocalizedString(
+            " You're in test mode.",
```
kgaidis-stripe, line 56: what do you think of putting the space separately from the localized string? (ex. maybe as simple as another `        attributedString.append(NSAttributedString(string: " ")`)

the practical fear I have is that localization teams could forget to put the space there, and then there's no spacing

the less-practical concern is just a separation of concerns: right now we are putting a space in the main text just to get a space (_but_ the space logically should be separate from text - it just "happens" that the text has a space because we added an icon)

StripeFinancialConnections/StripeFinancialConnections/Source/Common/TestModeAutofillBannerView.swift
kgaidis-stripe: I checked web and I think they just do an HStack (but, of course, we should do whatever is best for us as long as it handles everything equal/better):

Screenshot 2024-06-21 at 10 16 00 AM

StripeFinancialConnections/StripeFinancialConnections/Source/Common/TestModeAutofillBannerView.swift
```diff
@@ -0,0 +1,147 @@
...
+        autofillDataButton.setContentCompressionResistancePriority(.required, for: .horizontal)
+
+        stackView.addArrangedSubview(messageLabel)
+        stackView.addArrangedSubview(autofillDataButton)
+
+        stackView.translatesAutoresizingMaskIntoConstraints = false
+        addSubview(stackView)
+
+        NSLayoutConstraint.activate([
+            heightAnchor.constraint(equalToConstant: 38),
```
kgaidis-stripe, line 105: I think this might be an interesting test case on how this scales with dynamic font + potentially lengthy changes to text (or localization)

this is one reason why I often put everything in a UIStackView with padding because it automatically handles everything

StripeFinancialConnections/StripeFinancialConnections/Source/Native/ManualEntry/ManualEntryFormView.swift
```diff
@@ -82,15 +87,24 @@ final class ManualEntryFormView: UIView {
...
+            )
+            contentVerticalStackView.insertArrangedSubview(testModeBanner, at: 0)
```
kgaidis-stripe, line 103: _(this is completely subjective so feel free to ignore)_

one thing to consider is that now that we add multiple UIView's to the stack view, maybe it reads better like this (this is psuedocode):

```
var contentVerticalStackView = UIStackView()
if isTestMode {
   contentVerticalStackView.add(testModeBanner)
}
contentVerticalStackView.add(textFieldStackView)
```

the reason is just that one can read from top to bottom so it "flows better," while right now the `insertArrangedSubview(testModeBanner, at: 0)` is sort-of introducing a side-effect where the original UISTackView declaration gives the impression that `textFieldStackView` will be first:

```
let contentVerticalStackView = UIStackView(
    arrangedSubviews: [
         textFieldStackView,
    ]
)
```

...but the code later actually modifies this flow

StripeFinancialConnections/StripeFinancialConnections/Source/Common/TestModeAutofillBannerView.swift
mats-stripe: Here is the screen while using the largest dynamic font:

StripeFinancialConnections/StripeFinancialConnections/Source/Common/TestModeAutofillBannerView.swift
```diff
@@ -0,0 +1,155 @@
...
+//  StripeFinancialConnections
+//
+//  Created by Mat Schmid on 2024-06-20.
+//
+
+@_spi(STP) import StripeCore
+@_spi(STP) import StripeUICore
+import UIKit
+
+class TestModeAutofillBannerView: UIView {
```
kgaidis-stripe, line 12: good to have this abstracted in a separate class 🙏

StripeFinancialConnections/StripeFinancialConnections/Source/Common/TestModeAutofillBannerView.swift
```diff
@@ -0,0 +1,155 @@
...
+        let attributedString = NSMutableAttributedString()
+        attributedString.append(NSAttributedString(attachment: textAttachment))
+        // Add a spacer between the icon and the label in the form of a space.
+        attributedString.append(NSAttributedString(string: " "))
+        attributedString.append(NSAttributedString(string: STPLocalizedString(
+            "You're in test mode.",
+            "Message shown to a user who is in test mode, which will give them the option to autofill mock credentials."
+        )))
+
+        let label = UILabel()
```
kgaidis-stripe, line 62: it shouldn't matter too much here because its a single line text wrapped with padding, but generally prefer `AttributedLabel` because that supports line heights together with `FinancialConnectionsFont` (`UILabel` usage in Financial Connections is likely very small if one searches it)

kgaidis-stripe (approved):
thank you! and great that the test plan video shows both live mode and test mode 🤝 

---

Also extra info for learning:

If you type `bitrise workflows` in Terminal you'll see various options

one of them is:

```
⚡️ financial-connections-stability-tests
  Title: 
  Summary: 
  Run with: bitrise run financial-connections-stability-tests
  ```
  
So running `bitrise run financial-connections-stability-tests` will test the example app usage, and Financial Connections code quite throughly with UI tests. It also right now runs twice a day, but does not run on each PR (we might want to change this in the future).

kgaidis-stripe (approved):

PR #3692, opened 2024-06-24
FC Playground - Allow copying output session info

mats-stripe:
## Summary

This allows copying session info for a Financial Connections flow once it is completed in the playground app.

https://github.com/stripe/stripe-ios/assets/172562065/11b15f99-c7c1-4fd8-a0b7-8cbf9a73b1d0

## Motivation

This will make it easier to diagnose issues with the flow by making IDs and account details easy to export from the playground app.

## Testing

Manually tested the changes. See video posted above!

## Changelog

 N/a

### DIFF
--- a/Example/FinancialConnections Example/FinancialConnections Example/Playground/PlaygroundView.swift
+++ b/Example/FinancialConnections Example/FinancialConnections Example/Playground/PlaygroundView.swift
@@ -124,7 +124,7 @@ struct PlaygroundView: View {
                     }
 
                     Section(header: Text("Session output")) {
-                        if let output = viewModel.outputTextfieldText {
+                        if let output = viewModel.sessionOutput[.message] {
                             TextEditor(text: .constant(output))
                                 .accessibility(identifier: "playground-session-output-textfield")
                         }
@@ -132,19 +132,19 @@ struct PlaygroundView: View {
                         Button(action: viewModel.copySessionId) {
                             Text("Copy Session ID")
                         }
-                        .disabled(viewModel.outputSessionId == nil)
+                        .disabled(viewModel.sessionOutput[.sessionId] == nil)
                         .accessibility(identifier: "playground-session-output-copy-session-id")
 
                         Button(action: viewModel.copyAccountNames) {
                             Text("Copy Account Names")
                         }
-                        .disabled(viewModel.outputAccountName == nil)
+                        .disabled(viewModel.sessionOutput[.accountNames] == nil)
                         .accessibility(identifier: "playground-session-output-copy-account-names")
 
                         Button(action: viewModel.copyAccountIds) {
                             Text("Copy Account IDs")
                         }
-                        .disabled(viewModel.outputAccountIds == nil)
+                        .disabled(viewModel.sessionOutput[.accountIds] == nil)
                         .accessibility(identifier: "playground-session-output-copy-account-ids")
                     }
                 }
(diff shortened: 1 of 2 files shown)


kgaidis-stripe:
@mats-stripe 

I'm good to approve as-is, thank you! (lets just remove print statements)

---

If I had to add some commentary its (and feel free to ignore all of it!):

1) Again going with the theme of what is the simplest possible solution, I would have first just put a text box at the bottom and then only add extra stuff (ex. buttons like copy session, copy account names, copy ) through experience. So, after adding a simple text box, IF it proves to be supremely cumbersome to copy stuff from the output manually, then I'd add the extra buttons. Think "building an MVP and then seeing where the user takes the product."

2) Again going with the theme of what is the simplest possible solution (eliminating variables, state, etc.), if it did prove to be cumbersome, or I wanted buttons regardless for other reasons, to avoid extra logic scattered in multiple places (particularly the extra instance variables), I probably would have just put the output in a JSON, and then each "copy function" extracts the right info from the JSON. For example (and this is just badly named psuedocode):

```
    func copyAccountIds() {
        UIPasteboard.general.string = Extract(key: "account_ids", fromJSONString: outputTextfieldText)
    }
```

Both my suggestions can 100% be considered worse choices, and are subjective, but just sharing a different perspective.

kgaidis-stripe:
@mats-stripe 

Also extra info for learning:

If you type `bitrise workflows` in Terminal you'll see various options

one of them is:

```
⚡️ financial-connections-stability-tests
  Title: 
  Summary: 
  Run with: bitrise run financial-connections-stability-tests
  ```
  
So running `bitrise run financial-connections-stability-tests` will test the example app usage, and Financial Connections code quite throughly with UI tests. It also right now runs twice a day, but does not run on each PR (we might want to change this in the future). 

In this case, I doubt that its going to cause problems with UI tests, but sometimes its a nice sanity check as if the UI tests aren't built very robustly, sometimes moving some pixels around can cause them to fail.

**Comment by mats-stripe** [comment]
>Again going with the theme of what is the simplest possible solution, I would have first just put a text box at the bottom and then only add extra stuff (ex. buttons like copy session, copy account names, copy ) through experience. So, after adding a simple text box, IF it proves to be supremely cumbersome to copy stuff from the output manually, then I'd add the extra buttons. Think "building an MVP and then seeing where the user takes the product."

This is what I had originally, and I found it tricky to select a specific piece of text from the textbox. For instance, if you want to copy the session ID, double tapping the text doesn't select the whole ID: 

image

I found having these buttons made it much easier to use, but happy to change it up if you'd prefer the simplicity!

>Again going with the theme of what is the simplest possible solution (eliminating variables, state, etc.), if it did prove to be cumbersome, or I wanted buttons regardless for other reasons, to avoid extra logic scattered in multiple places (particularly the extra instance variables), I probably would have just put the output in a JSON, and then each "copy function" extracts the right info from the JSON. For example (and this is just badly named psuedocode): ...

I like the idea! Though JSON isn't _super_ easy to go from data -> JSON, and then back from JSON -> data, I might switch this up to use a dictionary! Something like this might be nice:

```swift
// Property:
@Published var sessionOutput: [String: String] = [:]

// Setting:
sessionOutput["session_id"] = session.id

// Getting:
func copySessionId() {
    UIPasteboard.general.string = sessionOutput["session_id"]
}
```

kgaidis-stripe (approved):
🎉 🙏

PR #3693, opened 2024-06-24
Track intent usage in PaymentSheet

porter-stripe:
## Summary
- Track what intent is loaded for a given PaymentSheet session

## Motivation
- Remediation

## Testing
- Manual

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift
@@ -95,7 +95,8 @@ final class PaymentSheetLoader {
                 analyticsClient.logPaymentSheetLoadSucceeded(
                     loadingStartDate: loadingStartDate,
                     linkEnabled: isLinkEnabled,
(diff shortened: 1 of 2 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift
```diff
@@ -95,7 +95,8 @@ final class PaymentSheetLoader {
                 analyticsClient.logPaymentSheetLoadSucceeded(
                     loadingStartDate: loadingStartDate,
                     linkEnabled: isLinkEnabled,
-                    defaultPaymentMethod: paymentOptionsViewModels.stp_boundSafeObject(at: defaultSelectedIndex)
+                    defaultPaymentMethod: paymentOptionsViewModels.stp_boundSafeObject(at: defaultSelectedIndex),
+                    intent: intent
```
wooj-stripe, line 99: nit: Feels a little awkward to send the entire intent, cause I'm worried we may end up logging more than we should-- maybe we change the signature of `logPaymentSheetLoadSucceeded` to accept:
`intentAnalyticsValue: String` instead?

wooj-stripe (approved):

PR #3705, opened 2024-06-25
Poll for Swish

porter-stripe:
## Summary
- Roughly 60% of the time the Swish API returns the intent in the `requires_action` state instead of the most up to date state. We start polling to account for this delay.
- To combat this we poll every second for five seconds.

## Motivation
- ir-copper-deep

## Testing
- Manual

## Changelog
See diff

### DIFF
--- a/StripePayments/StripePayments/Source/API Bindings/Models/PaymentMethods/STPPaymentMethodEnums.swift
+++ b/StripePayments/StripePayments/Source/API Bindings/Models/PaymentMethods/STPPaymentMethodEnums.swift
@@ -288,6 +288,17 @@ extension STPPaymentMethodType {
         }
     }
 
+    var pollingInterval: TimeInterval {
+        switch self {
+        case .amazonPay, .cashApp:
+            return 3
+        case .swish:
+            return 1
+        default:
+            return 0
+        }
+    }
+
     var supportsRefreshing: Bool {
         switch self {
         // Payment methods such as CashApp implement app-to-app redirects that bypass the "redirect trampoline" too give a more seamless user experience for app-to-app.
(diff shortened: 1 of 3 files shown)


StripePayments/StripePayments/Source/API Bindings/Models/PaymentMethods/STPPaymentMethodEnums.swift
```diff
@@ -281,13 +281,24 @@ extension STPPaymentMethodType {
...
         // However, when returning to the merchant app in this scenario, the intent often isn't updated instantaneously, requiring polling for intent status updates.
-        case .amazonPay, .cashApp:
+        case .amazonPay, .cashApp, .swish:
             return true
         default:
             return false
         }
     }
 
+    var pollingInterval: TimeInterval {
```
yuki-stripe, line 291: Ideally we shouldn't have `pollingInterval` if `requiresPolling` is false. Otherwise it introduces bad states like a PM that requires polling but has a polling interval of 0, etc.

How about 
```
extension STPPaymentMethodType {
  struct PollingRequirement {
    var pollingInterval: TimeInterval
  }
  /// If non-nil, Intents with this PM type do not update immediately after the next action is handled and require us to poll and this property contains the information needed to poll.
  var pollingRequirement: PollingRequirement? {
     switch self { ... }
  }
}
```

porter-stripe: Ah great suggestion. I've added this. It's a little weird w/ cards since there is an additional condition that determines if polling is required (web-based 3DS2), lmk what you think!

yuki-stripe (approved):

PR #3707, opened 2024-06-25
Add autofill to OTP field while in test mode

mats-stripe:
## Summary

This adds the test mode autofill banner (introduced here: https://github.com/stripe/stripe-ios/pull/3689) to OTP screens while in test mode.

## Motivation

This will make navigating through the networking flow much quicker while in test mode. Both web and Android have this implemented already!

## Testing

Added a UI test, and did some manual testing to ensure the autofill button is only shown while in test mode:

https://github.com/stripe/stripe-ios/assets/172562065/a16a1430-bff9-403b-a2e2-dfe9c04a2f72

## Changelog

N/a

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/Native/NetworkingLinkStepUpVerification/NetworkingLinkStepUpVerificationDataSource.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/Native/NetworkingLinkStepUpVerification/NetworkingLinkStepUpVerificationDataSource.swift
@@ -51,7 +51,7 @@ final class NetworkingLinkStepUpVerificationDataSourceImplementation: Networking
             apiClient: apiClient,
             clientSecret: clientSecret,
             analyticsClient: analyticsClient,
-            isTestmode: manifest.isTestMode
+            isTestMode: manifest.isTestMode
         )
         self.networkingOTPDataSource = networkingOTPDataSource
         networkingOTPDataSource.delegate = self
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/Native/NetworkingLinkStepUpVerification/NetworkingLinkStepUpVerificationDataSource.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/Native/NetworkingLinkStepUpVerification/NetworkingLinkStepUpVerificationDataSource.swift
@@ -50,7 +50,8 @@ final class NetworkingLinkStepUpVerificationDataSourceImplementation: Networking
             consumerSession: nil,
             apiClient: apiClient,
             clientSecret: clientSecret,
-            analyticsClient: analyticsClient
+            analyticsClient: analyticsClient,
+            isTestmode: manifest.isTestMode
         )
         self.networkingOTPDataSource = networkingOTPDataSource
         networkingOTPDataSource.delegate = self
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/Native/NetworkingLinkVerification/NetworkingLinkVerificationDataSource.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/Native/NetworkingLinkVerification/NetworkingLinkVerificationDataSource.swift
@@ -51,7 +51,8 @@ final class NetworkingLinkVerificationDataSourceImplementation: NetworkingLinkVe
             consumerSession: nil,
             apiClient: apiClient,
             clientSecret: clientSecret,
-            analyticsClient: analyticsClient
+            analyticsClient: analyticsClient,
+            isTestmode: manifest.isTestMode
         )
         self.networkingOTPDataSource = networkingOTPDataSource
         networkingOTPDataSource.delegate = self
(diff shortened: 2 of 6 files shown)


StripeFinancialConnections/StripeFinancialConnections/Source/Native/NetworkingLinkStepUpVerification/NetworkingLinkStepUpVerificationDataSource.swift
```diff
@@ -50,7 +50,8 @@ final class NetworkingLinkStepUpVerificationDataSourceImplementation: Networking
             consumerSession: nil,
             apiClient: apiClient,
             clientSecret: clientSecret,
-            analyticsClient: analyticsClient
+            analyticsClient: analyticsClient,
+            isTestmode: manifest.isTestMode
```
kgaidis-stripe, line 54:
```
isTestmode: manifest.isTestMode
```

one uses lowercase "m" (isTestmode) the other uses upper case "M" (manifest.isTestMode) - is that expected?

StripeFinancialConnections/StripeFinancialConnections/Source/Native/NetworkingLinkVerification/NetworkingLinkVerificationDataSource.swift
```diff
@@ -51,7 +51,8 @@ final class NetworkingLinkVerificationDataSourceImplementation: NetworkingLinkVe
             consumerSession: nil,
             apiClient: apiClient,
             clientSecret: clientSecret,
-            analyticsClient: analyticsClient
+            analyticsClient: analyticsClient,
+            isTestmode: manifest.isTestMode
```
kgaidis-stripe, line 55:
```
isTestmode: manifest.isTestMode
```

one uses lowercase "m" (isTestmode) the other uses upper case "M" (manifest.isTestMode) - is that expected?

StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/NetworkingOTPView/NetworkingOTPView.swift
```diff
@@ -30,15 +30,26 @@ protocol NetworkingOTPViewDelegate: AnyObject {
...
-            arrangedSubviews: [
-                otpTextField,
-            ]
-        )
+        let otpVerticalStackView = UIStackView()
+
+        if dataSource.isTestMode {
+            let testModeBanner = TestModeAutofillBannerView(
+                context: .otp,
+                didTapAutofill: applyTestModeValue
```
kgaidis-stripe, line 46: subjective but you could consider in-lining `applyTestModeValue` because its not referenced anywhere else

the thought pattern is that if a function is not used anywhere else but one place, then the simplest solution is to inline it - this theoretically reduces complexity because if one reads `NetworkingOTPView` they don't have to analyze whether `applyTestModeValue` is called multiple times in multiple places

kgaidis-stripe (approved):
pretty cool that it was easy to reuse

StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/NetworkingOTPView/NetworkingOTPView.swift
```diff
@@ -30,15 +30,26 @@ protocol NetworkingOTPViewDelegate: AnyObject {
 
 final class NetworkingOTPView: UIView {
 
+    enum TestModeValues {
+        static let otp = "000000"
```
kgaidis-stripe, line 34: subjective: because this is only used in one place, you could inline it too, but I break this rule sometimes too

the thought process is that its simpler to inline (ex. if both this solution, and inline solution, can be equally considered valid, then one could say that "in this battle of subjectivity" [the simplest solution is the winner](https://en.wikipedia.org/wiki/Occam%27s_razor) haha)

StripeFinancialConnections/StripeFinancialConnections/Source/Native/NetworkingLinkStepUpVerification/NetworkingLinkStepUpVerificationDataSource.swift
mats-stripe: Whoops - good catch!

kgaidis-stripe (approved):

StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/NetworkingOTPView/NetworkingOTPView.swift
mats-stripe: I know it's only used once, but I'm in favour of keeping this as is since in encapsulates a bit of the "business" logic away from the view logic. No strong opinion here though!

StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/NetworkingOTPView/NetworkingOTPView.swift
mats-stripe: I'm fine with either approach, but I will leave this as is for now since this is the same pattern that was used for the Manual Entry flow

PR #3711, opened 2024-06-25
Fix vertical mode accessibility of row button accessory

yuki-stripe:
## Testing
Manual

## Changelog
Not user facing


porter-stripe (approved):

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetVerticalUITest.swift
```diff
@@ -0,0 +1,255 @@
+//
+//  PaymentSheetVerticalUITest.swift
```
yuki-stripe, line 2: Please ignore the "Fix vertical mode tests subclassing.." commit - that just moves the vertical mode test methods out into this file but it blows up the diff.

porter-stripe (approved):

PR #3716, opened 2024-06-26
Adjust allow_redisplay values when using CustomerSession

wooj-stripe:
## Summary
Adjust allow_redisplay values when using CustomerSession

## Motivation
When not saving a payment method to a customer, we were specifying limited. This should have been unspecified

## Testing
Added unit tests

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/STPElementsSession.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/STPElementsSession.swift
@@ -211,15 +211,20 @@ extension STPElementsSession {
 
 extension STPElementsSession {
     func savePaymentMethodConsentBehavior() -> PaymentSheetFormFactory.SavePaymentMethodConsentBehavior {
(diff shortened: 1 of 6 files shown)


StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/STPElementsSession.swift
```diff
@@ -211,20 +211,20 @@ extension STPElementsSession {
...
-        } else {
+        guard let paymentMethodSave = paymentSheetPaymentMethodSave() else {
             return .legacy
         }
+        return paymentMethodSave
+        ? .paymentSheetWithCustomerSessionPaymentMethodSaveEnabled
+        : .paymentSheetWithCustomerSessionPaymentMethodSaveDisabled
+
+    }
+    func paymentSheetPaymentMethodSave() -> Bool? {
```
davidme-stripe, line 222: This name is hard to understand, maybe something like `var canSavePaymentMethod` with a comment about the default legacy behavior if it's nil? Possible I'm misunderstanding this too

davidme-stripe (approved):

PR #3720, opened 2024-06-27
Reanimate FC consent dots view on app foreground

mats-stripe:
## Summary

This fixes an issue where the consent logo view dots animation would be stopped and not restarted when the app was backgrounded, then returned to the foreground.

## Motivation

✨

## Testing

Verified the animation properly continues after the app is foregrounded:

https://github.com/stripe/stripe-ios/assets/172562065/6ecba189-f29a-4cde-9f1c-19bd95742713

## Changelog

N/a

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/Native/Consent/ConsentViewController.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/Native/Consent/ConsentViewController.swift
@@ -67,10 +67,6 @@ class ConsentViewController: UIViewController {
         fatalError("init(coder:) has not been implemented")
     }
 
-    deinit {
-        NotificationCenter.default.removeObserver(self)
-    }
-
     override func viewDidLoad() {
         super.viewDidLoad()
         view.backgroundColor = .customBackgroundColor
@@ -108,13 +104,6 @@ class ConsentViewController: UIViewController {
         paneLayoutView.addTo(view: view)
 
         dataSource.analyticsClient.logPaneLoaded(pane: .consent)
-    }
-
-    override func viewWillAppear(_ animated: Bool) {
-        super.viewWillAppear(animated)
-        // this fixes an issue where presenting a UIViewController
-        // on top of ConsentViewController would stop the dot animation
-        consentLogoView?.animateDots()
 
         NotificationCenter.default.addObserver(
             self,
@@ -122,6 +111,13 @@ class ConsentViewController: UIViewController {
             name: UIApplication.willEnterForegroundNotification,
             object: nil
         )
+    }
+
+    override func viewWillAppear(_ animated: Bool) {
+        super.viewWillAppear(animated)
+        // this fixes an issue where presenting a UIViewController
+        // on top of ConsentViewController would stop the dot animation
+        consentLogoView?.animateDots()
     }
 
     @objc private func appWillEnterForeground() {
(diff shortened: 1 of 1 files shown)


StripeFinancialConnections/StripeFinancialConnections/Source/Native/Consent/ConsentViewController.swift
```diff
@@ -67,6 +67,10 @@ class ConsentViewController: UIViewController {
         fatalError("init(coder:) has not been implemented")
     }
 
+    deinit {
+        NotificationCenter.default.removeObserver(self)
```
kgaidis-stripe, line 71: @mats-stripe 

Do you have to do this anymore? (for the selector apI)

https://developer.apple.com/documentation/foundation/nsnotificationcenter/1415360-addobserver

> If your app targets iOS 9.0 and later or macOS 10.11 and later, you do not need to unregister an observer that you created with this function. If you forget or are unable to remove an observer, the system cleans up the next time it would have posted to it.

StripeFinancialConnections/StripeFinancialConnections/Source/Native/Consent/ConsentViewController.swift
```diff
@@ -111,6 +115,19 @@ class ConsentViewController: UIViewController {
         // this fixes an issue where presenting a UIViewController
         // on top of ConsentViewController would stop the dot animation
         consentLogoView?.animateDots()
+
+        NotificationCenter.default.addObserver(
```
kgaidis-stripe, line 119: this likely doesn't matter too much, but wouldn't viewWillAppear potentially get called multiple times, so a potentially more appropriate place to "subscribe only once" is viewDidLoad (or something of the like)?

StripeFinancialConnections/StripeFinancialConnections/Source/Native/Consent/ConsentViewController.swift
```diff
@@ -111,6 +115,19 @@ class ConsentViewController: UIViewController {
...
+            self,
+            selector: #selector(appWillEnterForeground),
+            name: UIApplication.willEnterForegroundNotification,
+            object: nil
+        )
+    }
+
+    @objc private func appWillEnterForeground() {
```
kgaidis-stripe, line 127: I like the straightforward naming!  UIApplication.willEnterForegroundNotification > appWillEnterForeground is easy to understand (vs. naming it something more-specific like "reactiveDots" or else)

kgaidis-stripe (dismissed):

kgaidis-stripe:
marking this as done: https://jira.corp.stripe.com/browse/BANKCON-10932

StripeFinancialConnections/StripeFinancialConnections/Source/Native/Consent/ConsentViewController.swift
mats-stripe: TIL! Thanks for pointing this out, will remove this!

StripeFinancialConnections/StripeFinancialConnections/Source/Native/Consent/ConsentViewController.swift
mats-stripe: Good point! I'll move this

kgaidis-stripe (approved):

PR #3723, opened 2024-06-27
Update "US Bank Account" PaymentSheet label to "US bank account"

yuki-stripe:
# ⚠️ Will need to wait until translations are in, since this is string is used in production.

## Motivation
Match web PE


porter-stripe:
@yuki-stripe Changes look good, might need to get latest master for some test fixes it appears.

porter-stripe (approved):

PR #3731, opened 2024-06-28
Fixed an issue where certain cobranded cards showed a generic card ic…

yuki-stripe:
…on instead of using the other card brand

## Summary
We only use "preferred" card brand for icon. If it's unknown, we should fall back to the brand. This appears to match what Stripe.js does. 
Example: `4000050360000001` preferred brand is eftpos_australia and brand is visa. We show an unknown icon but stripe.js shows a visa icon.

## Motivation
https://docs.google.com/document/d/1tRLvYH4_UzRGevJuP15gU4SlU5zt9_pJBZtiOSMxy6A/edit search "AU Saved Payment Icon"

## Testing
Manually tested, added unit test.

## Changelog
See changelog


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentOption+Images.swift
```diff
@@ -55,14 +55,20 @@ extension PaymentOption {
 }
 
 extension STPPaymentMethod {
+    /// Returns the first non-unknown card brand, prioritizing the card's preferred network brand > display brand > brand
+    func calculateCardBrandToDisplay() -> STPCardBrand {
```
porter-stripe, line 59: ah clever!

porter-stripe (approved):

porter-stripe (approved):

PR #3739, opened 2024-07-01
Fix deleting payment method in vertical mode with CustomerSessions

porter-stripe:
## Summary
- We were not correctly reading the eph key from CustomSessions

## Motivation
- Bug bash

## Testing
- Manual
- Couldn't think of a better way to test this than a whole another E2E test with just CustomerSessions.

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/Vertical Saved Payment Method Screen/VerticalSavedPaymentMethodsViewController.swift
```diff
@@ -178,8 +181,7 @@ class VerticalSavedPaymentMethodsViewController: UIViewController {
     }
 
     private func remove(paymentMethod: STPPaymentMethod) {
-        guard let button = paymentMethodRows.first(where: { $0.paymentMethod.stripeId == paymentMethod.stripeId }),
-                let ephemeralKeySecret = configuration.customer?.ephemeralKeySecret else { return }
```
yuki-stripe, line 182: Not blocking: why doesn't SavedPaymentMethodManager do the work of getting the ephemeral key?  Maybe PSVerticalVC should create the Manager with the intent and configuration, let it be responsible for deriving the ephemeral key, and inject it into VerticalSavedPMsVC.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetVerticalViewController.swift
```diff
@@ -509,12 +509,17 @@ class PaymentSheetVerticalViewController: UIViewController, FlowControllerViewCo
             return
         }
 
+        guard let ephemeralKeySecret = configuration.customer?.ephemeralKeySecretBasedOn(intent: intent) else {
+            stpAssert(true, "Failed to read ephemeral key")
```
yuki-stripe, line 513: Should we send an unexpected error analytic?

StripePaymentSheet/StripePaymentSheet/Source/Helpers/SavedPaymentMethodManager.swift
```diff
@@ -6,26 +6,52 @@
...
 final class SavedPaymentMethodManager {
 
+    enum Error: Swift.Error {
+        case missingEphemeralKey
+    }
+
     let configuration: PaymentSheet.Configuration
+    let intent: Intent
 
-    init(configuration: PaymentSheet.Configuration) {
```
porter-stripe, line 16: I considered making this init throws if we can't find a valid eph key, but didn't love the ergonomics of it.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetVerticalViewController.swift
porter-stripe: Yeah probably, added one with some metadata.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/Vertical Saved Payment Method Screen/VerticalSavedPaymentMethodsViewController.swift
porter-stripe: Yeah that makes it a bit more testable, refactored so this is the case.

StripePaymentSheet/StripePaymentSheet/Source/Helpers/SavedPaymentMethodManager.swift
```diff
@@ -6,26 +6,52 @@
...
+    private var ephemeralKey: String? {
+        guard let ephemeralKey = configuration.customer?.ephemeralKeySecretBasedOn(intent: intent) else {
+            stpAssert(true, "Failed to read ephemeral key.")
+            let errorAnalytic = ErrorAnalytic(event: .unexpectedPaymentSheetError,
+                                              error: Error.missingEphemeralKey,
+                                              additionalNonPIIParams: ["customer_access_provider": configuration.customer?.customerAccessProvider.analyticValue as Any])
+            STPAnalyticsClient.sharedClient.log(analytic: errorAnalytic)
+            return nil
+        }
+        return ephemeralKey
```
yuki-stripe, line 31: nit: Why compute this?  Seems like it should be initialized once and send the error once?

StripePaymentSheet/StripePaymentSheet/Source/Helpers/SavedPaymentMethodManager.swift
```diff
@@ -6,26 +6,52 @@
...
+    init(configuration: PaymentSheet.Configuration, intent: Intent) {
         self.configuration = configuration
+        self.intent = intent
     }
 
     func update(paymentMethod: STPPaymentMethod,
-                with updateParams: STPPaymentMethodUpdateParams,
-                using ephemeralKey: String) async throws -> STPPaymentMethod {
+                with updateParams: STPPaymentMethodUpdateParams) async throws -> STPPaymentMethod {
+        guard let ephemeralKey = ephemeralKey else {
```
yuki-stripe, line 41:
```suggestion
        guard let ephemeralKey else {
```

StripePaymentSheet/StripePaymentSheet/Source/Helpers/SavedPaymentMethodManager.swift
```diff
@@ -6,26 +6,52 @@
...
+        }
+
         return try await configuration.apiClient.updatePaymentMethod(with: paymentMethod.stripeId,
                                                                      paymentMethodUpdateParams: updateParams,
                                                                      ephemeralKeySecret: ephemeralKey)
     }
 
-    func detach(paymentMethod: STPPaymentMethod, using ephemeralKey: String) {
+    func detach(paymentMethod: STPPaymentMethod) {
+        guard let ephemeralKey = ephemeralKey else {
```
yuki-stripe, line 51:
```suggestion
        guard let ephemeralKey else {
```

StripePaymentSheet/StripePaymentSheet/Source/Helpers/SavedPaymentMethodManager.swift
```diff
@@ -6,26 +6,52 @@
...
+
     let configuration: PaymentSheet.Configuration
+    let intent: Intent
+
+    private var ephemeralKey: String? {
+        guard let ephemeralKey = configuration.customer?.ephemeralKeySecretBasedOn(intent: intent) else {
+            stpAssert(true, "Failed to read ephemeral key.")
+            let errorAnalytic = ErrorAnalytic(event: .unexpectedPaymentSheetError,
+                                              error: Error.missingEphemeralKey,
+                                              additionalNonPIIParams: ["customer_access_provider": configuration.customer?.customerAccessProvider.analyticValue as Any])
```
yuki-stripe, line 27:
```suggestion
                                              additionalNonPIIParams: ["customer_access_provider": configuration.customer?.customerAccessProvider.analyticValue as String])
```

nit: Does that not work?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/Vertical Saved Payment Method Screen/VerticalSavedPaymentMethodsViewController.swift
```diff
@@ -28,12 +28,18 @@ class VerticalSavedPaymentMethodsViewController: UIViewController {
...
 
     private var updateViewController: UpdateCardViewController?
 
+    private var paymentMethodRemove: Bool {
+        intent.elementsSession.allowsRemovalOfPaymentMethodsForPaymentSheet()
+    }
+
+    private var isCBCEligible: Bool {
+        intent.cardBrandChoiceEligible
+    }
```
yuki-stripe, line 41: nit: These aren't really computed either, are they?  Should they be `let`s? Computed vars implies the value can change, which is harder to reason about compared to `let`.

StripePaymentSheet/StripePaymentSheet/Source/Helpers/SavedPaymentMethodManager.swift
porter-stripe: It doesn't since it's optional, we could unwrap it instead I suppose.
CleanShot 2024-07-03 at 13 52 29

yuki-stripe: `as String?`? Or provide a default value?  I'm just slightly wary of `Any`

porter-stripe: Yeah similar warning, I'll provide a default value.
CleanShot 2024-07-03 at 14 01 16

yuki-stripe (approved):

yuki-stripe (approved):

yuki-stripe (approved):

PR #3741, opened 2024-07-02
Notify delegate we should dismiss

porter-stripe:
## Summary
- We don't reopen to the manage screen in this case, we could try to track the state and re-open it to the manage screen but not sure it's totally worth it for this behavior.

https://github.com/stripe/stripe-ios/assets/88012362/68aebc52-a67f-4e77-8715-1e2ecf52596a

## Motivation
🐛 bash

## Testing
- Manual

## Changelog
N/A


yuki-stripe:
> We don't reopen to the manage screen in this case, we could try to track the state and re-open it to the manage screen but not sure it's totally worth it for this behavior.

PaymentSheet always starts fresh when you present it.  FlowController should re-open to the manage screen for free.

yuki-stripe (approved):

PR #3744, opened 2024-07-02
Show mandate if being reused

porter-stripe:
## Summary
- For saved payment methods like SEPA that show a mandate, we were not showing them. We need to take into account if the payment method is being re-used as well.
- We need a separate fix for US bank account since the mandate isn't technically a part of the form

### Horizontal for reference
Simulator Screenshot - iPhone 15 - 2024-07-02 at 12 05 19

### After fix
https://github.com/stripe/stripe-ios/assets/88012362/df522959-2ca6-4791-adb2-2876f71b45f4

## Motivation
🐛 

## Testing
- Manual

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/VerticalMandateView.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/VerticalMandateView.swift
@@ -46,12 +46,12 @@ final class VerticalMandateView: UIView {
         guard let form = formProvider(paymentMethodType) else {
             return
         }
-        
+
         if form.collectsUserInput && !state.isReusing {
             // If it collects user input and is not being re-used, the mandate will be displayed in the form and not here
             return
         }
-        
+
         // Get the mandate from the form, if available
         // 🙋‍♂️ Note: assumes mandates are SimpleMandateElement!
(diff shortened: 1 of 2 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/VerticalMandateView.swift
```diff
@@ -34,18 +38,20 @@ final class VerticalMandateView: UIView {
...
-        else {
-            // If it collects user input, the mandate will be displayed in the form and not here
+        guard let form = formProvider(paymentMethodType) else {
             return
         }
+
+        if form.collectsUserInput && !state.isReusing {
+            // If it collects user input and is not being re-used, the mandate will be displayed in the form and not here
+            return
+        }
```
yuki-stripe, line 53: What does isReusing mean? Does it mean like, whether or not we have a saved payment method selected? Maybe rename `isSavedPaymentMethod`?

IMO we should branch based on if it's a saved PM or not. If it's not, trigger the old form-based code. If it is, switch on the payment method type.  The "generate form" logic just doesn't hold up for saved PMs. 

```
if isSavedPaymentMethod {
  switch paymentMethodType {
   ... manually build mandates
  }
} else {
  ... original form-generating code
}
```

yuki-stripe (commented):
Can we add tests for this, too? We can expand `testDisplaysMandateBelowList`.

porter-stripe:
Closing from our Slack convo @yuki-stripe

PR #3746, opened 2024-07-03
FC Playground app - Make keyboard dismissible

mats-stripe:
## Summary

This updates the Financial Connections playground app such that the keyboard can now be dismissed interactively. Previously, there was no way to dismiss it, and it would block launching the FC flow.

## Motivation

✨

## Testing

https://github.com/stripe/stripe-ios/assets/172562065/24ad5f89-4ebd-4912-8488-f7d65730ff7a

## Changelog

N/a

### DIFF
--- a/Example/FinancialConnections Example/FinancialConnections Example/Playground/PlaygroundView.swift
+++ b/Example/FinancialConnections Example/FinancialConnections Example/Playground/PlaygroundView.swift
@@ -176,6 +176,9 @@ struct PlaygroundView: View {
         }
         .navigationTitle("Playground")
         .navigationBarTitleDisplayMode(.inline)
+        .onAppear {
+            UIScrollView.appearance().keyboardDismissMode = .interactive
+        }
     }
 }
 
--- a/Example/FinancialConnections Example/FinancialConnections Example/Playground/PlaygroundView.swift
+++ b/Example/FinancialConnections Example/FinancialConnections Example/Playground/PlaygroundView.swift
@@ -176,9 +176,11 @@ struct PlaygroundView: View {
         }
         .navigationTitle("Playground")
         .navigationBarTitleDisplayMode(.inline)
-        .onAppear {
-            UIScrollView.appearance().keyboardDismissMode = .interactive
-        }
+        .gesture(DragGesture().onChanged(hideKeyboard))
+    }
+
+    private func hideKeyboard(_ value: DragGesture.Value) {
+        UIApplication.shared.sendAction(#selector(UIResponder.resignFirstResponder), to: nil, from: nil, for: nil)
     }
 }
 


Example/FinancialConnections Example/FinancialConnections Example/Playground/PlaygroundView.swift
```diff
@@ -176,6 +176,9 @@ struct PlaygroundView: View {
         }
         .navigationTitle("Playground")
         .navigationBarTitleDisplayMode(.inline)
+        .onAppear {
+            UIScrollView.appearance().keyboardDismissMode = .interactive
```
kgaidis-stripe, line 180: Does this only affect `PlaygroundView`? Or can it also affect the Auth Flow? _(if it can affect the auth flow that would be subpar because we always want to test the actual experience the users would have integrating in any app)_

kgaidis-stripe (dismissed):

Example/FinancialConnections Example/FinancialConnections Example/Playground/PlaygroundView.swift
kgaidis-stripe: ChatGPT 😅 :

> Yes, setting UIScrollView.appearance().keyboardDismissMode = .interactive will affect all instances of UIScrollView within your app. The appearance() proxy is a way to configure default values for UI elements across the entire application, so any scroll view will adopt this keyboard dismiss mode unless it's specifically overridden in an individual scroll view instance.

---

@mats-stripe  If we are to blindly follow LLM's, what would be the explanation of how this works? How would calling `UIScrollView.appearance()` magically reference the correct thing? Not saying its impossible, but even the API feels very odd if its globally setting something, yet only affecting one thing?

There's also `.scrollDismissesKeyboard(.interactively)`?

kgaidis-stripe: This also doesn't increase my confidence that its not global?

https://developer.apple.com/documentation/uikit/uiappearance

Screenshot 2024-07-04 at 5 18 18 PM

mats-stripe: I tested this out, and as you guessed, changing the `UIScrollView.appearance()` here affects the app behaviour globally 😅 

I just pushed a change to dismiss the keyboard when a drag gesture is recognized, which works just as well and is local to only the playground view!

Example/FinancialConnections Example/FinancialConnections Example/Playground/PlaygroundView.swift
```diff
@@ -176,6 +176,11 @@ struct PlaygroundView: View {
         }
         .navigationTitle("Playground")
         .navigationBarTitleDisplayMode(.inline)
+        .gesture(DragGesture().onChanged(hideKeyboard))
+    }
+
+    private func hideKeyboard(_ value: DragGesture.Value) {
+        UIApplication.shared.sendAction(#selector(UIResponder.resignFirstResponder), to: nil, from: nil, for: nil)
```
kgaidis-stripe, line 183: that's quite clever, and good naming; I guess the `value: DragGesture.Value` is "needed" because of the `onChanged` definition

kgaidis-stripe (approved):

Example/FinancialConnections Example/FinancialConnections Example/Playground/PlaygroundView.swift
mats-stripe: That's right!

PR #3749, opened 2024-07-03
Add legal disclaimer text to card form when setting up

yuki-stripe:
Note it doesn't show up if 'save this card' checkbox is checked, like web.

## Motivation
https://jira.corp.stripe.com/browse/RUN_MOBILESDK-3345

## Testing
See tests.

## Changelog
See changelog


porter-stripe (approved):

porter-stripe:
Unsure if it's ok to ship w/o translations here, it wouldn't be a regression but I'm fud-ful about showing legal text to people in the wrong language.

yuki-stripe:
> Unsure if it's ok to ship w/o translations here, it wouldn't be a regression but I'm fud-ful about showing legal text to people in the wrong language.

Definitely shouldn't ship w/o translations since this is GA

porter-stripe (approved):

PR #3753, opened 2024-07-04
[FC] Consent pane - Snapshot tests

mats-stripe:


StripeFinancialConnections/StripeFinancialConnections/Source/FinancialConnectionsSheet.swift
```diff
@@ -70,7 +70,7 @@ final public class FinancialConnectionsSheet {
     /// Completion block called when the sheet is closed or fails to open
     private var completion: ((HostControllerResult) -> Void)?
 
-    private var hostController: HostController?
+    private(set) var hostController: HostController?
```
mats-stripe, line 73: @kgaidis-stripe I'm curious of your thoughts on this change. I did this to get access to the presented view controller from the `FinancialConnectionsSheet`. If you know of another way (without exposing this property) I'm happy to revise!

StripeFinancialConnections/StripeFinancialConnections/Source/Native/Consent/ConsentLogoView.swift
```diff
@@ -59,7 +59,11 @@ final class ConsentLogoView: UIView {
 
     func animateDots() {
 #if targetEnvironment(simulator)
-        if ProcessInfo.processInfo.environment["UITesting"] != nil {
+        guard ProcessInfo.processInfo.environment["UITesting"] == nil else {
+            return
+        }
+
+        guard ProcessInfo.processInfo.environment["XCTestSessionIdentifier"] == nil else {
```
mats-stripe, line 66: This is similar to the above condition, but catches all XCTests environments rather than just UI tests. Thinking about it now, maybe we no longer need the above condition with this here 🤔

StripeFinancialConnections/StripeFinancialConnections/Source/FinancialConnectionsSheet.swift
kgaidis-stripe: we should be good! its `internal` anyway so if we need to access it from tests it is what it is 👍

StripeFinancialConnections/StripeFinancialConnections/Source/Native/Consent/ConsentLogoView.swift
kgaidis-stripe: feel free to delete if you are 100% sure that UITesting is unnecessary

I am personally not familiar with `XCTestSessionIdentifier` vs. `UITesting`

StripeFinancialConnections/StripeFinancialConnectionsTests/ConsentPaneSnapshotTests.swift
```diff
@@ -0,0 +1,92 @@
...
+        )
+    }
+
+    override func tearDown() {
+        super.tearDown()
+        fcs = nil
+    }
+
+    func testConsentPane() {
+        presentAndLoadFcs()
```
kgaidis-stripe, line 32: naming is subjective and its 100% non-blocking, but am curious on your thoughts of why prefer to use abbreviations vs. typing the whole name out?

so lets say there's a someone just reading the code and they see

A:

```
func testConsentPane() {
   presentAndLoadFcs()
```

B: 

```
func testConsentPane() {
   presentAndLoadFinancialConnectionSheet()
```

(Similar story is `fcs = nil` variable vs. `financialConnectionSheet = nil`)

What would be the key arguments why (A) beats (B) overall?

StripeFinancialConnections/StripeFinancialConnectionsTests/ConsentPaneSnapshotTests.swift
```diff
@@ -0,0 +1,92 @@
...
+        let navigationController = fcs.hostController!.navigationController
+        let loadingExpectation = XCTestExpectation(description: "Loading completed")
+
+        // The Financial Connections sheet usually takes anywhere between 50ms-200ms (but once in a while 2-3 seconds).
+        // to present with the expected content. When the sheet is presented, it initially shows a loading screen,
+        // and when it is done loading, the loading screen is replaced with the expected content.
+        // Therefore, the following code polls every 50 milliseconds to check if the ConsentViewController
+        // is present. We then wait 2.5s to let the images load from remote.
+        DispatchQueue.global(qos: .background).async {
+            var count = 0
```
kgaidis-stripe, line 56: subjective: maybe more succinct to use something like `for _ in 0..<10` (avoids extra variables and manual increment); some people also do something like `(0..<10).forEach` but thats arguably getting fancy haha

StripeFinancialConnections/StripeFinancialConnections.xcodeproj/project.pbxproj
```diff
@@ -860,15 +882,20 @@
 			path = ManualEntry;
 			sourceTree = "<group>";
 		};
-		9D434FE45EB09749E475FED6 /* MockData */ = {
+		9D434FE45EB09749E475FED6 /* Ressources */ = {
```
kgaidis-stripe, line 885: `Ressources` vs. `Resources`

kgaidis-stripe (dismissed):

StripeFinancialConnections/StripeFinancialConnectionsTests/ConsentPaneSnapshotTests.swift
mats-stripe: I'll change this to B!

StripeFinancialConnections/StripeFinancialConnections/Source/Native/Consent/ConsentLogoView.swift
mats-stripe: Just checked, when UI Testing, the `XCTestSessionIdentifier` environment variable isn't set, so we'll keep both of these checks here for now 👍

kgaidis-stripe (approved):

kgaidis-stripe (dismissed):

kgaidis-stripe (approved):

PR #3761, opened 2024-07-09
[FC] Add logging to synchronize API call

mats-stripe:
Resolves part of [BANKCON-11464](https://jira.corp.stripe.com/browse/BANKCON-11464)

## Summary

This adds some V1 logging to our call to the `synchronize` API. We log:
- At the beginning of the API call
- When the API call completes. Also includes these parameters:
    - Name of class which called `synchronize`.
    -  Whether or not the API call resulted in a success.

## Motivation

Improves observability into our synchronize call. We'll know when we executed this request, and when it fails.

## Testing

N/a

## Changelog

N/a

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/Analytics/FinancialConnectionsSheetAnalytics.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/Analytics/FinancialConnectionsSheetAnalytics.swift
@@ -50,6 +50,29 @@ struct FinancialConnectionsSheetFailedAnalytic: FinancialConnectionsSheetAnalyti
     let error: Error
 }
 
+/// Logged at the begining of the initial `synchronize` API call.
+struct FinancialConnectionsSheetInitialSynchronizeStarted: FinancialConnectionsSheetAnalytic {
+    let event = STPAnalyticEvent.financialConnectionsSheetInitialSynchronizeStarted
+    let clientSecret: String
+    let additionalParams: [String: Any] = [:]
+}
+
+/// Logged when the initial `synchronize` API call completes. Includes whether or not the call was a success, and the error otherwise.
+struct FinancialConnectionsSheetInitialSynchronizeCompleted: FinancialConnectionsSheetAnalytic {
+    let event = STPAnalyticEvent.financialConnectionsSheetInitialSynchronizeCompleted
+    let clientSecret: String
+    let success: Bool
+    let possibleError: Error?
+
+    var additionalParams: [String: Any] {
+        var params: [String: Any] = ["success": success]
+        if let error = possibleError {
+            params["error"] = error.serializeForV1Analytics()
+        }
+        return params
+    }
+}
+
 /// Helper to determine if we should log a failed analytic or closed analytic from the sheet's completion block
 struct FinancialConnectionsSheetCompletionAnalytic {
     /// Returns either a `FinancialConnectionsSheetClosedAnalytic` or `FinancialConnectionsSheetFailedAnalytic` depending on the result
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/Common/HostController.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/Common/HostController.swift
@@ -39,16 +39,18 @@ class HostController {
 
     // MARK: - Properties
 
-    private let api: FinancialConnectionsAPIClient
+    private let apiClient: STPAPIClient
     private let clientSecret: String
     private let returnURL: String?
     private let analyticsClient: FinancialConnectionsAnalyticsClient
+    private let analyticsClientV1: STPAnalyticsClientProtocol
 
     private var nativeFlowController: NativeFlowController?
     lazy var hostViewController = HostViewController(
+        analyticsClientV1: analyticsClientV1,
         clientSecret: clientSecret,
         returnURL: returnURL,
-        apiClient: api,
+        apiClient: apiClient,
         delegate: self
     )
     lazy var navigationController = FinancialConnectionsNavigationController(rootViewController: hostViewController)
@@ -58,13 +60,15 @@ class HostController {
     // MARK: - Init
 
     init(
-        api: FinancialConnectionsAPIClient,
+        apiClient: STPAPIClient,
+        analyticsClientV1: STPAnalyticsClientProtocol,
         clientSecret: String,
         returnURL: String?,
         publishableKey: String?,
         stripeAccount: String?
     ) {
-        self.api = api
+        self.apiClient = apiClient
+        self.analyticsClientV1 = analyticsClientV1
         self.clientSecret = clientSecret
         self.returnURL = returnURL
         self.analyticsClient = FinancialConnectionsAnalyticsClient()
@@ -121,7 +125,7 @@ extension HostController: HostViewControllerDelegate {
                 visualUpdate: synchronizePayload.visual,
                 returnURL: returnURL,
                 consentPaneModel: synchronizePayload.text?.consentPane,
-                apiClient: api,
+                apiClient: apiClient,
                 clientSecret: clientSecret,
                 analyticsClient: analyticsClient
             )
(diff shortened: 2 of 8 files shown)


mats-stripe:
I modified the extension called out above by @github-actions to be `@_spi(STP) public`.

StripeFinancialConnections/StripeFinancialConnections/Source/Common/HostViewController.swift
```diff
@@ -106,6 +109,16 @@ extension HostViewController {
...
+
+                analyticsClient.log(
+                    analytic: FinancialConnectionsSheetSynchronizeCompleted(
+                        clientSecret: clientSecret,
+                        class: String(describing: self),
+                        success: result.success
+                    ),
+                    apiClient: .shared
+                )
+
```
kgaidis-stripe, line 121: Apologies in advance for the task not being clear, and it being confusing. I think we can put all the necessary logging in this function

https://github.com/stripe/stripe-ios/blob/9c3d6eb1/StripeFinancialConnections/StripeFinancialConnections/Source/Common/HostViewController.swift#L99-L126

To recap the goal: we lack logging early in the flow so it would be good to know if:
1) first synchronize was called
2) first synchronize succeeded
3) first synchronize error'd

"first synchronize" is not even a great name because it doesn't really matter whether we are calling `synchronize` or some other API call, we just want to know what's going on early in the flow, and we have no way of knowing without more analytics

The current PR theoretically works, but I'm worried about all the extra code that is likely of no big value anyway. For example, we had to add `@_spi(STP) public let analyticsClient: STPAnalyticsClientProtocol` to `STPAPIClient` which is a shared class across SDK's and if it has existed this long without extra dependencies, its likely not a big need to add more.

kgaidis-stripe: There's at least two ways we could go about modifying this PR.

# Option One (the quick one based off this PR)

1) Keep this `FinancialConnectionsSheetSynchronizeCompleted` here and maybe rename it to something like `FinancialConnectionsSheetInitialSynchronizeCompleted` (but maybe there's better naming like FirstSynchronize or else)
2) Add an extra `FinancialConnectionsSheetInitialSynchronizeStarted` in this same function right before API call is made
3) Delete the rest of the code

That should cover it?

# Option Two (needs research on how to include error details)

Same as Option One except we also figure out how to include error details

This gets tricky because there's some special rules or naming around logging errors so they get picked up by the analytics pipeline/table. [here is an example query that queries the table](https://hubble.corp.stripe.com/queries/kgaidis/63ee3d2d) where the logs go

kgaidis-stripe: Maybe this is the official way to log errors using AnalyticsClientV1?

https://github.com/stripe/stripe-ios/blob/9c3d6eb1/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+DeferredAPI.swift#L33-L36

mats-stripe: Updated to option 1!

StripeFinancialConnections/StripeFinancialConnections/Source/Common/HostViewController.swift
```diff
@@ -99,13 +102,29 @@ extension HostViewController {
...
             .observe { [weak self] result in
                 guard let self = self else { return }
+
+                analyticsClient.log(
+                    analytic: FinancialConnectionsSheetInitialSynchronizeCompleted(
+                        clientSecret: clientSecret,
+                        success: result.success,
+                        possibleError: result.error
+                    ),
+                    apiClient: .shared
```
kgaidis-stripe, line 125: I believe it would be cleaner (and mostly going off precedent) to avoid the singleton:

https://github.com/stripe/stripe-ios/blob/6968678b/StripeFinancialConnections/StripeFinancialConnections/Source/FinancialConnectionsSheet.swift#L216-L221

StripeFinancialConnections/StripeFinancialConnections/Source/Analytics/FinancialConnectionsSheetAnalytics.swift
```diff
@@ -50,6 +50,29 @@ struct FinancialConnectionsSheetFailedAnalytic: FinancialConnectionsSheetAnalyti
     let error: Error
 }
 
+/// Logged at the begining of the initial `synchronize` API call.
+struct FinancialConnectionsSheetInitialSynchronizeStarted: FinancialConnectionsSheetAnalytic {
+    let event = STPAnalyticEvent.financialConnectionsSheetInitialSynchronizeStarted
+    var clientSecret: String
+    var additionalParams: [String: Any] = [:]
```
kgaidis-stripe, line 57: if `var` is not needed, there's precedent to do `let` here:

https://github.com/stripe/stripe-ios/pull/3761/files#diff-a1d45f375cb5ca585cd6f3a0ef47b25cc99c5b4f3cc74518831ccbdcf0f916ffR35-R36

StripeFinancialConnections/StripeFinancialConnections/Source/Analytics/FinancialConnectionsSheetAnalytics.swift
```diff
@@ -50,6 +50,29 @@ struct FinancialConnectionsSheetFailedAnalytic: FinancialConnectionsSheetAnalyti
...
+}
+
+/// Logged when the initial `synchronize` API call completes. Includes whether or not the call was a success, and the error otherwise.
+struct FinancialConnectionsSheetInitialSynchronizeCompleted: FinancialConnectionsSheetAnalytic {
+    let event = STPAnalyticEvent.financialConnectionsSheetInitialSynchronizeCompleted
+    var clientSecret: String
```
kgaidis-stripe, line 63: if `var` is not needed, there's precedent to do `let` here:

https://github.com/stripe/stripe-ios/pull/3761/files#diff-a1d45f375cb5ca585cd6f3a0ef47b25cc99c5b4f3cc74518831ccbdcf0f916ffR35-R36

kgaidis-stripe (approved):
thank you! and sorry for the back-and-forth on the initial design

StripeFinancialConnections/StripeFinancialConnections/Source/Common/HostController.swift
```diff
@@ -39,7 +39,8 @@ class HostController {
 
     // MARK: - Properties
 
-    private let api: FinancialConnectionsAPIClient
+    private let apiClient: FinancialConnectionsAPIClient
+    private let analyticsApiClient: STPAPIClient
```
kgaidis-stripe, line 43: oh that is quite unfortunate that we have to do that haha...I didn't realize.

even the idea that one can't use the analytics client "out of the box" where you have to pass in an API client in each call is very odd too

🤷

kgaidis-stripe (approved):

StripeFinancialConnections/StripeFinancialConnections/Source/Common/HostController.swift
kgaidis-stripe: I wonder if you can just change `private let apiClient: FinancialConnectionsAPIClient` to `private let apiClient: STPAPIClient ` and it works just as well? without the 2 variables?

kgaidis-stripe: I guess the key insight is that you don't really need to use `FinancialConnectionsAPIClient` (just use `STPAPIClient`) unless there are some unit tests that break where stubbing no longer works

mats-stripe: I think I can just keep only the `STPAPIClient` here! It is a bummer that the relationship is `STPAPIClient` conforms to `FinancialConnectionsAPIClient`, and not the other way around 🤷‍♂️

kgaidis-stripe (approved):

StripeFinancialConnections/StripeFinancialConnections/Source/Common/HostViewController.swift
```diff
@@ -58,11 +59,13 @@ final class HostViewController: UIViewController {
     // MARK: - Init
 
     init(
+        analyticsClient: STPAnalyticsClientProtocol = STPAnalyticsClient.sharedClient,
```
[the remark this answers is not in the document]
mats-stripe: Fixed!

kgaidis-stripe (dismissed):

kgaidis-stripe (approved):

kgaidis-stripe (approved):

PR #3763, opened 2024-07-09
[FC] Add logging to FlowRouter decision

mats-stripe:
Resolves [BANKCON-11464](https://jira.corp.stripe.com/browse/BANKCON-11464)

## Summary

This refactors the financial connections `FlowRouter` to provide a `Flow` enum, which is then logged (via the V1 analytics logger). We log when a flow has been decided, including what the flow is and whether or not the native killswitch is active.

## Motivation

Improves observability into which flow is being used in the financial connections sheet.

## Testing

N/a

## Changelog

N/a

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/Common/HostController.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/Common/HostController.swift
@@ -100,41 +100,31 @@ extension HostController: HostViewControllerDelegate {
     ) {
         delegate?.hostController(self, didReceiveEvent: FinancialConnectionsEvent(name: .open))
 
-        if synchronizePayload.manifest.isProductInstantDebits {
-            continueWithWebFlow(synchronizePayload.manifest)
-        } else {
-            let flowRouter = FlowRouter(
-                synchronizePayload: synchronizePayload,
-                analyticsClient: analyticsClient
-            )
-            defer {
-                // no matter how we exit this function
-                // log exposure to one of the variants if appropriate.
-                flowRouter.logExposureIfNeeded()
-            }
+        let flowRouter = FlowRouter(
+            synchronizePayload: synchronizePayload,
+            analyticsClient: analyticsClient
+        )
 
-            guard flowRouter.shouldUseNative else {
-                continueWithWebFlow(synchronizePayload.manifest)
-                return
-            }
-
-            navigationController.configureAppearanceForNative()
-
-            let dataManager = NativeFlowAPIDataManager(
-                manifest: synchronizePayload.manifest,
-                visualUpdate: synchronizePayload.visual,
-                returnURL: returnURL,
-                consentPaneModel: synchronizePayload.text?.consentPane,
-                apiClient: apiClient,
+        let flow = flowRouter.flow
+        analyticsClientV1.log(
+            analytic: FinancialConnectionsSheetFlowDetermined(
                 clientSecret: clientSecret,
-                analyticsClient: analyticsClient
-            )
-            nativeFlowController = NativeFlowController(
-                dataManager: dataManager,
-                navigationController: navigationController
-            )
-            nativeFlowController?.delegate = self
-            nativeFlowController?.startFlow()
+                flow: flow,
+                killswitchActive: flowRouter.killswitchActive
+            ),
+            apiClient: apiClient
+        )
+
+        switch flow {
+        case .webInstantDebits:
+            continueWithWebFlow(synchronizePayload.manifest)
+        case .nativeInstantDebits:
+            // Not currently supported.
+            break
+        case .webFinancialConnections:
+            continueWithWebFlow(synchronizePayload.manifest)
+        case .nativeFinancialConnections:
+            continueWithNativeFlow(synchronizePayload)
         }
     }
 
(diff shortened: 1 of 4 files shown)


StripeFinancialConnections/StripeFinancialConnections/Source/Common/HostController.swift
```diff
@@ -62,11 +63,13 @@ class HostController {
         clientSecret: String,
         returnURL: String?,
         publishableKey: String?,
-        stripeAccount: String?
+        stripeAccount: String?,
+        v1AnalyticsClient: STPAnalyticsClientProtocol = STPAnalyticsClient.sharedClient
```
kgaidis-stripe, line 67: this is not too important but I feel like its cleaner: 

I think you can avoid accessing the `sharedClient` here, and instead just pass the correct value from `FinancialConnectionsSheet` (which has this as an instance value):

https://github.com/stripe/stripe-ios/blob/9c3d6eb1/StripeFinancialConnections/StripeFinancialConnections/Source/FinancialConnectionsSheet.swift#L207-L213

StripeFinancialConnections/StripeFinancialConnections/Source/Common/HostController.swift
```diff
@@ -96,41 +99,31 @@ extension HostController: HostViewControllerDelegate {
...
-            )
-            nativeFlowController?.delegate = self
-            nativeFlowController?.startFlow()
+                flow: flow,
+                killswitchActive: flowRouter.killswitchActive
+            ),
+            apiClient: .shared
+        )
+
+        switch flow {
```
kgaidis-stripe, line 117: this is cleaner than before 👍

StripeFinancialConnections/StripeFinancialConnections/Source/Common/HostController.swift
```diff
@@ -170,6 +163,26 @@ private extension HostController {
         webFlowViewController.delegate = self
         navigationController.setViewControllers([webFlowViewController], animated: true)
     }
+
+    func continueWithNativeFlow(_ synchronizePayload: FinancialConnectionsSynchronize) {
```
kgaidis-stripe, line 167: thats nice that its similar to `continueWithWebFlow` now (separate parts) - separation is not as important as just consistency

kgaidis-stripe (approved):

StripeFinancialConnections/StripeFinancialConnections/Source/Common/HostController.swift
```diff
@@ -42,6 +42,7 @@ class HostController {
     private let api: FinancialConnectionsAPIClient
     private let clientSecret: String
     private let returnURL: String?
+    private let v1AnalyticsClient: STPAnalyticsClientProtocol
```
kgaidis-stripe, line 45: its likely a good idea to include the "v1" to differentiate the two different ones - maybe we should even rename the instance variable for this in `FinancialConnectionsSheet`

what do you think of naming it something like `analyticsClientV1`? the logic is that the important thing to know first is that its an "analytics_client" and only after the context is known, is it needed to differentiate its type ("v1")

so, for example, a developer might be coding and they type "analyticsClient" as a guess to reference an analytics client and it be good for the compiler to suggest/recognize and allow dev to pick the right one (maybe IDE is smart enough to pick up v1AnalyticsClient but the odd prefix of v1 might confuse it)

StripeFinancialConnections/StripeFinancialConnections/Source/Common/HostController.swift
```diff
@@ -96,41 +99,31 @@ extension HostController: HostViewControllerDelegate {
...
+            ),
+            apiClient: .shared
+        )
+
+        switch flow {
+        case .webInstantDebits:
+            continueWithWebFlow(synchronizePayload.manifest)
+        case .nativeInstantDebits:
+            // Not currently supported.
+            break
```
kgaidis-stripe, line 122: just making a mental note: I believe this is safe and there's no need to add error logging, or route to web flow, because we can verify with code logic that this will never hit anyway

StripeFinancialConnections/StripeFinancialConnections/Source/Common/HostController.swift
mats-stripe: Done!

StripeFinancialConnections/StripeFinancialConnections/Source/Common/HostController.swift
mats-stripe: Done!

StripeFinancialConnections/StripeFinancialConnections/Source/Common/HostController.swift
```diff
@@ -96,41 +99,31 @@ extension HostController: HostViewControllerDelegate {
...
-            nativeFlowController = NativeFlowController(
-                dataManager: dataManager,
-                navigationController: navigationController
-            )
-            nativeFlowController?.delegate = self
-            nativeFlowController?.startFlow()
+                flow: flow,
+                killswitchActive: flowRouter.killswitchActive
+            ),
+            apiClient: .shared
```
kgaidis-stripe, line 114: I believe we can avoid the singleton, and pass in the variable as per precedent of passing in variables vs. singletons:

https://github.com/stripe/stripe-ios/pull/3763/files#diff-ab3536226ca6435b73fb87d20bbb229b3d6c948e2cf82207e31e2ee27aef39c6L124

kgaidis-stripe (approved):

kgaidis-stripe (approved):

PR #3764, opened 2024-07-09
Financial Connections: for server driven UX GenericInfoScreen, added SafeEnumCodable to have safer parsing

kgaidis-stripe:
## Summary

This ensures the enum types are safer as this is a new "server driven UX" framework and it could change

[slack discussion here from backend](https://stripe.slack.com/archives/C01DNBVURH9/p1720543981528089?thread_ts=1720198192.666669&cid=C01DNBVURH9)

## Testing

We can see that the screen that decodes this model loads/works:

Simulator Screenshot - iPhone 15 Pro - 2024-07-09 at 15 38 42

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/Models/FinancialConnectionsGenericInfoScreen.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/Models/FinancialConnectionsGenericInfoScreen.swift
@@ -21,10 +21,11 @@ struct FinancialConnectionsGenericInfoScreen: Decodable {
         let icon: FinancialConnectionsImage?
         let alignment: Alignment?
 
-        enum Alignment: String, Decodable {
+        enum Alignment: String, SafeEnumCodable {
             case left = "left"
             case center = "center"
             case right = "right"
+            case unparsable
         }
     }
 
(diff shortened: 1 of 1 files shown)


StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/Models/FinancialConnectionsGenericInfoScreen.swift
```diff
@@ -21,10 +21,11 @@ struct FinancialConnectionsGenericInfoScreen: Decodable {
         let icon: FinancialConnectionsImage?
         let alignment: Alignment?
 
-        enum Alignment: String, Decodable {
+        enum Alignment: String, SafeEnumCodable {
```
mats-stripe, line 24: These could just be `SafeEnumDecodable`, right? This is also fine though 👍

mats-stripe (approved):

StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/Models/FinancialConnectionsGenericInfoScreen.swift
kgaidis-stripe: ahh true, I just copied the enums in the `FinancialConnectionsSheet` so I'll just go with the precedent haha

https://github.com/stripe/stripe-ios/blob/6968678b/StripeFinancialConnections/StripeFinancialConnections/Source/API%20Bindings/Models/FinancialConnectionsSession.swift#L72-L78

PR #3765, opened 2024-07-10
[FC] Add instant debits support to the playground app

mats-stripe:
## Summary

This adds support for the instant debits experience in the Financial Connections playground app. This currently only exists as a web view, and is not yet officially supported by the SDK.

> [!NOTE]  
> The Instant Debits flow only supports the _Payment Intent_ use case. For convenience, when Instant Debits is selected, we automatically set the use case to `.paymentIntent`.

In order to support this on the [server side](https://glitch.com/edit/#!/financial-connections-playground-ios), we pass in `{ "product": "intant_debits" }` as a parameter to the `/link_account_sessions` API when the experience value is set to `instant_debits`.

## Motivation

This will allow for easily testing the Instant Debits flow.

## Testing

- Ensured the existing Financial Connections flow is unaffected
- Ensured the new Instant Debit flow is launched and works* as expected

*Session output will still display a failure alert. This is expected _for now_.

https://github.com/stripe/stripe-ios/assets/172562065/e0d5e8ca-2b87-4e3e-a42b-93d09e9887fe

## Changelog

N/a

### DIFF
--- a/Example/FinancialConnections Example/FinancialConnections Example/Playground/PlaygroundView.swift
+++ b/Example/FinancialConnections Example/FinancialConnections Example/Playground/PlaygroundView.swift
@@ -17,6 +17,18 @@ struct PlaygroundView: View {
         ZStack {
             VStack {
                 Form {
+                    Section(header: Text("Experience")) {
+                        VStack(alignment: .leading, spacing: 4) {
+                            Picker("Select Experience", selection: viewModel.experience) {
+                                ForEach(PlaygroundConfiguration.Experience.allCases) {
+                                    Text($0.displayName)
+                                        .tag($0)
+                                }
+                            }
+                            .pickerStyle(.segmented)
+                        }
+                    }
+
                     Section(header: Text("Select SDK Type")) {
                         VStack(alignment: .leading, spacing: 4) {
                             Picker("Select SDK Type", selection: viewModel.sdkType) {
@@ -63,13 +75,20 @@ struct PlaygroundView: View {
 
                     Section(header: Text("Select Use Case")) {
                         VStack(alignment: .leading, spacing: 4) {
-                            Picker("Select Use Case", selection: viewModel.useCase) {
-                                ForEach(PlaygroundConfiguration.UseCase.allCases) {
-                                    Text($0.rawValue.capitalized.replacingOccurrences(of: "_", with: " "))
-                                        .tag($0)
+                            if viewModel.experience.wrappedValue == .instantDebits {
+                                Text("Instant debits only supports the _Payment Intent_ use case.")
+                                    .font(.caption)
+                                    .foregroundColor(.secondary)
+                            } else {
+                                Picker("Select Use Case", selection: viewModel.useCase) {
+                                    ForEach(PlaygroundConfiguration.UseCase.allCases) {
+                                        Text($0.rawValue.capitalized.replacingOccurrences(of: "_", with: " "))
+                                            .tag($0)
+                                    }
                                 }
+                                .pickerStyle(.segmented)
                             }
-                            .pickerStyle(.segmented)
+
                         }
                     }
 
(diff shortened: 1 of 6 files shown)


StripeFinancialConnections/StripeFinancialConnections/Source/FinancialConnectionsSheet.swift
```diff
@@ -143,13 +143,20 @@ final public class FinancialConnectionsSheet {
                     switch completedResult {
                     case .financialConnections(let session):
                         completion(.completed(session: session))
-                    case .instantDebits:
-                        let errorDescription = "Instant Debits is not supported via this interface."
-                        assertionFailure(errorDescription)
+                    case .instantDebits(let linkedBank):
+                        // TODO(mats): Add support for instant debits.
```
kgaidis-stripe, line 147: you might have spoken to Till about this but right now we only support Instant Debits utilizing Financial Connections SDK through the Payment SDK payment sheet

not sure if this project would change that? would be interesting to hear! but I've only launched it through the Payment Sheet example app

Example/FinancialConnections Example/FinancialConnectionsUITests/FinancialConnectionsUITests.swift
```diff
@@ -391,6 +391,29 @@ final class FinancialConnectionsUITests: XCTestCase {
             // 'cancel' the test as the bank is under the maintenance
         }
     }
+
+    func testWebInstantDebitsFlow() throws {
```
kgaidis-stripe, line 395: might make sense to eventually split these into another UITest class if easy to do

Example/FinancialConnections Example/FinancialConnections Example/Playground/PlaygroundView.swift
```diff
@@ -63,13 +75,20 @@ struct PlaygroundView: View {
 
                     Section(header: Text("Select Use Case")) {
                         VStack(alignment: .leading, spacing: 4) {
-                            Picker("Select Use Case", selection: viewModel.useCase) {
-                                ForEach(PlaygroundConfiguration.UseCase.allCases) {
-                                    Text($0.rawValue.capitalized.replacingOccurrences(of: "_", with: " "))
-                                        .tag($0)
+                            if viewModel.experience.wrappedValue == .instantDebits {
```
kgaidis-stripe, line 78: IMO you could have just not shown this section as users of this likely don't know any different so the extra messaging just adds clutter? I also don't see a "use case" in the [Connections Dev App](https://stripejsdevapp.corp.stripe.com/connections-auth-flow)

Example/FinancialConnections Example/FinancialConnections Example/Playground/PlaygroundView.swift
```diff
@@ -17,6 +17,18 @@ struct PlaygroundView: View {
         ZStack {
             VStack {
                 Form {
+                    Section(header: Text("Experience")) {
```
kgaidis-stripe, line 20: great name choice because Connections Dev App uses the same name 👍

kgaidis-stripe (dismissed):
It would have been helpful to see the test plan contain output of `bitrise run financial-connections-stability-tests` as that gives confidence that playground app works as usual

one additional tricky thing that helps to cover is that you need to run UI tests on both the device/version that the CI (Bitrise) is using, _and_ whatever developers would run too - sometimes UI tests might not work on different iOS verisons, which then causes failures/confusion

kgaidis-stripe:
Also you maybe covered this/thought-about it but whenever you modify the playground app backend, you need to make sure that the changes are backward compatible. 

Sometimes I add/commit a temporary "is_playground_v2" boolean where the backend ONLY uses new logic if that boolean exists

Example/FinancialConnections Example/FinancialConnections Example/Playground/PlaygroundView.swift
mats-stripe: Updated to hide the use case section when instant debits is selected 👍  

https://github.com/stripe/stripe-ios/assets/172562065/48353c1e-df29-4f00-bc2b-9c07e4e6b4a3

kgaidis-stripe (approved):

kgaidis-stripe (approved):

PR #3766, opened 2024-07-10
Fix vertical mandates

yuki-stripe:
## Summary
- Makes vertical mode display bottomNoticeAttributedString, used by US Bank and Instant Debits.
- Makes vertical mode display mandates for saved US Bank and saved SEPA Debit

## Testing
- Updated US Bank Account test to check mandate is visible
- Added Instant Debit test
- Added snapshot tests for saved US Bank and SEPA

## Changelog
Not user facing


porter-stripe (approved):
nice!

PR #3770, opened 2024-07-12
Fix bottom sheet content not using bottom safe area, adjust primary button bottom padding

yuki-stripe:
## Summary
- Fixes long standing issue where content clips the safe area. The historical problem w/ making the content VCs respect the `safeAreaLayoutGuide` was that it introduced an infinite layout loop when animating the sheet height. This is fixed now!
- "Button should have 8px bottom padding to account for the home indicator / bottom bevel" design feedback from https://docs.google.com/document/d/1tRLvYH4_UzRGevJuP15gU4SlU5zt9_pJBZtiOSMxy6A/edit?pli=1 

## Testing

https://github.com/user-attachments/assets/9df74e87-57f5-4215-9f1c-5150af4f1100

### Infinite loop bug
It's easy to repro this w/ horizontal FC + all billing details = always. Just go from Card to US Bank Account and observe the infinite loop if you comment out `contentInsetAdjustmentBehavior = .never`

### CVC keyboard bug fixed

https://github.com/user-attachments/assets/fbd9c95f-a8c4-4919-836b-6419bb957590

## Changelog
I think this is too small of a UI tweak to call out


davidme-stripe (dismissed):
LGTM! Did you check it on iPadOS?

yuki-stripe:
> LGTM! Did you check it on iPadOS?

Oop!  I checked, it looks..well, the same as master.

davidme-stripe (approved):

PR #3775, opened 2024-07-15
[FC] Show native instant debits flow from the example app

mats-stripe:
## Summary

Shows the (still WIP and very buggy) native instant debits flow if and only if launched from the FC example app, and the `Native` SDK type is selected. 

## Motivation

This will let us test the flow while we build it, without exposing it to users.

## Testing

Ensured the Native flow is only shown when the `Native` SDK is selected from the example app:

https://github.com/user-attachments/assets/c0f91f71-d8e2-45fb-8091-5fe8a02b5503

## Changelog

N/a

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/Common/FlowRouter.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/Common/FlowRouter.swift
@@ -44,6 +44,10 @@ class FlowRouter {
 
     var flow: Flow {
         if synchronizePayload.manifest.isProductInstantDebits {
+            if ProcessInfo.processInfo.environment["UITesting"] != nil {
+                // Show web instant debits flow while UITesting for now.
+                return .webInstantDebits
+            }
             return exampleAppSdkOverride.shouldUseNativeFlow ? .nativeInstantDebits : .webInstantDebits
         } else {
             logExposureIfNeeded()
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/Common/FlowRouter.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/Common/FlowRouter.swift
@@ -17,6 +17,16 @@ class FlowRouter {
         case nativeFinancialConnections = "native_financial_connections"
     }
 
+    private enum ExampleAppOverride: Equatable {
+        case native
+        case web
+        case none
+
+        var shouldUseNativeFlow: Bool {
+            self == .native
+        }
+    }
+
     private let synchronizePayload: FinancialConnectionsSynchronize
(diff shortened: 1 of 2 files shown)


StripeFinancialConnections/StripeFinancialConnections/Source/Common/FlowRouter.swift
```diff
@@ -54,12 +54,18 @@ class FlowRouter {
...
         ) as? Bool {
-            return isNativeEnabled
+            return nativeOverride
         }
+        return false
+    }
+
+    private var shouldUseNative: Bool {
+        // Override all other conditions if the example app has native selected.
+        if exampleAppNativeOverrideEnabled { return true }
```
kgaidis-stripe, line 68: @mats-stripe

I think there _might_ be a change in behavior that I don't see called out?

I'm not 100% sure, if that causes issues, however. Maybe it wouldn't always return web flow when expected?

We always returned a value here IF `FINANCIAL_CONNECTIONS_EXAMPLE_APP_ENABLE_NATIVE` was set at all. This seemingly returns the value _only_ if it returns true? 

Did you run `bitrise run financial-connections-stability-tests`? (I generally recommend to include that in test plans as a check)

mats-stripe: Good catch, I'll look into this to make sure I don't change any existing behaviour!

mats-stripe: Updated! This should now have the same behaviour of returning early if any non-nil value is set for `FINANCIAL_CONNECTIONS_EXAMPLE_APP_ENABLE_NATIVE`! Thanks for catching this

kgaidis-stripe (approved):

StripeFinancialConnections/StripeFinancialConnections/Source/Common/FlowRouter.swift
```diff
@@ -33,13 +43,12 @@ class FlowRouter {
     // MARK: - Public
 
     var flow: Flow {
-        guard synchronizePayload.manifest.isProductInstantDebits == false else {
-            // We currently only support Instant Debits via a web flow.
-            return .webInstantDebits
+        if synchronizePayload.manifest.isProductInstantDebits {
```
kgaidis-stripe, line 46: I also felt like if-statement is better here than guard because its not really a "precondition" - its a whole different conditional/flow

mats-stripe: Exactly!

kgaidis-stripe (approved):

PR #3782, opened 2024-07-16
Fixes for Xcode 16 Beta 3

davidme-stripe:
## Summary
* Implement the async delegate method instead of the non-async one to avoid a compiler warning related to strict concurrency support
* Move the Comparable implementation for STPCardBrand to avoid a retroactive conformance warning

## Motivation
Fix a few compiler issues on Xcode 16 beta 3.

## Testing
CI, manually in Xcode 16 beta 3


davidme-stripe:
Adding the "modifies public API" flag, as we're moving the above conformance from StripePaymentsUI to StripePayments.

porter-stripe:
Still works in beta 6, should we land this or wait for GA @davidme-stripe?

porter-stripe (approved):

PR #3783, opened 2024-07-16
Update vertical CVC flow

yuki-stripe:
## Summary
Update vertical CVC flow to transition similar to transitioning to a form - retaining the primary button across the animation. If confirmation fails, it transitions back to the list unless the error was a CVC error.

https://github.com/user-attachments/assets/f770eb7a-dd15-4318-9890-1b2cd1329ada

It required many changes to make it work:
- Scrolls view down when animating error label in
- Adds `contentOffsetPercentage` to BottomSheetVC, so that we can manipulate the content offset without specifying exact amounts e.g. we can now tell it to scroll to the bottom by setting `contentOffsetPercentage = 1.0`  
- Restores contentOffset of the vertical list when we switch back to it
- Make vertical CVC recollect present keyboard in viewDidAppear b/c viewWillAppear doesn't play well with the custom BottomSheetVC animation 

Cleaned up some things along the way:
- Set `beginAppearanceTransition` and `endAppearanceTransition` to better control when the view lifecycle methods are called e.g. `viewDidAppear` is now called only after the transition completes.
- Removed `didFinishAnimatingHeight`, which is now unnecessary with this fix ^
- Combined CVCReconfirmationVC and PreConfirmationVC

## Testing
- Added hardcoded rule to CVC recollect flow so that values of "666" turn into "test_invalid_cvc" so that the API will return an invalid_cvc error.
- Uses that to test CVC error in vertical PS UI test
- Adds snapshot tests for CVC error logic depending on whether error is CVC related or not

Manually tested existing CVC flows (horizontal PS and FC).
## Changelog
Not user facing


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/CardSectionWithScanner/CVCRecollectionElement.swift
```diff
@@ -26,18 +30,36 @@ final class CVCRecollectionElement: Element {
...
+        )
     }()
 
     let defaultValues: DefaultValues
     var paymentMethod: STPPaymentMethod
     let appearance: PaymentSheet.Appearance
 
+    lazy var textFieldElement: TextFieldElement = {
+        let textFieldElement = TextFieldElement(configuration: cvcElementConfiguration, theme: appearance.asElementsTheme)
+        textFieldElement.view.backgroundColor = appearance.colors.componentBackground
```
porter-stripe, line 48: Does passing in the theme/appearance not do this already? Same question with the corner radius.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/CardSectionWithScanner/CVCRecollectionElement.swift
```diff
@@ -59,36 +81,54 @@ final class CVCRecollectionElement: Element {
...
-        return cvcRecollectionView.textFieldElement.validationState
+        return textFieldElement.validationState
     }
     func clearTextFields() {
-        self.cvcRecollectionView.textFieldElement.setText("")
+        textFieldElement.setText("")
+    }
+
+    func updateErrorLabel() {
+        if case let .invalid(error, shouldDisplay) = textFieldElement.validationState, shouldDisplay {
```
porter-stripe, line 96: Why don't we do `cvcRecollectionView.errorLabel.isHidden = shouldDisplay`? Seems like `shouldDisplay` is unused?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/CVCReconfirmationVerticalViewController.swift
```diff
@@ -0,0 +1,74 @@
...
+            cvcRecollectionElement.view,
+        ])
+        stackView.spacing = 16
+        stackView.axis = .vertical
+        view.addAndPinSubview(stackView)
+    }
+
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+            self.cvcRecollectionElement.beginEditing()
```
porter-stripe, line 72: nit: Weird tab/spaces

StripeCore/StripeCoreTestUtils/XCTestCase+Stripe.swift
```diff
@@ -52,6 +52,14 @@ extension XCTestCase {
             description: description
         )
     }
+
+    public func wait(seconds: TimeInterval) {
+        let e = expectation(description: "Wait for \(seconds) seconds")
+        DispatchQueue.main.asyncAfter(deadline: .now() + seconds) {
+            e.fulfill()
+        }
+        waitForExpectations(timeout: seconds + 0.1)
```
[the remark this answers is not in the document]
yuki-stripe: Whoops, not needed.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/CardSectionWithScanner/CVCRecollectionElement.swift
yuki-stripe: Probably not, this was carried over. I can try to remove.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/CardSectionWithScanner/CVCRecollectionElement.swift
yuki-stripe: shouldDisplay is used, its value determines if we end up in this case or the else case. In plain english this line says:
```
If the validation state has an error and we should display it, ...
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/CardSectionWithScanner/CVCRecollectionElement.swift
yuki-stripe: Yep, unnecessary. There's more we could do to clean this element up fwiw but I don't want to do more than necessary right now.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/CardSectionWithScanner/CVCRecollectionElement.swift
porter-stripe: Ah I missed the trailing `, shouldDisplay` on line 96, thanks!

porter-stripe (approved):

PR #3785, opened 2024-07-17
Try using firstMatch

porter-stripe:


porter-stripe:
Seems to have no impact on CI time

PR #3791, opened 2024-07-17
Remove `connections_consent_combined_logo` experiment check

mats-stripe:
## Summary

Looks like the server is [always returning `treatment`](https://livegrep.corp.stripe.com/view/stripe-internal/pay-server/lib/bank_connections/experiments/constants.rb#L75-82) for this experiment now, so it is safe to remove it from the client code, and keep the "enabled" behaviour.

[Web](https://livegrep.corp.stripe.com/search/stripe?q=connections_consent_combined_logo%20path%3Av3&fold_case=auto&regex=false&context=true) is also not checking if this experiment is enabled anymore in V3.

## Motivation

Removes completed experiments & keeps us in sync with web.

## Testing

Manually verified everything still works as it did when the experiment was considered enabled.

## Changelog

N/a


kgaidis-stripe (approved):
🧹

PR #3793, opened 2024-07-17
[FC] Create FinancialConnectionsTheme and apply to primary button

mats-stripe:
## Summary

This is the first PR of many for theming the Financial Connections flow based on the `theme` value we receive in the `synchronize` call. Specifically, this themes the primary button, and also updates the navigation bar theme values to be centralized in `FinancialConnectionsTheme`. 

This touches many files, as with dependency injection we need to pass in this value from view controllers down to the view builders.

## Motivation

Alignment with the [Instant Debits figma](https://www.figma.com/design/wXQ8NJApqSJiLmxxANDRTs/%E2%9C%A8-Connections-3.0?node-id=787-114662&t=C9zgpIX82HpMQJuU-0)

## Testing

Manually ensured the primary button colors in the Financial Connections flow are unchanged, and are using the appropriate colors in the Instant Debits flow.

| Financial Connections | Instant Debits |
|--------|--------|
| Simulator Screenshot - iPhone 12 mini - 2024-07-17 at 16 10 02 | Simulator Screenshot - iPhone 12 mini - 2024-07-17 at 16 09 47 | 

## Changelog

N/a

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/Models/FinancialConnectionsTheme.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/Models/FinancialConnectionsTheme.swift
@@ -1,0 +1,53 @@
+//
+//  FinancialConnectionsTheme.swift
+//  StripeFinancialConnections
+//
+//  Created by Mat Schmid on 2024-07-17.
+//
+
+import UIKit
+
+enum FinancialConnectionsTheme: String, SafeEnumCodable, Equatable {
+    case light = "light"
+    case dashboardLight = "dashboard_light"
+    case linkLight = "link_light"
+    case unparsable
+}
+
+extension FinancialConnectionsTheme? {
+    var logo: Image {
+        switch self {
+        case .linkLight:
+            return .link_logo
+        case .light, .dashboardLight, .unparsable, .none:
+            return .stripe_logo
+        }
+    }
+
+    var primaryColor: UIColor {
+        switch self {
+        case .linkLight:
+            return .linkGreen200
+        case .light, .dashboardLight, .unparsable, .none:
+            return .brand500
+        }
+    }
+
+    var primaryButtonTextColor: UIColor {
+        switch self {
+        case .linkLight:
+            return .linkGreen900
+        case .light, .dashboardLight, .unparsable, .none:
+            return .white
+        }
+    }
+
+    var logoColor: UIColor {
+        switch self {
+        case .linkLight:
+            return .linkGreen900
+        case .light, .dashboardLight, .unparsable, .none:
+            return .textActionPrimary
+        }
+    }
+}
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/Models/FinancialConnectionsTheme.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/Models/FinancialConnectionsTheme.swift
@@ -7,19 +7,26 @@
 
 import UIKit
 
-enum FinancialConnectionsTheme: String, SafeEnumCodable, Equatable {
-    case light = "light"
-    case dashboardLight = "dashboard_light"
-    case linkLight = "link_light"
-    case unparsable
+enum FinancialConnectionsTheme: String, Equatable {
+    case light
+    case linkLight
+
+    init(from theme: FinancialConnectionsSessionManifest.Theme?) {
+        switch theme {
+        case .linkLight:
+            self = .linkLight
+        case .light, .dashboardLight, .unparsable, .none:
+            self = .light
+        }
+    }
 }
 
-extension FinancialConnectionsTheme? {
+extension FinancialConnectionsTheme {
     var logo: Image {
         switch self {
         case .linkLight:
             return .link_logo
-        case .light, .dashboardLight, .unparsable, .none:
+        case .light:
             return .stripe_logo
         }
     }
@@ -28,7 +35,7 @@ extension FinancialConnectionsTheme? {
         switch self {
         case .linkLight:
             return .linkGreen200
-        case .light, .dashboardLight, .unparsable, .none:
+        case .light:
             return .brand500
         }
     }
@@ -37,7 +44,7 @@ extension FinancialConnectionsTheme? {
         switch self {
         case .linkLight:
             return .linkGreen900
-        case .light, .dashboardLight, .unparsable, .none:
+        case .light:
             return .white
         }
     }
@@ -46,7 +53,7 @@ extension FinancialConnectionsTheme? {
         switch self {
         case .linkLight:
             return .linkGreen900
-        case .light, .dashboardLight, .unparsable, .none:
+        case .light:
             return .textActionPrimary
         }
     }
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/Helpers/UIColor+Extensions.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/Helpers/UIColor+Extensions.swift
@@ -163,12 +163,16 @@ extension UIColor {
         return UIColor(red: 34 / 255.0, green: 132 / 255.0, blue: 3 / 255.0, alpha: 1)  // #228403
     }
 
-    static var linkBrand200: UIColor {
+    static var legacyLinkBrand200: UIColor {
         return UIColor(red: 166 / 255.0, green: 251 / 255.0, blue: 221 / 255.0, alpha: 1)  // #a6fbdd
     }
 
-    static var linkBrand600: UIColor {
+    static var legacyLinkBrand600: UIColor {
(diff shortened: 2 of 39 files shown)


StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/Button+Extensions.swift
```diff
@@ -92,3 +92,63 @@ private final class ButtonFeedbackGeneratorHandler: NSObject {
...
+                .padding()
+            PrimaryButtonViewRepresentable(theme: .linkLight, enabled: false)
+                .frame(height: 64)
+                .padding()
+            SecondaryButtonViewRepresentable(enabled: true)
+                .frame(height: 64)
+                .padding()
+            SecondaryButtonViewRepresentable(enabled: false)
+                .frame(height: 64)
+                .padding()
```
mats-stripe, line 149: Screenshot 2024-07-18 at 1 22 16 PM

tillh-stripe (dismissed):
❤️

StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/Models/FinancialConnectionsTheme.swift
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    var primaryColor: UIColor {
+        switch self {
+        case .linkLight:
+            return .linkGreen200
+        case .light, .dashboardLight, .unparsable, .none:
+            return .brand500
+        }
+    }
```
kgaidis-stripe, line 34: I like how this turned out in the sense that colors are still "theme independent" (doesn't know about theme)

And instead a Theme wraps colors

Makes sense; at the start I was (naively) afraid whether theme would leak into all the lower-level API's (ex. color or fonts) but that's not necessary

StripeFinancialConnections/StripeFinancialConnections/Source/Helpers/UIColor+Extensions.swift
```diff
@@ -163,14 +163,18 @@ extension UIColor {
         return UIColor(red: 34 / 255.0, green: 132 / 255.0, blue: 3 / 255.0, alpha: 1)  // #228403
     }
 
-    static var linkBrand200: UIColor {
+    static var legacyLinkBrand200: UIColor {
```
kgaidis-stripe, line 166: good rename; I guess these will eventually get deleted once all the colors are replaced? if so, also good to iterate this way instead of changing all the things

StripeFinancialConnections/StripeFinancialConnections/Source/Native/AccountPicker/AccountPickerAccountLoadErrorView.swift
```diff
@@ -122,24 +126,28 @@ struct AccountPickerAccountLoadErrorView_Previews: PreviewProvider {
...
         AccountPickerAccountLoadErrorViewUIViewRepresentable(
             institutionName: "Chase",
+            theme: .light,
             didSelectTryAgain: {},
             didSelectEnterBankDetailsManually: {}
         )
 
         AccountPickerAccountLoadErrorViewUIViewRepresentable(
             institutionName: "Ally",
+            theme: .light,
```
kgaidis-stripe, line 136: also kinda cool that we can just test all the themes using SwiftUI

StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/Button+Extensions.swift
kgaidis-stripe: cool!

StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/DataAccessNoticeViewController.swift
```diff
@@ -12,13 +12,16 @@ import UIKit
 final class DataAccessNoticeViewController: SheetViewController {
 
     private let dataAccessNotice: FinancialConnectionsDataAccessNotice
+    private let theme: FinancialConnectionsTheme?
     private let didSelectUrl: (URL) -> Void
 
     init(
         dataAccessNotice: FinancialConnectionsDataAccessNotice,
+        theme: FinancialConnectionsTheme?,
```
kgaidis-stripe, line 20: what are scenarios where theme is NULL? does it make sense that it should be NULL? or should we always have a default value?

What I'm thinking of is that doesn't that extra NULL just add extra complexity in handling another state? 

But all depends on whether there's a need for a NULL

kgaidis-stripe (dismissed):
💯 

I do wonder on the theme null part 🤔

StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/DataAccessNoticeViewController.swift
mats-stripe: Main reason the theme is optional is because it's an optional property in the [backend manifest model](https://livegrep.corp.stripe.com/view/stripe-internal/pay-server/lib/bank_connections/api/resource/link_account_session_manifest_resource.rb#L59), so we have it optional in the client-side manifest model. 

Another reason, which hasn't surfaced anywhere in this PR, is that there's situations where having a null theme is "valid". For instance, when the flow is doing its initial load to fetch the manifest, we show a spinner view, and that view should be themed. Since we don't have a theme from the manifest yet, we'll use a null theme, and color the view neutrally rather than using Stripe or Link colors.

That being said, we _could_ have the client-side model have a non-nullable `theme` property, and fallback on a "default" theme when we either haven't loaded it yet, or the value sent back from the server is null. Though, since we don't have a guarantee the server will give us a theme in the synchronize response, I lean towards keeping it optional. Let me know what you think!

kgaidis-stripe: edit: I actually didn't realize how you avoided scattering nullability logic throughout the code:

```
extension FinancialConnectionsTheme? {
```

its an extension on the optional...

but then what happens if you don't use it as an optional? it seems like its a compiler error? if true that seems off?

---

as discussed in the 1:1, feel free to do what's best as you are in the best position to make the judgement call

the key thing to think about is whether there's ever a valid reason for auth flow UI code (consent etc.) to ever NOT have a theme (I'm excluding the rare exceptions before synchronize); IF it's true that we always want to have a theme in certain parts, then we should likely not have an optional there

mats-stripe: >but then what happens if you don't use it as an optional

That's a good point I hadn't thought of. Will look into this all!

mats-stripe: I'm a little on the fence about what the best approach here is. With the field being optional in the manifest server response, we need an optional theme field on the client side model. i.e. 

```swift
enum FinancialConnectionsTheme: String, SafeEnumCodable, Equatable {
    case light = "light"
    case dashboardLight = "dashboard_light"
    case linkLight = "link_light"
    case unparsable
}

let theme: Theme?
```

What we _could_ do, is have an abstraction to this model, which is non-nullable, and absorbs the null-case into the default `light` theme. Something like this: 

```swift
enum FinancialConnectionsAppearance {
    case light
    case linkLight
    
    init(from theme: FinancialConnectionsTheme?) {
        switch theme {
        case .linkLight:
            self = .linkLight
        case .light, .dashboardLight, .unparsable, .none:
            self = .light
        }
    }
}
```

Reason for naming it `Appearance` is because the `theme` namespace is already reserved by the synthesized decodable model (unless we want to specify custom `CodingKeys` 😅). Then, the manifest model would have to look something like this: 

```swift
private let theme: FinancialConnectionsTheme?
var appearance: FinancialConnectionsAppearance {
    Appearance(from: theme)
}
```

With this change, we solve the issue of passing around a nullable theme instance, but we add complexity to the client side manifest model and possible confusion by having both a `theme` and `appearance` field. To help with this, we can keep `theme` private to indicate `appearance` is the field to use (but it does suck losing the "theme" namespace).

@kgaidis-stripe do you have a preference between what I proposed above, or keeping it all as is? Or maybe even an alternative approach? I'm fine either way, but if we chose to go with the above suggestion, I think that, despite the additional noise, using custom `CodingKeys` would be great so that we can continue to use the "theme" namespace, and access it via `manifest.theme`.

mats-stripe: Take a look at the [latest commit](https://github.com/stripe/stripe-ios/pull/3793/commits/25a388380bbfd0412bd3c48b7cf0f07e9c68e251) for my suggested approach!

kgaidis-stripe (approved):

PR #3798, opened 2024-07-18
[wip] Remove Giropay References post Deprecation

adityar-stripe:


yuki-stripe:
Thanks for working on this! Just a comment on timing: IMO we should wait to merge this until the next major version bump, since this is a breaking change. Can discuss more in slack if you disagree!

PR #3802, opened 2024-07-19
Add support for recording network mocks from Xcode

davidme-stripe:
## Summary
* Add a new "AllStripeFrameworks+NetworkRecording" scheme, which we can use to re-record individual tests or test suites from Xcode. When selected, it will pass an environment variable (`STP_RECORD_NETWORK`) which will record all network requests.
* Add a `strictParamsEnforcement` setting, which is now the default. When enabled, the tests will fail if the request parameters don't match the recorded params.
* Add a `followRedirects` setting to the recorder to match the behavior in STPPaymentHandler when needed. Kind of a hack, but I think this is the only place where we have custom URLSession behavior.
* In OHHTTPStubs, added support for request ordering. Mocks are now returned in alphanumeric order based on the serial numbers we record in the filename.
* In OHHTTPStubs, added `removeAfterUse` flag. When enabled, `removeAfterUse` will "consume" each mocked network requests. This means that multiple requests to e.g. "create_payment_intent" within a single test run will return each PaymentIntent in sequence, matching the order of the recorded test.
* Move from eurias' OHHTTPStubs fork to my OHHTTPStubs fork. We should probably do a better job of vendoring this.
* Moved one simple test case (STPAPIClientNetworkBridgeTest) to use mocks.

## Motivation
Preparing to mock more network requests in our tests.

## Testing
CI


porter-stripe (approved):
🚀

PR #3804, opened 2024-07-19
Faster machine, more workers

porter-stripe:
## Summary
- More workers
- Faster machine
- Anywhere from 10-15% reduction in run times for UI tests, marginal increase in credit usage.

## Motivation
- CI sprint

## Testing
- Manual via CI

## Changelog
N/A


bitrise.yml
```diff
@@ -583,14 +583,14 @@ workflows:
         - scheme: PaymentSheet Example
         - test_plan: PaymentSheet Example-Shard2
         - log_formatter: xcbeautify
-        - xcodebuild_options: -maximum-concurrent-test-simulator-destinations 2 -maximum-parallel-testing-workers 2
+        - xcodebuild_options: -maximum-concurrent-test-simulator-destinations 4 -maximum-parallel-testing-workers 4
```
porter-stripe, line 586: Putting this at 6 workers causes it to fail for some reason

davidme-stripe: We should be consistent between shards IMO. It gets flaky when we add too many, and I don't want engineers writing tests to have to micromanage the shard count

bitrise.yml
```diff
@@ -565,14 +565,14 @@ workflows:
...
-        - xcodebuild_options: -maximum-concurrent-test-simulator-destinations 2 -maximum-parallel-testing-workers 2
+        - xcodebuild_options: -maximum-concurrent-test-simulator-destinations 6 -maximum-parallel-testing-workers 6
     - deploy-to-bitrise-io@2: {}
     before_run:
     - prep_all
     meta:
       bitrise.io:
         stack: osx-xcode-15.0.x
-        machine_type_id: g2-m1.8core
+        machine_type_id: g2-m1-max.5core
```
davidme-stripe, line 575: Seems good!

bitrise.yml
porter-stripe: Turned out to be an actual flake not caused by more workers, it's now fixed. Both shards are now using 6 workers. https://stripe.slack.com/archives/C02HQBW38JX/p1721675483160809

davidme-stripe (approved):

PR #3807, opened 2024-07-19
Financial Connections: for networking manual entry, added support for networking warm up pane Not Now

kgaidis-stripe:
## Summary

This PR:
- This PR adds support for manual entry networking opt out. Press "Manually verify instead" > warm up pane presents > press "Not Now" (this then leads to manual entry INSTEAD OF institution picker as it does now).
- The way this works is that 1) we extract parameter from URL of pressing manual entry 2) we pass that parameter to warm up pane 3) when user presses Not Now, we pass that parameter to disableNetworking 4) disableNetworking/backend will return manual entry INSTEAD OF institution picker.

Context:
- Financial Connections is implementing support for "networking manual entry." This feature allows a user to enter their manually entered account _once_, and then we will allow them to reuse it later through Link. 
- This is just one PR of many where each PR will be merged to a feature branch `fc-networkingmanualentry`.
- It's expected that this PR may have some gaps or TODO's because its going to a feature branch. That being said, we still want to make sure there's no glaring logic issues/bugs. 

### New Flow (manual entry "Now Now")

In this video, I am pressing the manual entry button at the bottom:

https://github.com/user-attachments/assets/98c15134-cc76-4999-afef-e96ea970082e

### Usual Flow (consent/institution picker "Not Now")

https://github.com/user-attachments/assets/9033c563-6ea1-42a2-a280-096583129397

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
@@ -102,6 +102,7 @@ protocol FinancialConnectionsAPIClient {
 
     func disableNetworking(
         disabledReason: String?,
+        clientSuggestedNextPaneOnDisableNetworking: String?,
         clientSecret: String
     ) -> Future<FinancialConnectionsSessionManifest>
 
@@ -539,6 +540,7 @@ extension STPAPIClient: FinancialConnectionsAPIClient {
                 case .failure(let error):
                     self.disableNetworking(
                         disabledReason: "account_numbers_not_available",
+                        clientSuggestedNextPaneOnDisableNetworking: nil,
                         clientSecret: clientSecret
                     ).observe { _ in } // ignoring return is intentional
 
@@ -604,6 +606,7 @@ extension STPAPIClient: FinancialConnectionsAPIClient {
 
     func disableNetworking(
         disabledReason: String?,
+        clientSuggestedNextPaneOnDisableNetworking: String?,
         clientSecret: String
     ) -> Future<FinancialConnectionsSessionManifest> {
         var body: [String: Any] = [
@@ -611,6 +614,7 @@ extension STPAPIClient: FinancialConnectionsAPIClient {
             "expand": ["active_auth_session"],
         ]
         body["disabled_reason"] = disabledReason
+        body["client_requested_next_pane_on_disable_networking"] = clientSuggestedNextPaneOnDisableNetworking
         return post(resource: APIEndpointDisableNetworking, parameters: body)
     }
 
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/AuthFlowHelpers.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/AuthFlowHelpers.swift
@@ -36,10 +36,17 @@ final class AuthFlowHelpers {
         url: URL,
         pane: FinancialConnectionsSessionManifest.NextPane,
         analyticsClient: FinancialConnectionsAnalyticsClient,
-        handleStripeScheme: (_ urlHost: String?) -> Void
+        handleURL: (_ urlHost: String?, _ nextPaneOrDrawerOnSecondaryCta: String?) -> Void
     ) {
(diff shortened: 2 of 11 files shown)


StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/CreatePaneParameters.swift
```diff
@@ -0,0 +1,20 @@
...
+//  Created by Krisjanis Gaidis on 7/22/24.
+//
+
+import Foundation
+
+// A bag of extra parameters we can pass to `CreatePaneViewController` function
+// that creates new pane view controllers. This avoids preserving state in
+// `NativeFlowDataManager` where the state might be outdated after a specific
+// pane push.
+struct CreatePaneParameters {
```
kgaidis-stripe, line 14: historically, lots of parameters were set on `NativeFlowDataManager` (you can see a bunch of them now)

however, at least for some parameters, this was likely not the best choice as its potentially preserving state that is no longer true after some pane pushes

that being said, this was never such a clear problem until this feature where preserving `nextPaneOrDrawerOnSecondaryCta` can actually cause a bug:
1. Tap "Manually verify instead" in Consent
2. Drawer presents and we save `nextPaneOrDrawerOnSecondaryCta`
3. Exit the drawer
4. Press "Agree and conitnue"
5. See drawer
6. Press "Not Now"
7. Pressing "Not Now" will wrongly navigate to manual entry instead of institution picker (due to actions in (2))

kgaidis-stripe: [PR with similar change in Stripe.js](https://git.corp.stripe.com/stripe-internal/stripe-js-v3/pull/23245/files)

StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/AuthFlowHelpers.swift
```diff
@@ -36,10 +36,17 @@ final class AuthFlowHelpers {
...
         pane: FinancialConnectionsSessionManifest.NextPane,
         analyticsClient: FinancialConnectionsAnalyticsClient,
-        handleStripeScheme: (_ urlHost: String?) -> Void
+        handleURL: (_ urlHost: String?, _ nextPaneOrDrawerOnSecondaryCta: String?) -> Void
     ) {
-        if let urlParameters = URLComponents(url: url, resolvingAgainstBaseURL: true),
-            let eventName = urlParameters.queryItems?.first(where: { $0.name == "eventName" })?.value
+        let internalLinkToPaneId: [String: String] = [
+            "manual-entry": "manual_entry"
+        ]
```
kgaidis-stripe, line 43: this mapping is a bit unfortunate but this is what [stripe.js decided to do](https://git.corp.stripe.com/stripe-internal/stripe-js-v3/blob/4c7be206/src/linkedAccounts/shared-ui/v3/typography/serverTextParser.tsx#L38-L40) so just following the lead

StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
```diff
@@ -604,13 +606,15 @@ extension STPAPIClient: FinancialConnectionsAPIClient {
...
         disabledReason: String?,
+        clientSuggestedNextPaneOnDisableNetworking: String?,
         clientSecret: String
     ) -> Future<FinancialConnectionsSessionManifest> {
         var body: [String: Any] = [
             "client_secret": clientSecret,
             "expand": ["active_auth_session"],
         ]
         body["disabled_reason"] = disabledReason
+        body["client_requested_next_pane_on_disable_networking"] = clientSuggestedNextPaneOnDisableNetworking
```
kgaidis-stripe, line 617: [stripe.js](https://git.corp.stripe.com/stripe-internal/stripe-js-v3/pull/23246/files#diff-ee09cc629c5030b4f8cdb14513bc50a01d0e24dac4cf15d7073aac9a58e941b3R132-R143)

kgaidis-stripe: this is the core change that activates "Not Now" from manual entry button to keep going to manual entry (instead of institution picker)

mats-stripe (approved):

PR #3808, opened 2024-07-19
Rework some billing collection UI tests

yuki-stripe:
## Summary
See each commit for reasoning why we can remove the test / what we replaced it with.

More to come but thought I'd cut a PR here and make sure there's no problems with this approach.


davidme-stripe (approved):

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetLPMConfirmFlowTests.swift
```diff
@@ -379,6 +377,72 @@ final class PaymentSheet_LPM_ConfirmFlowTests: XCTestCase {
...
+        configuration.defaultBillingDetails.address.country = "US"
+        configuration.defaultBillingDetails.address.postalCode = "12345"
+
+        try await _testConfirm(
+            intentKinds: [.paymentIntent, .paymentIntentWithSetupFutureUsage, .setupIntent],
+            currency: "USD",
+            paymentMethodType: .stripe(.card),
+            configuration: configuration
+        ) { form in
+            print(form)
```
wooj-stripe, line 427: probably want to remove this print.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetFormFactorySnapshotTest.swift
```diff
@@ -395,38 +390,14 @@ final class PaymentSheetFormFactorySnapshotTest: STPSnapshotTestCase {
...
-    private func usAddressSpecProvider() -> AddressSpecProvider {
-        let specProvider = AddressSpecProvider()
-        specProvider.addressSpecs = [
-            "US": AddressSpec(
-                format: "NOACSZ",
-                require: "ACSZ",
-                cityNameType: .city,
-                stateNameType: .state,
-                zip: "",
-                zipNameType: .zip
```
wooj-stripe, line 407: Is this what is causing all the snapshot tests to default "State" to Alabama?

yuki-stripe: Yes. The current state is better; it's ~not possible for PaymentSheet to look like the old snapshots where US state is a textfield and not a dropdown.

wooj-stripe (approved):

PR #3811, opened 2024-07-22
[BANKCON-11624] FC - Theme spinner view

mats-stripe:
## Summary

This themes the financial connections `SpinnerView` based on the manifest's `theme`. It supports:
- `light`: Stripe colors
- `linkLight`: Link colors
- `null`: Neutral colors

> [!NOTE]  
> We support a null theme here because there is a spinner is shown while the manifest is being fetched. 

## Motivation

Alignment with the [Instant Debits figma](https://www.figma.com/design/wXQ8NJApqSJiLmxxANDRTs/%E2%9C%A8-Connections-3.0?node-id=787-114662&t=C9zgpIX82HpMQJuU-0)

## Testing

Manual testing:

- Neutral-themed spinner is shown on the initial loading screen, no matter the flow.
- Stripe themed spinners shown on the FC experience.
- Link themed spinners shown on the ID experience.

https://github.com/user-attachments/assets/5a8f686e-9245-4eb6-a0d8-3808da3885fc

## Changelog

N/a

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/Models/FinancialConnectionsTheme.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/Models/FinancialConnectionsTheme.swift
@@ -93,15 +93,22 @@ extension FinancialConnectionsTheme {
             return .brand600
         }
     }
-}
 
-extension FinancialConnectionsTheme? {
     var spinnerColor: UIColor {
         switch self {
         case .linkLight:
             return .linkGreen200
         case .light:
             return .brand500
+        }
+    }
+}
+
+extension FinancialConnectionsTheme? {
+    var spinnerColor: UIColor {
+        switch self {
+        case .some(let theme):
+            return theme.spinnerColor
         case .none:
             return .neutral200
         }
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/Models/FinancialConnectionsTheme.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/Models/FinancialConnectionsTheme.swift
@@ -94,3 +94,16 @@ extension FinancialConnectionsTheme {
         }
     }
 }
+
+extension FinancialConnectionsTheme? {
+    var spinnerColor: UIColor {
+        switch self {
+        case .linkLight:
+            return .linkGreen200
+        case .light:
+            return .brand500
+        case .none:
+            return .neutral200
+        }
+    }
+}
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/Native/InstitutionPicker/InstitutionCellView.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/Native/InstitutionPicker/InstitutionCellView.swift
@@ -10,7 +10,7 @@ import Foundation
 import UIKit
 
 final class InstitutionCellView: UIView {
-    private let theme: FinancialConnectionsTheme?
+    private let theme: FinancialConnectionsTheme!
 
     private lazy var horizontalStackView: UIStackView = {
         let horizontalStackView = UIStackView()
@@ -73,7 +73,7 @@ final class InstitutionCellView: UIView {
         return activityIndicator
     }()
 
-    init(theme: FinancialConnectionsTheme?) {
+    init(theme: FinancialConnectionsTheme) {
         self.theme = theme
         super.init(frame: .zero)
         backgroundColor = .clear
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/Native/InstitutionPicker/InstitutionCellView.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/Native/InstitutionPicker/InstitutionCellView.swift
@@ -10,7 +10,7 @@ import Foundation
 import UIKit
 
 final class InstitutionCellView: UIView {
-    private let theme: FinancialConnectionsTheme!
+    private let theme: FinancialConnectionsTheme
 
     private lazy var horizontalStackView: UIStackView = {
         let horizontalStackView = UIStackView()
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/Native/InstitutionPicker/InstitutionCellView.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/Native/InstitutionPicker/InstitutionCellView.swift
@@ -10,6 +10,7 @@ import Foundation
 import UIKit
 
 final class InstitutionCellView: UIView {
+    private let theme: FinancialConnectionsTheme?
 
     private lazy var horizontalStackView: UIStackView = {
         let horizontalStackView = UIStackView()
@@ -52,7 +53,7 @@ final class InstitutionCellView: UIView {
     private var iconView: UIView?
     private lazy var loadingView: ActivityIndicator = {
         let activityIndicator = ActivityIndicator(size: .medium)
-        activityIndicator.color = .iconActionPrimary
+        activityIndicator.color = theme.spinnerColor
         activityIndicator.startAnimating()
 
         // re-size `ActivityIndicator` to a size we desire
@@ -72,8 +73,9 @@ final class InstitutionCellView: UIView {
         return activityIndicator
     }()
 
-    override init(frame: CGRect) {
-        super.init(frame: frame)
+    init(theme: FinancialConnectionsTheme?) {
+        self.theme = theme
+        super.init(frame: .zero)
         backgroundColor = .clear
         addAndPinSubview(horizontalStackView)
     }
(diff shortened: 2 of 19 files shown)


StripeFinancialConnections/StripeFinancialConnections/Source/Web/FinancialConnectionsWebFlowViewController.swift
```diff
@@ -84,7 +84,8 @@ final class FinancialConnectionsWebFlowViewController: UIViewController {
         return item
     }()
 
-    private let loadingView = LoadingView(frame: .zero)
+    // Use nil theme so the spinner view doesn't flash to the theme's color before launching the webview.
+    private let loadingView = LoadingView(frame: .zero, theme: nil)
```
mats-stripe, line 88: This fixes a small issue where the spinner would flash to the theme's color right before launching the web flow.

Before:

https://github.com/user-attachments/assets/ab4c55b3-ba22-4b11-aa5b-9331b1cc4cc2

After:

https://github.com/user-attachments/assets/cfebaa54-2d24-49f9-b9fc-0e6a19203a1b

StripeFinancialConnections/StripeFinancialConnections/Source/Native/InstitutionPicker/InstitutionCellView.swift
```diff
@@ -72,8 +73,9 @@ final class InstitutionCellView: UIView {
         return activityIndicator
     }()
 
-    override init(frame: CGRect) {
-        super.init(frame: frame)
+    init(theme: FinancialConnectionsTheme?) {
```
kgaidis-stripe, line 76: is that weird that the theme is NULL here but `InstitutionCellView` is used in a place where it isn't expected to be? (...because its inside of Auth Flow where the the theme is now non-null for "after synchronize")? https://github.com/stripe/stripe-ios/pull/3793/files

kgaidis-stripe: this is in a bunch of places

StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/Models/FinancialConnectionsTheme.swift
```diff
@@ -58,3 +58,16 @@ extension FinancialConnectionsTheme {
...
+        switch self {
+        case .linkLight:
+            return .linkGreen200
+        case .light:
+            return .brand500
+        case .none:
+            return .neutral200
+        }
+    }
+}
```
kgaidis-stripe, line 73: I initially thought it be easier to just do something like:

`spinner.backgroundColor = theme?.spinnerColor ?? .neutral200` (or pass in the spinner color as the configuration)

but I guess sometimes you might need the whole theme instead of one color? another is to introduce a new theme "neutral" or something...but that seems a bit over engineery too

then we likely don't want to ONLY be able to access a color IF the theme is non-null; so maybe a better way is to write it something like 

```
extension FinancialConnectionsTheme? {
    var spinnerColor: UIColor {
        switch self {
        case .all: // this is psuedocode that doesn't compile
            return theme.spinnerColor
        case .none:
            return .neutral200
        }
    }
    ```

File: StripeFinancialConnections/StripeFinancialConnections/Source/Native/InstitutionPicker/InstitutionTableViewCell.swift [hunk 4]
```diff
@@ -121,17 +126,26 @@ struct InstitutionTableViewCell_Previews: PreviewProvider {
...
+                showLoadingView: true,
+                showOverlayView: false,
+                theme: .light
             ).frame(width: 343, height: 72)
 
             InstitutionTableViewCellUIViewRepresentable(
                 showLoadingView: true,
-                showOverlayView: false
+                showOverlayView: false,
+                theme: .linkLight
```
mats-stripe, line 142: Screenshot 2024-07-22 at 10 56 02 AM

StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/Models/FinancialConnectionsTheme.swift
mats-stripe: Good points! The reason for putting the `spinnerColor` extension variable on the optional theme was to keep all the color decisions centralized in one place. Though, I agree with the point of only being able to access the `spinnerColor` when the theme is optional, so I'll update this with your suggestion!

StripeFinancialConnections/StripeFinancialConnections/Source/Native/InstitutionPicker/InstitutionTableViewCell.swift
```diff
@@ -10,12 +10,13 @@ import Foundation
 import UIKit
 
 final class InstitutionTableViewCell: UITableViewCell {
+    private var theme: FinancialConnectionsTheme!
```
kgaidis-stripe, line 13: RE: https://github.com/stripe/stripe-ios/pull/3811#discussion_r1686967617

oh maybe this is why you had the nullls? (note that I could have been wrong where NULL was justified 😅 )

hmm it's hard to have the full picture without playing around with it but wouldn't this crash if something is wrongly accessed? for example any access to `institutionCellView` can crash if not accessed after `customize`? 

I can approve if you are OK with the risks, but I do feel like something likely has to be re-organized because there are several oddities that usually don't have to happen:
1. use of `!` (I almost never use it due to the risks)
2. I believe the moving of `contentView.addAndPinSubview(institutionCellView)` likely _works_ (its a no-op) because its just "re-adding" the same view(?) but it might be "logically wrong" because you are only supposed to add the view once (ex. table view reloads, but it reuses the same cells)
3. lazy property that requires `theme` to be there, but the `theme` is not passed in initializer, its in `customize`

overall, the core problem seems to be that the introduction of `theme` is breaking how the code existed today, and this is causing ripple effects

for example, one mind shift could be that: maybe due to the introduction of `theme` and the need/want for the theme to be passed in initializers (and UITableView not allowing to pass theme in init?), we can no longer construct the view in `init`; instead we need to construct it in `customize`, and this means how we build `InstitutionTableViewCell.swift` has to change from first-principles (ex. lazy variables, and non-nulls of _views_, potentially don't make sense because we don't have the info to construct them until `customize` is called etc. etc.)

---

💡  thinking out loud to unblock: a quick patch could be to just assign a theme `private var theme: FinancialConnectionsTheme = .light` - its still odd but at the very least avoids a crash

StripeFinancialConnections/StripeFinancialConnections/Source/Native/InstitutionPicker/InstitutionTableViewCell.swift
```diff
@@ -74,14 +74,17 @@ final class InstitutionTableViewCell: UITableViewCell {
...
+        self.theme = theme
         institutionIconView.setImageUrl(institution.icon?.default)
 
         institutionCellView.customize(
             iconView: institutionIconView,
             title: institution.name,
             subtitle: AuthFlowHelpers.formatUrlString(institution.url)
         )
+
+        contentView.addAndPinSubview(institutionCellView)
```
kgaidis-stripe, line 87: as mentioned in here:

https://github.com/stripe/stripe-ios/pull/3811/files#r1688386941

this is likely a no-op moving it here, but its "logically" off because technically we should only need to add a view once; if this was a large app, it could have caused perf issues (depending on how iOS handles this - ex. does it treat this as no-op, or does it re-add the view and constraints)

(I only noticed this because its similar to a common performance/coding problem of UITableView/UICollectionView where views are constructed in "customize-type-of-method", which causes these methods to be called on every scroll)

mats-stripe: I'm not too concerned about the view being added more than once, since that would be caused by developer error, but what I could do is add an `institutionCellView.removeFromSuperview()` before it is added in the `customize` call to prevent this from being an issue.

StripeFinancialConnections/StripeFinancialConnections/Source/Native/InstitutionPicker/InstitutionTableViewCell.swift
mats-stripe: Yeah this one is a bit tricky, because we need the `theme` before we add the view, but we can't pass it in an initializer, since we're using an implicit `UITableViewCell` initializer: 

```swift
let cell = tableView.dequeueReusableCell(withIdentifier: cellIdentifier) as? InstitutionTableViewCell
```

As for the force unwrapped property, I agree with your suggestion of changing it from a `lazy` var, and just making it an optional var that is set in the `customize` call. This way we don't even need to store `theme` as a property on `self`!

StripeFinancialConnections/StripeFinancialConnections/Source/Native/InstitutionPicker/InstitutionTableViewCell.swift
mats-stripe: Actually, without this change, this introduced a bug where the view _was_ being added multiple times when scrolling the Institution table view: 



But this `removeFromSuperview()` change fixes that: 

https://github.com/user-attachments/assets/61740ac0-7a3a-4947-821f-9f1e5e052b9f

kgaidis-stripe: sweet; these somewhat-minor changes completely shift the code around and it immediately compiles in my head 😅

kgaidis-stripe (approved):

PR #3818, opened 2024-07-23
Attempt to improve ui test run time

wooj-stripe:
## Summary
Noticed that waitforExistance can take up to 2 seconds to execute, but exists takes ~40ms to run.  If the item exits, we should forcibly tap it.  It introduced some flakiness in some tests, which I have fixed.

This change reduces the time spent on UI tests.
Assuming ui-tests-1 was taking about 26:53 and ui-tests-2 was taking about 25:39 minutes, I believe this saves about~20 seconds on ui-tests-1, ~2 minutes on ui-tests-2.

## Motivation
Reduce CI build time.

## Testing
Relying on CI

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheetUITest/LinkPaymentControllerUITest.swift
+++ b/Example/PaymentSheet Example/PaymentSheetUITest/LinkPaymentControllerUITest.swift
@@ -91,7 +91,7 @@ class LinkPaymentControllerUITest: XCTestCase {
             .withOffset(CGVector(dx: 0, dy: -130))
             .tap()
 
-        sleep(1) // wait for modal to disappear before pressing Buy
+        sleep(3) // wait for modal to disappear before pressing Buy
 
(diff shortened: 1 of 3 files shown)


Example/PaymentSheet Example/PaymentSheetUITest/LinkPaymentControllerUITest.swift
```diff
@@ -91,7 +91,7 @@ class LinkPaymentControllerUITest: XCTestCase {
             .withOffset(CGVector(dx: 0, dy: -130))
             .tap()
 
-        sleep(1) // wait for modal to disappear before pressing Buy
+        sleep(3) // wait for modal to disappear before pressing Buy
```
yuki-stripe, line 94: What happens if you tap the buy button immediately?  This sleep is long enough to make me wonder if it's possible for customers to run into whatever issue we fixed by sleeping.

yuki-stripe (approved):

PR #3823, opened 2024-07-24
[BANKCON-11512] Add killswitch for Native Instant Debits flow

mats-stripe:
Resolves https://jira.corp.stripe.com/browse/BANKCON-11512

## Summary

This makes use of the native experience killswitch (`bank_connections_mobile_native_version_killswitch` in the manifest's `feature` payload) in the native instant debits flow router. If the killswitch is active, we will always show the web instant debits flow.

In the second commit, I also added some unit tests for the `FlowRouter` to make sure everything works as expected.

## Motivation

The killswitch will allow us to remotely deactivate the native experience if we detect an issue with it. 

## Testing

Unit tests added! 

Also in this video you can see that with the `Automatic` flow selected, the Networking merchant, which does not have the killswitch activated (you can see the test merchant `carlosmuvi-us` [here](https://amp.corp.stripe.com/feature-flags/flag/bank_connections_mobile_native_instant_debits_eligible_merchants)), but for a different merchant which has the killswitch activated, they get the web flow:

https://github.com/user-attachments/assets/5c170605-f917-4bc1-92de-7c47af3a4910

## Changelog

N/a

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/Models/FinancialConnectionsSessionManifest.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/Models/FinancialConnectionsSessionManifest.swift
@@ -126,6 +126,84 @@ struct FinancialConnectionsSessionManifest: Decodable {
         !livemode
     }
 
+    init(
+        accountholderCustomerEmailAddress: String? = nil,
+        accountholderIsLinkConsumer: Bool? = nil,
+        accountholderPhoneNumber: String? = nil,
+        accountholderToken: String? = nil,
+        accountDisconnectionMethod: FinancialConnectionsSessionManifest.AccountDisconnectionMethod? = nil,
+        activeAuthSession: FinancialConnectionsAuthSession? = nil,
+        activeInstitution: FinancialConnectionsInstitution? = nil,
+        allowManualEntry: Bool,
+        assignmentEventId: String? = nil,
+        businessName: String? = nil,
+        cancelUrl: String? = nil,
+        consentRequired: Bool,
+        customManualEntryHandling: Bool,
+        disableLinkMoreAccounts: Bool,
+        displayText: FinancialConnectionsSessionManifest.DisplayText? = nil,
+        experimentAssignments: [String: String]? = nil,
+        features: [String: Bool]? = nil,
+        hostedAuthUrl: String? = nil,
+        initialInstitution: FinancialConnectionsInstitution? = nil,
+        instantVerificationDisabled: Bool,
+        institutionSearchDisabled: Bool,
+        isEndUserFacing: Bool? = nil,
+        isLinkWithStripe: Bool? = nil,
+        isNetworkingUserFlow: Bool? = nil,
+        isStripeDirect: Bool? = nil,
+        livemode: Bool,
+        manualEntryMode: FinancialConnectionsSessionManifest.ManualEntryMode,
+        manualEntryUsesMicrodeposits: Bool,
+        nextPane: FinancialConnectionsSessionManifest.NextPane,
+        paymentMethodType: FinancialConnectionsPaymentMethodType? = nil,
+        permissions: [StripeAPI.FinancialConnectionsAccount.Permissions],
+        product: String,
+        singleAccount: Bool,
+        skipSuccessPane: Bool? = nil,
+        stepUpAuthenticationRequired: Bool? = nil,
+        successUrl: String? = nil,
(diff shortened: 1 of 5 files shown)


StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/Models/FinancialConnectionsSessionManifest.swift
```diff
@@ -126,6 +126,84 @@ struct FinancialConnectionsSessionManifest: Decodable {
         !livemode
     }
 
+    init(
```
mats-stripe, line 129: I'm not _too_ sure why, but in order to initialize this in the unit tests, I had to add this explicit initializer 🤷‍♂️

StripeFinancialConnections/StripeFinancialConnections/Source/Analytics/FinancialConnectionsAnalyticsClient.swift
```diff
@@ -18,12 +18,12 @@ protocol FinancialConnectionsAnalyticsClientDelegate: AnyObject {
 
 final class FinancialConnectionsAnalyticsClient {
 
-    private let analyticsClient: AnalyticsClientV2
+    private let analyticsClient: AnalyticsClientV2Protocol
```
mats-stripe, line 21: This change allows us to dependency inject a mock analytics client in the unit tests. This should not cause any functional change to the implementation.

StripeFinancialConnections/StripeFinancialConnectionsTests/FlowRouterTests.swift
```diff
@@ -0,0 +1,187 @@
...
+//  StripeFinancialConnectionsTests
+//
+//  Created by Mat Schmid on 2024-07-24.
+//
+
+@_spi(STP) import StripeCoreTestUtils
+@testable import StripeFinancialConnections
+import XCTest
+
+class FlowRouterTests: XCTestCase {
```
tillh-stripe, line 12: ❤️

tillh-stripe (approved):

StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/Models/FinancialConnectionsSessionManifest.swift
kgaidis-stripe: just purely for curiosity, this is very interesting haha

I wonder if its `private let _theme: Theme?` - where the `private` is somehow confusing the compiler - but its still weird because you'd think Swift currently has an `internal` initializer generated (and it works?) so why the problem? 🤔

mats-stripe: Actually, yeah that's it! If I get rid of the `_theme` property, the initializer is generated... Wonder why 🤔

StripeFinancialConnections/StripeFinancialConnections/Source/Common/FlowRouter.swift
```diff
@@ -44,14 +44,10 @@ class FlowRouter {
 
     var flow: Flow {
         if synchronizePayload.manifest.isProductInstantDebits {
-            if ProcessInfo.processInfo.environment["UITesting"] != nil {
-                // Show web instant debits flow while UITesting for now.
-                return .webInstantDebits
-            }
-            return exampleAppSdkOverride.shouldUseNativeFlow ? .nativeInstantDebits : .webInstantDebits
+            return shouldUseNativeInstantDebits ? .nativeInstantDebits : .webInstantDebits
```
kgaidis-stripe, line 47: very clear naming 🙏 `shouldUseNativeInstantDebits` + `shouldUseNativeFinancialConnections`

StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/Models/FinancialConnectionsSessionManifest.swift
mats-stripe: So it seems like I could either keep the explicit initializer or make `_theme` no longer `private`. I lean towards the former, but I'll go with whichever you prefer!

kgaidis-stripe: oh no worries at all, lets go with your preference - I was just curious (it is interesting!)

PR #3841, opened 2024-07-26
Add PayPal, RevolutPay, CashApp to LPMConfirmFlowTests. Try fixing some flakes too.

yuki-stripe:
…flake.

## Summary

Also tried to fix the worst flake in the past 12 weeks according to bitrise's view. I'm not sure if it's better, I couldn't repro the flake locally.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PollingViewController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PollingViewController.swift
@@ -211,8 +211,8 @@ class PollingViewController: UIViewController {
                                   repeats: true)
         }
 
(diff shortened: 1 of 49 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PollingViewController.swift
```diff
@@ -211,8 +211,8 @@ class PollingViewController: UIViewController {
                                   repeats: true)
         }
 
-        DispatchQueue.main.asyncAfter(deadline: .now() + 5) {
-            self.intentPoller.beginPolling()
+        DispatchQueue.main.asyncAfter(deadline: .now() + 5) { [weak self] in
```
porter-stripe, line 214: 👍, wonder if we can have some sort of static analysis to catch these.

porter-stripe (approved):

PR #3844, opened 2024-07-29
Rebalance sharding across 4 workers

davidme-stripe:
## Summary
Rebalance the shards across 4 workers, split some LPM and deferred tests into multiple cases for better parallelization.

## Motivation
Reducing from 22min

## Testing
CI


porter-stripe (approved):
🚀

PR #3849, opened 2024-07-29
Better format `ordered_lpms` in metrics

porter-stripe:
## Summary
- Zoolander injests an array of strings in a weird way making it hard to query. 
- Currently looks like `"ordered_lpms[0]":"card","ordered_lpms[1]":"paypal"`
- Will now look like `"ordered_lpms"="card,paypal,amazon_pay"

## Motivation
Easier query

## Testing
Manual

## Changelog
N/A


porter-stripe:
cc @amk-stripe

wooj-stripe:
Seems fine, but would be better to extract params to another function and write a test.

wooj-stripe (approved):

PR #3851, opened 2024-07-29
Extract elementsSession from Intent. 

yuki-stripe:
## Motivation
It's not a property of the PaymentIntent/SetupIntent object. 

## Testing
Existing tests. 

## Changelog
Not user facing


davidme-stripe (approved):

davidme-stripe (approved):

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetGDPRConfirmFlowTests.swift
```diff
@@ -393,6 +395,7 @@ final class PaymentSheet_GDPR_ConfirmFlowTests: STPNetworkStubbingTestCase {
             configuration: configuration,
             authenticationContext: self,
             intent: intent,
+            elementsSession: ._testValue(intent: intent),
```
wooj-stripe, line 398: We should be passing in the value that is defined in each test case rather than passing in this test value.

PR #3858, opened 2024-07-30
Pass through correct value of elementsSession

wooj-stripe:
## Summary
In these test cases, we weren't passing in the correct instance of elements/sessions. Introduced here:
https://github.com/stripe/stripe-ios/pull/3851/files#r1697340803

## Motivation
Nightly non-stubbed tests were failing

## Testing
Manually ran tests on my box

### DIFF
--- a/StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetGDPRConfirmFlowTests.swift
+++ b/StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetGDPRConfirmFlowTests.swift
@@ -302,7 +302,6 @@ final class PaymentSheet_GDPR_ConfirmFlowTests: STPNetworkStubbingTestCase {
         let clientSecretResolved = expectation(description: "clientSecretResolved")
         var clientSecret: String!
         let intent = try await intentCreation(intentKind: intentKind,
-                                              elementsSession: elementsSession,
                                               apiClient: apiClient,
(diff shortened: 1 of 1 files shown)


StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetGDPRConfirmFlowTests.swift
```diff
@@ -302,7 +302,6 @@ final class PaymentSheet_GDPR_ConfirmFlowTests: STPNetworkStubbingTestCase {
         let clientSecretResolved = expectation(description: "clientSecretResolved")
         var clientSecret: String!
         let intent = try await intentCreation(intentKind: intentKind,
-                                              elementsSession: elementsSession,
```
yuki-stripe, line 305: Can you delete this arg from `func intentCreation`? 

Also nit the func should probably be named `createIntent`

yuki-stripe (dismissed):

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetGDPRConfirmFlowTests.swift
wooj-stripe: Yep. It can be deleted since it is no longer being used since elementsSessions is no longer attached to the `Intent`

yuki-stripe (approved):

PR #3861, opened 2024-07-30
Financial Connections: for networking manual entry, added support for mid-flow consent params

kgaidis-stripe:
## Summary

This PR:
- This PR adds support for "mid flow consent." IF a user goes through the "Return User Experience [RUX]" by tapping "manual entry" in consent, then we will show an "Agree and continue" button in RUX/Link Account Picker.
- The visual changes are run by the backend.
- This PR specifically passes on some parameters from API calls: it gets `acquireConsentOnPrimaryCtaClick`, and uses it to pass into `selectNetworkedAccounts`.
- We are replicating [this Stripe.js PR](https://git.corp.stripe.com/stripe-internal/stripe-js-v3/pull/23488)

Context:
- Financial Connections is implementing support for "networking manual entry." This feature allows a user to enter their manually entered account _once_, and then we will allow them to reuse it later through Link. 
- This is just one PR of many where each PR will be merged to a feature branch `fc-networkingmanualentry`.
- It's expected that this PR may have some gaps or TODO's because its going to a feature branch. That being said, we still want to make sure there's no glaring logic issues/bugs. 

## Testing

API param is returned from response:
acquireConsentOnClick

API param is passed into response:
consent_acquired_passed_in

Video of the experience:

https://github.com/user-attachments/assets/b99c847b-d5de-43b3-820e-493b7e6fa38b

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
@@ -115,7 +115,8 @@ protocol FinancialConnectionsAPIClient {
     func selectNetworkedAccounts(
         selectedAccountIds: [String],
         clientSecret: String,
-        consumerSessionClientSecret: String
+        consumerSessionClientSecret: String,
+        consentAcquired: Bool?
     ) -> Future<FinancialConnectionsInstitutionList>
 
     func markLinkStepUpAuthenticationVerified(
@@ -659,13 +660,15 @@ extension STPAPIClient: FinancialConnectionsAPIClient {
     func selectNetworkedAccounts(
         selectedAccountIds: [String],
         clientSecret: String,
-        consumerSessionClientSecret: String
+        consumerSessionClientSecret: String,
+        consentAcquired: Bool?
     ) -> Future<FinancialConnectionsInstitutionList> {
-        let parameters: [String: Any] = [
+        var parameters: [String: Any] = [
             "selected_accounts": selectedAccountIds,
             "client_secret": clientSecret,
             "consumer_session_client_secret": consumerSessionClientSecret,
         ]
+        parameters["consent_acquired"] = consentAcquired
         return post(resource: APIEndpointShareNetworkedAccount, parameters: parameters)
     }
 
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/Models/FinancialConnectionsNetworkedAccountsResponse.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/Models/FinancialConnectionsNetworkedAccountsResponse.swift
@@ -11,6 +11,7 @@ struct FinancialConnectionsNetworkedAccountsResponse: Decodable {
     let data: [FinancialConnectionsPartnerAccount]
     let display: Display?
     let nextPaneOnAddAccount: FinancialConnectionsSessionManifest.NextPane?
+    let acquireConsentOnPrimaryCtaClick: Bool?
 
     struct Display: Decodable {
         let text: Text?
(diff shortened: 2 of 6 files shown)


StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/Models/FinancialConnectionsNetworkedAccountsResponse.swift
```diff
@@ -11,6 +11,7 @@ struct FinancialConnectionsNetworkedAccountsResponse: Decodable {
     let data: [FinancialConnectionsPartnerAccount]
     let display: Display?
     let nextPaneOnAddAccount: FinancialConnectionsSessionManifest.NextPane?
+    let acquireConsentOnPrimaryCtaClick: Bool?
```
kgaidis-stripe, line 14: [backend definition](https://livegrep.corp.stripe.com/view/stripe-internal/pay-server/lib/bank_connections/api/networking/resources/networked_accounts_list_resource.rb#L43)

StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
```diff
@@ -659,13 +660,15 @@ extension STPAPIClient: FinancialConnectionsAPIClient {
...
+        consumerSessionClientSecret: String,
+        consentAcquired: Bool?
     ) -> Future<FinancialConnectionsInstitutionList> {
-        let parameters: [String: Any] = [
+        var parameters: [String: Any] = [
             "selected_accounts": selectedAccountIds,
             "client_secret": clientSecret,
             "consumer_session_client_secret": consumerSessionClientSecret,
         ]
+        parameters["consent_acquired"] = consentAcquired
```
kgaidis-stripe, line 671: [example Stripe.js](https://git.corp.stripe.com/stripe-internal/stripe-js-v3/pull/23488/files#diff-60871d17beee6942de7c86d06c3f88155b664634dd06b96c30c0e2479e6e7fb4R1001)

StripeFinancialConnections/StripeFinancialConnections/Source/Native/NetworkingLinkStepUpVerification/NetworkingLinkStepUpVerificationDataSource.swift
```diff
@@ -67,7 +67,8 @@ final class NetworkingLinkStepUpVerificationDataSourceImplementation: Networking
         return apiClient.selectNetworkedAccounts(
             selectedAccountIds: selectedAccountIds,
             clientSecret: clientSecret,
-            consumerSessionClientSecret: consumerSession.clientSecret
+            consumerSessionClientSecret: consumerSession.clientSecret,
+            consentAcquired: nil
```
kgaidis-stripe, line 71: this NULL is fine; there will be a separate PR that covers the "step up verification case"

StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
kgaidis-stripe: This is essentially step #2 of this PR: use `acquireConsentOnPrimaryCtaClick` and pass it as `consentAcquired`

mats-stripe: Should we do something like this, rather than possibly passing `nil` to the parameter?
```suggestion
        if let consentAcquired {
                parameters["consent_acquired"] = consentAcquired
        }
```

mats-stripe (approved):
LGTM, just one minor suggestion

StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
kgaidis-stripe: Do you have a example/source for it? I've always believed a `nil` is a way of clearing keys

One case where it does not is assigning an object: ex. NSNull object

I tried to pull in some docs (https://developer.apple.com/documentation/swift/dictionary):

Screenshot 2024-08-01 at 10 33 15 AM

kgaidis-stripe:
merging to keep the train going

PR #3868, opened 2024-07-31
Add StripeConnect module.

nschris-stripe:
## Summary
This is the initial commit for the StripeConnect module. Follow up PRs will add initial setup for snapshot tests as well as localization.

https://jira.corp.stripe.com/browse/MXMOBILE-2439

### DIFF
--- a/Stripe.xcworkspace/contents.xcworkspacedata
+++ b/Stripe.xcworkspace/contents.xcworkspacedata
@@ -48,6 +48,9 @@
       location = "group:StripeCardScan/StripeCardScan.xcodeproj">
    </FileRef>
    <FileRef
+      location = "group:StripeConnect/StripeConnect.xcodeproj">
+   </FileRef>
+   <FileRef
       location = "group:StripeCore/StripeCore.xcodeproj">
    </FileRef>
    <FileRef
--- a/StripeConnect/StripeConnect.xcodeproj/project.pbxproj
+++ b/StripeConnect/StripeConnect.xcodeproj/project.pbxproj
@@ -1,0 +1,414 @@
+// !$*UTF8*$!
+{
+	archiveVersion = 1;
+	classes = {
+	};
+	objectVersion = 56;
+	objects = {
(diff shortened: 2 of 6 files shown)


StripeConnect/StripeConnect/StripeConnect.docc/StripeConnect.md
```diff
@@ -0,0 +1 @@
+# ``StripeConnect``
```
mludowise-stripe, line 1: @porter-stripe Do we now use `.docc` files in the SDKs? I'm guessing this might have been auto-generated by Xcode and we might want to delete it since we do our own docs generation.

mludowise-stripe (commented):
Open question around the .docc directory, otherwise LGTM

StripeConnect/StripeConnect/StripeConnect.docc/StripeConnect.md
mludowise-stripe: I see that PaymentSheet has an empty placeholder for its `.docc` directory:
https://github.com/stripe/stripe-ios/blob/e9b3c5bb/StripePaymentSheet/StripePaymentSheet/Docs.docc/StripePaymentSheet.md

If Xcode needs this for something, we can probably keep it (maybe rename to `Docs.docc` it to be consistent with the other modules?) Otherwise I think we should remove it.

nschris-stripe: I removed it. I checked other SDKs and it looks like any that have the `.docc` have no content in them.

mludowise-stripe (approved):

StripeConnect/StripeConnect/StripeConnect.docc/StripeConnect.md
porter-stripe: Sounds good, yeah they are auto generated

porter-stripe (approved):

PR #3869, opened 2024-07-31
Fix some test code erroneously using background queue & race condition in Async.swift

yuki-stripe:
## Summary
I noticed these snapshot tests were flakey b/c they're using the background queue, which is unreliable & can take seconds (I saw up to 6 seconds!) to execute a work item. I originally just made them use `.userInteractive` but realized they could be simplified more.

I ran the CustomerSheetSnapshotTests tests 20 times in a row and realized there's still a flake - a race condition in `Async.swift` when you observe at the ~same time you set result! I put access to its two properties, `result` and `callbacks`, behind a serial queue.  The Promise's `observe` callback is now called on the main thread by default rather than the thread of the caller that set `result`.

Finally, running PayymentSheetSnapshotTests 20 times in a row showed a bizarre crash where https://github.com/stripe/stripe-ios/blob/e9b3c5bb/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift#L316 crashes accessing `paymentMethod.type` b/c `paymentMethod` is `Swift.__EmptyDictionarySingleton` instead of `STPPaymentMethodType`. I "fixed" that too, in `STPAPIClient+Payments.swift`

## Testing
Ran the snapshot tests 40 times in a row.  Before, it would consistently flake. Now it doesn't.
```
Test Suite 'CustomerSheetSnapshotTests' passed at 2024-07-31 12:13:54.387.
	 Executed 1720 tests, with 0 failures (0 unexpected) in 701.713 (702.001) seconds
```

## Changelog
This did fix a real bug but it's hard to know if it has an impact outside of tests.

### DIFF
--- a/StripeIdentity/StripeIdentityTests/Unit/NativeComponents/ViewControllers/DocumentCaptureViewControllerTest.swift
+++ b/StripeIdentity/StripeIdentityTests/Unit/NativeComponents/ViewControllers/DocumentCaptureViewControllerTest.swift
@@ -11,7 +11,6 @@ import Foundation
 @_spi(STP) import StripeCameraCoreTestUtils
 @_spi(STP) import StripeCore
 @_spi(STP) import StripeCoreTestUtils
-import StripeCoreTestUtils
 import XCTest
 
 @testable@_spi(STP) import StripeCameraCore
@@ -38,7 +37,7 @@ final class DocumentCaptureViewControllerTest: XCTestCase {
 
     let mockError = NSError(domain: "mock_error", code: 100, userInfo: nil)
 
-    override class func setUp() {
+    override static func setUp() {
         super.setUp()
         mockSampleBuffer = CapturedImageMock.frontDriversLicense.image.convertToSampleBuffer()
         guard let mockVerificationPage = try? VerificationPageMock.response200.make() else {
@@ -329,9 +328,14 @@ final class DocumentCaptureViewControllerTest: XCTestCase {
         // Request to save data
         vc.saveOrFlipDocument(scannedImage: mockBackImage, documentSide: .back)
 
-        guard case .success = mockSheetController.backUploadedDocumentsResult else {
-            return XCTFail("Expected success result")
+        let e = expectation(description: "back upload result")
+        mockDocumentUploader.backUploadPromise.observe { _ in
+            guard case .success = self.mockSheetController.backUploadedDocumentsResult else {
+                return XCTFail("Expected success result")
+            }
+            e.fulfill()
         }
+        waitForExpectations(timeout: 1)
 
         // Verify state
         verify(
--- a/StripeIdentity/StripeIdentityTests/Unit/NativeComponents/ViewControllers/DocumentCaptureViewControllerTest.swift
+++ b/StripeIdentity/StripeIdentityTests/Unit/NativeComponents/ViewControllers/DocumentCaptureViewControllerTest.swift
@@ -302,10 +302,14 @@ final class DocumentCaptureViewControllerTest: XCTestCase {
 
         // Request to save data
         vc.saveOrFlipDocument(scannedImage: mockFrontImage, documentSide: .front)
-
-        guard case .success = mockSheetController.frontUploadedDocumentsResult else {
-            return XCTFail("Expected success result")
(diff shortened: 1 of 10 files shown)


StripeIdentity/StripeIdentityTests/Unit/NativeComponents/ViewControllers/DocumentCaptureViewControllerTest.swift
```diff
@@ -38,7 +37,7 @@ final class DocumentCaptureViewControllerTest: XCTestCase {
 
     let mockError = NSError(domain: "mock_error", code: 100, userInfo: nil)
 
-    override class func setUp() {
+    override static func setUp() {
```
wooj-stripe, line 40: Curious-- did changing this from `class` to `static` change anything?

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetSnapshotTests.swift
```diff
@@ -1159,44 +1159,31 @@ class PaymentSheetSnapshotTests: STPSnapshotTestCase {
...
-        DispatchQueue.global(qos: .background).async {
-            var count = 0
-            while count < maxCount {
-                count += 1
-                usleep(500000)  // 500ms
-            }
+        DispatchQueue.main.asyncAfter(deadline: .now() + numSeconds) {
             waitExpectation.fulfill()
         }
         wait(for: [waitExpectation], timeout: 10.0)
```
wooj-stripe, line 1189: nit: We should probably remove the timeout.

StripePayments/StripePayments/Source/API Bindings/STPAPIClient+Payments.swift
```diff
@@ -920,7 +920,7 @@ extension STPAPIClient {
         )
 
         var params = [
-            "app": String(data: appData ?? Data(), encoding: .utf8) ?? "",
+            "app": String(decoding: appData ?? Data(), as: UTF8.self),
```
wooj-stripe, line 923: I've always used `String(data:`.. curious what the change here does?

StripePayments/StripePayments/Source/API Bindings/STPAPIClient+Payments.swift
```diff
@@ -1145,7 +1145,8 @@ extension STPAPIClient {
                         shared_lastError = error
                     }
                     if let paymentMethods = deserializer?.paymentMethods {
-                        shared_allPaymentMethods.append(contentsOf: paymentMethods)
+                        // For unknown reasons, `append(contentsOf:` here sometimes causes an EXC_BAD_INSTRUCTION if you repeatedly run tests
+                        paymentMethods.forEach { shared_allPaymentMethods.append($0) }
```
wooj-stripe, line 1149: since these are both of type `[STPPaymentMethod]` and `shared_allPaymentMethods` is a var, maybe we can avoid the `forEach` by using:
```
shared_allPaymentMethods += paymentMethods
```

StripeIdentity/StripeIdentityTests/Unit/NativeComponents/ViewControllers/DocumentCaptureViewControllerTest.swift
yuki-stripe: Doesn't change anything here, the linter made me

StripePayments/StripePayments/Source/API Bindings/STPAPIClient+Payments.swift
yuki-stripe: The new initializer is non-optional. The linter made do this 🤷. Maybe I have a different version from CI :(

StripePayments/StripePayments/Source/API Bindings/STPAPIClient+Payments.swift
yuki-stripe: Maybe, I didn't try that. But I don't want to re-run the tests 20 times to find out if this is flakey or not >_>

wooj-stripe (approved):

PR #3871, opened 2024-08-01
Add tests configuration for StripeConnect.

nschris-stripe:
## Summary
Follow up to #3868. This PR adds StripeCore as a dependency to StripeConnect and configures tests for StripeConnect.

https://jira.corp.stripe.com/browse/MXMOBILE-2439

PR #3873, opened 2024-08-01
Add localization for StripeConnect.

nschris-stripe:
## Summary
Follow up to #3868 & #3871 . This adds the necessary Localizable.strings file for every language connect embedded components supports ([Reference](https://docs.stripe.com/connect/get-started-connect-embedded-components#supported-languages)).

https://jira.corp.stripe.com/browse/MXMOBILE-2439


StripeConnect/Resources/Localizations/en.lproj/Localizable.strings
mludowise-stripe, line 1: Do we need `en.lproj` if we're supporting `en-US`?

I think we can select `en-US` as the development language in info.plist and remove this -- or we can drop `en-US` so that the `en.lproj` becomes the default for any unspecified english locales.

One thing I'm not sure of is what would happen if the app supports `en` or `fr` (without the region specified) and the user's locale is not one of the ones listed here (e.g. `en-CA` or `fr-MA`). Will the user's device think that the language is unsupported and therefore default to the developer language (English) or will it know to pick the closest region?

It's possible we might need to pick one locale as a default for each language (e.g. convert `en-US` -> `en` and `fr-FR` -> `fr`), but I'm not certain without testing. We could always merge this as-is then fix it later after testing.

mludowise-stripe (approved):
One fixit re `en` vs `en-US` + an open question

StripeConnect/Resources/Localizations/en.lproj/Localizable.strings
mludowise-stripe: For context – we don't plan to include any localized strings for the Connect SDK, but we do need to define the supported languages so that...
1. iOS will select the correct app language based on an intersection of:
   - The platform app's supported languages
   - The SDK's supported languages
   - The user's preferred languages
2. Forward the selected locale onto the Connect.js web SDK for localization

nschris-stripe: My takes on this is that we should just fix later if we need. I think there is a benefit to being explicit about the languages/regions we support. Unless the region selection process is documented somewhere I would rather not rely on current implementation. The only benefit I would see to not defining `en` directly is that we would save a little on size. That would really only be a benefit if we were planning to add many strings, which it sounds like we do not have plans to do.

PR #3876, opened 2024-08-01
Make vertical mode UI dim when disabled

yuki-stripe:
## Testing
See snapshot

Simulator Screen Recording -  snapshot tester  iPhone 12 mini 16 4 - 2024-08-01 at 11 39 48

## Changelog
Not shipped


porter-stripe:
It looks strange to me that the entire UI is dimmed, but the processing button isn't either, especially when it is disable too. Is this intentional? Also the wallet (Apple Pay and Link) buttons don't dim.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/WalletHeaderView.swift
```diff
@@ -182,3 +182,17 @@ extension PaymentSheetViewController {
...
 }
+
+// MARK: - EventHandler
+extension PaymentSheetViewController.WalletHeaderView: EventHandler {
+    func handleEvent(_ event: STPEvent) {
+        switch event {
+        case .shouldEnableUserInteraction:
+            alpha = 1
+        case .shouldDisableUserInteraction:
+            alpha = 0.8
```
porter-stripe, line 193: How'd we land on 0.6 for the rest of the UI for 0.8 here?

porter-stripe: I see some 0.5's too

porter-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/WalletHeaderView.swift
yuki-stripe: I just eyeballed it. Will get Hannah to review next week.

PR #3877, opened 2024-08-01
Support payment_method_save_allow_redisplay_override

wooj-stripe:
## Summary
Adds support for payment_method_save_allow_redisplay_override

## Motivation
Need the ability to override the allow_redisplay value when payment_method_save is false for PI+SFU & SI.

## Testing
Added confirmation tests


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/IntentConfirmParams.swift
```diff
@@ -133,7 +134,11 @@ class IntentConfirmParams {
         } else {
             if isSettingUp {
                 // Checkbox is hidden
-                paymentMethodParams.allowRedisplay = .limited
+                if let allowRedisplayOverride {
+                    paymentMethodParams.allowRedisplay = allowRedisplayOverride
+                } else {
+                    paymentMethodParams.allowRedisplay = .limited
+                }
```
yuki-stripe, line 141:
```suggestion
                paymentMethodParams.allowRedisplay = allowRedisplayOverride ?? .limited
```
nit

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/CustomerSession.swift
```diff
@@ -78,6 +85,7 @@ struct PaymentSheetComponent: Equatable, Hashable {
 struct PaymentSheetComponentFeature: Equatable, Hashable {
     let paymentMethodSave: Bool
     let paymentMethodRemove: Bool
+    let paymentMethodSaveAllowRedisplayOverride: STPPaymentMethodAllowRedisplay?
```
yuki-stripe, line 88: Can we document this somehow? Maybe just have a docstring for `PaymentSheetComponentFeature ` that links to the api https://docs.corp.stripe.com/api/customer_sessions/object#customer_session_object-components-payment_sheet

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/IntentConfirmParams.swift
```diff
@@ -133,7 +134,11 @@ class IntentConfirmParams {
         } else {
             if isSettingUp {
                 // Checkbox is hidden
```
yuki-stripe, line 136: Should we assert that `saveForFutureUseCheckboxState` is `.hidden` here?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/IntentConfirmParams.swift
```diff
@@ -108,6 +108,7 @@ class IntentConfirmParams {
         }
     }
     func setAllowRedisplay(paymentMethodSave: Bool?,
+                           allowRedisplayOverride: STPPaymentMethodAllowRedisplay?,
```
yuki-stripe, line 111: nit: I think this method could be called with invalid params, right? Like `setAllowRedisplay(paymentMethodSave: false, allowRedisplayOverride: .limited)`

Maybe it should take a `CustomerSession?` to eliminate that

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/IntentConfirmParamsTest.swift
```diff
@@ -15,23 +15,23 @@ class IntentConfirmParamsTest: XCTestCase {
...
+        intentConfirmParams.setAllowRedisplay(paymentMethodSave: nil, allowRedisplayOverride: nil, isSettingUp: false)
 
         XCTAssertEqual(.unspecified, intentConfirmParams.paymentMethodParams.allowRedisplay)
     }
     func testSetAllowRedisplay_legacyPI_deselected() {
         let intentConfirmParams = IntentConfirmParams(type: .stripe(.card))
 
         intentConfirmParams.saveForFutureUseCheckboxState = .deselected
-        intentConfirmParams.setAllowRedisplay(paymentMethodSave: nil, isSettingUp: false)
+        intentConfirmParams.setAllowRedisplay(paymentMethodSave: nil, allowRedisplayOverride: nil, isSettingUp: false)
```
yuki-stripe, line 34: Very much a rough thought but: these tests deeply bug me but I'm not exactly sure how to explain why and I don't have the knowledge to suggest something better. Like, we're just testing every possible combination of parameters right?

yuki-stripe: Maybe we do bite the bullet of testing all input combinations, in which case perhaps we should at least do like (copied from https://stackoverflow.com/questions/35441124/parametrized-unit-tests-in-swift):

```
func testDivision() {
    let testCases = [
        (a: 12, b: 3, result: 4),
        (a: 12, b: 2, result: 6),
        (a: 12, b: 6, result: 1),
        (a: 12, b: 4, result: 3),
    ]
    for (a, b, result) in testCases {
        XCTContext.runActivity(named: "Testing dividing \(a) by \(b) to get result \(result)") { activity in
            XCTAssertEqual(result, a/b)
        }
    }
}
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/IntentConfirmParams.swift
wooj-stripe: Sure, i will revisit this in a subsequent PR when hiding the checkbox when payment_method_save is set to disabled.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/IntentConfirmParams.swift
wooj-stripe: Good point, we can address this by passing in an optional "PaymentSheetComponentFeature".  Will address in subsequent PR when hiding checkbox.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/IntentConfirmParamsTest.swift
wooj-stripe: We are testing all use cases that we support (and two cases that technically shouldn't happen, but we are testing that our code is defensive enough to handle it).

Separating out each test into a separate function allows for narrowing down what exact use case is broken.

wooj-stripe:
public api is a false positive.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/IntentConfirmParams.swift
wooj-stripe: Addressed here:
https://github.com/stripe/stripe-ios/pull/3884

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/IntentConfirmParams.swift
wooj-stripe: Added assert here:
https://github.com/stripe/stripe-ios/pull/3884

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/IntentConfirmParamsTest.swift
yuki-stripe: Yeah I guess my suggestion is that the example above is IMO easier to read than separate testcases, and the `XCTContext.runActivity(named:` line lets you know the exact test case that breaks.

yuki-stripe (approved):

PR #3879, opened 2024-08-01
[BANKCON-11476] Refactor FinancialConnectionsAPIClient to support using consumer pub key

mats-stripe:
Resolves: https://jira.corp.stripe.com/browse/BANKCON-11476

## Summary

This is a refactor of our `FinancialConnectionsAPIClient` from a protocol to a class such that it can switch between using the merchant's publishable key and the consumer's publishable key. The conditions for using the consumer's publishable key are: 

```swift
/// Returns the `consumerPublishableKey` for scenarios where it is valid to do so. That is;
/// - `canUseConsumerKey` must be `true`. This is a flag passed in by each API request.
/// - `isLinkWithStripe` must be `true`. This represents whether we're in the Instant Debits flow.
/// - `consumerSession` must be verified. This represents whether we have a verified Link user.
func consumerPublishableKeyProvider(canUseConsumerKey: Bool) -> String? {
    guard canUseConsumerKey, isLinkWithStripe, consumerSession?.isVerified == true else {
        return nil
    }
    return consumerPublishableKey
}
```

As a rule of thumb, `useConsumerPublishableKeyIfNeeded` should be `true` for requests that can happen after the user is verified. However, there are some exceptions to this rules (such as the create payment method request).

## Motivation

Once a Link user is verified, we must use their publishable key for some requests. This enables that

## Testing

Unit tests added + `bitrise run financial-connections-stability-tests` passes!

## Changelog

N/a

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
@@ -8,49 +8,55 @@
 import Foundation
 @_spi(STP) import StripeCore
 
-class FinancialConnectionsAPIClient {
+final class FinancialConnectionsAPIClient {
     let backingAPIClient: STPAPIClient
 
     var isLinkWithStripe: Bool = false
     var consumerPublishableKey: String?
     var consumerSession: ConsumerSessionData?
 
-    static let shared = FinancialConnectionsAPIClient(apiClient: .shared)
-
     init(apiClient: STPAPIClient) {
         self.backingAPIClient = apiClient
     }
 
-    func consumerPublishableKeyProvider(canUseConsumerKey: Bool) -> String? {
+    /// Returns the `consumerPublishableKey` for scenarios where it is valid to do so. That is;
+    /// - `canUseConsumerKey` must be `true`. This is a flag passed in by each API request.
+    /// - `isLinkWithStripe` must be `true`. This represents whether we're in the Instant Debits flow.
+    /// - `consumerSession` must be verfied. This represents whether we have a verified Link user.
+    private func consumerPublishableKeyProvider(canUseConsumerKey: Bool) -> String? {
         guard canUseConsumerKey, isLinkWithStripe, consumerSession?.isVerified == true else {
             return nil
         }
         return consumerPublishableKey
     }
 
-    func get<T: Decodable>(
+    /// Passthrough to `STPAPIClient.get` which uses the `consumerPublishableKey` whenever it should be used.
+    /// As a rule of thumb, `useConsumerPublishableKeyIfNeeded` should be `true` for requests that happen after the user is verified.
+    /// However, there are some exceptions to this rules (such as the create payment method request).
+    private func get<T: Decodable>(
         resource: String,
         parameters: [String: Any],
-        canUseConsumerPublishableKey: Bool
+        useConsumerPublishableKeyIfNeeded: Bool
     ) -> Promise<T> {
-        let consumerPubKey = consumerPublishableKeyProvider(canUseConsumerKey: canUseConsumerPublishableKey)
+        let possibleConsumerPublishableKey = consumerPublishableKeyProvider(canUseConsumerKey: useConsumerPublishableKeyIfNeeded)
         return backingAPIClient.get(
             resource: resource,
             parameters: parameters,
-            ephemeralKeySecret: consumerPubKey
+            ephemeralKeySecret: possibleConsumerPublishableKey
         )
     }
 
-    func post<T: Decodable>(
+    /// Passthrough to `STPAPIClient.post` which uses the `consumerPublishableKey` whenever it should be used.
+    private func post<T: Decodable>(
         resource: String,
         parameters: [String: Any],
-        canUseConsumerPublishableKey: Bool
+        useConsumerPublishableKeyIfNeeded: Bool
     ) -> Promise<T> {
-        let consumerPubKey = consumerPublishableKeyProvider(canUseConsumerKey: canUseConsumerPublishableKey)
+        let possibleConsumerPublishableKey = consumerPublishableKeyProvider(canUseConsumerKey: useConsumerPublishableKeyIfNeeded)
         return backingAPIClient.post(
             resource: resource,
             parameters: parameters,
-            ephemeralKeySecret: consumerPubKey
+            ephemeralKeySecret: possibleConsumerPublishableKey
         )
     }
 }
@@ -217,7 +223,7 @@ extension FinancialConnectionsAPIClient: FinancialConnectionsAPI {
         return self.get(
             resource: APIEndpointListAccounts,
             parameters: parameters,
-            canUseConsumerPublishableKey: false
+            useConsumerPublishableKeyIfNeeded: false
         )
     }
 
@@ -225,7 +231,7 @@ extension FinancialConnectionsAPIClient: FinancialConnectionsAPI {
         return self.get(
             resource: APIEndpointSessionReceipt,
             parameters: ["client_secret": clientSecret],
-            canUseConsumerPublishableKey: false
+            useConsumerPublishableKeyIfNeeded: false
         )
     }
 
@@ -250,7 +256,7 @@ extension FinancialConnectionsAPIClient: FinancialConnectionsAPI {
         return self.post(
             resource: "financial_connections/sessions/synchronize",
             parameters: parameters,
-            canUseConsumerPublishableKey: true
+            useConsumerPublishableKeyIfNeeded: true
         )
     }
 
@@ -261,7 +267,7 @@ extension FinancialConnectionsAPIClient: FinancialConnectionsAPI {
         return self.post(
             resource: APIEndpointConsentAcquired,
             parameters: parameters,
-            canUseConsumerPublishableKey: false
+            useConsumerPublishableKeyIfNeeded: false
         )
     }
 
@@ -272,7 +278,7 @@ extension FinancialConnectionsAPIClient: FinancialConnectionsAPI {
         return self.get(
             resource: APIEndpointFeaturedInstitutions,
             parameters: parameters,
-            canUseConsumerPublishableKey: false
+            useConsumerPublishableKeyIfNeeded: false
         )
     }
 
@@ -285,7 +291,7 @@ extension FinancialConnectionsAPIClient: FinancialConnectionsAPI {
         return self.get(
             resource: APIEndpointSearchInstitutions,
             parameters: parameters,
-            canUseConsumerPublishableKey: false
+            useConsumerPublishableKeyIfNeeded: false
         )
     }
 
@@ -300,7 +306,7 @@ extension FinancialConnectionsAPIClient: FinancialConnectionsAPI {
         return self.post(
             resource: APIEndpointAuthSessions,
             parameters: body,
-            canUseConsumerPublishableKey: false
+            useConsumerPublishableKeyIfNeeded: false
         )
     }
 
@@ -312,7 +318,7 @@ extension FinancialConnectionsAPIClient: FinancialConnectionsAPI {
         return self.post(
             resource: APIEndpointAuthSessionsCancel,
             parameters: body,
-            canUseConsumerPublishableKey: false
+            useConsumerPublishableKeyIfNeeded: false
         )
     }
 
@@ -327,7 +333,7 @@ extension FinancialConnectionsAPIClient: FinancialConnectionsAPI {
         return self.post(
             resource: APIEndpointAuthSessionsRetrieve,
             parameters: body,
-            canUseConsumerPublishableKey: false
+            useConsumerPublishableKeyIfNeeded: false
         )
     }
 
@@ -348,7 +354,7 @@ extension FinancialConnectionsAPIClient: FinancialConnectionsAPI {
                 return self.post(
                     resource: APIEndpointAuthSessionsOAuthResults,
                     parameters: body,
-                    canUseConsumerPublishableKey: false
+                    useConsumerPublishableKeyIfNeeded: false
                 )
             },
             pollTimingOptions: APIPollingHelper<FinancialConnectionsMixedOAuthParams>.PollTimingOptions(
@@ -373,7 +379,7 @@ extension FinancialConnectionsAPIClient: FinancialConnectionsAPI {
         return self.post(
             resource: APIEndpointAuthSessionsAuthorized,
             parameters: body,
-            canUseConsumerPublishableKey: false
+            useConsumerPublishableKeyIfNeeded: false
         )
     }
 
@@ -397,7 +403,7 @@ extension FinancialConnectionsAPIClient: FinancialConnectionsAPI {
                 return self.post(
                     resource: APIEndpointAuthSessionsAccounts,
                     parameters: body,
-                    canUseConsumerPublishableKey: false
+                    useConsumerPublishableKeyIfNeeded: false
                 )
             },
             pollTimingOptions: APIPollingHelper<FinancialConnectionsAuthSessionAccounts>.PollTimingOptions(
@@ -421,7 +427,7 @@ extension FinancialConnectionsAPIClient: FinancialConnectionsAPI {
         return self.post(
             resource: APIEndpointAuthSessionsSelectedAccounts,
             parameters: body,
-            canUseConsumerPublishableKey: false
+            useConsumerPublishableKeyIfNeeded: false
         )
     }
 
@@ -432,7 +438,7 @@ extension FinancialConnectionsAPIClient: FinancialConnectionsAPI {
         return self.post(
             resource: APIEndpointLinkMoreAccounts,
             parameters: body,
-            canUseConsumerPublishableKey: false
+            useConsumerPublishableKeyIfNeeded: false
         )
     }
 
@@ -447,7 +453,7 @@ extension FinancialConnectionsAPIClient: FinancialConnectionsAPI {
         return self.post(
             resource: APIEndpointComplete,
             parameters: body,
-            canUseConsumerPublishableKey: true
+            useConsumerPublishableKeyIfNeeded: true
         )
         .chained { (session: StripeAPI.FinancialConnectionsSession) in
             if session.accounts.hasMore {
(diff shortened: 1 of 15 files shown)


StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
```diff
@@ -8,8 +8,54 @@
 import Foundation
 @_spi(STP) import StripeCore
 
-protocol FinancialConnectionsAPIClient {
+class FinancialConnectionsAPIClient {
```
kgaidis-stripe, line 11: `FinancialConnectionsAPIClient` is likely a good name because it matches the precedent of `STPAPIClient` 👍

kgaidis-stripe, line 11: Random analysis of whether this is the "simplest possible solution"...

Since it doesn't sound too feasible to "scatter" the logic of whether to use the new publishable key, then we must combine it.

The place where to combine it is somewhere in the API client. But right now our API client is a protocol. We can't(?) centralize logic in a protocol.

So we turn protocol into into a class to centralize logic.

...then its just a matter of asking how we shape the centralization logic (which is a bit separate from the first fundamental question of whether we "must" create a class/wrapper for the API)

kgaidis-stripe, line 11: We should consider to do some extra commenting/documenting so future devs can understand why this exists (ex. for Instant Debits)

StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
```diff
@@ -170,14 +216,16 @@ extension STPAPIClient: FinancialConnectionsAPIClient {
...
+            canUseConsumerPublishableKey: false
         )
     }
 
     func fetchFinancialConnectionsSession(clientSecret: String) -> Promise<StripeAPI.FinancialConnectionsSession> {
         return self.get(
             resource: APIEndpointSessionReceipt,
-            parameters: ["client_secret": clientSecret]
+            parameters: ["client_secret": clientSecret],
+            canUseConsumerPublishableKey: false
```
kgaidis-stripe, line 228: at first I was wondering whether we can avoid scattering `canUseConsumerPublishableKey` everywhere, but its likely a good idea to "force the developer to think" (...on whether this should be used or not) 👍 

that also goes to the note about commenting: maybe we should leave some basic guidelines/explanation of when one would want to set this or not (just some general guidance so a new dev is not completely confused what this param is as it is a bit ambigious)

StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
```diff
@@ -8,8 +8,54 @@
...
+        guard canUseConsumerKey, isLinkWithStripe, consumerSession?.isVerified == true else {
+            return nil
+        }
+        return consumerPublishableKey
+    }
+
+    func get<T: Decodable>(
+        resource: String,
+        parameters: [String: Any],
+        canUseConsumerPublishableKey: Bool
```
kgaidis-stripe, line 34: feel 100% free to ignore this:

maybe the name would be more idiomatic/clear if it was something like `useConsumerPublishableKeyIfNeeded`

the name I suggest is unfortunately more lengthy, however

I just use "if needed" because there's some Apple API's with it - ex. `layoutIfNeeded()`

`canUseConsumerPublishableKey` does say the same thing, but for some reason feels more confusing (though, again, feel 100% free to ignore my suggestion - not super bullish on my own name either)

StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
```diff
@@ -8,8 +8,54 @@
...
+    var consumerPublishableKey: String?
+    var consumerSession: ConsumerSessionData?
+
+    static let shared = FinancialConnectionsAPIClient(apiClient: .shared)
+
+    init(apiClient: STPAPIClient) {
+        self.backingAPIClient = apiClient
+    }
+
+    func consumerPublishableKeyProvider(canUseConsumerKey: Bool) -> String? {
```
kgaidis-stripe, line 24: if possible, mark private

StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
```diff
@@ -8,8 +8,54 @@
...
+            ephemeralKeySecret: consumerPubKey
+        )
+    }
+
+    func post<T: Decodable>(
+        resource: String,
+        parameters: [String: Any],
+        canUseConsumerPublishableKey: Bool
+    ) -> Promise<T> {
+        let consumerPubKey = consumerPublishableKeyProvider(canUseConsumerKey: canUseConsumerPublishableKey)
```
kgaidis-stripe, line 49: calling out different pattern of naming: 

`consumerPubKey` doesn't spell out the "publishable" while `canUseConsumerPublishableKey` does (I prefer the spelled out, personally [`conusmerPubKey` -> `consumerPublishableKey`])

StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
```diff
@@ -8,8 +8,54 @@
...
+    func get<T: Decodable>(
+        resource: String,
+        parameters: [String: Any],
+        canUseConsumerPublishableKey: Bool
+    ) -> Promise<T> {
+        let consumerPubKey = consumerPublishableKeyProvider(canUseConsumerKey: canUseConsumerPublishableKey)
+        return backingAPIClient.get(
+            resource: resource,
+            parameters: parameters,
+            ephemeralKeySecret: consumerPubKey
```
kgaidis-stripe, line 40: feel free to ignore 100% of this:

if you do end up changing some naming, the word "provider" sounds sort of ambiguous 

if some renaming is done, maybe this could all be written as:

```
        return backingAPIClient.get(
            resource: resource,
            parameters: parameters,
            ephemeralKeySecret: retrieveConsumerPublishableKeyIfNeeded(useConsumerPublishableKeyIfNeeded)
         )
```

some interesting things to note is that when one reads `ephemeralKeySecret: consumerPubKey` or `let consumerPubKey =` it makes it seem that "consumer publishable key" IS there because the "context of the computation of whether the key will be returned is elsewhere"

StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
```diff
@@ -8,8 +8,54 @@
...
 
+    var isLinkWithStripe: Bool = false
+    var consumerPublishableKey: String?
+    var consumerSession: ConsumerSessionData?
+
+    static let shared = FinancialConnectionsAPIClient(apiClient: .shared)
+
+    init(apiClient: STPAPIClient) {
+        self.backingAPIClient = apiClient
+    }
```
kgaidis-stripe, line 22: ⚠️ 

this part is likely the most critical part for this PR to think about

something feels a bit off about instantiation right now, and I wonder how we can clean this up so we can be "100%" sure that there is only one usage/instance of `FinancialConnectionsAPIClient`

---

it seems like `FinancialConnectionsAPIClient.shared` is only used in one place - SwiftUI previews? which maybe means we don't need it

---

then `FinancialConnectionsAPIClient(apiClient: apiClient)` is instantiated in `HostController`, but "the journey of using the API client" begins in `FinancialConnectionsSheet` - can we move the instantiation "up the stack to `FinancialConnectionsSheet` and then ensure it's 'never' done anywhere else? 

For example, maybe that also means that `STPAPIClient` should never be declared anywhere BUT `FinancialConnectionsSheet` (and inside of the new `FinancialConnectionsAPIClient` class).  (likely a grep should show that)

---

then another thing is that for some reason `apiClient` is public...what happens if a merchant swaps out `apiClient`? maybe it makes sense to have the instantiation of `FinancialConnectionsAPIClient` be done inside of `FinancialConnectionsSheet`, AND make sure that we re-instantiate it incase anyone changes `apiClient: STPAPIClient` with a public variable

this is very similar to what's happening with `APIVersion.configureFinancialConnectionsAPIVersion(apiClient: apiClient)`: its called in two places: `init` AND `didSet` of apiClient. likely the same has to be done for `FinancialConnectionsAPIClient`?

(small side-note: it does seem like we 'can't' "guarantee" that users of `FinancialConnectionsSheet` don't swap out the api client _after_ we already presented the sheet, but that's OK...that's already true today so we can just focus on the narrow/common use-case and treat that as user error if that ever does happen)

StripeFinancialConnections/StripeFinancialConnections/Source/Common/HostController.swift
```diff
@@ -67,7 +67,7 @@ class HostController {
         publishableKey: String?,
         stripeAccount: String?
     ) {
-        self.apiClient = apiClient
+        self.apiClient = FinancialConnectionsAPIClient(apiClient: apiClient)
```
kgaidis-stripe, line 70: related comment:

https://github.com/stripe/stripe-ios/pull/3879/files?diff=split&w=1#r1700961659

StripeFinancialConnections/StripeFinancialConnections/Source/Common/HostController.swift
```diff
@@ -112,7 +112,7 @@ extension HostController: HostViewControllerDelegate {
                 flow: flow,
                 killswitchActive: flowRouter.killswitchActive
             ),
-            apiClient: apiClient
+            apiClient: apiClient.backingAPIClient
```
kgaidis-stripe, line 115: I think `backingAPIClient` is a pretty good name 👍

StripeFinancialConnections/StripeFinancialConnections/Source/Native/NativeFlowController.swift
```diff
@@ -1001,7 +1001,7 @@ extension NativeFlowController: LinkLoginViewControllerDelegate {
         _ viewController: LinkLoginViewController,
         receivedLinkSignUpResponse linkSignUpResponse: LinkSignUpResponse
     ) {
-        // TODO(BANKCON-11476): Update apiClient to use new publishable key.
+        dataManager.consumerPublishableKey = linkSignUpResponse.publishableKey
```
kgaidis-stripe, line 1004: I don't know if this is necessary, but if you think there is some useful context of why we only set this here vs. other places, maybe worth adding

but, again, if there's no useful context then there's no need (ex. maybe we just set it here because this is one of the couple of places where it returns the key! then adding docs doesn't really add that much help)

StripeFinancialConnections/StripeFinancialConnections/Source/Native/NativeFlowController.swift
```diff
@@ -992,7 +992,7 @@ extension NativeFlowController: LinkLoginViewControllerDelegate {
         _ viewController: LinkLoginViewController,
         foundReturningUserWith lookupConsumerSessionResponse: LookupConsumerSessionResponse
     ) {
-        // TODO(BANKCON-11476): Save consumer publishable key for future use.
+        dataManager.consumerPublishableKey = lookupConsumerSessionResponse.publishableKey
```
kgaidis-stripe, line 995: I don't know if this is necessary, but if you think there is some useful context of why we only set this here vs. other places, maybe worth adding

but, again, if there's no useful context then there's no need (ex. maybe we just set it here because this is one of the couple of places where it returns the key! then adding docs doesn't really add that much help)

StripeFinancialConnections/StripeFinancialConnections/Source/Native/NativeFlowDataManager.swift
```diff
@@ -80,11 +81,22 @@ class NativeFlowAPIDataManager: NativeFlowDataManager {
     var errorPaneReferrerPane: FinancialConnectionsSessionManifest.NextPane?
     var paymentAccountResource: FinancialConnectionsPaymentAccountResource?
     var accountNumberLast4: String?
-    var consumerSession: ConsumerSessionData?
     var saveToLinkWithStripeSucceeded: Bool?
     var lastPaneLaunched: FinancialConnectionsSessionManifest.NextPane?
     var customSuccessPaneMessage: String?
 
+    var consumerSession: ConsumerSessionData? {
+        didSet {
```
kgaidis-stripe, line 89: didSet here is a good call!

StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/NetworkingOTPView/NetworkingOTPView.swift
```diff
@@ -257,7 +257,7 @@ private struct NetowrkingOTPViewRepresentable: UIViewRepresentable {
             connectionsMerchantName: nil,
             pane: .networkingLinkVerification,
             consumerSession: nil,
-            apiClient: STPAPIClient.shared,
+            apiClient: FinancialConnectionsAPIClient.shared,
```
kgaidis-stripe, line 260: relevant note: 

https://github.com/stripe/stripe-ios/pull/3879/files?diff=split&w=1#r1700961659

StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
kgaidis-stripe: could add `final` next to `class` here too for extra clarity

StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
mats-stripe: Good points - I'll move the initialization of `FinancialConnectionsAPIClient` to `FinancialConnectionsSheet`!

Thinking about how it should be setup on the `FinancialConnectionsSheet` (stored property vs computed property vs initialized inline when we initialized the `HostController`) - I don't think any approach is better than any other, knowing it needs to be re-initialized whenever `apiClient` changes. So, going with the "simplest approach", I'll probably just initialize it inline where `HostController` is created!

StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
mats-stripe: Will add some docs on this!

StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
mats-stripe: Will update to `useConsumerPublishableKeyIfNeeded`!

StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
mats-stripe: >some interesting things to note is that when one reads ephemeralKeySecret: consumerPubKey or let consumerPubKey = it makes it seem that "consumer publishable key" IS there because the "context of the computation of whether the key will be returned is elsewhere"

Makes sense! In these scenarios, I'm a fan of the `possible` nomenclature: 
```swift
let possibleConsumerPublishableKey = consumerPublishableKeyProvider(canUseConsumerKey: useConsumerPublishableKeyIfNeeded)
return backingAPIClient.get(
    resource: resource,
    parameters: parameters,
    ephemeralKeySecret: possibleConsumerPublishableKey
)
```

Let me know what you think!

kgaidis-stripe: yea all of that sounds good! this part isn't too critical and is very debatable haha

StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/Models/ConsumerSession/ConsumerSessionModels.swift
```diff
@@ -6,11 +6,38 @@
 //
 
 import Foundation
+@_spi(STP) import StripeCore
 
 struct ConsumerSessionData: Decodable {
     let clientSecret: String
     let emailAddress: String
     let redactedFormattedPhoneNumber: String
+    let verificationSessions: [VerificationSession]
```
kgaidis-stripe, line 15: you might have done this, but worth doing a double-check that this property is non-null in server

mats-stripe: [required](https://stripe.sourcegraphcloud.com/git.corp.stripe.com/stripe-internal/pay-server/-/blob/lib/consumer/api/resources/auth/consumer_session_resource.rb?L23)!

StripeFinancialConnections/StripeFinancialConnections/Source/FinancialConnectionsSheet.swift
```diff
@@ -211,8 +211,9 @@ final public class FinancialConnectionsSheet {
             }
         }
 
+        let financialConnectionsApiClient = FinancialConnectionsAPIClient(apiClient: apiClient)
```
kgaidis-stripe, line 214: oh so this is OK to do (you don't need code in `didSet`) _because_ this code only gets executed when we present the sheet (not when we initialize the sheet), so the presence of `apiClient` being a `public` key at this point (and its chance of being updated) doesn't matter 👍

kgaidis-stripe (approved):
nice! we should be good given that "bitrise run financial-connections-stability-tests passes" too 💯

PR #3883, opened 2024-08-02
Refactor STPAnalyticsClient+PaymentSheet

yuki-stripe:
## Summary
Replaces `STPAnalyticsClient+PaymentSheet.swift` log functions w/ `PaymentSheetAnalyticsHelper`

I also made “locale” a globally sent parameter, instead of just PaymentSheet

## Motivation
I was adding analytics parity to vertical mode and tried to clean up existing analytics code first.

Some problems w/ the old code:
#### 1. Inconsistent analytics parameters
There are a number of parameters that have specific meanings across PaymentSheet analytics and should be included consistently e.g. `linkEnabled`, `active_link_session`, `mpe_config`, `is_decoupled`, etc. This wasn't enforced at all today.

#### 2. Huge functions with lots of duplicate parameters
Before:
```
        STPAnalyticsClient.sharedClient.logPaymentSheetShow(
            isCustom: true,
            paymentMethod: mode.analyticsValue,
            linkEnabled: isLinkEnabled,
            activeLinkSession: LinkAccountContext.shared.account?.sessionState == .verified,
            currency: intent.currency,
            intentConfig: intent.intentConfig,
            apiClient: configuration.apiClient
        )
```

Related to (1), each function had tons of duplicate parameters - the intent was to make sure they sent the common parameters, but this was not a good way to do it. 

After:
```
        analyticsHelper.logShow(showingSavedPMList: mode == .selectingSaved)
```

AnalyticsHelper is initialized _once_ with things like `isCustom` and `configuration`, and then ensures it gets sent for all its analytics.

#### 3. AnalyticsHelper global singleton
We need to keep track of various bits of state across a PaymentSheet session, and we used this singleton. 

Now, we can put some of this state in a `PaymentSheetAnalyticsHelper` instance w/ the appropriate lifetime that is passed around. 

## Testing
Add unit tests

## Changelog
Not user facing


StripeCore/StripeCore/Source/Analytics/STPAnalyticEvent.swift
```diff
@@ -78,12 +78,8 @@ import Foundation
     // MARK: - PaymentSheet Show
     case mcShowCustomNewPM = "mc_custom_sheet_newpm_show"
     case mcShowCustomSavedPM = "mc_custom_sheet_savedpm_show"
-    case mcShowCustomApplePay = "mc_custom_sheet_applepay_show"
-    case mcShowCustomLink = "mc_custom_sheet_link_show"
     case mcShowCompleteNewPM = "mc_complete_sheet_newpm_show"
     case mcShowCompleteSavedPM = "mc_complete_sheet_savedpm_show"
-    case mcShowCompleteApplePay = "mc_complete_sheet_applepay_show"
-    case mcShowCompleteLink = "mc_complete_sheet_link_show"
```
porter-stripe, line 86: How are we thinking about how removing these events could impact dashboards? Do we not use them?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentOptionsViewController.swift
```diff
@@ -559,7 +572,7 @@ extension SavedPaymentOptionsViewController: PaymentOptionCellDelegate {
                                               error: Error.paymentOptionCellDidSelectRemoveOnNonSavedItem,
                                               additionalNonPIIParams: [
                                                 "indexPathRow": collectionView.indexPath(for: paymentOptionCell)?.row ?? "nil",
-                                                "viewModels": viewModels.map { $0.analyticsValue.rawValue },
+                                                "viewModels": viewModels.map { $0.analyticsValue },
```
porter-stripe, line 575: Similar, does this change existing metrics?

porter-stripe (approved):
Nice! Just a few non blocking Q's.

StripeCore/StripeCore/Source/Analytics/STPAnalyticEvent.swift
yuki-stripe: Oops, I commented on this exact question but didn't hit submit. We never sent these and never can. 'newpm' and 'savedpm' correspond to the two screens that PaymentSheet/FC can show.  There's no "Apple Pay" or "Link" screen.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentOptionsViewController.swift
yuki-stripe: It does not

PR #3889, opened 2024-08-05
Prevent cvc recollection callback from being called multiple times

wooj-stripe:
## Summary
The merchant supplied CVC Recollection called for deferred intents was being called twice in both of these scenarios:
* PaymentSheet.Complete (horizontal mode) -- Fix is in `PaymentSheetViewController.swift`
* PaymentSheet.Complete (vertical mode) -- Fix is in `PaymentSheet+API.swift` -- ensuring we check flowController prior to intent.cvcRecollectionEnabled.

## Motivation
There is no need to call the cvcRecollectionEnabled more than once in this context.

## Testing
Manual testing.

## Changelog
Updated


CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+## x.x.x y-y-y
```
porter-stripe, line 1: Add `### PaymentSheet` before the entry

porter-stripe (commented):
looks good, just a nit

porter-stripe (approved):

PR #3892, opened 2024-08-06
[BANKCON-11478] FC - Tweak warmup sheet for returning users

mats-stripe:
## Summary

This updates the Networking Link login warmup pane with the following change for the Instant Debits flow:

- Change secondary button text from `Not now` to `Cancel`.
- On tap of this button, dismiss the warmup pane rather than continuing without networking.

## Motivation

## Testing

Before:

https://github.com/user-attachments/assets/a3982abc-2980-4157-b99e-8afcfd3bbf33

After:

https://github.com/user-attachments/assets/60c1ffba-9454-4504-9b97-e475900720c2

## Changelog

N/a

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/Native/NetworkingLinkLoginWarmup/NetworkingLinkLoginWarmupViewController.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/Native/NetworkingLinkLoginWarmup/NetworkingLinkLoginWarmupViewController.swift
@@ -13,6 +13,9 @@ protocol NetworkingLinkLoginWarmupViewControllerDelegate: AnyObject {
     func networkingLinkLoginWarmupViewControllerDidSelectContinue(
         _ viewController: NetworkingLinkLoginWarmupViewController
     )
+    func networkingLinkLoginWarmupViewControllerDidSelectCancel(
+        _ viewController: NetworkingLinkLoginWarmupViewController
+    )
     func networkingLinkLoginWarmupViewController(
         _ viewController: NetworkingLinkLoginWarmupViewController,
         didSelectSkipWithManifest manifest: FinancialConnectionsSessionManifest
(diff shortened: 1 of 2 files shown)


tillh-stripe (dismissed):

StripeFinancialConnections/StripeFinancialConnections/Source/Native/NetworkingLinkLoginWarmup/NetworkingLinkLoginWarmupViewController.swift
```diff
@@ -95,29 +107,33 @@ final class NetworkingLinkLoginWarmupViewController: SheetViewController {
...
-                case .failure(let error):
-                    self.dataSource
-                        .analyticsClient
-                        .logUnexpectedError(
-                            error,
-                            errorName: "DisableNetworkingError",
-                            pane: .networkingLinkLoginWarmup
+        if dataSource.manifest.isProductInstantDebits {
+            delegate?.networkingLinkLoginWarmupViewControllerDidSelectCancel(self)
+        } else {
```
bric-stripe, line 112: non-blocking: any concern about a future change that accidentally didn't set the delegate when isProductInstantDebits which would then silently do nothing? 

worth a test for a nil delegate and fall through to the else case if nil? or do something to help debug if it did happen?

bric-stripe (dismissed):

StripeFinancialConnections/StripeFinancialConnections/Source/Native/NetworkingLinkLoginWarmup/NetworkingLinkLoginWarmupViewController.swift
mats-stripe: Hmm, that's a good point! To be honest, if the delegate is nil here I think it's better to do nothing than to fall back on the `else` case (since that behavior is incorrect for instant debits). I'll add some logging for this scenario though!

bric-stripe (approved):

PR #3901, opened 2024-08-07
Expand PaymentMethod in consumers/payment_details/share request inste…

yuki-stripe:
…ad of using dummy empty PaymentMethod

## Motivation
https://jira.corp.stripe.com/browse/RUN_MOBILESDK-3405

## Testing
Added assert and UI Test - it's indirect, but combined they do test what we want.  Directly testing this seems very hard, since IDK how to simulate the Link signup flow outside a UI test.

## Changelog
* [Fixed] Fixed an issue where signing up with Link and paying would vend an empty `STPPaymentMethod` object to an `IntentConfiguration` confirmHandler callback.

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
@@ -622,6 +622,11 @@ extension PlaygroundController {
     func confirmHandler(_ paymentMethod: STPPaymentMethod,
                         _ shouldSavePaymentMethod: Bool,
                         _ intentCreationCallback: @escaping (Result<String, Error>) -> Void) {
+        // Sanity check the payment method
+        if paymentMethod.type == .card {
+            assert(paymentMethod.card != nil)
+        }
+
         switch settings.integrationType {
         case .deferred_mp:
             // multiprocessor
--- a/Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
+++ b/Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
@@ -2216,6 +2216,17 @@ class PaymentSheetLinkUITests: PaymentSheetUITestCase {
 
         XCTAssertTrue(app.staticTexts["Success!"].waitForExistence(timeout: 10.0))
     }
+    
+    func testLinkInlineSignup_deferred() {
+        var settings = PaymentSheetTestPlaygroundSettings.defaultValues()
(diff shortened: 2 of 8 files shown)


Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
```diff
@@ -622,6 +622,11 @@ extension PlaygroundController {
     func confirmHandler(_ paymentMethod: STPPaymentMethod,
                         _ shouldSavePaymentMethod: Bool,
                         _ intentCreationCallback: @escaping (Result<String, Error>) -> Void) {
+        // Sanity check the payment method
+        if paymentMethod.type == .card {
+            assert(paymentMethod.card != nil)
+        }
```
yuki-stripe, line 628: Note this assertion fired in the `testLinkInlineSignup_deferred` added below before the change.

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/STPAPIClient+Link.swift
```diff
@@ -207,19 +207,26 @@ extension STPAPIClient {
         var parameters: [String: Any] = [
             "credentials": ["consumer_session_client_secret": consumerSessionClientSecret],
             "request_surface": "ios_payment_element",
+            "expand": ["payment_method"],
```
yuki-stripe, line 210: The ticket called for sending `expected_payment_method_type` but not sending it also works, which is good b/c IDK how we can tell what the value should be.

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
```diff
@@ -2265,6 +2276,11 @@ class PaymentSheetLinkUITests: PaymentSheetUITestCase {
             app.buttons["Continue"].tap()
             app.buttons["Confirm"].waitForExistenceAndTap()
         }
+        // Roundabout way to validate that signup completed successfully
+        let signupCompleteAnalytic = analyticsLog.first { payload in
+            payload["event"] as? String == "link.signup.complete"
+        }
+        XCTAssertNotNil(signupCompleteAnalytic)
```
yuki-stripe, line 2283: Signup fails silently (besides this analytic) so I'm not sure we checked signup was successful before this change!

davidme-stripe (approved):

davidme-stripe (approved):

davidme-stripe (approved):

PR #3902, opened 2024-08-07
[BANKCON-11812] Skip to nextPaneOnNewAccount if there are no accounts

mats-stripe:
## Summary

In order to avoid the scenario where we show the Link Account Picker pane with no account, we now push the `nextPaneOnAddAccount` pane automatically. We also hide the back button on that next pane to prevent the user from returning to an empty list of accounts. This PR also updates the Networking Link Verification pane to:
- No longer call the `mark_link_verified` endpoint during the instant debits flow
- Hardcode pushing to the Link Account Picker pane after a synchronize is complete (though we [might get this value from the server soon](https://stripe.slack.com/archives/C06UG8973J5/p1723055585416439?thread_ts=1723054458.490179&cid=C06UG8973J5))

## Motivation

Fixes some incorrect behavior for the instant debits flow, and 

## Testing

Here's a scenario where after verification, we push the Link Account Picker pane, but there are no accounts, so we then push to the Institution Picker pane:

https://github.com/user-attachments/assets/da46620d-57c5-4f76-97cc-ddbdff81664e

## Changelog

N/a - Instant Debits is not yet released.

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/Native/LinkAccountPicker/LinkAccountPickerViewController.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/Native/LinkAccountPicker/LinkAccountPickerViewController.swift
@@ -13,7 +13,8 @@ import UIKit
 protocol LinkAccountPickerViewControllerDelegate: AnyObject {
     func linkAccountPickerViewController(
         _ viewController: LinkAccountPickerViewController,
-        didRequestNextPane nextPane: FinancialConnectionsSessionManifest.NextPane
+        didRequestNextPane nextPane: FinancialConnectionsSessionManifest.NextPane,
+        hideBackButtonOnNextPane: Bool
     )
 
     func linkAccountPickerViewController(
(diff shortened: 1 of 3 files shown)


StripeFinancialConnections/StripeFinancialConnections/Source/Native/LinkAccountPicker/LinkAccountPickerViewController.swift
```diff
@@ -116,7 +117,15 @@ final class LinkAccountPickerViewController: UIViewController {
                 guard let self = self else { return }
                 switch result {
                 case .success(let networkedAccountsResponse):
-                    if let returningNetworkingUserAccountPicker = networkedAccountsResponse.display?.text?.returningNetworkingUserAccountPicker {
+                    if networkedAccountsResponse.data.isEmpty {
+                        let nextPaneOnAddAccount = dataSource.nextPaneOnAddAccount ?? .institutionPicker
```
mats-stripe, line 121: This is a value we get back when fetching networked accounts. It's [technically optional](https://stripe.sourcegraphcloud.com/git.corp.stripe.com/stripe-internal/pay-server/-/blob/lib/bank_connections/api/networking/resources/networked_accounts_list_resource.rb?L37-38), but seems to [always return a value](https://stripe.sourcegraphcloud.com/git.corp.stripe.com/stripe-internal/pay-server/-/blob/lib/bank_connections/api/networking/methods/link_account_session_networked_accounts_retrieve_method.rb?L177-196).

tillh-stripe (approved):
Looks good from a Financial Connections logic perspective!

bric-stripe (approved):

bric-stripe (approved):

PR #3908, opened 2024-08-08
Fix STPPaymentCardTextField cbc not calling didchange delegate

yuki-stripe:
## Summary
STPPaymentCardTextField didn't call its didChange delegate when you select a card brand.  I think it reasonably should, and for unclear reasons the RN SDK depends on didChange being called whenever the params changes. 

## Motivation
https://jira.corp.stripe.com/browse/RUN_MOBILESDK-3417

## Testing
Manually tested. 

## Changelog
See changelog

### DIFF
--- a/StripePaymentsUI/StripePaymentsUI/Source/Helpers/STPCBCController.swift
+++ b/StripePaymentsUI/StripePaymentsUI/Source/Helpers/STPCBCController.swift
@@ -146,11 +146,10 @@ class STPCBCController {
 
     var contextMenuConfiguration: UIContextMenuConfiguration {
         return UIContextMenuConfiguration(actionProvider: { _ in
-            let action = { (action: UIAction) -> Void in
+            let action = { (action: UIAction) in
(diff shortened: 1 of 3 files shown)


StripePaymentsUI/StripePaymentsUI/Source/Helpers/STPCBCController.swift
```diff
@@ -146,11 +146,10 @@ class STPCBCController {
 
     var contextMenuConfiguration: UIContextMenuConfiguration {
         return UIContextMenuConfiguration(actionProvider: { _ in
-            let action = { (action: UIAction) -> Void in
+            let action = { (action: UIAction) in
                 let brand = STPCard.brand(from: action.identifier.rawValue)
                 // Set the selected brand if a brand is selected
                 self.selectedBrand = brand != .unknown ? brand : nil
-                self.updateHandler?()
```
yuki-stripe, line 153: `selectedBrand`'s setter already calls `self.updateHandler?()`

davidme-stripe (approved):

PR #2729, opened 2023-07-12
Remove PaymentSheetError.unknown usage

porter-stripe:
## Summary
- Removes usage of `PaymentSheetError.unknown` in favor of specific errors
- `PaymentSheetError.unknown` was abused as a catch all error, now users will be encouraged to define more specific errors 

## Motivation
- Make errors more easily loggable into our analytics table
- Currently we are unable to identity errors by their domain or code since they all share the same given by `.unknown`.
- https://stripe.slack.com/archives/CFA2HJ99A/p1689180194110519

## Testing
Project compiles

## Changelog
N/A

PR #3918, opened 2024-08-09
Validate IntentConfiguration amount is nonzero

yuki-stripe:
Passing a non-zero amount is an integration error.

### Before
Fails at load time w/ a 'noPaymentMethodTypesAvailable' error because `v1/elements/sessions` returns an empty list of pm types. 

### After
In debug, we assert.

In production, we'll get a more accurate `intentConfigurationValidationFailed` error_code in the mc_load_failed analytic. Ideally we'd also get an error_message; I'll look into that in another PR.

## Testing
Manually tested.

## Changelog
See changelog

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetError.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetError.swift
@@ -63,66 +63,6 @@ public enum PaymentSheetError: Error, LocalizedError {
 }
 
 extension PaymentSheetError: CustomDebugStringConvertible {
-    /// A string that can safely be logged to our analytics service that does not contain any PII
-    public var safeLoggingString: String {
-        switch self {
-        case .unknown:
-            return "unknown"
-        case .missingClientSecret:
-            return "missingClientSecret"
-        case .invalidClientSecret:
-            return "invalidClientSecret"
-        case .unexpectedResponseFromStripeAPI:
-            return "unexpectedResponseFromStripeAPI"
-        case .applePayNotSupportedOrMisconfigured:
-            return "applePayNotSupportedOrMisconfigured"
-        case .alreadyPresented:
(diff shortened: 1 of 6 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetIntentConfiguration.swift
```diff
@@ -188,5 +193,15 @@ public extension PaymentSheet {
             // TODO
             self.isCVCRecollectionEnabledCallback = { return false }
         }
+
+        @discardableResult
+        func validate() -> Error? {
+            let errorMessage: String
+            if case .payment(let amount, _, _, _) = mode, amount > 0 {
```
wooj-stripe, line 200: Based on the error message, shouldn't this be `if amount <=0` ?

yuki-stripe: >_> yes

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetError.swift
```diff
@@ -62,66 +63,6 @@ public enum PaymentSheetError: Error, LocalizedError {
 }
 
 extension PaymentSheetError: CustomDebugStringConvertible {
-    /// A string that can safely be logged to our analytics service that does not contain any PII
-    public var safeLoggingString: String {
```
yuki-stripe, line 66: This should never have been public (added [here](https://github.com/stripe/stripe-ios/pull/2729/files#diff-9481030907897f02a164f3ed394a5c2e138b927a1d13fe0bec49a70956599f67R70)) so I'm adding the `modifies public API` label.

wooj-stripe (approved):

PR #3925, opened 2024-08-12
Move mandate to above call to action (v2)

wooj-stripe:
## Summary
Move mandate to above call to action

## Motivation
Legal ask

## Testing
Relying on existing tests

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/LinkEnabledPaymentMethodElement.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/LinkEnabledPaymentMethodElement.swift
@@ -10,7 +10,7 @@
 import UIKit
 
 final class LinkEnabledPaymentMethodElement: ContainerElement {
-    public lazy var elements: [Element] = { [paymentMethodElement, inlineSignupElement] }()
+    public lazy var elements: [Element] = { [inlineSignupElement] }()
 
     struct Constants {
         static let spacing: CGFloat = 12
@@ -24,7 +24,6 @@ final class LinkEnabledPaymentMethodElement: ContainerElement {
 
     private lazy var stackView: UIStackView = {
         let stackView = UIStackView(arrangedSubviews: [
-            paymentMethodElement.view,
             inlineSignupElement.view,
         ])
         stackView.axis = .vertical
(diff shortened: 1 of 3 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentMethodFormViewController.swift
```diff
@@ -24,14 +24,15 @@ class PaymentMethodFormViewController: UIViewController {
...
-        }
 
         let params = IntentConfirmParams(type: paymentMethodType)
         params.setDefaultBillingDetailsIfNecessary(for: configuration)
+
         if let params = form.updateParams(params: params) {
+            if let linkEnabledElement = form.getAllUnwrappedSubElements().compactMap({ $0 as? LinkEnabledPaymentMethodElement }).first {
+                return linkEnabledElement.makePaymentOption(intentConfirmParams: params)
+            }
+
```
wooj-stripe, line 35: Not sure if this is correct -- we were previously short circuiting setting default billing details if needed.  Seems like this was not intentional per the comment that I erased.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/LinkEnabledPaymentMethodElement.swift
```diff
@@ -35,38 +34,25 @@ final class LinkEnabledPaymentMethodElement: ContainerElement {
...
-        paymentMethodElement.delegate = self
         inlineSignupElement.delegate = self
     }
 
-    func makePaymentOption(intent: Intent, elementsSession: STPElementsSession) -> PaymentOption? {
-        guard let params = updateParams(params: .init(type: .stripe(paymentMethodType))) else {
-            return nil
-        }
-
+    func makePaymentOption(intentConfirmParams params: IntentConfirmParams) -> PaymentOption? {
```
yuki-stripe, line 55: I think we can move this method implementation to `LinkInlineSignupElement`'s `func updateParams(params: IntentConfirmParams) -> IntentConfirmParams?`.  That's the normal way that Elements update the IntentConfirmParams, I don't know why we did it differently here.

yuki-stripe: After that this class might not have a purpose anymore, it just contains `LinkInlineSignupElement`?

yuki-stripe (commented):
Love how simple this is!

yuki-stripe (approved):
🔥

PR #3932, opened 2024-08-14
[BANKCON-11874] Fix consumer publishable key usage for existing user flow

mats-stripe:
## Summary

In the Instant Debits flow, this fixes an issue for the existing user flow, where the consumer publishable key was not being used for API requests after the user had been verified. The change here is to propagate this publishable key back from the Networking OTP view, which we get from the `/consumer_session` API, to the `NativeFlowDataManager`, where it is set and used in subsequent API requests.

## Motivation

## Testing

Before: 

https://github.com/user-attachments/assets/76e02510-386f-40a0-a342-0f500e7d3e7b

After:

https://github.com/user-attachments/assets/3810389f-a879-423c-8880-2a84c781f6f7

## Changelog

N/a

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/Native/NativeFlowController.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/Native/NativeFlowController.swift
@@ -881,7 +881,7 @@ extension NativeFlowController: AttachLinkedPaymentAccountViewControllerDelegate
 // MARK: - NetworkingLinkVerificationViewControllerDelegate
 
 extension NativeFlowController: NetworkingLinkVerificationViewControllerDelegate {
-    func networkingLinkVerificationViewController(_ viewController: NetworkingLinkVerificationViewController, receivedConsumerPublishableKey consumerPublishableKey: String) {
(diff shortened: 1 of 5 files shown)


StripeFinancialConnections/StripeFinancialConnections/Source/Native/NativeFlowController.swift
```diff
@@ -881,6 +881,9 @@ extension NativeFlowController: AttachLinkedPaymentAccountViewControllerDelegate
 // MARK: - NetworkingLinkVerificationViewControllerDelegate
 
 extension NativeFlowController: NetworkingLinkVerificationViewControllerDelegate {
+    func networkingLinkVerificationViewController(_ viewController: NetworkingLinkVerificationViewController, receivedConsumerPublishableKey consumerPublishableKey: String) {
```
tillh-stripe, line 884: Nit: I guess this would need to be `didReceiveConsumerPublishableKey` to be consistent.

tillh-stripe (approved):

StripeFinancialConnections/StripeFinancialConnections/Source/Native/NativeFlowController.swift
mats-stripe: Updated - good call out!

bric-stripe (approved):

PR #3947, opened 2024-08-20
Dynamically set `request_surface` in FC API requests

mats-stripe:
## Summary

For all API requests to the `/consumers/...` path, we need to provide the `request_surface` body parameter. This is either `ios_instant_debits` if we're in the Native Instant Debits flow or `ios_connections` otherwise. This PR now dynamically sets that value according to the manifest's `isLinkWithStripe` field (set [here](https://stripe.sourcegraphcloud.com/github.com/stripe/stripe-ios/-/blob/StripeFinancialConnections/StripeFinancialConnections/Source/Native/NativeFlowDataManager.swift?L140)).

## Motivation

Consistency and correctness ✨ 

## Testing

N/a

## Changelog

N/a

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
@@ -15,6 +15,10 @@ final class FinancialConnectionsAPIClient {
     var consumerPublishableKey: String?
     var consumerSession: ConsumerSessionData?
 
+    var requestSurface: String {
+        isLinkWithStripe ? "ios_instant_debits" : "ios_connections"
+    }
+
     init(apiClient: STPAPIClient) {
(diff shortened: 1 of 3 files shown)


StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
```diff
@@ -589,7 +591,7 @@ extension FinancialConnectionsAPIClient: FinancialConnectionsAPI {
         return self.post(
             resource: APIEndpointAuthSessionsEvents,
             parameters: body,
-            useConsumerPublishableKeyIfNeeded: false
+            useConsumerPublishableKeyIfNeeded: true
```
mats-stripe, line 594: Drive-by change included in this PR: Use the consumer publishable key for this API request if needed.

tillh-stripe (dismissed):

PR #3948, opened 2024-08-20
Update PaymentSheet playground settings to save custom endpoint

mats-stripe:
## Summary

This fixes an issue I noticed with the PaymentSheet example app, where a custom checkout endpoint was not being included when sharing a configuration. As in, when launching the app by scanning the QR code, any custom endpoint would not be shared. This seems to be because the `checkoutEndpoint` property on the `PlaygroundController` was not using / updating the value in the `PlaygroundSettings` model.

## Motivation

This will make it a lot easier for bug bash testers when a custom endpoint is needed for testing!

## Testing

When setting a custom checkout endpoint: 

**Before:**
Generating a QR code for this configuration, the encoded settings string did not use the custom endpoint: (this is the default checkout endpoint)

```json
{
    "checkoutEndpoint": "https:\/\/stp-mobile-playground-backend-v7.stripedemos.com\/checkout"
}
```

This is a snippet of the whole encoded object, taken by putting a breakpoint in the [`base64Data` property](https://github.com/stripe/stripe-ios/blob/dce44782e444f0d8d88df9dc2bc17676caa22fe7/Example/PaymentSheet%20Example/PaymentSheet%20Example/PaymentSheetTestPlaygroundSettings.swift#L500-L503), then decoding the encoded string using [an online decoder](https://www.base64decode.org/).

**After:**

```json
{
    "checkoutEndpoint": "https:\/\/fc-instant-debits-playground.glitch.me\/checkout"
}
```

**Test it yourself!** 

1. Checkout and build this branch to a physical device.
2. Scan this QR code.
3. Verify the endpoint is set to https://fc-instant-debits-playground.glitch.me/checkout

image

## Changelog

N/a

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
@@ -283,10 +283,18 @@ class PlaygroundController: ObservableObject {
     var customerSessionClientSecret: String?
     var paymentMethodTypes: [String]?
     var amount: Int?
-    var checkoutEndpoint: String = PaymentSheetTestPlaygroundSettings.defaultCheckoutEndpoint
     var addressViewController: AddressViewController?
     var appearance = PaymentSheet.Appearance.default
     var currentDataTask: URLSessionDataTask?
+
+    var checkoutEndpoint: String {
+        get {
+            settings.checkoutEndpoint
+        }
+        set {
+            settings.checkoutEndpoint = newValue
+        }
+    }
 
     func makeAlertController() -> UIAlertController {
         let alertController = UIAlertController(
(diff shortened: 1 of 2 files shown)


Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
```diff
@@ -283,11 +283,19 @@ class PlaygroundController: ObservableObject {
...
     var paymentMethodTypes: [String]?
     var amount: Int?
-    var checkoutEndpoint: String = PaymentSheetTestPlaygroundSettings.defaultCheckoutEndpoint
     var addressViewController: AddressViewController?
     var appearance = PaymentSheet.Appearance.default
     var currentDataTask: URLSessionDataTask?
 
+    var checkoutEndpoint: String {
+        get {
+            settings.checkoutEndpoint
```
mats-stripe, line 292: This still has the same default value as before, which is set here: https://github.com/stripe/stripe-ios/blob/dce44782e444f0d8d88df9dc2bc17676caa22fe7/Example/PaymentSheet%20Example/PaymentSheet%20Example/PaymentSheetTestPlaygroundSettings.swift#L473

tillh-stripe, line 292: Should this be `settings.checkoutEndpoint ?? PaymentSheetTestPlaygroundSettings.defaultCheckoutEndpoint`?

tillh-stripe (approved):

porter-stripe (approved):

PR #3952, opened 2024-08-21
Add connect webview and connect component webview.

nschris-stripe:
## Summary
This adds the base web views that all connect components will utilize behind the scenes. It also adds and refines a few of our web view test helpers to be more reusable.

https://jira.corp.stripe.com/browse/MXMOBILE-2479
https://jira.corp.stripe.com/browse/MXMOBILE-2482

## Next PR
The next PR will add payouts view controller.


StripeConnect/StripeConnect/Source/Internal/Webview/ConnectWebView.swift
```diff
@@ -0,0 +1,187 @@
...
+        presentPopup(navController)
+        return popupVC.webView
+    }
+
+    // Opens the given URL in an SFSafariViewController
+    func openInAppSafari(url: URL) {
+        guard let presentPopup else {
+            assertionFailure("Cannot present popup")
+            return
+        }
```
mludowise-stripe, line 92: Optional:
Given that there are multiple callsites where we check for `presentPopup` and assertionFail, we could DRY this up into a helper:

```
private var optionalPresentPopup: ((UIViewController) -> Void)?

var presentPopup: (UIViewController) -> Void {
    get {
        assert(optionalPresentPopup != nil,  "Cannot present popup")
        return optionalPresentPopup ?? { _ in }
    }
    set {
        optionalPresentPopup = newValue
    }
}
```

StripeConnect/StripeConnectTests/Internal/Webview/ConnectWebViewTests.swift
```diff
@@ -0,0 +1,167 @@
...
+                return
+            }
+            
+            guard let userAgent = result as? String else {
+                XCTFail("User agent is not a string")
+                return
+            }
+                        
+            XCTAssertTrue(userAgent.hasSuffix("- stripe-ios/1.2.3"), "User agent should include the SDK identifier")
+            expectation.fulfill()
```
mludowise-stripe, line 44: nit: I think we want the expectation to fulfill, regardless of whether there's failures. XCTest continues to run the test, even after a failure occurs. So by returning early without fulfilling the expectation, the test can potentially take the full 2s to run before timing out.

You could do this by including a `defer { expectation.fulfill() }` or you could alternatively simplify the XCT checks:

```suggestion
            XCTAssertNil(error, "Error fetching user agent: \(error.localizedDescription)")
            XCTAssertEqual((result as? String).hasSuffix("- stripe-ios/1.2.3"), true, "User agent should include the SDK identifier but value was: \(String(describing: result))")
            expectation.fulfill()
```

^ I also included the user agent in the failure message. Then we won't have to insert breakpoints while running the test to figure out what the actual value.

StripeConnect/StripeConnectTests/Internal/Webview/ConnectWebViewTests.swift
```diff
@@ -0,0 +1,167 @@
...
+    }
+    
+    func testOpenSafariVCIfNotInAllowedHosts() {
+        var safariVC: SFSafariViewController?
+        webView.presentPopup = { vc in
+            safariVC = vc as? SFSafariViewController
+        }
+        let webView = webView.webView(webView,
+                        createWebViewWith: webView.configuration,
+                        for: MackNavigationAction(request: .init(url: URL(string: "https://google.com")!)),
```
mludowise-stripe, line 56: nit:
```suggestion
                        for: MackNavigationAction(request: .init(url: URL(string: "https://stripe.com")!)),
```

StripeConnect/StripeConnectTests/Internal/Webview/ConnectWebViewTests.swift
```diff
@@ -0,0 +1,167 @@
...
+                                      createWebViewWith: webView.configuration,
+                                      for: MackNavigationAction(request: .init(url: url), targetFrame: .init()),
+                                      windowFeatures: .init())
+        
+        XCTAssertNil(webView)
+    }
+}
+
+
+class MackNavigationAction: WKNavigationAction {
```
mludowise-stripe, line 137: typo?
```suggestion
class MockNavigationAction: WKNavigationAction {
```

StripeConnect/StripeConnectTests/Internal/Webview/ConnectWebViewTests.swift
```diff
@@ -0,0 +1,167 @@
...
+                
+            }
+            let webView = webView.webView(webView,
+                                          createWebViewWith: webView.configuration,
+                                          for: MackNavigationAction(request: .init(url: URL(string: domain)!)),
+                                          windowFeatures: .init())
+            
+            XCTAssertEqual(popUp?.webView, webView)
+            XCTAssertNotNil(popUp?.navigationItem.rightBarButtonItem)
+            XCTAssertEqual(popUp?.webView.bundleVersion, "1.2.3")
```
mludowise-stripe, line 86: nit: if this fails, we're not going to know which URL it failed for in the array. Adding it to the failure message will help us debug faster:
```suggestion
            XCTAssertEqual(popUp?.webView, webView, domain)
            XCTAssertNotNil(popUp?.navigationItem.rightBarButtonItem, domain)
            XCTAssertEqual(popUp?.webView.bundleVersion, "1.2.3", domain)
```

StripeConnect/StripeConnectTests/Internal/Webview/ConnectWebViewTests.swift
```diff
@@ -0,0 +1,167 @@
...
+                        createWebViewWith: webView.configuration,
+                        for: MackNavigationAction(request: .init(url: URL(string: "https://google.com")!)),
+                        windowFeatures: .init())
+                
+        XCTAssertNil(webView)
+        XCTAssertNotNil(safariVC)
+    }
+    
+    func testOpenAsPopUpIfInAllowedHosts() {
+        for domain in [
```
mludowise-stripe, line 64: nit: should this be `url` or `urlString` instead of `domain`?

StripeConnect/StripeConnectTests/Internal/Webview/ScriptWebTestBase.swift
```diff
@@ -0,0 +1,32 @@
...
+    
+    override func setUp() {
+        super.setUp()
+        webView = WKWebView(frame: .zero, configuration: .init())
+    }
+    
+    override func tearDown() {
+        webView = nil
+        super.tearDown()
+    }
```
mludowise-stripe, line 24: I'm curious, why do we need to set this to nil in tearDown? Shouldn't the garbage collector clean it up?

No need to make any changes, just interested for my own benefit.

StripeConnect/StripeConnectTests/StripeConnectTests.swift
```diff
@@ -1,11 +0,0 @@
...
-//  StripeConnectTests.swift
-//  StripeConnectTests
-//
-//  Created by Chris Mays on 7/31/24.
-//
-
-import XCTest
-@testable import StripeConnect
-
-final class StripeConnectTests: XCTestCase { }
```
mludowise-stripe, line 11: ➕

mludowise-stripe (commented):
This looks really great, the tests are extremely thorough!

I mostly have only nits and optional suggestions. I did find 3 blocking issues, but I'm going to pull those out into a followup pass so they don't get hidden in GitHub's UI.

StripeConnect/StripeConnect/Source/Internal/Webview/ConnectWebView.swift
```diff
@@ -0,0 +1,187 @@
...
+    let urlOpener: URLOpener
+    
+    /// The current version for the SDK
+    let bundleVersion: String?
+    
+    init(frame: CGRect,
+         configuration: WKWebViewConfiguration,
+         // Only override for tests
+         urlOpener: URLOpener = UIApplication.shared,
+         bundleVersion: String? = StripeConnectBundleLocator.bundleVersion) {
```
mludowise-stripe, line 42: 🚫 Blocker

We have a helper for this: `StripeAPIConfiguration.STPSDKVersion`. 

We use a hardcoded string instead of the `CFBundleVersion` because there are some cases where the CFBundleVersion doesn't map to the SDK package version, although I don't recall exactly what they were (building from source or maybe certain configurations of Cocoapods / Carthage?).

You'll need to add `@_spi(STP) import StripeCore` to access this:

```suggestion
         sdkVersion: String? = StripeAPIConfiguration.STPSDKVersion) {
```

StripeConnect/StripeConnect/Source/Internal/Webview/MessageHandlers/OnLoadErrorMessageHandler.swift
```diff
@@ -10,9 +10,9 @@ import Foundation
 // Emitted when there is an error loading connect js
 class OnLoadErrorMessageHandler: OnSetterFunctionCalledMessageHandler<OnLoadErrorMessageHandler.Values> {
     struct Values: Codable, Equatable {
-        let error: ErrorValue
+        let error: ConnectLoadingError
         
-        struct ErrorValue: Codable, Equatable {
+        struct ConnectLoadingError: Codable, Equatable, Error {
```
mludowise-stripe, line 15: 🚫 Blocker

This was renamed to `EmbeddedComponentError` in API review and should be moved into its own file + made public. Alternative is remove `Error` conformance from this PR and address decoding the JS error into a `EmbeddedComponentError` error in a separate PR.

I'm not certain if we want to necessarily make `EmbeddedComponentError` conform to `Codable` because I think it would mean that it has to have public conformance and would affect the public API. So we might want to keep this as an internal type that `EmbeddedComponentError` is constructed from.

StripeConnect/StripeConnect/Source/StripeConnectInstance.swift
```diff
@@ -0,0 +1,29 @@
...
+//
+//  Created by Mel Ludowise on 4/30/24.
+//
+
+import JavaScriptCore
+import StripeCore
+import UIKit
+
+@_spi(PrivateBetaConnect)
+public class StripeConnectInstance {
```
mludowise-stripe, line 13: 🚫 Blocker
This got renamed to `EmbeddedComponentManager` during API review

StripeConnect/StripeConnect/Source/Helpers/StripeConnectBundleLocator.swift
```diff
@@ -7,9 +7,15 @@
...
     static let internalClass: AnyClass = StripeConnectBundleLocator.self
     static let bundleName = "StripeConnectBundle"
     static let resourcesBundle = StripeConnectBundleLocator.computeResourcesBundle()
 }
+
+extension BundleLocatorProtocol {
+    static var bundleVersion: String? {
+        resourcesBundle.infoDictionary?["CFBundleVersion"] as? String
+    }
+}
```
mludowise-stripe, line 21: See my comment in `ConnectWebView` – CFBundleVersion isn't reliable, so we shouldn't add this extension

mludowise-stripe (changes requested):
Just a couple of blocking issues

StripeConnect/StripeConnect/Source/Internal/Webview/ConnectWebView.swift
```diff
@@ -0,0 +1,187 @@
...
+ - Camera access
+ - Popup windows
+ - Opening email links
+ - Downloads  TODO MXMOBILE-2485
+ */
+class ConnectWebView: WKWebView {
+
+    /// Closure to present a popup web view controller.
+    /// This is required for any components that can open a popup, otherwise an assertionFailure will occur.
+    var presentPopup: ((UIViewController) -> Void)?
```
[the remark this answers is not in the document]
nschris-stripe: Ah I can look into this.

StripeConnect/StripeConnectTests/Internal/Webview/ScriptWebTestBase.swift
nschris-stripe: It was mostly for symmetry. ARC will take care of this for us.

StripeConnect/StripeConnect/Source/Internal/Webview/ConnectComponentWebView.swift
```diff
@@ -0,0 +1,119 @@
...
+    private let webLocale: () -> String
+    
+    /// The current notification center instance
+    private let notificationCenter: NotificationCenter
+    
+    init(connectInstance: StripeConnectInstance,
+         componentType: ComponentType,
+         // Should only be overridden for tests
+         notificationCenter: NotificationCenter = NotificationCenter.default,
+         webLocale: @escaping () -> String = { Locale.autoupdatingCurrent.webIdentifier }) {
```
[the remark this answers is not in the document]
nschris-stripe: I did look into this, but as far as I can tell there isn't a good way of Faking locale's in iOS 15. I figured since for now we are just using the string it is probably best just to mock that. If we use other parts of the locale later I can look into a better solution.

nschris-stripe:
Diff: https://github.com/stripe/stripe-ios/compare/053f09c95e966af36dd5859bc4e6c8b22f860267..ea0d9142b631b93b3b22b74dd877c193aefdc495

StripeConnect/StripeConnect/Source/Internal/Webview/ConnectComponentWebView.swift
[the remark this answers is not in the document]
mludowise-stripe: You can construct a locale using `Locale(identifier: "fr_FR")`
Example: https://github.com/stripe/stripe-ios/blob/18010dba/Stripe/StripeiOSTests/NSString+StripeTest.swift#L73

StripeConnect/StripeConnect/Source/Internal/Webview/ConnectWebView.swift
```diff
@@ -0,0 +1,187 @@
...
+
+    /// Closure that executes when the view finishes loading.
+    /// - Note: If any JS needs to be evaluated immediately after instantiation, do that here.
+    var didFinishLoading: ((ConnectWebView) -> Void)?
```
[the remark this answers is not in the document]
mludowise-stripe: 🚫 Blocker
This var is no longer being referenced, did you mean to remove the declaration as well?

StripeConnect/StripeConnect/Source/Helpers/EmbeddedComponentError.swift
```diff
@@ -0,0 +1,38 @@
...
+//  StripeConnect
+//
+//  Created by Chris Mays on 8/22/24.
+//
+
+import Foundation
+
+/// An error that can occur loading a Connect embedded component
+@_spi(PrivateBetaConnect)
+public struct EmbeddedComponentError: Error, CustomDebugStringConvertible {
```
mludowise-stripe, line 12: ➕

StripeConnect/StripeConnect/Source/Internal/Webview/MessageHandlers/OnLoadErrorMessageHandler.swift
```diff
@@ -7,6 +7,13 @@
 
 import Foundation
 
+extension OnLoadErrorMessageHandler.Values.ErrorValue {
+    var connectEmbedError: EmbeddedComponentError {
+        // API Error is a catch all so defer to that if we get an unknown type.
```
mludowise-stripe, line 12: We also plan to log an error if the type is unrecognized
https://docs.google.com/document/d/1nzHGofoclzij4nP5n5qgSpYLia-93heStXvt7vOfrkQ/edit?pli=1#bookmark=id.fd4twivivcoi

```suggestion
        // API Error is a catch all so defer to that if we get an unknown type.
        // TODO(MXMOBILE-2491): Log error analytic if `type` is unrecognized
```

StripeConnect/StripeConnect/Source/Internal/Extensions/WKScriptMessage+extension.swift
```diff
@@ -0,0 +1,19 @@
...
+//  Created by Chris Mays on 8/22/24.
+//
+
+import WebKit
+
+extension WKScriptMessage {
+    func toDecodable<DecodableType: Decodable>() throws -> DecodableType {
+        if let payload = body as? DecodableType {
+            return payload
+        } else {
```
mludowise-stripe, line 14: minor nit: The `else` block isn't necessary since `if` returns. It just removes a level of nesting making it a little easier to read

StripeConnect/StripeConnect/Source/Internal/Extensions/WKScriptMessage+extension.swift
```diff
@@ -0,0 +1,19 @@
...
+//  WKScriptMessage+extension.swift
+//  StripeConnect
+//
+//  Created by Chris Mays on 8/22/24.
+//
+
+import WebKit
+
+extension WKScriptMessage {
+    func toDecodable<DecodableType: Decodable>() throws -> DecodableType {
```
mludowise-stripe, line 11: nice! this DRYs up some code nicely.

mludowise-stripe (changes requested):
Looks great! just a couple minor suggestions. Only blocker is the unused var

nschris-stripe:
Diff: https://github.com/stripe/stripe-ios/compare/08710eadc36b93fa434bea3a3d56e6229950acdf..d93f1393ebd2b6d5f61239df2e9a362477cbaf8d

nschris-stripe:
Diff: https://github.com/stripe/stripe-ios/compare/08710eadc36b93fa434bea3a3d56e6229950acdf..95a5947f549383562a4939b0a809581bf3295f8c

mludowise-stripe (approved):

PR #3966, opened 2024-08-27
MOBILESDK-2374: Add analytics for passthrough mode vs PM mode

joyceqin-stripe:
## Summary

Added metadata param for mc_load_succeeded "link_mode": "passthrough" if link enabled and passthroughEnabled, "payment_method_mode" otherwise.

## Testing

Saw the added analytics log when payment sheet loaded successfully


StripePaymentSheet/StripePaymentSheet/Source/Analytics/PaymentSheetAnalyticsHelper.swift
```diff
@@ -255,9 +255,20 @@ final class PaymentSheetAnalyticsHelper {
             guard let elementsSession else { return nil }
             return PaymentSheet.isLinkEnabled(elementsSession: elementsSession, configuration: configuration)
         }()
+        
         var additionalParams = [:] as [String: Any]
         additionalParams["duration"] = duration
         additionalParams["link_enabled"] = linkEnabled
+        if event == .paymentSheetLoadSucceeded {
```
porter-stripe, line 262: Instead of checking the event type here, can we instead find where we send `paymentSheetLoadSucceeded` and add `link_mode` as one of the `params` we pass in?

StripePaymentSheet/StripePaymentSheet/Source/Analytics/PaymentSheetAnalyticsHelper.swift
```diff
@@ -255,9 +255,20 @@ final class PaymentSheetAnalyticsHelper {
...
+            if linkEnabled == true {
```
porter-stripe, line 263: nit, can we replaced with just `linkEnabled`.

StripePaymentSheet/StripePaymentSheet/Source/Analytics/PaymentSheetAnalyticsHelper.swift
```diff
@@ -255,9 +255,20 @@ final class PaymentSheetAnalyticsHelper {
...
+                let linkPassthrough: Bool? = {
+                    guard let elementsSession else { return nil }
+                    return elementsSession.linkPassthroughModeEnabled
+                }()
+                let linkMode: String = linkPassthrough == true ? "passthrough" : "payment_method_mode"
```
porter-stripe, line 268: Should be able to write as 
`let linkMode: String = elementsSession.linkPassthroughModeEnabled ?? false ? "passthrough" : "payment_method_mode"`

porter-stripe (approved):

PR #3969, opened 2024-08-28
[StripeConnect] Add example app.

nschris-stripe:
## Summary
This adds a basic example app utilizing the StripeConnect SDK.

[MXMOBILE-2480](https://jira.corp.stripe.com/browse/MXMOBILE-2480)

## App Breakdown
- On app open retrieve app_info from the API. If loading takes longer than 1 second a reassurance message appears. On success push to the main components view. On Failure show error message and option to retry or go to app settings.
-  The components list shows the Payouts option, more will be available in the future. Selecting the payouts list will push on a PayoutsViewController.
- Tapping the top nav bar takes you to app settings where you can configure the server url and which account is currently selected.

## API Description
`GET: /app_info` : This is requested on app start and will return the public key as well as the available accounts. The display name is optional, but helpful.
Example Response:
```
{
  "publishable_key": "pk_test_<key>",
  "available_merchants": [
    {
      "display_name": "Name to display in app ",
      "merchant_id": "acct_123456"
    },
  ]
}
```
`POST /account_session` The will be called in fetchSession on the SDK. If the account header is provided then it will get the session for that account otherwise it will get a session from the default account which is currently acct_1N9FIXQ26HdRlxHg.
Example Response:
```
{
   "client_secret": "<secret>"
}
```

## Demo

https://github.com/user-attachments/assets/ccaef189-8161-4670-a719-d9184cd1763e

### DIFF
--- a/Example/StripeConnectExample/StripeConnect Example.xcodeproj/project.pbxproj
+++ b/Example/StripeConnectExample/StripeConnect Example.xcodeproj/project.pbxproj
@@ -1,0 +1,587 @@
+// !$*UTF8*$!
+{
+	archiveVersion = 1;
+	classes = {
+	};
+	objectVersion = 56;
+	objects = {
+
+/* Begin PBXBuildFile section */
+		413D18422C7FA30A0051AA42 /* TextInput.swift in Sources */ = {isa = PBXBuildFile; fileRef = 413D18412C7FA30A0051AA42 /* TextInput.swift */; };
+		413D18442C7FAE280051AA42 /* OptionListRow.swift in Sources */ = {isa = PBXBuildFile; fileRef = 413D18432C7FAE280051AA42 /* OptionListRow.swift */; };
+		413D184A2C7FCE5E0051AA42 /* StripeConnectExample.xctestplan in Resources */ = {isa = PBXBuildFile; fileRef = 413D18492C7FCE5E0051AA42 /* StripeConnectExample.xctestplan */; };
+		416E9E972C76BD4400A0B917 /* AppDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = 416E9E962C76BD4400A0B917 /* AppDelegate.swift */; };
+		416E9E992C76BD4400A0B917 /* SceneDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = 416E9E982C76BD4400A0B917 /* SceneDelegate.swift */; };
+		416E9E9B2C76BD4400A0B917 /* AppStartViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 416E9E9A2C76BD4400A0B917 /* AppStartViewController.swift */; };
+		416E9EA02C76BD4600A0B917 /* Assets.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 416E9E9F2C76BD4600A0B917 /* Assets.xcassets */; };
+		416E9EA32C76BD4600A0B917 /* Base in Resources */ = {isa = PBXBuildFile; fileRef = 416E9EA22C76BD4600A0B917 /* Base */; };
+		416E9EB82C76BD4600A0B917 /* StripeConnectExampleUITests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 416E9EB72C76BD4600A0B917 /* StripeConnectExampleUITests.swift */; };
+		416E9ECB2C76BE0C00A0B917 /* StripeConnect.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 416E9EC72C76BDFE00A0B917 /* StripeConnect.framework */; };
+		416E9ECC2C76BE0C00A0B917 /* StripeConnect.framework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = 416E9EC72C76BDFE00A0B917 /* StripeConnect.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
+		416E9ED62C78BD9900A0B917 /* API.swift in Sources */ = {isa = PBXBuildFile; fileRef = 416E9ED52C78BD9900A0B917 /* API.swift */; };
+		416E9EDA2C78BDC300A0B917 /* AccountSessionResponse.swift in Sources */ = {isa = PBXBuildFile; fileRef = 416E9ED92C78BDC300A0B917 /* AccountSessionResponse.swift */; };
+		416E9EDC2C78BE1300A0B917 /* APIError.swift in Sources */ = {isa = PBXBuildFile; fileRef = 416E9EDB2C78BE1300A0B917 /* APIError.swift */; };
+		416E9EDE2C78C0BA00A0B917 /* AppInfo.swift in Sources */ = {isa = PBXBuildFile; fileRef = 416E9EDD2C78C0BA00A0B917 /* AppInfo.swift */; };
+		416E9EE02C78C10800A0B917 /* MainViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 416E9EDF2C78C10800A0B917 /* MainViewController.swift */; };
+		416E9EE32C7B988100A0B917 /* AppSettings.swift in Sources */ = {isa = PBXBuildFile; fileRef = 416E9EE22C7B988100A0B917 /* AppSettings.swift */; };
+		41E9A1BD2C7CC85100EDE131 /* SwiftUIContainerViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 41E9A1BC2C7CC85100EDE131 /* SwiftUIContainerViewController.swift */; };
+		41E9A1BF2C7CD3C700EDE131 /* UIViewController+NavigationEmbed.swift in Sources */ = {isa = PBXBuildFile; fileRef = 41E9A1BE2C7CD3C700EDE131 /* UIViewController+NavigationEmbed.swift */; };
+		41E9A1C12C7CE30000EDE131 /* AppSettingsView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 41E9A1C02C7CE30000EDE131 /* AppSettingsView.swift */; };
+/* End PBXBuildFile section */
+
+/* Begin PBXContainerItemProxy section */
+		416E9EB42C76BD4600A0B917 /* PBXContainerItemProxy */ = {
+			isa = PBXContainerItemProxy;
+			containerPortal = 416E9E8B2C76BD4400A0B917 /* Project object */;
+			proxyType = 1;
+			remoteGlobalIDString = 416E9E922C76BD4400A0B917;
+			remoteInfo = StripeConnectExample;
+		};
+/* End PBXContainerItemProxy section */
+
+/* Begin PBXCopyFilesBuildPhase section */
+		416E9ECD2C76BE0C00A0B917 /* Embed Frameworks */ = {
+			isa = PBXCopyFilesBuildPhase;
+			buildActionMask = 2147483647;
+			dstPath = "";
+			dstSubfolderSpec = 10;
+			files = (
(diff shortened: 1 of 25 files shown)


Example/StripeConnectExample/StripeConnectExample/API/API.swift
```diff
@@ -0,0 +1,60 @@
...
+                let decoder = JSONDecoder()
+                decoder.keyDecodingStrategy = .convertFromSnakeCase
+                do {
+                    return .failure(.responseError(response: try decoder.decode(APIErrorResponse.self, from: data)))
+                } catch {
+                    return .failure(.failedToParse(error: error))
+                }
+            }
+            let decoder = JSONDecoder()
+            decoder.keyDecodingStrategy = .convertFromSnakeCase
```
mludowise-stripe, line 46: Could we reuse the same decoder instead of instantiating two of them? It looks like they use the same configuration

Example/StripeConnectExample/StripeConnectExample/MainViewController.swift
```diff
@@ -0,0 +1,129 @@
...
+        return button
+    }()
+
+    override var title: String? {
+        didSet {
+            navbarTitleButton.setTitle(title, for: .normal)
+        }
+    }
+
+    lazy var embeddedComponentInstance: EmbeddedComponentManager = {
```
mludowise-stripe, line 63: naming nit:
```suggestion
    lazy var embeddedComponentManager: EmbeddedComponentManager = {
```

Example/StripeConnectExample/StripeConnectExample/Views/OptionListRow.swift
```diff
@@ -0,0 +1,35 @@
...
+            onSelected()
+        } label: {
+            HStack {
+                VStack(alignment: .leading) {
+                    Text(title)
+                        .font(.body)
+                        .foregroundColor(Color(UIColor.label))
+                    subtitle.map(Text.init)
+                        .font(.caption)
+                        .foregroundColor(Color(UIColor.secondaryLabel))
```
mludowise-stripe, line 26: I think this is equivalent and more SwiftUI-y
```suggestion
                        .foregroundColor(.primary)
                    subtitle.map(Text.init)
                        .font(.caption)
                        .foregroundColor(.secondary)
```

Example/StripeConnectExample/StripeConnectExampleTests/StripeConnectExampleTests.swift
```diff
@@ -0,0 +1,15 @@
...
+//  StripeConnectExampleTests.swift
+//  StripeConnectExampleTests
+//
+//  Created by Chris Mays on 8/21/24.
+//
+
+import XCTest
+@testable import StripeConnectExample
+
+final class StripeConnectExampleTests: XCTestCase {
```
mludowise-stripe, line 11: I don't think we need to write any unit tests against the example app, but rather just E2E UI tests. So we can likely delete this file and the corresponding test target.

But we should keep the UITests target.

bitrise.yml
```diff
@@ -661,6 +661,24 @@ workflows:
...
+        machine_type_id: g2-m1-max.5core
+  ui-tests-5:
+    steps:
+    - xcode-test@4:
+        inputs:
+        - destination: $DEFAULT_TEST_DEVICE
+        - test_repetition_mode: retry_on_failure
+        - maximum_test_repetitions: "2"
+        - scheme: StripeConnect Example
+        - test_plan: StripeConnectExample
```
mludowise-stripe, line 676: 🤔 Interesting. It looks like PaymentSheet tests are broken into "shards", presumably so the tests can be run in parallel for faster CI times. Something we might want to consider when we start to write UI tests.

Example/StripeConnectExample/StripeConnectExample/AppSettingsView.swift
```diff
@@ -0,0 +1,90 @@
...
+                                      selected: selectedMerchant?.id == merchant.id,
+                                      onSelected: {
+                            selectedMerchant = merchant
+                        })
+                    }
+                } header: {
+                    Text("Select a demo account")
+                }
+                Section {
+                    TextInput(label: "", placeholder: "https://example.com", text: $serverURLString, isValid: isCustomEndpointValid)
```
mludowise-stripe, line 53: In addition to specifying a custom server, it's actually more important that we be able to specify additional accounts, beyond just the canonical ones hardcoded on the backend. This will enable us to quickly create test accounts and test them agains the example app during team bug bashes.

But I think you can address this in a separate PR – this work was spec'd in a separate ticket (MXMOBILE-2481) anyway and you ended up including most of it here 😄

Example/StripeConnectExample/StripeConnectExample/Base.lproj/LaunchScreen.storyboard
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="13122.16" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" launchScreen="YES" useTraitCollections="YES" useSafeAreas="YES" colorMatched="YES" initialViewController="01J-lp-oVM">
+    <dependencies>
+        <plugIn identifier="com.apple.InterfaceBuilder.IBCocoaTouchPlugin" version="13104.12"/>
+        <capability name="Safe area layout guides" minToolsVersion="9.0"/>
+        <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
+    </dependencies>
+    <scenes>
+        <!--View Controller-->
```
mludowise-stripe, line 9: nit: Can we add something to the launch screen so we see something on screen to know when it's loading vs. we should be seeing UI. It could be a label that just says "StripeConnect Example" or something.

Example/StripeConnectExample/StripeConnect Example.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,704 @@
+// !$*UTF8*$!
```
mludowise-stripe, line 1: This can be done as a followup PR, but we also need to enable all the same locales as the SDK in the example app in order to test them.

Example/StripeConnectExample/StripeConnectExample/MainViewController.swift
```diff
@@ -0,0 +1,129 @@
...
+//  StripeConnect Example
+//
+//  Created by Mel Ludowise on 4/30/24.
+//
+
+@_spi(PrivateBetaConnect) import StripeConnect
+import SwiftUI
+import UIKit
+
+class MainViewController: UITableViewController {
```
mludowise-stripe, line 12: Was there a reason you opted to keep this in UIKit instead of rewriting in SwiftUI? There's nothing wrong with keeping it as-is, but if it was in SwiftUI it might save us from needing `SwiftUIContainerViewController`

Example/StripeConnectExample/StripeConnectExample/AppStartViewController.swift
```diff
@@ -0,0 +1,95 @@
...
+    }
+    
+    var body: some View {
+        VStack(spacing: 10) {
+            switch loadingState {
+            case .loading:
+                ProgressView()
+                Text("Warming up the server...This might take several seconds")
+                    .opacity(displayLoadTimeReassurance ? 1.0 : 0.0)
+                    .multilineTextAlignment(.center)
```
mludowise-stripe, line 32: Nice touch!

mludowise-stripe (approved):
This looks really great! Way better than what I had 😛 

I only had minor optional nits, which you can decide whether or not to take.

There were a couple items that should be addressed, but it probably makes sense to do as a followup PR since this one is already quite large.
- Locale support
- Specific accounts

Example/StripeConnectExample/StripeConnectExample/AppStartViewController.swift
mludowise-stripe: supernit:
- Use ellipsis instead of three dots
- Adds space

```suggestion
                Text("Warming up the server… This might take several seconds")
                    .opacity(displayLoadTimeReassurance ? 1.0 : 0.0)
                    .multilineTextAlignment(.center)
```

Example/StripeConnectExample/StripeConnectExample.xctestplan
```diff
@@ -0,0 +1,29 @@
+{
```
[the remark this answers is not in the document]
nschris-stripe: I added this because we might want tests in the future (unsure) but it has the added benefit of allowing us to add it as a test run in Bit rise to ensure the app builds.

Example/StripeConnectExample/StripeConnectExample/MainViewController.swift
nschris-stripe: I figured a hybrid SwiftUI/UIKit app is what most people are operating on today. Also the SDK does not offer View equivalents in the API. I thought a UIKit version of this was the best way to show how to use the SDK without all of the added complexity of wrapping the view controllers in views.

Example/StripeConnectExample/StripeConnectExample/API/API.swift
```diff
@@ -0,0 +1,60 @@
...
+        await apiRequest(path: "account_session", 
+                         method: "POST",
+                         headers: merchantId.map { ["account": $0] } ?? [:],
+                         baseURL: baseURL)
+    }
+    
+    static func apiRequest<Response: Codable>(path: String,
+                                              method: String = "GET",
+                                              headers: [String: String] = [:],
+                                              baseURL: String = AppSettings.shared.selectedServerBaseURL) async -> Result<Response, APIError> {
```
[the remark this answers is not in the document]
nschris-stripe: The other thing that I like to avoid here is wrapping everything in a do/catch. I feel like that can sometimes clutter the code. What do you think?

nschris-stripe:
Diff: https://github.com/stripe/stripe-ios/compare/1b7f92fdb01318a38a8c68cccadaca01e4f40557..1b7eb824c0c2af7746650c093d6e9e3bf4b0f18c

nschris-stripe:
Updated based on feedback: https://github.com/stripe/stripe-ios/compare/1b7f92fdb01318a38a8c68cccadaca01e4f40557..1b7eb824c0c2af7746650c093d6e9e3bf4b0f18c

nschris-stripe:
Met with Lauren and we agreed upon a generic design. She is good with this, with the understanding that eventually we will support line height in our sail implementation:
Simulator Screenshot - iPhone 15 Pro Max - 2024-08-29 at 14 54 24


Here is the full diff since your last review:
https://github.com/stripe-ios/dashboard-ios/compare/4188b52b0436f9bcfc780817ce6761382e5f5fc6..5e40a36e1168a5a2e11f6373d53de5b3d3fca026

Example/StripeConnectExample/StripeConnectExample.xctestplan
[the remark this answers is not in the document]
mludowise-stripe: Good call – I think it is important to at least ensure the app builds in Bitrise

Example/StripeConnectExample/StripeConnectExample/API/API.swift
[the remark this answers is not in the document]
mludowise-stripe: >  Luckily we will get typed throws when we move to Swift 6

🥳

> The other thing that I like to avoid here is wrapping everything in a do/catch

I'm not strongly opinionated, so I think it's reasonable to keep this is a result. 

My personal opinion is that results force you to use switch statements, which isn't any cleaner than a try/catch. But throwing errors can potentially mean less logic handling because you can often defer it up the stack and consolidate the error handling in one place. But it's ultimately a personal preference 😄

mludowise-stripe (approved):

Issue #3358, opened 2024-03-04
[BUG - PaymentSheet UI] When selecting and dismissing iDEAL/PayPal webview authentication, after selecting card payment the authentication webview is being presented again 


MarcelStarczyk:
## Summary
I am using the prebuilt PaymentSheet UI. Reproducable in test mode 👍 

Steps to reproduce:
1. Open the Payment Sheet (using "setup" intent configuration mode)
2. Select Paypal or any other payment method that presents webview authentication
3. Dismiss the webview using upper-left close button on the webview (without choosing "Authorize Test Setup" or "Fail Test Setup")
4. Without dismissing the payment sheet, select Card payment method and fill out the required data (card number, date and CVC number) 
5. Press "Setup" button 

Result: Authentication web view is presented from the previously selected Paypal payment method even though user has "Card" selected as his payment method

## iOS version
17.2

## Installation method
SPM

## SDK version
23.23.0

yuki-stripe:
Hello @MarcelStarczyk,

This can happen if your backend returns the same SetupIntent when [`handleConfirm`](https://github.com/stripe/stripe-ios/blob/65670bdd425263fc57dadeac947ddada11bb989f/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetIntentConfiguration.swift#L48) is called. Your implementation should either
1. Create a new SetupIntent each time `handleConfirm` is called
2. Update the SetupIntent with the new payment method ID

Let me know if that solves your issue!

yuki-stripe:
Glad to hear it. I've added an action item on our side to add a guardrail, sorry you ran into this!  Closing this out.

PR #3972, opened 2024-08-29
MOBILESDK-1817 [iOS] Add validation to deferred flow to check Intent payment method id matches

joyceqin-stripe:
## Summary

Added validation functions to PaymentSheetDeferredValidator to check Intent payment method id matches the id of the STPPaymentMethod passed into the ConfirmHandler. Called this validator in 4a Client-side confirmation and 4b Server-side confirmation in PaymentSheet+DeferredAPI for both SetupIntent and PaymentIntent. Added simple tests for both.

## Motivation

Based on this issue https://github.com/stripe/stripe-ios/issues/3358, there's a bug merchants can make in their integration that won't get caught today.

## Testing

Reproduced the issue (Attempt Amazon Pay, cancel, and switch to card) and made unit tests for the validator method with the following checks:
1. Ids match (does not throw)
2. Ids do not match (throws)
3. Payment method on intent is nil (does not throw)


porter-stripe (approved):
nice!

PR #3974, opened 2024-08-30
Financial Connections: small fix for a networking manual entry edge case

kgaidis-stripe:
## Summary

^ see the comment in the code

## Testing

Video testing the edge case:

https://github.com/user-attachments/assets/82daccfe-5912-48b0-8c4d-30767bea9b31

We can see that save to link succeeds:

Screenshot 2024-08-30 at 2 01 39 PM


mats-stripe (approved):

kgaidis-stripe:
we went with a different solution

PR #3983, opened 2024-09-04
Start on embedded

porter-stripe:
## Summary
- Adds a new `UIView` `EmbeddedPaymentMethodsView` which is a raw view for embedded.
- Supports three styles, `flatRadio`, `floating`.
- Stubs out all new additions to the Appearance API for embedded
- Adds a skeleton playground for embedded
- Snapshot tests

#### Flat radio

https://github.com/user-attachments/assets/4e100fe7-474d-4ac0-b7b6-f65f47a53aea

#### Floating

https://github.com/user-attachments/assets/4536cc72-5a33-4404-8de8-2b4968cf9b34

#### Flat radio with saved payment method

https://github.com/user-attachments/assets/18766528-666d-443c-864f-8854e1b706fd

## Motivation
https://docs.google.com/document/d/1UbpF1Jt1Uv2NcB5WoGOn26lHpESd8JrH4y8pbAIJ2XI/edit?pli=1

## Testing
- Manual
- Snapshot

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButton.swift
```diff
@@ -107,12 +139,12 @@ class RowButton: UIView {
             labelsStackView.leadingAnchor.constraint(equalTo: imageView.trailingAnchor, constant: 12),
             labelsStackView.trailingAnchor.constraint(equalTo: rightAccessoryView?.leadingAnchor ?? trailingAnchor, constant: -12),
             labelsStackView.centerYAnchor.constraint(equalTo: centerYAnchor),
-            labelsStackView.topAnchor.constraint(greaterThanOrEqualTo: topAnchor, constant: 4),
-            labelsStackView.bottomAnchor.constraint(lessThanOrEqualTo: bottomAnchor, constant: -4),
+            labelsStackView.topAnchor.constraint(greaterThanOrEqualTo: topAnchor, constant: appearance.paymentOptionView.paymentMethodRow.additionalInsets),
+            labelsStackView.bottomAnchor.constraint(lessThanOrEqualTo: bottomAnchor, constant: -appearance.paymentOptionView.paymentMethodRow.additionalInsets),
```
porter-stripe, line 143: I could see some fud that now vertical mode gets some of the defaults in `paymentOptionView`, and if those change it'll impact vertical. I think we should make `paymentOptionView` on appearance optional, then, have other default values in RowButton. Thoughts?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetVerticalViewController.swift
```diff
@@ -363,6 +363,9 @@ class PaymentSheetVerticalViewController: UIViewController, FlowControllerViewCo
             allowsRemovalOfLastSavedPaymentMethod: configuration.allowsRemovalOfLastSavedPaymentMethod,
             allowsPaymentMethodRemoval: loadResult.elementsSession.allowsRemovalOfPaymentMethodsForPaymentSheet()
         )
+        // TODO(porter) Find a better way
+        var updatedAppearance = configuration.appearance
+        updatedAppearance.paymentOptionView.style = .floating
```
porter-stripe, line 368: This is how we work around the default value in appearance being flatRadio, this now means we need to update tests which test lower level components to update this default. :/

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentMethodsViewSnapshotTests.swift
```diff
@@ -0,0 +1,314 @@
...
+        let embeddedView = EmbeddedPaymentMethodsView(paymentMethodTypes: [.stripe(.card), .stripe(.cashApp)],
+                                                      savedPaymentMethod: ._testCard(),
+                                                      appearance: ._testMSPaintTheme,
+                                                      shouldShowApplePay: true,
+                                                      shouldShowLink: true,
+                                                      savedPaymentMethodAccessoryType: .viewMore)
+
+        // Simulate tapping a button
+        if let rowButton = embeddedView.stackView.arrangedSubviews.first(where: { $0 is RowButton }) as? RowButton {
+            embeddedView.handleRowSelection(selectedRowButton: rowButton)
```
yuki-stripe, line 176: Does this do anything? I think this taps the saved card or card row

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentMethodsViewSnapshotTests.swift
```diff
@@ -0,0 +1,314 @@
...
+        }
+
+        verify(embeddedView)
+    }
+
```
yuki-stripe, line 181: Can we add tests for font? Like giant font size, tiny font size, custom font.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentMethodsViewSnapshotTests.swift
```diff
@@ -0,0 +1,314 @@
...
+        appearance.paymentOptionView.paymentMethodRow.additionalInsets = 20
+
+        let embeddedView = EmbeddedPaymentMethodsView(paymentMethodTypes: [.stripe(.card), .stripe(.cashApp)],
+                                                      savedPaymentMethod: nil,
+                                                      appearance: appearance,
+                                                      shouldShowApplePay: true,
+                                                      shouldShowLink: true,
+                                                      savedPaymentMethodAccessoryType: .none)
+
+        verify(embeddedView)
```
yuki-stripe, line 252: same comment here about explicit height check

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentMethodsViewSnapshotTests.swift
```diff
@@ -0,0 +1,314 @@
...
+                                                      savedPaymentMethod: nil,
+                                                      appearance: .default,
+                                                      shouldShowApplePay: true,
+                                                      shouldShowLink: false,
+                                                      savedPaymentMethodAccessoryType: .none)
+
+        verify(embeddedView)
+    }
+
+    func testEmbeddedPaymentMethodsView_flatRadio_rowHeight() {
```
yuki-stripe, line 63: Could we test row height without any "two PM" rows as well?

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentMethodsViewSnapshotTests.swift
```diff
@@ -0,0 +1,314 @@
...
+import StripeCoreTestUtils
+@_spi(STP) @testable import StripePayments
+@_spi(EmbeddedPaymentMethodsViewBeta) @testable import StripePaymentSheet
+@testable import StripePaymentsTestUtils
+@_spi(STP) @testable import StripeUICore
+import XCTest
+
+class EmbeddedPaymentMethodsViewSnapshotTests: STPSnapshotTestCase {
+
+    // MARK: Flat radio snapshot tests
```
yuki-stripe, line 17: Add test for componentBackgroundColor?

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentMethodsViewSnapshotTests.swift
```diff
@@ -0,0 +1,314 @@
...
+                                                      savedPaymentMethodAccessoryType: .none)
+
+        // Simulate tapping a button
+        if let rowButton = embeddedView.stackView.arrangedSubviews.first(where: { $0 is RowButton }) as? RowButton {
+            embeddedView.handleRowSelection(selectedRowButton: rowButton)
+        }
+
+        verify(embeddedView)
+    }
+
```
yuki-stripe, line 290: Add test for border radius & componentBackground?

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentMethodsViewSnapshotTests.swift
```diff
@@ -0,0 +1,314 @@
...
+                                                      savedPaymentMethod: nil,
+                                                      appearance: appearance,
+                                                      shouldShowApplePay: true,
+                                                      shouldShowLink: true,
+                                                      savedPaymentMethodAccessoryType: .none)
+
+        verify(embeddedView)
+    }
+
+    func testEmbeddedPaymentMethodsView_flatRadio_separatorInset() {
```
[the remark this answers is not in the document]
porter-stripe: Looks like the translucent indicator/checkerboard pattern showing through for the .pngs. We could update the view to have a solid background color.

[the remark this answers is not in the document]
yuki-stripe: Ah I see, yeah. Is that right?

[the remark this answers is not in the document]
porter-stripe: Oh good question... maybe it should fill w/ the background color in that space. I think that probably sounds more right.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentMethodsViewSnapshotTests.swift
porter-stripe: It does select the saved PM, nothing is selected by default and is what I believe is expected based on the figmas.
testEmbeddedPaymentMethodsView_flatRadio_genericTestAppearance@3x

yuki-stripe: Ah ok. What's selected by default should (eventually) match vertical mode logic. This code might break when the view does what it really should do when you tap the saved PM (open the sheet)?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/RadioButton.swift
```diff
@@ -0,0 +1,127 @@
...
+    }()
+
+    /// Layer for the "on" state.
+    private lazy var onLayer: CALayer = {
+        let layer = CALayer()
+        layer.bounds = CGRect(x: 0, y: 0, width: Constants.diameter, height: Constants.diameter)
+        layer.cornerRadius = Constants.diameter / 2
+        layer.borderWidth = Constants.borderWidth
+        layer.anchorPoint = CGPoint(x: 0.5, y: 0.5)
+        layer.borderColor = selectedColor
```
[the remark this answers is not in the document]
porter-stripe: Beautiful!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/UIStackView+Separator.swift
```diff
@@ -0,0 +1,48 @@
...
+            addSeparator(color: color, thickness: thickness, inset: inset, at: 0)
+        }
+
+        if addBottomSeparator {
+            addSeparator(color: color, thickness: thickness, inset: inset, at: arrangedSubviews.count)
+        }
+    }
+
+    private func addSeparator(color: UIColor, thickness: CGFloat, inset: UIEdgeInsets, at index: Int) {
+        let containerView = UIView()
```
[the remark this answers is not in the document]
porter-stripe: Good suggestion, I've never used them so TIL. I actually ended up using the background color of this containerView to fix an issue pointed out in the snapshots of the translucent separators. https://github.com/stripe/stripe-ios/pull/3983#discussion_r1747628600

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetAppearance.swift
```diff
@@ -36,10 +36,16 @@ public extension PaymentSheet {
         /// - Note: The behavior of this property is consistent with the behavior of border width on `CALayer`
         public var borderWidth: CGFloat = 1.0
 
+        // TOOD(porter) Make public for GA
+        @_spi(EmbeddedPaymentMethodsViewBeta) public var borderWidthSelected: CGFloat?
```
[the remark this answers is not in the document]
porter-stripe: Added, let's flesh them our in API review.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButton.swift
```diff
@@ -14,6 +14,13 @@ import UIKit
 /// A selectable button used in vertical mode to display payment methods.
 class RowButton: UIView {
     private let shadowRoundedRect: ShadowedRoundedRectangle
+    private lazy var radioButton: RadioButton? = {
+        guard appearance.paymentOptionView.style == .flatRadio else { return nil }
+        return RadioButton(appearance: appearance) { [weak self] in
+            guard let self = self else { return }
```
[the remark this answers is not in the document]
porter-stripe: Wonder if we can add a lint rule for this.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentMethodsViewSnapshotTests.swift
porter-stripe: I think the initial selection point is a good one, but tapping the row shouldn't open the sheet IMO, just tapping the "View more" button. I'll update this to select the last row.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentMethodsViewSnapshotTests.swift
```diff
@@ -0,0 +1,314 @@
...
+
+        // Simulate tapping a button
+        if let rowButton = embeddedView.stackView.arrangedSubviews.first(where: { $0 is RowButton }) as? RowButton {
+            embeddedView.handleRowSelection(selectedRowButton: rowButton)
+        }
+
+        verify(embeddedView)
+    }
+
+    func testEmbeddedPaymentMethodsView_flatRadio_genericTestAppearance() {
```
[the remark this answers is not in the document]
porter-stripe: Hm, yeah fair. I removed them, I think we have good coverage with the others.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetVerticalViewController.swift
porter-stripe: I like 5, but upon trying to implement it I found I was just redeclaring much of Appearance and also needed helper instance font scaling functions. For simplicity, I decided to pass an `isEmbedded` flag into `RowButton` for now. It's easy to understand and implement. Happy to talk more about this. I feel like this may be a symptom of Appearance growing too big, or maybe Appearance shouldn't have a `PaymentOptionView` and instead `PaymentOptionView` should be it's own thing and contain an `Appearance` object.

yuki-stripe: > Setting appearance.paymentOptionView.whatever should not have an effect outside of the PaymentOptionView. 
I think the current implementation still has this problem.

> I feel like this may be a symptom of Appearance growing too large, or maybe Appearance shouldn’t define the style or have a PaymentOptionView
I see this problem as coming from the fact that we are trying to reuse PaymentMethodRow for both the embedded list (where we should respect the paymentOptionView appearance API) and not-embedded-list (including the saved PM screen which is notably still used in the embedded integration).

> I think this may indicate that we should reconsider the shape of the public API. IMO, it’s incorrect to have a populated paymentOptionView on Appearance when it’s not being used in embedded. Otherwise, we would need logic to determine when to ignore/modify the API values, like the options we've presented above.

I'm definitely open to making the PaymentOptionView appearance separate from PaymentSheet.Appearance, but it shouldn't be because of our internal implementation.

yuki-stripe (approved):

PR #3984, opened 2024-09-05
Change component name from payment_sheet to mobile_payment_element

wooj-stripe:
## Summary
Account for rename of CustomerSession `payment_sheet` to `mobile_payment_element`

## Motivation
Rename

## Testing
Relying on existing tests

### DIFF
--- a/Stripe/StripeiOSTests/STPPaymentMethodFunctionalTest.swift
+++ b/Stripe/StripeiOSTests/STPPaymentMethodFunctionalTest.swift
@@ -139,8 +139,7 @@ class STPPaymentMethodFunctionalTest: STPNetworkStubbingTestCase {
                                    ephemeralKeySecret: customerAndEphemeralKey.ephemeralKeySecret)
 
         // Element/Sessions endpoint should de-dupe payment methods with CustomerSesssion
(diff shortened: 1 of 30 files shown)


Stripe/StripeiOSTests/STPPaymentMethodFunctionalTest.swift
```diff
@@ -139,8 +139,9 @@ class STPPaymentMethodFunctionalTest: STPNetworkStubbingTestCase {
                                    ephemeralKeySecret: customerAndEphemeralKey.ephemeralKeySecret)
 
         // Element/Sessions endpoint should de-dupe payment methods with CustomerSesssion
-        let cscs = try await STPTestingAPIClient.shared().fetchCustomerAndCustomerSessionClientSecret(customerID: customerAndEphemeralKey.customer,
-                                                                                               merchantCountry: nil)
+        let cscs = try await STPTestingAPIClient.shared().fetchCustomerAndCustomerSessionClientSecret(componentName: "mobile_payment_element",
```
yuki-stripe, line 142: Do we need to pass in `componentName`? Can we hardcode it in the `fetch...` method instead? I don't see any other users.

yuki-stripe (approved):

PR #3985, opened 2024-09-05
[StripeConnect] Add appearance.

nschris-stripe:
## Summary
Adds the ability to set and update appearance for embedded components. Due to the fact that appearance cannot be made codable I also had to refactor some of the message receiving and sending code to make appearance compatible. 

## Features
- ComponentWebView uses the ComponentManager's appearance, and updates when the appearance or trait collection is updated. Colors are derived based on the webview's traitcollection.
- Example app has new theming functionality.

https://jira.corp.stripe.com/browse/MXMOBILE-2486

### DIFF
--- a/Example/StripeConnectExample/StripeConnectExample/Appearance/AppSettings+AppearanceInfo.swift
+++ b/Example/StripeConnectExample/StripeConnectExample/Appearance/AppSettings+AppearanceInfo.swift
@@ -1,0 +1,112 @@
+//
+//  AppSettings+AppearanceInfo.swift
+//  StripeConnect Example
+//
+//  Created by Chris Mays on 9/5/24.
+//
+
+import Foundation
+@_spi(PrivateBetaConnect) import StripeConnect
+import UIKit
+
+extension AppSettings {
+    var appearanceOptions: [AppearanceInfo] {
+        [
+            .default,
+            .hotDog,
+            .link,
+            .oceanBreeze,
+            .shrek
+        ]
+    }
+    
+    var appearanceInfo: AppearanceInfo {
+        get {
+            appearanceOptions.first(where: {
+                $0.id == AppSettings.shared.appearanceId
+            }) ?? .default
+        }
+        set {
+            appearanceId = newValue.id
+        }
+    }
+}
+
+
+extension AppearanceInfo {
+    static var `default`: AppearanceInfo {
+        .init(displayName: "Default", appearance: .default)
+    }
+    
+    static var shrek: AppearanceInfo {
+        var appearance = EmbeddedComponentManager.Appearance()
+        appearance.colors.primary = UIColor(red: 90/255, green: 233/255, blue: 43/255, alpha: 1)
+        appearance.colors.background = UIColor(red: 131/255, green: 116/255, blue: 17/255, alpha: 1)
+        appearance.buttonPrimary.colorBackground = UIColor(red: 218/255, green: 185/255, blue: 185/255, alpha: 1)
+        appearance.buttonPrimary.colorBorder = UIColor(red: 240/255, green: 0/255, blue: 0/255, alpha: 1)
+        appearance.buttonPrimary.colorText = UIColor(red: 0/255, green: 0/255, blue: 0/255, alpha: 1)
+        appearance.buttonSecondary.colorBackground = UIColor(red: 2/255, green: 95/255, blue: 8/255, alpha: 1)
+        appearance.colors.text = UIColor(red: 85/255, green: 65/255, blue: 37/255, alpha: 1)
+        appearance.badgeNeutral.colorText = UIColor(red: 40/255, green: 215/255, blue: 42/255, alpha: 1)
+        appearance.badgeNeutral.colorBackground = UIColor(red: 99/255, green: 136/255, blue: 99/255, alpha: 1)
+        appearance.buttonSecondary.colorText = UIColor(red: 0/255, green: 0/255, blue: 0/255, alpha: 1)
+        return .init(displayName: "Shrek", appearance: appearance)
+    }
+    
+    static var hotDog: AppearanceInfo {
+        var appearance = EmbeddedComponentManager.Appearance()
+
+        appearance.colors.primary = UIColor(red: 255/255, green: 34/255, blue: 0/255, alpha: 1)
+        appearance.colors.background = UIColor(red: 255/255, green: 255/255, blue: 0/255, alpha: 1)
+        appearance.buttonSecondary.colorBackground = UIColor(red: 198/255, green: 198/255, blue: 198/255, alpha: 1)
+        appearance.buttonSecondary.colorBorder = UIColor(red: 31/255, green: 31/255, blue: 31/255, alpha: 1)
+        appearance.buttonPrimary.colorBorder = UIColor(red: 31/255, green: 31/255, blue: 31/255, alpha: 1)
+        appearance.buttonPrimary.colorText = UIColor(red: 31/255, green: 31/255, blue: 31/255, alpha: 1)
+        appearance.buttonPrimary.colorBackground = UIColor(red: 198/255, green: 198/255, blue: 198/255, alpha: 1)
+        appearance.colors.offsetBackground = UIColor(red: 255/255, green: 34/255, blue: 0/255, alpha: 1)
+        appearance.colors.text = UIColor(red: 0/255, green: 0/255, blue: 0/255, alpha: 1)
+        appearance.colors.secondaryText = UIColor(red: 0/255, green: 0/255, blue: 0/255, alpha: 1)
+        appearance.badgeDanger.colorText = UIColor(red: 153/255, green: 20/255, blue: 0/255, alpha: 1)
+        appearance.badgeWarning.colorBackground = UIColor(red: 249/255, green: 164/255, blue: 67/255, alpha: 1)
+
+        appearance.cornerRadius.base = 0
+        
+        return .init(displayName: "Hot Dog Stand", appearance: appearance)
+    }
+    
+    static var oceanBreeze: AppearanceInfo {
+        var appearance = EmbeddedComponentManager.Appearance()
+        appearance.colors.background = UIColor(red: 234/255, green: 246/255, blue: 251/255, alpha: 1)  // #EAF6FB
+        appearance.colors.primary = UIColor(red: 21/255, green: 96/255, blue: 158/255, alpha: 1)  // #15609E
+        appearance.badgeSuccess.colorText = UIColor(red: 42/255, green: 96/255, blue: 147/255, alpha: 1)  // #2A6093
+        appearance.badgeNeutral.colorText = UIColor(red: 90/255, green: 98/255, blue: 29/255, alpha: 1)  // #5A621D
+        appearance.buttonSecondary.colorText = UIColor(red: 44/255, green: 147/255, blue: 232/255, alpha: 1)  // #2C93E8
+        appearance.buttonSecondary.colorBorder = UIColor(red: 44/255, green: 147/255, blue: 232/255, alpha: 1)  // #2C93E8
+        appearance.cornerRadius.base = 23
+
+        return .init(displayName: "Ocean Breeze", appearance: appearance)
+    }
+    
+    static var link: AppearanceInfo {
+        var appearance = EmbeddedComponentManager.Appearance()
+
+        appearance.colors.primary = UIColor(red: 28/255, green: 57/255, blue: 68/255, alpha: 1)
+        appearance.buttonPrimary.colorBackground = UIColor(red: 51/255, green: 220/255, blue: 179/255, alpha: 1)
+        appearance.buttonPrimary.colorBorder = UIColor(red: 51/255, green: 220/255, blue: 179/255, alpha: 1)
+        appearance.buttonPrimary.colorText = UIColor(red: 28/255, green: 57/255, blue: 68/255, alpha: 1)
+        appearance.colors.secondaryText = UIColor(red: 72/255, green: 91/255, blue: 97/255, alpha: 1)
+        appearance.colors.text = UIColor(red: 28/255, green: 57/255, blue: 68/255, alpha: 1)
+        appearance.colors.actionPrimaryText = UIColor(red: 51/255, green: 220/255, blue: 179/255, alpha: 1)
+        appearance.badgeSuccess.colorBackground = UIColor(red: 180/255, green: 254/255, blue: 225/255, alpha: 1)
+        appearance.badgeSuccess.colorBorder = UIColor(red: 192/255, green: 215/255, blue: 205/255, alpha: 1)
+        appearance.badgeSuccess.colorText = UIColor(red: 28/255, green: 57/255, blue: 68/255, alpha: 1)
+        appearance.badgeNeutral.colorBackground = UIColor(red: 222/255, green: 254/255, blue: 204/255, alpha: 1)
+        appearance.badgeNeutral.colorText = UIColor(red: 28/255, green: 57/255, blue: 68/255, alpha: 1)
+
+        appearance.cornerRadius.base = 5
+
+        appearance.spacingUnit = 9
+        
+        return .init(displayName: "Link", appearance: appearance)
+    }
+}
--- a/StripeConnect/StripeConnect/Source/EmbeddedComponentManager+Appearance.swift
+++ b/StripeConnect/StripeConnect/Source/EmbeddedComponentManager+Appearance.swift
@@ -1,0 +1,237 @@
+//
+//  EmbeddedComponentManager+Appearance.swift
+//  StripeConnect
+//
+//  Created by Chris Mays on 8/28/24.
+//
+
+@_spi(STP) import StripeCore
+import UIKit
+
+@_spi(PrivateBetaConnect)
+extension EmbeddedComponentManager {
+    /// Describes the appearance of embedded components.
+    /// - seealso: https://docs.stripe.com/connect/embedded-appearance-options
+    public struct Appearance {
+        public enum TextTransform {
+            /// The text doesn’t have a transform.
+            case none
+            /// Displays the text in all uppercase characters.
+            case uppercase
+            /// Displays the text in all lowercase characters.
+            case lowercase
+            /// Displays the text with the first character capitalized.
+            case capitalize
+            
+            // Since the public API does not call for TextTransform to be a string
+            // we manually create a raw value here.
+            var rawValue: String {
+                switch self {
+                case .none:
+                    return "none"
+                case .uppercase:
+                    return "uppercase"
+                case .lowercase:
+                    return "lowercase"
+                case .capitalize:
+                    return "capitalize"
+                }
+            }
+            
+            init?(rawValue: String) {
+                switch rawValue {
+                case "none":
+                    self = .none
+                case "uppercase":
+                    self = .uppercase
+                case "lowercase":
+                    self = .lowercase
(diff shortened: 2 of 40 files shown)


StripeConnect/StripeConnect/Source/Internal/Webview/MessageSenders/UpdateConnectInstanceSender.swift
```diff
@@ -9,9 +9,9 @@ import Foundation
 
 /// Updates appearance and locale options.
 struct UpdateConnectInstanceSender: MessageSender {
-    struct Payload: Codable, Equatable {
+    struct Payload: Encodable, Equatable {
         let locale: String
-        // TODO: Add appearance here.
+        private(set) var appearance: Appearance = .default
```
mludowise-stripe, line 14: Instead of defaulting appearance to `.default`, we could default it to `nil` since appearance and locale are both optional fields in the JS object passed to web.

I think this is better because it will prevent us from unintentionally resetting the appearance to the default when we actually intend to not set it.
```suggestion
        private(set) var locale: String?
        private(set) var appearance: Appearance?
```

StripeConnect/StripeConnectTests/AppearanceTests.swift
```diff
@@ -0,0 +1,186 @@
...
+//  StripeConnectTests
+//
+//  Created by Chris Mays on 9/4/24.
+//
+
+import XCTest
+import UIKit
+@_spi(PrivateBetaConnect) @testable import StripeConnect
+
+class AppearanceTests: XCTestCase {
```
mludowise-stripe, line 12: 😍

StripeConnect/StripeConnectTests/Internal/Webview/ConnectComponentWebViewTests.swift
```diff
@@ -13,30 +13,96 @@ import WebKit
...
-        try await webView.evaluateMessageWithReply(name: "fetchClientSecret",
+        try await webView.evaluateMessageWithReply(name: "fetchInitParams",
                                                    json: "{}",
-                                                   postReply: false)
+                                                   verifyResult: { result in
+            guard let json = try? JSONEncoder.connectEncoder.encode(message) else {
+                XCTFail("Failed to encode message \(message)")
+                return
+            }
+            XCTAssertEqual(String(data:result, encoding: .utf8), String(data:json, encoding: .utf8))
```
mludowise-stripe, line 52: Why does this need to be encoded to be compared? Could we instead directly compare `result` and `message`?

EDIT: I see now that the type of `evaluateMessageWithReply` changed to send a raw `Data` type instead of the payload type. This makes the function a lot more challenging to read. If the intent is to compare the encoded JSON string, then I think it would be better if the test helper returned the String directly instead of the data type and then having to decode it in the test.

So ideally, this would read something like:
```
try await webView.evaluateMessageWithReply(name: "fetchInitParams",
                                           json: "{}",
                                           verifyResult: { result in
    XCTAssertEqual(result, FetchInitParamsMessageHandler.Reply(locale: "fr-FR"))
}
```
...or this (if it's necessary to test the JSON encoding here as well):
```
    XCTAssertEqual(result, "{\"locale\":\"fr-FR\"}")
```

StripeConnect/StripeConnectTests/WebView+Tests.swift
```diff
@@ -135,23 +137,23 @@ extension WKWebView {
     @discardableResult
     func evaluateMessageWithReply(name: String,
                                   json: String,
-                                  postReply: Bool = true,
-                                  completionHandler: ((Any?, (any Error)?) -> Void)? = nil) async throws -> Any? {
+                                  verifyResult: ((Data) -> Void)? = nil) async throws -> Any? {
```
mludowise-stripe, line 140: Why does `verifyResult` need to be a `Data` type instead of decoding to the Response type like it was before?

StripeConnect/StripeConnectTests/WebView+Tests.swift
```diff
@@ -135,23 +137,23 @@ extension WKWebView {
     @discardableResult
     func evaluateMessageWithReply(name: String,
```
mludowise-stripe, line 138: optional:
I didn't catch this in the previous PR where this helper was added – given that this method is already async, does `verifyResult` need to be a closure? Instead, we could directly return the result and then assert on the result value in the test. This would eliminate the need for an expectation.

So this could look something like:

```
@MainActor
func testSomething() async {
    let result = await webView.evaluateMessageWithReply(name: "foo", json: "{}")
    XCTAssertEquals(result, ...)
}
```

StripeConnect/StripeConnectTests/Internal/Webview/ConnectComponentWebViewTests.swift
```diff
@@ -13,30 +13,96 @@ import WebKit
...
+        })
+        await fulfillment(of: [expectation], timeout: 0.2)
+    }
+    
+    @MainActor
+    func testUpdateAppearance() async throws {
+        let componentManager = EmbeddedComponentManager(apiClient: .init(publishableKey: "test"), fetchClientSecret: {
+            XCTFail("Client secret should not be retrieved in this test")
+            return ""
+        })
```
mludowise-stripe, line 63: optional nit: it looks like this is used a lot in almost every test. Should it get DRYed up into a helper?

StripeConnect/StripeConnectTests/Internal/Webview/ConnectComponentWebViewTests.swift
```diff
@@ -13,30 +13,96 @@ import WebKit
...
+        await fulfillment(of: [expectation], timeout: 1.0)
     }
     
     @MainActor
-    func testFetchClientSecret() async throws {
-        let expectation = XCTestExpectation(description: "Client secret is fetched")
+    func testFetchInitParams() async throws {
+        let expectation = self.expectation(description: "Message received")
+        let message = FetchInitParamsMessageHandler.Reply(locale: "fr-FR")
         let componentManager = EmbeddedComponentManager(apiClient: .init(publishableKey: "test"), fetchClientSecret: {
```
mludowise-stripe, line 35: 🚫 Blocker:
Can we also test that passing an appearance on init gets correctly sent to the webview on `fetchInitParams`?

StripeConnect/StripeConnect/Source/Internal/Webview/ConnectComponentWebView.swift
```diff
@@ -46,18 +47,31 @@ class ConnectComponentWebView: ConnectWebView {
...
+        appearance.traitCollection = self.traitCollection
+        sendMessage(UpdateConnectInstanceSender.init(payload: .init(locale: webLocale.webIdentifier, appearance: appearance)))
+    }
+    
     required init?(coder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
+    
+    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        updateAppearance(appearance: componentManager.appearance)
```
mludowise-stripe, line 73: 🚫 Blocker
I think that this needs to get wrapped in `DispatchQueue.main.async { }` – as of iOS 14, I believe that this is called off the main thread when the app is foregrounded, which causes a crash since it's accessing UIKit classes off the main thread. It's possible this may have been fixed in later iOS versions, but I suspect we still need to handle it since we're supporting 15+

mludowise-stripe (changes requested):
I only reviewed the SDK changes + tests and haven't gotten to the example app yet, but wanted to leave my feedback so far

StripeConnect/StripeConnect/Source/EmbeddedComponentManager+Appearance.swift
```diff
@@ -0,0 +1,239 @@
...
+
+@_spi(STP) import StripeCore
+import UIKit
+
+@_spi(PrivateBetaConnect)
+extension EmbeddedComponentManager {
+    /// Describes the appearance of embedded components.
+    /// - seealso: https://docs.stripe.com/connect/embedded-appearance-options
+    public struct Appearance {
+        var traitCollection: UITraitCollection = .current
```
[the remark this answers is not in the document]
mludowise-stripe: After looking at this again, I guess that `traitCollectionDidChange` is always called before rendering the view, so this is probably less of a concern. But I think it would still be a good idea to move it out of the `Appearance` struct since it's not really part of the Appearance but rather an attribute that determines how to encode the appearance dict.

StripeConnect/StripeConnect/Source/Helpers/Appearance+Encoding.swift
```diff
@@ -0,0 +1,174 @@
...
+        
+        dict["spacingUnit"] = spacingUnit?.pxString
+        
+        return dict
+        
+    }
+}
+
+extension Appearance.CornerRadius {
+    private var mappings: [(key: String, keyPath: WritableKeyPath<Self, CGFloat?>)] {
```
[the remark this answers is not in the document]
nschris-stripe: Yeah that makes sense at one point I had this as a more complicated structure then refactored it without moving it to a dictionary.

StripeConnect/StripeConnect/Source/Internal/Webview/MessageSenders/UpdateConnectInstanceSender.swift
nschris-stripe: Part of the public api for `EmbeddedComponentManager` is making appearance not optional and defaulted to .default. With that in mind is there ever a time where we will send a message that doesn't have an appearance? It seems like the right thing to do here is to not make make appearance optional but also not provide a default value that way it is required that we send some value (ideally always from EmbeddedComponentManager).
```
    public init(apiClient: STPAPIClient = STPAPIClient.shared,
                appearance: Appearance = .default,
                fonts: [CustomFontSource] = [],
                fetchClientSecret: @escaping () async -> String?)
```

StripeConnect/StripeConnectTests/WebView+Tests.swift
nschris-stripe: Because the send messages are no longer decodable. Previously I had everything set as codable, but Appearance can't be Codable.

StripeConnect/StripeConnect/Source/Helpers/CSSHelpers.swift
```diff
@@ -0,0 +1,72 @@
...
+        ]
+    }
+    
+    var cssValue: String? {
+        UIFont.Weight.stringMappings[self]
+    }
+}
+
+extension UIColor {
+    func cssValue(includeAlpha: Bool) -> String {
```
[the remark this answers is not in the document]
nschris-stripe: Since that will require more reviews I am going to add this as a todo: https://jira.corp.stripe.com/browse/MXMOBILE-2753

nschris-stripe:
Diff: https://github.com/stripe/stripe-ios/compare/bd21df9e6db2bd71cc21826ebde3e588cb3c5e24..ed1af4d35368c89f68c69fc154c30932a31ce4c2

nschris-stripe:
One more quick change: https://github.com/stripe/stripe-ios/compare/6af791184c18d38eabf65baa047c76b76d9d7a7b..38c29cd6683569ab007c6d5e3114431c37c3889e

StripeConnect/StripeConnect/Source/Helpers/Appearance+Encoding.swift
```diff
@@ -0,0 +1,135 @@
...
+        dict.mergeAssertingOnOverwrites(cornerRadius.asDictionary())
+        
+        dict["spacingUnit"] = spacingUnit?.pxString
+        
+        return dict
+    }
+}
+
+extension Appearance.CornerRadius {
+    private var mappings: [String: WritableKeyPath<Self, CGFloat?>] {
```
mludowise-stripe, line 34: Could this use `KeyPath` instead of `WritableKeyPath`? It looks like we're only reading from and never writing to the property

StripeConnect/StripeConnect/Source/Helpers/Dictionary+Merge.swift
```diff
@@ -0,0 +1,18 @@
...
+
+extension Dictionary {
+    func mergeKeepingOriginal(_ other: [Key: Value]) -> [Key: Value] {
+        return self.merging(other) { (original, _) in original }
+    }
+    
+    mutating func mergeKeepingOriginalInPlace(_ other: [Key: Value]) {
+        self.merge(other) { (original, _) in original }
+    }
+}
```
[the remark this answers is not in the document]
mludowise-stripe: I opened a separate PR to make these SPI-public: #3990

StripeConnect/StripeConnect/Source/Internal/Webview/MessageSenders/UpdateConnectInstanceSender.swift
mludowise-stripe: Correct, I think the public SDK should always default to the `default` value.

We could theoretically set just `locale` and not `appearance` when the locale is updated (or visa versa), but it's also valid for us to set both any time either is updated since we'll have both of them cached on the view.

My comment was intended to be specific to setting the default value in `UpdateConnectInstanceSender`, and not `EmbeddedComponentManager` with the rationale that we could unintentionally only set `locale` when instantiating `UpdateConnectInstanceSender` and forget to explicitly set appearance to the view's current appearance value. This would result in the appearance getting reset to the default value whenever the app's locale is updated.

Making both `appearance` and `locale` required init arguments in `UpdateConnectInstanceSender` also addresses the issue because it compiler-enforces us to always pass explicit values.

StripeConnect/StripeConnectTests/Internal/Webview/ConnectComponentWebViewTests.swift
```diff
@@ -13,49 +13,127 @@ import WebKit
...
+
+        webView.triggerTraitCollectionChange(style: .dark)
+        
+        let result = try await webView.evaluateMessageWithReply(name: "fetchInitParams",
+                                                                json: "{}")
+        print(result!)
+        XCTAssertEqual(
+            result,
+            """
+            {"appearance":{"variables":{"actionPrimaryColorText":"\(UIColor.white.cssRgbValue)","fontFamily":"-apple-system"}},"fonts":[],"locale":"fr-FR"}
```
mludowise-stripe, line 89: optional nit: What do you think about directly using `#FFFFFF` instead of `UIColor.white.cssRgbValue`?

I think it could make this easier to read but it also ensures the color is getting encoded correctly.

StripeConnect/StripeConnectTests/WebView+Tests.swift
mludowise-stripe: ah, that makes sense. I like your recent change though of comparing the JSON string instead, so I think that addresses this issue.

Example/StripeConnectExample/StripeConnectExample/Appearance/AppSettings+AppearanceInfo.swift
```diff
@@ -0,0 +1,65 @@
...
+}
+
+
+extension AppearanceInfo {
+    static var `default`: AppearanceInfo {
+        .init(displayName: "Default", appearance: .default)
+    }
+    
+    static var shrek: AppearanceInfo {
+        .init(displayName: "Shrek", appearance: .fromURL("https://stripe.com/docs/connect/customize-connect-embedded-components?enable_embedded_theming_demo=true&enable_uiconfig_copy_link=true&enable_new_theming=true&initOptionsOverride=%7B%22appearance%22%3A%7B%22colorPrimary%22%3A%22%235AE92B%22%2C%22colorBackground%22%3A%22%23837411%22%2C%22colorPrimaryButtonBackground%22%3A%22%23DAB9B9%22%2C%22colorPrimaryButtonBorder%22%3A%22%23F00000%22%2C%22colorPrimaryButtonText%22%3A%22%23000000%22%2C%22colorSecondaryButtonBackground%22%3A%22%23025F08%22%2C%22colorText%22%3A%22%23554125%22%2C%22badgeNeutralTextColor%22%3A%22%2328D72A%22%2C%22badgeNeutralBackgroundColor%22%3A%22%23638863%22%2C%22colorSecondaryButtonText%22%3A%22%23000000%22%2C%22fontFamily%22%3A%22fantasy%22%7D%7D")!)
```
mludowise-stripe, line 42: I'm hesitant hardcode the URL encoding from our docs site as a means of encoding the theme. The URL compatibility is not something we guarantee won't change in the future and we want to encourage best practices here as this is an example app for platform integrators to see how to use the SDK.

If the intent is to demonstrate how to easily translate the appearance from the appearance playground, then what I recommend doing instead is:
1. Hardcode the appearance using the `Appearance` object rather than decoding the URL
2. Add a comment above like "[You can view this appearance in our playground]({url})"
3. We can update the playground to also generate Swift code instead of just JS appearance code (see the bottom section in the screenshot below). We'll want to do this when we eventually publicly GA the SDK anyway.

image

Example/StripeConnectExample/StripeConnectExample/Helpers/AppearanceFromURL.swift
```diff
@@ -0,0 +1,192 @@
...
+              let initOptionsOverride = components.queryItems?.first(where: { $0.name == "initOptionsOverride" })?.value,
+              let data = initOptionsOverride.removingPercentEncoding?.data(using: .utf8),
+              let json = try? JSONSerialization.jsonObject(with: data, options: []) as? [String: Any],
+              let appearance = json["appearance"] as? [String: Any] else {
+            return nil
+        }
+        return createAppearance(from: appearance)
+    }
+
+    private static func createAppearance(from dict: [String: Any]) -> Appearance? {
```
mludowise-stripe, line 27: Instead of encoding this into the example app, I think it would make more sense to put into our docs site. See go/grep/generateInitScript for our JS appearance generation code -- we could make a similar one for Swift.

mludowise-stripe (changes requested):
Found a few smallish things and left some feedback on the example app

StripeConnect/StripeConnect/Source/Internal/Webview/ConnectComponentWebView.swift
```diff
@@ -46,18 +47,31 @@ class ConnectComponentWebView: ConnectWebView {
...
+        sendMessage(UpdateConnectInstanceSender.init(payload: .init(locale: webLocale.webIdentifier, appearance: .init(appearance: appearance, traitCollection: traitCollection))))
+    }
+    
     required init?(coder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
+    
+    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        DispatchQueue.main.async {
+            self.updateAppearance(appearance: self.componentManager.appearance)
```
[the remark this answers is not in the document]
nschris-stripe: I don't think it's needed unless we are worried about the webview sticking around until the next runloop.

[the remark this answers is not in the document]
mludowise-stripe: Makes sense -- dropping

nschris-stripe:
Diff: https://github.com/stripe/stripe-ios/compare/38c29cd6683569ab007c6d5e3114431c37c3889e..913d508d810588c48f3c10791ea6aa17c277bb24

mludowise-stripe:

This should pass if you rebase your branch – it's a result of making these SPI-public in #3990

mludowise-stripe (approved):
Looks great! 

Be sure to update the PR description before merging since this no longer includes the URL decoding

PR #3997, opened 2024-09-09
[BANKCON-14297] Require valid email to enter Instant Debits flow

mats-stripe:
## Summary

In the Payment Sheet form for Instant Debits, we now require an email be "valid" (i.e. matches the [Regex pattern for an email](https://www.regular-expressions.info/email.html)) before allowing the user continue. Previously, we enabled the Continue button we the email field was not empty, however, if the email was not valid, and error would be shown immediately when tapped. 

Before:

| Button enabled for non-email | Error when `Continue` is tapped |
|--------|--------|
| Simulator Screenshot - iPhone 15 - 2024-09-09 at 14 54 32 | Simulator Screenshot - iPhone 15 - 2024-09-09 at 14 54 36 | 

After:

https://github.com/user-attachments/assets/a5079026-ecf8-480e-a4f7-b3a13fb1f813

## Motivation

Better UX

## Testing

See before / after above!

## Changelog

N/a


kgaidis-stripe (approved):
💯

PR #3998, opened 2024-09-09
[BANKCON-14310] Add fallback + icon for Add Bank Account row

mats-stripe:
## Summary

Due to an issue with one of our remote icons being inaccessible, we learned that the icon for the `Add bank account` row does not have a local fallback image. This adds support for that in the `RoundedIconView` component. 

## Motivation

Resiliency

## Testing

Both of these screenshots are taken when the `imageUrl` for the icon view is inaccessible:

| Before | After |
|--------|--------|
| Simulator Screenshot - iPhone 15 - 2024-09-09 at 15 45 39 | Simulator Screenshot - iPhone 15 - 2024-09-09 at 16 19 30 | 

## Changelog

N/a

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/RoundedIconView.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/RoundedIconView.swift
@@ -12,8 +12,11 @@ import UIKit
 final class RoundedIconView: UIView {
 
     enum ImageType {
+        /// A local image. Images live in `/Resources/Images`, and defined in `/Source/Helpers/Image.swift`.
         case image(Image)
-        case imageUrl(String?)
+
+        /// A remote image from a given URL, and an optional local image to fallback to
+        case imageUrl(String?, fallback: Image? = nil)
     }
 
     enum Style {
@@ -46,10 +49,10 @@ final class RoundedIconView: UIView {
         switch image {
         case .image(let image):
             iconImageView.image = image.makeImage(template: true)
-        case .imageUrl(let imageUrl):
+        case .imageUrl(let imageUrl, let fallbackImage):
             iconImageView.setImage(
                 with: imageUrl,
-                placeholder: nil,
+                placeholder: fallbackImage?.makeImage(template: true),
                 useAlwaysTemplateRenderingMode: true
             )
         }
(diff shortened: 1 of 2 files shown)


tillh-stripe (approved):

StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/RoundedIconView.swift
```diff
@@ -46,10 +49,10 @@ final class RoundedIconView: UIView {
         switch image {
         case .image(let image):
             iconImageView.image = image.makeImage(template: true)
-        case .imageUrl(let imageUrl):
+        case .imageUrl(let imageUrl, let fallbackImage):
             iconImageView.setImage(
                 with: imageUrl,
-                placeholder: nil,
+                placeholder: fallbackImage?.makeImage(template: true),
```
kgaidis-stripe, line 55: note on consistency, right now this is introducing two different namings:
1) "fallback" image
2) "placeholder" image

is there a difference? if not, then it would make sense that both are the same?

kgaidis-stripe (approved):

StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/RoundedIconView.swift
```diff
@@ -46,10 +49,10 @@ final class RoundedIconView: UIView {
         switch image {
         case .image(let image):
             iconImageView.image = image.makeImage(template: true)
-        case .imageUrl(let imageUrl):
+        case .imageUrl(let imageUrl, let fallbackImage):
```
kgaidis-stripe, line 52: another thing to note (excluding naming), focusing on functionality (mostly), placeholder is an image that appears BEFORE the real image appears

so "fallbackImage" might be misleading (it makes it seem like its a fallback to errors [which is the original problem why this PR was created])

if, for example, the request takes a long time to load, the "placeholder" will keep showing (which may or may not be expected); in this case maybe it does not matter, but if there's a big difference between "placeholder" and "real image from backend" that might be jarring

https://github.com/stripe/stripe-ios/blob/aa04abcd/StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/UIImageView+Extensions.swift#L20

Google Gemini definition of placeholder: "A placeholder image is a temporary image that's used to indicate where an actual image will be placed in a design."

that does fit what the "image loading helper/library" does

mats-stripe: It's a good point - I renamed the field to `placeholder` to make this more clear. In terms of the implementation, the reason this works is because the imageview's image is never updated from the placeholder if there is an error loading the remote image. In terms of the possible jarring switch to the "real" image - I think this is standard behavior for a remote image view, like you said!

StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/RoundedIconView.swift
mats-stripe: Renamed the field to `placeholder`!

kgaidis-stripe (approved):

PR #4004, opened 2024-09-10
Switch UX to only use CVC instead of CVC/CVV

joyceqin-stripe:
## Summary

Replaced cvv with cvc for amex
## Motivation

When entering a CVC with an Amex card we currently show CVV. We should change this logic so that we show CVC like we do for all other card brands.
https://jira.corp.stripe.com/browse/MOBILESDK-1635
## Testing

Screenshot 2024-09-10 at 1 04 48 PM
Screenshot 2024-09-10 at 1 06 43 PM
Screenshot 2024-09-10 at 1 08 06 PM


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/TextField/TextFieldElement+Card.swift
```diff
@@ -168,11 +168,7 @@ extension TextFieldElement {
         let defaultValue: String?
         let cardBrandProvider: () -> (STPCardBrand)
         var label: String {
-            if cardBrandProvider() == .amex {
-                return String.Localized.cvv
-            } else {
-                return String.Localized.cvc
-            }
+            return String.Localized.cvc
```
wooj-stripe, line 171: can be:
```
        let label = String.Localized.cvc
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/CVCRecollectionView.swift
```diff
@@ -65,9 +65,7 @@ class CVCRecollectionView: UIView {
         super.init(frame: .zero)
 
         self.titleLabel.isHidden = mode == .detailedWithInput
-        let brand = (self.paymentMethod.card?.brand ?? .unknown) == .amex
-        ? String.Localized.cvv
-        : String.Localized.cvc
+        let brand = String.Localized.cvc
 
         self.titleLabel.text = String(format: String.Localized.cvc_section_title, brand)
```
wooj-stripe, line 70: No need for a variable anymore, can be:
```
self.titleLabel.text = String(format: String.Localized.cvc_section_title, String.Localized.cvc)
```

porter-stripe (approved):
nice!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/CVCReconfirmationViewController.swift
```diff
@@ -23,7 +23,7 @@ class CVCReconfirmationViewController: UIViewController {
         return navBar
     }()
     private lazy var headerLabel: UILabel = {
-        let headerLabelText = cardBrand == .amex ? String.Localized.confirm_your_cvv : String.Localized.confirm_your_cvc
```
wooj-stripe, line 26: No need for a variable, can be:
```
header.text = String.Localized.confirm_your_cvc
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/CVCReconfirmationVerticalViewController.swift
```diff
@@ -13,7 +13,7 @@ import UIKit
 
 class CVCReconfirmationVerticalViewController: UIViewController {
     private lazy var headerLabel: UILabel = {
-        let headerLabelText = cardBrand == .amex ? String.Localized.confirm_your_cvv : String.Localized.confirm_your_cvc
```
wooj-stripe, line 16: no need for variable. can be:
```
header.text = String.Localized.confirm_your_cvc
```

StripePaymentsUI/StripePaymentsUI/Source/Helpers/String+Localized.swift
```diff
@@ -76,12 +76,9 @@ extension String.Localized {
     @_spi(STP) public static var cvc_section_title: String {
         STPLocalizedString(
             "Confirm your %@",
-            "Section title for entering your CVC/CVV. e.g. 'Confirm your CVC' or 'Confirm your CVV'"
+            "Section title for entering your CVC. e.g. 'Confirm your CVC'"
         )
     }
```
wooj-stripe, line 81: Seems like we can get rid of `cvc_section_title`, and create a new string that doesn't take in a variable.

porter-stripe: ++ to leaving as is, otherwise we need to get new strings translated.

PR #4015, opened 2024-09-12
Add steps in release PR

wooj-stripe:
## Summary
Per last discussion we made some progress in improving the steps. Here is what I propose for now.

## Motivation
Clearer steps in release process until we can automate more

## Testing
None

### DIFF
--- a/ci_scripts/create_release.rb
+++ b/ci_scripts/create_release.rb
@@ -57,7 +57,7 @@ def create_pr
   # Create a new pull request from the branch
   pr_body = %{
   - [ ] Verify CHANGELOG
-    - [ ] Entries are intended to be read by humans. Does each entry make sense?
+    - [ ] Ensure notes for this release are not empty
     - [ ] Release date correct?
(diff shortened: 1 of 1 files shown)


ci_scripts/create_release.rb
```diff
@@ -56,7 +56,16 @@ def run_download_localized_strings
 def create_pr
   # Create a new pull request from the branch
   pr_body = %{
+  - [ ] Verify CHANGELOG
+    - [ ] Entries are intended to be read by humans. Does each entry make sense?
```
yuki-stripe, line 60: These are already reviewed, why review them again?

yuki-stripe (dismissed):

ci_scripts/create_release.rb
wooj-stripe: Ah, true, that should be done during code review, lets remove that and replace with make sure change log is not empty.

yuki-stripe (approved):

PR #4020, opened 2024-09-16
Add `link_card_brand` as LinkMode and synthetic payment method

mats-stripe:
## Summary

As part of [Panther](https://docs.google.com/document/d/1ErJVA3lLvNspPe3A8uYP9feK8Yvfq-Sw5aatNIvlGxk/edit?usp=sharing), this does two things:

- [[BANKCON-14081](https://jira.corp.stripe.com/browse/BANKCON-14081)] Add `link_mode` enum to LinkSettings model
- [[BANKCON-14083](https://jira.corp.stripe.com/browse/BANKCON-14083)] Add `linkCardBrand` as a synthetic payment type

More to come!

## Motivation

Building Panther support!

## Testing

No functional changes introduced yet, but here's the new linkMode value being set:

Screenshot 2024-09-16 at 9 48 33 AM

## Changelog

N/a

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
@@ -17,7 +17,8 @@ extension PaymentSheet {
         case stripe(STPPaymentMethodType)
         case external(ExternalPaymentMethod)
 
-        // Synthetic payment methods:
+        // Synthetic payment methods. These are payment methods which don't have an `STPPaymentMethodType` defined for the same name.
+        // That is, the payment method manifest for a synthetic PMT will show a PMT that doesn't match the form name.
         case instantDebits
         case linkCardBrand
 
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
@@ -16,7 +16,10 @@ extension PaymentSheet {
     enum PaymentMethodType: Equatable, Hashable {
         case stripe(STPPaymentMethodType)
         case external(ExternalPaymentMethod)
+
+        // Synthetic payment methods:
         case instantDebits
+        case linkCardBrand
 
         static var analyticLogForIcon: Set<PaymentMethodType> = []
         static let analyticLogForIconSemaphore = DispatchSemaphore(value: 1)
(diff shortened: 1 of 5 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
```diff
@@ -16,7 +16,10 @@ extension PaymentSheet {
     enum PaymentMethodType: Equatable, Hashable {
         case stripe(STPPaymentMethodType)
         case external(ExternalPaymentMethod)
+
+        // Synthetic payment methods:
```
tillh-stripe, line 20: Nit: Should we add an explanatory comment?

tillh-stripe (approved):

StripePayments/StripePayments/Source/API Bindings/Models/Shared/LinkSettings.swift
```diff
@@ -21,10 +21,17 @@ import Foundation
         case ephemeral
     }
 
+    @_spi(STP) @frozen public enum LinkMode: String {
+        case linkPaymentMethod = "LINK_PAYMENT_METHOD"
+        case passthrough = "PASSTHROUGH"
+        case linkCardBrand = "LINK_CARD_BRAND"
+    }
```
kgaidis-stripe, line 28: would be good to double-check this via backend link but looks good

(particularly important to be extra careful as this is payment SDK)

kgaidis-stripe (approved):

StripePayments/StripePayments/Source/API Bindings/Models/Shared/LinkSettings.swift
kgaidis-stripe: also wroth double checking that this is indeed `@frozen`

for example, can backend add new values? and if they can what would happen client side?

kgaidis-stripe: oh I guess the flatMap makes this very safe:

`(response["link_mode"] as? String).flatMap { LinkMode(rawValue: $0) }`

and since its `@_spi(STP)` it doesn't matter

I was worried about decoding errors

mats-stripe: Here's the enum definition: https://stripe.sourcegraphcloud.com/git.corp.stripe.com/stripe-internal/pay-server/-/blob/lib/consumer/payments_utils/data/link_mode.rb

Might not be a bad idea to get rid of `@frozen` since we don't know for sure things here won't change, but like you said, it doesn't make a huge difference because of `@spi_(STP)`

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
mats-stripe: Done!

PR #4021, opened 2024-09-16
[BANKCON-14524] Allow pay by bank when linkCardBrand criteria is met & dedupe instant debits

mats-stripe:
## Summary

Per the [Panther project plan](https://docs.google.com/document/d/1ErJVA3lLvNspPe3A8uYP9feK8Yvfq-Sw5aatNIvlGxk/edit?usp=sharing), this adds Link Card Brand as a payment method when the following criteria is met:

- `link_funding_sources` contains `BANK_ACCOUNT`
- `US_BANK_ACCOUNT` is not an available payment method.
- `link_mode` is `LINK_CARD_BRAND`

This also makes sure that both Instant Debits and Link Card Brand won't both be shown at the same time, since they appear as identical payment methods to a user (same name, icon, and elements form).

## Motivation

Building Panther support!

## Testing

Added a unit test, and manually verified the new payment method shows up when the conditions are met:

https://github.com/user-attachments/assets/28a1f8d6-abab-4143-ae1f-0a468340718d

## Changelog

N/a

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
+++ b/Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
@@ -196,7 +196,7 @@ class PaymentSheetStandardUITests: PaymentSheetUITestCase {
         // `mc_load_succeeded` event `selected_lpm` should be "apple_pay", the default payment method.
         XCTAssertEqual(analyticsLog[2][string: "selected_lpm"], "apple_pay")
         app.buttons["+ Add"].waitForExistenceAndTap()
-        XCTAssertTrue(app.staticTexts["Add a card"].waitForExistence(timeout: 2))
+        XCTAssertTrue(app.staticTexts["Card information"].waitForExistence(timeout: 2))
 
         // Should fire the `mc_form_shown` event w/ `selected_lpm` = card
         XCTAssertEqual(analyticsLog.last?[string: "event"], "mc_form_shown")
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
@@ -165,12 +165,29 @@ extension PaymentSheet {
                 // External Payment Methods
                 + elementsSession.externalPaymentMethods.map { PaymentMethodType.external($0) }
 
-            if
-                elementsSession.orderedPaymentMethodTypes.contains(.link),
-                !elementsSession.orderedPaymentMethodTypes.contains(.USBankAccount),
-                !intent.isDeferredIntent,
+            // We should manually add Instant Debits as a payment method when:
+            // - Link is an available payment method.
+            // - US Bank Account is *not* an available payment method.
+            // - Not a deferred intent flow.
+            // - Link Funding Sources contains Bank Account.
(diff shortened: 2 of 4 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
```diff
@@ -165,12 +165,29 @@ extension PaymentSheet {
...
+            // - US Bank Account is *not* an available payment method.
+            // - Not a deferred intent flow.
+            // - Link Funding Sources contains Bank Account.
+            var eligibleForInstantDebits: Bool {
+                elementsSession.orderedPaymentMethodTypes.contains(.link) &&
+                !elementsSession.orderedPaymentMethodTypes.contains(.USBankAccount) &&
+                !intent.isDeferredIntent &&
                 elementsSession.linkFundingSources?.contains(.bankAccount) == true
-            {
+            }
```
kgaidis-stripe, line 178: does this need to be a `var` instead of a `let`?

and out of curiosity/learning how does this compile?

I would have expected it to be something like:

```
            let eligibleForInstantDebits: Bool = {
                return elementsSession.orderedPaymentMethodTypes.contains(.link) &&
                !elementsSession.orderedPaymentMethodTypes.contains(.USBankAccount) &&
                !intent.isDeferredIntent &&
                elementsSession.linkFundingSources?.contains(.bankAccount) == true
            }()
```

I get the inferred `return` as they are no longer necessary.

But what are all the other pieces? 

it almost looks like an in-line computed property but the declaration is inside of a function 🤔 so I am not immediately sure the step-by-step execution for it to compile into a boolean

kgaidis-stripe (dismissed):
💯

kgaidis-stripe (approved):

mats-stripe:
Putting this PR on hold until we have [guardrails](https://docs.google.com/document/d/1ErJVA3lLvNspPe3A8uYP9feK8Yvfq-Sw5aatNIvlGxk/edit#bookmark=id.6ascwg3dbqkq) preventing this feature be shown to users while it is being built.

tillh-stripe (dismissed):

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
```diff
@@ -196,7 +196,7 @@ class PaymentSheetStandardUITests: PaymentSheetUITestCase {
         // `mc_load_succeeded` event `selected_lpm` should be "apple_pay", the default payment method.
         XCTAssertEqual(analyticsLog[2][string: "selected_lpm"], "apple_pay")
         app.buttons["+ Add"].waitForExistenceAndTap()
-        XCTAssertTrue(app.staticTexts["Add a card"].waitForExistence(timeout: 2))
+        XCTAssertTrue(app.staticTexts["Card information"].waitForExistence(timeout: 2))
```
mats-stripe, line 199: I had to update this UI test. With Panther enabled there's a second payment method available in this scenario now, and so the `Add a card` text was no longer shown. Here's the changes to this screen with Panther:

| Before (without Panther) | Now (with Panther) |
|--------|--------|
| Simulator Screenshot - iPhone 15 - 2024-09-25 at 11 28 33 | Simulator Screenshot - iPhone 15 - 2024-09-25 at 11 28 03 |

tillh-stripe (approved):

PR #4025, opened 2024-09-17
Add embedded to playground

porter-stripe:
## Summary
- Adds an embedded toggle to the playground that when enabled results in a new view controller being presented with the embedded integration
- Passes in the settings and appearance. In the future we may need to pass more info in, but can add as needed.
- When embedded is selected we enforce that we are not using CSC or horizontal mode 

https://github.com/user-attachments/assets/8ddd0f80-2fb7-433a-955f-3b3363c31ec3

## Motivation
- Easier embedded testing

## Testing
- Manual

## Changelog
N/A


Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
```diff
@@ -233,12 +233,54 @@ struct PaymentSheetTestPlayground: View {
...
         }
     }
+    
+    var uiStyleBinding: Binding<PaymentSheetTestPlaygroundSettings.UIStyle> {
+        Binding<PaymentSheetTestPlaygroundSettings.UIStyle> {
+            return playgroundController.settings.uiStyle
+        } set: { newUIStyle in
+            // If we switch to embedded put style back to vertical and default back to deferred CSC if we are in an intent first integration
+            if newUIStyle == .embedded {
+                playgroundController.settings.layout = .vertical
```
porter-stripe, line 243: Even though embedded doesn't use this layout API, I think it makes the most sense to change the toggle to vertical as that's what most resembles embedded.

yuki-stripe: Is it possible to disable the inapplicable toggles rather than changing their selection?

porter-stripe: Ah yes, easier than I anticipated! I left the logic for changing the confirmation type but now also disable it. I think it makes sense that it always goes back to deferred CSC, then disabling.


Simulator Screenshot - iOS 18  - 2024-09-24 at 14 32 12

yuki-stripe: Switching `normal` -> `deferred_csc` makes sense.

IMO I don't want it to switch to `.vertical` though; that's more confusing than helpful since the value is ignored either way. It should just stay whatever value it was.

porter-stripe: 🤦 I agree, I thought I had removed that but clearly didn't. Fixed!

yuki-stripe (approved):

PR #4027, opened 2024-09-18
Eventing framework v3

wooj-stripe:
## Summary
Adds an eventing early preview of an eventing framework (v3)

## Motivation
Access events for analytics

## Testing
Augmented ui tests w/ payload
Added tests for translating STPAnalyticEvents
Added tests for ensuring events are vended via NotificationCenter

## Example Integration
Users would need to add `@_spi`
```
+ @_spi(MobilePaymentElementAnalyticEventBeta) import StripeCore
```

Add an Observer
```
+        NotificationCenter.default.addObserver(self,
+                                               selector: #selector(mobilePaymentElementNotification(notification:)),
+                                               name: .mobilePaymentElement, object: nil)
```

Define a callback function to receive notifications
```
+    @objc
+    func mobilePaymentElementNotification(notification: NSNotification) {
+        guard let event = notification.object as? MobilePaymentElementEvent else {
+            return
+        }
+        switch event {
+          .............
+        } 
+    }
+
```


davidme-stripe (approved):
Great test coverage!

PR #4030, opened 2024-09-19
[testmode] Rename Down Bank

jlh-stripe:
## Summary
Rename Down Bank, so that its naming convention aligns with other similar institutions.

## Motivation
See above. See also https://github.com/stripe/stripe-android/pull/9299

## Testing
This is a testing-only change. The test will expectedly fail until we update the institution's name internally.

## Changelog

N/A


kgaidis-stripe (approved):
💯

kgaidis-stripe:
CC: @mats-stripe

mats-stripe (dismissed):
Thanks for fixing this!

mats-stripe (approved):

PR #4031, opened 2024-09-20
[StripeConnect] Add onboarding component.

nschris-stripe:
## Summary
This change:
- Implements the onboarding component (UI still work in progress)
- Updates the onSetterFunction. The previous implementation was incorrect because it assumed adding the onSetterFunction message handler for each setter. You can only add 1 message handler per name, so this makes it so you add setter handlers to the primary OnSetterFunctionCalledMessageHandler instead.
- Updates a few unit tests to not load web content.
- Adds an activity indicator. [MXMOBILE-2743](https://jira.corp.stripe.com/browse/MXMOBILE-2743)

## Demo
Video recorded before using StripeSpinner
https://github.com/user-attachments/assets/84e29814-ec25-4f9a-b7d6-d4167c4872c2

Video with Stripe spinner

https://github.com/user-attachments/assets/fbf1ae6c-7ee5-42ab-ac81-9e331bf47760

https://jira.corp.stripe.com/browse/MXMOBILE-2483

PR #4034, opened 2024-09-21
[Connect] Add support for Dashboard app

mludowise-stripe:
## Summary

Adds Dashboard app support for the StripeConnect SDK.

-  **Adds Dashboard app setting override keys**
  If the API client's key is a `uk_` key, then `apiKeyOverride`, `merchantIdOverride`, `platformIdOverride`, and `livemodeOverride` URL parameters are configured and sent to the connect webview, enabling it for use with direct accounts.

- **Fixes the `StripeConnect` package to compile as an SPM package**
  Removes `StripeConnectBundleLocator` and localization files. This file wasn't compiling because all the files in the package's `Resource` folder were empty so `Resource.bundle` wasn't being generated. Since we don't have localized strings currently, we can delete these helpers for now.

  Annotates all public types with `@available(iOS 15, *)`. The Package.swift file has a min supported iOS version of 13 but StripeConnect uses types only available on 15+.

- **Adds the payment-details component for use in the Dashboard**
  Because payment-details will not be included in the beta release, these interfaces are marked with `@_spi(DashboardOnly)` to only make them available to Dashboard.

- **Enables dynamic font scaling**
  Automatically scales appearance font sizes when the component's size class changes and updates the doc strings appropriately.

## Motivation

https://jira.corp.stripe.com/browse/MXMOBILE-2502

## Testing

Unit tests:
- AppearanceTests.testFontSizesChangeBasedOnTraitCollection
- ConnectJSURLParamsTests
- PaymentDetailsViewControllersTests

Manually testing dynamic font sizing and localization:

https://github.com/user-attachments/assets/6f492031-0ca3-4cac-bc09-226f7e9714b1


StripeConnect/StripeConnect/Source/EmbeddedComponentManager.swift
```diff
@@ -12,13 +12,14 @@ import UIKit
 @_spi(PrivateBetaConnect)
 public class EmbeddedComponentManager {
     let apiClient: STPAPIClient
-    
+
```
mludowise-stripe, line 15: This is from the pre-commit hooks auto-linting

StripeConnect/StripeConnect/Source/Helpers/ConnectJSURLParams.swift
```diff
@@ -0,0 +1,66 @@
...
+        self.component = component
+
+        // Validate that publishable key has been set
+        STPAPIClient.validateKey(apiClient.publishableKey)
+
+        if apiClient.publishableKeyIsUserKey {
+            // Dashboard app overrides
+            self.apiKeyOverride = apiClient.publishableKey
+            self.merchantIdOverride = apiClient.stripeAccount
+            self.platformIdOverride = apiClient.stripeAccount
```
mludowise-stripe, line 49: The web wrapper uses two account IDs so it can be flexible to potentially being used for Express as well as Dashboard, even though we don't plan to support that immediately.

If we eventually add support for the Express app, we'll need expose an SPI-public `apiClient.platformAccount` param to set the platform ID.

StripeConnect/StripeConnect/Source/Helpers/StripeConnectBundleLocator.swift
```diff
@@ -1,15 +0,0 @@
...
-//  StripeConnectBundleLocator.swift
-//  StripeConnect
-//
-//  Created by Chris Mays on 7/31/24.
-//
-
-import Foundation
-@_spi(STP) import StripeCore
-
-final class StripeConnectBundleLocator: BundleLocatorProtocol {
```
mludowise-stripe, line 11: To support SPM, we would have needed to add `static let spmResourcesBundle = Bundle.module` ([Example](https://github.com/stripe/stripe-ios/blob/5541e9b3/Stripe/StripeiOS/Source/StripeBundleLocator.swift#L16-L18)). However, SPM only generates this if there's a non-empty file in the `Resources` directory, which we don't yet have.

StripeConnect/StripeConnect/Source/Internal/Webview/ConnectComponentWebView.swift
```diff
@@ -48,25 +48,22 @@ class ConnectComponentWebView: ConnectWebView {
 
         super.init(frame: .zero, configuration: config)
         componentManager.registerChild(self)
-        guard let publishableKey = componentManager.apiClient.publishableKey else {
-            assertionFailure("A publishable key is required. For more info, see https://stripe.com/docs/keys")
-            return
-        }
```
mludowise-stripe, line 54: This is now handled in `ConnectJSURLParams` when we call `STPAPIClient.validateKey`

StripeConnect/StripeConnect/Source/Components/PaymentDetailsViewController.swift
```diff
@@ -0,0 +1,70 @@
...
+        webView = ConnectComponentWebView(
+            componentManager: componentManager,
+            componentType: .paymentDetails
+        )
+        super.init(nibName: nil, bundle: nil)
+        webView.addMessageHandler(OnLoadErrorMessageHandler { [weak self] value in
+            guard let self else { return }
+            self.delegate?.paymentDetailsLoadDidFail(self, withError: value.error.connectEmbedError)
+        })
+        // TODO: Add support for `setOnClose`
```
mludowise-stripe, line 29: Blocked on updating `OnSetterFunctionCalledMessageHandler`.

See https://github.com/stripe/stripe-ios/pull/4031#discussion_r1769634062

StripeCore/StripeCore/Source/API Bindings/STPAPIClient.swift
```diff
@@ -142,7 +142,7 @@ import UIKit
     // MARK: Helpers
 
     static var didShowTestmodeKeyWarning = false
-    class func validateKey(_ publishableKey: String?) {
+    @_spi(STP) public class func validateKey(_ publishableKey: String?) {
```
mludowise-stripe, line 145: Made SPI-public so it could be referenced in `ConnectJSURLParams`

StripeConnect/StripeConnect/Source/EmbeddedComponentManager+Appearance.swift
```diff
@@ -103,7 +103,12 @@ extension EmbeddedComponentManager {
             public var labelMd: Style = .init()
             /// Describes the font size and weight for the small label typography.
             public var labelSm: Style = .init()
-            
+
+            /// Treats the given font sizes as unscaled sizes and scales the
+            /// appearance when the component's trait collection is updated
+            /// - Note: This is an experimental feature that may be removed at any time.
+            @_spi(ExperimentalFontScaling) public var automaticallyScaleFontSize = false
```
mludowise-stripe, line 110: Chose to use `ExperimentalFontScaling` instead of `DashboardOnly` because it hasn't gone through API review and needs to be SPI-protected, but we don't want integrators to import with `DashboardOnly` just to get access to it.

If it works well for Dashboard, we'll likely put it through API review and expose publicly.

StripeConnect/StripeConnect/Source/EmbeddedComponentManager+Appearance.swift
```diff
@@ -80,9 +81,10 @@ extension EmbeddedComponentManager {
             ///   `CustomFontSource` when initializing the `EmbeddedComponentManager` before
             ///   referencing them in the appearance's `typography.font` property.
             public var font: UIFont?
-            /// The baseline font size set on the embedded component root.
-            /// This scales the value of other font size variables.
-            public var fontSizeBase: CGFloat?
+            /// The unscaled baseline font size set on the embedded component root.
+            /// This scales the value of other font size variables and is automatically scaled
+            /// when the component's size category is updated.
+            public var fontSizeBase: CGFloat? = 16
```
mludowise-stripe, line 87: I had to hardcode the default so that scaling was enabled by default. The only other option would be to send the default value from the web view on init.

This seems okay given that we put in our public docs that this has a default of 16pt

mludowise-stripe:
temporarily setting back to draft while i work on rebasing

StripeConnect/StripeConnect/Source/Components/PaymentDetailsViewController.swift
nschris-stripe: This should no longer be blocked, but it can go in another PR.

mludowise-stripe: I think I'm going to delete the comment – I think we actually don't want to support `setOnClose` anyway since it should get handled in the app layer (we need to likely remove the X button from the webview)

nschris-stripe (approved):

nschris-stripe (approved):

PR #4043, opened 2024-09-23
Adding new device information fields and updating older ones 

Pearapps:
Simply updates some fields according to the newest spec, and to get the device information test passing.


Stripe3DS2/Stripe3DS2/STDSDeviceInformationParameter.m
```diff
@@ -391,7 +428,7 @@ + (instancetype)systemLocale {
     return [[STDSDeviceInformationParameter alloc] initWithIdentifier:@"I010"
                                                       permissionCheck:nil
                                                            valueCheck:^id _Nullable{
-                                                               return [NSLocale currentLocale].localeIdentifier;
+                                                               return [[NSLocale currentLocale].localeIdentifier stringByReplacingOccurrencesOfString:@"_" withString:@"-"];
```
davidme-stripe, line 431: Note that this could be a string like "zh-Hant_HK_POSIX@collation=pinyin;currency=CNY". The spec says this "Returns the ID of the generic locale that contains fixed “backstop” settings that provide values for otherwise undefined keys." That matches better with Apple's description of [NSLocale systemLocale], so I wonder if that's what they're asking for? https://developer.apple.com/documentation/foundation/nslocale/1407691-systemlocale

PR #4050, opened 2024-09-24
[BANKCON-14528] Update FC SDK to accept `link_mode` value from the payment sheet

mats-stripe:
## Summary

As part of the [Panther](https://docs.google.com/document/d/1ErJVA3lLvNspPe3A8uYP9feK8Yvfq-Sw5aatNIvlGxk/edit#heading=h.p9ofrc4zbe1z) project, we'll need to call a different payment method endpoint at the end of the Financial Connections flow based on whether or not we're in a Panther flow. To know this, we need the `link_mode` value, which is fetched in the elements session response. This passes that value into the FC SDK for future use!

## Motivation

Building Panther!

## Testing

Verified the correct `linkMode` value is accessed in the `FinancialConnectionsSheet`:

Screenshot 2024-09-24 at 9 55 27 AM

## Changelog

N/a


tillh-stripe (dismissed):

StripeCore/StripeCore/Source/Connections Bindings/ElementsContext.swift
```diff
@@ -0,0 +1,16 @@
+//
+//  ElementsContext.swift
+//  StripeCore
+//
+//  Created by Mat Schmid on 2024-09-23.
+//
+
+import Foundation
+
+@_spi(STP) public struct ElementsContext {
```
porter-stripe, line 10: nit: `ElementsContext` is a bit vauge, can we either add a doc comment about what/where this is being populated from e.g. elements/session or rename to ElementsSessionContext

porter-stripe (dismissed):

porter-stripe (approved):

PR #4051, opened 2024-09-24
Add `transformed` operation to `Future`

mats-stripe:
## Summary

This adds a new operation to our fork of `Future`, `transformed(with:)`. This will allow us to apply a simple synchronous transform to a `Promise` value. 

> [!NOTE]  
>Our implementation of `Future` is [quoted](https://github.com/stripe/stripe-ios/blob/2936441f991dec772c8add8a0d318aacb577831c/StripeCore/StripeCore/Source/Helpers/Async.swift#L10) to being taken from [Swift by Sundell](https://github.com/JohnSundell/SwiftBySundell/blob/master/Blog/Under-the-hood-of-Futures-and-Promises.swift). This implementation of `transformed(with:)` also comes from the [same source](https://github.com/JohnSundell/SwiftBySundell/blob/7f16bece0b1de776b870a5ed8925e1b93726d512/Blog/Under-the-hood-of-Futures-and-Promises.swift#L114-L122). 
>
>Here is [the author's blog post](https://swiftbysundell.com/articles/under-the-hood-of-futures-and-promises-in-swift/) with some more details.

At its core, this unlocks basic functionally for transforming a `Promise` value:

```swift
let promise = Promise<Int>(value: 42)

let doubledFuture = promise.transformed { value in
    return value * 2
}

doubledFuture.observe { result in
    switch result {
    case .success(let value):
        print("Transformed value: \(value)") // Transformed value: 84
    case .failure(let error):
        print("Error: \(error)")
    }
}
```

At a practical level, this unlocks type erasure of `Future` values. Here's a simple example:

```swift
protocol Animal {
    var name: String { get }
}

struct Dog: Animal {
    let name: String
}

struct Cat: Animal {
    let name: String
}

func getRandomAnimal() -> Future<Animal> {
    let randomNumber = Int.random(in: 0...1)
    if randomNumber == 0 {
        return getRandomDog() // Cannot convert value of type 'Future<Dog>' to closure result type 'Future<any Animal>'
    } else {
        return getRandomCat()
            .transformed { $0 as Animal } // Works!
    }
}

func getRandomDog() -> Future<Dog> {
    let promise = Promise<Dog>()
    promise.resolve(with: Dog(name: "Buddy"))
    return promise
}

func getRandomCat() -> Future<Cat> {
    let promise = Promise<Cat>()
    promise.resolve(with: Cat(name: "Whiskers"))
    return promise
}
```

In this scenario,  the `getRandomAnimal()` function would get a compile error: 

```
Cannot convert value of type 'Future<Dog>' to closure result type 'Future<any Animal>'
```

However, we fix that error bellow using `.transformed { $0 as Animal }`

## Motivation

Unlocks type erasure of `Future` values.

## Testing

Added a bunch of unit tests for `Async.swift`!

## Changelog

N/a - these methods are all behind a `@_spi(STP)` class.

### DIFF
--- a/StripeCore/StripeCore.xcodeproj/project.pbxproj
+++ b/StripeCore/StripeCore.xcodeproj/project.pbxproj
@@ -37,6 +37,7 @@
 		45DAE581F74EF7E11C64212B /* InstallMethod.swift in Sources */ = {isa = PBXBuildFile; fileRef = 21E64986F72C7BD8B1105A95 /* InstallMethod.swift */; };
 		48A6CCB4008A5060C2655C5F /* XCTestCase+Stripe.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2DE48D0086BED21F9E837D0B /* XCTestCase+Stripe.swift */; };
 		4910B9282C3D8F3F00B030D4 /* Result+Extensions.swift in Sources */ = {isa = PBXBuildFile; fileRef = 4910B9272C3D8F3F00B030D4 /* Result+Extensions.swift */; };
+		49ECDA412CA340E100F647F0 /* AsyncTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 49ECDA402CA340E100F647F0 /* AsyncTests.swift */; };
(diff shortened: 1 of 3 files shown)


porter-stripe (approved):
✅  hopefully we can get FC to use Swift async/await in the future!

PR #4055, opened 2024-09-25
[WIP] [Connect] Add logout()

mludowise-stripe:
## Summary

Adds the `logout()` call to EmbeddedComponentManager

In order to ensure that logout works, even when no components have been created, a webview is instantiated in the background so we can call `window.logout()` to clear cookies and make a call to the logout endpoint.

## Motivation

https://jira.corp.stripe.com/browse/MXMOBILE-2488

### DIFF
--- a/StripeConnect/StripeConnectTests/Internal/Webview/MessageSenders/LogoutSenderTests.swift
+++ b/StripeConnect/StripeConnectTests/Internal/Webview/MessageSenders/LogoutSenderTests.swift
@@ -1,0 +1,24 @@
+//
+//  LogoutSenderTests.swift
+//  StripeConnectTests
+//
+//  Created by Mel Ludowise on 9/24/24.
+//
+
+@testable import StripeConnect
+import XCTest
+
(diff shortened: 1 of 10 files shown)


StripeConnect/StripeConnectTests/Internal/Webview/MessageSenders/LogoutSenderTests.swift
```diff
@@ -0,0 +1,24 @@
...
+class LogoutSenderTests: ScriptWebTestBase {
+    func testSendMessage() throws {
+        try validateMessageSent(sender: LogoutSender())
+    }
+
+    func testSenderSignature() {
+        XCTAssertEqual(
+            LogoutSender().javascriptMessage,
+            """
+            window.logout({});
```
mludowise-stripe, line 20: Ideally this would be empty args, but there's no easy way to do that with the MessageSender. I tested that sending extra arguments to this function doesn't seem to have any negative effect in JS.

PR #4057, opened 2024-09-25
[BANKCON-14856] Pass link_mode to the web flow params for Panther payments

mats-stripe:
Web PR: https://git.corp.stripe.com/stripe-internal/pay-server/pull/896974

## Summary

The web Instant Debits flow launched from the mobile FC SDK will have slightly different behaviour when going through the Panther flow ([context](https://docs.google.com/document/d/1ErJVA3lLvNspPe3A8uYP9feK8Yvfq-Sw5aatNIvlGxk/edit#bookmark=id.oljqxd4uzx7t)). This PR passes the `link_mode` to the web flow as a URL argument. `link_mode` will be used to determine if we're currently in a Panther flow.

## Motivation

Building Panther!

## Testing

Bunch of unit tests added!

## Changelog

N/a


StripeFinancialConnections/StripeFinancialConnections/Source/Web/FinancialConnectionsWebFlowViewController.swift
```diff
@@ -416,3 +420,20 @@ private extension FinancialConnectionsWebFlowViewController {
...
+
+extension FinancialConnectionsWebFlowViewController {
+    static func updateAdditionalParameters(
+        startingAdditionalParameters: String?,
+        isInstantDebits: Bool,
+        linkMode: LinkMode?
+    ) -> String? {
+        var additionalQueryParameters = startingAdditionalParameters
+        if isInstantDebits {
+            additionalQueryParameters = (additionalQueryParameters ?? "") + "&return_payment_method=true"
```
tillh-stripe, line 432: Nit: Should we just do the nil coalescing once at the start of this method? It would mean that we always return a non-nil string, but that seems fine to me.

tillh-stripe (dismissed):

StripeFinancialConnections/StripeFinancialConnections/Source/Web/FinancialConnectionsWebFlowViewController.swift
mats-stripe: Done! https://github.com/stripe/stripe-ios/pull/4057/commits/5a1b90f8d43a8e216ea615c2ccceb3d039421d1b Worked out pretty nicely!

porter-stripe (approved):

PR #4067, opened 2024-09-27
Embedded example 

yuki-stripe:
## Summary
- Add ExampleEmbeddedElementCheckoutViewController to example app
- Tweak Embedded API

## Testing
Manual for now

## Changelog
Not user facing

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/ExampleEmbeddedElementCheckoutViewController.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/ExampleEmbeddedElementCheckoutViewController.swift
@@ -1,0 +1,418 @@
+//
+//  ExampleEmbeddedElementCheckoutViewController.swift
+//  PaymentSheet Example
+//
+//  Created by Yuki Tokuhiro on 9/25/24.
+//
+
+// TODO: Not testable
+@_spi(STP) @testable import StripePaymentSheet
+import UIKit
+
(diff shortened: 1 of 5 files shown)


Example/PaymentSheet Example/PaymentSheet Example/ExampleEmbeddedElementCheckoutViewController.swift
```diff
@@ -0,0 +1,391 @@
...
+
+    @objc
+    func didTapPaymentMethodButton() {
+        let paymentMethodsViewController = PaymentMethodsViewController(embeddedPaymentElement: embeddedPaymentElement)
+        let navController = UINavigationController(rootViewController: paymentMethodsViewController)
+        present(navController, animated: true)
+    }
+
+    @objc
+    func didTapCheckoutButton() async {
```
porter-stripe, line 98: Surprised this works, didn't think you could use @objc + async on a function definition.

porter-stripe (approved):

Example/PaymentSheet Example/PaymentSheet Example/ExampleEmbeddedElementCheckoutViewController.swift
yuki-stripe: Me too!  This is actually untested, buy button won't work until we implement `paymentOption`.  So it might crash at runtime :D

PR #4068, opened 2024-09-27
[DRAFT] Start on embedded init/create

porter-stripe:

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/EmbeddedPlaygroundViewController.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/EmbeddedPlaygroundViewController.swift
@@ -6,7 +6,7 @@
 //
 
 import Foundation
-@_spi(EmbeddedPaymentElementPrivateBeta) import StripePaymentSheet
+@_spi(EmbeddedPaymentElementPrivateBeta) @_spi(STP) @_spi(ExperimentalAllowsRemovalOfLastSavedPaymentMethodAPI) import StripePaymentSheet
 import UIKit
 
 protocol EmbeddedPlaygroundViewControllerDelegate: AnyObject {
@@ -14,11 +14,20 @@ protocol EmbeddedPlaygroundViewControllerDelegate: AnyObject {
 }
 
 class EmbeddedPlaygroundViewController: UIViewController {
-
-    private let settings: PaymentSheetTestPlaygroundSettings
     private let appearance: PaymentSheet.Appearance
-
+    private let intentConfig: PaymentSheet.IntentConfiguration
+    private let configuration: EmbeddedPaymentElement.Configuration
+    
+    private var embeddedPaymentElement: EmbeddedPaymentElement!
     weak var delegate: EmbeddedPlaygroundViewControllerDelegate?
+    
+    private lazy var loadingIndicator: UIActivityIndicatorView = {
+        let indicator = UIActivityIndicatorView(style: .medium)
+        indicator.translatesAutoresizingMaskIntoConstraints = false
+        indicator.hidesWhenStopped = true
+        return indicator
+    }()
+    
 
     private lazy var checkoutButton: UIButton = {
         let checkoutButton = UIButton(type: .system)
@@ -31,9 +40,14 @@ class EmbeddedPlaygroundViewController: UIViewController {
         return checkoutButton
     }()
 
-    init(settings: PaymentSheetTestPlaygroundSettings, appearance: PaymentSheet.Appearance) {
-        self.settings = settings
+    init(configuration: PaymentSheet.Configuration, intentConfig: PaymentSheet.IntentConfiguration, appearance: PaymentSheet.Appearance, delegate: EmbeddedPlaygroundViewControllerDelegate?) {
         self.appearance = appearance
+        self.intentConfig = intentConfig
+        self.configuration = .init(from: configuration, formSheetAction: .confirm(completion: { result in
+            // TODO(porter) Probably pass in formSheetAction from PlaygroundController based on some toggle in the UI
+            delegate?.didComplete(with: result)
+        }), hidesMandateText: false)
+        
         super.init(nibName: nil, bundle: nil)
     }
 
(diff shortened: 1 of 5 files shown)


Example/PaymentSheet Example/PaymentSheet Example/EmbeddedPlaygroundViewController.swift
```diff
@@ -51,41 +64,97 @@ class EmbeddedPlaygroundViewController: UIViewController {
...
-        return STPPaymentMethod.decodedObject(fromAPIResponse: amex)
+extension EmbeddedPaymentElement.Configuration {
+
+    /// Initializes an EmbeddedPaymentElement.Configuration from a given PaymentSheet.Configuration.
+    ///
+    /// - Parameters:
+    ///   - paymentSheetConfig: The PaymentSheet.Configuration instance to convert from.
+    ///   - formSheetAction: The FormSheetAction specific to EmbeddedPaymentElement.Configuration.
+    ///   - hidesMandateText: Determines whether to hide mandate text. Defaults to `false`.
+    public init(
```
yuki-stripe, line 128: How will we add Embedded-specific stuff like `hidesMandateText` or `formSheetAction` if we go from playground settings -> PaymentSheet.Configuration -> EmbeddedConfiguration?

Maybe we should make the "playground settings -> EmbeddedConfiguration" logic happen in the PlaygroundController or something?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
```diff
@@ -55,19 +55,59 @@ public class EmbeddedPaymentElement {
         intentConfiguration: IntentConfiguration,
         configuration: Configuration
     ) async throws -> EmbeddedPaymentElement {
-        // TODO(https://jira.corp.stripe.com/browse/MOBILESDK-2525)
-        let dummyView = await EmbeddedPaymentMethodsView(
-            savedPaymentMethod: nil,
-            appearance: .default,
-            shouldShowApplePay: true,
-            shouldShowLink: true
+        let paymentSheetConfiguration = configuration.makePaymentSheetConfiguration()
```
yuki-stripe, line 58: This is a temporary hack right?  If so, let's document it

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
```diff
@@ -50,11 +46,20 @@ import UIKit
                     return nil
                 }
             }()
-            stackView.addArrangedSubview(RowButton.makeForSavedPaymentMethod(paymentMethod: savedPaymentMethod,
-                                                                             appearance: rowButtonAppearance,
-                                                                             rightAccessoryView: accessoryButton,
-                                                                             isEmbedded: true,
-                                                                             didTap: handleRowSelection(selectedRowButton:)))
```
yuki-stripe, line 57: Ah this makes me wonder, do the other RowButtons have a retain cycle if they pass a selector directly to didTap?   Like on line 44

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
```diff
@@ -180,3 +220,41 @@ extension EmbeddedPaymentElement {
     public typealias BillingDetailsCollectionConfiguration = PaymentSheet.BillingDetailsCollectionConfiguration
     public typealias ExternalPaymentMethodConfiguration = PaymentSheet.ExternalPaymentMethodConfiguration
 }
+
+// TODO(porter) Create a protocol for the commonalities between PaymentSheet.Configuration <> EmbeddedPaymentElement.Configuration
```
yuki-stripe, line 224: Are you gonna follow up on this immediately or should we make a ticket

Example/PaymentSheet Example/PaymentSheet Example/EmbeddedPlaygroundViewController.swift
porter-stripe: Yeah I think creating the EmbeddedConfiguration in the PlaygroundController sounds good, I was thinking we could just inject those two properties into this view controller but that doesn't scale well if we add more. Will update this in this PR.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentMethodsViewSnapshotTests.swift
```diff
@@ -472,6 +501,44 @@ class EmbeddedPaymentMethodsViewSnapshotTests: STPSnapshotTestCase {
         verify(embeddedView)
     }
 
+    // MARK: Initial selection tests
```
yuki-stripe, line 504: Might as well add an assert on selection, too.
> EmbeddedPaymentMethodsView.swift: warning: Property 'selection' is assigned, but never used

porter-stripe: Good idea!

Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
```diff
@@ -177,6 +177,75 @@ class PlaygroundController: ObservableObject {
         return configuration
     }
 
+    var embeddedConfiguration: EmbeddedPaymentElement.Configuration {
+        var configuration = EmbeddedPaymentElement.Configuration(formSheetAction: .confirm(completion: { [weak self] result in
+            // TODO(porter) Handle two step confirm
```
yuki-stripe, line 182: ticket or immediate followup? We should probably make an embedded-only `formSheetAction` toggle in the playground

yuki-stripe (approved):

Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
porter-stripe: I can make a ticket, in my head I was thinking the week of Oct. 7 it would be added but I can add sooner

porter-stripe: I'll make a follow up right now and add the hide mandate API too.

PR #4080, opened 2024-09-30
Add formSheetAction and hidesMandateText to playground

porter-stripe:


Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
```diff
@@ -151,6 +151,17 @@ struct PaymentSheetTestPlayground: View {
                         SettingView(setting: $playgroundController.settings.collectPhone)
                         SettingView(setting: $playgroundController.settings.collectAddress)
                     }
+                    Divider()
+                    Group {
```
yuki-stripe, line 155: Can we hide this if not embedded?

yuki-stripe (approved):

Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
porter-stripe: ah good idea, yes

Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
```diff
@@ -151,6 +151,17 @@ struct PaymentSheetTestPlayground: View {
                         SettingView(setting: $playgroundController.settings.collectPhone)
                         SettingView(setting: $playgroundController.settings.collectAddress)
                     }
+                    Divider().hidden(playgroundController.settings.uiStyle != .embedded)
```
yuki-stripe, line 154: ooc can you not put the divider inside the group below?

Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
```diff
@@ -577,3 +588,9 @@ struct PaymentSheetTestPlayground_Previews: PreviewProvider {
         PaymentSheetTestPlayground(settings: .defaultValues())
     }
 }
+
+extension View {
+    func hidden(_ shouldHide: Bool) -> some View {
+        opacity(shouldHide ? 0 : 1)
+    }
+}
```
yuki-stripe, line 596: iiuc this approach leaves an empty space when `shouldHide` is true, is that right?

If so, i think we want to hide by conditionally including the group itself.  Like
```
if playgroundController.settings.uiStyle == .embedded {
  Group { ... }
}
```

porter-stripe: oh yeah better nice

yuki-stripe (approved):

Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
porter-stripe: Yeah possible, for some reason we have all our dividers outside their groups 🤔

PR #4081, opened 2024-10-01
Introduce IntegrationConfigurable

porter-stripe:
## Summary
- Adds `IntegrationConfigurable` which is a protocol that defines similarities between `PaymentSheet.Configuration` and `EmbeddedPaymentElement.Configuration`.
- Injects a `IntegrationConfigurable` where currently needed
- Moves some extensions on `PaymentSheet.Configuration` to `IntegrationConfigurable`
- None of these changes broke CustomerSheet, so I did not conform `CustomerSheet.Configuration` to `IntegrationConfigurable` as it's not currently needed. It might be in the future however.

## Motivation
- Cleaner code
- https://jira.corp.stripe.com/browse/MOBILESDK-2533

## Testing
- Compiler
- CI

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Analytics/PaymentSheetAnalyticsHelper.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Analytics/PaymentSheetAnalyticsHelper.swift
@@ -12,7 +12,7 @@ import Foundation
 final class PaymentSheetAnalyticsHelper {
     let analyticsClient: STPAnalyticsClient
     let isCustom: Bool
-    let configuration: IntegrationConfigurable
+    let configuration: PaymentSheet.Configuration
 
     // Vars set later as PaymentSheet successfully loads, etc.
     var intent: Intent?
@@ -22,7 +22,7 @@ final class PaymentSheetAnalyticsHelper {
 
     init(
         isCustom: Bool,
-        configuration: IntegrationConfigurable,
+        configuration: PaymentSheet.Configuration,
         analyticsClient: STPAnalyticsClient = .sharedClient
     ) {
         self.isCustom = isCustom
@@ -352,38 +352,5 @@ extension PaymentSheet.Configuration {
         payload["preferred_networks"] = preferredNetworks?.map({ STPCardBrandUtilities.apiValue(from: $0) }).joined(separator: ", ")
         payload["payment_method_layout"] = paymentMethodLayout.description
         return payload
-    }
-}
-
-extension EmbeddedPaymentElement.Configuration {
-    /// Serializes the configuration into a safe dictionary containing no PII for analytics logging
-    var analyticPayload: [String: Any] {
-        var payload = [String: Any]()
-        payload["allows_delayed_payment_methods"] = allowsDelayedPaymentMethods
-        payload["apple_pay_config"] = applePay != nil
-        payload["style"] = style.rawValue
-        payload["customer"] = customer != nil
-        payload["customer_access_provider"] = customer?.customerAccessProvider.analyticValue
-        payload["return_url"] = returnURL != nil
-        payload["default_billing_details"] = defaultBillingDetails != PaymentSheet.BillingDetails()
-        payload["save_payment_method_opt_in_behavior"] = savePaymentMethodOptInBehavior.description
-        payload["appearance"] = appearance.analyticPayload
-        payload["billing_details_collection_configuration"] = billingDetailsCollectionConfiguration.analyticPayload
-        payload["preferred_networks"] = preferredNetworks?.map({ STPCardBrandUtilities.apiValue(from: $0) }).joined(separator: ", ")
-        payload["form_sheet_action"] = formSheetAction.analyticValue
-        payload["hide_mandate_text"] = hidesMandateText
-
-        return payload
-    }
-}
-
-extension EmbeddedPaymentElement.Configuration.FormSheetAction {
-    var analyticValue: String {
-        switch self {
-        case .confirm:
-            return "confirm"
-        case .continue:
-            return "continue"
-        }
     }
 }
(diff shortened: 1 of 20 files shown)


StripePaymentSheet/StripePaymentSheet/Source/Analytics/PaymentSheetAnalyticsHelper.swift
```diff
@@ -354,3 +354,36 @@ extension PaymentSheet.Configuration {
...
+        payload["customer"] = customer != nil
+        payload["customer_access_provider"] = customer?.customerAccessProvider.analyticValue
+        payload["return_url"] = returnURL != nil
+        payload["default_billing_details"] = defaultBillingDetails != PaymentSheet.BillingDetails()
+        payload["save_payment_method_opt_in_behavior"] = savePaymentMethodOptInBehavior.description
+        payload["appearance"] = appearance.analyticPayload
+        payload["billing_details_collection_configuration"] = billingDetailsCollectionConfiguration.analyticPayload
+        payload["preferred_networks"] = preferredNetworks?.map({ STPCardBrandUtilities.apiValue(from: $0) }).joined(separator: ", ")
+        payload["form_sheet_action"] = formSheetAction.analyticValue
+        payload["hide_mandate_text"] = hidesMandateText
```
porter-stripe, line 374: Note these new embedded only params.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/IntegrationConfigurable.swift
```diff
@@ -0,0 +1,63 @@
...
+//
+
+import Foundation
+import UIKit
+@_spi(STP) import StripePayments
+
+/// Represents shared configuration properties shared between integration surfaces in mobile payment element.
+/// - Note: See the concrete implementations of `IntegrationConfigurable` for detailed doc comments.
+/// - Note: Not currently used by CustomerSheet.
+protocol IntegrationConfigurable: PaymentMethodRequirementProvider {
```
porter-stripe, line 15: Open to naming suggestions

porter-stripe:
I initially thought the dead code check was confused, but it's not, it's correct! I can comment out all those properties on `IntegrationConfigurable` and the project still compiles. This is b/c no where in the project where we inject a `IntegrationConfigurable` do we use any of those properties (yet).

I added the skip dead code check b/c I feel the spirit of the protocol is to define the commonalities between config objects and even though those properties are not used yet they probably will be as we build out embedded.

Options:
1. Skip dead code check, add all commonalities to `IntegrationConfigurable` as we have in the PR
2. Remove the flagged code, add properties to `IntegrationConfigurable` as needed in future PRs.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerAddPaymentMethodViewController.swift
```diff
@@ -244,7 +244,7 @@ class CustomerAddPaymentMethodViewController: UIViewController {
             isSettingUp: true,
             countryCode: nil,
             savePaymentMethodConsentBehavior: savePaymentMethodConsentBehavior,
-            analyticsHelper: .init(isCustom: false, configuration: .init()) // Just use a dummy analytics helper; we don't look at these analytics.
+            analyticsHelper: .init(isCustom: false, configuration: PaymentSheet.Configuration.init()) // Just use a dummy analytics helper; we don't look at these analytics.
```
yuki-stripe, line 247: Can you add https://jira.corp.stripe.com/browse/MOBILESDK-2548 to this comment

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/IntegrationConfigurable.swift
yuki-stripe: Let's call it PaymentElementConfiguration? I think that's the most accurate name, since this is Configuration for all Mobile Payment Element integration variants.  We just need to start thinking of PaymentSheet, FlowController, EmbeddedPaymentElement as Payment Element variants.

StripePaymentSheet/StripePaymentSheet/Source/Analytics/PaymentSheetAnalyticsHelper.swift
```diff
@@ -12,7 +12,7 @@ import Foundation
 final class PaymentSheetAnalyticsHelper {
     let analyticsClient: STPAnalyticsClient
     let isCustom: Bool
-    let configuration: PaymentSheet.Configuration
+    let configuration: IntegrationConfigurable
```
yuki-stripe, line 15: I wonder if we should keep this `PaymentSheet.Configuration`?  I'm not sure this class can reasonably also handle analytics for EMPE

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/IntegrationConfigurable.swift
yuki-stripe: This means we should probably rename things like `PaymentSheetLoader` to be `PaymentElementLoader`.

porter-stripe: Agree on renaming other PaymentSheet prefixed classes, maybe something we should tackle when we rename the module.

StripePaymentSheet/StripePaymentSheet/Source/Analytics/PaymentSheetAnalyticsHelper.swift
porter-stripe: Yeah looking at the class more, it'll be pretty hard. We'll want some embedded specific one.

yuki-stripe (approved):

PR #4084, opened 2024-10-02
Add `device_id` to common V2 analytics payload

mludowise-stripe:
## Summary
Adds a `device_id` param to the analytics payload which contains the device's vendor ID. The vendor ID is unique per app install and can enable us to dedupe analytic events by device/user in situations we otherwise might not be able to.

Practical examples:
- Identity: Allows us to determine if multiple verifications are created for a single user rather than reusing the same verification
- Connect: Allows us to understand how many users are actively using embedded components on the same connected account
- Connect: Logs a unique ID before the component has loaded, enabling us to track if the same user is repeatedly encountering errors preventing the component to load vs. spread across multiple users

Note: Because `device_id` is getting added to the common V2 analytics payload, this means that the Identity and FinancialConnections SDKs will also send `device_id` to r.stripe.com, however additional work will be needed to consume this value from the request in the analytics table.

## Motivation
https://jira.corp.stripe.com/browse/MXMOBILE-2490

## Testing
Unit tests

### DIFF
--- a/StripeCore/StripeCore/Source/Analytics/AnalyticsClientV2.swift
+++ b/StripeCore/StripeCore/Source/Analytics/AnalyticsClientV2.swift
@@ -134,6 +134,9 @@ extension AnalyticsClientV2Protocol {
             "install": InstallMethod.current.rawValue,
             "app_bundle_id": Bundle.stp_applicationBundleId() ?? "",
         ]
+        if let deviceId = UIDevice.current.identifierForVendor?.uuidString {
+            payload["device_id"] = deviceId
+        }
 
         return payload
     }
--- a/StripeCore/StripeCoreTests/Analytics/AnalyticsClientV2Test.swift
+++ b/StripeCore/StripeCoreTests/Analytics/AnalyticsClientV2Test.swift
@@ -58,9 +58,13 @@ final class AnalyticsClientV2Test: XCTestCase {
         XCTAssertNotNil(commonPayload["app_name"] as? String)
         XCTAssertNotNil(commonPayload["app_version"] as? String)
 
+        // Verify this is a valid UUID
+        XCTAssertNotNil((commonPayload["device_id"] as? String).map(UUID.init))
+
         let platformInfo = commonPayload["platform_info"] as? [String: Any]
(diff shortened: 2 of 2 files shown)


StripeCore/StripeCore/Source/Analytics/AnalyticsClientV2.swift
```diff
@@ -134,6 +134,9 @@ extension AnalyticsClientV2Protocol {
             "install": InstallMethod.current.rawValue,
             "app_bundle_id": Bundle.stp_applicationBundleId() ?? "",
         ]
+        if let deviceId = UIDevice.current.identifierForVendor?.uuidString {
```
mats-stripe, line 137: Is there a mapping somewhere of UUIDs to device model names? I'm familiar with [this list of device identifiers](https://gist.github.com/adamawolf/3048717), but it wouldn't be the same as `UIDevice.current.identifierForVendor?.uuidString`

mats-stripe (approved):
Nice! LGTM, just one question

StripeCore/StripeCore/Source/Analytics/AnalyticsClientV2.swift
nschris-stripe: This is a unique identifier for the specific device/app install, so it does not map to any device types. This means that if one device had multiple apps with the SDK integrated then the identifier will be different (unless the apps were distributed by the same developer).

aywang-stripe (approved):

StripeCore/StripeCore/Source/Analytics/AnalyticsClientV2.swift
mludowise-stripe: The Vendor ID is unrelated to the device model and is a per-device x per-app hash. When the user installs the app, the system vends a unique UUID identifier:
https://developer.apple.com/documentation/uikit/uidevice/1620059-identifierforvendor/

If you want to know the device model, then I think you probably want to use `device_type` which is populated by this:
https://github.com/stripe/stripe-ios/blob/1494f048/StripeCore/StripeCore/Source/Helpers/STPDeviceUtils.swift

PR #4096, opened 2024-10-03
Embedded mandates

porter-stripe:
## Summary
- Displays mandates in embedded for no form LPMs, like Cash App Pay, and saved payment methods like US bank acct.
- Consolidates mandate building between vertical and embedded into a single class that is now unit tested
- Notifies the delegate when the height changes, updates the playground to implement the delegate

Future PR will notify the delegate when a selection is made.

Demo:

### Without saved bank

https://github.com/user-attachments/assets/6aaa7179-469d-4675-b166-e5ca1b98527f

### With saved bank

https://github.com/user-attachments/assets/fdba1595-f58d-43dd-8be3-5dfcd79498c2

## Motivation
- Embedded

## Testing
- Manual
- Snapshot tests
- Unit tests

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/SimpleMandateContainerView.swift
```diff
@@ -0,0 +1,42 @@
...
+//  SimpleMandateContainerView.swift
+//  StripePaymentSheet
+//
+//  Created by Nick Porter on 10/3/24.
+//
+
+import Foundation
+import UIKit
+
+class SimpleMandateContainerView: UIView {
```
porter-stripe, line 11: I tried using a UILayoutGuide but could just not get it working... I think it's b/c the mandate is contain within a stack view.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
```diff
@@ -98,9 +98,18 @@ public class EmbeddedPaymentElement {
...
-            savedPaymentMethodAccessoryType: savedPaymentMethodAccessoryType
+            savedPaymentMethodAccessoryType: savedPaymentMethodAccessoryType,
+            mandateProvider: MandateProvider(configuration: configuration,
+                                             elementsSession: loadResult.elementsSession,
+                                             intent: .deferredIntent(intentConfig: intentConfiguration))
         )
-        return .init(view: embeddedPaymentMethodsView, configuration: configuration)
+
+        let embeddedPaymentElement: EmbeddedPaymentElement = .init(view: embeddedPaymentMethodsView, configuration: configuration)
+        await MainActor.run {
```
porter-stripe, line 108: Maybe this public API should be: `@MainActor public static func create(...`

yuki-stripe: I think once we're interacting with UIKit it should be on the main thread. Not sure if the whole func should, wdyt?  

AFAIU we still need to use `await` no matter what when we call async methods

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedMandateContainerView.swift
```diff
@@ -0,0 +1,47 @@
...
+//  StripePaymentSheet
+//
+//  Created by Nick Porter on 10/3/24.
+//
+
+import Foundation
+import UIKit
+
+/// A simple container that holds a mandate text view for the embedded payment element
+class EmbeddedMandateContainerView: UIView {
```
yuki-stripe, line 12: ooc any reason to not put `SimpleMandateTextView` directly in `EmbeddedPaymentElement`?  Maybe we can add the reason to the docstring

yuki-stripe: I think it's to encapsulate the constraint code, b/c we're putting it in a stackview and have to do some things to make it look right.

I'm not convinced that it should be in the stackview though.  Like, should it respect the stackview's `spacing`?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/MandateTextProvider.swift
```diff
@@ -0,0 +1,101 @@
...
+//  StripePaymentSheet
+//
+//  Created by Nick Porter on 10/3/24.
+//
+
+import Foundation
+@_spi(STP) import StripeCore
+@_spi(STP) import StripeUICore
+
+protocol MandateTextProvider {
```
yuki-stripe, line 12: Why the protocol?  Can we delete this and rename FormMandateProvider to `MandateTextProvider`?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/MandateTextProvider.swift
```diff
@@ -0,0 +1,101 @@
...
+    init(configuration: PaymentElementConfiguration, elementsSession: STPElementsSession, intent: Intent) {
+        self.configuration = configuration
+        self.elementsSession = elementsSession
+        self.intent = intent
+    }
+
+    /// Builds the attributed string for a given payment method type
+    /// - Parameter paymentMethodType: The payment method type who's mandate should be constructed
+    /// - Parameter savedPaymentMethod: The currently selected saved payment method if any
+    /// - Parameter bottomNoticeAttributedString: The
```
yuki-stripe, line 31: nit this is missing

yuki-stripe: "Passing this in just makes this method return it as long as `configuration` doesn't hide mandate text"

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/MandateTextProvider.swift
```diff
@@ -0,0 +1,101 @@
...
+                    type: paymentMethodType,
+                    intent: intent,
+                    elementsSession: elementsSession,
+                    previousCustomerInput: nil,
+                    formCache: .init(),
+                    configuration: configuration,
+                    headerView: nil,
+                    analyticsHelper: .init(isCustom: false, configuration: PaymentSheet.Configuration()), // Dummy, not used
+                    delegate: self
+                ).form
```
yuki-stripe, line 65: We should probably just use `FormFactory` directly and get rid of the `PaymentMethodFormViewControllerDelegate` methods.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactoryConfig.swift
```diff
@@ -8,7 +8,7 @@ import Foundation
 @_spi(STP) import StripePayments
 
 enum PaymentSheetFormFactoryConfig {
-    case paymentSheet(PaymentSheet.Configuration)
+    case paymentSheet(PaymentElementConfiguration) // TODO(porter) Change this back to PaymentSheet.Configuration when implementing embedded showing forms, and add a .embedded(EmbeddedPaymentElement.Configuration) case.
```
yuki-stripe, line 11: Interesting, ooc why a new case? Maybe this case is just renamed to `paymentElement`?  Is the form supposed to be different between embedded and PS?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
```diff
@@ -25,17 +30,25 @@ import UIKit
         return stackView
     }()
 
+    private lazy var mandateView = {
+        EmbeddedMandateContainerView(appearance: appearance)
+    }()
```
yuki-stripe, line 35: Apparently we can do this now:

```
    private lazy var mandateView = EmbeddedMandateContainerView(appearance: appearance)
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
```diff
@@ -139,11 +154,31 @@ import UIKit
...
         self.selection = selection
+        updateMandate()
     }
 
     func didTapAccessoryButton() {
         // TODO(porter)
     }
+
+    // MARK: Mandate handling
+    private func updateMandate(updateHeight: Bool = true) {
```
yuki-stripe, line 165: We'll want this to animate, too, outside of the init.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
```diff
@@ -139,11 +154,31 @@ import UIKit
...
+        guard updateHeight else { return }
+
+        // Force the layout to update so we get the new height
+        setNeedsLayout()
+        layoutIfNeeded()
+
+        let newHeight = frame.size.height
+        if previousHeight != newHeight {
+            delegate?.heightDidChange()
+        }
```
yuki-stripe, line 180: I wonder if there's a way we could do this more generically. Like override `layoutSubviews` and check there.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/FormMandateProviderTests.swift
```diff
@@ -0,0 +1,180 @@
...
+    override func setUp() async throws {
+        await withCheckedContinuation { continuation in
+            Task {
+                AddressSpecProvider.shared.loadAddressSpecs {
+                    FormSpecProvider.shared.load { _ in
+                        continuation.resume()
+                    }
+                }
+            }
+        }
```
yuki-stripe, line 25:
```suggestion
        await PaymentSheetLoader.loadMiscellaneousSingletons()
```

😌

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentMethodsViewSnapshotTests.swift
```diff
@@ -541,12 +572,223 @@ class EmbeddedPaymentMethodsViewSnapshotTests: STPSnapshotTestCase {
...
+            shouldShowLink: true,
+            savedPaymentMethodAccessoryType: .none,
+            mandateProvider: mockMandateProvider
+        )
+
+        verify(embeddedView)
+    }
+
+    // MARK: EmbeddedPaymentMethodsViewDelegate test
+    func testEmbeddedPaymentMethodsView_heightDidChange() {
```
yuki-stripe, line 736: Let's move this to its own test file (not a snapshot test)

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/FormMandateProviderTests.swift
```diff
@@ -0,0 +1,180 @@
...
+//  Created by Nick Porter on 10/4/24.
+//
+
+import Foundation
+@testable @_spi(STP) import StripeCore
+@testable @_spi(EmbeddedPaymentElementPrivateBeta) @_spi(STP) import StripePaymentSheet
+@testable @_spi(STP) import StripeUICore
+import XCTest
+
+class FormMandateProviderTests: XCTestCase {
```
yuki-stripe, line 14: 👏

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/MandateTextProvider.swift
yuki-stripe: Ah okay the test was hidden by default, I see how we use it.  IMO I'd rather have the snapshot tests simply build the actual Provider and use it directly; that way those tests can also exercise the code and they're more realistic.  I'm not sure what the benefit of the protocol is; it's not like this is making network calls that we don't want to deal with.  But I don't feel strongly and am not asking you to re-write all that code!

yuki-stripe (commented):
NICE!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
porter-stripe: Ah good idea, I overrode `layoutSubviews`, only caveat is I needed an instance variable to keep track of the previous height.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
porter-stripe: ah nice!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/MandateTextProvider.swift
porter-stripe: good idea!

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/FormMandateProviderTests.swift
porter-stripe: TIL

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
porter-stripe: Threw this inside an animation block, I think it looks pretty good. 

https://github.com/user-attachments/assets/6d78c406-f3c3-415b-aa98-596e8bf98916

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/MandateTextProvider.swift
porter-stripe: Yep just for testing. I like this kind of pattern! Some of our types (`STPElementsSession` and `STPPaymentMethod`) are quite cumbersome to instantiate just how you want to test a specific scenario. I like that I can focus on testing the embedded view with specific scenarios, then write tests for the provider in `FormMandateProviderTests`.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedMandateContainerView.swift
porter-stripe: Yeah it pretty much just encapsulates the constraint code so we get proper padding. I updated the doc comment a bit.

If we directly add the mandate view to the `EmbeddedPaymentMethodsView`, when we hide it we need to do some manual calculation to get the height of the `EmbeddedPaymentMethodsView` correct (at least from what I could see in my testing). But when it's inside the stack view, the stack view auto resizes itself when we hide the mandate which makes it easy to get the height of the EmbeddedPaymentMethodsView.

re: Spacing good callout, I set custom spacing before mandate to be 0.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/MandateTextProvider.swift
yuki-stripe: I'd rather solve this:
> Some of our types (STPElementsSession and STPPaymentMethod) are quite cumbersome to instantiate just how you want to test a specific scenario.

by making it easier to instantiate vs. creating the protocol.  

We will need to create STPElementsSession and the IntentConfiguration to initialize EmbeddedPaymentElement anyways - it will need those things - so I don't think it's even buying us anything in the long term by mocking out MandateTextProvider.

yuki-stripe: I think thinking about it more, I feel we aren't going to like having the mock in the long term. It will be strange to initialize EmbeddedPaymentElement w/ STPElementSession + IntentConfiguration _and_ a MandateProvider which uses those same things?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedMandateContainerView.swift
yuki-stripe: I think what you're describing is the fact that a stackview will automatically remove a hidden arranged subview / make its height zero. That's a nice benefit of it being in a stackview for sure.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/MandateTextProvider.swift
yuki-stripe: Talked offline - actually we probably won't initialize EPE with that stuff!

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentMethodsViewTests.swift
```diff
@@ -0,0 +1,84 @@
...
+//
+
+@_spi(EmbeddedPaymentElementPrivateBeta) @testable import StripePaymentSheet
+import XCTest
+
+final class EmbeddedPaymentMethodsViewTests: XCTestCase {
+
+    // MARK: EmbeddedPaymentMethodsViewDelegate.heightDidChange test
+
+    func testEmbeddedPaymentMethodsView_heightDidChange() {
```
yuki-stripe, line 15: great test btw

yuki-stripe (approved):

PR #4097, opened 2024-10-04
Fixes `testWebInstantDebitsFlow()`

mats-stripe:
## Summary

This fixes the UI test `testWebInstantDebitsFlow()` which has been failing on master due to a server-side string change in the consent pane title. 

Copy before:

```
<brand> uses Link to connect your account
```

Now it says:

```
Pay with your bank using Link
```

I kept the expected value pretty vague (`Link`) since this string might change again in the future.

## Motivation

Fix CI

## Testing

Test passes

## Changelog

N/a

### DIFF
--- a/Example/FinancialConnections Example/FinancialConnectionsUITests/FinancialConnectionsUITests.swift
+++ b/Example/FinancialConnections Example/FinancialConnectionsUITests/FinancialConnectionsUITests.swift
@@ -396,11 +396,12 @@ final class FinancialConnectionsUITests: XCTestCase {
         app.fc_playgroundCell.tap()
         app.fc_playgroundShowAuthFlowButton.tap()
 
-        let usesLinkText = app.webViews
-            .staticTexts
-            .containing(NSPredicate(format: "label CONTAINS 'Link'"))
+        let authFlowWebViewUrl = app
+            .otherElements["TopBrowserBar"]
+            .otherElements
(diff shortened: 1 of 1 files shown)


Example/FinancialConnections Example/FinancialConnectionsUITests/FinancialConnectionsUITests.swift
```diff
@@ -398,7 +398,7 @@ final class FinancialConnectionsUITests: XCTestCase {
 
         let usesLinkText = app.webViews
             .staticTexts
-            .containing(NSPredicate(format: "label CONTAINS 'uses Link to connect your account'"))
+            .containing(NSPredicate(format: "label CONTAINS 'Link'"))
```
tillh-stripe, line 401: Can we use a different label on this screen that isn’t affected by the backend changes, or just generally wait for the page to be loaded?

tillh-stripe, line 401: Can we use a different label on this screen that isn’t affected by the backend changes, or just generally wait for the page to be loaded?

mats-stripe: Good point - updated to look for the presence of `auth.stripe.com` in the top browser bar: 

image

PR #4100, opened 2024-10-04
Vertical mode preserves previous list selection when going back to li…

yuki-stripe:
…st from form

Also threw in an unrelated fix for the embedded example

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-2521

## Testing
Existing tests. Manually tested.

## Changelog
Not user facing enough


porter-stripe (dismissed):
🚢

porter-stripe (approved):

PR #4101, opened 2024-10-07
Add `CustomerSession` analytics for `CustomerSheet`

samer-stripe:
## Summary
Add `CustomerSession` analytics for `CustomerSheet`

## Motivation
Track what integrations merchants are initializing `CustomerSheet` with.


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSheet.swift
```diff
@@ -212,6 +221,15 @@ public class CustomerSheet {
         }
         return nil
     }
+    
+    private func createInitEvent() -> STPAnalyticEvent {
```
mats-stripe, line 225: Super minor (not blocking): This could be a computed property instead of a function:
```suggestion
    private var initEvent: STPAnalyticEvent {
```

mats-stripe (dismissed):

PR #4115, opened 2024-10-09
[Connect] Display errors in alert in example app

mludowise-stripe:
## Summary
- If a `didFailLoadWithError` event occurs, displays the error in an alert
- When failing to retrieve the client secret, displays error description in the alert instead of generic "Failed to retrieve client secret." message
- Fixes bug where error alert wouldn't display if component was modally displayed

## Motivation
https://jira.corp.stripe.com/browse/MXMOBILE-2836

## Testing

Displays server error when attempting to test with a non-custom account type:

Note: I was unable to trigger a `didFailLoadWithError` error to test

### DIFF
--- a/Example/StripeConnectExample/StripeConnectExample/MainViewController.swift
+++ b/Example/StripeConnectExample/StripeConnectExample/MainViewController.swift
@@ -246,7 +246,7 @@ class MainViewController: UITableViewController {
 
 extension MainViewController: AccountOnboardingViewControllerDelegate {
     func accountOnboarding(_ accountOnboarding: AccountOnboardingViewController, didFailLoadWithError error: any Error) {
-        presentAlert(title: "Error loading account onboarding", message: "\((error as NSError).debugDescription)")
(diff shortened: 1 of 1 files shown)


Example/StripeConnectExample/StripeConnectExample/MainViewController.swift
```diff
@@ -198,7 +202,7 @@ class MainViewController: UITableViewController {
     func presentAlert(title: String, message: String) {
         let alert = UIAlertController(title: title, message: message, preferredStyle: .alert)
         alert.addAction(UIAlertAction(title: "OK", style: .default))
-        self.present(alert, animated: true)
+        (presentedViewController ?? self).present(alert, animated: true)
```
mludowise-stripe, line 205: When this VC is already presenting a component VC modally, the system won't let us present another modal. So instead, check if we're already presenting something modally and stack on top of that, otherwise present on self.

kush-stripe (approved):

PR #4117, opened 2024-10-09
Misc playground improvements

yuki-stripe:
## Summary
Bunch of playground improvements:

- Combine amount & currency for some space savings
- Only show "Supported Payment Methods" text field when APMs is off
- Only show "Layout" when UI != embedded
- Hide "Client-side confirmation" from type picker when UI == embedded
- Add animation when settings are added/removed

https://github.com/user-attachments/assets/6d9926e0-a728-40ff-b51a-30353545e3c5


porter-stripe (approved):
nice!

porter-stripe (dismissed):

porter-stripe (approved):

PR #4118, opened 2024-10-09
[BANKCON-11801] Pass amount, currency, and intent ID to Link signup endpoint

mats-stripe:
## Summary

Using the new `ElementsSessionContext`, we can now pass `amount`, `currency`, and `intentId` from payment elements to the Instant Debits Link signup endpoint. This will be used for Link incentives.

## Motivation

[BANKCON-11801](https://jira.corp.stripe.com/browse/BANKCON-11801)

## Testing

Verified API parameters correctly make their way from the payment sheet to the signup endpoint: 

Screenshot 2024-10-09 at 11 08 00 AM

## Changelog

N/a

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
@@ -217,7 +217,10 @@ protocol FinancialConnectionsAPI {
     func linkAccountSignUp(
         emailAddress: String,
         phoneNumber: String,
-        country: String
+        country: String,
+        amount: Int?,
+        currency: String?,
+        intentId: ElementsSessionContext.IntentID?
     ) -> Future<LinkSignUpResponse>
 
     func attachLinkConsumerToLinkAccountSession(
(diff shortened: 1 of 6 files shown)


StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
```diff
@@ -915,6 +921,27 @@ extension FinancialConnectionsAPIClient: FinancialConnectionsAPI {
             "locale": Locale.current.toLanguageTag(),
             "consent_action": "entered_phone_number_clicked_save_to_link",
         ]
+
+        if let amount, let currency {
```
porter-stripe, line 925: Do we test somehow that we are sending the right params?

mats-stripe: I'll try think of how we could test this. It'll be a little tricky to test that the values from the `ElementsSessionContext` passed in the `FinancialConnectionsSheet` will be used in the signup API params. 

Couple reasons why: the Financial Connections SDK could use more dependency injection (in this case, we would want to inject an instance of `FinancialConnectionsAPI` to the sheet), and there's also an issue that most of the FC types aren't `public` (or even `@_SPI(STP) public`), which makes DI here tough to implement.

There's definitely a broader action item here, which is to make the FC SDK more testable. I'll track that here [BANKCON-15481](https://jira.corp.stripe.com/browse/BANKCON-15481) so that it doesn't block this work.

porter-stripe (approved):

PR #4124, opened 2024-10-10
Card Brand Filtering in PAN field and saved PMs

porter-stripe:
## Summary
- Adds the CBF public API to PaymentSheet and CustomerSheet under SPI
- Adds unit and UI tests
- Updates playground with a CBF toggle
- Filters cards in the card text field based on brand
- Hides saved payment methods of cards with disallowed brands

In a future PR:
- Filter card networks in Apple Pay
- Filter card brands in CBC dropdown

### Blocking AmEx then paying with Visa

https://github.com/user-attachments/assets/ccafb592-337a-48cd-af12-729705d455da

## Motivation
https://docs.google.com/document/d/1DD7mncHsHbMtU_jqr_TnaLL8ltjO61pPmdOAlJXxYHY/edit

## Testing
- Manual
- Unit tests
- UI tests

## Changelog
N/A for now

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CardBrandFilter.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CardBrandFilter.swift
@@ -1,0 +1,82 @@
+//
+//  CardBrandFilter.swift
+//  StripePaymentSheet
+//
+//  Created by Nick Porter on 10/10/24.
+//
+
+import Foundation
+@_spi(STP) import StripeCore
+
+struct CardBrandFilter {
+
+    static let `default`: CardBrandFilter = .init(cardBrandAcceptance: .all)
+
+    private let cardBrandAcceptance: PaymentSheet.CardBrandAcceptance
+
+    init(cardBrandAcceptance: PaymentSheet.CardBrandAcceptance) {
+        self.cardBrandAcceptance = cardBrandAcceptance
+    }
+
+    /// Determines if a merchant can accept a card brand based on `cardBrandAcceptance`
+    /// - Parameter cardBrand: The `STPCardBrand` to determine if acceptance
+    /// - Returns: Returns true if this merchant can accept this card brand, false otherwise
+    func isAccepted(cardBrand: STPCardBrand) -> Bool {
+        switch cardBrandAcceptance {
+        case .all:
+            return true
+        case .allowed(let allowedCardBrands):
+            // If a merchant has specified a list of brands to allow, block unknown brands
+            guard let brandCategory = cardBrand.asBrandCategory else {
+                STPAnalyticsClient.sharedClient.logPaymentSheetEvent(event: .paymentSheetDisallowedCardBrand,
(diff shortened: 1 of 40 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/TextField/TextFieldElement+Card.swift
```diff
@@ -93,6 +108,9 @@ extension TextFieldElement {
                     return String.Localized.your_card_number_is_incomplete
                 case .invalidBrand, .invalidLuhn:
                     return String.Localized.your_card_number_is_invalid
+                case .disallowedBrand(let brand):
+                    let cardBrandDisplayName =  STPCardBrandUtilities.stringFrom(brand) ?? "This card brand"
```
davidme-stripe, line 112: I don't think this would localize cleanly — could we have a second string with the full "The selected brand is not allowed" message?

porter-stripe: Ah you're probably right! I updated this to use a hard coded string for English, and a generic one for non-english that will be localized.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CardBrandFilter.swift
```diff
@@ -0,0 +1,82 @@
...
+    /// Determines if a merchant can accept a card brand based on `cardBrandAcceptance`
+    /// - Parameter cardBrand: The `STPCardBrand` to determine if acceptance
+    /// - Returns: Returns true if this merchant can accept this card brand, false otherwise
+    func isAccepted(cardBrand: STPCardBrand) -> Bool {
+        switch cardBrandAcceptance {
+        case .all:
+            return true
+        case .allowed(let allowedCardBrands):
+            // If a merchant has specified a list of brands to allow, block unknown brands
+            guard let brandCategory = cardBrand.asBrandCategory else {
```
davidme-stripe, line 30: This looks like it will log many times: Continuously as the user types a card number, and for each enumeration in the `STPCardBrand.allCases.filter { cardFilter.isAccepted(cardBrand: $0) }` code below. Maybe we could have it log less often, maybe just from the PaymentSheet UI somehow when the card number entry is completed?

porter-stripe: You're right! That's gonna be super spammy! I changed this to only log from the PAN element, and not log if it's the same disallowed brand.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/TextField/TextFieldElement+Card.swift
porter-stripe: Oh got it, fixed!

davidme-stripe (approved):

PR #4133, opened 2024-10-14
Show forms in LPM only mode for vertical mode

porter-stripe:
## Summary
- If we only have one LPM, we directly show the LPM form if it collects user input
- If the LPM does not collect input we just show the standard list with it selected
- If there are wallet options, we show the wallet buttons with the form (if it collects user input)
- If there are wallet options, we do not auto select the lone LPM

## Motivation
https://jira.corp.stripe.com/browse/RUN_MOBILESDK-3605

## Testing
- Manual
- Updated/new snapshots

## Changelog
N/A (private beta)


Tests/ReferenceImages_64/StripePaymentSheetTests.PaymentSheetVerticalViewControllerSnapshotTest/testDisplaysList_one_non_card_pm_no_input@3x.png
porter-stripe, line 1: Should we auto select Cash App Pay here? I'm leaning towards yes.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetVerticalViewController.swift
```diff
@@ -86,9 +86,9 @@ class PaymentSheetVerticalViewController: UIViewController, FlowControllerViewCo
     weak var paymentSheetDelegate: PaymentSheetViewControllerDelegate?
     let shouldShowApplePayInList: Bool
     let shouldShowLinkInList: Bool
-    /// Whether or not we are in the special case where we don't show the list and show the card form directly
-    var shouldDisplayCardFormOnly: Bool {
-        return paymentMethodTypes.count == 1 && paymentMethodTypes[0] == .stripe(.card)
+    /// Whether or not we are in the special case where we don't show the list and show the form directly
+    var shouldDisplayFormOnly: Bool {
```
yuki-stripe, line 90: This isn't accurate, is it?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetVerticalViewController.swift
```diff
@@ -322,7 +328,15 @@ class PaymentSheetVerticalViewController: UIViewController, FlowControllerViewCo
                 case .link:
                     return isFlowController ? .link : nil // Only default to Link in flow controller mode
                 case .stripeId, nil:
-                    return savedPaymentMethods.first.map { .saved(paymentMethod: $0) }
+                    if let savedSelection = savedPaymentMethods.first {
+                        return .saved(paymentMethod: savedSelection)
+                    }
+                    // If we only have one PM in the list with no wallet options or saved payment methods then select it
+                    if shouldDisplayFormOnly, makeWalletHeaderView() == nil, let paymentMethodType = loadResult.paymentMethodTypes.first {
```
yuki-stripe, line 335: It's confusing that we would be in this method when `shouldDisplayFormOnly` is true; that implies we display the form only, not the list.

yuki-stripe (changes requested):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetVerticalViewController.swift
porter-stripe: Hrm, I guess not in the case there is no form to present when we don't collect user input. I've updated this flag to also do the checking for if the form requires user input.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetVerticalViewController.swift
porter-stripe: Yeah this usage was a bit overloaded, I've re-written this `if` to check for other more relevant flags to this code path.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetVerticalViewController.swift
```diff
@@ -174,35 +175,29 @@ class PaymentSheetVerticalViewController: UIViewController, FlowControllerViewCo
         if let paymentMethodFormViewController {
             remove(childViewController: paymentMethodFormViewController)
         }
-
-        // If we'd only show one PM in the vertical list, and it collects user input, display the form instead of the payment method list.
         if shouldDisplayFormOnly, let paymentMethodType = loadResult.paymentMethodTypes.first {
+            // If we'd only show one PM in the vertical list, display the form instead of the payment method list.
```
yuki-stripe, line 179: I think the old comment was better!

>         // If we'd only show one PM in the vertical list, and it collects user input, display the form instead of the payment method list.

yuki-stripe (approved):

yuki-stripe (approved):

PR #4136, opened 2024-10-14
Fix bug where product_usage is not being sent with FC analytics

wooj-stripe:
## Summary
PaymentSheetLoader.load() was being called prior to setting the product usage for FlowController use cases.  This PR changes it so that we set product usage before any analytics calls are made.

## Motivation
Initial metrics for FC are being sent out without product_usage being sent -- specifically the first time FC loads.

## Testing
Manual testing


yuki-stripe (approved):
ooh nice!

PR #4138, opened 2024-10-15
[Connect] Add error handling, default disable swipe-to-dismiss, and refactor web views into view controllers

mludowise-stripe:
## Summary

1. **Calls the `didFailLoadWithError` when the web page fails to load**
   If the web page is not successfully loaded, the `didFailLoadWithError` delegates on the component VCs are called. In this case, the web page would never load far enough to receive an Embedded Component error from JS, but we still want to notify the app that the component could not be loaded.

   This could happen in the following scenarios:
   - The web page fails to load ([didFailNavigation](https://developer.apple.com/documentation/webkit/wknavigationdelegate/1455637-webview))
   - The web page receives a 4xx or 5xx status code

2. **Default disables swipe to dismiss**
    Sets `isModalInPresentation = true` on the component view controller to prevent users from swiping to dismiss the view in case their intent was to swipe to dismiss a bottom sheet being displayed inside the web view.

3. **Refactors `ConnectWebView` and `ConnectComponentWebView` into view controllers**
   Advantages to this are:
   - Removes the need to set a `presentPopup` or `didClose` closure from each view controller containing a web view and instead directly call the VC's `present` and `dismiss` methods
   - Pre-work needed for analytics ([MXMOBILE-2491](https://jira.corp.stripe.com/browse/MXMOBILE-2491)) so we can log when the view is displayed to the user in `viewDidAppear`
   - Makes it easier to DRY up setting `isModalInPresentation = true` by doing this in the ConnectWebViewController's `didMove` method

## Motivation
https://jira.corp.stripe.com/browse/MXMOBILE-2842
https://jira.corp.stripe.com/browse/MXMOBILE-2837

## Testing
- [x] Unit test coverage
- [x] Manually tested swipe-to-dismiss (see video)
- [x] Manually retested popups and downloads to ensure presenting view controllers still worked (see video)
- [x] Manually tested hitting a 404 caused an HTTPStatusError by temporarily changing `connectJSBaseURL` to `"https://connect-js.stripe.com/v1.0/notapage.html"` (see video)

### Swipe to dismiss

https://github.com/user-attachments/assets/6114ae0a-a1bb-4923-b2a0-5aafb1afbeef

### Retesting popups

https://github.com/user-attachments/assets/04240096-f101-49cc-83bc-babc984ce5af

### HTTP Status error

https://github.com/user-attachments/assets/fbfce3ed-74c2-42f4-abb3-f6d13c3dea01


StripeConnect/StripeConnect/Source/Components/AccountManagementViewController.swift
```diff
@@ -22,39 +22,31 @@ public class AccountManagementViewController: UIViewController {
...
-            componentType: .accountManagement
+            componentType: .accountManagement,
+            loadContent: loadContent
         ) {
             Props(collectionOptions: collectionOptions)
-        }
-        super.init(nibName: nil, bundle: nil)
-
-        webView.addMessageHandler(OnLoadErrorMessageHandler { [weak self] value in
+        } didFailLoadWithError: { [weak self] error in
```
mludowise-stripe, line 39: Made the `didFailLoadWithError` a required property passed on `init`. This now gets used in more than just the `OnLoadErrorMessageHandler` but also when the web page fails to load.

By making it an argument passed on init, it ensures we never forget to set the handler. It does force us to make `webVC` an implicitly unwrapped optional so that we can reference `self` when initializing the webVC, but personally I think it's worth the tradeoff and better than forgetting to set the `didFail` handler.

StripeConnect/StripeConnect/Source/Internal/Webview/ConnectComponentWebViewController.swift
```diff
@@ -82,44 +87,68 @@ class ConnectComponentWebView: ConnectWebView {
...
+                  loadContent: loadContent,
                   fetchInitProps: VoidPayload.init,
+                  didFailLoadWithError: didFailLoadWithError,
                   notificationCenter: notificationCenter,
-                  webLocale: webLocale,
-                  loadContent: loadContent)
-    }
-    func updateAppearance(appearance: Appearance) {
-        sendMessage(UpdateConnectInstanceSender.init(payload: .init(locale: webLocale.webIdentifier, appearance: .init(appearance: appearance, traitCollection: traitCollection))))
-        updateColors(appearance: appearance)
```
mludowise-stripe, line 105: Moved to "Internal" extension

StripeConnect/StripeConnect/Source/Internal/Webview/HTTPStatusError.swift
```diff
@@ -0,0 +1,28 @@
...
+
+import Foundation
+@_spi(STP) import StripeCore
+
+/// Error passed to the when the `ConnectComponentWebViewController.didFailLoadWithError`
+/// when receiving an error status code loading the component web page
+struct HTTPStatusError: Error, CustomNSError {
+    /// The HTTP status code
+    let errorCode: Int
+}
```
mludowise-stripe, line 15: Note: This is not public but it can potentially get passed to the integrator in `didFailLoadWithError`. I considered putting this through API review, but ultimately decided it was unnecessary since there's no value in the integrator explicitly checking for this error code. 

The error will have:
- Domain = `StripeConnect.HTTPStatusError`
- Code = the http status code
- Debug description = `Component loaded with HTTP status {errorCode}`
- Localized description = `There was an unexpected error -- try again in a few seconds`

StripeConnect/StripeConnectTests/Internal/Webview/ConnectComponentWebViewControllerTests.swift
```diff
@@ -0,0 +1,259 @@
+//
+//  ConnectComponentWebViewControllerTests.swift
```
mludowise-stripe, line 2: This was renamed from `ConnectComponentWebViewTests` but I added enough new tests to it that GH no longer recognized the move (highlighted in comments below)

Or you can diff these locally with this:
```sh
git fetch mludowise/refactor_to_vc
git difftool origin/master:StripeConnect/StripeConnectTests/Internal/Webview/ConnectComponentWebViewTests.swift origin/mludowise/refactor_to_vc:StripeConnect/StripeConnectTests/Internal/Webview/ConnectComponentWebViewControllerTests.swift
```

StripeConnect/StripeConnect/Source/Internal/Webview/ConnectComponentWebViewController.swift
```diff
@@ -40,17 +43,19 @@ class ConnectComponentWebView: ConnectWebView {
     init<InitProps: Encodable>(
         componentManager: EmbeddedComponentManager,
         componentType: ComponentType,
+        loadContent: Bool,
```
mludowise-stripe, line 46: I made `loadContent` a required param to ensure we always explicitly pass the value from the Component manager (I wasn't doing this for AccountManagement, PaymentDetails, or NotificationBanner which meant it was loading in tests)

StripeConnect/StripeConnectTests/Internal/Webview/ConnectComponentWebViewControllerTests.swift
```diff
@@ -0,0 +1,259 @@
...
+private extension ConnectComponentWebViewControllerTests {
+    func componentManagerAssertingOnFetch(appearance: Appearance = .default, fonts: [EmbeddedComponentManager.CustomFontSource] = []) -> EmbeddedComponentManager {
+        EmbeddedComponentManager(apiClient: .init(publishableKey: "test"),
+                                 appearance: appearance,
+                                 fonts: fonts,
+                                 fetchClientSecret: {
+            XCTFail("Client secret should not be retrieved in this test")
+            return ""
+        })
+    }
```
mludowise-stripe, line 258: Moved `componentManagerAssertingOnFetch` into an extension since it was sandwiched between multiple test functions

StripeConnect/StripeConnectTests/Internal/Webview/ConnectComponentWebViewControllerTests.swift
```diff
@@ -0,0 +1,259 @@
...
+
+        try await webVC.webView.evaluateMessageWithReply(name: "fetchInitParams",
+                                                         json: "{}",
+                                                         expectedResponse: """
+                                                            {"appearance":{"variables":{"fontFamily":"-apple-system","fontSizeBase":"16px"}},"fonts":[{"family":".AppleSystemUIFont","src":"url(data:font\\/txt;charset=utf-8;base64,dGVzdAo=)","weight":"400"}],"locale":"fr-FR"}
+                                                            """)
+    }
+
+    @MainActor
+    func testLoaderStartShowsLoadingIndicator() async throws {
```
mludowise-stripe, line 171: 👀 Tests from here down are all new test coverage

StripeConnect/StripeConnectTests/Internal/Webview/ConnectWebViewControllerTests.swift
```diff
@@ -132,28 +131,17 @@ class ConnectWebViewTests: XCTestCase {
...
-    /// `window.close()` should close the view
-    func testWindowCloseDismissesView() {
-        var didCloseCalled = false
-        webView.didClose = { wv in
-            XCTAssertEqual(wv, self.webView)
-            didCloseCalled = true
-        }
-        webView.uiDelegate?.webViewDidClose?(webView)
-        XCTAssertTrue(didCloseCalled)
-    }
```
mludowise-stripe, line 155: No value in testing this since the subclass just needs to override this now

StripeConnect/StripeConnectTests/WebView+Tests.swift
```diff
@@ -28,8 +28,8 @@ extension WKWebView {
                         """)
     }
 
-    func evaluateOnLoaderStart(elementTagName: String) {
-        evaluateMessage(name: "onSetterFunctionCalled",
+    func evaluateOnLoaderStart(elementTagName: String) async throws {
+        try await evaluateMessage(name: "onSetterFunctionCalled",
```
mludowise-stripe, line 32: Had to change this to an async method so we could wait for it to finish in `testJSOnLoadError`

StripeConnect/StripeConnect/Source/Components/AccountManagementViewController.swift
```diff
@@ -22,39 +22,31 @@ public class AccountManagementViewController: UIViewController {
...
             guard let self else { return }
-            self.delegate?.accountManagement(self, didFailLoadWithError: value.error.connectEmbedError)
-        })
-
-        // TODO(MXMOBILE-2796): Send collection options to web view
```
mludowise-stripe, line 44: This is being done in the init params, just forgot to delete the TODO

StripeConnect/StripeConnect/Source/Internal/Webview/ConnectComponentWebViewController.swift
```diff
@@ -82,44 +87,68 @@ class ConnectComponentWebView: ConnectWebView {
...
+                  loadContent: loadContent,
                   fetchInitProps: VoidPayload.init,
+                  didFailLoadWithError: didFailLoadWithError,
                   notificationCenter: notificationCenter,
-                  webLocale: webLocale,
-                  loadContent: loadContent)
-    }
-    func updateAppearance(appearance: Appearance) {
-        sendMessage(UpdateConnectInstanceSender.init(payload: .init(locale: webLocale.toLanguageTag(), appearance: .init(appearance: appearance, traitCollection: traitCollection))))
-        updateColors(appearance: appearance)
```
mludowise-stripe, line 105: Moved to "Internal" extension

StripeConnect/StripeConnect/Source/Internal/Webview/HTTPStatusError.swift
```diff
@@ -0,0 +1,27 @@
...
+
+extension HTTPStatusError: LocalizedError {
+    var errorDescription: String? {
+        NSError.stp_unexpectedErrorMessage()
+    }
+}
```
mludowise-stripe, line 21: Is there any value in having a localized error message? The only thing I can think of is that if the app decides displays the error message to the user that it's at least localized and not super technical. 

If we decide to keep this then I think I should also add it to `EmbeddedComponentError` since its debug description isn't localized.

mludowise-stripe:
> 
> ```diff
> ConnectComponentWebViewController.swift:20 warning: Property 'componentType' is assigned, but never used
> ConnectWebViewController.swift:108 warning: Function 'showErrorAlert(for:)' is unused
> ConnectWebViewController.swift:119 warning: Function 'cleanupDownloadedFile()' is unused
> ConnectWebViewController.swift:248 warning: Function 'download(decideDestinationUsing:suggestedFilename:)' is unused
> ConnectWebViewController.swift:274 warning: Function 'download(didFailWithError:resumeData:)' is unused
> ConnectWebViewController.swift:279 warning: Function 'downloadDidFinish()' is unused
> ConnectWebViewController.swift:303 warning: Function 'download(_:decideDestinationUsing:suggestedFilename:)' is unused
> ConnectWebViewController.swift:310 warning: Function 'download(_:didFailWithError:resumeData:)' is unused
> ConnectWebViewController.swift:316 warning: Function 'downloadDidFinish(_:)' is unused
> ```
> 
> 
> 

These are all called from delegate methods and are not actually dead code.

StripeConnect/StripeConnect/Source/Internal/Extensions/UIViewController+StripeConnect.swift
```diff
@@ -0,0 +1,19 @@
...
+
+import UIKit
+
+extension UIViewController {
+    /// Helper that adds a child view controller and pins its view to this view controller's view
+    func addChildAndPinView(_ child: UIViewController) {
+        child.willMove(toParent: self)
+        addChild(child)
+        view.addSubview(child.view)
+        child.view.frame = view.bounds
```
nschris-stripe, line 16: Shouldn't this use constraints?

mludowise-stripe: We don't need to if `translatesAutoresizingMaskIntoConstraints = true`, which it is by default

StripeConnect/StripeConnect/Source/Internal/Webview/ConnectComponentWebViewController.swift
```diff
@@ -81,44 +82,63 @@ class ConnectComponentWebView: ConnectWebView {
...
     }
 
     required init?(coder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
 
+    // MARK: - UIViewController
+
     override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        super.traitCollectionDidChange(previousTraitCollection)
```
nschris-stripe, line 113: oops

StripeConnect/StripeConnect/Source/Internal/Webview/HTTPStatusError.swift
nschris-stripe: I would assume that integrators would appreciate the convenience of being able to directly use a user facing error message from the SDK. We could add a task for this.

nschris-stripe (approved):

PR #4139, opened 2024-10-15
[Connect] Assert app includes NSCameraUsageDescription

mludowise-stripe:
## Summary
Throws an assertion failure when instantiating `EmbeddedComponentManager` if the app does not include `NSCameraUsageDescription` in its info.plist

## Motivation
During the bug bash, we forgot to include `NSCameraUsageDescription` in our app and it took us awhile to figure out that the reason was we forgot to enable camera permissions. Even though we will be including instructions to do this in our integration guide, in the event the app integrator skips over it, an assertion should hopefully help them catch the issue early.

https://jira.corp.stripe.com/browse/MXMOBILE-2839

## Testing
Removed `NSCameraUsageDescription` from the example app's info.plist and verified a console warning displayed.

https://github.com/user-attachments/assets/8e558128-8394-4bff-aa9e-f5fdeecf87a0


StripeConnect/StripeConnect/Source/Helpers/Log.swift
```diff
@@ -0,0 +1,21 @@
...
+/// Wraps `os_log` and prepends messages with "[StripeConnect]"
+struct Log {
+    static let log = OSLog(subsystem: "StripeConnect", category: "general")
+
+    static func debug(_ message: String) {
+        os_log("[StripeConnect] %{public}@", log: log, type: .debug, message)
+    }
+
+    static func warn(_ message: String) {
+        os_log("[StripeConnect] Warning: %{public}@", log: log, type: .error, message)
```
mludowise-stripe, line 19: Using the `.error` type makes the message show up in yellow in the console

StripeConnect/StripeConnect/Source/Internal/Webview/MessageHandlers/DebugMessageHandler.swift
```diff
@@ -9,7 +9,7 @@ import Foundation
 
 // Emitted when the SDK should print to the console in debug mode.
 class DebugMessageHandler: ScriptMessageHandler<String> {
-    init(didReceiveMessage: @escaping (String) -> Void = { Swift.debugPrint($0) }) {
```
mludowise-stripe, line 12: Use our Log helper so it prepends "[StripeConnect] " to the message when we receive a debug message from the web view

kush-stripe (commented):
Not including the camera permission might be a pretty common failure mode. Additionally, the console box can be bit of a dumping ground for all sorts of log messages so this could be easily missed. Is there something more aggressive that we could do here? E.g throwing an assertion failure when the SDK gets initialized or something along those lines?

kush-stripe (changes requested):
Not including the camera permission might be a pretty common failure mode. Additionally, the console box can be bit of a dumping ground for all sorts of log messages so this could be easily missed. Is there something more aggressive that we could do here? E.g throwing an assertion failure when the SDK gets initialized or something along those lines?

kush-stripe (approved):

PR #4141, opened 2024-10-15
Implement embedded `update` (part 1)

yuki-stripe:
## Summary
Implements `update` w/ cancel functionality.

Still to come:
1. Make `view` a view that _contains_ embedded view so we can swap to the updated embedded view with an animation.
1. Restore previous customer input + E2E tests (e.g. load PI -> fill out card form -> update to SI -> expect form to be preserved but w/o checkbox) 
1. Make `confirm` handle in-flight and failed `update` calls.
1. (Bonus) Cancel network calls etc. from previous `update` to reduce battery/network usage. Can apply this to FC.update as well.

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-2524

## Testing
See tests.

## Changelog
Not user facing.


StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentElementTest.swift
```diff
@@ -0,0 +1,134 @@
...
+        let sut = try await EmbeddedPaymentElement.create(intentConfiguration: paymentIntentConfig, configuration: configuration)
+        // ...updating w/ an invalid intent config should fail...
+        var intentConfig = paymentIntentConfig
+        intentConfig.mode = .setup(currency: "Invalid currency", setupFutureUsage: .offSession)
+        let updateResult = await sut.update(intentConfiguration: intentConfig)
+        switch updateResult {
+        case .failed(error: let error):
+            print(error.nonGenericDescription)
+        default:
+            break
```
wooj-stripe, line 87: Instead of breaking, seems like it should be an `XCTFailure()`

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentElementTest.swift
```diff
@@ -0,0 +1,134 @@
...
+            secondUpdateExpectation.fulfill()
+        }
+        await fulfillment(of: [secondUpdateExpectation])
+    }
+
+    func testUpdateCancelsInFlightUpdate() async throws {
+        // Given a EmbeddedPaymentElement instance...
+        let sut = try await EmbeddedPaymentElement.create(intentConfiguration: paymentIntentConfig, configuration: configuration)
+        // ...updating...
+        async let _updateResult = await sut.update(intentConfiguration: paymentIntentConfig)
```
wooj-stripe, line 105: If we don't want the first update to finish, I would have expected the first call to update to not have an await?

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentElementTest.swift
yuki-stripe: Oops, ty

wooj-stripe (approved):

PR #4144, opened 2024-10-16
[Connect] Support opening authenticated web views

mludowise-stripe:
## Summary
Adds support for the `openAuthenticatedWebView` JS messenger. When the web view sends this message to the mobile SDK, we open an `ASWebAuthenticatedSession` to the given URL and then call `window.returnedFromAuthenticatedWebView` with the return URL.

Also adds id to `returnedFromAuthenticatedWebView` payload so that web layer can associate the callback with the original open request.

## Motivation
https://jira.corp.stripe.com/browse/MXMOBILE-2578

## Testing

| Scenario                         | Unit test coverage | Manually tested  |
| -------------------------------- | ------------------ | ---------------- |
| User cancel action               |         ✓          | ✓ modal + navbar |
| Error state: multiple sessions   |         ✓          | ✓                |
| Error state: session can't start |         ✓          | ✓ background app |
| Error state: no window           |         ✓          |                  |
| Error state: misc error          |         ✓          |                  |
| Successfully redirect            |         ✓          | ✓                |

Because this callback is not yet implemented on web, manual testing was performed by directly invoking the handler in the Safari debug console (sound on):

https://github.com/user-attachments/assets/f8a0b764-6d4d-4d68-8d95-fd325dae3fee


StripeConnect/StripeConnect/Source/Internal/AuthenticatedWebView/AuthenticatedWebViewManager.swift
```diff
@@ -0,0 +1,79 @@
...
+
+        return returnUrl
+    }
+}
+
+// MARK: - ASWebAuthenticationPresentationContextProviding
+
+extension AuthenticatedWebViewManager: ASWebAuthenticationPresentationContextProviding {
+    func presentationAnchor(for session: ASWebAuthenticationSession) -> ASPresentationAnchor {
+        return window ?? .init()
```
mludowise-stripe, line 77: It's possible that calling `UIWindow.init()` could cause cause a fatalError since this is called off the main thread, but we do check that `window` is non-nil before starting the session, so I don't think this could realistically happen. 

When I tested locally and changed this to always return `.init()` I was triggering the UI thread checker. The only solution I can think of to work around this is to instantiate a fallback UIWindow on init and change this to:
```
return window ?? fallbackWindow
```

But then we would also need to enforce that `init` is always called on the main thread. I considered annotating the class with `@MainActor` but then it complicates the callsites where it gets instantiated.

StripeConnect/StripeConnect/Source/Internal/Webview/MessageHandlers/Helpers/ScriptMessageHandlerWithReply.swift
```diff
@@ -36,8 +36,8 @@ class ScriptMessageHandlerWithReply<Payload: Decodable, Response: Encodable>: NS
 
             return (response, nil)
         } catch {
-            debugPrint("Error processing message: \(error.localizedDescription)")
-            return (nil, error.localizedDescription)
+            debugPrint("Error processing message: \((error as NSError).debugDescription)")
+            return (nil, (error as NSError).debugDescription)
```
mludowise-stripe, line 40: I think the debug description is more valuable than the user-facing localized description for logging in the web layer. For errors that don't have a localized description implemented, this was just showing the error domain & code, but I believe that most NSError implementations have this encoded in the debugDescription along with some other useful things.

StripeConnect/StripeConnect/Source/Internal/Webview/MessageHandlers/OpenAuthenticatedWebViewMessageHandler.swift
```diff
@@ -8,14 +8,20 @@
 import Foundation
 
 /// Indicates to open the provided URL in an `ASWebAuthenticationSession`.
-class OpenAuthenticatedWebViewMessageHandler: ScriptMessageHandler<OpenAuthenticatedWebViewMessageHandler.Payload> {
+class OpenAuthenticatedWebViewMessageHandler: ScriptMessageHandlerWithReply<OpenAuthenticatedWebViewMessageHandler.Payload, OpenAuthenticatedWebViewMessageHandler.Response> {
```
mludowise-stripe, line 11: Proposing that we combine `OpenAuthenticatedWebViewMessageHandler` and `ReturnedFromAuthenticatedWebViewSender` into a single async handler.

See [slack discussion](https://stripe.slack.com/archives/C06JRADKJF7/p1729033305657649?thread_ts=1729031087.701449&cid=C06JRADKJF7)

StripeConnect/StripeConnectTests/Internal/AuthenticatedWebView/AuthenticatedWebViewManagerTests.swift
```diff
@@ -0,0 +1,163 @@
...
+//  StripeConnect
+//
+//  Created by Mel Ludowise on 10/15/24.
+//
+
+import AuthenticationServices
+@testable import StripeConnect
+import XCTest
+
+class AuthenticatedWebViewManagerTests: XCTestCase {
```
mludowise-stripe, line 12: I had to get creative to figure out how to unit test this, so lmk if you think this is easy enough to follow

StripeConnect/StripeConnect/Source/Internal/Webview/MessageHandlers/OpenAuthenticatedWebViewMessageHandler.swift
mludowise-stripe: Opted not to do this since Android cannot support this

StripeConnect/StripeConnect/Source/Internal/Webview/MessageSenders/ReturnedFromAuthenticatedWebViewSender.swift
```diff
@@ -10,8 +10,10 @@ import Foundation
 /// Notifies that the user finished the flow within the `ASWebAuthenticationSession`
 struct ReturnedFromAuthenticatedWebViewSender: MessageSender {
     struct Payload: Codable, Equatable {
-        /// The return URL from the `ASWebAuthenticationSession` redirect. This value will be nil if the user canceled out fo the view
-        let url: String?
+        /// The return URL from the `ASWebAuthenticationSession` redirect. This value will be nil if the user canceled out of the view
+        let url: URL?
+        /// The unique identifier sent from the web view in `openAuthenticatedWebView`
+        let id: String
```
mludowise-stripe, line 16: 👀 @mxl-stripe 

Added `id` so we could round-trip track the original open request

StripeConnect/StripeConnect/Source/Internal/AuthenticatedWebView/AuthenticatedWebViewManager.swift
nschris-stripe: Yeah the only risk I see here is that since window is weak it could potentially become nil after checking it. That is unlikely and I can't think of a scenario where that could happen. That got me thinking though, do we really need to make window weak? As long as we nil out window after the continuation is complete I think it's reasonable to hold a strong reference to it.

Even better if you created a class dedicated to being the presentationContextProvider and have that take in a non optional then you will always be guaranteed a window and you don't need to worry about nil-ing anything out because that will happen automatically. The below is all just code I typed here to get the point across you might have to tweak it a bit:

```
class AuthenticatedWebViewPresentationContextProvider: NSObject, ASWebAuthenticationPresentationContextProviding {
   let window: UIWindow
   init(window: UIWindow) {
    self.window = window
   }

    func presentationAnchor(for session: ASWebAuthenticationSession) -> ASPresentationAnchor {
        return window
   }
 }

   class AuthenticatedWebViewManager: NSObject {
    ...

     func present(with url: URL, in window: UIWindow?) async throws -> URL? {
        guard authSession == nil else {
            throw AuthenticatedWebViewError.alreadyPresenting
        }
        guard let window else {
            throw AuthenticatedWebViewError.noWindow
        }

        // Context provider holds onto window but is only scoped to this function.
        let contextProvider = .init(window: window)

        let returnUrl: URL? = try await withCheckedThrowingContinuation { continuation in
            let authSession = sessionFactory(url, StripeConnectConstants.authenticatedWebViewReturnUrlScheme) { returnUrl, error in

                if let authenticationSessionError = error as? ASWebAuthenticationSessionError,
                   authenticationSessionError.code == .canceledLogin {
                    // The user either selected "Cancel" in the initial modal
                    // prompting them to "Sign In" or they hit the "Cancel"
                    // button in presented browser view
                    continuation.resume(returning: nil)
                    return
                } else if let error {
                    continuation.resume(throwing: error)
                    return
                }

                continuation.resume(returning: returnUrl)
            }
            authSession.presentationContextProvider = contextProvider

            self.authSession = authSession

            guard authSession.canStart,
                  authSession.start() else {
                continuation.resume(throwing: AuthenticatedWebViewError.cannotStartSession)
                return
            }
        }
        return returnUrl
    }
}


```

StripeConnect/StripeConnect/Source/Internal/AuthenticatedWebView/AuthenticatedWebViewManager.swift
```diff
@@ -0,0 +1,79 @@
+//
+//  AuthenticatedWebViewManager.swift
+//  StripeConnect
+//
+//  Created by Mel Ludowise on 10/15/24.
+//
+
+import AuthenticationServices
+
+/// Singleton helper to manage
```
nschris-stripe, line 10: I don't think this is a singleton anymore

StripeConnect/StripeConnect/Source/Internal/AuthenticatedWebView/AuthenticatedWebViewManager.swift
nschris-stripe: The other option here is that present probably shouldn't take in a nil window if it's just going to immediately error. This seems like something that should be handled at the layer that is trying to present.

StripeConnect/StripeConnect/Source/Internal/Webview/ConnectComponentWebViewController.swift
```diff
@@ -229,4 +239,19 @@ private extension ConnectComponentWebViewController {
...
+    func openAuthenticatedWebView(_ payload: OpenAuthenticatedWebViewMessageHandler.Payload) {
+        Task { @MainActor in
+            do {
+                // TODO: MXMOBILE-2491 log `component.authenticated_web.*` analytic
+                let returnUrl = try await authenticatedWebViewManager.present(with: payload.url, in: view.window)
+
+                sendMessage(ReturnedFromAuthenticatedWebViewSender(payload: .init(url: returnUrl, id: payload.id)))
+            } catch {
+                // TODO: MXMOBILE-2491 log `component.authenticated_web.error` analytic
+                debugPrint("Error returning from authenticated web view: \(error)")
```
nschris-stripe, line 253: If the ConnectComponentWebViewController has a parent ViewController and parent view then I would say at that point it would indicate an integration error. Would we still want to log this to our analytics?

nschris-stripe (dismissed):
Approving, but I have made some suggestions for improvements. Up to you whether you want to follow up on these in a later PR or address them here.

StripeConnect/StripeConnect/Source/Internal/AuthenticatedWebView/AuthenticatedWebViewManager.swift
mludowise-stripe: yeah, I did that for conv

> The other option here is that present probably shouldn't take in a nil window if it's just going to immediately error. This seems like something that should be handled at the layer that is trying to present.

Yeah, I did that for convenience just so that I could throw one error and check for it in the ComponentWebViewController but probably not the best setup.

mludowise-stripe: > if you created a class dedicated to being the presentationContextProvider and have that take in a non optional then you will always be guaranteed a window

I like this idea. I'm gonna see if I can get this working

StripeConnect/StripeConnect/Source/Internal/Webview/ConnectComponentWebViewController.swift
mludowise-stripe: The ASWebAuthenticationSession can also return an error, although I wasn't able to trigger this case myself other than `ASWebAuthenticationSessionError.canceledLogin`. In that case, I don't think it's an integration error.

If the app is backgrounded (`canStart` is false) or it's not in the view hierarchy (window is nil), then I do think we want to log it since it means we're triggering an authenticated webview to open when the user hasn't taken any action to do so.

StripeConnect/StripeConnect/Source/Internal/AuthenticatedWebView/AuthenticatedWebViewManager.swift
mludowise-stripe: - I took your suggestion of using a `AuthenticatedWebViewPresentationContextProvider` – much cleaner!
- I updated `present` to take a UIView and then grab its window param. I think this makes the `AuthenticatedWebViewError.noWindow` error handling make a lot more sense since the intent was to consolidate all the error handling into one place
- I annotated `present` with `@MainActor` to ensure we're always accessing the window on the main thread.

StripeConnect/StripeConnect/Source/Internal/AuthenticatedWebView/AuthenticatedWebViewManager.swift
mludowise-stripe: Good catch – I didn't even finish that sentence 🤦‍♀️ 

Updated

StripeConnect/StripeConnect/Source/Internal/Webview/MessageHandlers/Helpers/ScriptMessageHandlerWithReply.swift
mludowise-stripe: Note: I originally made this change because I was playing around with making `openAuthenticatedWebView` an async function that returned a debug error message to the web view, but ended up reverting that change. But I think we till want the debug description here anyway so leaving it in.

StripeConnect/StripeConnect/Source/Internal/Webview/ConnectComponentWebViewController.swift
mludowise-stripe: - I renamed `noWindow` to `notInViewHierarchy` for clarity
- I also removed the debug descriptions from these errors since we won't be logging them anywhere (originally I had planned to return them to the web layer, but we ended up not doing that)

mludowise-stripe:
> 
> ### StripePaymentSheet
> ...

This was a temporary rebasing issue that should be fixed now.

mludowise-stripe:
> 
> ### StripePaymentSheet
> ...

This was a temporary rebasing issue that should be fixed now.

nschris-stripe (approved):

PR #4145, opened 2024-10-16
Pass `link_mode` to LAS creation

tillh-stripe:
## Summary

This pull request adds a `link_mode` field to the LAS creation request. If Link is not enabled, we send `LINK_DISABLED` just like Web does.

(cc @selinafeng-stripe)

## Motivation

[BANKCON-15655](https://jira.corp.stripe.com/browse/BANKCON-15655)

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkPaymentController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkPaymentController.swift
@@ -147,7 +147,7 @@ import UIKit
                                                    paymentMethodType: .link,
                                                    customerName: configuration.defaultBillingDetails.name,
                                                    customerEmailAddress: configuration.defaultBillingDetails.email,
-                                                   linkMode: .linkPaymentMethod,
+                                                   linkMode: nil,
                                                    additionalParameters: parameters) { [weak self] linkAccountSession, error in
                     self?.generateManifest(continuation: continuation, error: error, emailAddress: self?.configuration.defaultBillingDetails.email, linkAccountSession: linkAccountSession)
                 }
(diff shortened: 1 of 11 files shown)


StripePayments/StripePayments/Source/API Bindings/STPAPIClient+LinkAccountSession.swift
```diff
@@ -75,6 +80,7 @@ public extension STPAPIClient {
         parameters["amount"] = amount
         parameters["currency"] = currency
         parameters["on_behalf_of"] = onBehalfOf
+        parameters["link_mode"] = linkMode?.rawValue ?? "LINK_DISABLED"
```
mats-stripe, line 83: Do we need to pass `LINK_DISABLED`, or could we only pass the `link_mode` parameter when it's non-nil?

```swift
if let linkMode {
    parameters["link_mode"] = linkMode.rawValue
}
```

mats-stripe (approved):
LGTM, just one non-blocking suggestion

StripePayments/StripePayments/Source/API Bindings/STPAPIClient+LinkAccountSession.swift
tillh-stripe: Yep, we want to pass `LINK_DISABLED` to be in line with Web’s behavior.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkPaymentController.swift
```diff
@@ -147,7 +147,8 @@ import UIKit
                                                    paymentMethodType: .link,
                                                    customerName: configuration.defaultBillingDetails.name,
                                                    customerEmailAddress: configuration.defaultBillingDetails.email,
-                                                   additionalParameteres: parameters) { [weak self] linkAccountSession, error in
+                                                   linkMode: .linkPaymentMethod,
```
tillh-stripe, line 150: @carlosmuvi-stripe Do you think it’s fine to pass this for the custom integration?

mats-stripe (approved):

PR #4147, opened 2024-10-16
Fix product usage test for flow controller

wooj-stripe:
## Summary
Running this test in isolation was failing due to the movement of where we are setting the static variable.  To remedy this, make the instance injectable on the non-public interface, and utilize the @testable import to call it

## Motivation
Test is failing in isolation.

## Testing
locally

### DIFF
--- a/StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetAnalyticsHelperTest.swift
+++ b/StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetAnalyticsHelperTest.swift
@@ -28,9 +28,11 @@ final class PaymentSheetAnalyticsHelperTest: XCTestCase {
         // Clear product usage prior to testing PaymentSheet.FlowController
         STPAnalyticsClient.sharedClient.productUsage = Set()
         XCTAssertTrue(STPAnalyticsClient.sharedClient.productUsage.isEmpty)
-
-        PaymentSheet.FlowController.create(paymentIntentClientSecret: "", configuration: PaymentSheet.Configuration()) { _ in
+        let expectation = expectation(description: "create")
+        PaymentSheet.FlowController.create(paymentIntentClientSecret: "pi_123_secret_456", configuration: PaymentSheet.Configuration()) { _ in
+            expectation.fulfill()
         }
+        wait(for: [expectation], timeout: 2.0)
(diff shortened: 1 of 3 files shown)


StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetAnalyticsHelperTest.swift
```diff
@@ -21,16 +21,7 @@ final class PaymentSheetAnalyticsHelperTest: XCTestCase {
...
-            configuration: PaymentSheet.Configuration(),
-            loadResult: .init(
-                intent: .paymentIntent(STPFixtures.paymentIntent()),
-                elementsSession: .makeBackupElementsSession(with: STPFixtures.paymentIntent()),
-                savedPaymentMethods: [],
-                paymentMethodTypes: [.stripe(.card)]
-            ), analyticsHelper: .init(isCustom: true, configuration: PaymentSheet.Configuration())
-        )
-        XCTAssertTrue(STPAnalyticsClient.sharedClient.productUsage.contains("PaymentSheet.FlowController"))
+        // Test for PaymentSheet.FlowController is added to productUsage is in PaymentSheetAPITest.testUpdate()
```
yuki-stripe, line 24: It might make more sense to make this call `PaymentSheet.FlowController.create(...)`

wooj-stripe: We could, but create calls load(), which is a bit more difficult to stub out.  We'd need to make these tests network stubbable and/or mocked, which is a bit more work than (imho) is worth to test that product usage is being set.

yuki-stripe: I think we don't even need to wait for the `create` call to complete, we can just call it and forget about it.  No need to stub/mock, is there?

Ooh also just realized, we should clear `STPAnalyticsClient.sharedClient.productUsage` at the start of this test so it fails when it should have.  Right now I guess it's hardly testing anything, just that `STPAnalyticsClient.sharedClient.productUsage` was set at some point in the test suite beforehand 😂

wooj-stripe: Good point on not needing to mock as well as clearing out product usage!

yuki-stripe (approved):

PR #4150, opened 2024-10-16
Embedded update pt 2 - update view

yuki-stripe:
## Summary
Previous PR: https://github.com/stripe/stripe-ios/pull/4141

Makes EmbeddedPaymentElement view a view that contains embedded view so we can swap to the updated embedded view with an animation.

https://github.com/user-attachments/assets/533e0fb1-c848-4fea-9793-30347bbd90f1

Still to come:

- Restore previous customer input + E2E tests (e.g. load PI -> fill out card form -> update to SI -> expect form to be preserved but w/o checkbox)
- Make confirm handle in-flight and failed update calls.
- (Bonus) Cancel network calls etc. from previous update to reduce battery/network usage. Can apply this to FC.update as well.

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-2583

## Testing
See snapshot test. 

## Changelog
Not user facing


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
```diff
@@ -178,6 +179,14 @@ public final class EmbeddedPaymentElement {
             configuration: configuration,
             loadResult: loadResult
         )
+        self.containerView = EmbeddedPaymentElementContainerView(
+            embeddedPaymentMethodsView: embeddedPaymentMethodsView
+        )
+
+        self.containerView.updateSuperviewHeight = { [weak self] in
```
porter-stripe, line 186: Any reason to not just remove `containerView` and just return it directly on `view`? Guessing it's to avoid casting between UIView and the container view?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElementContainerView.swift
```diff
@@ -0,0 +1,78 @@
...
+//  StripePaymentSheet
+//
+//  Created by Yuki Tokuhiro on 10/16/24.
+//
+import UIKit
+
+/// The view that's vended to the merchant, containing the embedded view.  We use this to be able to swap out the embedded view with an animation when `update` is called.
+class EmbeddedPaymentElementContainerView: UIView {
+    var updateSuperviewHeight: () -> Void = {}
+    var view: UIView
```
porter-stripe, line 12: Any reason these aren't private (excluding `updateSuperviewHeight`, although that feels like it should come in at init time and be optional)? Also wondering it it ever makes sense to use this class when `view` is not of type `EmbeddedPaymentMethodsView`. Should it just be `EmbeddedPaymentMethodsView` over `UIView` in this class?

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentElementSnapshotTests.swift
```diff
@@ -0,0 +1,52 @@
...
+//
+
+import StripeCoreTestUtils
+@_spi(STP) @testable import StripePayments
+@_spi(EmbeddedPaymentElementPrivateBeta) @testable import StripePaymentSheet
+@testable import StripePaymentsTestUtils
+@_spi(STP) @testable import StripeUICore
+import XCTest
+
+class EmbeddedPaymentElementSnapshotTests: STPSnapshotTestCase, EmbeddedPaymentElementDelegate {
```
porter-stripe, line 15: Are we planning to add more tests to this file? If so do we have a ticket or will that happen as a follow up?

Tests/ReferenceImages_64/StripePaymentSheetTests.EmbeddedPaymentElementSnapshotTests/testUpdateFromCardToCardAndUSBankAccountWithCompletionHandler_@3x.png
porter-stripe, line 1: Is all the transparent padding expected? Would that be the merchant's UI/background in the real world?

porter-stripe (approved):
Few non blocking Q's, nice!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
yuki-stripe: hm can you elaborate, not sure I understand.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElementContainerView.swift
yuki-stripe: `updateSuperviewHeight` is a var for convenience; `self` isn't available when `EmbeddedPaymentElement` initializes this.

Will make them private & change the UIView type though!

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentElementSnapshotTests.swift
yuki-stripe: Yep will add more tests with the next PRs (see bullets in summary)!

Tests/ReferenceImages_64/StripePaymentSheetTests.EmbeddedPaymentElementSnapshotTests/testUpdateFromCardToCardAndUSBankAccountWithCompletionHandler_@3x.png
yuki-stripe: Padding is expected; that is the default `layoutMargins`. Whether it should be transparent or appearance.colors.background, IDK. But the background color of the view can be set by the merchant so I think it doesn't matter?

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentElementSnapshotTests.swift
porter-stripe: ACK, saw bullets just didn't think they'd be snapshot tests? Haven't thought much about it though.

porter-stripe (approved):

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentElementSnapshotTests.swift
yuki-stripe: Oh, missed "in this file".  I think we'll eventually add but I don't have any specific tests in mind for `update`.

PR #4154, opened 2024-10-17
Support Instant Debits with deferred intents

tillh-stripe:
## Summary

This pull request adds support for using Instant Debits with deferred intents in MPE. We keep this behind a local feature flag for now and will enable it after a successful bug bash.

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/Web/FinancialConnectionsWebFlowViewController.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/Web/FinancialConnectionsWebFlowViewController.swift
@@ -296,6 +296,8 @@ extension FinancialConnectionsWebFlowViewController {
 
 private extension URL {
     
+    /// The URL contains a base64-encoded payment method. We store its values in `LinkBankPaymentMethod` so that
+    /// we can parse it back in StripeCore.
     func extractLinkBankPaymentMethod() -> LinkBankPaymentMethod? {
         guard let encodedPaymentMethod = extractValue(forKey: "payment_method") else {
             return nil
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/Web/FinancialConnectionsWebFlowViewController.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/Web/FinancialConnectionsWebFlowViewController.swift
@@ -168,22 +168,20 @@ extension FinancialConnectionsWebFlowViewController {
                 switch result {
                 case .success(.success(let returnUrl)):
                     if manifest.isProductInstantDebits {
-                        if
-                            let paymentMethodId = Self.extractValue(from: returnUrl, key: "payment_method_id")
-                        {
+                        if let paymentMethod = returnUrl.extractLinkBankPaymentMethod() {
                             let instantDebitsLinkedBank = InstantDebitsLinkedBank(
-                                paymentMethodId: paymentMethodId,
-                                bankName: Self.extractValue(from: returnUrl, key: "bank_name")?
+                                paymentMethod: paymentMethod,
+                                bankName: returnUrl.extractValue(forKey: "bank_name")?
                                 // backend can return "+" instead of a more-common encoding of "%20" for spaces
                                     .replacingOccurrences(of: "+", with: " "),
-                                last4: Self.extractValue(from: returnUrl, key: "last4"),
+                                last4: returnUrl.extractValue(forKey: "last4"),
                                 linkMode: elementsSessionContext?.linkMode
                             )
                             self.notifyDelegateOfSuccess(result: .instantDebits(instantDebitsLinkedBank))
                         } else {
                             self.notifyDelegateOfFailure(
                                 error: FinancialConnectionsSheetError.unknown(
-                                    debugDescription: "payment_method_id was not returned"
+                                    debugDescription: "Invalid payment_method returned"
                                 )
                             )
                         }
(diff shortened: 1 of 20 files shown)


StripeFinancialConnections/StripeFinancialConnections/Source/Web/FinancialConnectionsWebFlowViewController.swift
```diff
@@ -293,9 +291,24 @@ extension FinancialConnectionsWebFlowViewController {
 
         notifyDelegate(result: .failed(error: error))
     }
+}
 
-    private static func extractValue(from url: URL, key: String) -> String? {
-        guard let components = URLComponents(url: url, resolvingAgainstBaseURL: false) else {
+private extension URL {
+    
+    func extractLinkBankPaymentMethod() -> LinkBankPaymentMethod? {
```
mats-stripe, line 298: Minor: A code comment here explaining that this object comes base64 encoded would be nice, since it's unlike the other return URL parameters!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentMethodFormViewController.swift
```diff
@@ -439,7 +438,7 @@ extension PaymentMethodFormViewController {
         }
         let additionalParameters: [String: Any] = [
             "product": "instant_debits",
-            "attach_required": true,
+//            "attach_required": true,
```
mats-stripe, line 441: Is this intentionally left commented out?

mats-stripe (approved):
LGTM!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentMethodFormViewController.swift
tillh-stripe: Nope, that’s an oversight.

StripeFinancialConnections/StripeFinancialConnections/Source/Web/FinancialConnectionsWebFlowViewController.swift
tillh-stripe: Will do!

mats-stripe (dismissed):
LGTM, pending sign off on the small public API changes

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
```diff
@@ -198,12 +198,27 @@ extension PaymentSheet {
                 )
             // MARK: ↪ Deferred Intent
             case .deferredIntent(let intentConfig):
-                handleDeferredIntentConfirmation(
-                    confirmType: .new(
+                let confirmType: ConfirmPaymentMethodType? = if let bank = confirmParams.instantDebitsLinkedBank {
```
yuki-stripe, line 201: Can you add a comment explaining this a bit? IDK maybe something like "Instant Debits should be handled as if it were a saved payment method because x"

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentMethodFormViewController.swift
```diff
@@ -437,11 +436,21 @@ extension PaymentMethodFormViewController {
...
             "product": "instant_debits",
-            "attach_required": true,
             "hosted_surface": "payment_element",
         ]
+        
+        switch intent {
+        case .paymentIntent:
+            additionalParameters["attach_required"] = true
+        case .setupIntent:
+            additionalParameters["attach_required"] = true
```
yuki-stripe, line 449:
```suggestion
        case .paymentIntent, .setupIntent:
            additionalParameters["attach_required"] = true
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
yuki-stripe: Also instead of this hacky thing where we're using `Paymentoption.new` but then overriding logic to make it confirm like it's saved, did you consider other options like
1. handling instant debits as a `PaymentOption.saved` 
2. making a new `PaymentOption.instantDebits` type

tillh-stripe: I went with option 1 and created a `PaymentOption.saved` in `PaymentMethodFormViewController.paymentOption`. What do you think?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
```diff
@@ -198,27 +198,12 @@ extension PaymentSheet {
                 )
             // MARK: ↪ Deferred Intent
             case .deferredIntent(let intentConfig):
-                let confirmType: ConfirmPaymentMethodType? = if let bank = confirmParams.instantDebitsLinkedBank {
```
yuki-stripe: Can we delete the other instances of this too (search `instantDebitsLinkedBank` in this file)

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentMethodFormViewController.swift
```diff
@@ -54,6 +54,16 @@ class PaymentMethodFormViewController: UIViewController {
...
                 return .external(paymentMethod: paymentMethod, billingDetails: params.paymentMethodParams.nonnil_billingDetails)
             }
+            
+            if case .instantDebits = paymentMethodType {
+                // We create the final payment method in the bank auth flow, therefore treating the Instant Debits
+                // result like a saved payment option.
+                guard let paymentMethod = instantDebitsFormElement?.getLinkedBank()?.paymentMethod.decode() else {
+                    return nil
+                }
+                return .saved(paymentMethod: paymentMethod, confirmParams: nil)
```
yuki-stripe, line 64: There's maybe something smelly about this being inside the `if let params = form.updateParams(..)` block when we ignore the params. We could use the `params.instantDebitsLinkedBank` instead of this but I'm not sure that gets at the crux of the issue. 

I'll leave this half-baked comment here but it's not at all blocking b/c I'm not sure what the problem is, if any.

tillh-stripe: True, that is weird. I’ll use `params.instantDebitsLinkedBank` for now, which seems like a marginal improvement.

yuki-stripe (approved):
This looks good to me (minus the comments)

mats-stripe (approved):

mats-stripe (approved):

PR #4159, opened 2024-10-18
[Connect] Using FinancialConnections SDK in Connect

mludowise-stripe:
## Summary
Enables using the mobile native FinancialConnections SDK from Connect web components to retrieve the user's bank token.

* Adds a dependency from `StripeConnect` -> `StripeFinancialConnections`
* Adds a new JS message handler and sender to trigger the FinancialConnections flow from web and pass back the bank token:
  * `openFinancialConnections` message handler: Triggers the FinancialConnectionsSheet to open:
    * Param `clientSecret`: The Financial Connections Session client secret for the connected account
    * Param `connectedAccountId`: The ID of the connected account
    * Param `id`: A unique UUID that gets passed back to the web layer on `returnFromFinancialConnections`
  * `returnedFromFinancialConnections` message sender: Returns the connected bank token to the web layer
    * Param `bankToken`: The bank account token of the connected account. This value will be null if the user canceled out of the flow or an error occurred
    * Param `id`: The `id` originally passed to the mobile layer from `openFinancialConnections`

## Motivation
https://jira.corp.stripe.com/browse/MXMOBILE-2526
https://docs.google.com/document/d/1WUTvnK8kqsB00zcmUzwVC_CW5MMmXC_CDk8eQTAzaaQ

## Testing

https://github.com/user-attachments/assets/abcb5236-17c9-4833-9788-350f0aa497c2

### DIFF
--- a/Example/StripeConnectExample/StripeConnect Example.xcodeproj/project.pbxproj
+++ b/Example/StripeConnectExample/StripeConnect Example.xcodeproj/project.pbxproj
@@ -40,6 +40,8 @@
 		41E9A1C12C7CE30000EDE131 /* AppSettingsView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 41E9A1C02C7CE30000EDE131 /* AppSettingsView.swift */; };
 		E62B3CD72C99EA020098B607 /* StripeCore.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = E62B3CD62C99EA020098B607 /* StripeCore.framework */; };
 		E62B3CD82C99EA020098B607 /* StripeCore.framework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = E62B3CD62C99EA020098B607 /* StripeCore.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
+		E640C9D92CC1970F009D0C6E /* StripeFinancialConnections.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = E640C9D82CC1970F009D0C6E /* StripeFinancialConnections.framework */; };
+		E640C9DA2CC1970F009D0C6E /* StripeFinancialConnections.framework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = E640C9D82CC1970F009D0C6E /* StripeFinancialConnections.framework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
 		E6C5E52B2CA76B0F0021444D /* PresentationSettingsView.swift in Sources */ = {isa = PBXBuildFile; fileRef = E6C5E52A2CA76B030021444D /* PresentationSettingsView.swift */; };
 		E6C5E52D2CA771730021444D /* PresentationSettings.swift in Sources */ = {isa = PBXBuildFile; fileRef = E6C5E52C2CA771700021444D /* PresentationSettings.swift */; };
 /* End PBXBuildFile section */
@@ -64,6 +66,7 @@
 				E62B3CD82C99EA020098B607 /* StripeCore.framework in Embed Frameworks */,
 				416E9ECC2C76BE0C00A0B917 /* StripeConnect.framework in Embed Frameworks */,
 				4161C28A2CA1B438005BD67C /* StripeUICore.framework in Embed Frameworks */,
+				E640C9DA2CC1970F009D0C6E /* StripeFinancialConnections.framework in Embed Frameworks */,
 			);
 			name = "Embed Frameworks";
 			runOnlyForDeploymentPostprocessing = 0;
@@ -104,6 +107,7 @@
 		41E9A1BE2C7CD3C700EDE131 /* UIViewController+NavigationEmbed.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "UIViewController+NavigationEmbed.swift"; sourceTree = "<group>"; };
 		41E9A1C02C7CE30000EDE131 /* AppSettingsView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AppSettingsView.swift; sourceTree = "<group>"; };
 		E62B3CD62C99EA020098B607 /* StripeCore.framework */ = {isa = PBXFileReference; explicitFileType = wrapper.framework; path = StripeCore.framework; sourceTree = BUILT_PRODUCTS_DIR; };
+		E640C9D82CC1970F009D0C6E /* StripeFinancialConnections.framework */ = {isa = PBXFileReference; explicitFileType = wrapper.framework; path = StripeFinancialConnections.framework; sourceTree = BUILT_PRODUCTS_DIR; };
 		E6C5E52A2CA76B030021444D /* PresentationSettingsView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = PresentationSettingsView.swift; sourceTree = "<group>"; };
 		E6C5E52C2CA771700021444D /* PresentationSettings.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = PresentationSettings.swift; sourceTree = "<group>"; };
 /* End PBXFileReference section */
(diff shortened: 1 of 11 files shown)


Example/StripeConnectExample/StripeConnectExample/MainViewController.swift
```diff
@@ -63,7 +63,18 @@ class MainViewController: UITableViewController {
...
+        /*
+         NOTE:
+         We are using a custom URL scheme in this example, but Stripe recommends
+         against doing this in production code to prevent malicious apps from
+         reusing the same URL scheme to intercept sensitive user data.
+
+         Instead, Stripe recommends using associated domains and adding the
+         return url to your websites `apple-app-site-association` file:
+         https://developer.apple.com/documentation/xcode/supporting-associated-domains
+         */
```
mludowise-stripe, line 75: Wording here needs to be tweaked – it sounds like we disallow custom URL schemes in livemode

Context:
https://stripe.slack.com/archives/C02KW8G938W/p1729217697015449

StripeConnect/StripeConnect/Source/EmbeddedComponentManager.swift
```diff
@@ -28,10 +28,14 @@ public class EmbeddedComponentManager {
...
     var shouldLoadContent: Bool = true
 
+    /// Return URL used for FinancialConnections
+    let returnUrl: String
+
     /**
      Initializes a StripeConnect instance.
 
      - Parameters:
+       - returnUrl: A return URL that can be used to return to the embedded component
```
mludowise-stripe, line 38: This will trigger another API review

StripeFinancialConnections/StripeFinancialConnections/Source/Common/HostController.swift
```diff
@@ -10,12 +10,12 @@ import UIKit
...
 /// An internal result type that helps us handle both
 /// Financial Connections and Instant Debits
-@_spi(STP) public enum HostControllerResult {
+@_spi(STP) @frozen public enum HostControllerResult {
     case completed(Completed)
     case failed(error: Error)
     case canceled
 
-    @_spi(STP) public enum Completed {
+    @_spi(STP) @frozen public enum Completed {
```
mludowise-stripe, line 18: Froze these so that we wouldn't have to handle `@unknown default` when switching on the result.

If we ever do add more results, this will trigger a compile error in `StripeConnect` and ensure we think through what the expected behavior should be.

Example/StripeConnectExample/StripeConnect Example.xcodeproj/project.pbxproj
```diff
@@ -597,6 +603,7 @@
 				DEVELOPMENT_TEAM = "";
 				GENERATE_INFOPLIST_FILE = YES;
 				INFOPLIST_FILE = StripeConnectExample/Info.plist;
+				INFOPLIST_KEY_NSCameraUsageDescription = "We use the camera to scan your ID documents";
```
mludowise-stripe, line 606: IDK why Xcode pulled this out of Info.plist

StripeConnect/StripeConnect/Source/EmbeddedComponentManager.swift
mludowise-stripe: Actually, we might not need to do this – it looks like the returnURL is optional and only used if want to enable the native banking app to oauth and return back to the app.

Context:
https://stripe.slack.com/archives/C02KW8G938W/p1729221773775259?thread_ts=1729217697.015449&cid=C02KW8G938W

Example/StripeConnectExample/StripeConnectExample/MainViewController.swift
mludowise-stripe: turns out we don't need this

StripeConnect/StripeConnect/Source/EmbeddedComponentManager.swift
mludowise-stripe: turns out we don't need a return url

StripeFinancialConnections/StripeFinancialConnections/Source/Common/HostController.swift
mludowise-stripe: this is no longer needed after changing to `presentForToken`

Example/StripeConnectExample/StripeConnect Example.xcodeproj/project.pbxproj
mludowise-stripe: Reverted this

StripeConnect/StripeConnect/Source/Internal/Webview/MessageSenders/ReturnedFromFinancialConnections.swift
```diff
@@ -0,0 +1,19 @@
...
+//
+//  Created by Mel Ludowise on 10/17/24.
+//
+
+/// Notifies that the user finished the FinancialConnections flow
+struct ReturnedFromFinancialConnections: MessageSender {
+    struct Payload: Codable, Equatable {
+        /// The linked bank account token.
+        /// This value will be nil if the user canceled the flow or an error occurred.
+        let bankToken: String?
```
mludowise-stripe, line 13: Alternatively, if we don't need to return the bank token, we could send a `success | cancelled | failed` enum value to the web view

StripeConnect/StripeConnect/Source/Internal/Webview/ConnectComponentWebViewController.swift
```diff
@@ -49,13 +53,15 @@ class ConnectComponentWebViewController: ConnectWebViewController {
         // Should only be overridden for tests
         notificationCenter: NotificationCenter = NotificationCenter.default,
         webLocale: Locale = Locale.autoupdatingCurrent,
-        authenticatedWebViewManager: AuthenticatedWebViewManager = .init()
+        authenticatedWebViewManager: AuthenticatedWebViewManager = .init(),
+        financialConnectionsPresenter: FinancialConnectionsPresenter = .init()
```
nschris-stripe, line 57: I'm starting to second guess whether these should have default values or not. I'm worried that one day we will do a handoff to a new ConnectComponentWebViewController instance and not provide the existing managers.

StripeConnect/StripeConnect/Source/Internal/FinancialConnectionsPresenter.swift
```diff
@@ -0,0 +1,34 @@
...
+        from presentingViewController: UIViewController
+    ) async -> FinancialConnectionsSheet.TokenResult {
+        let financialConnectionsSheet = FinancialConnectionsSheet(
+            financialConnectionsSessionClientSecret: clientSecret
+        )
+        // FC needs the connected account ID to be configured on the API Client
+        // Make a copy before modifying so we don't unexpectedly modify the shared API client
+        financialConnectionsSheet.apiClient = apiClient.makeCopy()
+        financialConnectionsSheet.apiClient.stripeAccount = connectedAccountId
+        return await withCheckedContinuation { continuation in
```
nschris-stripe, line 28: Just wanted to check to see if we have weighed the pros and cons of checked vs unchecked continuation here. If we stick with checked then this __will__ cause a crash if resume is called more than once. It shouldn't get called more than once in this instance, but maybe we can add some extra protections here to ensure we don't call resume more than once and instead of crashing we log an error.

mats-stripe (approved):
FC integration looks good to me!

StripeConnect/StripeConnect/Source/Internal/FinancialConnectionsPresenter.swift
mludowise-stripe: I didn't spend too much time thinking about it in this instance since it shouldn't be possible for it to get called more than once. I don't think we have a strong opinion as to which to use the SDK.

nschris-stripe: Ok that's ok, I don't think this should be a blocker or anything. Just wanted to call out that we have been bitten by this before. Especially within an SDK it would be ideal to not crash on a programmer error like this.

StripeConnect/StripeConnect/Source/Internal/Webview/ConnectComponentWebViewController.swift
mludowise-stripe: If we end up having that future case, then I agree we should stop configuring default values here since it could lead to issues. But given that we don't have that case today and these are only explicitly set for test injection, I'm not too concerned about it right now.

nschris-stripe (approved):

PR #4164, opened 2024-10-18
Add embedded config to analytics payload

wooj-stripe:
## Summary
Adds configuration for EmbeddedPaymentElement.Configuration to Analytics payload

## Motivation
Need to track

## Testing
manual testing

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElementConfiguration.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElementConfiguration.swift
@@ -141,15 +141,6 @@ extension EmbeddedPaymentElement {
 
             /// The button says “Continue”. When tapped, the form sheet closes.
             case `continue`
-
-            var analyticValue: String {
-                switch self {
-                case .confirm:
-                    return "confirm"
-                case .`continue`:
-                    return "continue"
-                }
-            }
         }
 
         /// The view can display payment methods like “Card” that, when tapped, open a sheet where customers enter their payment method details. The sheet has a button at the bottom. `formSheetAction` controls the action the button performs.
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElementConfiguration.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElementConfiguration.swift
@@ -141,6 +141,15 @@ extension EmbeddedPaymentElement {
 
             /// The button says “Continue”. When tapped, the form sheet closes.
             case `continue`
+
+            var analyticValue: String {
+                switch self {
+                case .confirm:
+                    return "confirm"
+                case .`continue`:
+                    return "continue"
+                }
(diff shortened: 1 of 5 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetAppearance.swift
```diff
@@ -306,5 +308,18 @@ public extension PaymentSheet {
...
+            var payload = [String: Any]()
+            payload["style"] = style != PaymentSheet.Appearance.EmbeddedPaymentElement.default.style
+            payload["row"] = row != PaymentSheet.Appearance.EmbeddedPaymentElement.default.row
+            // Convenience payload item to make querying high level appearance usage easier
+            payload["usage"] = payload.values.contains(where: { value in
+                if let boolValue = value as? Bool {
+                    return boolValue == true
+                }
+                return false
+            })
```
porter-stripe, line 321: Do we need this?

wooj-stripe: I thought it was a common pattern we were using -- happy to remove if we don't want it.

porter-stripe: I think the only place we use this is for top level appearance stuff so we can easily answer "Is appearance being used at all". Peaceful to remove this

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetAppearance.swift
```diff
@@ -306,5 +308,11 @@ public extension PaymentSheet {
                 public var spacing: CGFloat = 12.0
             }
         }
+        var analyticPayload: [String: Any] {
```
yuki-stripe, line 311: Nit but wdyt about defining this and the `static let` above in an extension in `STPAnalyticsClient+PaymentSheet`?

yuki-stripe: @porter-stripe Outside scope of this PR  and embedded too really but I wonder if it'd be more useful to send the actual value of these things.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElementConfiguration.swift
```diff
@@ -141,6 +141,15 @@ extension EmbeddedPaymentElement {
...
             case `continue`
+
+            var analyticValue: String {
+                switch self {
+                case .confirm:
+                    return "confirm"
+                case .`continue`:
+                    return "continue"
+                }
+            }
```
yuki-stripe, line 152: Ah sorry, could we move this to `STPAnalyticsClient+PaymentSheet` too?

wooj-stripe: oops, my bad, thanks!

yuki-stripe (approved):

PR #4182, opened 2024-10-24
Pass email from MPE to Link Login pane

mats-stripe:
## Summary

Due to a backend bug, emails which don't have an existing Link account are not passed to the Financial Connections session manifest yet. This is intended as a quick fix / workaround for that issue while it gets resolved.

## Motivation

Fixes an issue reported in the Instant Debits & Panther bug bash.

## Testing

https://github.com/user-attachments/assets/87c019cf-9197-4cdf-a6b5-e6908c914325

## Changelog

N/a


mats-stripe:
Latest commit also passes the phone number from MPE to the Link Login pane: 

https://github.com/user-attachments/assets/10d68cc4-fcce-4935-9954-bc310c1b2447

StripeFinancialConnections/StripeFinancialConnections/Source/Native/LinkLogin/LinkLoginViewController.swift
```diff
@@ -118,9 +118,12 @@ final class LinkLoginViewController: UIViewController {
         paneLayoutView?.scrollView.keyboardDismissMode = .onDrag
         #endif
 
-        let emailAddress = dataSource.manifest.accountholderCustomerEmailAddress
+        let emailAddress = dataSource.manifest.accountholderCustomerEmailAddress ?? dataSource.elementsSessionContext?.email
         if let emailAddress, !emailAddress.isEmpty {
             formView.prefillEmailAddress(emailAddress)
+
+            let phoneNumber = dataSource.manifest.accountholderPhoneNumber ?? dataSource.elementsSessionContext?.phoneNumber
```
tillh-stripe, line 125: With the current approach, it seems like `elementsSessionContext.phoneNumber` is only being populated if the field was visible in PaymentSheet, even if the merchant provided a default value. Can we change that to also prefill if the field is not displayed?

mats-stripe: >it seems like `elementsSessionContext.phoneNumber` is only being populated if the field was visible in PaymentSheet

`elementsSessionContext.phoneNumber` is either the value entered in the phone number field, or the default value if `attachDefaultsToPaymentMethod` is true. The field does not need to be visible for the latter!

https://github.com/stripe/stripe-ios/blob/c8f86df3b8e4b36c6c145147875a4a46bab9d4bc/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/USBankAccount/InstantDebitsPaymentMethodElement.swift#L89-L96

mats-stripe: Updated to disregard `attachDefaultsToPaymentMethod` here in https://github.com/stripe/stripe-ios/pull/4182/commits/fa43cc638a1f185730c098c2e1ac0fdb067857a1

tillh-stripe (approved):
❤️

PR #4186, opened 2024-10-25
Fix synchronous download manager

yuki-stripe:
## Summary
Deleted the synchronous path of code in DownloadManager; this is unsafe (as the comment indicates!) and can deadlock. 

The purpose of the synchronous code was, I believe, to make the method return the *cached* image by synchronously making a network request that would then access the URLSession cache. The better way to do this is to maintain an in-memory cache, implemented in this PR. 

I tried to make a minimal change here to fix the deadlock; broader refactor effort is needed and ticketed here: https://jira.corp.stripe.com/browse/MOBILESDK-2604

## Testing
See tests

## Changelog
See changelog


davidme-stripe (dismissed):

mats-stripe (approved):
If only we could make `DownloadManager` an `actor` 🥲

yuki-stripe:
> If only we could make `DownloadManager` an `actor` 🥲

Agreed @mats-stripe!  Some discussion here: https://stripe.slack.com/archives/CFA2HJ99A/p1729811584450199?thread_ts=1729805357.152749&cid=CFA2HJ99A

davidme-stripe (approved):

PR #4189, opened 2024-10-25
Fix embedded update preserving invalid paymentoption

yuki-stripe:
## Summary
Previously, embedded would always preserve the selected payment option across `update`, even if it's no longer valid.  This PR fixes that.

Also contains some minor unrelated tweaks.

## Testing
See test

## Changelog
Not user facing

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example-Shard1.xctestplan
+++ b/Example/PaymentSheet Example/PaymentSheet Example-Shard1.xctestplan
@@ -31,6 +31,8 @@
     {
       "parallelizable" : true,
       "skippedTests" : [
+        "EmbeddedUITests",
+        "EmbeddedUITests\/testUpdate()",
         "PaymentSheetBillingCollectionLPMUITests",
         "PaymentSheetBillingCollectionUICardTests",
         "PaymentSheetBillingCollectionUITestCase",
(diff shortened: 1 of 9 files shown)


porter-stripe (dismissed):

Example/PaymentSheet Example/PaymentSheet Example-Shard1.xctestplan
```diff
@@ -31,6 +31,8 @@
     {
       "parallelizable" : true,
       "skippedTests" : [
+        "EmbeddedUITests",
+        "EmbeddedUITests\/testUpdate()",
```
porter-stripe, line 35: Hmm, this doesn't look right. Why is it including the test name `testUpdate` not just the class?

porter-stripe (approved):

Example/PaymentSheet Example/PaymentSheet Example-Shard1.xctestplan
yuki-stripe: IDK, this is what Xcode did. In Xcode it looks the same as other tests:

CleanShot 2024-10-25 at 17 47 30

yuki-stripe: Adding another test to EmbeddedUITest automatically does the expected thing, too - skipped in shards 1, 2, 3 and added to shard 4.

PR #4194, opened 2024-10-28
Fix embedded mandate padding

yuki-stripe:
## Summary
- Fixed mandate having extra horizontal padding. 
- Fixed mandate sometimes having background color. It should be transparent, like its parent view background color, ie the default background color of UIViews.

Simulator Screen Recording -  default tester  iPhone 12 mini 16 1  - 2024-10-28 at 11 59 12

## Testing
See snapshot test + gif

## Changelog
Not user facing


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
```diff
@@ -218,23 +217,31 @@ class EmbeddedPaymentMethodsView: UIView {
...
+        let shouldDisplayMandate: Bool = {
+            guard let mandateText else {
+                return false
+            }
+            return shouldShowMandate && !mandateText.string.isEmpty
+        }()
+        mandateView.attributedText = mandateText
+        let updateMandateUI = {
+            let spacing = shouldDisplayMandate ? 12.0 : 0
+            let mandateViewIndex = self.stackView.arrangedSubviews.firstIndex(of: self.mandateView)! - 1
```
wooj-stripe, line 234: 1. Rather than crashing if the mandate (and the sep isn't there, maybe we should bail, and send an event?

2. nit: the name `mandateViewIndex` isn't that accurate since we are subtracting one from the index.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
```diff
@@ -218,23 +217,31 @@ class EmbeddedPaymentMethodsView: UIView {
...
         }
-
         UIView.animate(withDuration: 0.25, animations: {
-            self.mandateView.setHiddenIfNecessary(
-                (self.mandateView.attributedText?.string.isEmpty ?? true) ||
-                !self.shouldShowMandate
-            )
+            updateMandateUI()
             self.setNeedsLayout()
             self.layoutIfNeeded()
```
wooj-stripe, line 246: I would have expected `setNeedsLayout` and `layoutIfNeeded` to be inside the `updateMandateUI()` closure?

yuki-stripe: I think it makes more sense if `layoutIfNeeded` isn't called if animation = false so that callers can control when exactly the view is laid out.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
yuki-stripe: event is overkill IMO, our tests will exercise this code well, but I'll make this not crash in production

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
wooj-stripe: ah! makes sense!

wooj-stripe (approved):

PR #4200, opened 2024-10-30
Use Xcode Cloud to build SDK releases

davidme-stripe:
## Summary
We set up a "Stripe.framework" project in Xcode Cloud, which will build the new 'ReleaseFrameworksTarget' target on release branches. All this does is run the `export_builds.rb` script and copy the results to the `BUILT_PRODUCTS_DIR`.

The deploy script will no longer build the frameworks locally. Instead, it will fetch the artifact from the Xcode build matching the current release hash and copy it to the build products directory.

This is pretty brittle, but should be fine for our purposes. It'll also break loudly if anything fails.

## Motivation
Allow us to deploy from machines running Sequoia

## Testing
Tested locally. Pushed a `releases/23.test.test` branch, waited for CI to complete, then ran the first step of `deploy_release.rb`.

## Changelog
None


mludowise-stripe (approved):
This is awesome!

bitrise.yml
```diff
@@ -151,6 +151,11 @@ workflows:
         - content: bundle exec ./ci_scripts/deploy_release.rb --dry-run
         is_always_run: true
         title: Deploy release
+    - script@1:
+        inputs:
+        - content: bundle exec ./ci_scripts/export_builds.rb
+        is_always_run: true
+        title: Export builds
```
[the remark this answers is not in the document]
davidme-stripe: Huh, I thought deploy-dry-run *was* nightly! Looks like we're also running it on main branches too, which is incorrect. I'll fix it

bitrise.yml
```diff
@@ -48,7 +48,6 @@ stages:
     - ui-tests-3: {}
     - ui-tests-4: {}
     - integration-all: {}
-    - deploy-dry-run: {}
```
mludowise-stripe, line 51: This also removes it from `releases-trigger-pipeline`. Are we okay with that?

I'm guessing this is okay given that we would be manually executing the deploy for releases, but it would mean that we wouldn't catch any breakages until we attempt to run the deploy scripts as opposed to catching them earlier in CI.

davidme-stripe: I think it was never running in `releases-trigger-pipeline`, it was only running in the main pipeline. We should generally catch release breakages in the nightly build. (And even those are unlikely, as the other builds should catch most breakages)

PR #4204, opened 2024-10-30
SPM for embedded payment element

wooj-stripe:
## Summary
Adds SPM experience to Embedded for CBC and non CBC use cases

A couple things need to be updated, which will be handled in another PR
* Need to use "X" button on sheet when presenting SPM or update sheet (and is at the root)

## Motivation
SPM for embedded payment element

## Testing
Added ui tests

## Changelog
N/A

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
+++ b/Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
@@ -57,27 +57,27 @@ class EmbeddedUITests: PaymentSheetUITestCase {
         // Ensure card preference is cartes bancaires
         XCTAssertTrue(app.buttons["•••• 1001"].waitForExistence(timeout: 3))
         XCTAssertTrue(app.images["stp_card_cartes_bancaires"].waitForExistence(timeout: 3))
-        XCTAssertTrue(app.images["stp_card_visa"].waitForNonExistence(timeout: 1))
+        XCTAssertFalse(app.images["stp_card_visa"].waitForExistence(timeout: 3))
 
         app.buttons["Edit"].waitForExistenceAndTap()
         app.otherElements["Card Brand Dropdown"].waitForExistenceAndTap()
         app.pickerWheels.firstMatch.swipeUp()
         app.buttons["Done"].waitForExistenceAndTap()
         app.buttons["Update"].waitForExistenceAndTap()
-        app.staticTexts["Update card brand"].waitForNonExistence(timeout: 10)
+        XCTAssertFalse(app.staticTexts["Update card brand"].waitForExistence(timeout: 3))
 
         // Ensure card preference is switched to visa
         XCTAssertTrue(app.buttons["•••• 1001"].waitForExistence(timeout: 3))
         XCTAssertTrue(app.images["stp_card_visa"].waitForExistence(timeout: 3))
-        XCTAssertTrue(app.images["stp_card_cartes_bancaires"].waitForNonExistence(timeout: 1))
+        XCTAssertFalse(app.images["stp_card_cartes_bancaires"].waitForExistence(timeout: 3))
 
         // Now remove card
         app.buttons["Edit"].waitForExistenceAndTap()
         app.buttons["Remove card"].waitForExistenceAndTap()
         dismissAlertView(alertBody: "Visa •••• 1001", alertTitle: "Remove card?", buttonToTap: "Remove")
 
-        XCTAssertTrue(app.images["stp_card_visa"].waitForNonExistence(timeout: 1))
-        XCTAssertTrue(app.images["stp_card_cartes_bancaires"].waitForNonExistence(timeout: 1))
+        XCTAssertFalse(app.images["stp_card_visa"].waitForExistence(timeout: 3))
+        XCTAssertFalse(app.images["stp_card_cartes_bancaires"].waitForExistence(timeout: 3))
     }
 
     func testMulipleCardWith_updateCBCWithinViewMore() {
@@ -113,7 +113,7 @@ class EmbeddedUITests: PaymentSheetUITestCase {
         app.buttons["Present embedded payment element"].waitForExistenceAndTap()
 
         XCTAssertTrue(app.buttons["•••• 4242"].waitForExistence(timeout: 3.0))
-        XCTAssertTrue(app.buttons["•••• 1001"].waitForNonExistence(timeout: 1.0))
+        XCTAssertFalse(app.buttons["•••• 1001"].waitForExistence(timeout: 3.0))
 
         // Switch from 4242 to 4444
         app.buttons["View more"].waitForExistenceAndTap()
@@ -129,7 +129,7 @@ class EmbeddedUITests: PaymentSheetUITestCase {
         app.buttons["•••• 1001"].waitForExistenceAndTap()
 
         XCTAssertTrue(app.buttons["•••• 1001"].waitForExistence(timeout: 3.0))
-        XCTAssertTrue(app.buttons["•••• 4242"].waitForNonExistence(timeout: 1.0))
+        XCTAssertFalse(app.buttons["•••• 4242"].waitForExistence(timeout: 3.0))
     }
 
     func testMulipleCard_update_and_remove() {
@@ -165,14 +165,14 @@ class EmbeddedUITests: PaymentSheetUITestCase {
         app.buttons["Present embedded payment element"].waitForExistenceAndTap()
 
         XCTAssertTrue(app.buttons["•••• 4242"].waitForExistence(timeout: 3.0))
-        XCTAssertTrue(app.buttons["•••• 4444"].waitForNonExistence(timeout: 1.0))
+        XCTAssertFalse(app.buttons["•••• 4444"].waitForExistence(timeout: 3.0))
 
         // Switch from 4242 to 4444
         app.buttons["View more"].waitForExistenceAndTap()
         app.buttons["•••• 4444"].waitForExistenceAndTap()
 
-        XCTAssertTrue(app.buttons["•••• 4242"].waitForNonExistence(timeout: 3.0))
-        XCTAssertTrue(app.buttons["•••• 4444"].waitForExistence(timeout: 1.0))
+        XCTAssertFalse(app.buttons["•••• 4242"].waitForExistence(timeout: 3.0))
+        XCTAssertTrue(app.buttons["•••• 4444"].waitForExistence(timeout: 3.0))
 
         // Remove selected 4444 card
         app.buttons["View more"].waitForExistenceAndTap()
@@ -183,15 +183,15 @@ class EmbeddedUITests: PaymentSheetUITestCase {
 
         // Since there is only one PM left (4242), sheet dismisses automatically on tapping Done. Verify that only 4242 exists
         XCTAssertTrue(app.buttons["•••• 4242"].waitForExistence(timeout: 3.0))
-        XCTAssertTrue(app.buttons["•••• 4444"].waitForNonExistence(timeout: 1.0))
+        XCTAssertFalse(app.buttons["•••• 4444"].waitForExistence(timeout: 3.0))
 
         // Remove 4242 & verify
         app.buttons["Edit"].waitForExistenceAndTap()
         app.buttons["CircularButton.Remove"].firstMatch.waitForExistenceAndTap()
         dismissAlertView(alertBody: "Visa •••• 4242", alertTitle: "Remove card?", buttonToTap: "Remove")
 
-        XCTAssertTrue(app.buttons["•••• 4242"].waitForNonExistence(timeout: 3.0))
-        XCTAssertTrue(app.buttons["•••• 4444"].waitForNonExistence(timeout: 1.0))
+        XCTAssertFalse(app.buttons["•••• 4242"].waitForExistence(timeout: 3.0))
+        XCTAssertFalse(app.buttons["•••• 4444"].waitForExistence(timeout: 3.0))
     }
 
     func dismissAlertView(alertBody: String, alertTitle: String, buttonToTap: String) {
--- a/Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
+++ b/Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
@@ -73,9 +73,14 @@ class EmbeddedUITests: PaymentSheetUITestCase {
 
         // Now remove card
         app.buttons["Edit"].waitForExistenceAndTap()
+
+        // Ensure Popup is presented
+        XCTAssertTrue(app.staticTexts["Update card brand"].waitForExistence(timeout: 3.0))
         app.buttons["Remove card"].waitForExistenceAndTap()
         dismissAlertView(alertBody: "Visa •••• 1001", alertTitle: "Remove card?", buttonToTap: "Remove")
 
+        // Ensure popup is implicitly dismissed
+        XCTAssertFalse(app.staticTexts["Update card brand"].waitForExistence(timeout: 3.0))
         XCTAssertFalse(app.images["stp_card_visa"].waitForExistence(timeout: 3))
         XCTAssertFalse(app.images["stp_card_cartes_bancaires"].waitForExistence(timeout: 3))
     }
@@ -83,39 +88,21 @@ class EmbeddedUITests: PaymentSheetUITestCase {
     func testMulipleCardWith_updateCBCWithinViewMore() {
         var settings = PaymentSheetTestPlaygroundSettings.defaultValues()
         settings.mode = .paymentWithSetup
-        settings.uiStyle = .paymentSheet
+        settings.uiStyle = .embedded
         settings.customerKeyType = .legacy
-        settings.customerMode = .new
+        settings.customerMode = .returning
         settings.merchantCountryCode = .FR
         settings.currency = .eur
         settings.applePayEnabled = .on
         settings.apmsEnabled = .off
 
         loadPlayground(app, settings)
-
-        // Add CBC eligible card
-        app.buttons["Present PaymentSheet"].waitForExistenceAndTap()
-        try! fillCardData(app, cardNumber: "4000002500001001", postalEnabled: true)
-        app.buttons["Pay €50.99"].tap()
-        XCTAssertTrue(app.staticTexts["Success!"].waitForExistence(timeout: 10.0))
-
-        reload(app, settings: settings)
-
-        // Add Visa
-        app.buttons["Present PaymentSheet"].waitForExistenceAndTap()
-        app.buttons["+ Add"].waitForExistenceAndTap()
-        try! fillCardData(app, postalEnabled: true)
-        app.buttons["Pay €50.99"].tap()
-        XCTAssertTrue(app.staticTexts["Success!"].waitForExistence(timeout: 10.0))
-
-        // Switch to embedded mode kicks off a reload
-        app.buttons["embedded"].waitForExistenceAndTap(timeout: 5.0)
         app.buttons["Present embedded payment element"].waitForExistenceAndTap()
 
-        XCTAssertTrue(app.buttons["•••• 4242"].waitForExistence(timeout: 3.0))
-        XCTAssertFalse(app.buttons["•••• 1001"].waitForExistence(timeout: 3.0))
+        XCTAssertTrue(app.buttons["•••• 1001"].waitForExistence(timeout: 3.0))
+        XCTAssertFalse(app.buttons["•••• 4242"].waitForExistence(timeout: 3.0))
 
-        // Switch from 4242 to 4444
+        // Switch from 1001 to 4242
         app.buttons["View more"].waitForExistenceAndTap()
         app.buttons["Edit"].waitForExistenceAndTap()
         app.buttons["CircularButton.Edit"].waitForExistenceAndTap()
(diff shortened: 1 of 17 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
```diff
@@ -195,7 +196,9 @@ public final class EmbeddedPaymentElement {
     internal private(set) var embeddedPaymentMethodsView: EmbeddedPaymentMethodsView
     internal private(set) var loadResult: PaymentSheetLoader.LoadResult
     internal private(set) var latestUpdateTask: Task<UpdateResult, Never>?
-    private let analyticsHelper: PaymentSheetAnalyticsHelper
+    internal let analyticsHelper: PaymentSheetAnalyticsHelper
+    internal lazy var paymentHandler: STPPaymentHandler = { STPPaymentHandler(apiClient: configuration.apiClient) }()
+    internal var savedPaymentMethods: [STPPaymentMethod]
```
yuki-stripe, line 201: I'm wary of having two sources of truth for savedPaymentMethods - this and `loadResult.savedPaymentMethods`.  Maybe we should remove `var loadResult` and replace with the `var intent, var elementSession` etc.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
```diff
@@ -213,7 +205,84 @@ class EmbeddedPaymentMethodsView: UIView {
...
     func didTapAccessoryButton() {
-        // TODO(porter)
+        guard let savedPaymentMethod = savedPaymentMethods.first else {
+            stpAssertionFailure("Tapped on accessory button but there are no saved payment methods")
+            return
+        }
+        delegate?.presentSavedPaymentMethods(selectedSavedPaymentMethod: savedPaymentMethod)
+    }
+
+    func updateSavedPaymentMethods(_ savedPaymentMethods: [STPPaymentMethod],
```
yuki-stripe, line 215: I think this function should be `updateSavedPaymentMethodRow(savedPaymentMethod: STPPaymentMethod?, ...)`

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
```diff
@@ -213,7 +205,84 @@ class EmbeddedPaymentMethodsView: UIView {
...
+                                   userSelectedPaymentMethod: STPPaymentMethod?,
```
yuki-stripe, line 216: What is `userSelectedPaymentMethod`?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
```diff
@@ -213,7 +205,84 @@ class EmbeddedPaymentMethodsView: UIView {
...
+                }
+            }
+
+            // Update instance states
+            self.savedPaymentMethods = savedPaymentMethods
+            self.savedPaymentMethodButton = nil
+        }
+    }
+
+    func createSavedPaymentMethodButton(savedPaymentMethod: STPPaymentMethod,
```
yuki-stripe, line 266:
```suggestion
    func makeSavedPaymentMethodButton(savedPaymentMethod: STPPaymentMethod,
```
nit per https://www.swift.org/documentation/api-design-guidelines/  (search "make")

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -93,6 +93,120 @@ extension EmbeddedPaymentElement: EmbeddedPaymentMethodsViewDelegate {
...
+                                                                appearance: configuration.appearance,
+                                                                hostedSurface: .paymentSheet,
+                                                                canRemoveCard: configuration.allowsRemovalOfLastSavedPaymentMethod && elementsSession.allowsRemovalOfPaymentMethodsForPaymentSheet(),
+                                                                isTestMode: configuration.apiClient.isTestmode,
+                                                                cardBrandFilter: configuration.cardBrandFilter)
+            updateViewController.delegate = self
+            let bottomSheetVC = BottomSheetViewController(contentViewController: updateViewController,
+                                                          appearance: configuration.appearance,
+                                                          isTestMode: configuration.apiClient.isTestmode, didCancelNative3DS2: {
+                self.paymentHandler.cancel3DS2ChallengeFlow()
```
yuki-stripe, line 113: I think it makes more sense to no-op or assertionFailure or something here, why would update invoke 3ds2?

yuki-stripe: (It's very stinky that BottomSheetVC has this parameter, I have a ticket somewhere to refactor this)

yuki-stripe (commented):
Leaving some early feedback, will review more later!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
wooj-stripe: Just want to verify -- by removing loadResult and replacing with intent & elementSession, it implies making an initial copy of the saved payment methods from the load result into the proposed `var savedPaymentMethods`, then updating it as we add/remove/update payment methods?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
wooj-stripe: I may need to re-think this. I was using it to decide whether or not to update the selected payment method.  I'll revisit the code and figure out if this is needed.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
wooj-stripe: If we changed this function to only pass in a single savedPaymentMethod, I'm not quite sure how we'd make it work if we update to multiple items in the list.  For example:
1. You have one CBC card and one regular card, then present Embedded Payment Element.
2. Tap view more on the SPM to reveal the `VerticalSavedPaymentMethodsViewController`
3. Tap edit
4. Update first card to switch the cbc
5. Remove the regular card.
6. Hit back to reveal the `EmbeddedPaymentMethodsView`

At this point, we've updated one card's CBC, and removed another card, and then as soon as we hit 'back', the code calls "VerticalSavedPaymentMethodsViewControllerDelegate.didComplete" with `latestPaymentMethods: [STPPaymentMethods]` which contains both of those changes.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
wooj-stripe: I believe this is needed for the use case of when a user selects a saved payment method, and we want to update the main view so that it updates the selection.  It was incorrectly being passed in for the 'update' case, which has been updated.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
yuki-stripe: Yeah that's the idea. Don't feel strongly, maybe it's worse in some other way?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
yuki-stripe: I'm not understanding the example but I'm not seeing where EmbeddedPaymentMethodsView uses savedPaymentMethods in the code and I wouldn't expect it to need more than the one optional saved payment method that it displays.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
wooj-stripe: I think your original proposal of storing savedPaymentMethods & elementsSessions separately is better, as I just noticed that the `update()` code overrides the loadResult.  Instead of `update()` setting `loadResult` we can have `update()` directly set `savedPaymentMethods`, `elementsSessions` and any other items that come from loadResult.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -93,6 +93,119 @@ extension EmbeddedPaymentElement: EmbeddedPaymentMethodsViewDelegate {
...
+            isFirstCardCoBranded: savedPaymentMethods.first?.isCoBrandedCard ?? false,
+            isCBCEligible: elementsSession.isCardBrandChoiceEligible,
+            allowsRemovalOfLastSavedPaymentMethod: configuration.allowsRemovalOfLastSavedPaymentMethod,
+            allowsPaymentMethodRemoval: elementsSession.allowsRemovalOfPaymentMethodsForPaymentSheet()
+        )
+    }
+}
+
+extension EmbeddedPaymentElement: VerticalSavedPaymentMethodsViewControllerDelegate {
+    nonisolated func didComplete(viewController: VerticalSavedPaymentMethodsViewController,
```
yuki-stripe, line 191: same here, `@MainActor`

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
wooj-stripe: Sorry, I think I was looking at a function that belonged to EmbeddedPaymentElement, not EmbeddedPaymentMethodsView.

Yes, we can get away with only passing in a single savedPaymentMethod to this view!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
```diff
@@ -213,7 +205,84 @@ class EmbeddedPaymentMethodsView: UIView {
...
+            if selection?.isSaved ?? false || userSelectedPaymentMethod != nil {
+                self.stackView.arrangedSubviews.forEach { view in
+                    (view as? RowButton)?.isSelected = false
+                }
+                updatedSavedPaymentMethodButton.isSelected = true
+
+                // When selection is set, updateMandate() should be called without animation to
+                // avoid animation jank when adding and removing items within the stack view
+                self.animateUpdateMandate = false
+                self.selection = .saved(paymentMethod: savedPaymentMethod)
```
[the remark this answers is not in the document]
wooj-stripe: Resolved offline

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
```diff
@@ -213,7 +205,84 @@ class EmbeddedPaymentMethodsView: UIView {
...
+        delegate?.presentSavedPaymentMethods(selectedSavedPaymentMethod: savedPaymentMethod)
+    }
+
+    func updateSavedPaymentMethods(_ savedPaymentMethods: [STPPaymentMethod],
+                                   userSelectedPaymentMethod: STPPaymentMethod?,
+                                   savedPaymentMethodAccessoryType: RowButton.RightAccessoryButton.AccessoryType?) {
+        guard let previousSavedPaymentMethodButton = self.savedPaymentMethodButton,
+              let viewIndex = stackView.arrangedSubviews.firstIndex(of: previousSavedPaymentMethodButton) else {
+            return
+        }
```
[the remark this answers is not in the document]
wooj-stripe: resolved offline

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
```diff
@@ -14,14 +14,17 @@ import UIKit
...
     func selectionDidUpdate()
+    func presentSavedPaymentMethods(selectedSavedPaymentMethod: STPPaymentMethod?)
 }
 
 /// The view for an embedded payment element
 class EmbeddedPaymentMethodsView: UIView {
 
     typealias Selection = VerticalPaymentMethodListSelection // TODO(porter) Maybe define our own later
 
+    private var savedPaymentMethod: STPPaymentMethod?
```
yuki-stripe, line 25: Per the dead code detection bot, this appears assigned but never used.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
```diff
@@ -193,9 +195,11 @@ public final class EmbeddedPaymentElement {
 
     internal private(set) var containerView: EmbeddedPaymentElementContainerView
     internal private(set) var embeddedPaymentMethodsView: EmbeddedPaymentMethodsView
-    internal private(set) var loadResult: PaymentSheetLoader.LoadResult
+    internal private(set) var elementsSession: STPElementsSession
+    internal private(set) var intent: Intent
     internal private(set) var latestUpdateTask: Task<UpdateResult, Never>?
-    private let analyticsHelper: PaymentSheetAnalyticsHelper
+    internal let analyticsHelper: PaymentSheetAnalyticsHelper
+    internal var savedPaymentMethods: [STPPaymentMethod]
```
yuki-stripe, line 202: should this be `private(set)`?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
```diff
@@ -213,7 +202,73 @@ class EmbeddedPaymentMethodsView: UIView {
...
-        // TODO(porter)
+        delegate?.presentSavedPaymentMethods(selectedSavedPaymentMethod: selection?.savedPaymentMethod)
+    }
+
+    func updateSavedPaymentMethodRow(_ savedPaymentMethod: STPPaymentMethod?,
+                                     isSelected: Bool,
+                                     accessoryType: RowButton.RightAccessoryButton.AccessoryType?) {
+        guard let previousSavedPaymentMethodButton = self.savedPaymentMethodButton,
+              let viewIndex = stackView.arrangedSubviews.firstIndex(of: previousSavedPaymentMethodButton) else {
+            stpAssertionFailure("Function supports only when there is an existing savedPaymentMethodButton")
```
yuki-stripe, line 213: Can we write out why? I think that would be most helpful for future readers.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
```diff
@@ -213,7 +202,73 @@ class EmbeddedPaymentMethodsView: UIView {
...
+            return
+        }
+
+        if let savedPaymentMethod {
+            // Replace saved payment method button at same index
+            let updatedSavedPaymentMethodButton = makeSavedPaymentMethodButton(savedPaymentMethod: savedPaymentMethod,
+                                                                               savedPaymentMethodAccessoryType: accessoryType)
+            // If the previous item was as selected spm or the user explicitly selected a spm, then auto-select saved pm
+            if selection?.isSaved ?? false || isSelected {
```
yuki-stripe, line 222: Why isn't this just `if isSelected {`?  

Is it intentional that this method can select the PM even when you pass `isSelected=false`?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
wooj-stripe: analyticsHelper can be `private(set)` but savedPaymentMethods can not be since we are updating it in the +Internal extension.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
wooj-stripe: 🤔 -- indeed, we can get away with not keeping it as an instance variable

Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
```diff
@@ -28,4 +28,177 @@ class EmbeddedUITests: PaymentSheetUITestCase {
...
+        app.otherElements["Card Brand Dropdown"].waitForExistenceAndTap()
+        app.pickerWheels.firstMatch.swipeUp()
+        app.buttons["Done"].waitForExistenceAndTap()
+        app.buttons["Update"].waitForExistenceAndTap()
+        XCTAssertFalse(app.staticTexts["Update card brand"].waitForExistence(timeout: 3))
+
+        // Ensure card preference is switched to visa
+        XCTAssertTrue(app.buttons["•••• 1001"].waitForExistence(timeout: 3))
+        XCTAssertTrue(app.images["stp_card_visa"].waitForExistence(timeout: 3))
+        XCTAssertFalse(app.images["stp_card_cartes_bancaires"].waitForExistence(timeout: 3))
```
yuki-stripe, line 72: We should hit done/back and make sure that the embedded view correctly updates. I'm not seeing this behavior in the branch.

Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
```diff
@@ -28,4 +28,177 @@ class EmbeddedUITests: PaymentSheetUITestCase {
...
+        XCTAssertTrue(app.staticTexts["Success!"].waitForExistence(timeout: 10.0))
+
+        reload(app, settings: settings)
+
+        // Add Visa
+        app.buttons["Present PaymentSheet"].waitForExistenceAndTap()
+        app.buttons["+ Add"].waitForExistenceAndTap()
+        try! fillCardData(app, postalEnabled: true)
+        app.buttons["Pay €50.99"].tap()
+        XCTAssertTrue(app.staticTexts["Success!"].waitForExistence(timeout: 10.0))
```
yuki-stripe, line 109: Can we use the returning customer now?

Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
```diff
@@ -28,4 +28,177 @@ class EmbeddedUITests: PaymentSheetUITestCase {
...
+        XCTAssertFalse(app.images["stp_card_cartes_bancaires"].waitForExistence(timeout: 3))
+
+        // Now remove card
+        app.buttons["Edit"].waitForExistenceAndTap()
+        app.buttons["Remove card"].waitForExistenceAndTap()
+        dismissAlertView(alertBody: "Visa •••• 1001", alertTitle: "Remove card?", buttonToTap: "Remove")
+
+        XCTAssertFalse(app.images["stp_card_visa"].waitForExistence(timeout: 3))
+        XCTAssertFalse(app.images["stp_card_cartes_bancaires"].waitForExistence(timeout: 3))
+    }
```
yuki-stripe, line 81: We should validate that the form is dismissed and the embedded view is updated to not have a saved pm row.

Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
```diff
@@ -28,4 +28,177 @@ class EmbeddedUITests: PaymentSheetUITestCase {
...
+        XCTAssertTrue(app.staticTexts["Success!"].waitForExistence(timeout: 10.0))
+
+        // Switch to embedded mode kicks off a reload
+        app.buttons["embedded"].waitForExistenceAndTap(timeout: 5.0)
+        app.buttons["Present embedded payment element"].waitForExistenceAndTap()
+
+        XCTAssertTrue(app.buttons["•••• 4242"].waitForExistence(timeout: 3.0))
+        XCTAssertFalse(app.buttons["•••• 1001"].waitForExistence(timeout: 3.0))
+
+        // Switch from 4242 to 4444
```
yuki-stripe, line 118: Is this a typo? I think we're switching from Cartes Bancaires to Visa

Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
```diff
@@ -28,4 +28,177 @@ class EmbeddedUITests: PaymentSheetUITestCase {
...
+        XCTAssertTrue(app.staticTexts["Success!"].waitForExistence(timeout: 10.0))
+
+        reload(app, settings: settings)
+
+        // Add visa
+        app.buttons["Present PaymentSheet"].waitForExistenceAndTap()
+        app.buttons["+ Add"].waitForExistenceAndTap()
+        try! fillCardData(app, postalEnabled: true)
+        app.buttons["Pay $50.99"].tap()
+        XCTAssertTrue(app.staticTexts["Success!"].waitForExistence(timeout: 10.0))
```
yuki-stripe, line 161: Can we use the returning Customer here too? That way we can also test that the saved US Bank Account shows mandate when selected.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
wooj-stripe: It is intentional, assuming that this is the expected behavior

Consider the scenario:
1. User has a single CBC enabled card
2. User selects the single CBC card, putting in a selected state
3. User taps edit (UpdateCardViewController is presented) 
4. User Switches from Visa -> CB (or CB->VISA), then taps update, UpdateCardViewController is dismissed

Expected Result:
The updated single CBC card should preserve the selected state (which happened in step 2). 

If we don't check `selection?.isSaved`, then the updated card will be unselected.

Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
wooj-stripe: I'm not following your comment.

In this test, the UpdateCardViewController is presented, and CBC is updated. On line 66, update is tapped, and the sheet dismisses.  What behavior are you not seeing in the branch?

Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
wooj-stripe: Added `app.staticTexts["Update card brand"].waitForExistence`

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
```diff
@@ -210,7 +208,10 @@ class EmbeddedPaymentMethodsView: UIView {
                                      accessoryType: RowButton.RightAccessoryButton.AccessoryType?) {
         guard let previousSavedPaymentMethodButton = self.savedPaymentMethodButton,
               let viewIndex = stackView.arrangedSubviews.firstIndex(of: previousSavedPaymentMethodButton) else {
-            stpAssertionFailure("Function supports only when there is an existing savedPaymentMethodButton")
+            stpAssertionFailure("""
+            Function supports only when there is an existing savedPaymentMethodButton because this function is only used
+            for updating the view after selecting from a larger list of saved payment methods or updating a saved payment method.
```
yuki-stripe, line 213: "This function should never be called when there isn't already a saved PM row because there's no way for Embedded to _add_ a saved payment method today; you can only update or remove them" - something like this makes sense to me!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
yuki-stripe: This seems like a code smell; `isSelected` doesn't do what you'd expect as a caller (it does not necessarily select the PM).

Maybe we could either:
- Rename `isSelected` to something that better matches its behavior.
- Rework so that `isSelected` has the expected behavior. Maybe the caller can pass `isSelected: true` in the scenario above?

yuki-stripe (approved):

PR #4205, opened 2024-10-31
Add internal "Native Link" switch

davidme-stripe:
## Summary
* Add a client-side switch to enable Link.
* Connect the switch to an internal PaymentSheet config var.
* Move some existing Link-related protocol implementations from PaymentSheet.swift to PaymentSheet+Link.swift.

## Motivation
Enabling testing of native Link

## Testing
Tested in PS Example

## Changelog
No user-facing changes


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+Link.swift
```diff
@@ -0,0 +1,163 @@
...
+extension PaymentSheet: PayWithLinkWebControllerDelegate {
+
+    func payWithLinkWebControllerDidComplete(
+        _ PayWithLinkWebController: PayWithLinkWebController,
+        intent: Intent,
+        elementsSession: STPElementsSession,
+        with paymentOption: PaymentOption
+    ) {
+        let backgroundColor = self.configuration.appearance.colors.background.withAlphaComponent(0.85)
+        self.bottomSheetViewController.addBlurEffect(animated: false, backgroundColor: backgroundColor) {
```
wooj-stripe, line 21: Seems like we should find paymentSheetViewController before presenting a blur effect, otherwise, users may end up in a bad UI state.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
```diff
@@ -193,6 +193,8 @@ extension PaymentSheet {
         // MARK: Internal
         internal var linkPaymentMethodsOnly: Bool = false
 
+        @_spi(STP) public var forceEnableNativeLink: Bool = false
```
wooj-stripe, line 196: nit: is "force" really necessary?   Maybe `nativeLinkEnabled`.. but if we do want the word 'force', then maybe `forceNativeLinkEnabled`

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
```diff
@@ -315,11 +315,22 @@ extension PaymentSheet: PaymentSheetViewControllerDelegate {
...
-            intent: paymentSheetViewController.intent,
-            elementsSession: paymentSheetViewController.elementsSession
-        )
+        if configuration.forceEnableNativeLink {
+            self.presentPayWithNativeLinkController(
+                from: paymentSheetViewController,
+                intent: paymentSheetViewController.intent,
+                elementsSession: paymentSheetViewController.elementsSession,
+                shouldOfferApplePay: false,
+                shouldFinishOnClose: true,
```
wooj-stripe, line 324: `presentPayWithNativeLinkController` is only being called here. There doesn't seem to be a need to pass in `shouldOfferApplePay` nor `shouldFinishOnClose`

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
davidme-stripe: This is an override switch for testing, it's going away once the server-side behavior is implemented.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
davidme-stripe: I'm working on another PR right now that sets these to the opposite values — if we're launching directly into Link before PS, then we do want to offer Apple Pay.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+Link.swift
davidme-stripe: Fixed!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+Link.swift
```diff
@@ -0,0 +1,166 @@
...
+private func findPaymentSheetViewController() -> PaymentSheetViewController? {
+    for vc in bottomSheetViewController.contentStack {
+        if let paymentSheetVC = vc as? PaymentSheetViewController {
+            return paymentSheetVC
+        }
+    }
+
+    return nil
+}
+}
```
wooj-stripe, line 166: Spacing looks off here.

wooj-stripe (approved):

Issue #4195, opened 2024-10-28
ERROR An error occurred in PaymentSheet.There is a mismatch between the payment method ID on your Intent: pm_XXX and the payment method passed into the `confirmHandler`: pm_YYY


Ariandr:
## Summary
We are migrating from Basic Integration to new Payment Sheet, using iOS and backend with payment intents. We use Stripe connect where we save the payment methods on the main platform account and then share these payment methods to make the Direct payments to Connected accounts.

With the Basic Integration (and with the PaymentSheet August version of SDK) it worked perfectly for us, but with the new Payment Sheet releases (after https://github.com/stripe/stripe-ios/pull/3972) there is an issue that we cannot confirm it on the iOS side later due to a problem that the initial payment method doesn't match the one inside the payment intent.

https://docs.stripe.com/connect/direct-charges-multiple-accounts

## iOS version
iOS 18, but not related to iOS version

## Installation method
Cocoapods

## SDK version
23.32

## Other information
Essentially, our flow it this:
1) Use the PaymentSheet.FlowController to select a payment method
2) Send the request to the backend with the selected paymentMethod.stripeId
3) Backend shares this payment method to a connected account and puts the shared one inside the payment intent. Then, the backend tries to confirm the payment intent.
4) Then backend sends the payment intent secret back to iOS
5) On iOS, we are tying to call  intentCreationCallback?(.success(clientSecret))
6) If the intent requires next action (e.g. 3DSecure), we get this `ERROR An error occurred in PaymentSheet.There is a mismatch between the payment method ID on your Intent: pm_XXX and the payment method passed into the `confirmHandler`: pm_YYY`

With Basic Integration and older PaymentSheet versions this action is handled without issues and we confirm the payment with 3DSecure.
But with the new PaymentSheet releases this got broken.

I saw you added this check in this pull request, but it breaks our use case unfortunately.
https://github.com/stripe/stripe-ios/pull/3972

It works with the older (August) version of SDK, e.g. 23.29.2. Can we add a way (maybe additional configuration option) to omit this check, so our use case is not broken?

yuki-stripe:
Hello @Ariandr, sorry you're running into this. How are you "sharing" the payment method? In the integration you linked (https://docs.stripe.com/connect/direct-charges-multiple-accounts), cloning happens *after* the confirmation succeeds and you shouldn't run into this issue.

Ariandr:
@yuki-stripe 
To make it simpler, probably the steps will be easier:
[...]

PR #4207, opened 2024-10-31
Added fingerprint check validation for card and us bank

joyceqin-stripe:
## Summary

Added fingerprint checking logic in deferred validation for card and us_bank_account if PM ids don't match

## Motivation

https://github.com/stripe/stripe-ios/issues/4195
https://jira.corp.stripe.com/browse/RUN_MOBILESDK-3670


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetDeferredValidator.swift
```diff
@@ -48,12 +49,41 @@ struct PaymentSheetDeferredValidator {
         }
         try validatePaymentMethodId(intentPaymentMethod: setupIntent.paymentMethod, paymentMethod: paymentMethod)
     }
-    
+
     static func validatePaymentMethodId(intentPaymentMethod: STPPaymentMethod?, paymentMethod: STPPaymentMethod) throws {
         guard let intentPaymentMethod = intentPaymentMethod else { return }
         guard intentPaymentMethod.stripeId == paymentMethod.stripeId else {
+            if intentPaymentMethod.type == paymentMethod.type {
+                switch paymentMethod.type.identifier {
```
yuki-stripe, line 57: We should add a comment explaining why we're validating fingerprints if the id is not equal so that future readers know about the Connect edge case.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetDeferredValidator.swift
```diff
@@ -48,12 +49,41 @@ struct PaymentSheetDeferredValidator {
         }
         try validatePaymentMethodId(intentPaymentMethod: setupIntent.paymentMethod, paymentMethod: paymentMethod)
     }
-    
+
     static func validatePaymentMethodId(intentPaymentMethod: STPPaymentMethod?, paymentMethod: STPPaymentMethod) throws {
```
yuki-stripe, line 53: Now that this no longer only validates the payment method id, we should rename it to match its new behavior.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetDeferredValidator.swift
```diff
@@ -48,12 +49,41 @@ struct PaymentSheetDeferredValidator {
...
+                }
+            }
             let errorMessage = """
                 \nThere is a mismatch between the payment method ID on your Intent: \(intentPaymentMethod.stripeId) and the payment method passed into the `confirmHandler`: \(paymentMethod.stripeId).
+            
+                To resolve this issue, you can:
+                1. Create a new Intent each time before you call the `confirmHandler`, or
+                2. Update the existing Intent with the desired `paymentMethod` before calling the `confirmHandler`.
+            """
+            let errorAnalytic = ErrorAnalytic(event: .paymentSheetDeferredIntentPaymentMethodIdMismatch, error: PaymentSheetError.unknown(debugDescription: errorMessage))
```
yuki-stripe, line 75: We send this "...IDMismatch" event for validating fingerprints below, which isn't quite right.

yuki-stripe (commented):
Nice!

StripeCore/StripeCore/Source/Analytics/STPAnalyticEvent.swift
```diff
@@ -217,6 +217,7 @@ import Foundation
     case paymentSheetFormCompleted = "mc_form_completed"
     case paymentSheetCardNumberCompleted = "mc_card_number_completed"
     case paymentSheetDeferredIntentPaymentMethodIdMismatch = "mc_deferred_intent_payment_method_id_mismatch"
+    case paymentSheetDeferredIntentPaymentMethodFingerprintMismatch = "mc_deferred_intent_payment_method_fingerprint_mismatch"
```
yuki-stripe, line 220: Let's just combine these into a single `paymentSheetDeferredIntentPaymentMethodMismatch` event and send more details in the analytic payload to distinguish between the specific failures (id mismatch vs fingerprint mismatch) by supplying `additionalNonPIIParams` in the `ErrorAnalytic` initializer. 

Whenever we change analytics we need to make sure we update relevant queries / make sure we don't break existing queries but in this case I don't think we specifically alert/search for this today.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetDeferredValidator.swift
```diff
@@ -46,24 +47,55 @@ struct PaymentSheetDeferredValidator {
...
     }
-    
-    static func validatePaymentMethodId(intentPaymentMethod: STPPaymentMethod?, paymentMethod: STPPaymentMethod) throws {
+
+    static func validatePaymentMethod(intentPaymentMethod: STPPaymentMethod?, paymentMethod: STPPaymentMethod) throws {
         guard let intentPaymentMethod = intentPaymentMethod else { return }
         guard intentPaymentMethod.stripeId == paymentMethod.stripeId else {
+            if intentPaymentMethod.type == paymentMethod.type {
+                // Payment methods of type card and us_bank_account can be cloned, leading to mismatched pm ids, but their fingerprints should still match
+                switch paymentMethod.type.identifier {
```
yuki-stripe, line 58: Why do we switch on `identifier` instead of `type`?  Ideally we can switch on specific enum cases to avoid potential String typos

yuki-stripe (commented):
Let's add a CHANGELOG entry as well

yuki-stripe (dismissed):

yuki-stripe (commented):
Oh I just realized this is untested right? Could we add a unit test?

yuki-stripe (approved):
Excellent tests, thank you

PR #4208, opened 2024-10-31
Fix ui-tests

joyceqin-stripe:
## Summary

Some UI tests were failing while waiting for a button that didn't exist, and it's because it was expecting the horizontal layout but found the vertical layout instead. Also increased the timeout for one case.

## Motivation

https://app.bitrise.io/build/bdbb441b-becc-4c63-80b8-c394f2e3d2f3?tab=tests
https://app.bitrise.io/build/40c55a29-1cd9-4a90-b653-6188c5edad52?tab=tests

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
+++ b/Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
@@ -1638,7 +1638,7 @@ class PaymentSheetCVCRecollectionUITests: PaymentSheetUITestCase {
         payButton.tap()
 
         let successText = app.staticTexts["Success!"]
-        XCTAssertTrue(successText.waitForExistence(timeout: 10.0))
+        XCTAssertTrue(successText.waitForExistence(timeout: 60.0))
 
         // Reload w/ same customer
         reload(app, settings: settings)
--- a/Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
+++ b/Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
@@ -1638,7 +1638,7 @@ class PaymentSheetCVCRecollectionUITests: PaymentSheetUITestCase {
         payButton.tap()
 
         let successText = app.staticTexts["Success!"]
-        XCTAssertTrue(successText.waitForExistence(timeout: 60.0))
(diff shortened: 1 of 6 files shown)


Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
```diff
@@ -1638,7 +1638,7 @@ class PaymentSheetCVCRecollectionUITests: PaymentSheetUITestCase {
         payButton.tap()
 
         let successText = app.staticTexts["Success!"]
-        XCTAssertTrue(successText.waitForExistence(timeout: 10.0))
```
wooj-stripe, line 1641: 10 seconds already feels like a long time, and hesitant to bump up to 60s.

Did we increase the time because we thought this test was "flakey?"  If so, my guess of why it is flake is due to not using `waitForExistanceAndTap()` on `app.buttons["Pay $50.99"]`

joyceqin-stripe: It was for a flaky test, but I think I actually changed the timeout length in a place I didn't mean to, so I'll revert that. Thanks for catching it!

joyceqin-stripe: The actual thing that it was flaky on was [this line](https://github.com/stripe/stripe-ios/blob/master/Example/PaymentSheet%20Example/PaymentSheetUITest/XCUITest%2BUtilities.swift#L311)

joyceqin-stripe: Do you think it's worth extending the timeout on that one?

wooj-stripe: No, I dont think we should update that-- Do you remember which test was flakey?  Maybe there is something we were doing before calling reload which made reload flakey?

wooj-stripe (approved):

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
joyceqin-stripe: It was testCVCRecollectionComplete_deferredCSC()

PR #4211, opened 2024-10-31
Update navigation icons for SPM on EPE

wooj-stripe:
## Summary
Update navigation icons for SPM on EPE

## Motivation
We should be showing a close button when presenting from embedded

## Testing
Added, and updated snapshot tests


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/Vertical Saved Payment Method Screen/VerticalSavedPaymentMethodsViewController.swift
```diff
@@ -206,6 +206,16 @@ class VerticalSavedPaymentMethodsViewController: UIViewController {
         minHeightConstraint.isActive = true
     }
 
+    private func navigationBarStyle() -> SheetNavigationBar.Style {
+        if let bottomSheet = self.bottomSheetController,
+           bottomSheet.contentStack.first is VerticalSavedPaymentMethodsViewController,
```
yuki-stripe, line 211: Interesting, ooc why do we need this check?

yuki-stripe (approved):
The whole navigation bar architecture is a bit funky; these VCs shouldn't be doing all this repeated logic to determine back button vs close. Something to refactor another day though.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/Vertical Saved Payment Method Screen/VerticalSavedPaymentMethodsViewController.swift
wooj-stripe: AH! Not needed!  I must have added it for safety, but is not necessary. thank you!

yuki-stripe (approved):

PR #4212, opened 2024-11-01
Fix `LinkPaymentControllerUITest`, add tests for native & web flow

mats-stripe:
## Summary

This fixes the recently disabled `LinkPaymentControllerUITest`, which were failing since we switched to using the native instant debits flow. It now has a test for the web flow, and a test for the native flow.

## Motivation

Restores a disabled test

## Testing

✅

## Changelog

N/a


tillh-stripe (dismissed):
❤️

tillh-stripe (approved):

PR #4219, opened 2024-11-04
Link v3: Connect passthrough mode

davidme-stripe:
## Summary
If passthrough mode is enabled, call the `/share` endpoint to get the passthrough PM, then confirm using that PM.

## Motivation
Enable passthrough mode in native Link.

## Testing
In PaymentSheet Example.

## Changelog
None

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
@@ -480,8 +480,23 @@ extension PaymentSheet {
             case .withPaymentMethod(let paymentMethod):
                 confirmWithPaymentMethod(paymentMethod, nil, false)
             case .withPaymentDetails(let linkAccount, let paymentDetails):
-                // shouldSave is false, as we don't show a save checkbox in the Link VC
-                confirmWithPaymentDetails(linkAccount, paymentDetails, paymentDetails.cvc, false)
+                let shouldSave = false // always false, as we don't show a save-to-merchant checkbox in Link VC
+
+                if elementsSession.linkPassthroughModeEnabled {
+                    linkAccount.sharePaymentDetails(id: paymentDetails.stripeID, cvc: paymentDetails.cvc) { result in
+                        switch result {
+                        case .success(let paymentDetailsShareResponse):
+                            // shouldSave is false, as we don't show a save checkbox in the Link VC
+                            confirmWithPaymentMethod(paymentDetailsShareResponse.paymentMethod, linkAccount, shouldSave)
+                        case .failure(let error):
+                            STPAnalyticsClient.sharedClient.logLinkSharePaymentDetailsFailure(error: error)
+                            paymentHandlerCompletion(.failed, error as NSError)
+                        }
+                    }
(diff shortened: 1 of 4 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
```diff
@@ -454,8 +454,21 @@ extension PaymentSheet {
             case .withPaymentMethod(let paymentMethod):
                 confirmWithPaymentMethod(paymentMethod, nil, false)
             case .withPaymentDetails(let linkAccount, let paymentDetails):
-                // shouldSave is false, as we don't show a save checkbox in the Link VC
-                confirmWithPaymentDetails(linkAccount, paymentDetails, paymentDetails.cvc, false)
+                let shouldSave = false // always false, as we don't show a save-to-merchant checkbox in Link VC
+
+                if elementsSession.linkPassthroughModeEnabled {
+                    linkAccount.sharePaymentDetails(id: paymentDetails.stripeID, cvc: paymentDetails.cvc) { result in
+                        switch result {
```
wooj-stripe, line 461: Seems like we should add a test to validate that these functions are being called when `linkPassthroughModeEnabled` == true.

wooj-stripe (approved):

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheet+APIMockTest.swift
```diff
@@ -0,0 +1,284 @@
...
+            configuration: configuration,
+            authenticationContext: self,
+            intent: .deferredIntent(intentConfig: MockParams.deferredPaymentIntentConfiguration(clientSecret: MockParams.paymentIntentClientSecret)),
+            elementsSession: elementsSession,
+            paymentOption: .link(option: .withPaymentDetails(account: .init(email: "test@example.com", session: .init(clientSecret: "cs_xxx", emailAddress: "test@example.com", redactedPhoneNumber: "+1-555-xxx-xxxx", verificationSessions: [.init(type: .sms, state: .verified)], supportedPaymentDetailsTypes: [.card]), publishableKey: MockParams.publicKey), paymentDetails: .init(stripeID: "pd1", details: .card(card: .init(expiryYear: 2055, expiryMonth: 12, brand: "visa", last4: "1234", checks: nil)), billingAddress: nil, billingEmailAddress: nil, isDefault: true))),
+            paymentHandler: paymentHandler,
+            analyticsHelper: ._testValue(),
+            completion: { _, _ in
+                exp.fulfill()
+            }
```
wooj-stripe, line 176: Seems like we should do some asserting here? e.g. ensure it has been completed?

wooj-stripe (approved):

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheet+APIMockTest.swift
davidme-stripe: We do, see the "confirm completed" expectation defined above. Additionally, each of the above stubs have their own expectations.

PR #4224, opened 2024-11-05
Fix embeddedSettingsView displaying stale values by passing a closure

porter-stripe:
## Summary
This PR addresses an issue where embeddedSettingsView in PaymentSheetTestPlayground occasionally shows outdated values because state changes weren't propagating to the embedded SwiftUI view within UIKit. To fix this, we modified the presentEmbedded method to accept a closure that returns a view (@escaping () -> SettingsView) instead of a static view. This change ensures that embeddedSettingsView is re-evaluated whenever it's accessed, allowing it to reflect the latest state updates correctly.

## Motivation
- Embedded

## Testing
-  Manual

## Changelog
N/A

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/EmbeddedPlaygroundViewController.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/EmbeddedPlaygroundViewController.swift
@@ -8,8 +8,14 @@
 import Foundation
 @_spi(EmbeddedPaymentElementPrivateBeta) @_spi(STP) @_spi(ExperimentalAllowsRemovalOfLastSavedPaymentMethodAPI) import StripePaymentSheet
 import UIKit
+import Combine
+import SwiftUI
 
 class EmbeddedPlaygroundViewController: UIViewController {
+    private var hostingController: UIHostingController<AnyView>?
+    private var cancellables = Set<AnyCancellable>()
+    private var playgroundController: PlaygroundController
+    
     var isLoading: Bool = false {
         didSet {
             if isLoading {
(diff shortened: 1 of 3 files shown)


Example/PaymentSheet Example/PaymentSheet Example/EmbeddedPlaygroundViewController.swift
```diff
@@ -146,9 +155,34 @@ class EmbeddedPlaygroundViewController: UIViewController {
...
+        hostingController.didMove(toParent: self)
+    }
+    
+    private func observePlaygroundController() {
+        playgroundController.objectWillChange
+            .sink { [weak self] _ in
+                DispatchQueue.main.async {
+                    self?.hostingController?.rootView = AnyView(
+                        EmbeddedSettingsView().environmentObject(self?.playgroundController ?? PlaygroundController(settings: PaymentSheetTestPlaygroundSettings.defaultValues()))
+                    )
```
wooj-stripe, line 182: If we pass in a new instance of PlaygroundController, then seems like it would no longer be observed?

Example/PaymentSheet Example/PaymentSheet Example/EmbeddedPlaygroundViewController.swift
```diff
@@ -8,8 +8,14 @@
 import Foundation
 @_spi(EmbeddedPaymentElementPrivateBeta) @_spi(STP) @_spi(ExperimentalAllowsRemovalOfLastSavedPaymentMethodAPI) import StripePaymentSheet
 import UIKit
+import Combine
+import SwiftUI
 
 class EmbeddedPlaygroundViewController: UIViewController {
+    private var hostingController: UIHostingController<AnyView>?
+    private var cancellables = Set<AnyCancellable>()
+    private var playgroundController: PlaygroundController
```
wooj-stripe, line 17: Seems like `playgroundController` should be a weak reference to avoid a retain cycle?

Example/PaymentSheet Example/PaymentSheet Example/EmbeddedPlaygroundViewController.swift
porter-stripe: Yep good point

Example/PaymentSheet Example/PaymentSheet Example/EmbeddedPlaygroundViewController.swift
porter-stripe: This shouldn't happen, but I re-wrote this so it's better!

wooj-stripe (approved):

PR #4226, opened 2024-11-05
Update embedded appearance API with changes from API review

porter-stripe:
## Summary
- Moved EmbeddedPaymentElement.Style struct to EmbeddedPaymentElement.Row.Style
- Moved EmbeddedPaymentElement.style property to EmbeddedPaymentElement.row.style

## Motivation
- Embedded appearance API changes

## Testing
- Manual/CI

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetAppearance.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetAppearance.swift
@@ -237,25 +237,25 @@ public extension PaymentSheet {
     /// Describes the appearance of the Embedded Mobile Payment Element
     @_spi(EmbeddedPaymentElementPrivateBeta) struct EmbeddedPaymentElement: Equatable {
 
-        /// The display style options for the Embedded Mobile Payment Element
-        public enum Style: CaseIterable {
-            /// A flat style with radio buttons
-            case flatWithRadio
-            /// A floating button style
-            case floatingButton
-        }
-
         /// Creates a `PaymentSheet.Appearance.EmbeddedPaymentElement` with default values
         public init() {}
 
         /// Describes the appearance of the row in the Embedded Mobile Payment Element
         public var row: Row = Row()
 
-        /// The display style of the Embedded Mobile Payment Element
-        public var style: Style = .flatWithRadio
-
         /// Describes the appearance of the row in the Embedded Mobile Payment Element
         public struct Row: Equatable {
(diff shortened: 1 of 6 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetAppearance.swift
```diff
@@ -237,25 +237,25 @@ public extension PaymentSheet {
...
+            public enum Style: CaseIterable {
+                /// A flat style with radio buttons
+                case flatWithRadio
+                /// A floating button style
+                case floatingButton
+            }
+            
+            /// The display style of the Embedded Mobile Payment Element
+            public var style: Style = .flatWithRadio
+
```
wooj-stripe, line 258: The doc strings for these fields should be updated to match what is in the API review. E.g. for `var style: Style`, it should be:
```
            /// The display style of the row
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/STPAnalyticsClient+PaymentSheet.swift
```diff
@@ -86,7 +86,7 @@ extension PaymentSheet.Appearance.EmbeddedPaymentElement {
 
     var analyticPayload: [String: Any] {
         var payload = [String: Any]()
-        payload["style"] = style != PaymentSheet.Appearance.EmbeddedPaymentElement.default.style
+        payload["style"] = row.style != PaymentSheet.Appearance.EmbeddedPaymentElement.default.row.style
         payload["row"] = row != PaymentSheet.Appearance.EmbeddedPaymentElement.default.row
```
wooj-stripe, line 90: While we still have the chance (pre beta/GA), seems like we should update the key for this to `row_style`?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetAppearance.swift
porter-stripe: Ah, didn't catch those were changed, thanks!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/STPAnalyticsClient+PaymentSheet.swift
wooj-stripe: Hmm at some point in time, we should update our `PaymentSheetAnalyticsHelperTest` to include unit tests for this.

PR #4229, opened 2024-11-06
Fix removal dialog showing incorrect brand

porter-stripe:
## Summary
- When a co-branded card does not have a preferred brand set, we should fall back to the display brand not the brand.

## Motivation
- Bug fix

## Testing
- Manual
- New unit test

## Changelog
See diff

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentOptionsViewController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentOptionsViewController.swift
@@ -682,7 +682,7 @@ extension STPPaymentMethod {
     var removalMessage: (title: String, message: String) {
         switch type {
         case .card:
-            let brandString = STPCardBrandUtilities.stringFrom(card?.networks?.preferred?.toCardBrand ?? card?.brand ?? .unknown) ?? ""
+            let brandString = STPCardBrandUtilities.stringFrom(card?.preferredDisplayBrand ?? .unknown) ?? ""
             let last4 = card?.last4 ?? ""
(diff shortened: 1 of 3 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentOptionsViewController.swift
```diff
@@ -682,7 +682,7 @@ extension STPPaymentMethod {
     var removalMessage: (title: String, message: String) {
         switch type {
         case .card:
-            let brandString = STPCardBrandUtilities.stringFrom(card?.networks?.preferred?.toCardBrand ?? card?.brand ?? .unknown) ?? ""
+            let brandString = STPCardBrandUtilities.stringFrom(card?.preferredDisplayBrand ?? .unknown) ?? ""
```
wooj-stripe, line 685: Do we know when we should be using `preferredDisplayBrand` vs card?.networks?.preferred?  The description says that we should "fallback" but this approach seems to use the preferredDisplayBrand and fallback to unknown.

porter-stripe: preferredDisplayBrand is

```swift
    var preferredDisplayBrand: STPCardBrand {
        return networks?.preferred?.toCardBrand ?? displayBrand?.toCardBrand ?? brand
    }
```

porter-stripe: So we attempt to read the preferred brand, if nil fallback to displayBrand then fallback to brand.

wooj-stripe (approved):

PR #4232, opened 2024-11-06
Present forms in embedded

porter-stripe:
## Summary
- Now when tapping a selection in embedded, if the PM requires input we display a form to collect input and use that as the selected payment option.
- Generate a new form upon update to preserve selections
- Handle edge cases around canceling and reselecting the old option

## Motivation
- Embedded

## Testing
- Manual
- Existing tests
- New UI tests

## Changelog
N/A


porter-stripe:
Added skip dead code check as all the flags are for unused protocol function parameters for functions that will be implementing in following PRs.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedFormViewController.swift
```diff
@@ -70,25 +70,29 @@ class EmbeddedFormViewController: UIViewController {
...
         return paymentMethodFormViewController.paymentOption
     }
-
-    private let loadResult: PaymentSheetLoader.LoadResult
+    
     private let paymentMethodType: PaymentSheet.PaymentMethodType
     private let configuration: EmbeddedPaymentElement.Configuration
     private let intent: Intent
     private let elementsSession: STPElementsSession
+    private let savedPaymentMethods: [STPPaymentMethod]?
```
wooj-stripe, line 89: 1. Seems like `savedPaymentMethods` can be non-optional
2. We technically don't even have to pass in the entire array. Instead we can either pass in:

* a boolean -- "shouldUseNewCardNewCardHeader: Bool" or
* `savedPaymentMethod.first` type: `STPPaymentMethod?`

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -90,8 +92,55 @@ extension EmbeddedPaymentElement: EmbeddedPaymentMethodsViewDelegate {
         delegate?.embeddedPaymentElementDidUpdateHeight(embeddedPaymentElement: self)
     }
 
-    func selectionDidUpdate() {
-        delegate?.embeddedPaymentElementDidUpdatePaymentOption(embeddedPaymentElement: self)
+    func selectionWasTapped(didUpdate: Bool) {
+        // Defer notifying delegate until we are about to exit this function to give time to either allow the new payment option to come from
+        // The new `EmbeddedFormViewController` instance if needed
```
wooj-stripe, line 97: Saying "to give time to either allow the new.." does not sound precise.  How about:
```
Deferring notifying delegate until the exit of this function guarantees the new payment option comes from the new instance of `EmbeddedFormViewController`
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -90,8 +92,55 @@ extension EmbeddedPaymentElement: EmbeddedPaymentMethodsViewDelegate {
...
+            }
+        }
+        
+        guard case let .new(paymentMethodType) = embeddedPaymentMethodsView.selection else {
+            // This can occur when selection is being reset to nothing selected, so don't assert.
+            return
+        }
+
+        guard let presentingVC = presentingViewController else {
+            assertionFailure("Presenting view controller not found, set EmbeddedPaymentElement.presentingViewController.")
```
wooj-stripe, line 110: stpAssertionFailure("")

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -90,8 +92,55 @@ extension EmbeddedPaymentElement: EmbeddedPaymentMethodsViewDelegate {
...
+                delegate?.embeddedPaymentElementDidUpdatePaymentOption(embeddedPaymentElement: self)
+            }
+        }
+        
+        guard case let .new(paymentMethodType) = embeddedPaymentMethodsView.selection else {
+            // This can occur when selection is being reset to nothing selected, so don't assert.
+            return
+        }
+
+        guard let presentingVC = presentingViewController else {
```
wooj-stripe, line 109: nit:
        guard let presentingViewController else {

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -90,8 +92,55 @@ extension EmbeddedPaymentElement: EmbeddedPaymentMethodsViewDelegate {
...
+            paymentMethodType: paymentMethodType,
+            previousPaymentOption: self.formViewController?.previousPaymentOption,
+            analyticsHelper: analyticsHelper,
+            formCache: formCache
+        )
+        formViewController.delegate = self
+
+        // Only show forms that require user input
+        guard formViewController.collectsUserInput else {
+            self.formViewController = nil  // Clear out any previous form view controller to update self._paymentOption
```
wooj-stripe, line 128: 1. s/self._paymentOption/self.paymentOption ??
2. I don't understand this comment -- I would have expected to see a didSet for `formViewController` to clear out `paymentOption`

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -90,8 +92,55 @@ extension EmbeddedPaymentElement: EmbeddedPaymentMethodsViewDelegate {
...
+        }
+
+        let bottomSheet = BottomSheetViewController(
+            contentViewController: formViewController,
+            appearance: configuration.appearance,
+            isTestMode: configuration.apiClient.isTestmode,
+            didCancelNative3DS2: {
+                stpAssertionFailure("3DS2 was triggered unexpectedly")
+            }
+        )
```
wooj-stripe, line 139: nit: since i'm seeing duplicate code in 3 spots now... maybe should add a function:
```
    func bottomSheetController(with viewController: BottomSheetContentViewController) -> BottomSheetViewController {
        return BottomSheetViewController(contentViewController: viewController,
                                         isTestMode: configuration.apiClient.isTestmode,
                                         didCancelNative3DS2: {
            stpAssertionFailure("3DS2 was triggered unexpectedly")
        })
    }
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
```diff
@@ -127,15 +127,30 @@ public final class EmbeddedPaymentElement {
...
+            var formViewController: EmbeddedFormViewController?
+            if let formPaymentMethodType = self?.formViewController?.selectedPaymentOption?.paymentMethodType {
+                formViewController = EmbeddedFormViewController(configuration: configuration,
+                                                                intent: loadResult.intent,
+                                                                elementsSession: loadResult.elementsSession,
+                                                                savedPaymentMethods: loadResult.savedPaymentMethods,
+                                                                paymentMethodType: formPaymentMethodType,
+                                                                previousPaymentOption: self?.formViewController?.selectedPaymentOption,
+                                                                analyticsHelper: analyticsHelper,
+                                                                formCache: .init()) // Always use a fresh cache after an update
```
wooj-stripe, line 141: nit: instead of having formCache a required arg, make the default parameter use a fresh cache?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
```diff
@@ -127,15 +127,30 @@ public final class EmbeddedPaymentElement {
...
+                
+            }
+            
+            // 2.4 Re-initialize embedded view to update the UI to match the newly loaded data.
             let embeddedPaymentMethodsView = Self.makeView(
                 configuration: configuration,
                 loadResult: loadResult,
                 analyticsHelper: analyticsHelper,
-                previousPaymentOption: self?._paymentOption,
+                previousPaymentOption: formViewController != nil ? formViewController?.selectedPaymentOption : self?._paymentOption, // Use the possibly new payment option from the formVC if it exists, self?._paymentOption is not yet updated with the new formVC
```
wooj-stripe, line 150: 1. we should break this into a separate line
2. I'm not following what we're doing here, because the computed variable has:
```
        if let formPaymentOption = formViewController?.selectedPaymentOption {
            return formPaymentOption
        }
```
I'm sure i'm missing something, but it seems like we could just pass in self?._paymentOption without checking formViewController here.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
```diff
@@ -155,6 +170,8 @@ public final class EmbeddedPaymentElement {
             self.intent = loadResult.intent
             self.embeddedPaymentMethodsView = embeddedPaymentMethodsView
             self.containerView.updateEmbeddedPaymentMethodsView(embeddedPaymentMethodsView)
+            self.formViewController = formViewController
+            self.formCache = .init()
```
wooj-stripe, line 174: I'm not following why this class needs a formCache. Seems like only `formViewController` would need a copy since we aren't preserving the cache between update calls?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
```diff
@@ -24,12 +24,20 @@ class EmbeddedPaymentMethodsView: UIView {
...
     private(set) var selection: Selection? {
+        willSet {
+            previousSelection = selection
+        }
         didSet {
             updateMandate()
-            if oldValue != selection {
-                delegate?.selectionDidUpdate()
-            }
+            delegate?.selectionWasTapped(didUpdate: oldValue != selection)
```
wooj-stripe, line 40: I think there are times where the selectionDidUpdate, but a user didn't actually tap (e.g. removal use case), so I think the name selectionDidUpdate() is more appropriate.

wooj-stripe (commented):
Leaving comments a bit early, didn't review test code yet.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
porter-stripe: The cache needs to be kept in storage between creations (presentations) of formViewController, to preserve input of multiple PMs, and partial user input. That way if we
1. Present Klarna enter email
2. Close Klarna form
3. Open card, start typing a number
4. Close card
5. Open Klarna, we still have the cached input in Klarna email field.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedFormViewController.swift
porter-stripe: Ah, I like this better

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
porter-stripe: 1. Broke this into it's own line
2. paymentOption looks at `self.formViewController`, but at this point in the function `self.formViewController` is not yet updated with the latest instance, `self.formViewController` is updated with this new instance later in the function with the rest of the properties. So we need to look at the freshly created formViewController if it exists for a previous payment option. I added a better comment here.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
porter-stripe: Updated!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
```diff
@@ -189,9 +197,20 @@ class EmbeddedPaymentMethodsView: UIView {
             delegate?.heightDidChange()
         }
     }
+    
+    // MARK: Internal functions
+    func resetSelectionToLastSelection() {
```
wooj-stripe, line 202: 1. `resetSelectionToLastSelection` looks super similar to `didTap` -- maybe we find a way to share some code?
2.  I am seeing a case where i instantiate embedded, saved card is selected, i click on "new card" then tap cancel, and my original selection is not preserved.

porter-stripe: Refactored #1 and fixed #2.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
```diff
@@ -114,6 +129,7 @@ class EmbeddedPaymentMethodsView: UIView {
             if initialSelection == selection {
                 applePayRowButton.isSelected = true
                 self.selection = initialSelection
+                self.selectedRowButton = savedPaymentMethodButton
```
wooj-stripe, line 132: I'm having trouble following this logic.. here's how i'm reading it:
1.  If the initial selection is apple pay, then..
  i. applePayRowButton should be selected
  ii. set the self.selection = initialSelection
  iii. set the selectedRowButton back to the savedPaymentMethodButton (but wait, in "i", we've set apple pay row button to be selected?)

Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
```diff
@@ -14,11 +14,43 @@ class EmbeddedUITests: PaymentSheetUITestCase {
...
         app.buttons["Present embedded payment element"].waitForExistenceAndTap()
-        // TODO: Test card form (see PaymentSheetVerticalUITests testUpdate)
-
-        // Selecting Alipay w/ deferred PaymentIntent...
+        
+        // Entering a card w/ deferred PaymentIntent...
+        app.buttons["Card"].waitForExistenceAndTap()
+        XCTAssertTrue(app.staticTexts["Add card"].waitForExistence(timeout: 10))
+        try! fillCardData(app, postalEnabled: true)
+        app.toolbars.buttons["Done"].tap()
```
wooj-stripe, line 25: 1. waitForExistenceAndTap() is probably more reliable.
2. TIL: didn't realize there was toolbars.buttons, i've always used "app.buttons["Done"]"

Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
```diff
@@ -14,11 +14,43 @@ class EmbeddedUITests: PaymentSheetUITestCase {
...
+        XCTAssertTrue(app.buttons["Continue"].isEnabled)
+        app.buttons["Continue"].waitForExistenceAndTap()
+        XCTAssertTrue(app.staticTexts["Payment method"].waitForExistence(timeout: 10))
+        XCTAssertEqual(app.staticTexts["Payment method"].label, "•••• 4242")
+        
+        // ...and *updating* to a SetupIntent...
+        app.buttons.matching(identifier: "Setup").element(boundBy: 1).tap()
```
wooj-stripe, line 32: waitForExistenceAndTap() is probably more reliable.

Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
```diff
@@ -14,11 +14,43 @@ class EmbeddedUITests: PaymentSheetUITestCase {
...
+        // ...(wait for it to finish updating)...
+        _ = app.buttons["Reload"].waitForExistence(timeout: 10)
```
wooj-stripe, line 34: XCTAssertTrue on waitForExistence to get a better fail state in our tests. -- this should be applied to any time we "Reload"

Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
```diff
@@ -14,11 +14,43 @@ class EmbeddedUITests: PaymentSheetUITestCase {
...
+        
+        // ....Tapping card should show the card form with details preserved
+        app.buttons["Card"].waitForExistenceAndTap()
+        // ...thus the Continue button should be enabled
+        app.buttons["Continue"].waitForExistenceAndTap()
+        // ...should cause the card ending in 4242 that was previously entered to be the selected payment method
+        XCTAssertTrue(app.staticTexts["Payment method"].waitForExistence(timeout: 10))
+        XCTAssertEqual(app.staticTexts["Payment method"].label, "•••• 4242")
+        // ...switching from setup to payment should preserve this card as the selected payment method
+        app.buttons.matching(identifier: "Payment").element(boundBy: 1).tap()
```
wooj-stripe, line 46: s/tap/waitForExistenceAndTap

Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
```diff
@@ -27,6 +59,49 @@ class EmbeddedUITests: PaymentSheetUITestCase {
...
+        // ...go back into deferred PaymentIntent mode
+        app.buttons.matching(identifier: "Payment").element(boundBy: 1).tap()
+        // ...(wait for it to finish updating)...
+        _ = app.buttons["Reload"].waitForExistence(timeout: 10)
+        //...selecting Klarna w/ deferred PaymentIntent...
+        app.buttons["Klarna"].waitForExistenceAndTap()
+        // ...fill out the form for Klarna
+        let emailField = app.textFields["Email"]
+        XCTAssertTrue(emailField.waitForExistence(timeout: 10))
+        emailField.tap()
```
wooj-stripe, line 86: s/tap/waitForExistenceAndTap

Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
```diff
@@ -293,6 +368,110 @@ class EmbeddedUITests: PaymentSheetUITestCase {
...
+        settings.mode = .paymentWithSetup
+        settings.uiStyle = .paymentSheet
+        settings.customerKeyType = .legacy
+        settings.formSheetAction = .continue
+        settings.customerMode = .new
+        loadPlayground(app, settings)
+        
+        // Start by saving a new card
+        app.buttons["Present PaymentSheet"].waitForExistenceAndTap()
+        try! fillCardData(app, postalEnabled: true)
```
wooj-stripe, line 383: I think we can use a .returning customer here so we don't have to fillCardData?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
porter-stripe: Good point, the applePayRowButton.isSelected = true is a bit repetitive and reads weird. I moved this below as just one line for all.

Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
porter-stripe: Yeah I'd like to, but the returning customer has two PMs saved and the one that is selected by default changes randomly between a card and bank making it not great for tests.

Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
```diff
@@ -293,6 +367,110 @@ class EmbeddedUITests: PaymentSheetUITestCase {
...
     }
+    
+    func testSelection() {
+        var settings = PaymentSheetTestPlaygroundSettings.defaultValues()
+        settings.mode = .paymentWithSetup
+        settings.uiStyle = .paymentSheet
+        settings.customerKeyType = .legacy
+        settings.formSheetAction = .continue
+        settings.customerMode = .new
+        loadPlayground(app, settings)
```
wooj-stripe, line 378: We should add a test, which tests that selection disappears when a payment method disappears.
e.g.
1. Have mobilepay selected w/ payment mode.
2. Switch to payment+sfu (which forces an update)

Ensure that mobile pay is not the selected payment method, and it is not showing on screen.

This is important because the init code for `EmbeddedPaymentMethodsView` is a bit complex -- it's easy to think  we can just set `self.selection = initialSelection` prior to calling `super.init(frame: .zero)` but this would result in showing that the unavailable payment method is still available (despite it not being selected)

wooj-stripe (approved):

PR #4234, opened 2024-11-06
Fix issue with bank form in vertical mode

tillh-stripe:
## Summary

This pull request fixes an issue with bank forms (ACH, Instant Debits, and Link Card Brand) in vertical mode when email collection (and name collection, for the ACH case) was set to `never`.

This billing details collection configuration results in an empty form, which in vertical mode means “show the pay button instead of the form”. That doesn’t work for any bank payment methods, since they need to “hijack” the primary button to open the bank auth flow.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
@@ -474,3 +474,32 @@ extension PaymentElementConfiguration {
         (defaultBillingDetails.email?.isEmpty == false && billingDetailsCollectionConfiguration.attachDefaultsToPaymentMethod)
     }
 }
+
+extension PaymentSheet.PaymentMethodType {
+    
+    var isLinkBankPayment: Bool {
+        switch self {
+        case .stripe:
+            return false
+        case .external:
+            return false
+        case .instantDebits:
+            return true
+        case .linkCardBrand:
+            return true
+        }
+    }
+    
+    var isBankPayment: Bool {
+        switch self {
+        case .stripe(let type):
+            return type == .USBankAccount
+        case .external:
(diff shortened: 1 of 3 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
```diff
@@ -474,3 +474,32 @@ extension PaymentElementConfiguration {
...
+        switch self {
+        case .stripe:
+            return false
+        case .external:
+            return false
+        case .instantDebits:
+            return true
+        case .linkCardBrand:
+            return true
+        }
```
mats-stripe, line 490: Minor refactor suggestions:

```suggestion
        [.instantDebits, .linkCardBrand].contains(self)
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
```diff
@@ -474,3 +474,32 @@ extension PaymentElementConfiguration {
...
+        switch self {
+        case .stripe(let type):
+            return type == .USBankAccount
+        case .external:
+            return false
+        case .instantDebits:
+            return true
+        case .linkCardBrand:
+            return true
+        }
```
mats-stripe, line 503:
```suggestion
        self == .stripe(.USBankAccount) || isLinkBankPayment
```

mats-stripe (approved):
LGTM!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
tillh-stripe: Same as above.

PR #4242, opened 2024-11-11
Confirm payments in embedded

porter-stripe:
## Summary
- Supports confirming payments in the confirm and continue formSheetAction for embedded
- Fixes a bug where selecting a form PM and filling it out, then selecting back to a saved card did not update the payment option correctly. Added a test for this.
- A PR immediately following this will add more UI tests for more edge cases like CVC recollection (which was tested manually)

## Motivation
- Embedded

## Testing
- Manual
- Unit tests
- UI tests

## Changelog
N/A

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/EmbeddedPlaygroundViewController.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/EmbeddedPlaygroundViewController.swift
@@ -39,7 +39,7 @@ class EmbeddedPlaygroundViewController: UIViewController {
     private(set) var embeddedPaymentElement: EmbeddedPaymentElement?
 
     private lazy var loadingIndicator: UIActivityIndicatorView = {
-        let indicator = UIActivityIndicatorView(style: .medium)
+        let indicator = UIActivityIndicatorView(style: .large)
         indicator.translatesAutoresizingMaskIntoConstraints = false
         indicator.hidesWhenStopped = true
         return indicator
@@ -53,7 +53,8 @@ class EmbeddedPlaygroundViewController: UIViewController {
         checkoutButton.setTitle("Checkout", for: .normal)
         checkoutButton.setTitleColor(.white, for: .normal)
         checkoutButton.translatesAutoresizingMaskIntoConstraints = false
-        checkoutButton.isEnabled = false
+        checkoutButton.isEnabled = embeddedPaymentElement?.paymentOption != nil
+        checkoutButton.addTarget(self, action: #selector(pay), for: .touchUpInside)
         return checkoutButton
     }()
 
@@ -189,6 +190,24 @@ class EmbeddedPlaygroundViewController: UIViewController {
             }
             .store(in: &cancellables)
     }
+    
+    @objc func pay() {
+        Task { @MainActor in
+            guard let embeddedPaymentElement else { return }
+            self.isLoading = true
+            let result = await embeddedPaymentElement.confirm()
+            self.isLoading = false
+            
+            switch result {
+            case .completed, .failed:
+                playgroundController?.lastPaymentResult = result
+                self.dismiss(animated: true)
+            case .canceled:
+                break
+            }
+        }
+    }
+
 }
 
 // MARK: - EmbeddedPaymentElementDelegate
--- a/Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
+++ b/Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
@@ -106,9 +106,7 @@ class EmbeddedUITests: PaymentSheetUITestCase {
         XCTAssertTrue(app.buttons["Checkout"].isEnabled)
         app.buttons["Checkout"].waitForExistenceAndTap()
         let springboard = XCUIApplication(bundleIdentifier: "com.apple.springboard")
-        let sbContinueButton = springboard.buttons["Continue"]
-        XCTAssertTrue(sbContinueButton.waitForExistence(timeout: 10.0))
-        sbContinueButton.tap()
+        springboard.buttons["Continue"].waitForExistenceAndTap()
         // Stop here; Klarna's test playground is out of scope
     }
 
@@ -533,9 +531,7 @@ class EmbeddedUITests: PaymentSheetUITestCase {
         app.buttons["Link"].waitForExistenceAndTap()
         app.buttons["Checkout"].waitForExistenceAndTap()
         let springboard = XCUIApplication(bundleIdentifier: "com.apple.springboard")
-        let sbContinueButton = springboard.buttons["Continue"]
-        XCTAssertTrue(sbContinueButton.waitForExistence(timeout: 10.0))
-        sbContinueButton.tap()
+        springboard.buttons["Continue"].waitForExistenceAndTap()
         // Stop here; Links's test playground is out of scope
     }
 
--- a/Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
+++ b/Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
@@ -495,6 +495,11 @@ class EmbeddedUITests: PaymentSheetUITestCase {
         XCTAssertEqual(app.staticTexts["Payment method"].label, "••••6789")
         XCTAssertTrue(app.buttons["Checkout"].isEnabled)
         
+        // Confirm with the saved card
+        app.buttons["•••• 4242"].waitForExistenceAndTap()
+        XCTAssertEqual(app.staticTexts["Payment method"].label, "•••• 4242")
+        XCTAssertTrue(app.buttons["Checkout"].isEnabled)
+        
         app.buttons["Checkout"].waitForExistenceAndTap()
         XCTAssertTrue(app.staticTexts["Success!"].waitForExistence(timeout: 20))
     }
--- a/Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
+++ b/Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
@@ -498,8 +498,8 @@ class EmbeddedUITests: PaymentSheetUITestCase {
         // Confirm with the saved card
         app.buttons["•••• 4242"].waitForExistenceAndTap()
         XCTAssertEqual(app.staticTexts["Payment method"].label, "•••• 4242")
+        app.swipeUp() // scroll to see the checkout button
         XCTAssertTrue(app.buttons["Checkout"].isEnabled)
-        
         app.buttons["Checkout"].waitForExistenceAndTap()
         XCTAssertTrue(app.staticTexts["Success!"].waitForExistence(timeout: 20))
     }
--- a/Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
+++ b/Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
@@ -103,8 +103,9 @@ class EmbeddedUITests: PaymentSheetUITestCase {
         XCTAssertEqual(app.staticTexts["Payment method"].label, "Klarna")
         
         // Confirm the Klarna payment
+        XCTAssertTrue(app.buttons["Checkout"].waitForExistence(timeout: 10))
         XCTAssertTrue(app.buttons["Checkout"].isEnabled)
-        app.buttons["Checkout"].waitForExistenceAndTap()
+        app.buttons["Checkout"].tap()
         let springboard = XCUIApplication(bundleIdentifier: "com.apple.springboard")
         springboard.buttons["Continue"].waitForExistenceAndTap()
         // Stop here; Klarna's test playground is out of scope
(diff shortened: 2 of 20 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedFormViewController.swift
```diff
@@ -18,15 +18,13 @@ import UIKit
...
-    ///   - paymentOption: The `PaymentOption` to be confirmed.
     ///   - completion: A completion handler to call with the `PaymentSheetResult` from the confirmation attempt.
     func embeddedFormViewControllerShouldConfirm(
         _ embeddedFormViewController: EmbeddedFormViewController,
-        with paymentOption: PaymentOption,
         completion: @escaping (PaymentSheetResult, STPAnalyticsClient.DeferredIntentConfirmationType?) -> Void
     )
 
-    /// This method is called when the user taps the primary button (e.g., "Buy") while `formSheetAction` is set to `.continue`.
+    /// This method is called when the user taps the primary button (e.g., "Buy") while `formSheetAction` is set to `.confirm` after confirmation.
```
wooj-stripe, line 27: 1. Maybe rewrite to:
When `formSheetAction` is set to `.confirm`, this method is called after a user taps the primary button (e.g, "Buy") followed by a successful confirmation.

2. The name of the function is awkward to me.  "xxxxShouldContinue" seems like the function would return a boolean to describe xxxx to continue or not.

3. The name "Continue" clashes with the formSheetAction value .continue -- maybe we can find a better name here?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -293,6 +300,68 @@ extension EmbeddedPaymentElement: EmbeddedFormViewControllerDelegate {
...
+                    // formViewController is never currently presented during confirmation in continue mode
+                    if let presentingViewController {
+                        return STPAuthenticationContextWrapper(presentingViewController: presentingViewController)
+                    }
+                    return nil
+                }
+            }()
+            
+            guard let authContext else {
+                return (.failed(error: PaymentSheetError.unknown(debugDescription: "Unexpectedly found nil on both bottomSheetController and presentingViewController.")), STPAnalyticsClient.DeferredIntentConfirmationType.none)
```
wooj-stripe, line 327: Probably better just to put that authContext is nil.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -293,6 +300,68 @@ extension EmbeddedPaymentElement: EmbeddedFormViewControllerDelegate {
...
+    
+    func _confirm() async -> (result: PaymentSheetResult, deferredIntentConfirmationType: STPAnalyticsClient.DeferredIntentConfirmationType?) {
+        let result: (PaymentSheetResult, STPAnalyticsClient.DeferredIntentConfirmationType?) = await {
+            
+            let authContext: STPAuthenticationContext? = {
+                switch configuration.formSheetAction {
+                case .confirm:
+                    if let formViewController {
+                        return formViewController
+                    }
```
wooj-stripe, line 312: nit: seems like we should check to make sure things are on screen before presenting?

```
                    if let formViewController,
                       formViewController.view.window != nil{
                        return formViewController
                    } else if let presentingViewController {
                        return STPAuthenticationContextWrapper(presentingViewController: presentingViewController)
                    }
                    return nil
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -293,6 +300,68 @@ extension EmbeddedPaymentElement: EmbeddedFormViewControllerDelegate {
...
+                                                    intent: intent,
+                                                    elementsSession: elementsSession,
+                                                    paymentOption: paymentOption,
+                                                    paymentHandler: paymentHandler,
+                                                    integrationShape: .embedded)
+        }()
+        
+        // Wait for the last update to finish and fail if didn't succeed. A failure means the view is out of sync with the intent and could e.g. not be showing a required mandate.
+        if let latestUpdateTask {
+            switch await latestUpdateTask.value {
```
wooj-stripe, line 346: I'm a bit confused on how this function works. It seems like you'd want to wait for latestUpdateTask to finish before attempting to confirm, but the way the code is written, it looks like `await PaymentSheet.confirm` could execute before checking `latestUpdateTask`?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/PayWithLinkController.swift
```diff
@@ -64,7 +64,7 @@ extension PayWithLinkController: PayWithLinkWebControllerDelegate {
             elementsSession: elementsSession,
             paymentOption: paymentOption,
             paymentHandler: paymentHandler,
-            isFlowController: false
+            integrationShape: .complete
```
wooj-stripe, line 67: Maybe instead of passing in `integrationShape`, we just pass in the value for requiresInterstitialForCVC?

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentElementTest.swift
```diff
@@ -23,6 +23,19 @@ class EmbeddedPaymentElementTest: XCTestCase {
...
+    let paymentIntentConfigWithConfirmHandler = EmbeddedPaymentElement.IntentConfiguration(mode: .payment(amount: 1000, currency: "USD"), paymentMethodTypes: ["card", "cashapp"]) {paymentMethod, _, intentCreationCallback in
+        STPTestingAPIClient.shared.fetchPaymentIntent(types: ["card"],
+                                currency: "USD",
+                                paymentMethodID: paymentMethod.stripeId,
+                                confirm: true) { result in
+            switch result {
+            case .success(let clientSecret):
+                intentCreationCallback(.success(clientSecret))
+            case .failure(let error):
+                print(error)
```
wooj-stripe, line 35: XCTFail("") ??

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentElementTest.swift
```diff
@@ -185,6 +212,68 @@ class EmbeddedPaymentElementTest: XCTestCase {
...
+        sut.delegate = self
+        sut.presentingViewController = UIViewController()
+        sut.view.autosizeHeight(width: 320)
+        
+        // Create test confirmParams with valid card details
+        let confirmParams = IntentConfirmParams(type: .stripe(.card))
+        confirmParams.paymentMethodParams.card = STPPaymentMethodCardParams()
+        confirmParams.paymentMethodParams.card?.number = "4242424242424242"
+        confirmParams.paymentMethodParams.card?.expMonth = NSNumber(value: 12)
+        confirmParams.paymentMethodParams.card?.expYear = NSNumber(value: 34)
```
wooj-stripe, line 228: Maybe we add a common function like:
```
    func nextYear() -> Int {
        let year = Calendar.current.component(.year, from: Date())
        let yearPlus1 = year.advanced(by: 1)
        return yearPlus1 - 2000
    }
```

wooj-stripe: We have some example of this already in `STPPaymentIntentFunctionalTest`

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedFormViewController.swift
porter-stripe: Renaming to `embeddedFormViewControllerDidCompleteConfirmation` and made a minor change to the doc comment.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
porter-stripe: Added this, although I can't conceive a scenario where this would occur.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
porter-stripe: `result` is a closure that gets [invoked](https://github.com/stripe/stripe-ios/pull/4242/files#diff-b66bb3ee17a6e8b6b2c784056b390b3ee60b34d788c38ba8aac66ef1ded1558dR349) if the last update has completed or just [returns](https://github.com/stripe/stripe-ios/pull/4242/files#diff-b66bb3ee17a6e8b6b2c784056b390b3ee60b34d788c38ba8aac66ef1ded1558dR362) result if there was no last update. We have a test for this [here](https://github.com/stripe/stripe-ios/pull/4242/files#diff-8908c6f71c82abec6b4f4989d9fa0ce5750cfa6e76702fb67d8c27ae86e9e1d7R175).

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/PayWithLinkController.swift
porter-stripe: We did something similar to integration shape in PaymentSheetLoader, and I tend to prefer this over the flag as it's more extensible for possible future changes in confirmation depending on surface. The flag `isFlowController` is a good example of why it's limiting IMO.

Example/PaymentSheet Example/PaymentSheet Example/EmbeddedPlaygroundViewController.swift
```diff
@@ -189,6 +190,24 @@ class EmbeddedPlaygroundViewController: UIViewController {
...
+            self.isLoading = true
+            let result = await embeddedPaymentElement.confirm()
+            self.isLoading = false
+            
+            switch result {
+            case .completed, .failed:
+                playgroundController?.lastPaymentResult = result
+                self.dismiss(animated: true)
+            case .canceled:
+                break
```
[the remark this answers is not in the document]
porter-stripe: This only happens when the user closes the sheet, I feel just dismissing the sheet is enough of an indication it was canceled and that's probably what I'd expect merchants to do in the real world too.

Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
```diff
@@ -101,6 +101,13 @@ class EmbeddedUITests: PaymentSheetUITestCase {
         XCTAssertTrue(app.buttons["Reload"].waitForExistence(timeout: 10))
         // ... Klarna should still be selected
         XCTAssertEqual(app.staticTexts["Payment method"].label, "Klarna")
+        
+        // Confirm the Klarna payment
+        XCTAssertTrue(app.buttons["Checkout"].isEnabled)
+        app.buttons["Checkout"].waitForExistenceAndTap()
```
wooj-stripe, line 107: Check existence before checking isEnabled?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
porter-stripe: You're understanding is right, mine wasn't. Fixed this.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentElementTest.swift
```diff
@@ -138,14 +151,28 @@ class EmbeddedPaymentElementTest: XCTestCase {
...
+        let sut = try await EmbeddedPaymentElement.create(intentConfiguration: paymentIntentConfigWithConfirmHandler, configuration: configuration)
+        sut.presentingViewController = UIViewController()
+        sut.view.autosizeHeight(width: 320)
+        
+        // Create test confirmParams with valid card details
+        let confirmParams = IntentConfirmParams(type: .stripe(.card))
+        confirmParams.paymentMethodParams.card = STPPaymentMethodCardParams()
+        confirmParams.paymentMethodParams.card?.number = "4242424242424242"
+        confirmParams.paymentMethodParams.card?.expMonth = NSNumber(value: 12)
+        confirmParams.paymentMethodParams.card?.expYear = NSNumber(value: 34)
```
wooj-stripe, line 163: extend +5 years? (doesnt need to be in this pr)

wooj-stripe (approved):

PR #4246, opened 2024-11-12
Fix Retain Cycles in `CustomerSheet.swift` and `STPCameraView.swift`

DrHazemAli:
## Summary
This pull request resolves retain cycle issues identified in `CustomerSheet.swift` and `STPCameraView.swift` within the `StripePaymentSheet` module. These changes improve memory management by ensuring proper deallocation of resources and preventing memory leaks.

## Motivation
The retain cycles were identified as potential causes of memory leaks, especially during interactions involving the payment sheet and camera preview. To demonstrate the problem:
- In `CustomerSheet.swift`, closures retained strong references to `self`, resulting in retained memory even after the payment flow was complete.
- In `STPCameraView.swift`, the `playSnapshotAnimation` function held a strong reference to `self` inside a `DispatchQueue.main.async` closure, leading to memory retention of the `flashLayer` object.

The changes involve modifying the closure behavior by using `[weak self]`, thus breaking the retain cycles and allowing proper deallocation.

## Testing
- **CustomerSheet.swift**:
  - The `CustomerSheet` presentation and dismissal were tested across different scenarios involving multiple payment methods and cancellation flows.
  - Confirmed that no retain cycles remained, and memory was correctly released after each interaction.

- **STPCameraView.swift**:
  - The `playSnapshotAnimation` functionality was tested on both simulator and real devices to ensure the snapshot effect occurs as intended.
  - Verified that the flash animation did not retain memory unnecessarily, ensuring proper memory release when the view was dismissed.

## Changelog
- `[Fixed]` Retain cycle issues in `CustomerSheet.swift` by adding `[weak self]` to closures.
- `[Fixed]` Retain cycle issues in `STPCameraView.swift` by updating the closure handling in `playSnapshotAnimation`.

### References
- [Automatic Reference Counting (ARC) - Swift Programming Language Guide](https://developer.apple.com/documentation/swift/automatic_reference_counting)
- [Strong Reference Cycles for Closures - Swift Language Guide](https://developer.apple.com/documentation/swift/automatic_reference_counting/avoiding_strong_reference_cycles_for_closures)

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Helpers/STPCameraView.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Helpers/STPCameraView.swift
@@ -36,7 +36,8 @@ class STPCameraView: UIView {
             x: 0, y: 0, width: layer.bounds.size.width, height: layer.bounds.size.height)
         flashLayer?.opacity = 1.0
         CATransaction.commit()
-        DispatchQueue.main.async(execute: {
+        DispatchQueue.main.async(execute: { [weak self] in
+            guard let self = self else { return }
             let fadeAnim = CABasicAnimation(keyPath: "opacity")
             fadeAnim.fromValue = NSNumber(value: 1.0)
             fadeAnim.toValue = NSNumber(value: 0.0)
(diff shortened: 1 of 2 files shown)


davidme-stripe:
Thanks for filing this! In both cases, the reference count of the retained `self` should decrease once the closure itself has executed and deallocated, so I don't think this is necessarily causing a retain cycle. And in some cases (like the `shouldDismiss` function), we *do* want to retain self through the end of the block to ensure the nested completion block is called.

PR #4260, opened 2024-11-13
Added cvc and expiry to UpdateCardViewController

joyceqin-stripe:
## Summary

Added CVC and expiration date fields to the existing UpdateCardViewController and updated the button
## Motivation

Default pms sync
## Testing

test_UpdateCardViewControllerLightMode@3x


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/TextField/TextFieldElement+Card.swift
```diff
@@ -288,7 +295,12 @@ extension TextFieldElement {
                 }
                 // Is the date expired?
                 guard let expiryDate = CardExpiryDate(text), !expiryDate.expired() else {
-                    return .invalid(Error.invalid)
+                    if !isEditable {
+                        return .invalid(Error.expired)
+                    }
+                    else {
+                        return .invalid(Error.invalid)
+                    }
```
porter-stripe, line 303: Using `isEditable` probably isn't the best way to determine which error we should show, they should be independent.

You can break this up into

````swift
guard let expiryDate = CardExpiryDate(text) else {
   return .invalid(Error.invalid)
}

guard !expiryDate.expired() else {
   return .invalid(Error.expired)
}

File: StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdateCardViewController.swift [hunk 2]
```diff
@@ -62,12 +62,12 @@ final class UpdateCardViewController: UIViewController {
 
     private lazy var headerLabel: UILabel = {
         let label = PaymentSheetUI.makeHeaderLabel(appearance: appearance)
-        label.text = .Localized.update_card_brand
+        label.text = .Localized.manage_card
         return label
     }()
 
     private lazy var updateButton: ConfirmButton = {
-        return ConfirmButton(state: .disabled, callToAction: .custom(title: .Localized.update), appearance: appearance, didTap: {  [weak self] in
```
porter-stripe, line 70: `.update` string can be removed, it's unused.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdateCardViewController.swift
```diff
@@ -123,11 +134,29 @@ final class UpdateCardViewController: UIViewController {
...
+        let expiryDate = CardExpiryDate(month: paymentMethod.card?.expMonth ?? 0, year: paymentMethod.card?.expYear ?? 0)
+        let expiryDateElement = TextFieldElement.ExpiryDateConfiguration(defaultValue: expiryDate.displayString, isEditable: false).makeElement(theme: appearance.asElementsTheme)
+        return expiryDateElement
+
+    }()
+
+    private lazy var cvcElement: TextFieldElement = {
+        let cardBrandProvider = { [weak self] in
+            self?.paymentMethod.card?.brand ?? .unknown
+        }
```
porter-stripe, line 147: Let's use `paymentMethod.card?.card.preferredDisplayBrand` to better support CBC cards.

StripeUICore/StripeUICore/Source/Elements/TextField/TextFieldElement.swift
```diff
@@ -109,7 +114,12 @@ import UIKit
 
     /// Call this to manually set the text of the text field.
     public func setText(_ text: String) {
-        self.text = sanitize(text: text)
+        if configuration.isEditable {
+            self.text = sanitize(text: text)
+        }
+        else {
+            self.text = text
+        }
```
porter-stripe, line 122: Having this dupe'd code in two places doesn't feel right, and not sanitizing due to not editing doesn't make sense to me.

What happens if we remove this? Maybe we should have a text field configuration that would handle the sanitization correctly, similar to what we do for last 4 configuration.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/TextField/TextFieldElement+Card.swift
```diff
@@ -224,17 +224,43 @@ extension TextFieldElement {
...
+extension TextFieldElement {
+    struct CensoredCVCConfiguration: TextFieldElementConfiguration {
+        init(cardBrandProvider: @escaping () -> (STPCardBrand)) {
+            self.cardBrandProvider = cardBrandProvider
+            let maxLength = Int(STPCardValidator.maxCVCLength(for: cardBrandProvider()))
+            self.defaultValue = String(repeating: "•", count: maxLength)
+        }
+
+        let defaultValue: String?
+        let cardBrandProvider: () -> (STPCardBrand)
```
porter-stripe, line 237: Is there any reason to pass in the provider rather than just the brand? Seems like we can just pass in the brand.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/TextField/TextFieldElement+Card.swift
```diff
@@ -287,9 +316,14 @@ extension TextFieldElement {
                     return .invalid(Error.invalidMonth)
                 }
                 // Is the date expired?
```
porter-stripe, line 318: Nit: This comment should be above the second guard.

porter-stripe (commented):
looks good just one question

porter-stripe (commented):
This looks good, let's wait until after Monday's release to ship this to prevent regressions. In the future let's use more descriptive commit messages and reply back to feedback with what changes were needed/answering the questions.

joyceqin-stripe:
> This looks good, let's wait until after Monday's release to ship this to prevent regressions. In the future let's use more descriptive commit messages and reply back to feedback with what changes were needed/answering the questions.

Noted— I'll be sure to do so in the future. Thanks!

porter-stripe (approved):

PR #4271, opened 2024-11-15
Added a flag for default spm navigation

joyceqin-stripe:
## Summary

Added a flag defaultSPMNavigation in PaymentSheet and CustomerSheet. Off by default.
When on, edit icon shows for all cards regardless of cbc eligibility.

## Motivation

[MOBILESDK-2663](https://jira.corp.stripe.com/browse/MOBILESDK-2663)
[MOBILESDK-2664](https://jira.corp.stripe.com/browse/MOBILESDK-2664)
## Testing

Simple UI tests that check to make sure that both the cbc card and non-cbc card have edit buttons


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSavedPaymentMethodsCollectionViewController.swift
```diff
@@ -537,6 +539,10 @@ extension CustomerSavedPaymentMethodsCollectionViewController: UpdateCardViewCon
 
         let updatedViewModel: Selection = .saved(paymentMethod: updatedPaymentMethod)
         viewModels[row] = updatedViewModel
+        // Update savedPaymentMethods
```
joyceqin-stripe, line 542: This is unrelated to the default spm navigation but it's a fix for a bug I found— should I make a separate pr for that?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSheetConfiguration.swift
```diff
@@ -81,6 +81,11 @@ extension CustomerSheet {
         /// Note: Card brand filtering is not currently supported by Link.
         @_spi(CardBrandFilteringBeta) public var cardBrandAcceptance: PaymentSheet.CardBrandAcceptance = .all
 
+        /// This is an experimental feature that may be removed at any time.
+        /// If true, when editing, cards and us bank accounts will have the edit icon, users cannot remove them from the list screen.
+        /// If false (default), only cbc eligible cards can be edited and users can remove payment methods from the list screen.
+        @_spi(DefaultSPM) public var defaultSPMNavigation = false
```
porter-stripe, line 87: Can we find a better name for this? I don't think `DefaultSPM` and `defaultSPMNavigation` really describe what this flag is doing. It's also odd that it is "default" but by default is turned off, and turning it on results in non-default behavior.

porter-stripe: Let's also replace CBC with card brand choice cards, merchants may not know what "cbc" is.

Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
```diff
@@ -184,6 +184,7 @@ class PlaygroundController: ObservableObject {
         case .allowVisa:
             configuration.cardBrandAcceptance = .allowed(brands: [.visa])
         }
+        configuration.defaultSPMNavigation = settings.defaultSPMNavigation == .on
```
porter-stripe, line 187: I don't think the new behavior should be on by default, same applies below.

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
```diff
@@ -2395,6 +2395,41 @@ class PaymentSheetLinkUITests: PaymentSheetUITestCase {
 //    }
 }
 
+class PaymentSheetDefaultSPMUITests: PaymentSheetUITestCase {
+    func testDefaultSPMHorizontalNavigation() {
```
porter-stripe, line 2399: This is good! Maybe add the inverse for when defaultSPMNavigation is off and we only get 1 circular edit button.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSavedPaymentMethodsCollectionViewController.swift
porter-stripe: Nice, can you explain the bug fix and add a test for it?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentMethodCollectionView.swift
```diff
@@ -112,12 +112,18 @@ extension SavedPaymentMethodCollectionView {
...
         var allowsPaymentMethodRemoval: Bool = true
+        var defaultSPMNavigation: Bool = false
 
         /// Indicates whether the cell should be editable or just removable.
         /// If the card is a co-branded card and the merchant is eligible for card brand choice, then
         /// the cell should be editable. Otherwise, it should be just removable.
         var shouldAllowEditing: Bool {
-            return (viewModel?.isCoBrandedCard ?? false) && cbcEligible
+            if defaultSPMNavigation {
+                return viewModel?.savedPaymentMethod?.type == .card
```
porter-stripe, line 122: This will change to support bank accounts in the future right?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSavedPaymentMethodsCollectionViewController.swift
joyceqin-stripe: Yeah, originally, there was a behavior I noticed in the horizontal flow where if the merchant country were FR and the currency were eur (leading to one CBC card and one non-editable visa card in the playground), if you change the CBC card brand and then remove the non-editable visa card, the card brand displayed for the CBC card will revert to what it was. I'm attaching a screen recording for reference— the screen recording was from before the changes we made in UI, but the behavior was always the same

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/Vertical Saved Payment Method Screen/VerticalSavedPaymentMethodsViewController.swift
```diff
@@ -48,8 +48,14 @@ class VerticalSavedPaymentMethodsViewController: UIViewController {
...
-                paymentMethodRows.forEach { $0.state = .editing(allowsRemoval: canRemovePaymentMethods,
-                                                                allowsUpdating: $0.paymentMethod.isCoBrandedCard && isCBCEligible) }
+                if configuration.defaultSPMNavigation {
+                    paymentMethodRows.forEach { $0.state = .editing(allowsRemoval: false,
+                                                                    allowsUpdating: $0.paymentMethod.type == .card) }
+                }
+                else {
+                    paymentMethodRows.forEach { $0.state = .editing(allowsRemoval: canRemovePaymentMethods,
+                                                                    allowsUpdating: $0.paymentMethod.isCoBrandedCard && isCBCEligible) }
+                }
```
porter-stripe, line 58: Maybe simplify to just 
```swift
let allowsUpdating = ($0.paymentMethod.isCoBrandedCard && isCBCEligible) || (configuration.defaultSPMNavigation && $0.paymentMethod.type == .card)
paymentMethodRows.forEach { $0.state = .editing(allowsRemoval: canRemovePaymentMethods,
                                                                    allowsUpdating: $0.paymentMethod.isCoBrandedCard && isCBCEligible)
```

Might be some other areas we can do this to

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSheetConfiguration.swift
joyceqin-stripe: I'm not really sure of a better name— it seems like we'd need two separate flags for this project: one specifically to trigger the updated navigation flow and one to opt in to allowing default SPMs. The reasoning I had for why there would be two flags is for the case that would lead us to the remove-only screen, in which case the navigation flag would be on and the set as default flag would be off. My thought was that the navigation flag would eventually be removed altogether as the new navigation flow would completely replace the old one. Maybe there's a better way though

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSavedPaymentMethodsCollectionViewController.swift
joyceqin-stripe: https://github.com/user-attachments/assets/5631b9d1-63a4-45dd-bcc7-e57f07237589

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSheetConfiguration.swift
joyceqin-stripe: Would newUpdatePaymentMethodNavigation be better?

joyceqin-stripe: I've refactored it to alternateUpdatePaymentMethodNavigation, but I'm open to any other change suggestions

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSavedPaymentMethodsCollectionViewController.swift
joyceqin-stripe: This screen recording was before we decided on the changes, but the underlying behavior of the bug was the same. I can do a screen recording from master that shows the same.

joyceqin-stripe: https://github.com/user-attachments/assets/f5f8b05d-99b1-4238-9cd4-1a1d6780413e

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RightAccessoryButton.swift
```diff
@@ -16,29 +16,34 @@ extension RowButton {
...
                 }
             }
         }
 
         private var label: UILabel {
             let label = UILabel()
-            label.text = accessoryType.text
+            if accessoryType != .update {
+                label.text = accessoryType.text
+            }
```
porter-stripe, line 46: I don't understand why we have `.update` returning .Localized.update_card for it's text but then not setting the text? Why not just have nil for the text?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RightAccessoryButton.swift
```diff
@@ -53,7 +58,12 @@ extension RowButton {
         private var imageView: UIImageView? {
             guard let image = accessoryType.accessoryImage else { return nil }
             let imageView = UIImageView(image: image)
-            imageView.tintColor = appearance.colors.primary // TODO(porter) use secondary action color
+            if accessoryType == .update {
+                imageView.tintColor = appearance.colors.icon
+            }
+            else {
+                imageView.tintColor = appearance.colors.primary // TODO(porter) use secondary action color
+            }
```
porter-stripe, line 66: Let's move this out to be a property on `AccessoryType`  like `text` and `accessoryImage`

porter-stripe: Ah nvm, we don't have an appearance object in there, let's leave it.

porter-stripe:
I'm noticing a bug that when the new flag is enabled, I cannot delete a bank account. I'm assuming this will be handled in a future PR, but is this intended for now?

Simulator Screenshot - iOS 18  - 2024-11-19 at 11 38 48

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RightAccessoryButton.swift
joyceqin-stripe: I could have it return nil, but the accessibility label text is set to .update_card

joyceqin-stripe:
> I'm noticing a bug that when the new flag is enabled, I cannot delete a bank account. I'm assuming this will be handled in a future PR, but is this intended for now?
> 
> Simulator Screenshot - iOS 18 - 2024-11-19 at 11 38 48

Yes, I'm changing that in my pr for the bank account screen.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RightAccessoryButton.swift
porter-stripe: I don't think this should have an accessibility label, it's only really an indicator, not a button like "View more" and "Edit"

joyceqin-stripe: That could be valid, my UI test just wouldn't be able to find the buttons by identifier the way it did with CircularButton.Edit or Update card. I can still make the change though

porter-stripe: I think you can update your tests to look for the `accessoryImage` instead, you can see an example of looking for the image in `testSingleCardCBC_update_and_remove_selectStateApplePay`, then have nil text here.

joyceqin-stripe: The image for the chevron doesn't show up in app.images. I've implemented a workaround that sets the accessibility identifier to "chevron" instead

porter-stripe (approved):

PR #4275, opened 2024-11-17
STPFormView and StackViewWithSeparator better SwiftUI support

wooj-stripe:
## Summary
Surrouding higher priority views around STPCardFormView was causing the view to completely collapse depsite already setting contentCompressionResistancePriority.  To remedy, determine the intrinsic content size such that swift ui can lay this view out better.

## Motivation
Adding Spacers() w/ higher priority around this view can cause this view to completely collapse.

## Testing
Manual testing

|  Before | after   |
|---|---|
|  image |  image |


wooj-stripe (commented):
Seems that the spacing is off for a few. I will try and match this up the best i can, and there are some places that I've obviously broken... From a quick review:
- Not sure what's going on with: `testEmptyHiddenPostalCode`
- In testSingleSectionWithTItle -- the title is missing.

wooj-stripe:
https://github.com/stripe/stripe-ios/pull/4284

PR #4280, opened 2024-11-18
Extend UpdateCardViewController to bank accounts and sepa debit

joyceqin-stripe:
## Summary

Refactored UpdateCardViewController to UpdatePaymentMethodViewController and extended its functionality to include US bank accounts and SEPA debit. Created an UpdatePaymentMethodViewModel to determine which form should be shown. Conditionally displays "Save" button and "_ details cannot be changed." label.
## Motivation

[MOBILESDK-2743](https://jira.corp.stripe.com/browse/MOBILESDK-2743)
[MOBILESDK-2662](https://jira.corp.stripe.com/browse/MOBILESDK-2662)
## Testing

https://github.com/user-attachments/assets/49331f2f-1fe7-45bf-91eb-7554c22eecdf
test_UpdatePaymentMethodViewControllerUSBankAccountLightMode@3x
test_UpdatePaymentMethodViewControllerSEPADebitLightMode@3x
test_UpdatePaymentMethodViewControllerLightMode@3x
test_UpdatePaymentMethodViewControllerExpiredCard@3x
test_UpdatePaymentMethodViewControllerRemoveOnlyCard@3x


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/TextField/TextFieldElement+Card.swift
```diff
@@ -361,4 +361,23 @@ extension TextFieldElement {
             return TextFieldElement.PANConfiguration(cardBrandDropDown: cardBrandDropDown).accessoryView(for: lastFourFormatted, theme: theme)
         }
     }
+    struct USBankNumberConfiguration: TextFieldElementConfiguration {
```
porter-stripe, line 364: Let's put this in it's own file, it doesn't belong in the `+Card` file.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/TextField/TextFieldElement+Card.swift
```diff
@@ -361,4 +361,23 @@ extension TextFieldElement {
...
+        let label = String.Localized.bank_account
+        let bankName: String
+        let lastFour: String
+        let isEditable = false
+
+        private var lastFourFormatted: String {
+            "\(bankName) •••• \(lastFour)"
```
porter-stripe, line 371: I notice we have a space between the bullet points and the last 4, typically for bank accounts we don't do this but I see in the figma we do.

Can you raise the point that typically we don't have this space and confirm if this is intentional?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
```diff
@@ -0,0 +1,378 @@
...
+                fatalError("Updating payment method has not been implemented for \(viewModel.paymentMethod.type)")
+            }
+        }()
+        let paymentMethodDetails = UIStackView(arrangedSubviews: [details, detailsCannotBeChangedLabel])
+        paymentMethodDetails.axis = .vertical
+        paymentMethodDetails.setCustomSpacing(8, after: details) // custom spacing from figma
+        return paymentMethodDetails
+    }()
+
+    private lazy var detailsCannotBeChangedLabel: UITextView = {
```
porter-stripe, line 125: In the future I think this could say something else like only card brand can be updated. Maybe name this "footnoteLabel" or something more generic.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
```diff
@@ -0,0 +1,378 @@
...
+        button.titleLabel?.textAlignment = .center
+        button.titleLabel?.font = appearance.scaledFont(for: appearance.font.base.medium, style: .callout, maximumPointSize: 25)
+        button.titleLabel?.adjustsFontForContentSizeCategory = true
+        button.addTarget(self, action: #selector(removePaymentMethod), for: .touchUpInside)
+        button.isHidden = !viewModel.canRemove
+        return button
+    }()
+
+    private lazy var paymentMethodDetails: UIStackView = {
+        lazy var details: UIView = {
```
porter-stripe, line 107: Why does `details` need to be lazy? a lazy inside of a lazy seems smelly.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
```diff
@@ -0,0 +1,378 @@
...
+        button.setTitle(.Localized.remove, for: .normal)
+        button.titleLabel?.textAlignment = .center
+        button.titleLabel?.font = appearance.scaledFont(for: appearance.font.base.medium, style: .callout, maximumPointSize: 25)
+        button.titleLabel?.adjustsFontForContentSizeCategory = true
+        button.addTarget(self, action: #selector(removePaymentMethod), for: .touchUpInside)
+        button.isHidden = !viewModel.canRemove
+        return button
+    }()
+
+    private lazy var paymentMethodDetails: UIStackView = {
```
porter-stripe, line 106: Nit: `paymentMethodDetails` makes this sound like it's an object contains data for a payment method. Maybe rename this to `paymentMethodForm`

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
```diff
@@ -0,0 +1,378 @@
...
+            case .SEPADebit:
+                return sepaDebitSection
+            default:
+                fatalError("Updating payment method has not been implemented for \(viewModel.paymentMethod.type)")
+            }
+        }()
+        let paymentMethodDetails = UIStackView(arrangedSubviews: [details, detailsCannotBeChangedLabel])
+        paymentMethodDetails.axis = .vertical
+        paymentMethodDetails.setCustomSpacing(8, after: details) // custom spacing from figma
+        return paymentMethodDetails
```
porter-stripe, line 122: Feels odd we are making another stack view here, why not just add detailsCannotBeChangedLabel to formStackView and have this be a UIView?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
porter-stripe: Let's make this optional, and instead of hiding it return nil if we don't need to show it. Minor performance bump. We can check the cases where we set hidden to true before creating the label and return nil if we need to.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
```diff
@@ -0,0 +1,378 @@
...
+    }()
+
+    private lazy var errorLabel: UILabel = {
+        let label = ElementsUI.makeErrorLabel(theme: appearance.asElementsTheme)
+        label.isHidden = true
+        return label
+    }()
+
+    private lazy var cardBrandDropDown: DropdownFieldElement? = {
+        if viewModel.paymentMethod.isCoBrandedCard {
```
porter-stripe, line 153: This would be easier to read if we did

```swift
guard viewModel.paymentMethod.isCoBrandedCard else { return nil }
/// continue creating dropdown.
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
```diff
@@ -0,0 +1,378 @@
...
+                cardBrandDropDown.select(index: indexToSelect, shouldAutoAdvance: false)
+            }
+            
+            return cardBrandDropDown
+        }
+        return nil
+    }()
+
+    private lazy var cardSection: SectionElement = {
+        lazy var panElement: TextFieldElement = {
```
porter-stripe, line 182: Same comment about any lazy's inside of a lazy, I haven't seen that before and don't think we need it.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
```diff
@@ -0,0 +1,378 @@
...
+        }()
+        let stackView = UIStackView(arrangedSubviews: [nameElement.view, emailElement.view, bankAccountElement.view])
+        stackView.isLayoutMarginsRelativeArrangement = true
+        stackView.axis = .vertical
+        stackView.setCustomSpacing(8, after: nameElement.view) // custom spacing from figma
+        stackView.setCustomSpacing(8, after: emailElement.view) // custom spacing from figma
+        return stackView
+    }()
+
+    private lazy var sepaDebitSection: UIStackView = {
```
porter-stripe, line 225: I think it would be good to move all the form creation for card, us bank account and sepa into it's own class.

Then it can have a function where you pass it in a view model and it just returns a view.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewModel.swift
```diff
@@ -0,0 +1,49 @@
...
+//  Created by Joyce Qin on 11/15/24.
+//
+
+import Foundation
+@_spi(STP) import StripeCore
+import UIKit
+
+class UpdatePaymentMethodViewModel {
+    let paymentMethod: STPPaymentMethod
+    static let supportedPaymentMethods: [STPPaymentMethodType] = [.card, .USBankAccount, .SEPADebit]
```
porter-stripe, line 14: Nit: On the ordering here, let's put all static variables up on their own and not in between more "normal" variables for readability. 

```swift
class UpdatePaymentMethodViewModel {
    static let supportedPaymentMethods: [STPPaymentMethodType] = [.card, .USBankAccount, .SEPADebit]


    let paymentMethod: STPPaymentMethod
}

File: StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift (hunk 6 above)
Comment (joyceqin-stripe) [reply StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift:122]: The idea of that was that in the Figma, if there's a label, it's 8px from the rest of the form and there are 32px between that group (form + label) and the Save button, whereas if there were no label, there would be just 32px between the form and the button. I could just do conditional spacing in formStackView

File: StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewModel.swift (hunk 10 above)
Comment (joyceqin-stripe) [reply StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewModel.swift:14]: It's used in VerticalSavedPaymentMethodsViewController and SavedPaymentMethodCollectionView when checking if a payment method should allow editint

File: StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift (hunk 6 above)
Comment (porter-stripe) [reply StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift:122]: Conditional spacing on formStackView sounds right to me!

File: StripePaymentSheet/StripePaymentSheet/Source/Categories/String+Localized.swift [hunk 11]
```diff
@@ -101,10 +136,10 @@ extension String.Localized {
         )
     }
 
-    static var update_card_brand: String {
+    static var update_payment_method: String {
         STPLocalizedString(
-            "Update card brand",
-            "Title for a screen for updating a card brand."
+            "Update payment method",
+            "Accessibility label for a button that leads to a screen for updating a payment method."
```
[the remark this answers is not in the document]
joyceqin-stripe: Good to note— I thought accessibility labels did need to be localized since show_menu had a description that said it was an accessibility label for something, but good to know for future reference.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/TextField/TextFieldElement+IBAN.swift
```diff
@@ -179,6 +179,24 @@ extension TextFieldElement {
             }
         }
     }
+
+    struct LastFourIBANConfiguration: TextFieldElementConfiguration {
+        let label: String = STPLocalizedString("IBAN", "Label for an IBAN field")
```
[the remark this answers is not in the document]
joyceqin-stripe: I removed the STPLocalizedString part and just have it as a hardcoded String "IBAN" now.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/TextField/TextFieldElement+Card.swift
joyceqin-stripe: Created a new file, TextFieldElement+Card.swift specifically for USBankNumberConfiguration

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
joyceqin-stripe: Done and done

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
joyceqin-stripe: I didn't have the greatest grasp of what lazy variables were, and I was mostly just moving the elements into another view block. I'll know better in the future.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
joyceqin-stripe: Done

StripeUICore/StripeUICore/Source/Elements/Factories/TextFieldElement+Factory.swift
```diff
@@ -73,14 +75,16 @@ import UIKit
...
         public let defaultValue: String?
         public let isOptional: Bool
+        public let isEditable: Bool
         public let disallowedCharacters: CharacterSet = .whitespacesAndNewlines
         let invalidError = Error.invalid(
             localizedDescription: String.Localized.invalid_email
         )
 
-        init(defaultValue: String? = nil, isOptional: Bool = false) {
+        public init(defaultValue: String? = nil, isOptional: Bool = false, isEditable: Bool = true) {
```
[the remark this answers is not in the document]
joyceqin-stripe: It doesn't— I don't remember what led me to add that in. I've reverted that part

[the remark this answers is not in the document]
joyceqin-stripe: Oh wait actually, I think it was because I was getting inaccessible due to '@_spi' protection level errors

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/CircularButton.swift
```diff
@@ -103,7 +103,7 @@ class CircularButton: UIControl {
             accessibilityIdentifier = "CircularButton.Remove"
         case .edit:
             imageView.image = Image.icon_edit.makeImage(template: true)
-            accessibilityLabel = String.Localized.update_card_brand
```
porter-stripe, line 106: I was wrong here, sorry! I got this mixed up w/ accessibilityIdentifier! Let's localize this, apologies for the churn on this.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/TextField/TextFieldElement+IBAN.swift
```diff
@@ -179,6 +179,24 @@ extension TextFieldElement {
...
         }
     }
+
+    struct LastFourIBANConfiguration: TextFieldElementConfiguration {
+        let label: String = "IBAN"
+        let lastFour: String
+        let isEditable = false
+
+        private var lastFourFormatted: String {
+            "•••• \(lastFour)"
```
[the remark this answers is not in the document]
joyceqin-stripe: I remember being told that the IBAN shouldn't have the space between those digits, but I'll verify in the channel

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
```diff
@@ -0,0 +1,256 @@
...
+            stackView.addArrangedSubview(footnoteLabel)
+            stackView.setCustomSpacing(8, after: paymentMethodForm) // custom spacing from figma
+            stackView.setCustomSpacing(32, after: footnoteLabel) // custom spacing from figma
+        }
+        else {
+            stackView.setCustomSpacing(32, after: paymentMethodForm) // custom spacing from figma
+        }
+        stackView.addArrangedSubview(updateButton)
+        stackView.addArrangedSubview(removeButton)
+        stackView.addArrangedSubview(errorLabel)
```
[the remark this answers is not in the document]
joyceqin-stripe: I don't know how to insert footnoteLabel at an index into the stackView array— addArrangedSubview adds to the end of the array. If I add these to the array on line 50, then when I add footnoteLabel, it will show up below the remove button.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/TextField/TextFieldElement+IBAN.swift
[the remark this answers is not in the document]
joyceqin-stripe: Bella confirmed there shouldn't be a space there.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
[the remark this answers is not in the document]
joyceqin-stripe: Never mind, found out that there's insertArrangedSubview(view, at)

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/SavedPaymentMethodFormFactory.swift
```diff
@@ -0,0 +1,148 @@
...
+                    // pre-select current card brand
+                    if let currentCardBrand = viewModel.paymentMethod.card?.preferredDisplayBrand,
+                       let indexToSelect = cardBrandDropDown.items.firstIndex(where: { $0.rawData == STPCardBrandUtilities.apiValue(from: currentCardBrand) }) {
+                        cardBrandDropDown.select(index: indexToSelect, shouldAutoAdvance: false)
+                        viewModel.selectedCardBrand = currentCardBrand
+                    }
+                    return cardBrandDropDown
+    }()
+
+    private lazy var panElement: TextFieldElement = {
```
porter-stripe, line 65: Probably not in this PR but another one, it might be a good idea to move these all to their own file, it reads kinda hard right now with all the lazy variables (most of which aren't used for certain payment methods). See PaymentSheetFormFactory for what that may look like.

porter-stripe (approved):

PR #4283, opened 2024-11-19
Fix when selecting a saved payment method

porter-stripe:
## Summary
- When we would have something other than the saved payment method selected, then hit "View more" and select a payment method it would not be selected.
- The selected saved payment method should become the payment option to match the behavior of vertical mode and not require users to tap the saved payment method twice.
- I don't believe there to be any scenario where the saved payment method view controller completes with a selected payment method that we would not want it to be selected.

### Before
https://github.com/user-attachments/assets/31081f58-cedb-4003-a161-8bea4154b876

## Motivation
- Bug fix

## Testing
- Manual
- New UI test

## Changelog
N/A

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
+++ b/Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
@@ -408,7 +408,8 @@ class EmbeddedUITests: PaymentSheetUITestCase {
         XCTAssertTrue(app.buttons["View more"].waitForExistenceAndTap())
         XCTAssertTrue(app.buttons["•••• 4242"].waitForExistenceAndTap())
         
-        // Card should now be selected
+        // Verify we have dismissed the saved payment method view and have the correct card selected
+        XCTAssertTrue(app.buttons["View more"].waitForExistence(timeout: 10))
         XCTAssertTrue(app.staticTexts["•••• 4242"].waitForExistence(timeout: 10))
         XCTAssertTrue(app.buttons["Checkout"].isEnabled)
     }
--- a/Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
+++ b/Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
@@ -410,7 +410,7 @@ class EmbeddedUITests: PaymentSheetUITestCase {
         XCTAssertTrue(app.buttons["•••• 4242"].waitForExistenceAndTap())
         
         // Verify we have dismissed the saved payment method view and have the correct card selected
-        XCTAssertTrue(app.buttons["View more"].waitForExistence(timeout: 10))
+        XCTAssertFalse(app.staticTexts["Select payment method"].waitForExistence(timeout: 2))
         XCTAssertTrue(app.staticTexts["•••• 4242"].waitForExistence(timeout: 10))
         XCTAssertTrue(app.buttons["Checkout"].isEnabled)
         
--- a/Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
+++ b/Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
@@ -406,12 +406,31 @@ class EmbeddedUITests: PaymentSheetUITestCase {
         
         // Select a saved card
         XCTAssertTrue(app.buttons["View more"].waitForExistenceAndTap())
+        XCTAssertTrue(app.staticTexts["Select payment method"].waitForExistence(timeout: 10))
         XCTAssertTrue(app.buttons["•••• 4242"].waitForExistenceAndTap())
         
         // Verify we have dismissed the saved payment method view and have the correct card selected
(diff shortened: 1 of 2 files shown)


Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
```diff
@@ -389,6 +389,29 @@ class EmbeddedUITests: PaymentSheetUITestCase {
...
+        
+        // Select Cash App Pay
+        app.buttons["Cash App Pay"].waitForExistenceAndTap()
+        XCTAssertTrue(app.staticTexts["Cash App Pay"].waitForExistence(timeout: 10))
+        XCTAssertTrue(app.buttons["Checkout"].isEnabled)
+        
+        // Select a saved card
+        XCTAssertTrue(app.buttons["View more"].waitForExistenceAndTap())
+        XCTAssertTrue(app.buttons["•••• 4242"].waitForExistenceAndTap())
+
```
wooj-stripe, line 410: Can we validate that the form has been dismissed somehow?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -228,9 +228,11 @@ extension EmbeddedPaymentElement: VerticalSavedPaymentMethodsViewControllerDeleg
             savedPaymentMethods: latestPaymentMethods
         )
 
-        // If there are still saved payment methods & the saved payment method was previously selected to presenting
+        // Select the selected payment method if we selected a payment method and have more than one saved PM
+        // Or if there are still saved payment methods & the saved payment method was previously selected to presenting
         // the list of saved payment methods, then the embedded view should continue to show it is selected, otherwise unselected.
-        let isSelected: Bool = latestPaymentMethods.count > 0 && embeddedPaymentMethodsView.selection?.isSaved ?? false
+        let isSelected = (latestPaymentMethods.count > 1 && selectedPaymentMethod != nil) ||
```
wooj-stripe, line 234: is there any reason to check `latestPaymentMethods.count > 1` ? seems like we can just check `selectedPaymentMethod != nil || (embeddedPaymentMethodsView.selection?.isSaved ?? false && latestPaymentMethods.count > 0)`

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -228,9 +228,11 @@ extension EmbeddedPaymentElement: VerticalSavedPaymentMethodsViewControllerDeleg
             savedPaymentMethods: latestPaymentMethods
         )
 
-        // If there are still saved payment methods & the saved payment method was previously selected to presenting
+        // Select the selected payment method if we selected a payment method and have more than one saved PM
+        // Or if there are still saved payment methods & the saved payment method was previously selected to presenting
```
wooj-stripe, line 232: // Select the *saved payment method if we selected a payment method
// or
//  there are still saved payment methods & the saved payment method was previously selected to presenting

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
porter-stripe: Yes we do need this! The saved payment screen when you delete all but the last payment method auto selects it, even if the user did not tap it. I think i this case we should not auto select that saved payment method in the embedded view. I extended the existing test to cover this now.

wooj-stripe: Ah! right! thank you!

Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
```diff
@@ -389,6 +389,49 @@ class EmbeddedUITests: PaymentSheetUITestCase {
...
+        XCTAssertTrue(app.staticTexts["Cash App Pay"].waitForExistence(timeout: 10))
+        XCTAssertTrue(app.buttons["Checkout"].isEnabled)
+        
+        // Select a saved card
+        XCTAssertTrue(app.buttons["View more"].waitForExistenceAndTap())
+        XCTAssertTrue(app.staticTexts["Select payment method"].waitForExistence(timeout: 10))
+        XCTAssertTrue(app.buttons["•••• 4242"].waitForExistenceAndTap())
+        
+        // Verify we have dismissed the saved payment method view and have the correct card selected
+        XCTAssertTrue(app.buttons["View more"].waitForExistence(timeout: 10))
```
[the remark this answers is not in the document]
porter-stripe: yeah good idea

wooj-stripe (approved):

PR #4288, opened 2024-11-20
Selecting a new saved payment method bug fix

porter-stripe:
## Summary
- If you would select a saved payment method by opening the manage screen (tapping "View more"), then tap new card, then dismiss the card form the row but the previously selected payment method would not be selected.
- This was due to `updateSavedPaymentMethodRow` updating the selection before the saved payment method button.
- This PR improves the selection logic making it less bug prone by keeping a map of selections to row buttons
- We also improve the selection UI tests

## Motivation
- Embedded

## Testing
- Manual
- Improved existing tests

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
@@ -220,6 +220,7 @@ class EmbeddedPaymentMethodsView: UIView {
                                      isSelected: Bool,
                                      accessoryType: RowButton.RightAccessoryButton.AccessoryType?) {
         guard let previousSavedPaymentMethodButton = self.savedPaymentMethodButton,
+              let previousSelection = selectionButtonMapping.first(where: { $0.value == previousSavedPaymentMethodButton })?.key,
               let viewIndex = stackView.arrangedSubviews.firstIndex(of: previousSavedPaymentMethodButton) else {
             stpAssertionFailure("""
             This function should never be called when there isn't already a saved PM row because there's no way for Embedded
@@ -229,9 +230,7 @@ class EmbeddedPaymentMethodsView: UIView {
         }
         
         // Remove old mapping from selectionButtonMapping
-        if let previousSelection = selectionButtonMapping.first(where: { $0.value == previousSavedPaymentMethodButton })?.key {
-            selectionButtonMapping.removeValue(forKey: previousSelection)
-        }
+        selectionButtonMapping.removeValue(forKey: previousSelection)
 
         if let savedPaymentMethod {
             // Replace saved payment method button at same index
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
@@ -25,23 +25,14 @@ class EmbeddedPaymentMethodsView: UIView {
     private let appearance: PaymentSheet.Appearance
     private let rowButtonAppearance: PaymentSheet.Appearance
     private let customer: PaymentSheet.CustomerConfiguration?
+    private var selectionButtonMapping = [Selection: RowButton]()
     private var previousSelection: Selection?
-    private var previousSelectedRowButton: RowButton?
-    private var selectedRowButton: RowButton? {
-        willSet {
-            previousSelectedRowButton = selectedRowButton
-        }
-        didSet {
-            for case let rowButton as RowButton in stackView.arrangedSubviews {
-                rowButton.isSelected = rowButton === selectedRowButton
-            }
-        }
-    }
     private(set) var selection: Selection? {
-        willSet {
-            previousSelection = selection
-        }
         didSet {
+            previousSelection = oldValue
+            selectionButtonMapping.forEach { (key, button) in
+                button.isSelected = key == selection
+            }
             updateMandate()
             delegate?.selectionDidUpdate(didChange: oldValue != selection)
         }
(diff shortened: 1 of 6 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
```diff
@@ -238,6 +227,11 @@ class EmbeddedPaymentMethodsView: UIView {
             """)
             return
         }
+        
+        // Remove old mapping from selectionButtonMapping
+        if let previousSelection = selectionButtonMapping.first(where: { $0.value == previousSavedPaymentMethodButton })?.key {
+            selectionButtonMapping.removeValue(forKey: previousSelection)
+        }
```
wooj-stripe, line 234: This function is only called when there was a previouslySavedPaymentMethod button, which, if i understand correctly means that there must exist a mapping in `selectionButtonMapping`, correct?

If so, seems like `let previousSelection = selectionButtonMapping.first(where..` should be part of the guard statement?

porter-stripe: Yep I think you're right, I don't think this would ever be called when we don't have an existing mapping to a saved button. This step is pretty much optional/nice-to-have/clean up that we should do. I think we could actually get away w/o doing it at all.

Since it's mostly optional, I don't feel like it should block the execution of the rest of the function, even though I can't think of a state where that would happen. Feels slightly safer outside the guard to me.

wooj-stripe: If this function's precondition is to have `self.savedPaymentMethodButton` set, it seems that there would be a bug if there was no entry in selectionButtonMapping, correct?

Seems like moving it to the guard would strengthen the guard condition and has the potential to catch bugs if this condition is no longer being met.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/VerticalPaymentMethodListViewController.swift
```diff
@@ -230,6 +230,21 @@ enum VerticalPaymentMethodListSelection: Equatable {
...
+            hasher.combine(paymentMethodType.identifier)
+        case .saved(let paymentMethod):
+            hasher.combine(1)
+            hasher.combine(paymentMethod.stripeId)
+        case .applePay:
+            hasher.combine(2)
+        case .link:
+            hasher.combine(3)
+        }
+    }
```
wooj-stripe, line 247: Is this needed? I can compile without this function?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
porter-stripe: Hrm, I don't feel strongly. I'll add it.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/VerticalPaymentMethodListViewController.swift
porter-stripe: Yeah it compiles without this function, but the auto synthesized version is a black box. Given we have non trivial associated values such as `PaymentSheet.PaymentMethodType` and `STPPaymentMethod`, I'd prefer to define how and what we hash here rather than hoping the auto synthesized one is what we want.

Similarly we could remove the `static func == (lhs: VerticalPaymentMethodListSelection, rhs: VerticalPaymentMethodListSelection) -> Bool` and keep the `Equatable` conformance and it still compiles.

porter-stripe: In fact, if you remove this function and run `EmbeddedUITest.testSelection()` it fails!

wooj-stripe: Hmm, I'd argue that unless it is absolutely needed, it should be deleted, so in this case, I think it would be better to delete both functions.

Looking at other examples in the code base, we tend to do it for specific reasons:
  PaymentMethodAvailabilityStatus: Equtable -- needed because we don't care about ordering.
  EmbeddedPaymentElement.UpdateResult: Equatable -- needed because underlying types are not Equatable.

wooj-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/VerticalPaymentMethodListViewController.swift
porter-stripe: Bringing this out from Slack convo

```swift
    func testHashSelection() {
        let selection = VerticalPaymentMethodListSelection.saved(paymentMethod: ._testCard())
        let selection1 = VerticalPaymentMethodListSelection.saved(paymentMethod: ._testCard())
        
        // Fails with auto synthesized Hashable implementation, passes with custom
        XCTAssertEqual(selection.hashValue, selection1.hashValue)
    }
    
    func testHashSelection_sameInstance() {
        let testCard: STPPaymentMethod = ._testCard()
        let selection = VerticalPaymentMethodListSelection.saved(paymentMethod: testCard)
        let selection1 = VerticalPaymentMethodListSelection.saved(paymentMethod: testCard)
        
        // Passes with auto synthesized Hashable implementation, passes with custom
        XCTAssertEqual(selection.hashValue, selection1.hashValue)
    }
```

PR #4300, opened 2024-11-25
Embedded quality review pt. 1

porter-stripe:
## Summary
- Increase row height to 52
- Fixed a crash when removing the last saved payment method in floating style (there are no separators)
- Reduce mandate font size by 1pt to 11pt
- Hide selected checkmark in saved PM VC when showing in embedded
- Round image view for US bank account when shown in saved
- When removing last card change card row to "Card" instead of "New card"

## Motivation
- https://docs.google.com/document/d/1CIt73R5FFuV_si3Y9UUnTRsUoIrGeqO9gXMtqtZdDps/edit?tab=t.0

## Testing
- Manual
- CI

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
```diff
@@ -253,7 +260,9 @@ class EmbeddedPaymentMethodsView: UIView {
         } else {
             // No more saved payment methods
             let separatorIndex = stackView.arrangedSubviews.index(before: viewIndex)
-            stackView.removeArrangedSubview(at: separatorIndex, animated: false)
+            if separatorIndex >= 0 {
+                stackView.removeArrangedSubview(at: separatorIndex, animated: false)
+            }
```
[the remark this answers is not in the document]
wooj-stripe: Chatted offline -- due to the floating version

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
```diff
@@ -52,7 +52,14 @@ class EmbeddedPaymentMethodsView: UIView {
         return stackView
     }()
 
-    private lazy var mandateView = SimpleMandateTextView(theme: appearance.asElementsTheme)
+    private lazy var mandateView = {
+        // Special font size for mandates in embedded
+        var theme = appearance.asElementsTheme
+        theme.fonts.caption = UIFont.systemFont(ofSize: 11, weight: .regular).scaled(
+            withTextStyle: .caption2,
+            maximumPointSize: 20)
```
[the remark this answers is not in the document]
porter-stripe: Good catch, fixed this!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
```diff
@@ -263,6 +272,30 @@ class EmbeddedPaymentMethodsView: UIView {
             // Update instance states
             self.savedPaymentMethodButton = nil
         }
+        
+        // Update text on card row to say just "Card" instead of "New card" if needed
+        if let oldCardButton = selectionButtonMapping[.new(paymentMethodType: .stripe(.card))],
```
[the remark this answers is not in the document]
wooj-stripe: Sorry, I'm not understanding when we should show "new card" vs "card". It's confusing because assuming a user has two payment methods (1) bank account and (1) card, switching between the payment methods changes "card" -> "new card" , and it looks like a bug-- but i'm guessing this is expected behavior?

1. Can we sync offline about this?
2. If this is really a requirement, we should add some UI tests to test for this.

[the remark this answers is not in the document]
porter-stripe: > Sorry, I'm not understanding when we should show "new card" vs "card"

Whenever the selected saved payment method (if any) is a card we should display "New card", otherwise just "Card". Note: In this context "selected saved payment method" refers to what the saved PM row is displaying it may not actually be "selected" in the embedded view.

>  switching between the payment methods changes "card" -> "new card" , and it looks like a bug-- but i'm guessing this is expected behavior

Yep this is expected behavior, the previous behavior was a bug of sorts, check vertical mode for expected behavior. The current behavior on master is actually somewhat inconsistent, the copy on the card button depends entirely on the default saved payment method at load time. So if bank is selected by default it is just "Card", but if card is selected by default it is "New card". But this copy never updates depending on the user interactions.

> If this is really a requirement, we should add some UI tests to test for this.

Yeah, this behavior should match vertical mode, so it makes sense to me that it is a requirement and was called out in the quality review.

re tests: We did update a [UI test](https://github.com/stripe/stripe-ios/pull/4300/files#diff-1528ed9f3c25572a57550dea4569a6e9e75a440611c0d09d316ba1f4fc7101e8R236) for this change, and many of our UI tests depend on the text on this button. The updated UI test exercises the exact flow that was called out in quality review. Then we have snapshot tests in `EmbeddedPaymentMethodsViewSnapshotTests`. I think that's probably enough tests for the slight copy difference. I don't think it warrants a dedicated UI test above what we already have IMO.


Happy to sync offline if needed!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
```diff
@@ -217,6 +224,7 @@ class EmbeddedPaymentMethodsView: UIView {
     }
 
     func updateSavedPaymentMethodRow(_ savedPaymentMethod: STPPaymentMethod?,
+                                     _ savedPaymentMethods: [STPPaymentMethod],
```
wooj-stripe, line 227: Seems we can remove `_ savedPaymentMethod: STPPaymentMethod?,` as an argument?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
```diff
@@ -263,6 +273,31 @@ class EmbeddedPaymentMethodsView: UIView {
             // Update instance states
             self.savedPaymentMethodButton = nil
         }
+        
+        // Update text on card row based on the new selected payment method
+        // It can vary between "Card" and "New card" depending on the payment method type of the selected saved payment method
```
wooj-stripe, line 278: update comment to say that it should say "card" if there are no saved cards and "new card" if there is at least one saved card.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
```diff
@@ -95,7 +102,7 @@ class EmbeddedPaymentMethodsView: UIView {
             let selection: Selection = .new(paymentMethodType: .stripe(.card))
             let cardRowButton = RowButton.makeForPaymentMethodType(
                 paymentMethodType: .stripe(.card),
-                savedPaymentMethodType: savedPaymentMethod?.type,
+                hasSavedCard: !savedPaymentMethods.filter{$0.type == .card}.isEmpty,
```
wooj-stripe, line 105: To avoid duplicating this logic, we can add:
```
extension Array where Element == STPPaymentMethod {
    func hasSavedCard() -> Bool {
        return !self.filter{$0.type == .card}.isEmpty
    }
}
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
porter-stripe: yeah probably can

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
porter-stripe: cool! made it a var though

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/VerticalPaymentMethodListViewController.swift
```diff
@@ -114,7 +114,7 @@ class VerticalPaymentMethodListViewController: UIViewController {
             let rowButton = RowButton.makeForPaymentMethodType(
                 paymentMethodType: paymentMethodType,
                 subtitle: Self.subtitleText(for: paymentMethodType),
-                savedPaymentMethodType: savedPaymentMethod?.type,
+                hasSavedCard: savedPaymentMethod?.type == .card, // TODO(RUN_MOBILESDK-3708)
```
wooj-stripe, line 117: By passing in `savedPaymentMethod?.type == .card` we end up showing "New Card", if we want to do this work later as part of 3708, and continue to keep this as just "Card", seems like we just need to pass `false` into here?

wooj-stripe: Nevermind - this will maintain the existing behavior, and the jira is actually fix an existing bug where we show "card" even if we have a saved card that isn't selected.

wooj-stripe (approved):

PR #4302, opened 2024-11-25
Don't show selected in embedded view with forms unless we have a vali…

porter-stripe:
…d payment option

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
@@ -93,7 +93,7 @@ extension EmbeddedPaymentElement: EmbeddedPaymentMethodsViewDelegate {
         delegate?.embeddedPaymentElementDidUpdateHeight(embeddedPaymentElement: self)
     }
 
-    func selectionDidUpdate(didChange: Bool) {
+    func selectionDidUpdate(didChange: Bool) -> Bool {
         // Deferring notifying delegate until the exit of this function guarantees the new payment option comes from the new instance of `EmbeddedFormViewController`
         defer {
             if didChange {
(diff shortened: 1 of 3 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -127,14 +127,16 @@ extension EmbeddedPaymentElement: EmbeddedPaymentMethodsViewDelegate {
...
             self.formViewController = nil  // Clear out any previous form view controller to update self._paymentOption
-            return
+            return true
         }
 
         let bottomSheet = bottomSheetController(with: formViewController)
         delegate?.embeddedPaymentElementWillPresent(embeddedPaymentElement: self)
         presentingViewController.presentAsBottomSheet(bottomSheet, appearance: configuration.appearance)
         self.formViewController = formViewController
+        return formViewController.selectedPaymentOption != nil // Don't show row as selected when presenting a form unless we have a valid payment option
```
wooj-stripe, line 137: Nit, this might improve readability
```
        let formHasValidPaymentOption = formViewController.selectedPaymentOption != nil
        return formHasValidPaymentOption // Show row selected only if payment option is valid
```

wooj-stripe: I may have found a bug... It's possible to fill out a payment method, then tap outside the sheet, which results in no payment method being selected.  Will DM you with a video

porter-stripe: Good catch on this bug! It actually has some unwanted behavior on master too. The fix was quite simple since we have access to the previous payment option on the form VC.

Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
```diff
@@ -482,8 +490,8 @@ class EmbeddedUITests: PaymentSheetUITestCase {
         // Try to fill a card
         app.buttons["New card"].waitForExistenceAndTap()
         XCTAssertTrue(app.staticTexts["Add new card"].waitForExistence(timeout: 10))
-        try! fillCardData(app, cardNumber: "5555555555554444", postalEnabled: true)
-        app.toolbars.buttons["Done"].waitForExistenceAndTap()
+        XCTAssertTrue(app.buttons["New card"].isSelected) // Should show selected, the form should be filled with a valid card from above
```
porter-stripe, line 493: This should now show as selected since we are presenting the form with a filled out card form (valid payment option)

wooj-stripe (approved):

PR #4313, opened 2024-11-27
Read the default_payment_method field from elements session and use that as the default PM

joyceqin-stripe:
## Summary

Added allowsSetAsDefaultPM flag. Only shown if we're using customer session.
If allowsSetAsDefaultPM is on, then instead of reading from the locally stored default, we read from ElementsSession.
## Motivation

[MOBILESDK-2799](https://jira.corp.stripe.com/browse/MOBILESDK-2799)
## Testing

customer_session, allowsSetAsDefaultPM on
Hardcoded the value that was "read" from default_payment_method (set to the 4242 Visa stripe id)
After changing the selected payment method and paying with it, upon reload the PaymentSheet default is back to our hardcoded default.

https://github.com/user-attachments/assets/38601a0e-513b-40a8-a3d8-f30069723f97


Example/PaymentSheet Example/PaymentSheet Example/CustomerSheetTestPlayground.swift
```diff
@@ -53,6 +53,9 @@ struct CustomerSheetTestPlayground: View {
                                 SettingPickerView(setting: $playgroundController.settings.paymentMethodRemove)
                                 SettingPickerView(setting: $playgroundController.settings.paymentMethodRemoveLast)
                                 SettingPickerView(setting: $playgroundController.settings.paymentMethodAllowRedisplayFilters)
+                                if playgroundController.settings.alternateUpdatePaymentMethodNavigation == .on {
```
porter-stripe, line 56: When can we remove alternateUpdatePaymentMethodNavigation?

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/ElementsCustomer.swift
```diff
@@ -36,6 +36,8 @@ struct ElementsCustomer: Equatable, Hashable {
         }
 
         // Optional
+        // to test default payment methods reading from back end, hard-code a valid default payment method
+        // later, when API calls to get and update default payment method are available, that will no longer be needed
```
porter-stripe, line 40: This comment doesn't make much sense to me, I don't see any code changes near this comment indicating we are hard coding a default PM.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSessionAdapter/CustomerSessionAdapter.swift
```diff
@@ -107,7 +107,13 @@ extension CustomerSessionAdapter {
         return stripePaymentMethodId
     }
 
-    func fetchSelectedPaymentOption(for customerId: String) -> CustomerPaymentOption? {
+    func fetchSelectedPaymentOption(for customerId: String, elementsSession: STPElementsSession? = nil) -> CustomerPaymentOption? {
+        if configuration.allowsSetAsDefaultPM,
```
porter-stripe, line 111: Nit; I think a guard here is more appropriate as it enforces that you return a value: 
```swift
guard configuration.allowsSetAsDefaultPM,
      let elementsSession = elementsSession,
      let customer = elementsSession.customer,
      let defaultPaymentMethod = customer.defaultPaymentMethod else {
    return CustomerPaymentOption.defaultPaymentMethod(for: customerId)
}

return CustomerPaymentOption.stripeId(defaultPaymentMethod)
```

porter-stripe: Also, it seems unnecessary to pass in the entire `STPElementsSession` here, we only really need the customer.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSheetConfiguration.swift
```diff
@@ -86,6 +86,11 @@ extension CustomerSheet {
         /// If false (default), only card brand choice eligible cards can be edited and users can remove payment methods from the list screen.
         @_spi(AlternateUpdatePaymentMethodNavigation) public var alternateUpdatePaymentMethodNavigation = false
 
+        /// This is an experimental feature that may be removed at any time.
+        /// If true, users can set a payment method as default.
```
porter-stripe, line 90: We should add that this also influences where we read the default payment method from, not just allow setting.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSheet.swift
```diff
@@ -326,7 +326,25 @@ extension CustomerSheet {
         switch customerSheetDataSource.dataSource {
         case .customerSession(let customerSessionAdapter):
             let (elementsSession, customerSessionClientSecret) = try await customerSessionAdapter.elementsSessionWithCustomerSessionClientSecret()
-            let selectedPaymentOption = CustomerPaymentOption.defaultPaymentMethod(for: customerSessionClientSecret.customerId)
+
+            var selectedPaymentOption: CustomerPaymentOption?
+
+            // read from back end
```
porter-stripe, line 332: I think this comment could be better!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSheet.swift
```diff
@@ -326,7 +326,25 @@ extension CustomerSheet {
...
+                if let defaultPaymentMethod = defaultPaymentMethod {
+                    selectedPaymentOption = CustomerPaymentOption.stripeId(defaultPaymentMethod.stripeId)
+                }
+                else {
+                    selectedPaymentOption = CustomerPaymentOption.defaultPaymentMethod(for: customerSessionClientSecret.customerId)
+                }
+            }
+            else {
+                selectedPaymentOption = CustomerPaymentOption.defaultPaymentMethod(for: customerSessionClientSecret.customerId)
+            }
```
porter-stripe, line 347: Having multiple code paths/branches that lead to the same code `CustomerPaymentOption.defaultPaymentMethod(for: customerSessionClientSecret.customerId)` is smelly. I think there's a better way to write this.

Any reason we can't do
```swift
if configuration.allowsSetAsDefaultPM,
   let customer = elementsSession.customer,
   let defaultPaymentMethodId = customer.defaultPaymentMethod,
   let defaultPaymentMethod = customer.paymentMethods.first(where: { $0.stripeId == defaultPaymentMethodId }) {
    selectedPaymentOption = .stripeId(defaultPaymentMethod.stripeId)
} else {
    selectedPaymentOption = .defaultPaymentMethod(for: customerSessionClientSecret.customerId)
}
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElementConfiguration.swift
```diff
@@ -136,6 +136,11 @@ extension EmbeddedPaymentElement {
         /// If false (default), only card brand choice eligible cards can be edited and users can remove payment methods from the list screen.
         @_spi(AlternateUpdatePaymentMethodNavigation) public var alternateUpdatePaymentMethodNavigation = false
 
+        /// This is an experimental feature that may be removed at any time.
+        /// If true, users can set a payment method as default.
+        /// If false (default), users cannot set default payment methods.
+        @_spi(AllowsSetAsDefaultPM) public var allowsSetAsDefaultPM = false
```
porter-stripe, line 142: Same comment from about in the CustomerSheetConfiguration. Also same in PaymentSheetConfiguration.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -37,6 +37,18 @@ extension EmbeddedPaymentElement {
...
+               let customer =  loadResult.elementsSession.customer {
+                let defaultPaymentMethod = customer.paymentMethods.filter {
+                    $0.stripeId == customer.defaultPaymentMethod
+                }.first
+                if let defaultPaymentMethod = defaultPaymentMethod {
+                    return .saved(paymentMethod: defaultPaymentMethod)
+                }
+                
+            }
+
```
porter-stripe, line 51: Somewhat similar comment to above, about the "read from back end" comment and the nested if's

Maybe:
```swift
if configuration.allowsSetAsDefaultPM,
   let customer = loadResult.elementsSession.customer,
   let defaultPaymentMethodId = customer.defaultPaymentMethod,
   let defaultPaymentMethod = customer.paymentMethods.first(where: { $0.stripeId == defaultPaymentMethodId }) {
    return .saved(paymentMethod: defaultPaymentMethod)
}
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift
```diff
@@ -316,9 +318,25 @@ final class PaymentSheetLoader {
 
         // Move default PM to front
         if let customerID = configuration.customer?.id {
-            let defaultPaymentMethod = CustomerPaymentOption.defaultPaymentMethod(for: customerID)
+            var defaultPaymentMethodOption: CustomerPaymentOption?
+            // read from back end
+            if configuration.allowsSetAsDefaultPM,
+               let customer =  elementsSession.customer {
```
porter-stripe, line 324: Similar here w/ the comment and nested if's and multiple branches leading to the same `CustomerPaymentOption.defaultPaymentMethod(for: customerID)`. I think there's a cleaner way to write this.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentOptionsViewController.swift
```diff
@@ -438,9 +444,22 @@ class SavedPaymentOptionsViewController: UIViewController {
 
     /// Creates the list of viewmodels to display in the "saved payment methods" carousel e.g. `["+ Add", "Apple Pay", "Link", "Visa 4242"]`
     /// - Returns defaultSelectedIndex: The index of the view model that is the default e.g. in the above list, if "Visa 4242" is the default, the index is 3.
-    static func makeViewModels(savedPaymentMethods: [STPPaymentMethod], customerID: String?, showApplePay: Bool, showLink: Bool) -> (defaultSelectedIndex: Int, viewModels: [Selection]) {
+    static func makeViewModels(savedPaymentMethods: [STPPaymentMethod], customerID: String?, showApplePay: Bool, showLink: Bool, allowsSetAsDefaultPM: Bool, elementsSession: STPElementsSession) -> (defaultSelectedIndex: Int, viewModels: [Selection]) {
```
porter-stripe, line 447: Similar to how I don't think we need to pass in the entire ElementsSession, maybe just the customer or default PM. That way we can consolidate all this unwrapping/duplicate code.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetVerticalViewController.swift
```diff
@@ -288,6 +288,16 @@ class PaymentSheetVerticalViewController: UIViewController, FlowControllerViewCo
...
+            // read from back end
+            if configuration.allowsSetAsDefaultPM,
+               let customer =  elementsSession.customer {
+                let defaultPaymentMethod = customer.paymentMethods.filter {
+                    $0.stripeId == customer.defaultPaymentMethod
+                }.first
+                if let defaultPaymentMethod = defaultPaymentMethod {
+                    return .saved(paymentMethod: defaultPaymentMethod)
+                }
+            }
```
porter-stripe, line 300: This code is duplicated in a few spots, I think we should consolidate this logic into one spot and pass around the defaultPM or something.

porter-stripe (commented):
There is a lot of duplicated logic that is very similar, we should try to consolidate that into one place as much as possible.

Example/PaymentSheet Example/PaymentSheet Example/CustomerSheetTestPlayground.swift
joyceqin-stripe: I suppose if we're trying to release this is the default flow, it should be removed now

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/ElementsCustomer.swift
joyceqin-stripe: What I was trying to say is that the only way I knew to test it was to add a payment method, get its stripe ID, and then hardcode that here so that the elements session will set that as its default. And then after adding other cards, even after paying with them, when the payment sheet is reloaded, it will have the hardcoded default selected, which shows that it's reading from the elements session instead of the local default.

Right now, there's a duping process where when you read from "default_payment_method", it returns the locally saved default that you pass into the API call— it's not actually retrieving a default from the back end. When the API endpoints are ready, then we should be able to actually get the value stored in the back end, but until then, the only way I could think to test that it's reading from the elementsSession value instead of the local default was to hard-code a default (e.g. let defaultPaymentMethod = "pm_1QSOUILu5o3P18ZpN4PY49SE")

joyceqin-stripe:
> There is a lot of duplicated logic that is very similar, we should try to consolidate that into one place as much as possible.

I've made a static func getDefaultPaymentMethod(from customer: ElementsCustomer?) -> STPPaymentMethod? in ElementsCustomer.swift

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSheet.swift
joyceqin-stripe: Changed it to "get default payment method from elements session"— is that better?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSheetConfiguration.swift
joyceqin-stripe: What about "users can set a payment method as default and sync their default payment method across web and mobile"

Example/PaymentSheet Example/PaymentSheet Example/CustomerSheetTestPlayground.swift
joyceqin-stripe: I do anticipate possibly having to update a lot of tests though, so might be changing a lot of files

joyceqin-stripe:
Okay, I've extracted reading the default payment method from the elements to a function in ElementsCustomer, edited some of the comments, and changed some parameters from ElementsSession to ElementsCustomer

porter-stripe (commented):
Let's make sure we are also adding tests for this new functionality.

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/ElementsCustomer.swift
```diff
@@ -36,7 +36,14 @@ struct ElementsCustomer: Equatable, Hashable {
...
         let defaultPaymentMethod = response["default_payment_method"] as? String
         return ElementsCustomer(paymentMethods: paymentMethods, defaultPaymentMethod: defaultPaymentMethod, customerSession: customerSession)
     }
+
+    static func getDefaultPaymentMethod(from customer: ElementsCustomer?) -> STPPaymentMethod? {
+        guard let customer = customer else { return nil }
+        return customer.paymentMethods.first { $0.stripeId == customer.defaultPaymentMethod }
+    }
```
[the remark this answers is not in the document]
joyceqin-stripe: I can't really remember any reasoning of mine, you're right that it should be an instance function on ElementsCustomer

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -37,6 +37,12 @@ extension EmbeddedPaymentElement {
             isFlatCheckmarkStyle: configuration.appearance.embeddedPaymentElement.row.style == .flatWithCheckmark
         )
         let initialSelection: EmbeddedPaymentMethodsView.Selection? = {
+            // get default payment method from elements session
+            if configuration.allowsSetAsDefaultPM,
+               let defaultPaymentMethod = ElementsCustomer.getDefaultPaymentMethod(from: loadResult.elementsSession.customer) {
+                    return .saved(paymentMethod: defaultPaymentMethod)
+            }
+
```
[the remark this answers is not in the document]
joyceqin-stripe: You're right, I didn't consider that the view shouldn't automatically initially select the default because there could have been previous user input. I'll fix that

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -60,7 +54,19 @@ extension EmbeddedPaymentElement {
...
-            let customerDefault = CustomerPaymentOption.defaultPaymentMethod(for: configuration.customer?.id)
+            var customerDefault: CustomerPaymentOption?
+            // if opted in to the "set as default" feature, try to get default payment method from elements session
+            if configuration.allowsSetAsDefaultPM {
+                if let defaultPaymentMethod = loadResult.elementsSession.customer?.getDefaultOrFirstPaymentMethod() {
+                    customerDefault = CustomerPaymentOption.stripeId(defaultPaymentMethod.stripeId)
+                }
+                else {
+                    customerDefault = nil
+                }
```
porter-stripe, line 65: We should be able to remove this, customerDefault is implied `nil` from it's definition. Same with other similar code paths in the PR.

porter-stripe (commented):
Almost there, one small clean up

porter-stripe (approved):

PR #4322, opened 2024-12-02
Pass incentive information to IBP flow

tillh-stripe:
## Summary

This pull request passes the incentive information into the Instant Bank Payments flow. We set an `incentiveEligibilitySession` if the session is eligible, or `nil` otherwise.

## Motivation

[CONSUMERBANK-569](https://jira.corp.stripe.com/browse/CONSUMERBANK-569)

### DIFF
--- a/StripeCore/StripeCore/Source/Connections Bindings/ElementsSessionContext.swift
+++ b/StripeCore/StripeCore/Source/Connections Bindings/ElementsSessionContext.swift
@@ -12,6 +12,14 @@ import Foundation
     @_spi(STP) @frozen public enum IntentID {
         case payment(String)
         case setup(String)
+        case deferred(String)
+        
+        @_spi(STP) public var id: String {
+            switch self {
+            case let .payment(string), let .setup(string), let .deferred(string):
+                return string
+            }
+        }
     }
 
     /// These fields will be used to prefill the Financial Connections Link Login pane.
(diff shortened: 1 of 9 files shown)


tillh-stripe:
Applied `skip dead code check` because the check didn’t work correctly.

StripeCore/StripeCore/Source/Connections Bindings/ElementsSessionContext.swift
```diff
@@ -12,6 +12,14 @@ import Foundation
     @_spi(STP) @frozen public enum IntentID {
         case payment(String)
         case setup(String)
+        case deferred(String)
+        
+        @_spi(STP) public var id: String {
+            switch self {
+            case let .payment(string), let .setup(string), let .deferred(string):
+                return string
```
mats-stripe, line 20: Very minor:

```suggestion
            case let .payment(id), let .setup(id), let .deferred(id):
                return id
```

mats-stripe (approved):

StripeCore/StripeCore/Source/Connections Bindings/ElementsSessionContext.swift
tillh-stripe: ok ok ok ok 😉

tillh-stripe: ok ok ok ok 😉

mats-stripe (approved):

PR #4326, opened 2024-12-04
Improve console logging for Instant Bank Payments and Panther

mats-stripe:
## Summary

This adds some console logging to help users debug why they might not be seeing Instant Bank Payment or Link Card Brands as payment methods in MPE. This is the result of feedback from a user who was trying to integrate, and couldn't figure out why these payment methods weren't appearing.

In order to not pollute the logs with unnecessary information, this only logs these events when a "primary condition" is met. Meaning, we suspect the user is trying to add these payment methods, but the other requirements are not met.

## Motivation

Improve debugability (<- new word) of these payment methods

## Testing

image

## Changelog

N/a

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
@@ -141,6 +141,14 @@ extension PaymentSheet {
         ///   - configuration: A `PaymentSheet` configuration.
         static func filteredPaymentMethodTypes(from intent: Intent, elementsSession: STPElementsSession, configuration: PaymentElementConfiguration, logAvailability: Bool = false) -> [PaymentMethodType]
         {
+            func logIfNecessary(_ message: String) {
+                if logAvailability {
+                    #if DEBUG
+                    print("[Stripe SDK]: \(message)")
+                    #endif
+                }
+            }
+
             var recommendedStripePaymentMethodTypes = elementsSession.orderedPaymentMethodTypes
             recommendedStripePaymentMethodTypes = recommendedStripePaymentMethodTypes.filter { paymentMethodType in
                 let availabilityStatus = PaymentSheet.PaymentMethodType.supportsAdding(
@@ -151,11 +159,9 @@ extension PaymentSheet {
                     supportedPaymentMethods: PaymentSheet.supportedPaymentMethods
                 )
 
-                if logAvailability && availabilityStatus != .supported {
+                if availabilityStatus != .supported {
                     // This payment method is being filtered out, log the reason/s why
-                    #if DEBUG
-                    print("[Stripe SDK]: PaymentSheet could not offer \(paymentMethodType.displayName):\n\t* \(availabilityStatus.debugDescription)")
-                    #endif
+                    logIfNecessary("PaymentSheet could not offer \(paymentMethodType.displayName):\n\t* \(availabilityStatus.debugDescription)")
                 }
 
                 return availabilityStatus == .supported
@@ -164,8 +170,8 @@ extension PaymentSheet {
             // Log a warning if elements session doesn't contain all the merchant's desired external payment methods
             let missingExternalPaymentMethods = Set(configuration.externalPaymentMethodConfiguration?.externalPaymentMethods.map { $0.lowercased() } ?? [])
                 .subtracting(Set(elementsSession.externalPaymentMethods.map { $0.type }))
-            if logAvailability && !missingExternalPaymentMethods.isEmpty {
-                print("[Stripe SDK]: PaymentSheet could not offer these external payment methods: \(missingExternalPaymentMethods). See https://stripe.com/docs/payments/external-payment-methods#available-external-payment-methods")
+            if !missingExternalPaymentMethods.isEmpty {
+                logIfNecessary("PaymentSheet could not offer these external payment methods: \(missingExternalPaymentMethods). See https://stripe.com/docs/payments/external-payment-methods#available-external-payment-methods")
             }
 
             // The full ordered list of recommended payment method types:
@@ -203,12 +209,8 @@ extension PaymentSheet {
             switch instantBankPaymentsAvailability {
             case .supported:
                 eligibleForInstantDebits = true
-            case .primaryRequirementMetButMissingOtherRequirements:
-                if logAvailability {
-                    #if DEBUG
-                    print("[Stripe SDK]: PaymentSheet could not offer Instant Bank Payments: \(instantBankPaymentsAvailability.debugDescription)")
-                    #endif
-                }
+            case .missingRequirements:
+                logIfNecessary("PaymentSheet could not offer the 'Bank' payment method type. See https://docs.stripe.com/payments/link/instant-bank-payments. Missing requirements: \(instantBankPaymentsAvailability.debugDescription)")
                 fallthrough
             default:
                 eligibleForInstantDebits = false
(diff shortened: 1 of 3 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
```diff
@@ -165,51 +165,73 @@ extension PaymentSheet {
...
-                    recommendedPaymentMethodTypes.append(.linkCardBrand)
+            case .primaryRequirementMetButMissingOtherRequirements:
+                if logAvailability {
+                    #if DEBUG
+                    print("[Stripe SDK]: PaymentSheet could not offer the Link Card integration: \(linkCardBrandAvailability.debugDescription)")
+                    #endif
                 }
+                fallthrough
+            default:
+                eligibleForLinkCardBrand = false
```
tillh-stripe, line 227: Should this log `[Stripe SDK]: PaymentSheet is not adding the 'Bank' payment method type: \(description)`?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
```diff
@@ -165,51 +165,73 @@ extension PaymentSheet {
...
+            )
+
+            let eligibleForInstantDebits: Bool
+            switch instantBankPaymentsAvailability {
+            case .supported:
+                eligibleForInstantDebits = true
+            case .primaryRequirementMetButMissingOtherRequirements:
+                if logAvailability {
+                    #if DEBUG
+                    print("[Stripe SDK]: PaymentSheet could not offer Instant Bank Payments: \(instantBankPaymentsAvailability.debugDescription)")
```
tillh-stripe, line 199:
```suggestion
                    print("[Stripe SDK]: PaymentSheet could not offer the 'Bank' payment method type: \(instantBankPaymentsAvailability.debugDescription)")
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
```diff
@@ -320,6 +342,130 @@ extension PaymentSheet {
...
+
+            // - We collect an email, or a default non-empty email has been provided.
+            if !configuration.isEligibleForBankTab {
+                missingRequirements.insert(.instantBankPaymentRequirement(.invalidEmailCollectionConfiguration))
+            }
+
+            let hasMissingRequirements = !missingRequirements.isEmpty
+            if primaryRequirementMet, hasMissingRequirements {
+                return .primaryRequirementMetButMissingOtherRequirements(
+                    "The Link mode was set to Link Card Brand, but other requirements were not met for the Link Card integration. See https://docs.stripe.com/payments/link/link-payment-integrations?link-integrations=link-card-integrations",
```
tillh-stripe, line 459: With the other suggestions, do we need this statement still?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+PaymentMethodAvailability.swift
```diff
@@ -172,11 +175,36 @@ extension PaymentSheet {
...
+            case .unexpectedUsBankAccount:
+                return "US Bank Account should not be set as a payment method"
+            case .invalidEmailCollectionConfiguration:
+                return "The provided configuration must either collect an email, or a default non-empty email must be provided"
+            case .linkFundingSourcesMissingBankAccount:
+                return "Link funding sources must contain bank account"
+            case .unexpectedLinkMode:
+                return "The Link Mode received is not Link Card Brand"
+            }
+        }
```
tillh-stripe, line 205: We should reword these a little. And with the other suggestions, we could probably drop two cases that would be covered by returning `.notSupported`.

```suggestion
        var debugDescription: String {
            switch self {
            case .unexpectedUsBankAccount:
                return "The list of payment method types includes 'us_bank_account', which prevents the 'Bank' tab from being displayed."
            case .invalidEmailCollectionConfiguration:
                return "The provided configuration must either collect an email, or a default non-empty email must be provided."
            case .linkFundingSourcesMissingBankAccount:
                return "Your account isn't set up to process Instant Bank Payments. Reach out to Stripe support."
            }
        }
```

mats-stripe: Much better than what I had - will update!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+PaymentMethodAvailability.swift
```diff
@@ -186,8 +214,11 @@ extension PaymentSheet {
         case unactivated
         /// This payment method has requirements not met by the configuration or intent
         case missingRequirements(Set<PaymentMethodTypeRequirement>)
+        /// The primary requirement for this payment method were met, but other requirements are missing, preventing us from displaying this payment method.
+        case primaryRequirementMetButMissingOtherRequirements(String, Set<PaymentMethodTypeRequirement>)
```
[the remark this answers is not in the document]
mats-stripe: I see! So the logic would be:
- if primary requirement met, but not all other requirements are -> `.missingRequirements`
- if primary requirement not met -> `.notSupported`

And with this, we would log the missing requirements when the availability is `.missingRequirements` and can drop this newly added enum case 👍

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
mats-stripe: The primary requirement is not met in this scenario, so we don't log anything here

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
```diff
@@ -165,51 +171,63 @@ extension PaymentSheet {
                 // External Payment Methods
                 + elementsSession.externalPaymentMethods.map { PaymentMethodType.external($0) }
 
-            // We should manually add Instant Debits as a payment method when:
-            // - Link is an available payment method.
+            // We support Instant Bank Payments as a payment method when:
+            // - (Primary condition) Link is an available payment method.
+            // - The Financial Connections SDK is available.
             // - US Bank Account is *not* an available payment method.
-            // - Not a deferred intent flow.
```
tillh-stripe, line 171: Forgot about that — thank you 🙏

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
```diff
@@ -320,6 +338,124 @@ extension PaymentSheet {
...
+            )
+
+            if case .missingRequirements(let requirements) = configurationAvailabilityStatus {
+                missingRequirements.formUnion(requirements)
+            }
+
+            // Primary requirement:
+            // - Link is an available payment method.
+            if elementsSession.orderedPaymentMethodTypes.contains(.link) {
+                primaryRequirementMet = true
```
tillh-stripe, line 370: Can we use a `guard` for an early return here?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
```diff
@@ -320,6 +338,124 @@ extension PaymentSheet {
...
+            )
+
+            if case .missingRequirements(let requirements) = configurationAvailabilityStatus {
+                missingRequirements.formUnion(requirements)
+            }
+
+            // Primary:
+            // - Link Mode is set to Link Card Brand.
+            if elementsSession.isLinkCardBrand {
+                primaryRequirementMet = true
```
tillh-stripe, line 429: Same here.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
```diff
@@ -141,6 +141,14 @@ extension PaymentSheet {
         ///   - configuration: A `PaymentSheet` configuration.
         static func filteredPaymentMethodTypes(from intent: Intent, elementsSession: STPElementsSession, configuration: PaymentElementConfiguration, logAvailability: Bool = false) -> [PaymentMethodType]
         {
+            func logIfNecessary(_ message: String) {
```
tillh-stripe, line 144: Nice!

tillh-stripe (approved):

PR #4327, opened 2024-12-04
Applied fixes from update PM bug bash

joyceqin-stripe:
## Summary

Bug: “Pay” button is enabled when in edit mode, after deleting a card; fix: make button disabled whenever savedPaymentOptionsViewController.isRemovingPaymentMethods (editing saved pms)

Bug: Holding the Remove button changes the color from Red to Blue; fix: setTitleColor for: highlighted and then change border color on touch

Bug: Janky animation after deleting the last payment method; fix: wait until update pm screen popped before expanding add new pm screen

Bug: Label in manage card should not be selectable; fix: disabled user interaction for footnote

Bug: We are showing visa brand instead of MasterCard; fix: pass in card brand to the PAN configuration (it was previously looking at the text in the text field, sanitizing it, and figuring out what card brand it was from there. However, since our text in the last 4 configuration is just dots and the last 4, the PAN configuration was only taking in the last 4 and using that to guess the card brand, which was incorrect)
## Motivation

[Mobile SDK](https://jira.corp.stripe.com/browse/MOBILESDK)[MOBILESDK-2797](https://jira.corp.stripe.com/browse/MOBILESDK-2797)
https://docs.google.com/document/d/1OlfqsbflpEEjDsB2g5yxWHJr1ZVlinpmTHgLOFkg3TE/edit?usp=sharing


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
```diff
@@ -119,6 +130,7 @@ final class UpdatePaymentMethodViewController: UIViewController {
         }
         let label = ElementsUI.makeSmallFootnote(theme: viewModel.appearance.asElementsTheme)
         label.text = viewModel.footnote
+        label.isUserInteractionEnabled = false
```
porter-stripe, line 133: This should be in makeSmallFootnote(), the label should never be selectable, not just in this context.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
```diff
@@ -95,6 +95,9 @@ final class UpdatePaymentMethodViewController: UIViewController {
             button.contentEdgeInsets = UIEdgeInsets(top: 10, left: 16, bottom: 10, right: 16)
         }
         button.setTitleColor(viewModel.appearance.colors.danger, for: .normal)
+        button.setTitleColor(viewModel.appearance.colors.danger.disabledColor, for: .highlighted)
+        button.addTarget(self, action: #selector(buttonTouchDown(_:)), for: .touchDown)
+        button.addTarget(self, action: #selector(buttonTouchUp(_:)), for: [.touchUpInside, .touchUpOutside])
```
porter-stripe, line 100: Why do we need `touchUpOutside`?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/TextField/TextFieldElement+Card.swift
```diff
@@ -54,6 +56,13 @@ extension TextFieldElement {
                 }
             }
 
+            // If this is coming from the LastFourConfiguration, cardBrand(for: text) will retrieve a card brand from •••• •••• •••• last4, which may be incorrect, so we pass in the card brand for that case
+            if let cardBrand = cardBrand,
+               cardBrandDropDown == nil {
+                rotatingCardBrandsView.cardBrands = [cardBrand]
+                return rotatingCardBrandsView
+            }
```
porter-stripe, line 64: Any reason we can't have this below on line 77.
`rotatingCardBrandsView.cardBrands = [overrideCardBrand ?? cardBrand]`

joyceqin-stripe: cardBrand(for: text) takes the text in the text field (•••• •••• •••• last4) and sanitizes it, making it just the last4, and it tries to figure out what the card brand is from there. It could potentially return an unknown, so we can't guarantee it'll go into the else.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
joyceqin-stripe: If you hold down and press the button but then drag your finger and lift it outside of the bounds of the button, without touchUpOutside, it won't go back to its original color

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
joyceqin-stripe: That's true, I didn't consider that— I'll fix that

porter-stripe (approved):

PR #4331, opened 2024-12-05
Fix embedded example

porter-stripe:

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/ExampleEmbeddedElementCheckoutViewController.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/ExampleEmbeddedElementCheckoutViewController.swift
@@ -91,6 +91,7 @@ class ExampleEmbeddedElementCheckoutViewController: UIViewController {
     @objc
     func didTapPaymentMethodButton() {
         let paymentMethodsViewController = PaymentMethodsViewController(embeddedPaymentElement: embeddedPaymentElement, needsDismissal: { [weak self] in
+            self?.embeddedPaymentElement.presentingViewController = self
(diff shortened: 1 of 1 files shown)


Example/PaymentSheet Example/PaymentSheet Example/ExampleEmbeddedElementCheckoutViewController.swift
```diff
@@ -91,6 +91,7 @@ class ExampleEmbeddedElementCheckoutViewController: UIViewController {
     @objc
     func didTapPaymentMethodButton() {
         let paymentMethodsViewController = PaymentMethodsViewController(embeddedPaymentElement: embeddedPaymentElement, needsDismissal: { [weak self] in
+            self?.embeddedPaymentElement.presentingViewController = self
```
wooj-stripe, line 94: I'm not sure why this is needed -- it seems to be set in the init of PyamentMethodsViewController

porter-stripe: That is because the presenting view controller is set when the PaymentMethodsViewController is presented, so when it dismisses we need to set the presentingViewController back to self.

https://github.com/stripe/stripe-ios/blob/master/Example/PaymentSheet%20Example/PaymentSheet%20Example/ExampleEmbeddedElementCheckoutViewController.swift#L365

wooj-stripe: chatted offline -- this is needed due to authContext on confirm.

wooj-stripe (approved):

PR #4333, opened 2024-12-05
Display a “default” badge for the default PM

joyceqin-stripe:
## Summary

Display a "default" badge for the default PM in vertical and horizontal mode across all UI shapes (payment sheet, flow controller, embedded)
## Motivation

[MOBILESDK-2800](https://jira.corp.stripe.com/browse/MOBILESDK-2800)
## Testing

https://github.com/user-attachments/assets/e2dd8f48-e23c-4e4a-8732-bb55a98efa80

https://github.com/user-attachments/assets/8a0c25a5-7384-483d-9814-d8f8c70d898f

test_all_saved_pms_and_apple_pay_and_link_default_badge@3x

test_VerticalSavedPaymentOptionsViewControllerSnapshotTestsDefaultBadge@3x


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSavedPaymentMethodsCollectionViewController.swift
```diff
@@ -386,7 +386,8 @@ extension CustomerSavedPaymentMethodsCollectionViewController: UICollectionViewD
 
         cell.setViewModel(viewModel.toSavedPaymentOptionsViewControllerSelection(),
                           cbcEligible: cbcEligible,
-                          allowsPaymentMethodRemoval: configuration.paymentMethodRemove)
+                          allowsPaymentMethodRemoval: configuration.paymentMethodRemove,
+                          allowsSetAsDefaultPM: false)
```
porter-stripe, line 390: Why always pass false here? Maybe add a TODO here for yourself so we don't forget to update this if that's the plan.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentMethodCollectionView.swift
```diff
@@ -43,13 +45,22 @@ class SavedPaymentMethodCollectionView: UICollectionView {
     }
 
     var isRemovingPaymentMethods: Bool = false
+    var showDefaultPMBadge: Bool = false
```
porter-stripe, line 48: This should be a `let` like `appearance`, and we shouldn't need to define `false` as it's default value twice.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentMethodCollectionView.swift
```diff
@@ -91,11 +103,29 @@ extension SavedPaymentMethodCollectionView {
             button.accessibilityLabel = String.Localized.edit
             return button
         }()
+        lazy var defaultBadge: UILabel = {
+            let label = UILabel()
+            label.font = appearance.scaledFont(for: appearance.font.base.medium, style: .footnote, maximumPointSize: 20)
+            label.textColor = appearance.colors.textSecondary
+            label.adjustsFontForContentSizeCategory = true
+            label.text = STPLocalizedString("Default", "Label for identifying the default payment method.")
```
porter-stripe, line 111: Let's move this into a variable into `StripePaymentSheet/StripePaymentSheet/Source/Categories/String+Localized.swift` and update the usage in LinkPaymentMethodPicker-Cell. This way we can easily reuse the translations w/o worrying about matching the comment on it.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentMethodCollectionView.swift
```diff
@@ -91,11 +103,29 @@ extension SavedPaymentMethodCollectionView {
...
 
         fileprivate var viewModel: SavedPaymentOptionsViewController.Selection?
+        var isDefaultPM: Bool = false
 
         var isRemovingPaymentMethods: Bool = false {
             didSet {
+                if allowsSetAsDefaultPM {
+                    if isRemovingPaymentMethods {
+                        addDefaultBadge()
+                        defaultBadge.isHidden = !isDefaultPM
```
porter-stripe, line 123: I find it a little odd that we add the default badge to PMs that aren't the default then remove them. We should move this logic into the `addDefaultBadge()` function and maybe rename it to `addDefaultBadgeIfNecessary`.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentMethodCollectionView.swift
```diff
@@ -91,11 +103,29 @@ extension SavedPaymentMethodCollectionView {
...
+            label.text = STPLocalizedString("Default", "Label for identifying the default payment method.")
+            return label
+        }()
 
         fileprivate var viewModel: SavedPaymentOptionsViewController.Selection?
+        var isDefaultPM: Bool = false
 
         var isRemovingPaymentMethods: Bool = false {
             didSet {
+                if allowsSetAsDefaultPM {
```
porter-stripe, line 120: Any harm in just `if allowsSetAsDefaultPM, isRemovingPaymentMethods` rather than the nesting?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentMethodCollectionView.swift
```diff
@@ -16,6 +16,7 @@ import UIKit
 // MARK: - Constants
 /// Entire cell size
 private let cellSize: CGSize = CGSize(width: 106, height: 94)
+private let cellSizeWithDefaultBadge: CGSize = CGSize(width: 106, height: 112)
```
porter-stripe, line 19: What's the harm in making all the cells the new size to account for the possible default badge? What happens if we use the old size? Would be preferred if we can avoid swapping sizes.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentMethodCollectionView.swift
```diff
@@ -372,6 +407,42 @@ extension SavedPaymentMethodCollectionView {
...
+                NSLayoutConstraint.deactivate([
+                    labelHeightConstraint
+                ])
+                defaultBadge.removeFromSuperview()
+                NSLayoutConstraint.activate([
+                    labelBottomConstraint
+                ])
+                contentView.layoutIfNeeded()
+            }
+        }
```
porter-stripe, line 445: I think a simpler approach would be to add the default badge to all PMs, then just hide/show it when needed.
This way you can use one size for all the cells and not need to swap the cell sizes, and can move adding the default badge up to where the rest of the cell adds it's view's, and you won't need to track the extra constraints. You may also not need to reload the collection view either.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/Vertical Saved Payment Method Screen/SavedPaymentMethodRowButton.swift
```diff
@@ -47,6 +47,8 @@ final class SavedPaymentMethodRowButton: UIView {
         }
     }
 
+    var showDefaultPMBadge: Bool = false
```
porter-stripe, line 50: Let's make this a let.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButton.swift
```diff
@@ -95,6 +97,19 @@ class RowButton: UIView {
         } else {
             self.sublabel = nil
         }
+
+        if showDefaultPMBadge {
+            let defaultBadge = UILabel()
+            defaultBadge.font = appearance.scaledFont(for: appearance.font.base.medium, style: .footnote, maximumPointSize: 20)
```
porter-stripe, line 103: Figma shows in vertical mode this being a weight of regular and caption1 for style.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButton.swift
```diff
@@ -95,6 +97,19 @@ class RowButton: UIView {
         } else {
             self.sublabel = nil
         }
+
+        if showDefaultPMBadge {
+            let defaultBadge = UILabel()
```
porter-stripe, line 102: Nit: I know you're following the `sublabel` pattern, but might be cleaner if you followed the `checkmarkImageView` pattern, up to you.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButton.swift
```diff
@@ -115,6 +130,11 @@ class RowButton: UIView {
         labelsStackView.axis = .vertical
         labelsStackView.alignment = .leading
 
+        let horizontalStackView = UIStackView(arrangedSubviews: [labelsStackView, defaultBadge].compactMap { $0 })
```
porter-stripe, line 133: Let's name this something more descriptive.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButton.swift
```diff
@@ -175,13 +195,31 @@ class RowButton: UIView {
...
             view.isAccessibilityElement = false
             addSubview(view)
         }
 
+        if let defaultBadge = defaultBadge {
+            for view in [labelsStackView, defaultBadge].compactMap({ $0 }) {
+                view.translatesAutoresizingMaskIntoConstraints = false
+                view.isUserInteractionEnabled = false
+                view.isAccessibilityElement = false
+                addSubview(view)
```
porter-stripe, line 210: Why are we adding `labelsStackView` and `defaultBadge` to the view again when it's already added to the `horizontalStackView` and `horizontalStackView` is added to the view? Also why not just iterate the arrangedSubviews in the `horizontalStackView`?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButton.swift
```diff
@@ -175,13 +195,31 @@ class RowButton: UIView {
...
+            }
+            let stackViewConstraints = [
```
porter-stripe, line 212: Could use a better name too, there's lots of stack views in here, it's ambiguous.

porter-stripe (commented):
Seeing a bug where we show "Default" under "+ Add", let's add a snapshot test for this case.
Simulator Screenshot - iOS 18  - 2024-12-17 at 11 15 27

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSavedPaymentMethodsCollectionViewController.swift
joyceqin-stripe: CustomerSheet should not display a default badge

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentMethodCollectionView.swift
joyceqin-stripe: If you look in the Figma, the sheet gets taller when entering edit mode due to the default badge. The way it currently works, it updates the item size and intrinsic content size and adds the default badge to all PMs upon entering edit mode and hides/shows when needed. When it exits edit mode, the default badge is removed from all the cells and the item size and intrinsic content size return to normal. If the default badge is there the whole time and the item size and intrinsic content sizes are constant the whole time, then there is extra space under the PMs, and the height of the sheet doesn't change.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentMethodCollectionView.swift
joyceqin-stripe: If something doesn't exist in the view, then I suppose it can't be removed from it, so I guess that would be fine

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentMethodCollectionView.swift
joyceqin-stripe: I've figured out that I can have the default badge added to the view but change the constraints so that it won't take up the extra space, but I don't see a way around the dynamically changing cell sizes yet

joyceqin-stripe:
I've changed the way I'm laying out the default badge in Vertical Mode by having the constraint just be 8px right of the end of the label and centerY

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/ElementsCustomer.swift
```diff
@@ -40,9 +40,14 @@ struct ElementsCustomer: Equatable, Hashable {
         return ElementsCustomer(paymentMethods: paymentMethods, defaultPaymentMethod: defaultPaymentMethod, customerSession: customerSession)
     }
 
+    func getDefaultPaymentMethod() -> STPPaymentMethod? {
+        let defaultSavedPaymentMethod = paymentMethods.first { $0.stripeId == defaultPaymentMethod }
+        return defaultSavedPaymentMethod
+    }
+
     func getDefaultOrFirstPaymentMethod() -> STPPaymentMethod? {
```
wooj-stripe, line 48: in PaymentSheetVerticalViewController, we have this logic for the initialSelection
```
                    if let customer = elementsSession.customer,
                       let defaultPaymentMethod = customer.getDefaultOrFirstPaymentMethod() {
                        customerDefault = CustomerPaymentOption.stripeId(defaultPaymentMethod.stripeId)
                    }
```
This seems to indicate that the first payment method can become the default. Is this true?  Ultimately, i'm trying to wrap my head around when we would access defaultPaymentMethod vs defaultOrFirst.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentMethodCollectionView.swift
```diff
@@ -208,15 +245,20 @@ extension SavedPaymentMethodCollectionView {
             }
         }
 
-        // MARK: - Internal Methods
+        private var labelBottomConstraint: NSLayoutConstraint?
+        private var labelHeightConstraint: NSLayoutConstraint?
```
wooj-stripe, line 249: Seems like these do not need to be optional.  We can initialize them in the init(), and then activate/deactivate when needed.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSavedPaymentMethodsCollectionViewController.swift
```diff
@@ -386,7 +386,9 @@ extension CustomerSavedPaymentMethodsCollectionViewController: UICollectionViewD
 
         cell.setViewModel(viewModel.toSavedPaymentOptionsViewControllerSelection(),
                           cbcEligible: cbcEligible,
-                          allowsPaymentMethodRemoval: configuration.paymentMethodRemove)
+                          allowsPaymentMethodRemoval: configuration.paymentMethodRemove,
+                          allowsSetAsDefaultPM: false,
+                          showDefaultPMBadge: false)
```
wooj-stripe, line 391: I'm guessing customerSheet changes will come at a later time?

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/ElementsCustomer.swift
joyceqin-stripe: From my understanding, the selected payment method should be the default if it exists or the first payment method in the customer's list of saved payment methods. However, the default badge itself should only be displayed if there is a set default in the elementsSession customer.

So we access defaultOrFirst when trying to determine the selected payment method, while we access defaultPaymentMethod to determine if the default badge should be shown and which payment method to show it on.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSavedPaymentMethodsCollectionViewController.swift
joyceqin-stripe: Bella said that the CustomerSheet should not ever display a default badge because it's implied that the selected payment method in CustomerSheet is your default.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentMethodCollectionView.swift
```diff
@@ -91,11 +103,31 @@ extension SavedPaymentMethodCollectionView {
...
 
         fileprivate var viewModel: SavedPaymentOptionsViewController.Selection?
+        var isDefaultPM: Bool = false
 
         var isRemovingPaymentMethods: Bool = false {
             didSet {
+                if showDefaultPMBadge {
+                    if isRemovingPaymentMethods {
+                        activateDefaultBadge()
+                        defaultBadge.isHidden = !isDefaultPM
```
wooj-stripe, line 124: We should avoid setting .isHidden directly as it can lead to unexpected behavior if setting it to true multiple times in a row. Instead, it is safer to call our helper function `setHiddenIfNecessary`  on it instead.

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/ElementsCustomer.swift
wooj-stripe: Ah thanks for the explanation!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentMethodCollectionView.swift
```diff
@@ -24,12 +25,13 @@ private let paymentMethodLogoSize: CGSize = CGSize(width: 54, height: 40)
 /// For internal SDK use only
 @objc(STP_Internal_SavedPaymentMethodCollectionView)
 class SavedPaymentMethodCollectionView: UICollectionView {
-    init(appearance: PaymentSheet.Appearance) {
+    init(appearance: PaymentSheet.Appearance, showDefaultPMBadge: Bool = false) {
```
wooj-stripe, line 28: Variable name is awkward since we technically aren't showing a default PM badge in this context -- I think we should rename this to something like, `needsVerticalPaddingForBadge`.  If we end up needing a 3rd size, then we should change it from a Bool to a Float and call it `verticalPaddingForBadge` and just calculate it.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentMethodCollectionView.swift
```diff
@@ -43,13 +45,23 @@ class SavedPaymentMethodCollectionView: UICollectionView {
...
     }
 
     override var intrinsicContentSize: CGSize {
-        return CGSize(width: UIView.noIntrinsicMetric, height: 100)
+        return showDefaultPMBadge && isRemovingPaymentMethods ? CGSize(width: UIView.noIntrinsicMetric, height: 118) : CGSize(width: UIView.noIntrinsicMetric, height: 100)
+    }
+
+    func updateLayout() {
+        guard let layout = collectionViewLayout as? UICollectionViewFlowLayout else { return }
+        let newItemSize = showDefaultPMBadge && isRemovingPaymentMethods ? cellSizeWithDefaultBadge : cellSize
```
wooj-stripe, line 60: nit: s/newItemSize/newCellSize/

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentMethodCollectionView.swift
```diff
@@ -208,15 +245,20 @@ extension SavedPaymentMethodCollectionView {
             }
         }
 
-        // MARK: - Internal Methods
+        private var labelBottomConstraint: NSLayoutConstraint = NSLayoutConstraint()
+        private var labelHeightConstraint: NSLayoutConstraint = NSLayoutConstraint()
```
wooj-stripe, line 249: We can turn these into lazy vars?
```
        lazy var labelBottomConstraint: NSLayoutConstraint = {
            return label.bottomAnchor.constraint(equalTo: contentView.bottomAnchor)
        }()
        lazy var labelHeightConstraint: NSLayoutConstraint = {
            return label.heightAnchor.constraint(equalToConstant: 20)
        }()
        lazy var defaultBadgeConstraints: [NSLayoutConstraint] = {
            return [
                defaultBadge.topAnchor.constraint(
                    equalTo: label.bottomAnchor, constant: 4),
                defaultBadge.bottomAnchor.constraint(equalTo: contentView.bottomAnchor),
                defaultBadge.leadingAnchor.constraint(equalTo: contentView.leadingAnchor, constant: 2),
                defaultBadge.trailingAnchor.constraint(equalTo: contentView.trailingAnchor)
            ]
        }()
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentMethodCollectionView.swift
```diff
@@ -372,6 +414,34 @@ extension SavedPaymentMethodCollectionView {
...
+
+        private func deactivateDefaultBadge() {
+            NSLayoutConstraint.deactivate(defaultBadgeConstraints)
+            NSLayoutConstraint.deactivate([
+                labelHeightConstraint
+            ])
+            NSLayoutConstraint.activate([
+                labelBottomConstraint
+            ])
+        }
```
wooj-stripe, line 443: Naming on these functions feels awkward since we aren't activating a badge.  Maybe a better name might be "activateConstraintsDefaultBadge" and "deactivateConstraintsforDefaultBadge"?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentOptionsViewController.swift
```diff
@@ -508,8 +518,15 @@ extension SavedPaymentOptionsViewController: UICollectionViewDataSource, UIColle
...
         }
-        cell.setViewModel(viewModel, cbcEligible: cbcEligible, allowsPaymentMethodRemoval: self.configuration.allowsRemovalOfPaymentMethods)
+        cell.setViewModel(viewModel, cbcEligible: cbcEligible, allowsPaymentMethodRemoval: self.configuration.allowsRemovalOfPaymentMethods, allowsSetAsDefaultPM: configuration.allowsSetAsDefaultPM, showDefaultPMBadge: hasDefault)
         cell.delegate = self
+        let isDefaultPM: Bool = {
+            guard self.hasDefault,
+                  let savedPMId = viewModel.savedPaymentMethod?.stripeId,
+                  let defaultPMId = self.elementsSession.customer?.defaultPaymentMethod else { return false }
+            return savedPMId == defaultPMId
+        }()
```
wooj-stripe, line 528: This logic feel awkward because self.hasDefault is calling `self.elementsSession.customer?.defaultPaymentMethod`, which is also being called here. Seems like there is a way to simplify this logic -- maybe we can introduce a function which takes the stripeId and returns a boolean to determine if it is the default payment method?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/Vertical Saved Payment Method Screen/VerticalSavedPaymentMethodsViewController.swift
```diff
@@ -177,7 +177,8 @@ class VerticalSavedPaymentMethodsViewController: UIViewController {
     private func buildPaymentMethodRows(paymentMethods: [STPPaymentMethod]) -> [SavedPaymentMethodRowButton] {
         return paymentMethods.map { paymentMethod in
             let button = SavedPaymentMethodRowButton(paymentMethod: paymentMethod,
-                                                     appearance: configuration.appearance)
+                                                     appearance: configuration.appearance,
+                                                     showDefaultPMBadge: configuration.allowsSetAsDefaultPM && paymentMethod.stripeId == elementsSession.customer?.defaultPaymentMethod)
```
wooj-stripe, line 181: This feels repetitive:
```
configuration.allowsSetAsDefaultPM && paymentMethod.stripeId == elementsSession.customer?.defaultPaymentMethod)
```

How about we create a new function:
```
    func isDefaultPaymentMethod(paymentMethodId: String) -> Bool {
        return configuration.allowsSetAsDefaultPM && paymentMethodId == elementsSession.customer?.defaultPaymentMethod
    }

```

Then, calling it looks like:
```
 let button = SavedPaymentMethodRowButton(paymentMethod: paymentMethod,
                                                     showDefaultPMBadge: isDefaultPaymentMethod(paymentMethodId: paymentMethod.stripeId))

....
let newButton = SavedPaymentMethodRowButton(paymentMethod: updatedPaymentMethod,
                                                    showDefaultPMBadge: isDefaultPaymentMethod(paymentMethodId: updatedPaymentMethod.stripeId))
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentMethodCollectionView.swift
```diff
@@ -208,15 +243,32 @@ extension SavedPaymentMethodCollectionView {
...
+                    equalTo: label.bottomAnchor, constant: 4),
+                defaultBadge.bottomAnchor.constraint(equalTo: contentView.bottomAnchor),
+                defaultBadge.leadingAnchor.constraint(equalTo: contentView.leadingAnchor, constant: 2),
+                defaultBadge.trailingAnchor.constraint(equalTo: contentView.trailingAnchor)
+            ]
+        }()
 
-        func setViewModel(_ viewModel: SavedPaymentOptionsViewController.Selection, cbcEligible: Bool, allowsPaymentMethodRemoval: Bool) {
+        // MARK: - Internal Methods
+        func setViewModel(_ viewModel: SavedPaymentOptionsViewController.Selection, cbcEligible: Bool, allowsPaymentMethodRemoval: Bool, allowsSetAsDefaultPM: Bool = false, showDefaultPMBadge: Bool = false) {
```
wooj-stripe: Is there a way to make `allowsSetAsDefaultPM`, `showDefaultPMBadge` and `isDefaultPM` less confusing?  I understand isDefaultPM actually shows the badge, but I'm getting confused between `allowsSetAsDefaultPM`, and `showDefaultPMBadge`.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButton.swift
```diff
@@ -69,6 +70,7 @@ class RowButton: UIView {
         imageView: UIImageView,
         text: String,
         subtext: String? = nil,
+        showDefaultPMBadge: Bool = false,
```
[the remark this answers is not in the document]
joyceqin-stripe: I just recently noticed that the Figma shows the badge with a different font weight when selected vs unselected, so I changed it to pass in defaultBadge as a UILabel? and adjust the weight when the state changes instead— does that work?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentMethodCollectionView.swift
joyceqin-stripe: I've changed it to allowsSetAsDefaultPM, needsVerticalPaddingForBadge, and showDefaultPMBadge. 

allowsSetAsDefaultPM is for the purpose of displaying the circular edit icon when "edit" is pressed— if a user is allowed to set a default payment method, if saved payment methods can be removed, or if a card is cbc eligible, then the user should have the option of managing their payment method and navigating to the update payment method screen.

needsVerticalPaddingForBadge indicates that there exists a set default payment method, and therefore, the cells need padding on the bottom to accommodate for the default badge. The default badge may not be shown on this cell, but it needs the padding regardless to match the cell that has the badge.

showDefaultPMBadge, now, is only true for the cell with the default payment method. When this is true, the default badge is displayed.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentMethodCollectionView.swift
```diff
@@ -91,11 +103,30 @@ extension SavedPaymentMethodCollectionView {
...
+            label.text = String.Localized.default_text
+            label.isHidden = true
+            return label
+        }()
 
         fileprivate var viewModel: SavedPaymentOptionsViewController.Selection?
 
         var isRemovingPaymentMethods: Bool = false {
             didSet {
+                if needsVerticalPaddingForBadge {
```
wooj-stripe, line 120: Wdyt: I think the code could be written more linearly:
```
        var isRemovingPaymentMethods: Bool = false {
            didSet {
                updateVerticalConstraintsIfNeeded()
                update()
            }
        }

        func updateVerticalConstraintsIfNeeded() {
            guard needsVerticalPaddingForBadge else {
                return
            }
            if isRemovingPaymentMethods {
                activateDefaultBadgeConstraints()
                defaultBadge.setHiddenIfNecessary(!showDefaultPMBadge)
            } else {
                deactivateDefaultBadgeConstraints()
                defaultBadge.setHiddenIfNecessary(true)
            }
        }
```
This way, when reading `didSet` I don't have to read the implementation of how verticalConstraints get updated-- i can read that constraints get updated if needed, and then a call to update happens irrespective of updating constraints.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButton.swift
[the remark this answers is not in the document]
wooj-stripe: I think it is a bit odd to pass in UILabel from a responsibilities perspective (e.g. If i have a view, feels like i should tell it what text to render, and it's up to that class to determine how to render it (might be a UILabel, might be a UITextView, etc..)

It seems that the selected state is available in the RowButton -- can't we hook into that to update the font weight?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentMethodCollectionView.swift
joyceqin-stripe: Yeah, I think that makes sense

Wmc05061970:
What is this

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButton.swift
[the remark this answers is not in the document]
joyceqin-stripe: Yeah, that's true— just updated to reflect that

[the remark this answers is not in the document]
wooj-stripe: In SavedPaymentMethodRowButton.swift, can we add:
```
    private lazy var rowButton: RowButton = {
        return .makeForSavedPaymentMethod(paymentMethod: paymentMethod, appearance: appearance, badgeText: badgeText, rightAccessoryView: chevronButton, didTap: handleRowButtonTapped)
    }()

    private lazy var badgeText: String? = {
        return showDefaultPMBadge ? String.Localized.default_text : nil
    }()
```

wooj-stripe (approved):

porter-stripe (approved):

PR #4338, opened 2024-12-06
Add initial StripeAttest code

davidme-stripe:
## Summary
Adds StripeAttest, a class for implementing basic App Attestation against the Stripe API.

## Motivation
Enabling basic App Attestation support in the SDK.

## Testing
Added unit tests and additional tests for a few edge cases.

## Changelog
Not user-visible.

### DIFF
--- a/StripeCore/StripeCore/Source/Attestation/StripeAttest.swift
+++ b/StripeCore/StripeCore/Source/Attestation/StripeAttest.swift
@@ -70,11 +70,11 @@ import UIKit
     // MARK: - Internal
 
     // MARK: - Device-local settings
-    private enum DefaultsKeys: String {
+    enum DefaultsKeys: String {
         /// The ID of the attestation key stored in the keychain.
         case keyID = "STPAttestKeyID"
         /// The last date we generated an attestation key, used for rate limiting.
-        case lastGeneratedDate = "STPAttestKeyLastGenerated"
+        case lastAttestedDate = "STPAttestKeyLastAttestedDate"
         /// Whether the current keyID key has been attested successfully.
         case successfullyAttested = "STPAttestKeySuccessfullyAttested"
     }
@@ -97,12 +97,12 @@ import UIKit
         }
     }
 
-    private static var lastGeneratedDate: Date? {
+    private static var lastAttestedDate: Date? {
         get {
-            UserDefaults.standard.object(forKey: DefaultsKeys.lastGeneratedDate.rawValue) as? Date
+            UserDefaults.standard.object(forKey: DefaultsKeys.lastAttestedDate.rawValue) as? Date
         }
         set {
-            UserDefaults.standard.set(newValue, forKey: DefaultsKeys.lastGeneratedDate.rawValue)
+            UserDefaults.standard.set(newValue, forKey: DefaultsKeys.lastAttestedDate.rawValue)
         }
     }
 
@@ -159,6 +159,13 @@ import UIKit
     }
 
     func _attest() async throws {
+        // It's dangerous to attest, as it increments a permanent counter for the device.
+        // First, make sure we the last time we called this is more than 24 hours away from now.
+        // (Either in the future or the past, who knows what people are doing with their clocks)
+        if let lastGenerated = StripeAttest.lastAttestedDate, abs(lastGenerated.timeIntervalSinceNow) < Self.minDurationBetweenKeyGenerationAttempts {
+            throw AttestationError.keygenRateLimitExceeded
+        }
+
         let keyId = try await self.getOrCreateKeyID()
         let challenge = try await getChallenge()
         guard let challengeData = challenge.data(using: .utf8) else {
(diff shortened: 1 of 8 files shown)


StripeCore/StripeCore/Source/Attestation/StripeAttest.swift
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    private static var keyID: String? {
+        get {
+            UserDefaults.standard.string(forKey: DefaultsKeys.keyID.rawValue)
+        }
+        set {
+            UserDefaults.standard.set(newValue, forKey: DefaultsKeys.keyID.rawValue)
+        }
+    }
```
mats-stripe, line 89: Out of scope, but makes me wonder if we should pull out our `@UserDefaults` [property wrapper](https://github.com/stripe/stripe-ios/blob/master/Example/FinancialConnections%20Example/FinancialConnections%20Example/Playground/helpers/UserDefaults.swift) to StripeCore 👀 😅 

```swift
@UserDefault(DefaultsKeys.keyID.rawValue, defaultValue: nil)
static var keyID: String?
```

mats-stripe (commented):
Really nice job documenting all this complex code - it made it very easy to follow along from having nearly no context on how attestation works. The API seems like it will be super easy to integrate in FC! 👏 Happy to ✅ this once it's ready to go!

StripeCore/StripeCore/Source/Attestation/StripeAttest.swift
```diff
@@ -0,0 +1,278 @@
...
+            // If error is DCErrorInvalidKey (3) and the domain is DCErrorDomain,
+            // then the key is either unattested or corrupted.
+            let error = error as NSError
+            if error.domain == DCErrorDomain && error.code == DCError.invalidKey.rawValue {
+                if retryIfNeeded {
+                    // We'll try to attest again, maybe our initial attestation was unsuccessful?
+                    // `DCError.invalidKey` could mean a lot of things, unfortunately.
+                    // If this doesn't work, then in `attest()` we'll deem the key to be corrupted
+                    // and throw it out.
+                    try await attest()
```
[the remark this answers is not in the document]
davidme-stripe: It isn't a straightforward retry, it's re-attesting the key (in case it wasn't correctly attested) and *then* retrying with the re-attested key or a new key. So a timeout wouldn't change the behavior. I can make that clearer in the comments/naming!

StripeCore/StripeCoreTests/Attestation/StripeAttestTest.swift
```diff
@@ -0,0 +1,67 @@
...
+        let invalidKeyError = NSError(domain: DCErrorDomain, code: DCError.invalidKey.rawValue, userInfo: nil)
+        // But fail the assertion, causing the key to be reset
+        mockAttestService.shouldFailAssertionWithError = invalidKeyError
+        do {
+            _ = try await stripeAttest.assert()
+            XCTFail("Should not succeed")
+        } catch {
+            // Should get a rate limiting error when we try to generate the second key:
+            XCTAssertEqual(error as! StripeAttest.AttestationError, StripeAttest.AttestationError.keygenRateLimitExceeded)
+        }
```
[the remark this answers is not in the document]
davidme-stripe: I tried that at first, but it unfortunately doesn't appear to support async functions!

StripeCore/StripeCore/Source/Attestation/StripeAttest.swift
davidme-stripe: Oooh, neat! We should!

StripeCore/StripeCoreTests/Attestation/StripeAttestTest.swift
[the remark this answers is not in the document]
mats-stripe: Ahhh darn!

StripeCore/StripeCore/Source/Attestation/StripeAttest.swift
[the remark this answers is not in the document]
mats-stripe: Ahh gotcha, that makes sense!

modules.yaml
```diff
@@ -155,8 +155,8 @@ modules:
     output: docs/stripeapplepay
     readme: StripeApplePay/README.md
   size_report:
-    max_compressed_size: 500
-    max_uncompressed_size: 1200
+    max_compressed_size: 700
+    max_uncompressed_size: 1700
```
mats-stripe, line 159: Does this PR increase the `StripeApplePay` pod size by ~40%? 😮

mats-stripe (approved):

davidme-stripe:
Unfortunately I had to bump up the size of StripeApplePay and StripeFinancialConnections due to the inclusion of `async` support, but I don't think it will cause problems for most users: If the minimum iOS version of their app clip is high enough, Xcode won't bundle  libswift_Concurrency.dylib.

modules.yaml
davidme-stripe: Yeah, see below comment, this is because of `libswift_Concurrency.dylib` back-deployment. :( Fortunately it should only increase the size for apps targeting older iOS versions, though I'm not sure of the exact versions.

davidme-stripe: I also considered putting this somewhere else, but we need to use it in FinancialConnections, Payments, and PaymentSheet, so there isn't a better location for it. We're planning to add more async support in 2025, so I think it's okay to move to it now.

PR #4346, opened 2024-12-09
Add dashboard only parameters to connect sdk.

nschris-stripe:
## Summary
Introduces a few fields that will be useful for testing the connect sdk. This required:
- Making allowedHosts dynamic since the baseURL can now be changed
- Lightly refactoring sanitized key logic so that we can sanitize any key string. This is useful because the publishable key isn't necessarily coming from the api client.
- Adding tests to validate functionality.


StripeCore/StripeCore/Source/Categories/String+StripeCore.swift
```diff
@@ -20,6 +20,11 @@ import Foundation
     public var nonEmpty: String? {
         stringIfHasContentsElseNil(self)
     }
+    
+    @_spi(STP) public var sanitizedKey: String {
+        return (self.isSecretKey || self.hasPrefix("uk_"))
+            ? "[REDACTED_LIVE_KEY]" : self
+    }
```
mludowise-stripe, line 27: ➕

StripeConnect/StripeConnect/Source/EmbeddedComponentManager.swift
```diff
@@ -33,6 +33,12 @@ public class EmbeddedComponentManager {
         ComponentAnalyticsClient(client: AnalyticsClientV2.sharedConnect,
                                  commonFields: $0)
     }
+    
+    @_spi(DashboardOnly)
+    public var baseURL: URL = StripeConnectConstants.connectJSBaseURL
+
+    @_spi(DashboardOnly)
+    public var publicKeyOverride: String? = nil
```
mludowise-stripe, line 41: optional – given that this is required for Dashboard, we could add a `DashboardOnly` convenience init that takes this argument in place of `fetchClientSecret`.

StripeConnect/StripeConnect/Source/Internal/Analytics/ComponentAnalyticsClient.swift
```diff
@@ -240,23 +241,15 @@ class ComponentAnalyticsClient {
 extension ComponentAnalyticsClient.CommonFields {
     init(apiClient: STPAPIClient,
          component: ComponentType,
+         publicKeyOverride: String?,
          componentInstance: UUID = .init()
     ) {
```
mludowise-stripe, line 246: optional - what do you think about passing the `ConnectJSURLParams` directly in `init` instead of separate `apiClient`, `component`, and `publicKeyOverride` params? We already have to construct the URL params in the ComponentManager, so we could just do that earlier in the stack and forward it here.

StripeConnect/StripeConnect/Source/Helpers/ConnectJSURLParams.swift
```diff
@@ -36,27 +36,27 @@ struct ConnectJSURLParams: Encodable {
...
 extension ConnectJSURLParams {
-    init(component: ComponentType, apiClient: STPAPIClient) {
+    init(component: ComponentType, apiClient: STPAPIClient, publicKeyOverride: String?) {
         self.component = component
 
         // Validate that publishable key has been set
         STPAPIClient.validateKey(apiClient.publishableKey)
-
         if apiClient.publishableKeyIsUserKey {
-            // Dashboard app overrides
```
mludowise-stripe, line 46: nit: I think keeping this comment will help future devs understand the intention of this check

StripeConnect/StripeConnect/Source/Internal/Webview/ConnectWebViewController.swift
```diff
@@ -41,8 +42,11 @@ class ConnectWebViewController: UIViewController {
     /// The current version for the SDK
     let sdkVersion: String?
 
+    let allowedHosts: [String]
```
mludowise-stripe, line 45: nit: docstring
```suggestion
    /// Popups with an allowed host will open in a PopupViewController and all others will open in a SafariVC.
    /// Camera permission requests from allowed hosts use the app's camera permissions while all other requests will explicitly ask for user permission.
    let allowedHosts: [String]
```

mludowise-stripe (approved):
Left some minor suggestions, but no blockers.

nschris-stripe:
Diff: https://github.com/stripe/stripe-ios/compare/8f797e2b1dd8a834cf533a2803a28273a253c3ec..8f7f90584c0f660450a2361b3b59be1a3a2a3414

StripeConnect/StripeConnect/Source/Internal/StripeConnectConstants.swift
```diff
@@ -15,11 +15,10 @@ enum StripeConnectConstants {
      - Open popups in PopupWebViewController (instead of Safari)
      */
     static let allowedHosts: Set<String> = [
-        "connect-js.stripe.com",
```
nschris-stripe, line 18: No longer needed since this is the default base url. We have logic to add the base url to the allowed host list.

nschris-stripe: Also added tests to cover this.

StripeConnect/StripeConnect/Source/EmbeddedComponentManager.swift
```diff
@@ -33,7 +33,33 @@ public class EmbeddedComponentManager {
         ComponentAnalyticsClient(client: AnalyticsClientV2.sharedConnect,
                                  commonFields: $0)
     }
+    
+    @_spi(DashboardOnly)
+    var baseURL: URL = StripeConnectConstants.connectJSBaseURL
 
+    @_spi(DashboardOnly)
+    var publicKeyOverride: String? = nil
```
mludowise-stripe, line 41: does this still need to be `@_spi(DashboardOnly)` since it's now settable in `init`? I think it could be made private.

If we decide to keep it, then I think it will need to be marked public to make it accessible from Dashboard.

StripeConnect/StripeConnect/Source/EmbeddedComponentManager.swift
```diff
@@ -33,7 +33,33 @@ public class EmbeddedComponentManager {
         ComponentAnalyticsClient(client: AnalyticsClientV2.sharedConnect,
                                  commonFields: $0)
     }
+    
+    @_spi(DashboardOnly)
+    var baseURL: URL = StripeConnectConstants.connectJSBaseURL
```
mludowise-stripe, line 38: does `baseURL` need to be public so it can be set from Dashboard?

mludowise-stripe (approved):
I think `baseURL` and `publicKeyOverride` need to be public, but otherwise changes look good. Leaving a shipit in case I'm wrong so it won't block from merging.

StripeConnect/StripeConnect/Source/EmbeddedComponentManager.swift
nschris-stripe: I removed the annotation but I can't make it private because we need to access it for tests.

StripeConnect/StripeConnect/Source/EmbeddedComponentManager.swift
nschris-stripe: baseURL has been added to the initializer so I treated it the same as publicKeyOverride

mludowise-stripe (approved):

PR #4359, opened 2024-12-16
[Connect] Update SPI shape for returning from FinancialConnections

mludowise-stripe:
## Summary

The StripeConnect web layer requires the complete json object for the `BankToken` and `FinancialConnectionsSession` accounts. This change encodes these API objects to be compatible with their equivalent Stripe.js types and includes them in the webview JS message payload. To achieve this:
- Some additional properties were added to the FinancialConnections types that are required by Stripe.js
- The Stripe.js types use snake_case encoding while the Connect JS MessageSender uses camelCase, so custom key-encoding overrides were added to `MessageSender`
- The `returnedFromFinancialConnections` message sender is replaced with a `setCollectMobileFinancialConnectionsResult` setter.

Example message payload:
```
{
    "setter": "setCollectMobileFinancialConnectionsResult",
    "value": {
        "financialConnectionsSession": {  // camelCase
            "accounts": [
                {
                    "balance_refresh": ... // snake_case
                }
            ]
        },
        token: {
            "account_holder_name": ... // snake_case
        }
    }
}
```

The payload of the messenger matches the following shape where [FinancialConnectionsSession.Account](https://github.com/stripe/stripe-js/blob/d212300e2796954faa7f3b4f916847f31b78be10/types/api/financial-connections.d.ts#L125), [BankAccountToken](https://github.com/stripe/stripe-js/blob/d212300e2796954faa7f3b4f916847f31b78be10/types/api/tokens.d.ts#L50-L52), and [StripeError](https://github.com/stripe/stripe-js/blob/d212300e2796954faa7f3b4f916847f31b78be10/types/stripe-js/stripe.d.ts#L1477) match the shapes defined in Stripe.js.
```
{
    financialConnectionsSession?: {
        accounts: FinancialConnectionsSession.Account[];
    };
    token?: BankAccountToken;
    error?: StripeError
}
```

The payload contains the following values depending on the result of the FinanicalConnectionsSheet:
- **Success:** 
  - `financialConnectionsSession.accounts`:  non-empty
  - `token`: non-null
  - `error`: null
- **Canceled:**
  - `financialConnectionsSession.accounts`:  empty
  - `token`: null
  - `error`: null
- **Error:**
  - `financialConnectionsSession.accounts`:  null
  - `token`: null
  - `error`: If the error is an API error, its value will be returned here. If it was a client-side error, the value returned will be null but the error will be logged to analytics.

## Motivation
https://jira.corp.stripe.com/browse/MXMOBILE-2526

## Testing
Unit tests

Manual testing from @mxl-stripe against WIP web changes:

https://github.com/user-attachments/assets/b7b5cc0d-a6c6-4bce-85ef-6457f75a982b

## Changelog
n/a


StripeFinancialConnections/StripeFinancialConnections/Source/FinancialConnectionsSheet.swift
```diff
@@ -26,17 +26,22 @@ final public class FinancialConnectionsSheet {
...
                 session: StripeAPI.FinancialConnectionsSession,
                 token: StripeAPI.BankAccountToken?
             )
         )
-        // Failed with error
+        /// Failed with error
         case failed(error: Error)
-        // User canceled out of the financialConnections session
+        /// User canceled out of the financialConnections session
         case canceled
```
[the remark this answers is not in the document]
mats-stripe: Yup! Thank you 👍

StripeConnect/StripeConnect/Source/Internal/Webview/MessageSenders/SetCollectMobileFinancialConnectionsResultSender.swift
```diff
@@ -0,0 +1,110 @@
...
+            analyticsClient.logClientError(error)
+            return .init(
+                financialConnectionsSession: nil,
+                token: nil,
+                error: nil
+            )
+
+        case .canceled:
+            return .init(
+                financialConnectionsSession: .init(accounts: []),
```
mats-stripe, line 104: I'm curious of the distinction here between initializing this with an empty list of accounts, vs a nil session (in the `.failed` case). Will these scenarios be treated differently, besides an error being logged in the latter case?

mats-stripe (approved):
LGTM!

StripeConnect/StripeConnect/Source/Internal/Webview/MessageSenders/SetCollectMobileFinancialConnectionsResultSender.swift
mxl-stripe: From the stripe js documentation: https://docs.stripe.com/js/financial_connections/collect_financial_connections_accounts, https://docs.stripe.com/js/financial_connections/collect_bank_account_token

> The accounts property will contain the list of accounts collected during this session. If the user chooses to link no accounts, or exits the Authentication Flow early, the accounts array will be empty.

> If the user chooses to link no accounts, or exits the Authentication Flow early, the result object will contain: `result.financialConnectionsSession`: the updated [Financial Connections Session](https://docs.stripe.com/api/financial_connections/session) object. The accounts array will be empty.

So the calling code expects the `accounts` array to be empty when the sheet is cancelled and `null` when the sheet is errored.

mats-stripe: Ahh gotcha, that makes sense! Thanks for the context! 👍

PR #4365, opened 2024-12-17
Print all requests to console in debug mode

wooj-stripe:
## Summary
It's kind of difficult to see when requests are failing. I propose we print requestID to the console so that it's easier to see/debug.

## Motivation
Difficult to view requestId's unless you have a ton of breakpoints in code.

## Testing
Manual testing.

### DIFF
--- a/Stripe/StripeiOSTests/APIRequestTest.swift
+++ b/Stripe/StripeiOSTests/APIRequestTest.swift
@@ -140,6 +140,7 @@ class APIRequestTest: STPNetworkStubbingTestCase {
 
         APIRequest<STPCard>.parseResponse(
             httpURLResponse,
+            method: "GET",
             body: body,
             error: errorParameter
         ) { (object: STPCard?, response, error) in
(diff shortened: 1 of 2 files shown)


davidme-stripe (dismissed):

wooj-stripe:
Changed the approach to log on all requests, and print out the url's relative path.  For example, output looks like:
```
[Stripe SDK]: Request req_T3UJiX9Rbhlh8V, /v1/elements/sessions
[Stripe SDK]: Request req_hM8KJQuvSUvI9r, /v1/customers/cus_RQ08km9gxe2npT
```

mats-stripe (approved):
Nice, might steal this for the FC API client! 

Just curious, can users do anything with these request IDs that might help them debug an error? Or is this ID just something we can use internally to debug?

wooj-stripe:
> Nice, might steal this for the FC API client!
> 
> Just curious, can users do anything with these request IDs that might help them debug an error? Or is this ID just something we can use internally to debug?

I believe uses can log into their dashboard and access request logs based on requestID, but the idea is if they open an issue w/ us and include requests, we'd have more info on figuring out what's going on.

davidme-stripe (dismissed):

davidme-stripe (approved):

PR #4366, opened 2024-12-17
Set correct consent action for Link signups

davidme-stripe:
## Summary
* Send the correct consent action, which is the one from the older Link SDK. Documented it in the consent action chart.
## Motivation
Finalizing Link v3 on Mobile

## Testing
CI

## Changelog
None


StripePaymentSheet/StripePaymentSheet/Source/Helpers/PaymentSheetLinkAccount.swift
```diff
@@ -47,6 +47,8 @@ class PaymentSheetLinkAccount: PaymentSheetLinkAccountInfoProtocol {
         // Inline, email-prefilled
         case implied_v0_0 = "implied_consent_withspm_mobile_v0_0"
 
+        // Clicked button in separate Link sheet
+        case clicked_button_mobile = "clicked_button_mobile"
```
wooj-stripe, line 51: Seems like we should version this similar to the other ones? e.g.:
clicked_button_mobile_v0

wooj-stripe (approved):

PR #4367, opened 2024-12-17
Fetch available incentives after linking bank account

tillh-stripe:
## Summary

This pull request adds a call to `consumers/incentives/update_available` at the end of the Instant Bank Payments flow. This endpoint doesn’t return any available incentives yet, but will do so soon.

Depending on the response, we return `incentiveEligible: Bool` to the MPE. With a future pull request, it’ll update the payment methods list based on the response.

## Motivation

[CONSUMERBANK-570](https://jira.corp.stripe.com/browse/CONSUMERBANK-570)


mats-stripe (dismissed):
LGTM!

mats-stripe (approved):

PR #4369, opened 2024-12-17
Integrate App Attestation backend

davidme-stripe:
## Summary
Integrates App Attest using the backend APIs. This is less scary than it looks, most of the changes are in the tests or just plumbing the `use_attestation_endpoints` value down into Link.

### StripeAttest
* We're now observing the `initial_challenge_required` boolean as another signal of whether to attest. If we're asked to attest an already attested key (or not attest an unattested one), we'll log an error and reset the key.
* We'll also reset the key if any asserted requests return an error. The new `receivedAssertionError` function can be used to react to that error.
* Added a new `prepareAttestation` function, which checks if we're able to perform attestation and attempts to attest a key if we don't have one already. This can be used to switch between the web and native flows, depending on the likelihood that future assertions will succeed.

### PaymentSheet
* Removed the useNativeLink flag, as this behavior is now configured by elements/sessions.
* We now pass the `mobile_app_id` to elements/sessions, enabling it to decide whether to enable `link_mobile_use_attestation_endpoints`.
* Updated STPAPIClient+Link to optionally use the modern attestation-backed APIs if available. Piped the `link_mobile_use_attestation_endpoints` value from elements/sessions to the Link components and API client. (But I'm unhappy that the signup form has to know so much about the API calls!)
* Added a `deviceCanUseNativeLink` to tell us (synchronously) whether the device is capable of attestation, and a `prepareNativeLink` to attempt to attest a key and then return whether we can sign assertions. We use these to help switch between the web and native flows.
* In the niche case where we're in the modern bindings and attestation *doesn't* work, we'll disable signup completely.

### Errata
* Start sending the sessionID and Email source in Link lookup requests. (These should have been sent before, and are required in the new endpoints.)
* Stop logging out of Link automatically after an unsuccessful payment. (This broke 3DS2 testing)
* Enabled attestation in the PaymentSheet Playground app's entitlements.

## Motivation
Enabling App Attestation in Link.

## Testing
Updated all impacted tests. Re-recorded network tests to include the new `mobile_app_id` parameter.

## Changelog
Not yet released to users.


StripeCore/StripeCore/Source/Attestation/StripeAttestBackend.swift
```diff
@@ -6,23 +6,37 @@
...
 
     @_spi(STP) public init(apiClient: STPAPIClient) {
         self.apiClient = apiClient
     }
 
-    public func attest(appId: String, deviceId: String, keyId: String, attestation: Data) async throws {
-        stpAssertionFailure("Not implemented")
+    @_spi(STP) public func attest(appId: String, deviceId: String, keyId: String, attestation: Data) async throws {
+        let _: EmptyResponse = try await withCheckedThrowingContinuation { continuation in
+            apiClient.post(resource: "mobile_sdk_attestation/ios_attest", parameters: ["app_id": appId, "device_id": deviceId, "key_id": keyId, "attestation_object": attestation.base64EncodedString()]) { result in
```
mats-stripe, line 27: Couple suggestions here that I think would help clean this up a tiny bit; 

1. Could the base parameters (`appId`, `deviceId`, `keyId`) be set on class / protocol? This way they don't need to be passed in both methods requesting them. Though this wouldn't really work if the parameters could change between the initialization time and the function call time. Something like:

```swift
protocol StripeAttestBackend {
    var appId: String { get }
    var deviceId: String { get }
    var keyId: String { get }

    func getChallenge() async throws -> StripeChallengeResponse
    func attest(attestation: Data) async throws
}
```

2. It'd be nice to maybe encapsulate some of this endpoint logic into an enum along the lines of: 

```swift
enum Endpoint {
    case attest(attestation: Data)
    case challenge
    
    var path: String {
        switch self {
        case .attest: "mobile_sdk_attestation/ios_attest"
        case .challenge: "mobile_sdk_attestation/ios_challenge"
        }
    }
}
```

3. That alone would be nice, but it would pair well with a function on `StripeAPIAttestationBackend` to build the parameters object:

```swift
private func parameters(for endpoint: Endpoint) -> [String: String] {
    var baseParams: [String: String] = [
        "app_id": appId,
        "device_id": deviceId,
        "key_id": keyId,
    ]
    
    switch endpoint {
    case .attest(let attestation):
        let attestationObject = ["attestation_object": attestation.base64EncodedString()]
        return baseParams.merging(attestationObject) { (current, _) in current }
    case .challenge:
        return baseParams
    }
}
```

4. And finally, these two functions would be implemented as:

```swift
public func attest(attestation: Data) async throws {
    let _: EmptyResponse = try await withCheckedThrowingContinuation { continuation in
        let endpoint = Endpoint.attest(attestation: attestation)
        apiClient.post(resource: endpoint.path, parameters: parameters(for: endpoint))) { result in
            continuation.resume(with: result)
        }
    }
}
```

Feel free to cherry pick anything you like or even disregard! It really all comes down to personal preference

mats-stripe (dismissed):
Lot going on here, especially in areas I'm not too familiar with, so I might've missed something. Would be great to get another review before merging! 

However, this is all very helpful for my upcoming work in FC! Left a few comments and suggestions, but overall LGTM!

Example/PaymentSheet Example/PaymentSheet Example/PaymentSheet Example.entitlements
```diff
@@ -2,6 +2,8 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>com.apple.developer.devicecheck.appattest-environment</key>
+	<string>development</string>
```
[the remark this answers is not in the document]
davidme-stripe: It applies to the app, users will need to flip it on in Xcode. (Will send you the doc)

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
```diff
@@ -315,7 +315,9 @@ extension PaymentSheet {
                         with: authenticationContext,
                         completion: { actionStatus, _, error in
                             paymentHandlerCompletion(actionStatus, error)
-                            linkAccount?.logout()
+                            if actionStatus == .succeeded {
+                                linkAccount?.logout()
```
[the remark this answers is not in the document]
davidme-stripe: Yes, otherwise if a payment is cancelled (e.g. during a 3DS2 challenge) or fails (for insufficient funds), the user would need to log in again before trying with another card.

StripeCore/StripeCore/Source/Attestation/StripeAttestBackend.swift
davidme-stripe: Thanks! I'm thinking about this, but I'd rather not share the parameter generating code between these two API calls: They do very different things (even if the first few parameters are the same) and I'd find the above `Endpoint` enum to be a little harder to reason about when debugging. I might just be stuck in Objective-C brain though 😅 

> Though this wouldn't really work if the parameters could change between the initialization time and the function call time. 

Yeah, that's the main issue — the key could change between two subsequent requests, so I think it's clearer if `StripeAttest.keyId` is the source of truth instead of having the backend object store its own key. App ID and Device ID will never change during the lifetime of the process though.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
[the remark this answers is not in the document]
mats-stripe: Makes sense!

StripeCore/StripeCore/Source/Attestation/StripeAttestBackend.swift
mats-stripe: Makes sense! This definitely added some complexity that ultimately probably isn't worth it 👍

StripeCore/StripeCore/Source/Attestation/StripeAttest.swift
```diff
@@ -26,10 +26,51 @@ import UIKit
...
+        // If the backend responds with this specific error, we should reset the key.
+        if let error = error as? StripeCore.StripeError,
+           case let .apiError(apiError) = error,
+           apiError.code == "invalid_parameter",
+           apiError.type == .invalidRequestError,
+           apiError.param == nil {
+            let resetKeyAnalytic = ErrorAnalytic(event: .resetKeyForAssertionError, error: error)
+            STPAnalyticsClient.sharedClient.log(analytic: resetKeyAnalytic, apiClient: apiClient)
+            resetKey()
+        }
```
wooj-stripe, line 66: Seems odd that we only log invalid_parameter errors.  Shouldn't we log others?

StripePaymentSheet/StripePaymentSheet/Source/Helpers/PaymentSheetLinkAccount.swift
```diff
@@ -53,6 +53,10 @@ class PaymentSheetLinkAccount: PaymentSheetLinkAccountInfoProtocol {
     let apiClient: STPAPIClient
     let cookieStore: LinkCookieStore
 
+    let useModernMobileEndpoints: Bool
```
wooj-stripe, line 56: I don't have much context here, but "modern" MobileEndpoints feels weird.  Feels like we'll look back on this in a year and think that it isn't so modern anymore.

StripeCore/StripeCore/Source/Attestation/StripeAttest.swift
davidme-stripe: StripeAttest isn't interested in other errors — this is the error from the endpoint being asserted against (might be a `Link` endpoint, might be `PaymentIntents/create`, etc), so we only handle it if it's one specific type of error.

StripePaymentSheet/StripePaymentSheet/Source/Helpers/PaymentSheetLinkAccount.swift
davidme-stripe: Ideally we'd delete the old ones someday! I also considered something like `useMobileSpecificEndpoints`, but I'm not sure if that's much clearer. `useAttestationEndpoints`?

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/STPAPIClient+Link.swift
```diff
@@ -110,11 +112,13 @@ extension STPAPIClient {
             }
 
             post(
-                resource: useModernMobileEndpoints ? modernEndpoint : legacyEndpoint,
+                resource: useMobileEndpoints ? modernEndpoint : legacyEndpoint,
```
wooj-stripe, line 115: nit: maybe we should update modern endpoint to `mobileEndpoint` (At this point modernEndpoint looks like a typo)

wooj-stripe (approved):

PR #4372, opened 2024-12-18
"Card details cannot be changed" text size closer to form label size

joyceqin-stripe:
## Summary

Changed smallFootnote to font size 11.5
The form label font is subheadline, but when the field has text or is being edited, it moves up and scales to 75%. I tried making the smallFootnote font subheadline and scaling it, but it scaled the label position. 75% of 14pt is 10.5

## Motivation

https://jira.corp.stripe.com/browse/MOBILESDK-2878
## Testing

image
image


StripeUICore/StripeUICore/Source/Elements/ElementsUI.swift
```diff
@@ -100,7 +100,7 @@ import UIKit
                                             withTextStyle: .caption1,
                                             maximumPointSize: 20)
         public var footnote = UIFont.preferredFont(forTextStyle: .footnote, weight: .regular, maximumPointSize: 20)
-        public var smallFootnote = UIFont.preferredFont(forTextStyle: .footnote, weight: .medium, maximumPointSize: 10)
+        public var smallFootnote = UIFontMetrics(forTextStyle: .footnote).scaledFont(for: .systemFont(ofSize: 10.5, weight: .medium))
```
porter-stripe, line 103: My slight preference is to not use half point sizes, and prefer the text styles. Can we get away with using `caption2`, it's 11.

joyceqin-stripe: My assumption is that we'd want the sublabel and the error text to be the same size— is that right? If so, I'm currently sending Hannah a screenshot with the error text set to caption2, and if she says that's okay, then we can go through with it.

porter-stripe (approved):

PR #4373, opened 2024-12-18
Added background color to disabled text fields on manage PM screen

joyceqin-stripe:
## Summary

Added a translucent mask over componentBackground color to make a disabled appearance
## Motivation

https://jira.corp.stripe.com/browse/MOBILESDK-2880
## Testing

test_UpdatePaymentMethodViewControllerLightMode@3x
test_UpdatePaymentMethodViewControllerUSBankAccountDarkMode@3x
test_UpdatePaymentMethodViewControllerSEPADebitAppearance@3x


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewModel.swift
```diff
@@ -56,7 +93,9 @@ class UpdatePaymentMethodViewModel {
             fatalError("Unsupported payment type \(paymentMethod.type) in UpdatePaymentMethodViewModel")
         }
         self.paymentMethod = paymentMethod
-        self.appearance = appearance
+        var disabledAppearance = appearance
+        disabledAppearance.colors.componentBackground = disabledBackgroundColor(componentBackground:    disabledAppearance.colors.componentBackground)
+        self.appearance = disabledAppearance
```
porter-stripe, line 98: Do we not use this new disabled look all though PaymentSheet? The fields are also disable when we are in the processing state.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewModel.swift
```diff
@@ -12,6 +12,43 @@ import Foundation
 @_spi(STP) import StripeUICore
 import UIKit
 
+// calculate the resulting color of the mask overlaying the component background
+private func disabledBackgroundColor(componentBackground: UIColor) -> UIColor {
```
porter-stripe, line 16: This feels quite complex for generating a background color. What's the formula for this in english? Is it really necessary?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewModel.swift
joyceqin-stripe: I'm asking to check if we're supposed to use it throughout all of PaymentSheet. If so, then I'll redo this so that we have a translucent view overlaying the text fields and just set the background color of the overlay view. I didn't want to do that when I thought it was only for the update pm screen because that would affect the text color, which is why I did the calculations for the background color instead.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewModel.swift
porter-stripe: If we keep it we should put it in it's own file or find an appropriate file and make it an extension on `UIColor`.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewModel.swift
joyceqin-stripe: Hannah said it should only apply to the text fields on the update pm screen. I'll move the color calculation to another file though

StripeUICore/StripeUICore/Source/Categories/UIColor+StripeUICore.swift
```diff
@@ -215,4 +230,30 @@ private extension UIColor {
         })
     }
 
+    /// Applies a mask over a base color
+    ///
+    /// - Parameters:
+    ///   - overlayColor: The color to overlay
+    ///   - baseColor: The base color
+    /// - Returns: The computed color of the overlay color on the base color.
+    func overlayColor(overlayColor: UIColor, baseColor: UIColor) -> UIColor {
```
porter-stripe, line 239: Is Android computing the overlay color the same?

joyceqin-stripe: I'm not sure, I'll ask

StripeUICore/StripeUICore/Source/Elements/Section/SectionContainerView.swift
```diff
@@ -31,25 +31,27 @@ class SectionContainerView: UIView {
     }()
 
     lazy var stackView: StackViewWithSeparator = {
-        let view = buildStackView(views: views, theme: theme)
+        let view = buildStackView(views: views, theme: theme, setMask: setMask)
```
porter-stripe, line 34: I think this is getting closer to what we want, but I think `setMask` is a little too vague, and I don't think it should be a required parameter for creating a SectionContainerView. We should look at doing something like adding a function/API to the SectionContainerView to put it in a "disabled" state.

joyceqin-stripe: I've added a function to the SectionElement called disableAppearance, and it calls a function on the SectionContainerView that applies the mask

porter-stripe (approved):

PR #4378, opened 2024-12-19
Attest separately for each API key and environment

davidme-stripe:
## Summary
We don't share attestation keys across API keys, so we need a separate attestation key for each API key. This slightly increases the number of attestations, but prevents situations where switching between merchants causes attestations to fail until the hardcoded timeout.

We'll also relax the 24 hour limit when using the development App Attest environment, as it doesn't increment the device's risk metric. We will still enforce it for generating keys in the production environment. To determine if we're in the development environment, we'll look for the appattestdevelop aaguid in the attestation data blob.

## Motivation
Switching between merchants and environments without waiting for the 24 hour attestation failsafe to time out.

## Testing
Updated existing tests, added a new test, tested attestations on physical device.

## Changelog
None


StripeCore/StripeCore/Source/Attestation/StripeAttest.swift
```diff
@@ -117,33 +117,42 @@ import UIKit
...
         }
         set {
-            UserDefaults.standard.set(newValue, forKey: DefaultsKeys.lastAttestedDate.rawValue)
+            UserDefaults.standard.set(newValue, forKey: defaultsKeyForSetting(.lastAttestedDate))
         }
     }
 
+    /// The key to use for storing an attestation key in NSUserDefaults.
+    func defaultsKeyForSetting(_ setting: SettingsKeys) -> String {
+        var key = "\(setting.rawValue):\(apiClient.publishableKey ?? "unknown")"
```
wooj-stripe, line 149: Storing "unknown:xxx" in NSUserDefaults feels awkward. If we don't have a publishable key, seems like we should just bail?

StripeCore/StripeCoreTests/Attestation/MockAppAttestService.swift
```diff
@@ -66,10 +67,14 @@ class MockAppAttestService: AppAttestService {
...
         // Generate a fake attestion
-        let attestation = ["keyID": key.id, "counter": key.counter, "clientDataHash": clientDataHash.base64EncodedString()] as [String: Any]
+        let attestation = ["keyID": key.id, "counter": key.counter, "clientDataHash": clientDataHash.base64EncodedString(), "isDevelopmentEnvironment": attestationUsingDevelopmentEnvironment] as [String: Any]
         return try JSONSerialization.data(withJSONObject: attestation)
     }
 
+    @_spi(STP) public func attestationDataIsDevelopmentEnvironment(_ data: Data) -> Bool {
+        let decodedKey = try! JSONSerialization.jsonObject(with: data) as! [String: Any]
+        return decodedKey["isDevelopmentEnvironment"] as! Bool
+    }
```
wooj-stripe, line 77: Feels like we can avoid using `as!` and fail gracefully?  E.g. if we can't decode, default to development.

davidme-stripe: This is test code, so I want it to fail loudly

StripeCore/StripeCore/Source/Attestation/StripeAttest.swift
davidme-stripe: Let's bail above in `getOrCreateKey()` (though I'll still safely unwrap here so we don't crash)

StripeCore/StripeCore/Source/Attestation/StripeAttest.swift
```diff
@@ -117,33 +119,42 @@ import UIKit
         case successfullyAttested = "STPAttestKeySuccessfullyAttested"
     }
 
-    private static var keyID: String? {
+    private var keyID: String? {
```
wooj-stripe, line 122: keyID is used way too many times in this file, can we rename this to something like `storedKeyID` ?

StripeCore/StripeCore/Source/Attestation/StripeAttest.swift
wooj-stripe: Maybe we take a copy of apiClient.publishableKey when the class is initialized?

wooj-stripe (approved):

PR #4379, opened 2024-12-19
Update StripeConnect's FinancialConnections integration to work with public key override.

nschris-stripe:
## Summary
- Ensures the financial connection integration uses the public key override.
- Updates implementation to make it more testable.


StripeConnect/StripeConnect/Source/Internal/FinancialConnectionsPresenter.swift
```diff
@@ -9,21 +9,35 @@
...
-            financialConnectionsSessionClientSecret: clientSecret
-        )
+        let financialConnectionsSheet = financialConnectionSheetInitializer(clientSecret, nil)
         // FC needs the connected account ID to be configured on the API Client
         // Make a copy before modifying so we don't unexpectedly modify the shared API client
-        financialConnectionsSheet.apiClient = apiClient.makeCopy()
+        financialConnectionsSheet.apiClient = componentManager.apiClient.makeCopy()
+        
+        // FC expects a public key and not a UK. If there is a public key override we should use that.
+        financialConnectionsSheet.apiClient.publishableKey = componentManager.publicKeyOverride ?? financialConnectionsSheet.apiClient.publishableKey
```
mludowise-stripe, line 40: nit: I'd rather keep the logic for copying the PK encapsulated in `makeCopy()`, in the event that ever needs to change. So instead of defaulting this value to `financialConnectionsSheet.apiClient.publishableKey`, can we instead only override it if there is a PK override?
```suggestion
        if let publicKeyOverride = componentManager.publicKeyOverride {
            financialConnectionsSheet.apiClient.publishableKey = publicKeyOverride
        }
```

StripeConnect/StripeConnect/Source/Internal/FinancialConnectionsPresenter.swift
```diff
@@ -9,21 +9,35 @@
...
 import UIKit
 
+protocol FinancialConnectionSheetImplementation: AnyObject {
+    init(financialConnectionsSessionClientSecret: String, returnURL: String?)
+    var apiClient: STPAPIClient { get set}
+    func presentForToken(
+        from presentingViewController: UIViewController,
+        completion: @escaping (FinancialConnectionsSheet.TokenResult) -> Void
+    )
+}
```
mludowise-stripe, line 19: Instead of creating a DI protocol, you could add a separate unit-testable function that is a factory method for the sheet:
```
class FinancialConnectionsPresenter {
    func makeSheet(
        componentManager: EmbeddedComponentManager,
        clientSecret: String,
        connectedAccountId: String
    ) -> FinancialConnectionsSheet {
        let financialConnectionsSheet = FinancialConnectionsSheet(...)
        financialConnectionsSheet.apiClient = componentManager.apiClient.makeCopy()
        financialConnectionsSheet.apiClient.stripeAccount = ...
        financialConnectionsSheet.apiClient.publishableKey = ...
        return financialConnectionsSheet
    }

    func presentForToken(...) {
        let financialConnectionsSheet = makeSheet()
        ...
    }
}

class FinancialConnectionsPresenterTests: XCTestCase {
    func testMakeSheet() {
        let sheet = FinancialConnectionsPresenter().makeSheet(...)
        // Verify APIClient configured on `sheet`
    }
}
```

mludowise-stripe (changes requested):

StripeConnect/StripeConnect/Source/Internal/FinancialConnectionsPresenter.swift
nschris-stripe: This is definitely simpler, but it does have the drawback of not being able to test that the client secret was set correctly.

mludowise-stripe: Why not? The `financialConnectionsSessionClientSecret` property is publicly accessible, so we should able to XCAssertEqual on it.
https://github.com/stripe/stripe-ios/blob/d623be31/StripeFinancialConnections/StripeFinancialConnections/Source/FinancialConnectionsSheet.swift#L53

nschris-stripe: 🤦‍♂️ Oops, I was looking for clientSecret.

mludowise-stripe (dismissed):
lgtm -- not sure if you wanted to add an xctassertequal on the client secret. If so, lmk and I'll reshipit

mludowise-stripe:
Looks like `testOpenFinancialConnections` tests need to get updated `ConnectComponentWebViewControllerTests`. Looks like they're failing because the method signature changed for `overridePresentForToken`

StripeConnect/StripeConnectTests/Internal/Webview/ConnectComponentWebViewControllerTests.swift
```diff
@@ -463,8 +463,9 @@ class ConnectComponentWebViewControllerTests: XCTestCase {
         let componentManager = componentManagerAssertingOnFetch()
         let session = try FinancialConnectionsSessionMock.default.make()
 
-        let financialConnectionsPresenter = MockFinancialConnectionsPresenter { apiClient, secret, connectedAccountId, vc in
-            XCTAssert(apiClient === componentManager.apiClient)
+        let financialConnectionsPresenter = MockFinancialConnectionsPresenter { compManager, secret, connectedAccountId, vc in
+            XCTAssert(compManager.apiClient == componentManager.apiClient)
+            XCTAssert(compManager.publicKeyOverride == componentManager.publicKeyOverride)
```
mludowise-stripe, line 468: Doublechecking – STPAPIClient subclasses NSObject and doesn't override `isEqual`, so I think this should just do pointer-comparison, right? Just want to make sure the equality comparison does what we think it should.

Non-blocking: Also, if we're using `==` anyway, this could be `XCTAssertEqual` which would provide a better failure message if the test fails.
```suggestion
            XCTAssertEqual(compManager.apiClient, componentManager.apiClient)
            XCTAssertEqual(compManager.publicKeyOverride, componentManager.publicKeyOverride)
```

mludowise-stripe (approved):

StripeConnect/StripeConnectTests/Internal/Webview/ConnectComponentWebViewControllerTests.swift
nschris-stripe: Yeah this should be doing pointer comparison.

PR #4382, opened 2024-12-19
Fix issues with concurrent attestations

davidme-stripe:
## Summary
I realized that some of this code is unsafe if run concurrently, so I:
* Adopted actors to protect stored properties in StripeAttest and the MockAppAttestService
* Used a stored `Task` to only allow one simultaneous attestation to occur

Now we can kick off as many simultaneous `assert()` requests as we'd like, and they all seem to complete without issues.

Note that I marked the appAttestService as `nonisolated` in order to access the `isSupported` boolean synchronously. I _think_ this is fine (DCAppAttestService appears to be set up to handle this correctly), but this is our first use of actors in the SDK, so let me know if it seems odd!

## Motivation
Fixing concurrency-related crashes and unexpected behavior

## Testing
Added a test, tested attestation on device

## Changelog
None

### DIFF
--- a/StripeCore/StripeCore/Source/Attestation/AppAttestService.swift
+++ b/StripeCore/StripeCore/Source/Attestation/AppAttestService.swift
@@ -7,7 +7,7 @@ import DeviceCheck
 import Foundation
 
 @_spi(STP) public protocol AppAttestService {
-    var isSupported: Bool { get }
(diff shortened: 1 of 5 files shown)


mats-stripe (approved):
While I'm not support familiar with `actor`s and `nonisolated` properties, these change seem sensible and the test coverage looks good!

PR #4422, opened 2024-12-20
Remove space for bank and SEPA

tillh-stripe:
## Summary

This pull request removes the space before the last4 again for bank and SEPA.


davidme-stripe (approved):

davidme-stripe:
Actually, I think we should use a tab

PR #4424, opened 2024-12-23
Change default playground & example app layout from horizontal to automatic

yuki-stripe:
## Summary
See title

## Motivation
It's the default API now so our playground should match.


davidme-stripe (approved):

davidme-stripe:
Unfortunate that all the tests are forced to `.horizontal`, but I think we should rethink some of these E2E tests anyway in the new year. (Many are redundant!)

porter-stripe (dismissed):

porter-stripe (approved):

PR #4426, opened 2024-12-23
Fix concurrent assertions

davidme-stripe:
## Summary
The server enforces that the [counter attached to the assertion increases with each assertion](https://developer.apple.com/documentation/devicecheck/validating-apps-that-connect-to-your-server?language=objc#Verify-the-assertion). This means if one client happens to attempt to send multiple assertions at once, and the network requests arrive out of order, we'll fail the assertion.

To work around this, we'll change the API so that the `StripeAttest` returns an `AssertionHandle`. It's required to call `complete()` with this handle, or no future assertions will proceed.

I'm also going to make `stripeAttest` a parameter of an STPAPIClient — now that there's more state involved, it makes more sense for an STPAPIClient to own this instead of creating one each time.

## Testing
CI, added a new test, tested attestation on device.


StripeCore/StripeCore/Source/Attestation/StripeAttest.swift
```diff
@@ -365,10 +389,43 @@ import UIKit
...
+            self.assertion = assertion
+            self.stripeAttest = stripeAttest
+        }
+
+        // Must be called by the caller when done with the assertion
+        public func complete() {
+            guard let stripeAttest = stripeAttest else { return }
+            Task {
+                await stripeAttest.assertionCompleted()
+            }
```
wooj-stripe, line 428: If stripeAttest goes out of scope it seems that all the callers of stripeattest.assert() will be awaiting forever. It think is this unlikely to happen, and the correct way to implement this, but maybe we should add an stpAssert in the else clause of the guard e.g.:
```
guard let stripeAttest else {
    stpAssert(...)
    return
}
 Task {
     await ..
 } 
```

StripeCore/StripeCore/Source/Attestation/StripeAttest.swift
```diff
@@ -18,18 +18,31 @@ import UIKit
...
             let assertion = try await _assert()
             let successAnalytic = GenericAnalytic(event: .assertionSucceeded, params: [:])
-            STPAnalyticsClient.sharedClient.log(analytic: successAnalytic, apiClient: apiClient)
-            return assertion
+            if let apiClient {
+                STPAnalyticsClient.sharedClient.log(analytic: successAnalytic, apiClient: apiClient)
+            }
+            return AssertionHandle(assertion: assertion, stripeAttest: self)
         } catch {
             let errorAnalytic = ErrorAnalytic(event: .assertionFailed, error: error)
```
wooj-stripe, line 39: If there is an exception, are users expected to call `complete()`? If so, that seems odd.  Seems like if there is an exception, we should set assertionInProgress to false and continue with the next one that is queue'd up?

davidme-stripe: Good catch! Will fix this and add a test.

StripeCore/StripeCore/Source/Attestation/StripeAttest.swift
davidme-stripe: I'll add an assert here, though I don't think it's possible for `stripeAttest` to go out of scope while it has other callers waiting on it.

wooj-stripe (approved):

PR #4428, opened 2024-12-24
Fix missing Link logo when logging out

davidme-stripe:
## Summary
If the Link button is set to show the user's email address, then the user logs out, the button would be blank. This is because we forgot to unset `isHidden` on the payWithLink logo view when updating the UI.

## Testing
Manually in PS Example, and added a snapshot test.


Stripe/StripeiOSTests/PayWithLinkButtonSnapshotTests.swift
```diff
@@ -69,6 +69,13 @@ class PayWithLinkButtonSnapshotTests: STPSnapshotTestCase {
         verify(sut)
     }
 
+    func testAddThenRemoveAccount() {
+        let sut = makeSUT()
```
sfriedman-stripe, line 73: What is a `sut`?

sfriedman-stripe (approved):

Stripe/StripeiOSTests/PayWithLinkButtonSnapshotTests.swift
davidme-stripe: "System Under Test": It's a standard term in our tests (since 2015!), but I don't know who introduced it.

PR #4432, opened 2024-12-24
Add interactive 2FA refresh for Link

davidme-stripe:
## Summary
We currently have some code to refresh the session using a token, but we don't store a persistent login token in the app, so it won't work. Instead, we'll pop the 2FA prompt again. This requires communicating that state up from the PaymentSheetLinkAccount to the PayWithLinkViewController so that we can present the authentication view controller.

## Motivation
Allowing use of Link after the token expires.

## Testing
Added logic test, tested manually in PS Example.

## Changelog
None


StripePaymentSheet/StripePaymentSheet/Source/Helpers/PaymentSheetLinkAccount.swift
```diff
@@ -238,19 +236,19 @@ class PaymentSheetLinkAccount: PaymentSheetLinkAccountInfoProtocol {
...
-        guard let session = currentSession else {
-            stpAssertionFailure()
-            completion(
-                .failure(
-                    PaymentSheetError.unknown(
-                        debugDescription: "Linking account session without valid consumer session"
+        retryingOnAuthError(completion: completion) { [publishableKey] completionWrapper in
+            guard let session = self.currentSession else {
+                stpAssertionFailure()
+                completion(
```
wooj-stripe, line 242: This code is getting harder to read, because it's unclear when to call the completion vs the completionWrapper. I think your code is correct, but do you have any thoughts on how to make it more clear?

StripePaymentSheet/StripePaymentSheet/Source/Helpers/PaymentSheetLinkAccount.swift
```diff
@@ -238,19 +236,19 @@ class PaymentSheetLinkAccount: PaymentSheetLinkAccountInfoProtocol {
     func createLinkAccountSession(
         completion: @escaping (Result<LinkAccountSession, Error>) -> Void
     ) {
-        guard let session = currentSession else {
-            stpAssertionFailure()
-            completion(
-                .failure(
-                    PaymentSheetError.unknown(
-                        debugDescription: "Linking account session without valid consumer session"
+        retryingOnAuthError(completion: completion) { [publishableKey] completionWrapper in
```
wooj-stripe, line 239: Is there a concern that the publishableKey will be changing after making this call? I think this might be introducing more complexity than needed?

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController.swift
```diff
@@ -103,7 +103,11 @@ final class PayWithLinkViewController: UINavigationController {
     private var accountContext: LinkAccountContext = .shared
 
     private var linkAccount: PaymentSheetLinkAccount? {
-        get { accountContext.account }
+        get {
+            let account = accountContext.account
+            account?.paymentSheetLinkAccountDelegate = self
+            return account
```
wooj-stripe, line 109: Sorry, i'm not following why this is needed.

StripePaymentSheet/StripePaymentSheet/Source/Helpers/PaymentSheetLinkAccount.swift
davidme-stripe: The reformatting of this file is making the diff confusing, but I'm only moving the `guard let session = currentSession` into the retryingOnAuthError block: The session could change after making this call, as `retryingOnAuthError` could create a new session and change the `currentSession`.

StripePaymentSheet/StripePaymentSheet/Source/Helpers/PaymentSheetLinkAccount.swift
davidme-stripe: Agreed, let me try making this more obvious. I haven't really changed this code, I'm just moving it around.

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController.swift
davidme-stripe: The `PaymentSheetLinkAccount` needs to be able to communicate with the `PayWithLinkViewController` to inform it of updates. So when a `linkAccount` is set, the PayWithLinkViewController needs to set itself as the PaymentSheetLinkAccount's delegate. But maybe I can make this cleaner.

StripePaymentSheet/StripePaymentSheet/Source/Helpers/PaymentSheetLinkAccount.swift
davidme-stripe: I'll remove the explicit captures and just use `self` in these functions, maybe that'll make it less confusing.

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController.swift
davidme-stripe: Maybe this is clearer? https://github.com/stripe/stripe-ios/pull/4432/commits/b4e379716d7b10204086f8fdff2c1c3572b89e3c#diff-7d5339851903df687dc841f3deb388ba566a2f0b319c57c38671cb529b80e7a4R185

StripePaymentSheet/StripePaymentSheet/Source/Helpers/PaymentSheetLinkAccount.swift
davidme-stripe: Okay, renamed the variable to make it clearer. `completion` calls the completion block, while `completionRetryingOnAuthErrors` is the wrapped completion that checks for auth errors before completing.

StripePaymentSheet/StripePaymentSheet/Source/Helpers/PaymentSheetLinkAccount.swift
```diff
@@ -452,38 +460,9 @@ private extension PaymentSheetLinkAccount {
...
-                case .noAvailableLookupParams:
-                    completion(
-                        .failure(PaymentSheetError.missingClientSecret)
-                    )
-                }
-            case .failure(let error):
-                completion(.failure(error))
-            }
-        }
+        paymentSheetLinkAccountDelegate?.refreshLinkSession(completion: completion)
```
wooj-stripe, line 465: Seems like if `paymentSheetLinkAccountDelegate` is not set, the completion will not get called.  Seems like we should add a guard statement like..
```
guard let delegate = paymentSheetLinkAccountDelegate {
   stpAssertionFailure()
   completion(failure)
   return 
}
delegate.refreshLinkSession(completion:completion)
```

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController.swift
wooj-stripe: If we add onAccountChange Observer, then seems like we can revert the this change back to:
```
get { accountContext.account }
```
?

davidme-stripe: Ah yes, I did but accidentally stashed. Pushing that!

wooj-stripe (approved):

PR #4435, opened 2024-12-24
Bail to web-based Link when needed

davidme-stripe:
## Summary
Add support for bailing back to the web Link flow when attestation fails. This makes launching Link more responsive for the expected case (that attestation will succeed if the app has attestation enabled and the device supports attestation), while allowing a graceful fallback in the unexpected case (the attestation or assertion fails). We'll bail if either:
1. Initial attestation fails just after the sheet opens
2. If any asserted network request fails

## Motivation
Handle unexpected errors in a more user-friendly way.

## Testing
Tested manually by intentionally breaking assertion and attestation in both cases.
Added a test that uses mocks to fail attestation and check that the ASWebAuthenticationSession is presented.

## Changelog
None


StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkWebController.swift
```diff
@@ -72,23 +72,27 @@ final class PayWithLinkWebController: NSObject, ASWebAuthenticationPresentationC
...
         var lastAddedPaymentDetails: ConsumerPaymentDetails?
+        let alwaysUseEphemeralSession: Bool
 
         /// Creates a new Context object.
         /// - Parameters:
         ///   - intent: Intent.
         ///   - elementsSession: STPElementsSession.
         ///   - configuration: PaymentElementConfiguration configuration.
         ///   - callToAction: A custom CTA to display on the confirm button. If `nil`, will display `intent`'s default CTA.
+        ///   - alwaysUseEphemeralSession: If `true`, always use an ephemeral session.
```
wooj-stripe, line 83: what is the significance of 'always' here?  Seems like it could just be called useEphemeralSession?

davidme-stripe: https://github.com/stripe/stripe-ios/pull/4435/files#diff-3586b686dbee34f78db7cb9f98263219ac168f2ead0b422686cab88b80f767f8R144

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController.swift
```diff
@@ -354,6 +369,30 @@ extension PayWithLinkViewController: PayWithLinkCoordinating {
...
+        let payWithLinkVC = PayWithLinkWebController(
+            intent: context.intent,
+            elementsSession: context.elementsSession,
+            configuration: context.configuration,
+            alwaysUseEphemeralSession: true
+        )
+        payWithLinkVC.payWithLinkDelegate = payWithLinkDelegate as? PayWithLinkWebControllerDelegate
+        self.dismiss(animated: false)
+        payWithLinkVC.present(over: presentingViewController)
+        STPAnalyticsClient.sharedClient.logLinkBailedToWebFlow()
```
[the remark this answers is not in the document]
davidme-stripe: It could also be caused by failing to assert, not just attest. We don't want to send the analytic until after we've performed the above checks (ensuring we only bail once), so I think it makes sense to do it at the end here.

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController.swift
```diff
@@ -354,6 +369,30 @@ extension PayWithLinkViewController: PayWithLinkCoordinating {
...
+        // Set up a web controller with the same settings and swap to it
+
+        let payWithLinkVC = PayWithLinkWebController(
+            intent: context.intent,
+            elementsSession: context.elementsSession,
+            configuration: context.configuration,
+            alwaysUseEphemeralSession: true
+        )
+        payWithLinkVC.payWithLinkDelegate = payWithLinkDelegate as? PayWithLinkWebControllerDelegate
+        self.dismiss(animated: false)
```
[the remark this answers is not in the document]
davidme-stripe: * The presentingViewController could be anything (I think even a user's VC in the FlowController case), so we don't really care what type it is. It's just the VC that is presenting us.
* I'll add comments to make it more obvious what we're trying to do here — we're removing ourselves and swapping in the ASWebAuthenticationSession.
* If payWithLinkDelegate is not set, I'd assume we it had fallen out of scope, so we'd probably want to bail and assert yes.

StripeCore/StripeCore/Source/Analytics/STPAnalyticEvent.swift
```diff
@@ -149,6 +149,7 @@ import Foundation
     case link2FAComplete = "link.2fa.complete"
     case link2FACancel = "link.2fa.cancel"
     case link2FAFailure = "link.2fa.failure"
+    case linkBailedToWebFlow = "link.native.bailed"
```
wooj-stripe, line 152: nit: case should be named `linkNativeBailed`

wooj-stripe (dismissed):

StripeCore/StripeCore/Source/Analytics/STPAnalyticEvent.swift
davidme-stripe: Fixed!

wooj-stripe (approved):

PR #4438, opened 2025-01-02
Remove extra CY country code mapping

jaynewstrom-stripe:
# Summary

This was incorrect. See https://en.wikipedia.org/wiki/Telephone_numbers_in_Cyprus

# Motivation

https://jira.corp.stripe.com/browse/RUN_MOBILESDK-3773


StripeUICore/StripeUICore/Source/Validators/PhoneNumber.swift
```diff
@@ -343,7 +343,6 @@ import UIKit
             Metadata(prefix: "+509", regionCode: "HT", pattern: "## ## ####"),
             Metadata(prefix: "+51", regionCode: "PE", pattern: "### ### ###"),
             Metadata(prefix: "+52", regionCode: "MX", pattern: "### ### ####"),
-            Metadata(prefix: "+537", regionCode: "CY", pattern: ""),
```
jaynewstrom-stripe, line 346: See the existing line that correctly handles this here: https://github.com/stripe/stripe-ios/blob/master/StripeUICore/StripeUICore/Source/Validators/PhoneNumber.swift#L292

joyceqin-stripe (approved):

PR #4449, opened 2025-01-09
Header text says manage cards instead of manage payment methods when only cards available

joyceqin-stripe:
## Summary

In vertical mode, when the only saved payment methods are cards, in edit mode, it now says "Manage cards" instead of "Manage payment methods" since in non-edit mode it says "Select card" instead of "Select payment method"
## Motivation

[MOBILESDK-2890](https://jira.corp.stripe.com/browse/MOBILESDK-2890)


porter-stripe (commented):
Can you explain why it's ok to skip translations here? As far as I can tell this will be immediately visible, resulting in un-localized portions of our UI, which would be a regression. We probably need to wait for translations to come in until we ship this.

joyceqin-stripe:
> Can you explain why it's ok to skip translations here? As far as I can tell this will be immediately visible, resulting in un-localized portions of our UI, which would be a regression. We probably need to wait for translations to come in until we ship this.

I forgot this wasn't specifically for the default sync project, my bad

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/Vertical Saved Payment Method Screen/VerticalSavedPaymentMethodsViewController.swift
```diff
@@ -70,12 +70,15 @@ class VerticalSavedPaymentMethodsViewController: UIViewController {
     }
 
     private var headerText: String {
+        let nonCardPaymentMethods = paymentMethods.filter({ $0.type != .card })
+        let onlyCards = nonCardPaymentMethods.isEmpty
```
wooj-stripe, line 74: I typically like to name booleans in a way that answers a yes or no question.. e.g.(`is...`, `has...`, etc) -- e.g. hasOnlyCards

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/Vertical Saved Payment Method Screen/VerticalSavedPaymentMethodsViewController.swift
```diff
@@ -70,12 +70,15 @@ class VerticalSavedPaymentMethodsViewController: UIViewController {
     }
 
     private var headerText: String {
+        let nonCardPaymentMethods = paymentMethods.filter({ $0.type != .card })
+        let hasOnlyCards = nonCardPaymentMethods.isEmpty
```
porter-stripe, line 74: Nit: Could be `let hasOnlyCards = paymentMethods.filter({ $0.type != .card }).isEmpty`

porter-stripe (approved):

PR #4457, opened 2025-01-11
Use mobile signup endpoint for verified FC flows

mats-stripe:
## Summary

This updates our Link account sign up API to call the new `/consumers/mobile/sign_up` endpoint during verified flows. In these verified flows, we also attach attestation parameters to the API request. Here are the changes made to the consumers of this API:

- `LinkLogin` (Instant Debits): Call the new mobile sign up endpoint (`/consumers/mobile/sign_up`) during verified flows, and continue to call the non-mobile sign up endpoint (`/consumers/accounts/sign_up`) for non-verified flows.
- `NetworkingLinkSignup` (Networking): Call the new mobile sign up endpoint (`/consumers/mobile/sign_up`) during verified flows, passing the app attestation token, then call the `save_accounts_to_link` API (with no email / phone number). Continue to just call `save_accounts_to_link` in non-verified flows.

## Motivation

https://docs.google.com/document/d/1joKz5UZHLVazmecfMHbq6gB6n4wj5u8To6AtqYgq_tc/edit?usp=sharing

## Testing

Hard to test the _success_ case (AFAIK, only real devices can use App Attest), but tested that, when hardcoding a `verified = true` state, the app attestation parameters are being based to the signup endpoint:

Screenshot 2025-01-13 at 10 46 55 AM

And the request fails with the expected error: 

Screenshot 2025-01-13 at 10 47 40 AM

## Changelog

N/a

### DIFF
--- a/StripeCore/StripeCore/Source/Attestation/StripeAttest.swift
+++ b/StripeCore/StripeCore/Source/Attestation/StripeAttest.swift
@@ -82,6 +82,15 @@ import UIKit
         return appAttestService.isSupported
     }
 
+    @_spi(STP) public static func isLinkAssertionError(error: Error) -> Bool {
+        if let error = error as? StripeCore.StripeError,
+           case let .apiError(apiError) = error,
+           apiError.code == "link_failed_to_attest_request" {
+            return true
+        }
+        return false
+    }
+
     // MARK: Public structs
 
     /// Contains the signed data and various information used to sign the request.
--- a/StripeCore/StripeCore/Source/Attestation/StripeAttest.swift
+++ b/StripeCore/StripeCore/Source/Attestation/StripeAttest.swift
@@ -403,7 +403,7 @@ import UIKit
     // MARK: Assertion concurrency
 
     // Called when an assertion handle is completed or times out
-    private func assertionCompleted() {
+    @_spi(STP) public func assertionCompleted() {
         assertionInProgress = false
 
         // Resume the next waiter if there is one
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
@@ -38,11 +38,14 @@ final class FinancialConnectionsAPIClient {
         return consumerPublishableKey
     }
 
-    /// Forwards attestation errors to the `StripeAttest` client for logging.
-    func reportAttestationErrorIfNeeded(error: Error) {
-        guard StripeAttest.isLinkAssertionError(error: error) else { return }
+    /// Marks the assertion as completed and forwards attestation errors to the `StripeAttest` client for logging.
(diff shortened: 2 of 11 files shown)


StripeCore/StripeCore/Source/Attestation/StripeAttest.swift
```diff
@@ -82,6 +82,15 @@ import UIKit
         return appAttestService.isSupported
     }
 
+    @_spi(STP) public static func isLinkAssertionError(error: Error) -> Bool {
```
mats-stripe, line 85: Moved this from `StripePaymentSheet` -> `StripeCore` so that it can be reused in Financial Connections. I also thought it made sense to keep it as part of the `StripeAttest` namespace

StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
```diff
@@ -89,9 +97,32 @@ final class FinancialConnectionsAPIClient {
...
     }
+
+    private func applyAttestationParameters(
+        to baseParameters: [String: Any]
+    ) -> Future<[String: Any]> {
+        let promise = Promise<[String: Any]>()
+        Task {
+            do {
+                let attest = backingAPIClient.stripeAttest
+                let handle = try await attest.assert()
```
mats-stripe, line 108: Hmm, I'm realizing now that this `handle` needs to be marked as `complete()` once the request completes. This might be tricky with this Futures / Promises based API client - I'll try to come up with something

mats-stripe: Done in https://github.com/stripe/stripe-ios/pull/4457/commits/c4e592ded8519f67c44d2be1c8c5384c7681158e

StripeCore/StripeCore/Source/Attestation/StripeAttest.swift
```diff
@@ -394,7 +403,7 @@ import UIKit
     // MARK: Assertion concurrency
 
     // Called when an assertion handle is completed or times out
-    private func assertionCompleted() {
+    @_spi(STP) public func assertionCompleted() {
```
mats-stripe, line 406: Marking an assertion as completed was previously done by holding onto the assertion "handle" and calling its [`complete()` API](https://github.com/stripe/stripe-ios/blob/05f7f74f1af226fbd0fc51196927b2e15421c1f5/StripeCore/StripeCore/Source/Attestation/StripeAttest.swift#L426), but it's pretty tricky to do that using a Futures / Promises based API client. Making this method internally public makes that much simpler

davidme-stripe: Sounds good — just make sure the calls to it are balanced so it doesn't hang. I was thinking about making `complete()` in the handle keep a reference to the individual task for debugging purposes, but it sounds like that wouldn't work here.

davidme-stripe (approved):

PR #4460, opened 2025-01-13
Fix incorrect Panther behavior in Embedded MPE

tillh-stripe:
## Summary

This pull request fixes an issue where the Embedded MPE would show the incorrect last4 for Instant Bank Payments in passthrough mode.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
@@ -284,8 +284,8 @@ extension EmbeddedPaymentElement.PaymentOptionDisplayData {
             label = String.Localized.apple_pay
             paymentMethodType = "apple_pay"
             billingDetails = nil
-        case .saved(let paymentMethod, _):
-            label = paymentMethod.paymentSheetLabel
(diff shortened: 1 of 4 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -281,8 +281,8 @@ extension EmbeddedPaymentElement.PaymentOptionDisplayData {
             label = String.Localized.apple_pay
             paymentMethodType = "apple_pay"
             billingDetails = nil
-        case .saved(let paymentMethod, _):
-            label = paymentMethod.paymentSheetLabel
+        case .saved(let paymentMethod, let confirmParams):
+            label = paymentMethod.paymentOptionLabel(confirmParams: confirmParams)
```
yuki-stripe, line 285: What's the difference between these two? It'd be great if we could only have one, unless they have clear distinctions

yuki-stripe: Ahh sorry this feels hacky b/c `STPPaymentMethod` represents the Stripe API PaymentMethod object. I prefer the original proposal. 

I don't have a good idea for how to fix my discomfort with `paymentSheetLabel` and `paymentOptionLabel`, so let's ship as-is for now.

PR #4473, opened 2025-01-15
Break testAllIntegrationMethods into a few examples

davidme-stripe:
## Summary
This is pretty flaky, and most of these are tested in the PaymentSheet LPM confirm tests. Let's just test a few to make sure the complex UI stuff works (e.g. switching out to the Safari app then getting the `UIApplication.didBecomeActiveNotification` when we switch back).

## Motivation
Reducing CI flake

## Testing
CI

### DIFF
--- a/Testers/IntegrationTester/IntegrationTesterUITests/IntegrationTesterUITests.swift
+++ b/Testers/IntegrationTester/IntegrationTesterUITests/IntegrationTesterUITests.swift
@@ -407,7 +407,8 @@ class IntegrationTesterUITests: XCTestCase {
         if integrationMethod == .paypal {
             // PayPal uses ASWebAuthenticationSession, tap continue:
             let springboard = XCUIApplication(bundleIdentifier: "com.apple.springboard")
-            XCTAssertTrue(springboard.waitForExistence(timeout: 10.0))
+            let continueButton = springboard.buttons["Continue"]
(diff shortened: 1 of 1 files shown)


Testers/IntegrationTester/IntegrationTesterUITests/IntegrationTesterUITests.swift
```diff
@@ -459,7 +445,7 @@ class IntegrationTesterUITests: XCTestCase {
         completeAuth.forceTapElement()
 
         let safariOpenButton = safari.buttons["Open"]
-        XCTAssertTrue(safariOpenButton.waitForExistence(timeout: 5.0))
+        XCTAssertTrue(safariOpenButton.waitForExistence(timeout: 30.0))
```
porter-stripe, line 448: Why so long??

porter-stripe (approved):

Testers/IntegrationTester/IntegrationTesterUITests/IntegrationTesterUITests.swift
davidme-stripe: This is the time between hitting "authorize", waiting for the PI to authorize, getting sent to the redirect URL, and getting the "Open in PaymentSheet Example?" popup from Safari. Combination of slow network, slowness on our side in testmode, and slow machines running the tests. It shouldn't take 30 seconds, but it definitely takes more than 5.

PR #4475, opened 2025-01-16
Hide non-card and non-US bank pms on CustomerSheet,

joyceqin-stripe:
## Summary

On CustomerSheet, when opted in to the set as default feature, only show cards and US bank accounts
## Motivation

[MOBILESDK-2700](https://jira.corp.stripe.com/browse/MOBILESDK-2700)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSavedPaymentMethodsCollectionViewController.swift
```diff
@@ -241,7 +242,14 @@ class CustomerSavedPaymentMethodsCollectionViewController: UIViewController {
...
     ) {
-        self.savedPaymentMethods = savedPaymentMethods
+        if configuration.allowsSetAsDefaultPM {
+            self.savedPaymentMethods = savedPaymentMethods.filter{ savedPaymentMethod in CustomerSheet.supportedDefaultPaymentMethods.contains{paymentMethodType in
+                savedPaymentMethod.type == paymentMethodType}
+            }
+        }
+        else {
+            self.savedPaymentMethods = savedPaymentMethods
+        }
```
porter-stripe, line 252: Can we add a comment about why we're doing this? Also, do we plan to "undo" this ever in the future?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSavedPaymentMethodsViewController.swift
```diff
@@ -63,8 +63,13 @@ class CustomerSavedPaymentMethodsViewController: UIViewController {
     }()
     private var cachedClientSecret: String?
 
+    var showApplePay: Bool {
```
porter-stripe, line 66: Can we update the CustomerSheet playground with a toggle to set the allows default flag for easier testing?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+PaymentMethodAvailability.swift
```diff
@@ -41,6 +41,12 @@ extension PaymentSheet {
         .multibanco,
     ]
 
+    /// An unordered list of paymentMethod types that can be saved in PaymentSheet
+    @_spi(STP) public static let supportedSavedPaymentMethods: [STPPaymentMethodType] = [.card, .USBankAccount, .SEPADebit]
```
porter-stripe, line 45: "unordered", how is this unordered?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+PaymentMethodAvailability.swift
```diff
@@ -41,6 +41,12 @@ extension PaymentSheet {
...
+
+    /// An unordered list of paymentMethod types that can be set as default in PaymentSheet when opted in to the "save as default" feature
```
porter-stripe, line 47: Same on the unordered bit, also maybe "paymentMethod types" -> `STPPaymentMethodType`.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+PaymentMethodAvailability.swift
porter-stripe: In PaymentSheetLoader we have "savedPaymentMethodTypes", can we consolidate these to have a single source of truth?

porter-stripe: Also do these need to be public? I don't think they are used outside the PaymentSheet module.

Wmc05061970:
What is this

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSavedPaymentMethodsViewController.swift
joyceqin-stripe: There's already a toggle, but you have to use customer session for it to appear

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSavedPaymentMethodsCollectionViewController.swift
joyceqin-stripe: Undo in what way? When opted in to the set as default feature, I don't think we ever want to show payment methods on customer sheet that don't support being set as default.

porter-stripe: Undo as-in support setting wallets as defaults: https://stripe.slack.com/archives/C07RF1UM08M/p1737151131938219

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+PaymentMethodAvailability.swift
joyceqin-stripe: It's unordered because it's not specifically placing priority on any payment methods, it's just a collection of payment methods that can be saved

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSavedPaymentMethodsCollectionViewController.swift
joyceqin-stripe: Well the case of apple pay isn't handled in savedPaymentMethods since apple pay isn't an STPPaymentMethod— apple pay is added in the viewModels based on the configuration's value for showApplePay

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+PaymentMethodAvailability.swift
joyceqin-stripe: And it says only public and open declarations can be @_spi

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSavedPaymentMethodsCollectionViewController.swift
porter-stripe: Ah, I meant to ask about SEPA. Will we ever support setting that as default in CustomerSheet?

joyceqin-stripe: Possibly— we're not planning on doing that this project because web doesn't support it, but unless someone requests setting SEPA as default I don't think it'll happen for some time

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+PaymentMethodAvailability.swift
porter-stripe: Yep, doesn't need to be spi.

porter-stripe (approved):

PR #4477, opened 2025-01-17
Refactor embedded view selection flow

yuki-stripe:
## Summary
Refactored the `updateSelectionState` into two separate, smaller methods with single responsibilities: 
- `embeddedPaymentMethodsViewDidUpdateSelection`, called whenever the selection changes
- `embeddedPaymentMethodsViewDidTapPaymentMethodRow`, called whenever a row is tapped

and fixed some bugs from the old implementation.

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-2974

The old `updateSelectionState` code had some bugs. It should work like this: 
- If the customer taps the card row, we should present the form, send a "tapped" analytic, etc.
- If we reset the selection back to the card row, we shouldn't do those things.

but instead it:
- Always try to present the form
- Always send a "tapped" analytic if the selection changed

It was also doing too many different things.

## Testing
See expanded view tests.

The old delegate method implementation should probably have been unit tested given its complexity, but IMO they don't need to be now that they're so simple and small.

## Changelog
See changelog

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
+++ b/Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
@@ -377,7 +377,7 @@ class EmbeddedUITests: PaymentSheetUITestCase {
         XCTAssertFalse(app.textViews["By continuing, you agree to authorize payments pursuant to these terms."].waitForExistence(timeout: 3.0))
         let events = analyticsLog.compactMap({ $0[string: "event"] })
             .filter({ !$0.starts(with: "luxe") })
-            .suffix(7)
+            .suffix(5)
 
         XCTAssertEqual(
             events,
@@ -452,36 +452,36 @@ class EmbeddedUITests: PaymentSheetUITestCase {
         payWithApplePay()
         XCTAssertTrue(app.staticTexts["Success!"].waitForExistence(timeout: 10))
     }
-    
+
     func testSelection_savedPaymentMethod() {
         var settings = PaymentSheetTestPlaygroundSettings.defaultValues()
         settings.uiStyle = .embedded
         settings.customerMode = .returning
         settings.applePayEnabled = .on
-        
+
         loadPlayground(app, settings)
         app.buttons["Present embedded payment element"].waitForExistenceAndTap()
-        
+
         // Select Cash App Pay
         app.buttons["Cash App Pay"].waitForExistenceAndTap()
         XCTAssertTrue(app.staticTexts["Cash App Pay"].waitForExistence(timeout: 10))
         XCTAssertTrue(app.buttons["Checkout"].isEnabled)
-        
+
         // Select a saved card
         XCTAssertTrue(app.buttons["View more"].waitForExistenceAndTap())
         XCTAssertTrue(app.staticTexts["Select payment method"].waitForExistence(timeout: 10))
         XCTAssertTrue(app.buttons["•••• 4242"].firstMatch.waitForExistenceAndTap())
-        
+
         // Verify we have dismissed the saved payment method view and have the correct card selected
         XCTAssertFalse(app.staticTexts["Select payment method"].waitForExistence(timeout: 2))
         XCTAssertTrue(app.staticTexts["•••• 4242"].waitForExistence(timeout: 10))
         XCTAssertTrue(app.buttons["Checkout"].isEnabled)
-        
+
         // Select Cash App Pay again
         app.buttons["Cash App Pay"].waitForExistenceAndTap()
         XCTAssertTrue(app.staticTexts["Cash App Pay"].waitForExistence(timeout: 10))
         XCTAssertTrue(app.buttons["Checkout"].isEnabled)
-        
+
         // Delete one payment method so we only have one left, we should not auto select the last remaining saved PM
         XCTAssertTrue(app.buttons["View more"].waitForExistenceAndTap())
         XCTAssertTrue(app.buttons["Edit"].waitForExistenceAndTap())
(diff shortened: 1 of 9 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -323,25 +328,22 @@ extension EmbeddedPaymentElement: EmbeddedFormViewControllerDelegate {
     func embeddedFormViewControllerDidCancel(_ embeddedFormViewController: EmbeddedFormViewController) {
         // If the formViewController was populated with a previous payment option don't reset
         if embeddedFormViewController.previousPaymentOption == nil {
-            self.formViewController = nil
             embeddedPaymentMethodsView.resetSelectionToLastSelection()
         }
         embeddedFormViewController.dismiss(animated: true)
     }
 
-    func embeddedFormViewControllerShouldClose(_ embeddedFormViewController: EmbeddedFormViewController) {
```
yuki-stripe, line 332: This method name implies it would be called when you cancel, since you should close then too, so I renamed it to be clearer

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -92,38 +89,23 @@ extension EmbeddedPaymentElement {
...
-                }
-            }
-        }
 
-        guard case let .new(paymentMethodType) = embeddedPaymentMethodsView.selection else {
-            // This can occur when selection is being reset to nothing selected or to a saved payment method, so don't assert.
-            self.formViewController = nil
-            return
+    /// Helper method to inform delegate only if the payment option changed
+    func informDelegateIfPaymentOptionUpdated() {
```
yuki-stripe, line 98: There's a footgun here where if you call the delegate method directly instead of calling this helper, the helper will break b/c it won't know the correct last updated payment option. Not sure there's a good way to prevent that in code.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentMethodsViewTests.swift
```diff
@@ -44,62 +44,74 @@ final class EmbeddedPaymentMethodsViewTests: XCTestCase {
...
-        XCTAssertFalse(mockDelegate.didCallHeightDidChange, "heightDidChange should not be called on init")
-        XCTAssertFalse(mockDelegate.didCallSelectionDidUpdate, "selectionDidUpdate should not be called on init")
+        XCTAssertEqual(mockDelegate.calls, [])
         XCTAssertNil(embeddedView.selection)
 
         // Simulate tapping Cash App and verify delegate is called
         embeddedView.didTap(selection: .new(paymentMethodType: .stripe(.cashApp)))
-        XCTAssertTrue(mockDelegate.didCallHeightDidChange, "didCallHeightDidChange should be called when selection changes to show a mandate")
-        XCTAssertTrue(mockDelegate.didCallSelectionDidUpdate, "selectionDidUpdate should be called when selection changes")
+        XCTAssertEqual(mockDelegate.calls, [.didUpdateHeight, .didUpdateSelection, .didTapPaymentMethodRow])
```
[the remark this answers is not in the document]
porter-stripe: I like this!

Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
```diff
@@ -382,8 +382,8 @@ class EmbeddedUITests: PaymentSheetUITestCase {
         XCTAssertEqual(
             events,
             ["mc_embedded_paymentoption_savedpm_select",
-             "mc_carousel_payment_method_tapped", "mc_open_edit_screen", "mc_embedded_paymentoption_removed",
-             "mc_carousel_payment_method_tapped", "mc_open_edit_screen", "mc_embedded_paymentoption_removed",
+             "mc_open_edit_screen", "mc_embedded_paymentoption_removed",
+             "mc_open_edit_screen", "mc_embedded_paymentoption_removed",
```
yuki-stripe, line 386: There's no payment methods being tapped here! We were incorrectly sending the 'tapped' analytic when you delete a saved PM and the selected saved PM changes. 

That doesn't match vertical mode behavior but also doesn't match the name of the analytic since there's no tap happening.

cc @wooj-stripe

porter-stripe (approved):
nice, ty!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -323,25 +328,22 @@ extension EmbeddedPaymentElement: EmbeddedFormViewControllerDelegate {
     func embeddedFormViewControllerDidCancel(_ embeddedFormViewController: EmbeddedFormViewController) {
         // If the formViewController was populated with a previous payment option don't reset
         if embeddedFormViewController.previousPaymentOption == nil {
-            self.formViewController = nil
```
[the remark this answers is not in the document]
yuki-stripe: We definitely need to make sure the form VC is up to date!

`embeddedPaymentMethodsViewDidUpdateSelection` is called whenever the view updates its selection and ensures the form VC matches the selection. https://github.com/stripe/stripe-ios/pull/4477/files/18741b63730895f925c2d26cf526889372c1cbf5#diff-b66bb3ee17a6e8b6b2c784056b390b3ee60b34d788c38ba8aac66ef1ded1558dR136-R143

PR #4481, opened 2025-01-20
Add attestation event logging to FC API clients

mats-stripe:
## Summary

Adds logging for various attestation-related events in our Financial Connections API clients (both legacy and async clients). Most of the event logging logic has been consolidated in `FinancialConnectionsAPIClientLogger`. As part of this, I also had to make a few secondary changes to the API clients;

1. Only apply attestation-related parameters in the _initial_ `/synchronize` call.
2. Pass the corresponding `pane` and `api` through the consumer session lookup and sign up APIs for logging purposes.

## Motivation

https://docs.google.com/document/d/1joKz5UZHLVazmecfMHbq6gB6n4wj5u8To6AtqYgq_tc/edit?tab=t.0#bookmark=kix.oi6g969wwrz0

## Testing

N/a

## Changelog

N/a

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClientLogger.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClientLogger.swift
@@ -6,26 +6,28 @@
 //
 
 import Foundation
+@_spi(STP) import StripeCore
 
 struct FinancialConnectionsAPIClientLogger {
     private var analyticsClient = FinancialConnectionsAnalyticsClient()
 
+    enum API: String {
+        case consumerSessionLookup = "consumer_session_lookup"
+        case linkSignUp = "link_sign_up"
+    }
+
     enum Event {
-        /// When checking if generating attestation is supported succeeds.
-        case attestationInitSucceeded
         /// When checking if generating attestation is supported does not succeed.
         case attestationInitFailed
         /// When an attestation token gets generated successfully.
-        case attestationRequestTokenSucceeded
+        case attestationRequestTokenSucceeded(API)
         /// When a token generation attempt fails client-side.
-        case attestationRequestTokenFailed
+        case attestationRequestTokenFailed(API, Error)
         /// When an attestation verdict fails backend side and we get an attestation related error.
-        case attestationVerdictFailed
+        case attestationVerdictFailed(API)
 
         var name: String {
             switch self {
-            case .attestationInitSucceeded:
-                return "attestation.init_succeeded"
             case .attestationInitFailed:
                 return "attestation.init_failed"
             case .attestationRequestTokenSucceeded:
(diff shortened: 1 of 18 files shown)


StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClientLogger.swift
```diff
@@ -0,0 +1,65 @@
...
+//
+
+import Foundation
+
+struct FinancialConnectionsAPIClientLogger {
+    private var analyticsClient = FinancialConnectionsAnalyticsClient()
+
+    enum Event {
+        /// When checking if generating attestation is supported succeeds.
+        case attestationInitSucceeded
```
mats-stripe, line 15: @carlosmuvi-stripe I added this event for now, but we can drop it if we think it will be too noisy.

StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClientLogger.swift
```diff
@@ -0,0 +1,65 @@
...
+            case .attestationRequestTokenSucceeded:
+                return "attestation.request_token_succeeded"
+            case .attestationRequestTokenFailed:
+                return "attestation.request_token_failed"
+            case .attestationVerdictFailed:
+                return "attestation.verdict_failed"
+            }
+        }
+
+        var parameters: [String: Any] {
```
mats-stripe, line 40: @carlosmuvi-stripe let me know which other parameters we should include for each event

StripeFinancialConnections/StripeFinancialConnections/Source/Analytics/FinancialConnectionsAnalyticsClient.swift
```diff
@@ -174,6 +174,7 @@ extension FinancialConnectionsAnalyticsClient {
         additionalParameters["single_account"] = manifest.singleAccount
         additionalParameters["allow_manual_entry"] = manifest.allowManualEntry
         additionalParameters["account_holder_id"] = manifest.accountholderToken
+        additionalParameters["app_verification_enabled"] = manifest.appVerificationEnabled
```
mats-stripe, line 177: Includes `manifest.appVerificationEnabled` as a common field, passed along to most logs we emit.

StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClientLogger.swift
mats-stripe: Dropped it in the latest commit

StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClientLogger.swift
carlosmuvi-stripe: I think we should be good with these + the default parameters sent on all requests

carlosmuvi-stripe (approved):

tillh-stripe (approved):

PR #4488, opened 2025-01-22
Make minor tweaks to IBP incentives flow

tillh-stripe:
## Summary

This pull request fixes some issues in the IBP incentives flow:
1. We simplified `LinkConsumerIncentive` by removing the custom parsing, which is no longer needed.
2. We now pass the `incentiveEligible` field to the bank account view after the flow.

This pull request previously included other changes, but [a backend change](https://git.corp.stripe.com/stripe-internal/pay-server/pull/984495) made them unnecessary.

### DIFF
--- a/StripeCore/StripeCore/Source/Connections Bindings/ElementsSessionContext.swift
+++ b/StripeCore/StripeCore/Source/Connections Bindings/ElementsSessionContext.swift
@@ -10,13 +10,20 @@ import Foundation
 /// Contains elements session context useful for the Financial Connections SDK.
 @_spi(STP) public struct ElementsSessionContext {
     @_spi(STP) @frozen public enum IntentID {
-        case payment(String)
-        case setup(String)
-        case deferred(String)
+        case payment(id: String, elementsSessionID: String)
+        case setup(id: String, elementsSessionID: String)
+        case deferred(id: String)
 
         @_spi(STP) public var id: String {
             switch self {
-            case let .payment(id), let .setup(id), let .deferred(id):
+            case let .payment(id, _), let .setup(id, _), let .deferred(id):
+                return id
+            }
+        }
+        
+        @_spi(STP) public var elementsSessionID: String {
+            switch self {
+            case let .payment(_, id), let .setup(_, id), let .deferred(id):
                 return id
             }
         }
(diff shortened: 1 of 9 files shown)


StripeCore/StripeCore/Source/Connections Bindings/ElementsSessionContext.swift
```diff
@@ -10,13 +10,20 @@ import Foundation
 /// Contains elements session context useful for the Financial Connections SDK.
 @_spi(STP) public struct ElementsSessionContext {
     @_spi(STP) @frozen public enum IntentID {
-        case payment(String)
-        case setup(String)
-        case deferred(String)
+        case payment(id: String, elementsSessionID: String)
+        case setup(id: String, elementsSessionID: String)
+        case deferred(id: String)
```
mats-stripe, line 15: Does `id` for the deferred case represents both the intent ID and the elements session ID? Could you clarify why it's being returned for both properties below?

StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPIClient.swift
```diff
@@ -1215,7 +1217,7 @@ extension FinancialConnectionsAPIClient: FinancialConnectionsAPI {
         return post(
             resource: APIEndpointAvailableIncentives,
             parameters: parameters,
-            useConsumerPublishableKeyIfNeeded: true
+            useConsumerPublishableKeyIfNeeded: false
```
mats-stripe, line 1220: This change will also need to be made in the async API client (sorry, we have to maintain both right now 😅). It would change here: 

https://github.com/stripe/stripe-ios/blob/659309da5a1f254a40c4e34cd5cd4bb462ec3140/StripeFinancialConnections/StripeFinancialConnections/Source/API%20Bindings/FinancialConnectionsAsyncAPIClient.swift#L1098-L1114

mats-stripe (dismissed):
LGTM, pending a small change to the async API client, and a question about the IDs

StripeCore/StripeCore/Source/Connections Bindings/ElementsSessionContext.swift
tillh-stripe: We no longer require `elementsSessionID`, so I undid that change.

mats-stripe (approved):
😚👌

PR #4489, opened 2025-01-22
Show blocked brands in CBC dropdown, but block selection

porter-stripe:
## Summary
- Instead of just hiding disallowed card brands in the CBC dropdown, we should instead show them as disabled and prevent them from being selected. When the user attempts to scroll to a disallowed brand it is "rolled back" to the previous selection.

### Update card flow

https://github.com/user-attachments/assets/de5547b8-c95a-4e0c-a0cb-c96b439caf8d

### New card flow

https://github.com/user-attachments/assets/8acf80e5-1936-407f-a402-5b1d0e50137b

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-3035

## Testing
- New unit test
- Manual

## Changelog
N/A (private beta for now)


porter-stripe:
Added the "hip without translations" b/c:
1. This feature is in private beta
2. None of the merchants using this feature use CBC

By the time we GA this feature translations will have come in.

StripeUICore/StripeUICoreTests/Unit/Elements/DropdownFieldElementTest.swift
```diff
@@ -93,4 +93,28 @@ final class DropdownFieldElementTest: XCTestCase {
...
     }
+    
+    func testCantSelectDisabledItem() {
+        let disabledItem = DropdownFieldElement.DropdownItem(pickerDisplayName: "Disabled",
+                                                             labelDisplayName: "Disabled",
+                                                             accessibilityValue: "Disabled",
+                                                             rawData: "Disabled",
+                                                             isDisabled: true)
+        let itemsWithDisabled = items + [disabledItem]
+
```
wooj-stripe, line 104: Seems like we should assert the number of items before directly accessing specific indexes.

StripeUICore/StripeUICore/Source/Elements/DropdownFieldElement.swift
```diff
@@ -265,6 +271,13 @@ extension DropdownFieldElement {
     }
 
     public func pickerView(_ pickerView: UIPickerView, didSelectRow row: Int, inComponent component: Int) {
+        let item = items[row]
```
wooj-stripe, line 274: Seems like we should add to the top of the fn:
```
        guard row < items.count else {
            stpAssertionFailure("...")
            return
        }
```

StripePaymentsUI/StripePaymentsUI/Source/Internal/UI/Elements/DropDownFieldElement+CardBrand.swift
```diff
@@ -14,14 +14,15 @@ import UIKit
 extension DropdownFieldElement {
 
     @_spi(STP) public static func makeCardBrandDropdown(cardBrands: Set<STPCardBrand> = Set<STPCardBrand>(),
+                                                        disallowedCardBrands: Set<STPCardBrand> = Set<STPCardBrand>(),
```
wooj-stripe, line 17: Is there a strong reason to have a default?  Seems like something each caller should be required to pass in.

StripePaymentsUI/StripePaymentsUI/Source/Internal/Categories/STPCardBrand+PaymentsUI.swift
```diff
@@ -26,17 +27,21 @@ extension STPCardBrand {
         return NSAttributedString(attachment: brandImageAttachment)
     }
 
-    func cardBrandItem(theme: ElementsAppearance = .default, maxWidth: CGFloat? = nil) -> DropdownFieldElement.DropdownItem {
+    func cardBrandItem(theme: ElementsAppearance = .default, isAllowed: Bool, maxWidth: CGFloat? = nil) -> DropdownFieldElement.DropdownItem {
```
wooj-stripe, line 30: Would be good to keep a consistent nomenclature  of `disallowed` instead of introducing `isAllowed`

StripePaymentsUI/StripePaymentsUI/Source/Internal/UI/Elements/DropDownFieldElement+CardBrand.swift
porter-stripe: We do have the "cardBrands" parameter with a default value too, inside of CardSectionElement we initialize the drop down before we have fetched brands if the merchant is eligible for CBC, so it's nice for convince.

StripePaymentsUI/StripePaymentsUI/Source/Internal/Categories/STPCardBrand+PaymentsUI.swift
porter-stripe: Yeah agree!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/CardSection/CardSectionElement.swift
```diff
@@ -245,19 +245,30 @@ final class CardSectionElement: ContainerElement {
...
-                } else if cardBrands.count == 1 && self.cardBrandFilter != .default {
-                    // If we only fetched one card brand auto select it, 1 index due to 0 index being the placeholder.
+                } else if validBrandSelections.count == 1,
+                          !disallowedCardBrands.isEmpty,
+                          let firstItem = validBrandSelections.first,
+                          let indexToSelect = cardBrandDropDown.items.firstIndex(where: { $0.rawData == firstItem.rawData }) {
+                    // If we only fetched one card brand that is not disallowed, auto select it.
                     // This case typically only occurs when card brand filtering is used with CBC and one of the fetched brands is filtered out.
-                    cardBrandDropDown.select(index: 1, shouldAutoAdvance: false)
+                    cardBrandDropDown.select(index: indexToSelect, shouldAutoAdvance: false)
```
wooj-stripe, line 271: I'm having a tough time following this logic, but let me know what you think about breaking up this logic in the following way:

 ```
    ....
                cardBrandDropDown.update(items: DropdownFieldElement.items(...)
    ....
                // Prioritize merchant preference if we did not have brands prior to calling .possibleBrands, otherwise use default logic
                if !hadBrands, let indexToSelect = hasPreferredBrandIndex(fetchedCardBrands: fetchedCardBrands, disallowedCardBrands: disallowedCardBrands, cardBrandDropDown: cardBrandDropDown) {
                    cardBrandDropDown.select(index: indexToSelect, shouldAutoAdvance: false)
                } else if let indexToSelect = useDefaultSelectionLogic(disallowedCardBrands: disallowedCardBrands, cardBrandDropDown: cardBrandDropDown) {
                    cardBrandDropDown.select(index: indexToSelect, shouldAutoAdvance: false)
                }
 ```

Then we take the logic out of those ifstatements into two separate functions:
```
        // Select the first brand in the fetched brands that appears earliest in the merchants preferred networks
        func hasPreferredBrandIndex(fetchedCardBrands: Set<STPCardBrand>, disallowedCardBrands: Set<STPCardBrand>, cardBrandDropDown: DropdownFieldElement) -> Int? {
            guard let preferredNetworks = self.preferredNetworks,
                  let brandToSelect = preferredNetworks.first(where: { fetchedCardBrands.contains($0) && !disallowedCardBrands.contains($0) }),
                  let indexToSelect = cardBrandDropDown.items.firstIndex(where: { $0.rawData == STPCardBrandUtilities.apiValue(from: brandToSelect) }) else {
                      return nil
                  }
            return indexToSelect

        }

        // If we only fetched one card brand that is not disallowed, auto select it.
        // This case typically only occurs when card brand filtering is used with CBC and one of the fetched brands is filtered out.
        func useDefaultSelectionLogic(disallowedCardBrands: Set<STPCardBrand>, cardBrandDropDown: DropdownFieldElement) -> Int? {
            let validBrandSelections = cardBrandDropDown.items.filter { !$0.isPlaceholder && !$0.isDisabled }
            guard validBrandSelections.count == 1,
                  !disallowedCardBrands.isEmpty,
                  let firstItem = validBrandSelections.first,
                  let indexToSelect = cardBrandDropDown.items.firstIndex(where: { $0.rawData == firstItem.rawData }) else {
                return nil
            }
            return indexToSelect
        }
```

porter-stripe: Ah I like this better!

wooj-stripe (approved):

PR #4491, opened 2025-01-22
Fix embedded deferred intent confirmation type

yuki-stripe:
## Summary
`STPAnalyticsClient.DeferredIntentConfirmationType.none` should only be sent when we complete confirmation using `COMPLETE_WITHOUT_CONFIRMING_INTENT`. 

It's understandably confusing so I also renamed `none` to `completeWithoutConfirmingIntent`.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Analytics/PaymentSheetAnalyticsHelper.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Analytics/PaymentSheetAnalyticsHelper.swift
@@ -389,7 +389,7 @@ extension STPAnalyticsClient {
     enum DeferredIntentConfirmationType: String {
         case server = "server"
         case client = "client"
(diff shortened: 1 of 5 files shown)


StripePaymentSheet/StripePaymentSheet/Source/Analytics/PaymentSheetAnalyticsHelper.swift
```diff
@@ -389,7 +389,8 @@ extension STPAnalyticsClient {
     enum DeferredIntentConfirmationType: String {
         case server = "server"
         case client = "client"
-        case none = "none"
+        /// The merchant backend used the special string instead of a intent client secret, so we completed the payment without confirming an intent.
```
porter-stripe, line 392: Nit: Maybe we mention what this special string is?

porter-stripe (approved):

porter-stripe (approved):

PR #4494, opened 2025-01-22
Refactor embedded confirm to pass authContext & paymentOption

yuki-stripe:
## Summary
I noticed there was some unnecessarily complicated logic/error handling in `_confirm` to find the `authContext` and `paymentOption` when we could either:
1. just pass those things directly from the FormVC, eliminating the possibility of them being nil
2. Guard they're non-nil when the merchant calls`confirm`

## Testing
Relying on existing tests

## Changelog
Not user facing


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetError.swift
```diff
@@ -20,6 +20,7 @@ public enum PaymentSheetError: Error, LocalizedError {
     case unknown(debugDescription: String)
 
     // MARK: Generic errors
+    case integrationError(debugDescription: String)
     case missingClientSecret
     case invalidClientSecret
```
yuki-stripe, line 25: I don't see the value in having specific cases for various integration errors. It's not like anyone would want to switch over or otherwise identify a specific integration error _in code_. I'll discuss reworking PaymentSheetError in our next sync.

porter-stripe: IIRC, we did this instead of having generic "unknown" cases where we pass in a string to be really sure the errors that we log never contain PII.

yuki-stripe: Yeah I recall that too. Does it really help, though? Nothing stops someone from adding a `case invalidClientSecret(valueWhichCouldContainPII: String)` except code review.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetError.swift
```diff
@@ -129,6 +130,8 @@ extension PaymentSheetError: CustomDebugStringConvertible {
                 return message
             case .embeddedPaymentElementAlreadyConfirmedIntent:
                 return "This instance of EmbeddedPaymentElement has already confirmed an intent successfully. Create a new instance of EmbeddedPaymentElement to confirm a new intent."
+            case .integrationError(debugDescription: let debugDescription):
```
yuki-stripe, line 133: I could modify this to `case integrationError(nonPIIDebugDescription)` to assuage FUD, we've done that elsewhere

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetError.swift
porter-stripe: Yeah the benefit is debatable, there's nothing totally stopping someone from doing it, just guardrails trying to prevent the accidental addition of PII in error messages by making devs think more about the contents of the error message.

porter-stripe (approved):
very nice!

PR #4500, opened 2025-01-23
Adds HSBC Web Flow Integration Test

Twigz:
Adds an integration test against the previously failing HSBC web view flow.

## Summary
Adds an integration test against the card number for HSBC to make sure that the flow remains tappable.

I made it specific to HSBC because the actual web view automation relies on the specific elements of the web view, which may (likely will) be different for other web flows.

## Motivation
This [PR](https://github.com/stripe/stripe-ios/pull/4277/files) added functionality fixing an issue where this particular flow was not tappable.

## Testing
This is the test code!


reuben-stripe (approved):
LGTM. Thanks for making this change.

davidme-stripe (approved):

PR #4501, opened 2025-01-23
Set correct allow_redisplay value for spms with link

wooj-stripe:
## Summary
Set allow_redisplay value when signing up for link and saving a new payment method

## Motivation
allow_redisplay was not being set.

## Testing
Updated existing test

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
@@ -453,7 +453,7 @@ extension PaymentSheet {
                         case .success(let paymentDetails):
                             if elementsSession.linkPassthroughModeEnabled {
                                 // If passthrough mode, share payment details
-                                linkAccount.sharePaymentDetails(id: paymentDetails.stripeID, cvc: paymentMethodParams.card?.cvc) { result in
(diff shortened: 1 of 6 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
```diff
@@ -505,7 +510,8 @@ extension PaymentSheet {
                 let shouldSave = false // always false, as we don't show a save-to-merchant checkbox in Link VC
 
                 if elementsSession.linkPassthroughModeEnabled {
-                    linkAccount.sharePaymentDetails(id: paymentDetails.stripeID, cvc: paymentDetails.cvc) { result in
+                    // allowRedisplay is nil since we are not saving a payment method.
+                    linkAccount.sharePaymentDetails(id: paymentDetails.stripeID, cvc: paymentDetails.cvc, allowRedisplay: nil) { result in
```
wooj-stripe, line 514: @davidestes  - a bit unclear here on capturing allowRedisplay.  Is it possible to have a checkbox when given .withPaymentDetails?

davidme-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
wooj-stripe: Chatted offline -- we are not showing a save payment method checkbox in the link flow, so we will never be saving the payment method back to the user, therefore there is no reason to send an allow_redisplay value.

PR #4503, opened 2025-01-24
Show checkbox for sepa based payment methods to set allow_redisplay

wooj-stripe:
## Summary
Adds the consent checkbox for sepa based payment methods when saving and customerSession.features.payment_method_save is set to enabled.

## Motivation
allow_redisplay value is not being set for sepa based payment methods

## Testing
existing tests & will add additional tests.


Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetVerticalUITest.swift
```diff
@@ -30,7 +30,7 @@ class PaymentSheetVerticalUITests: PaymentSheetUITestCase {
         settings.mode = .setup
         settings.customerMode = .new
         settings.currency = .eur
-        settings.customerKeyType = .legacy // TODO: Change to customerSessions when adding save checkbox for SEPA
```
porter-stripe, line 33: I also see a TODO like this in PaymentSheetStandardLPMUIOneTests.testSEPADebitPaymentMethod_PaymentSheet. Should we update that one too?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
```diff
@@ -603,6 +626,26 @@ extension PaymentSheetFormFactory {
...
 
+    // Only show checkbox for PI+SFU & Setup Intent
+    func makeSepaBasedPMCheckbox() -> Element? {
+        let isSaving = BoolReference()
+        let saveCheckbox = makeSaveCheckbox(
+            label: String(
+                format: STPLocalizedString(
+                    "Save this account for future %@ payments",
+                    "Prompt next to checkbox to save bank account."
+                ),
```
porter-stripe, line 637: I see this string and comment hardcoded twice in this file, should we move it to String.Localized?

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetVerticalUITest.swift
wooj-stripe: YES! thank you!!

porter-stripe (dismissed):

porter-stripe (approved):

PR #4507, opened 2025-01-27
Embedded GA metrics

porter-stripe:
## Summary
- Distinguish between embedded in load events
- Add analytics for update API calls
- Track confirm calls

## Motivation
- Embedded GA

## Testing
- Updating existing tests
- New unit tests

## Changelog
N/A

### DIFF
--- a/StripeCore/StripeCore/Source/Analytics/STPAnalyticEvent.swift
+++ b/StripeCore/StripeCore/Source/Analytics/STPAnalyticEvent.swift
@@ -78,11 +78,6 @@ import Foundation
     // MARK: - Embedded Payment Element init
     case mcInitEmbedded = "mc_embedded_init"
     
-    // MARK: - Embedded Payment Element loading
-    case mcLoadStartedEmbedded = "mc_load_started_embedded"
-    case mcLoadSucceededEmbedded = "mc_load_succeeded_embedded"
-    case mcLoadFailedEmbedded = "mc_load_failed_embedded"
-    
     // MARK: - Embedded Payment Element confirm
     case mcConfirmEmbedded = "mc_embedded_confirm"
     
--- a/StripeCore/StripeCore/Source/Analytics/STPAnalyticEvent.swift
+++ b/StripeCore/StripeCore/Source/Analytics/STPAnalyticEvent.swift
@@ -89,6 +89,8 @@ import Foundation
     // MARK: - Embedded Payment Element update
     case mcUpdateStartedEmbedded = "mc_embedded_update_started"
     case mcUpdateFinishedEmbedded = "mc_embedded_update_finished"
+    case mcUpdateFailedEmbedded = "mc_embedded_update_failed"
+    case mcUpdateCanceledEmbedded = "mc_embedded_update_canceled"
     
     // MARK: - PaymentSheet Show
     case mcShowCustomNewPM = "mc_custom_sheet_newpm_show"
--- a/StripeCore/StripeCore/Source/Analytics/STPAnalyticEvent.swift
+++ b/StripeCore/StripeCore/Source/Analytics/STPAnalyticEvent.swift
@@ -84,8 +84,6 @@ import Foundation
     // MARK: - Embedded Payment Element update
     case mcUpdateStartedEmbedded = "mc_embedded_update_started"
     case mcUpdateFinishedEmbedded = "mc_embedded_update_finished"
-    case mcUpdateFailedEmbedded = "mc_embedded_update_failed"
-    case mcUpdateCanceledEmbedded = "mc_embedded_update_canceled"
     
     // MARK: - PaymentSheet Show
     case mcShowCustomNewPM = "mc_custom_sheet_newpm_show"
--- a/StripeCore/StripeCore/Source/Analytics/STPAnalyticEvent.swift
+++ b/StripeCore/StripeCore/Source/Analytics/STPAnalyticEvent.swift
@@ -77,7 +77,19 @@ import Foundation
 
     // MARK: - Embedded Payment Element init
     case mcInitEmbedded = "mc_embedded_init"
-
+    
+    // MARK: - Embedded Payment Element loading
+    case mcLoadStartedEmbedded = "mc_load_started_embedded"
+    case mcLoadSucceededEmbedded = "mc_load_succeeded_embedded"
+    case mcLoadFailedEmbedded = "mc_load_failed_embedded"
+    
+    // MARK: - Embedded Payment Element confirm
+    case mcConfirmEmbedded = "mc_embedded_confirm"
+    
+    // MARK: - Embedded Payment Element update
+    case mcUpdateStartedEmbedded = "mc_embedded_update_started"
+    case mcUpdateFinishedEmbedded = "mc_embedded_update_finished"
+    
     // MARK: - PaymentSheet Show
     case mcShowCustomNewPM = "mc_custom_sheet_newpm_show"
     case mcShowCustomSavedPM = "mc_custom_sheet_savedpm_show"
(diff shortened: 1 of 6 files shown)


StripeCore/StripeCore/Source/Analytics/STPAnalyticEvent.swift
```diff
@@ -77,7 +77,21 @@ import Foundation
 
     // MARK: - Embedded Payment Element init
     case mcInitEmbedded = "mc_embedded_init"
-
+    
+    // MARK: - Embedded Payment Element loading
+    case mcLoadStartedEmbedded = "mc_load_started_embedded"
+    case mcLoadSucceededEmbedded = "mc_load_succeeded_embedded"
+    case mcLoadFailedEmbedded = "mc_load_failed_embedded"
```
yuki-stripe, line 84: Let's reuse `mc_load_started` etc and send a new param that says what type of integration it is (PS vs FC vs Embedded).

StripeCore/StripeCore/Source/Analytics/STPAnalyticEvent.swift
```diff
@@ -77,7 +77,21 @@ import Foundation
...
+    
+    // MARK: - Embedded Payment Element confirm
+    case mcConfirmEmbedded = "mc_embedded_confirm"
+    
+    // MARK: - Embedded Payment Element update
+    case mcUpdateStartedEmbedded = "mc_embedded_update_started"
+    case mcUpdateFinishedEmbedded = "mc_embedded_update_finished"
+    case mcUpdateFailedEmbedded = "mc_embedded_update_failed"
+    case mcUpdateCanceledEmbedded = "mc_embedded_update_canceled"
```
yuki-stripe, line 93: Let's use a single `mc_embedded_update_finished` and send status = canceled, failed, succeeded. Easier to query in hubble.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
```diff
@@ -115,6 +115,12 @@ public final class EmbeddedPaymentElement {
         _ = await latestUpdateTask?.value
         // Start the new update task
         let currentUpdateTask = Task { @MainActor [weak self, configuration, analyticsHelper] in
+            analyticsHelper.logEmbeddedUpdateStarted()
```
yuki-stripe, line 118: Why log here and not at the start of the `update` method?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
```diff
@@ -115,6 +115,12 @@ public final class EmbeddedPaymentElement {
...
+            var updateResult: UpdateResult = .canceled // Default to canceled if unhandled
+            defer {
+                analyticsHelper.logEmbeddedUpdateFinished(result: updateResult)
```
yuki-stripe, line 121: Hm can we not log the result immediately before this method returns?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
```diff
@@ -202,6 +212,7 @@ public final class EmbeddedPaymentElement {
             stpAssertionFailure(errorMessage)
             return .failed(error: PaymentSheetError.integrationError(nonPIIDebugDescription: errorMessage))
         }
+        analyticsHelper.log(event: .mcConfirmEmbedded)
```
yuki-stripe, line 215: Same here, why not log first thing in this method?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
porter-stripe: Good question! I think we should log when the update work actually begins not when it is scheduled. If we log it when it is scheduled (at the top of the function) it makes it harder to reason about the duration metric, since some will account for awaiting the previous update canceling and some won't. If we want consistency of the duration metric across all update events we should only measure the time the actual work associated with the update took, not accounting for any time it spent waiting for a previous update to cancel.

It's also convenient from an implementation standpoint with how we currently handle dates in the analytics helper. If we allow multiple update start events to be fired before the finished event it will overwrite the updateStartDate before the previous finished event could send.

If we fire the start event at the top we could get this
mc_embedded_update_started -> set's updateStartDate to x
mc_embedded_update_started -> set's updateStartDate to y
mc_embedded_update_finished -> reads updateStartDate set to y
mc_embedded_update_finished -> reads updateStartDate set to y

We could get around this by mapping update events and their start dates to a UUID so we know what start date to compute the duration on for all the finished events.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
porter-stripe: No strong reason on this one, I'll move it up.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
porter-stripe: Correct me if I'm wrong, but I believe there is a race condition. By the time you reach `logEmbeddedUpdateFinished(result:)` at the end of the function, a second `update` could have already started its own update and logged a start event. There’s no guarantee about which event comes first unless you serialize the calls within the same Task. Logging right before the method returns can record an “update finished” event out of order, or with an incorrect start date, if another update began in between.

```swift
latestUpdateTask?.cancel()
_ = await latestUpdateTask?.value
// Define task

let updateResult = await currentUpdateTask.value
analyticsHelper.logEmbeddedUpdateFinished(result: updateResult)
// After we await the update task we cannot guarantee this code runs before the next update task starts and logs an update started event, resetting the updateStartDate.
```

As I write this out, maybe we should just map update calls to a UUID and a start date to avoid this, but logging all the update events within the Task seems the safest with the current way we handle dates in the analytics helper.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
yuki-stripe: If we moved this to the start of `update`, duration represents the time it takes for the `update` method to finish. That feels quite simple and straightforward.

The total time the app spends waiting for `update` to finish also seems like a more useful duration metric. Like, why ignore the time spent waiting for previous update calls to finish? That's still time the app spends waiting and something we control (e.g. we could try to cancel in-flight tasks more aggressively).

Moving to the top has the additional benefit of capturing all possible errors.

Re: implementation - what if we keep track of the start time within the `update` method?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
yuki-stripe: Imagine the merchant does the following
1. Call `update`
2. Call `update` again before (1) finishes.
3. Receive Update #1 result
4. Receive Update #2 result

I think you're saying the correct analytics ordering is:
1. update_started
2. update_finished
3. update_started
4. update_finished

That matches the ordering of the Tasks inside the `update` method.

I'm proposing (along with my other comment above) to model this analytic around the entire update method,  not just the inner Task. This would make the ordering match the merchant's POV:
1. update_started
2. update_started
3. update_finished
4. update_finished

The ordering is still deterministic; `update_started` fires when the `update` API is called and `update_finished` fires when it returns.

porter-stripe: Sure we can do that 👍

yuki-stripe (approved):

PR #4512, opened 2025-01-29
Fix forcing height for flat_with_checkmark saved PM row

yuki-stripe:
## Summary
RowButton has a `calculateTallestHeight` method to ensure that row buttons are all the same height.  

This was being used to set the flat_with_checkmark saved PM row height, even though that row shouldn't be the same height as the rest.

## Motivation
- flat_with_checkmark saved PM row "View more" row was rendering smaller than intended
- This code was causing problems when I was modifying it to handle something new related to the accessory view

## Testing
I manually verified the "View more" label was getting squished to below its intrinsic content size before this PR and is fixed after.

## Changelog
Technically user facing but too small to mention


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RightAccessoryButton.swift
```diff
@@ -43,12 +43,7 @@ extension RowButton {
         private var label: UILabel {
             let label = UILabel()
             label.text = accessoryType.text
-            switch accessoryType {
-            case .edit, .viewMoreChevron, .update:
-                label.font = appearance.scaledFont(for: appearance.font.base.medium, style: .footnote, maximumPointSize: 20)
-            case .viewMore:
-                label.font = appearance.scaledFont(for: appearance.font.base.medium, size: 14, maximumPointSize: 20)
```
yuki-stripe, line 50: They should be the same size (they are in the figma), I'm assuming this was an attempt to work around the issue I fixed here

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButton.swift
```diff
@@ -367,13 +353,28 @@ class RowButton: UIView {
...
         }
     }
+
+    func makeSameHeightAsOtherRowButtonsIfNecessary() {
+        // To make all RowButtons the same height, set our height to the tallest variant (a RowButton w/ text and subtext)
+        // Don't do this if we are flat_with_checkmark style and have an accessory view - this row button is allowed to be taller than the rest
+        let displayingRightAccessoryView = !(rightAccessoryView?.isHidden ?? true)
+        guard !(isFlatWithCheckmarkStyle && displayingRightAccessoryView) else {
+            return
+        }
```
porter-stripe, line 363: I find this code hard to reason about what it's doing, maybe we can do:
```swift
let displayingRightAccessoryView = rightAccessoryView?.isHidden == false 

if isFlatWithCheckmarkStyle && displayingRightAccessoryView {
    return
}
```

It's nice the guard enforces that we return, but it makes it a bit tough to grok imo.

porter-stripe (approved):
Good catch on this, minor non blocking comment on readability.

porter-stripe (approved):

PR #4514, opened 2025-01-29
Embedded update form pm row pt 1

yuki-stripe:
## Summary
The big changes here are
1. add a hidden Change accessory button to embedded rows
2. make RowButton.sublabel non-optional and hidden if it's empty

In another PR we'll wire it up so that these two things get toggled appropriately, something like this:

https://github.com/user-attachments/assets/5f749241-9b60-40c1-9043-93bf4dabc920

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-3057

## Testing
No new tests b/c PR should be a no-op in terms of behavior changes. 

## Changelog
Not user facing


Tests/ReferenceImages_64/StripePaymentSheetTests.EmbeddedPaymentMethodsViewSnapshotTests/testEmbeddedPaymentMethodsView_flatRadio_rowHeight@3x.png
porter-stripe, line 1: The font size on "Buy now or pay later with Klarna" got bigger .Is this intentional? It now looks bigger than the Link subtext which didn't change in size.

Tests/ReferenceImages_64/StripePaymentSheetTests.EmbeddedPaymentMethodsViewSnapshotTests/testEmbeddedPaymentMethodsView_floating_rowHeight@3x.png
porter-stripe, line 1: Same w/ the Afterpay label

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButton.swift
```diff
@@ -393,18 +386,6 @@ extension RowButton: UIGestureRecognizerDelegate {
         true
     }
 
-    func gestureRecognizer(_ gestureRecognizer: UIGestureRecognizer, shouldReceive touch: UITouch) -> Bool {
```
porter-stripe, line 396: Hm, why don't we need this anymore? I recall needing it otherwise the accessoryView would not get taps.

Tests/ReferenceImages_64/StripePaymentSheetTests.EmbeddedPaymentMethodsViewSnapshotTests/testEmbeddedPaymentMethodsView_flatRadio_rowHeight@3x.png
yuki-stripe: Yeah I meant to comment before pinging you but had to run to lunch - I can't for the life of me figure out why the screenshots changed here. The frames of the Klarna and Link sublabels are identical in both the tests and in the example app and in the example app there is no visual difference.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButton.swift
yuki-stripe: https://developer.apple.com/documentation/uikit/touches_presses_and_gestures/using_responders_and_the_responder_chain_to_handle_events touch events go in order from the deepest subviews to the parent until something handles it, so there shouldn't ever be a need to do this.  Tapping e.g. "View more" works fine.

porter-stripe: On your branch, when I change to to flat w/ checkmark, the "View more" button is not tappable anymore.

yuki-stripe: Thanks for finding that! I understand we added this now, but this isn't a great solution. The real problem is we were incorrectly setting isUserInteractionEnabled to false on the stackview that contains the "View more" button. When iOS goes down the responder chain with a touch to find which view it belongs to, it skips views (and its subviews) if they have `isUserInteractionEnabled = false`.

There were some other bugs uncovered that I fixed as well. Looks like this now:

https://github.com/user-attachments/assets/685ca1f0-c8f9-4cb0-8175-45bd9b4065df

yuki-stripe: LMK if you want to talk more about the responder chain, it's really mind opening when you grok it.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButton.swift
```diff
@@ -199,7 +189,6 @@ class RowButton: UIView {
 
         for view in [radioButton, imageView, labelsStackView, defaultBadge].compactMap({ $0 }) {
             view.translatesAutoresizingMaskIntoConstraints = false
-            view.isUserInteractionEnabled = false
```
porter-stripe, line 202: Interesting, won't this make the image view, labels, and badge intercept taps that should go to the RowButton?

porter-stripe (approved):
nice!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButton.swift
yuki-stripe: They get the tap first, but they don't do anything with it, so it goes to their parent view. Once you understand the responder chain this will all make sense!

PR #4516, opened 2025-01-30
Disable embedded UI during confirmation

porter-stripe:
## Summary
- During confirmation we should disable the user interaction on the embedded view until we have a result
- Re-enable interaction for failed or canceled cases

## Motivation
https://stripe.slack.com/archives/C058XM768SE/p1738259151501839

## Testing
- Manual

## Changelog
N/A (private beta)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -357,7 +358,8 @@ extension EmbeddedPaymentElement {
                 return (.failed(error: error), nil)
             }
         }
-
+        
+        embeddedPaymentMethodsView.isUserInteractionEnabled = false
```
porter-stripe, line 362: I'm thinking we only need to disable interaction after we have waited for the previous update (if any) since that should handle UI enable/disable for the update call.

yuki-stripe (approved):

PR #4520, opened 2025-01-31
UIViewRepresentable for embedded

porter-stripe:
## Summary
- Adds EmbeddedViewRepresentable which is a SwiftUI view that can display the embedded payment element's view in SwiftUI
- Adds EmbeddedPaymentElementView, which is the public API of the EmbeddedViewRepresentable
- Adds a snapshot test for it

## Motivation
- Embedded SwiftUI

## Testing
- Manual
- New snapshots

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedViewRepresentable.swift
```diff
@@ -0,0 +1,92 @@
...
+            .first(where: \.isKeyWindow)
+        #endif
+    }
+    
+    static var visibleViewController: UIViewController? {
+        current?.rootViewController?.topMostViewController
+    }
+}
+
+extension UIViewController {
```
porter-stripe, line 70: I tested this VC stuff out in a very barebones SwiftUI app and confirmed it works. https://github.com/porter-stripe/swift-ui-embedded/tree/main

Tests/ReferenceImages_64/StripePaymentSheetTests.EmbeddedPaymentMethodsViewSnapshotTests/testEmbeddedPaymentMethodsView_flatRadio_rowHeight@3x.png
porter-stripe, line 1: Not sure why, but this was failing in CI for me but passed locally. I don't see any diff, and don't see how my changes would impact this.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+SwiftUI.swift
```diff
@@ -134,17 +137,43 @@ import Combine
         embeddedPaymentElement?.testHeightChange()
     }
 #endif
+    
+    private func calculateAndPublishHeight() {
+        guard let embeddedPaymentElement else { return }
```
yuki-stripe, line 142: nit: take an `embeddedPaymentElement` and we can remove the need to think (eg wonder how we end up in that situation, if we should assert, etc) about it being nil.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedViewRepresentable.swift
```diff
@@ -0,0 +1,92 @@
...
+    var topMostViewController: UIViewController {
```
yuki-stripe, line 71: We have a couple methods already that do this (search "topmost" in the repo), should we use one of them?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+SwiftUI.swift
```diff
@@ -134,17 +137,43 @@ import Combine
...
+    /// Initializes a new instance of `EmbeddedPaymentElementView`.
+    ///
+    /// - Parameter viewModel: The view model for this payment element view.
+    public init(viewModel: EmbeddedPaymentElementViewModel) {
+        self.viewModel = viewModel
+    }
+
+    public var body: some View {
+        EmbeddedViewRepresentable(viewModel: viewModel)
+            .frame(height: viewModel.height)
```
yuki-stripe, line 177: Curious to understand this better (is this the standard way to deal with UIViews that can change size? what happens if you don't do this?), maybe we can chat live.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedViewRepresentable.swift
```diff
@@ -0,0 +1,92 @@
...
+
+import SwiftUI
+@_spi(STP) import StripeCore
+
+struct EmbeddedViewRepresentable: UIViewRepresentable {
+    @ObservedObject var viewModel: EmbeddedPaymentElementViewModel
+
+    init(viewModel: EmbeddedPaymentElementViewModel) {
+        self.viewModel = viewModel
+        STPAnalyticsClient.sharedClient.addClass(toProductUsageIfNecessary: EmbeddedSwiftUIProduct.self)
```
yuki-stripe, line 16: Why add it here vs the `EmbeddedPaymentElementViewModel` init?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedViewRepresentable.swift
```diff
@@ -0,0 +1,92 @@
...
+    }
+    
+    public func makeUIView(context: Context) -> UIView {
+        let containerView = UIView()
+        containerView.backgroundColor = .clear
+        containerView.layoutMargins = .zero
+
+        guard let embeddedPaymentElement = viewModel.embeddedPaymentElement else { return containerView }
```
yuki-stripe, line 24: Should this error?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedViewRepresentable.swift
porter-stripe: I _very_ slightly prefer to put this here, since it's an internal class, and keep the public EmbeddedPaymentElementViewModel init empty. In an effort to keep our public classes a bit cleaner in case anyone is looking through them, kinda like EmbeddedPaymentElement+Internal motivation.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+SwiftUI.swift
porter-stripe: Ah, nice!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedViewRepresentable.swift
porter-stripe: Yeah, that's a good thought. I think it should! This case should only happen if you show the EmbeddedPaymentElementView in your SwiftUI view before isLoaded is true on the view model. According to the [docs](https://developer.apple.com/documentation/swiftui/uiviewrepresentable/makeuiview(context:)) makeUIView is only ever called once, so this looks safe and appears so from testing in an example integration.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedViewRepresentable.swift
porter-stripe: Yeah, I tried the one in StripeUICore, but it actually doesn't work, it can't find the topmost VC and hits an assert that we do not have a presenting VC. We have a few others in FinancialConnections and one in the PaymentSheet+SwiftUI. All three of these use pretty much the same logic as the StripeUICore one. They don't seem to account for NavigationControllers or TabBarControllers, which may be the issue.

I think it'd be best to consolidate all our logic for this kind of work, but I'm a little worried about changing it for the others in this PR and verifying we didn't break anything because I'm not sure of the context they're being used in. Apparently, they work for their use cases, but makes me question if they actually work given they don't work in our example app. Maybe a run ticket to investigate the others and see if they can be updated safely? WDYT?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedViewRepresentable.swift
```diff
@@ -0,0 +1,92 @@
...
+    public func updateUIView(_ uiView: UIView, context: Context) {
+        // Update the presenting view controller in case it has changed
+        viewModel.embeddedPaymentElement?.presentingViewController = UIWindow.visibleViewController
+    }
+}
+
+// MARK: UIWindow and UIViewController helpers
+
+extension UIWindow {
+    static var current: UIWindow? {
```
yuki-stripe, line 52: This one is also similar to `stp_hackilyFumbleAroundUntilYouFindAKeyWindow`

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedViewRepresentable.swift
porter-stripe: Hmm, I see, I'll move it.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedViewRepresentable.swift
porter-stripe: Ah didn't see that, I tested this out in our example app and in the barebones SwiftUI app and it works!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedViewRepresentable.swift
porter-stripe: Based on quick debugging, it appears that failing to handle the NavigationController case is causing the failure.

```swift
        if let nav = self as? UINavigationController {
            // Use visibleViewController for navigation stacks
            return nav.visibleViewController?.topMostViewController ?? nav
```

I updated the SwiftUICore implementation to handle this case now, I did a quick test of everywhere that calls this function and things appear fine.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+SwiftUI.swift
porter-stripe: If you remove this line SwiftUI thinks our view has a height of zero, so the layout of things is totally busted! SwiftUI doesn't seem to play very nice when it comes to view sizing and AutoLayout. It looks at the intrinsicContentSize of the view to determine the height (if you don't set it explicitly like we're doing). The intrinsicContentSize of our embedded view always has a height of zero, I believe this is b/c we use AutoLayout to handle the height. The width is computed correctly on the intrinsicContentSize but height is always zero.

An alternative approach would be to define the intrinsicContentSize on the container view and give it a non-zero height, we could compute the height the same way we currently do. This works and the view sizes itself correctly, however the tradeoff becomes that there doesn't seem to be a way to animate these height changes smoothly. The view jumps between heights.

Those are the two standard ways I've found, [here's](https://stackoverflow.com/questions/68719533/how-do-i-change-the-height-of-this-uiviewrepresentable) someone proposing what I am for a similar question. What I have in the PR seems to be the most reliable from what I have found, providing smooth animations.

There is one another option [sizeThatFits(_:uiView:context:)](https://developer.apple.com/documentation/swiftui/uiviewrepresentable/sizethatfits(_:uiview:context:)). But requires iOS 16+, and I haven't found it to be super easy to use, mainly due to the "proposed size" being somewhat confusing/not explained well in the docs. We can chat more about it.

porter-stripe: Yeah, intrinsicContentSize seems not viable due to animation limitations. Given that, I think it's between what is proposed in the PR or the iOS 16 approach commented above.

porter-stripe: I think I'm leaning towards the iOS 14 compatible approach, the implementation is easier to understand IMO. Lmk what you think, we can also discuss during co-working hours too.

yuki-stripe: Yep agreed. It's a shame that correctly defining intrinsicContentSize on our embedded UIKit view causes SwiftUI to automatically lay out when it changes (good) but not animate with no way to make it animated.

We should override `intrinsicContentSize` in our embedded view, make it return super.intrinsicContentSize, and add a comment explaining why we can't actually implement it correctly. O/w some enterprising future Stripe might implement it and not realize.

cc @davidme-stripe fyi

porter-stripe: Good idea, I added the comment in both the EmbeddedPaymentMethodsView and EmbeddedPaymentElementContainerView. 

I also confirmed out snapshot tests added in this PR would catch this.

yuki-stripe: Oh woah how does the snapshot test catch it?  I'd expect the final snapshot to be the same, regardless of whether it was animated or not.

porter-stripe: Oh you're probably right, I hardcoded in a bad size of 200x200 to test it, but if it had the actual "correct" size it probably wouldn't catch it!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+SwiftUI.swift
```diff
@@ -50,7 +51,9 @@ import Combine
     // MARK: - Public APIs
 
     /// Creates an empty view model. Call `load` to initialize the `EmbeddedPaymentElementViewModel`
-    public init() {}
+    public init() {
+        STPAnalyticsClient.sharedClient.addClass(toProductUsageIfNecessary: SwiftUIProduct.self)
```
yuki-stripe, line 55: Let's pass self and get rid of SwiftUIProduct?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElementContainerView.swift
```diff
@@ -8,6 +8,14 @@ import UIKit
 
 /// The view that's vended to the merchant, containing the embedded view.  We use this to be able to swap out the embedded view with an animation when `update` is called.
 class EmbeddedPaymentElementContainerView: UIView {
+    
+    /// Return the default size to let Auto Layout manage the height.
+    /// Overriding intrinsicContentSize values force SwiftUI to update layout immediately,
```
yuki-stripe, line 13:
```suggestion
    /// Overriding intrinsicContentSize values and setting `invalidIntrinsicContentSize` forces SwiftUI to update layout immediately,
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
```diff
@@ -25,7 +25,14 @@ protocol EmbeddedPaymentMethodsViewDelegate: AnyObject {
 
 /// The view for an embedded payment element
 class EmbeddedPaymentMethodsView: UIView {
-
+    
+    /// Return the default size to let Auto Layout manage the height.
+    /// Overriding intrinsicContentSize values force SwiftUI to update layout immediately,
```
yuki-stripe, line 30:
```suggestion
    /// Overriding intrinsicContentSize values and setting `invalidIntrinsicContentSize` forces force SwiftUI to update layout immediately,
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+SwiftUI.swift
porter-stripe: 👍 👍

PR #4526, opened 2025-01-31
Refactor VerticalPaymentMethodListSelection -> RowButton.Type

yuki-stripe:
## Summary
This renames VerticalPaymentMethodListSelection to RowButton.Type and adds a `type` property to RowButton.

## Motivation
There was potential value in having RowButton be agnostic to whether it was a saved PM / new PM / Apple Pay / Link, in particular if we ever wanted to move it to StripeCore.  But it's too useful for RowButton to hold onto its type and I don't think it's worse off for it. We can simplify `EmbeddedPaymentMethodsView` a lot with this change.

Vertical mode could also benefit from this change but I'll leave that for another PR.

## Testing
Existing tests, no new behavior

## Changelog
Not user facing


StripeCore/StripeCore/Source/Categories/NSArray+Stripe.swift
```diff
@@ -28,4 +28,11 @@ extension Array where Element: Equatable {
         guard let index = firstIndex(of: element) else { return nil }
         return remove(at: index)
     }
+    
+    @discardableResult public mutating func replace(_ element: Element, with newElement: Element) -> Element? {
```
porter-stripe, line 32: Should be behind spi?

porter-stripe (approved):

PR #4529, opened 2025-01-31
Fix embedded bug where update didnt set form VC delegate and clean up…

yuki-stripe:
… implementation

## Motivation
If you fill out a form (eg card), then update, then tap back into the form, the sheet is stuck open and can't be dismissed. 

## Testing
Manual tests. Added checks/asserts so that tests will fail if this regresses.

## Changelog
Not user facing, bug didn't get released.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedFormViewController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedFormViewController.swift
@@ -155,8 +155,6 @@ class EmbeddedFormViewController: UIViewController {
 
     weak var delegate: EmbeddedFormViewControllerDelegate?
 
-    // MARK: - Initializers
-
     init(configuration: EmbeddedPaymentElement.Configuration,
          intent: Intent,
          elementsSession: STPElementsSession,
@@ -164,7 +162,9 @@ class EmbeddedFormViewController: UIViewController {
          paymentMethodType: PaymentSheet.PaymentMethodType,
          previousPaymentOption: PaymentOption? = nil,
          analyticsHelper: PaymentSheetAnalyticsHelper,
-         formCache: PaymentMethodFormCache = .init()) {
+         formCache: PaymentMethodFormCache = .init(),
+         delegate: EmbeddedFormViewControllerDelegate
+    ) {
         self.intent = intent
         self.elementsSession = elementsSession
         self.shouldUseNewCardNewCardHeader = shouldUseNewCardNewCardHeader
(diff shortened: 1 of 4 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
```diff
@@ -133,54 +133,26 @@ public final class EmbeddedPaymentElement {
...
-                self?.selectedFormViewController = EmbeddedFormViewController(configuration: configuration,
-                                                                intent: loadResult.intent,
-                                                                elementsSession: loadResult.elementsSession,
-                                                                shouldUseNewCardNewCardHeader: loadResult.savedPaymentMethods.first?.type == .card,
-                                                                paymentMethodType: formPaymentMethodType,
-                                                                previousPaymentOption: self?.selectedFormViewController?.selectedPaymentOption,
-                                                                analyticsHelper: analyticsHelper)
-
-            }
-
```
yuki-stripe, line 154: Bug was here - we didn't set delegate!  I made it a required param on the initializer so we can't forget it.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
```diff
@@ -133,54 +133,26 @@ public final class EmbeddedPaymentElement {
...
             )
-
-            // 3. Pre-load image into cache
-            // Call this on a detached Task b/c this synchronously (!) loads the image from network and we don't want to block the main actor
-            let fetchPaymentOption = Task.detached(priority: .userInitiated) {
-                // This has the nasty side effect of synchronously downloading the image (see https://jira.corp.stripe.com/browse/MOBILESDK-2604)
-                // This caches it so that DownloadManager doesn't block the main thread when the merchant tries to access the image
-                return await embeddedPaymentMethodsView.selection?.paymentMethodType?.makeImage(updateHandler: nil)
-            }
-            _ = await fetchPaymentOption.value
```
yuki-stripe, line 171: I realized we don't need to do this; there's still a bad problem where `paymentOption` will block the main thread if the image isn't loaded, but it can't happen in practice b/c either:
1. We're preserving the old paymentOption, in which case the image was already loaded
2. The payment option changed to the default, which is always of a type we have a hardcoded image for (saved PM, apple pay, link).

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedFormViewController.swift
```diff
@@ -250,7 +251,12 @@ class EmbeddedFormViewController: UIViewController {
     }
 
     private func didCancel() {
-        delegate?.embeddedFormViewControllerDidCancel(self)
+        if let delegate {
+            delegate.embeddedFormViewControllerDidCancel(self)
+        } else {
+            stpAssertionFailure()
```
porter-stripe, line 257: Should we add an error message here?

porter-stripe (approved):
Looks good! Just a super nit, up to you.

PR #4534, opened 2025-02-03
Enable bank payments in native Link

tillh-stripe:


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
```diff
@@ -733,3 +743,15 @@ private func isEqual(_ lhs: STPPaymentIntentShippingDetails?, _ rhs: STPPaymentI
 
     return rhs == lhsConverted
 }
+
+private extension ConsumerPaymentDetails {
+    
+    var expectedPaymentMethodTypeForPassthroughMode: String? {
+        switch type {
+        case .card, .unparsable:
```
davidme-stripe, line 751: Just checking, should we not send this for `.card`?

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/STPAPIClient+Link.swift
```diff
@@ -330,7 +335,7 @@ extension STPAPIClient {
         let parameters: [String: Any] = [
             "credentials": ["consumer_session_client_secret": consumerSessionClientSecret],
             "request_surface": "ios_payment_element",
-            "types": ["card"],
+            "types": supportedPaymentDetailsTypes.map(\.rawValue),
```
davidme-stripe, line 338: Why do we use a Set instead of an Array for supportedPaymentDetailsType? It's a little clunkier in Swift, and I'm not sure if the underlying JSON response can really represent a Set. It was causing issues with ordering in our tests: https://github.com/stripe/stripe-ios/pull/4549

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
tillh-stripe: As far as I know, that’s not necessary. Let me double-check!

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/STPAPIClient+Link.swift
tillh-stripe: I’ll change it to a list, now that I’ve learned about the test issues it can cause.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
tillh-stripe: The backend will fall back to `card`, but I added it anyway since that’s what Web is doing 🤷‍♂️

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Components/PaymentMethodPicker/LinkPaymentMethodPicker.swift
```diff
@@ -50,7 +50,12 @@ final class LinkPaymentMethodPicker: UIView {
         }
     }
 
-    var supportedPaymentMethodTypes = Set(ConsumerPaymentDetails.DetailsType.allCases)
+    var supportedPaymentMethodTypes = Set(ConsumerPaymentDetails.DetailsType.allCases) {
+        didSet {
+            // TODO(tillh-stripe) Update this as soon as adding bank accounts is supported
+            addPaymentMethodButton.isHidden = !supportedPaymentMethodTypes.contains(.card)
```
tillh-stripe, line 56: I noticed that pressing `Add payment method` in a bank-only session (`orderedPaymentMethodTypes == [.USBankAccount]`) does nothing, so let’s hide it, @davidme-stripe?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
```diff
@@ -733,3 +744,21 @@ private func isEqual(_ lhs: STPPaymentIntentShippingDetails?, _ rhs: STPPaymentI
...
+    ) -> String? {
+        switch type {
+        case .card:
+            return "card"
+        case .unparsable:
+            return nil
+        case .bankAccount:
+            let canAcceptACH = elementsSession.orderedPaymentMethodTypes.contains(.USBankAccount)
+            let isLinkCardBrand = elementsSession.linkSettings?.linkMode?.isPantherPayment ?? false
+            return isLinkCardBrand && !canAcceptACH ? "card" : "bank_account"
```
tillh-stripe, line 761: This mimics behavior on the Web: If we’re in the Link card brand mode, but the merchant accepts ACH, then we want the payment method to be a full bank account.

cc @abagasra-stripe and @constantine-stripe

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Components/PaymentMethodPicker/LinkPaymentMethodPicker.swift
davidme-stripe: Yep, it's not hooked up yet, I have a JIRA for this somewhere. We should plug it in once all this is working!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
davidme-stripe: Interesting, okay!

davidme-stripe (approved):

PR #4538, opened 2025-02-04
Set 'Change >' and sublabel to selected embedded form rows

yuki-stripe:
## Summary
Adds logic to set and unset the "Change >" + sublabel.

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-3057

## Testing
See new tests

## Changelog
See changelog


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -331,14 +331,63 @@ extension EmbeddedPaymentElement: EmbeddedFormViewControllerDelegate {
...
+            switch type.paymentMethodType {
+            case .stripe(.card):
+                guard
+                    case .new(confirmParams: let params) = _paymentOption,
+                    let cardNumber = params.paymentMethodParams.card?.number
+                else {
+                    return nil
+                }
+                let brand = STPCardValidator.brand(forNumber: cardNumber)
+                let brandString = STPCardBrandUtilities.stringFrom(brand)
```
porter-stripe, line 380: This doesn't seem to work with co-branded cards, when I enter a co-branded (Visa/CB) card and choose CB, the copy says "Visa ...". You'll need to also check the preferred network (if it exists) on the card params.

`params.paymentMethodParams.card?.networks?.preferred`

There's also a weird edge case where you can fill out a form for a co-branded card and don't select a brand since it's optional, so the card brand gets decided on the backend.

yuki-stripe: Thanks for catching this! Fixed it up. Are there things we can do to make it harder to make this mistake?

porter-stripe: We did something similar to what you did with `func brand(for card: STPPaymentMethodCardParams?) -> STPCardBrand` on `STPPaymentMethodCard`. But this is the first instance AFAIK where we have needed to read it off the `STPPaymentMethodCardParams`.  Maybe we "internally" deprecate STPCardBrandUtilities.stringFrom(brand) somehow, since the card number is not the single source of truth for the card brand for a number.

porter-stripe (dismissed):

porter-stripe (dismissed):

porter-stripe (dismissed):

yuki-stripe:
I approved a bunch of snapshot tests that are not quite right; the labels can get squished a bit w/ the hidden "Change" button. Debugging & fixing this is too toilsome with the current RowButton architecture relative to the tiny UI problem here, so I'm going to wait to fix this until we refactor RowButton as a GA fast-follow (likely after we submit for Implementation Review & finish the next major version) @porter-stripe

porter-stripe (approved):

PR #4539, opened 2025-02-04
Add SwiftUI example

porter-stripe:
## Summary
- Adds SwiftUI example integration
- Adds UI test

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-3002

## Testing
- UI test

## Changelog
N/A

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/ExampleSwiftUIEmbeddedPaymentElement.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/ExampleSwiftUIEmbeddedPaymentElement.swift
@@ -1,0 +1,368 @@
+//
+//  ExampleSwiftUIEmbeddedPaymentElement.swift
+//  PaymentSheet Example
+//
+//  Created by Nick Porter on 2/4/25.
+//
+// This is an example integration using Embedded Payment Element in SwiftUI
+
+import SwiftUI
+@_spi(EmbeddedPaymentElementPrivateBeta) import StripePaymentSheet
+
+// MARK: - BackendViewModel
+class BackendViewModel: ObservableObject {
+    struct BackendResponse {
+        let publishableKey: String
+        let customerID: String
+        let ephemeralKey: String
+    }
+
+    private let baseUrl = "https://stripe-mobile-payment-sheet-custom-deferred.glitch.me"
+    private lazy var checkoutUrl = URL(string: baseUrl + "/checkout")!
+    private lazy var confirmIntentUrl = URL(string: baseUrl + "/confirm_intent")!
+    private lazy var computeTotalsUrl = URL(string: baseUrl + "/compute_totals")!
+    
+    /// Store the “current total” (in cents) that we last fetched from the backend
+    @Published private(set) var currentTotal: Int = 0
+    private var response: BackendResponse?
+    
+    // MARK: Step 1) Load ephemeral key & publishable key from /checkout
+    @MainActor
+    func loadCheckout() async throws -> BackendResponse {
+        var request = URLRequest(url: checkoutUrl)
+        request.httpMethod = "POST"
+        request.setValue("application/json", forHTTPHeaderField: "Content-type")
+        
+        let body: [String: Any] = [
+            "hot_dog_count": 5,
+            "salad_count": 0,
+            "is_subscribing": false
+        ]
+        request.httpBody = try JSONSerialization.data(withJSONObject: body, options: [])
+        
+        let (data, _) = try await URLSession.shared.data(for: request)
+        guard
+            let json = try JSONSerialization.jsonObject(with: data, options: []) as? [String: Any],
+            let pk = json["publishableKey"] as? String,
+            let cid = json["customer"] as? String,
+            let ekey = json["ephemeralKey"] as? String,
+            let total = json["total"] as? Int
+        else {
+            throw URLError(.badServerResponse)
+        }
+        
+        self.currentTotal = total
+        
+        // Set the publishable key on the Stripe SDK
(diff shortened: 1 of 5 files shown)


Example/PaymentSheet Example/PaymentSheet Example/ExampleSwiftUIEmbeddedPaymentElement.swift
```diff
@@ -0,0 +1,368 @@
...
+    @StateObject var embeddedViewModel = EmbeddedPaymentElementViewModel()
+    @StateObject var backendViewModel = BackendViewModel()
+    @State var confirmationResult: EmbeddedPaymentElementResult?
+    @State private var isSubscribing: Bool = false
+    
+    @Environment(\.dismiss) private var dismiss
+    
+    // MARK: Body
+    var body: some View {
+        VStack(spacing: 24) {
```
yuki-stripe, line 174: ooc why VStack here? Seems like there's at most one thing we're displaying, not a list.

Example/PaymentSheet Example/PaymentSheet Example/ExampleSwiftUIEmbeddedPaymentElement.swift
```diff
@@ -0,0 +1,368 @@
...
+                    .foregroundColor(.white)
+                    .background(Color.orange)
+                    .cornerRadius(6)
+                }
+            } else {
+                if !embeddedViewModel.isLoaded {
+                    ProgressView("Preparing Payment...")
+                } else {
+                    Text("Payment element not loaded.")
+                }
```
yuki-stripe, line 252: Can delete all of this except for ProgressView

Example/PaymentSheet Example/PaymentSheet Example/ExampleSwiftUIEmbeddedPaymentElement.swift
```diff
@@ -0,0 +1,368 @@
...
+        )
+        
+        // Set the customer ID & ephemeral key from the backendViewModel
+        configuration.customer = .init(
+            id: response.customerID,
+            ephemeralKeySecret: response.ephemeralKey
+        )
+        configuration.returnURL = "payments-example://stripe-redirect"
+        
+        try? await embeddedViewModel.load(intentConfiguration: intentConfig, configuration: configuration)
```
yuki-stripe, line 304: we should handle the error

Example/PaymentSheet Example/PaymentSheet Example/ExampleSwiftUIEmbeddedPaymentElement.swift
```diff
@@ -0,0 +1,368 @@
...
+    // MARK: Body
+    var body: some View {
+        VStack(spacing: 24) {
+            if embeddedViewModel.isLoaded {
+                ScrollView {
+                    // Embedded Payment Element
+                    EmbeddedPaymentElementView(viewModel: embeddedViewModel)
+                    
+                    // Display the selected payment option
+                    if let paymentOption = embeddedViewModel.paymentOption {
```
yuki-stripe, line 181: Hrmm we don't expect/want any merchants to actually display the payment option immediately below the embedded so I'm not sure if we should show it as an example. 

But there's some value to showing how to use `paymentOption`. Maybe we can just add a comment here saying like
// A real integration probably wouldn't show the selected payment option on the same screen as the embedded payment element. We display it as an example.

Example/PaymentSheet Example/PaymentSheet Example/ExampleSwiftUIEmbeddedPaymentElement.swift
```diff
@@ -0,0 +1,368 @@
...
+            } else {
+                if !embeddedViewModel.isLoaded {
+                    ProgressView("Preparing Payment...")
+                } else {
+                    Text("Payment element not loaded.")
+                }
+            }
+        }
+        .padding()
+        .onAppear {
```
yuki-stripe, line 256: should this be .task?

porter-stripe: Yes!!!

Example/PaymentSheet Example/PaymentSheet Example/ExampleSwiftUIEmbeddedPaymentElement.swift
porter-stripe: Yep, updated!

Example/PaymentSheet Example/PaymentSheet Example/ExampleSwiftUIEmbeddedPaymentElement.swift
porter-stripe: Not required! I removed the VStack in favor of a high level ScrollView instead.

Example/PaymentSheet Example/PaymentSheet Example/ExampleSwiftUIEmbeddedPaymentElement.swift
```diff
@@ -166,94 +166,102 @@ struct MyEmbeddedCheckoutView: View {
...
+                            await prepareEmbeddedPaymentElement()
+                        }
+                    }
                 }
             } else {
                 if !embeddedViewModel.isLoaded {
                     ProgressView("Preparing Payment...")
-                } else {
-                    Text("Payment element not loaded.")
                 }
```
yuki-stripe, line 260: I think this should just be 
```
                    ProgressView("Preparing Payment...")
```

PR #4542, opened 2025-02-08
Keep track of selected row button change state and persist it across updates

yuki-stripe:
## Motivation
Last piece for this change button row state feature - persist across updates

## Testing
See unit test


porter-stripe (approved):
🚢 !

PR #4547, opened 2025-02-10
Fix bug after updating selected CBC card

porter-stripe:
## Summary
- Fixed a bug where when updating a co-branded card, and coming back it was not still selected. See the JIRA ticket for the before.
- Simplified some logic in the SavedPaymentMethodRowButton, and refactored to make it less prone to bugs like this.

After:

https://github.com/user-attachments/assets/736a9951-1524-4357-b609-077aa46fe6fe

## Motivation
- https://jira.corp.stripe.com/browse/MOBILESDK-2725

## Testing
- Manual
- New unit tests

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/Vertical Saved Payment Method Screen/VerticalSavedPaymentMethodsViewController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/Vertical Saved Payment Method Screen/VerticalSavedPaymentMethodsViewController.swift
@@ -375,11 +375,14 @@ extension VerticalSavedPaymentMethodsViewController: UpdatePaymentMethodViewCont
         }
 
         // Create the new button
-        let newButton = SavedPaymentMethodRowButton(paymentMethod: updatedPaymentMethod, appearance: configuration.appearance, showDefaultPMBadge: isDefaultPaymentMethod(paymentMethodId: updatedPaymentMethod.stripeId))
+        let isDefaultPaymentMethod = isDefaultPaymentMethod(paymentMethodId: updatedPaymentMethod.stripeId)
+        let newButton = SavedPaymentMethodRowButton(paymentMethod: updatedPaymentMethod,
+                                                    appearance: configuration.appearance,
+                                                    showDefaultPMBadge: isDefaultPaymentMethod,
+                                                    previousSelectedState: oldButton.previousSelectedState,
+                                                    currentState: oldButton.state)
 
         newButton.delegate = self
-        newButton.previousSelectedState = oldButton.previousSelectedState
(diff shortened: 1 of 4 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/Vertical Saved Payment Method Screen/VerticalSavedPaymentMethodsViewController.swift
```diff
@@ -375,11 +375,14 @@ extension VerticalSavedPaymentMethodsViewController: UpdatePaymentMethodViewCont
         }
 
         // Create the new button
-        let newButton = SavedPaymentMethodRowButton(paymentMethod: updatedPaymentMethod, appearance: configuration.appearance, showDefaultPMBadge: isDefaultPaymentMethod(paymentMethodId: updatedPaymentMethod.stripeId))
+        let isDefaultPaymentMethod = isDefaultPaymentMethod(paymentMethodId: updatedPaymentMethod.stripeId)
+        let newButton = SavedPaymentMethodRowButton(paymentMethod: updatedPaymentMethod,
+                                                    appearance: configuration.appearance,
+                                                    showDefaultPMBadge: isDefaultPaymentMethod,
+                                                    previousSelectedState: oldButton.previousSelectedState,
+                                                    currentState: oldButton.state)
```
wooj-stripe, line 383: Just double checking here -- the bug fix is to pass these values need through init() because addAndPinSubview() uses these state values to render the view?

wooj-stripe: Ah.. right.. setting `newButton.state` set's the previousSelected state, so we just overwrote setting `newButton.previousSelectedState = oldButton.previousSelectedState` -- nice catch!

And also thank you for making `previousSelectedState` have the `private(set)` modifier!

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/SavedPaymentMethodRowButtonTests.swift
```diff
@@ -0,0 +1,89 @@
...
+            appearance: appearance,
+            currentState: .unselected
+        )
+        
+        sut.state = .selected
+        
+        XCTAssertTrue(sut.isSelected, "Button should become selected")
+        XCTAssertEqual(sut.previousSelectedState, .unselected, "previousSelectedState should track the old value")
+        XCTAssertTrue(sut.chevronButton.isHidden, "Chevron is hidden when not editing.")
+    }
```
wooj-stripe, line 72: nice test!

wooj-stripe (approved):

PR #4549, opened 2025-02-11
Deflake testURLParamsWithCardAndBankFundingSources

wooj-stripe:
## Summary
This test has about a 50% chance of failing. To remedy this, I am proposing we continue to deserialize as a Set(), but during encoding, convert to an orderedlist.

## Motivation
Deflaking test

## Testing
Using existing tests

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Utils/LinkURLGenerator.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Utils/LinkURLGenerator.swift
@@ -137,14 +137,7 @@ class LinkURLGenerator {
 extension Set where Element == LinkSettings.FundingSource {
     func toSortedArray() -> [LinkSettings.FundingSource] {
         return self.sorted { a, b in
-            switch (a, b) {
-            case (.card, .bankAccount):
-                return true
-            case (.bankAccount, .card):
-                return false
-            default:
-                return false
-            }
+            a.rawValue.localizedCaseInsensitiveCompare(b.rawValue) == .orderedAscending
         }
     }
 }
--- a/StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Utils/LinkURLGenerator.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Utils/LinkURLGenerator.swift
@@ -33,7 +33,7 @@ struct LinkURLParams: Encodable {
         var stripePreferredNetworks: [String]
         var supportedCobrandedNetworks: [String: Bool]
     }
-    
+
     var path = "mobile_pay"
     var integrationType = "mobile"
     var paymentObject: PaymentObjectMode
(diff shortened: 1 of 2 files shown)


davidme-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Utils/LinkURLGenerator.swift
```diff
@@ -133,6 +133,22 @@ class LinkURLGenerator {
...
+        return self.sorted { a, b in
+            switch (a, b) {
+            case (.card, .bankAccount):
+                return true
+            case (.bankAccount, .card):
+                return false
+            default:
+                return false
+            }
+        }
```
mats-stripe, line 148: Do we specifically want `card` before `bankAccount` here, or could we sort alphabetically by raw value? It would future-proof this logic for any new funding source that gets added (you never know! 😅)

```suggestion
        return self.sorted { a, b in
            a.rawValue.localizedCaseInsensitiveCompare(b.rawValue) == .orderedAscending
        }
```

mats-stripe (approved):
Drive-by review with one non-blocking suggestion

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Utils/LinkURLGenerator.swift
wooj-stripe: Sounds good, but will need to also update the test to reflect this.

davidme-stripe (dismissed):

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Utils/LinkURLGenerator.swift
```diff
@@ -50,7 +50,7 @@ struct LinkURLParams: Encodable {
     var intentMode: IntentMode
     var setupFutureUsage: Bool
     var cardBrandChoice: CardBrandChoiceInfo?
-    var linkFundingSources: Set<LinkSettings.FundingSource>
+    var linkFundingSources: [LinkSettings.FundingSource]
```
tillh-stripe, line 53: The idea behind using a set was to not care about the ordering in this test (and in general), but I guess the custom `Equatable` implementation doesn’t really allow for that 😔

mats-stripe (dismissed):

tillh-stripe (dismissed):

tillh-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Utils/LinkURLGenerator.swift
davidme-stripe: The issue is that we're serializing it to JSON data, and we sort the keys when doing that, but if we use a `Set` then the ordering is non-deterministic during the test.

PR #4552, opened 2025-02-11
set as default test flakiness

joyceqin-stripe:
## Summary

Added a delay in testPaymentSheetLoadAndConfirmWithPaymentIntentSetAsDefault, which was occasionally failing with nil default payment method in the CI. Tested locally running the relevant tests 100 times each.


porter-stripe (approved):

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheet+APITest.swift
```diff
@@ -397,6 +397,7 @@ class PaymentSheetAPITest: STPNetworkStubbingTestCase {
                         ) { result, _ in
                             switch result {
                             case .completed:
+                                self.waitForDefaultPaymentMethodToBePersisted()
```
wooj-stripe, line 400: Have we opened a thread for improving this on the backend?

joyceqin-stripe: I don't think so— it didn't seem like the delay would have too much of an impact, right?

porter-stripe (approved):

PR #4553, opened 2025-02-11
Update embedded example

yuki-stripe:
- Use client-side confirmation
- Use continue form sheet action
- Fix bug where we didn't disable interaction after you hit buy

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-2577

## Testing
Manual

## Changelog
not user facing

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/ExampleEmbeddedElementCheckoutViewController.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/ExampleEmbeddedElementCheckoutViewController.swift
@@ -55,7 +55,7 @@ class ExampleEmbeddedElementCheckoutViewController: UIViewController {
                         intentCreationCallback(.failure(ExampleError()))
                         return
                     }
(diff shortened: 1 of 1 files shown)


Example/PaymentSheet Example/PaymentSheet Example/ExampleEmbeddedElementCheckoutViewController.swift
```diff
@@ -106,8 +106,10 @@ class ExampleEmbeddedElementCheckoutViewController: UIViewController {
     func didTapCheckoutButton() {
         Task {
             // MARK: - Confirm the payment
+            view.isUserInteractionEnabled = false
             let result = await embeddedPaymentElement.confirm()
             handlePaymentResult(result)
+            view.isUserInteractionEnabled = true
```
porter-stripe, line 112: FYI, this is fixed here and probably shouldn't be handled by the merchant: https://github.com/stripe/stripe-ios/pull/4516

porter-stripe: Actually, disabling their own UI sounds right too, but ours should be disabled too

PR #4556, opened 2025-02-12
Fix a bug on setting presentingViewController in SwiftUI

porter-stripe:
## Summary
- There was a bug where we would set the presentingViewController to one of our own VCs which is incorrect and would cause the "Change" button to no-op.
- The bug was caused in the following scenario: User fills out card, hits continue, then hits change. When tapping change no form would be presented, it would no-op.
- This was caused by `updateUIView` on our UIViewRepresentable being called while our form VC is being dismissed. So when we compute the top most VC, we get back our bottom sheet VC even though it's in the process of dismissing.
- The fix is two fold: 1. Make sure we never set presenting VC to one of our own. 2. If the top most VC is in the process of dismissing when `updateUIView` is called, wait 0.1 seconds and try again.

## Motivation
- Embedded SwiftUI bug fix

## Testing
- Manual
- Updated UI test

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedViewRepresentable.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedViewRepresentable.swift
@@ -37,8 +37,21 @@ struct EmbeddedViewRepresentable: UIViewRepresentable {
     }
 
     public func updateUIView(_ uiView: UIView, context: Context) {
-        // Update the presenting view controller in case it has changed
-        viewModel.embeddedPaymentElement?.presentingViewController = UIWindow.visibleViewController
+        guard let visibleVC = UIWindow.visibleViewController else { return }
+        
+        // If visibleVC in the process of dismissing, skip for now and retry shortly.
+        // updateUIView can be trigged by a view controller (such as a form) being dismissed
+        guard !visibleVC.isBeingDismissed else {
+            DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) {
+                // Re-trigger SwiftUI’s update cycle
+                viewModel.objectWillChange.send()
+            }
+            return
+        }
+
+        if !(visibleVC is StripePaymentSheet.BottomSheetViewController) {
+            viewModel.embeddedPaymentElement?.presentingViewController = visibleVC
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedViewRepresentable.swift
```diff
@@ -30,15 +30,28 @@ struct EmbeddedViewRepresentable: UIViewRepresentable {
...
+        guard !visibleVC.isBeingDismissed else {
+            DispatchQueue.main.asyncAfter(deadline: .now() + 0.1) {
+                // Re-trigger SwiftUI’s update cycle
+                viewModel.objectWillChange.send()
+            }
+            return
+        }
+
+        if !(visibleVC is StripePaymentSheet.BottomSheetViewController) {
+            viewModel.embeddedPaymentElement?.presentingViewController = visibleVC
```
wooj-stripe, line 53: 1. I'm having trouble figuring out why we need to set`viewModel.embeddedPaymentElement?.presentingViewController` in `updateUIView`.  Seems like when the view is created, we set `embeddedPaymentElement.presentingViewController = UIWindow.visibleViewController` so curious as to know why this is needed.

2. I thought about adding a failsafe to have a maximum number of retires to set the controller, but seems necessary.

3. Timeout of .1s seems fine -- most animations are about .3s, so attempting multiple times at this rate seems good.

4. Your code/logic does make sense, but I wish that there was another solution, given that this is a piece of code that will probably need to execute often --e.g. filling out a card, hitting continue, then hitting "change" seems like it should not put our code in a fragile spot where we are retrying/fumbling around the viewController stack.  Have we explored other ways of handling this?

porter-stripe: The reason for needing to re-set the presentingViewController even after creation is the same as [here](https://github.com/stripe/stripe-ios/blob/master/Example/PaymentSheet%20Example/PaymentSheet%20Example/ExampleEmbeddedElementCheckoutViewController.swift#L95).

The merchant could show the embedded view in a standalone view, then call confirm in a different view. Calling confirm could then require us to show a VC, like CVC collection or just a web view to continue. I've attached a video of this flow.

Agree the solution isn't amazing but it's one of those concessions we make w/ SwiftUI <> UIKit imo.


https://github.com/user-attachments/assets/2563274f-9a9b-45cd-a9b8-5053f830cf41

wooj-stripe (approved):

PR #4559, opened 2025-02-12
Fail confirm if update is in progress

porter-stripe:
## Summary
- Fails a confirm if the previous update has not yet finished

## Motivation
- https://jira.corp.stripe.com/browse/MOBILESDK-3066

## Testing
- Updated and added a new unit test

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
@@ -139,7 +139,7 @@ extension EmbeddedPaymentElement: EmbeddedPaymentMethodsViewDelegate {
         // Note `paymentOption` derives from this property
         self.selectedFormViewController = Self.makeFormViewControllerIfNecessary(
             selection: embeddedPaymentMethodsView.selectedRowButton?.type,
-            previousPaymentOption:  selectedFormViewController?.previousPaymentOption,
+            previousPaymentOption: selectedFormViewController?.previousPaymentOption,
             configuration: configuration,
             intent: intent,
             elementsSession: elementsSession,
@@ -361,7 +361,7 @@ extension EmbeddedPaymentElement: EmbeddedFormViewControllerDelegate {
         embeddedFormViewController.dismiss(animated: true)
         informDelegateIfPaymentOptionUpdated()
     }
-    
+
     func getChangeButtonState(for type: RowButtonType) -> (shouldShowChangeButton: Bool, sublabel: String?) {
         guard let _paymentOption, let displayData = paymentOption else {
             return (false, nil)
@@ -373,7 +373,7 @@ extension EmbeddedPaymentElement: EmbeddedFormViewControllerDelegate {
             }
             return false
         }()
-        
+
         // Add a sublabel to the selected row for cards and us bank account like "Visa 4242"
         let sublabel: String? = {
             switch type.paymentMethodType {
@@ -390,7 +390,7 @@ extension EmbeddedPaymentElement: EmbeddedFormViewControllerDelegate {
                 return nil
             }
         }()
-        
+
         return (shouldShowChangeButton: shouldShowChangeButton, sublabel: sublabel)
     }
 }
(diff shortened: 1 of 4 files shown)


wooj-stripe (approved):

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentElementTest.swift
```diff
@@ -196,7 +196,7 @@ class EmbeddedPaymentElementTest: XCTestCase {
         sut.embeddedPaymentMethodsView.didTap(rowButton: sut.embeddedPaymentMethodsView.getRowButton(accessibilityIdentifier: "Cash App Pay"))
         // ...updating w/ a broken config...
         let brokenConfig = EmbeddedPaymentElement.IntentConfiguration(mode: .payment(amount: -1000, currency: "bad currency"), confirmHandler: { _, _, _ in })
-        async let _ = sut.update(intentConfiguration: brokenConfig)
+        _ = await sut.update(intentConfiguration: brokenConfig)
```
yuki-stripe, line 199: What's the reason behind this change?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
```diff
@@ -104,6 +104,11 @@ public final class EmbeddedPaymentElement {
     public func update(
         intentConfiguration: IntentConfiguration
     ) async -> UpdateResult {
+        isUpdating = true
+        // Ensure `isUpdating` is reset by using a defer block, preventing accidental omissions in early returns.
+        defer {
+            isUpdating = false
+        }
```
yuki-stripe, line 111: Does this work?

If you submit two updates U1 and U2 in a row, when U1 finishes it sets `isUpdating = false`.  What if U2 is still in progress?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -404,6 +404,11 @@ extension EmbeddedPaymentElement {
         guard !hasConfirmedIntent else {
             return (.failed(error: PaymentSheetError.embeddedPaymentElementAlreadyConfirmedIntent), nil)
         }
+
+        guard !isUpdating else {
+            return (.failed(error: PaymentSheetError.embeddedPaymentElementConfirmDuringUpdate), nil)
+        }
```
yuki-stripe, line 410: can we add a comment for context on why we don't just wait for `update` to finish here.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -404,6 +404,11 @@ extension EmbeddedPaymentElement {
...
+
         // Wait for the last update to finish and fail if didn't succeed. A failure means the view is out of sync with the intent and could e.g. not be showing a required mandate.
         if let latestUpdateTask {
```
yuki-stripe, line 413: Is this code still needed?

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentElementTest.swift
```diff
@@ -196,7 +196,7 @@ class EmbeddedPaymentElementTest: XCTestCase {
...
         // ...and immediately calling confirm, before the 1st update finishes...
```
yuki-stripe, line 200: Isn't this the behavior being tested by the test you added?  Calling confirm when an update is in progress?

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentElementTest.swift
porter-stripe: We may actually have a threading problem on master! In this test if you change it from an invalid config to a valid one, the code just deadlocks. I think it may be due to the test code but this shouldn't occur.

PR #4562, opened 2025-02-12
Floating button refactor in RowButton

porter-stripe:
## Summary
- Refactors the second button style, floating into it's own content view

## Motivation
- RowButton too big

## Testing
- Manual
- Snapshots

## Changelog
N/A


porter-stripe:
@yuki-stripe apologies for the commit history here, rebase didn't go as planned!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButton.swift
```diff
@@ -342,7 +351,13 @@ class RowButton: UIView {
             return
         }
         // Don't do this if we *are* the tallest variant; otherwise we'll infinite loop!
-        guard sublabel.text?.isEmpty ?? true else {
+        let isSublabelTextEmpty = {
+            if let content {
+                return !content.hasSubtext
+            }
```
yuki-stripe, line 357: Why is this needed?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButtonContent.swift
```diff
@@ -5,20 +5,43 @@
 //  Created by Nick Porter on 2/12/25.
 //
 
+@_spi(STP) import StripeUICore
 import UIKit
 
 /// Defines a view that is contained with a `RowButton`
-protocol RowButtonContent: UIView {
+protocol RowButtonContent: UIView, EventHandler {
```
yuki-stripe, line 12: I'm growing more convinced that subclassing is the right strategy here. I feel like these RowButtonX classes *are* RowButtons. They'll always all share a common set of properties. They all have the same behaviors except for a handful which they override.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButton.swift
porter-stripe: In the current state of the refactor it's not required as we still have the "self.sublabel" available, but in the next PR sublabel will be removed and we'll need to look at the content view as we're doing here. Happy to remove and add this later when we really need it.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButtonContent.swift
porter-stripe: Yeah, I agree that subclassing could work too, but I’m not seeing compelling benefits over our current composition approach, particularly given that we’d have to restart a good chunk of the refactor. Both approaches could scale similarly, though subclassing could turn RowButton into a ‘god class’ if more variations appear in the future, but composition isn’t immune to that risk either, but it does fit better with our existing pattern of avoiding UIKit subclasses when possible. Unless there’s a strong argument for subclassing, I’d prefer to stick with composition as we discussed. Definitely willing to hear the case for subclassing if you think there's some benefits!

yuki-stripe: Benefits are that subclasses can inherit an required initializer so that you can't forget to handle a new subview in one particular RowButtonContent, subclasses can inherit and override and call super on methods like `setKeyContent`, they inherit all the shared properties (so we don't have to repeat the 8 properties 3-4 times).

It'd probably make most sense to try subclassing after you finish the third style.

yuki-stripe (approved):

PR #4573, opened 2025-02-14
Update mandate before notifying delegate

porter-stripe:
## Summary
- Fixes a bug introduced in https://github.com/stripe/stripe-ios/pull/4538 where we update the mandate after notifying the delegate the payment option has changed
- This bug was only evident in our playground example app when you would have the merchant display the mandate not us
- This resulted in a nil mandateText being provided on our payment option to merchants

## Motivation
- 🐛 

## Testing
- Manual
- New test

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentMethodsViewTests.swift
+++ b/StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentMethodsViewTests.swift
@@ -50,13 +50,13 @@ final class EmbeddedPaymentMethodsViewTests: XCTestCase {
 
         // Simulate tapping Cash App and verify delegate is called
         embeddedView.didTap(rowButton: embeddedView.getRowButton(accessibilityIdentifier: "Cash App Pay"))
-        XCTAssertEqual(mockDelegate.calls, [.didUpdateSelection, .didUpdateHeight, .didTapPaymentMethodRow])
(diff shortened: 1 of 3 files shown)


StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentMethodsViewTests.swift
```diff
@@ -50,13 +50,13 @@ final class EmbeddedPaymentMethodsViewTests: XCTestCase {
 
         // Simulate tapping Cash App and verify delegate is called
         embeddedView.didTap(rowButton: embeddedView.getRowButton(accessibilityIdentifier: "Cash App Pay"))
-        XCTAssertEqual(mockDelegate.calls, [.didUpdateSelection, .didUpdateHeight, .didTapPaymentMethodRow])
+        XCTAssertEqual(mockDelegate.calls, [.didUpdateHeight, .didUpdateSelection, .didTapPaymentMethodRow])
```
porter-stripe, line 53: Now that we update the mandate before notifying delegate of the payment option changing this ordering changes. This seems reasonable as it's pretty much simultaneous.

porter-stripe: We use `layoutSubviews` to auto listen for these height changes which triggers the didUpdateHeight, I think I prefer keeping it there than trying to keep track of height changes on our own to control the ordering, so I'm proposing we call didUpdateHeight before didUpdateSelection in this case.

yuki-stripe (approved):
Nice test and fix, thanks!

PR #4583, opened 2025-02-18
Refactor Flat with check 

porter-stripe:
## Summary
- Refactored the flat with checkmark style into it's own UIView
- Removed all old layout code from RowButton

## Motivation
- Clean up row button

## Testing
- Manual
- Existing tests

## Changelog
N/A


StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/VerticalPaymentMethodListViewControllerTest.swift
```diff
@@ -74,7 +74,7 @@ final class VerticalPaymentMethodListViewControllerTest: XCTestCase {
             incentive: nil,
             delegate: self
         )
-        XCTAssertEqual(["SEPA Debit", "Card", "Apple Pay", "Link"], sut.rowButtons.map { $0.label.text })
+        XCTAssertEqual(["SEPA Debit", "Card", "Apple Pay", "Link"], sut.rowButtons.map { $0.text })
```
porter-stripe, line 77: `text` on RowButton is flagged as dead but used here, lmk what you think.

yuki-stripe: Looks like `text` just contains the value that was passed as the initializer's `text` parameter. What's the intention?  Based on this use case, `text` should be a computed var that returns the label's text.

yuki-stripe (approved):
My hunch is if you try to subclass after this PR you'll be pleasantly surprised by how well it fits the code here, and how much code you can delete/get for free.

PR #4585, opened 2025-02-19
Set as default API

joyceqin-stripe:
## Summary

Added internal endpoint API, refactored analytics, and hid default checkbox when it's already the default
## Testing

testSetAsDefault_CustomerSession()
call apiClient.setAsDefaultPaymentMethod, load PaymentSheet and check that the elements session matches default


StripeCore/StripeCore/Resources/Localizations/en.lproj/Localizable.strings
```diff
@@ -1,12 +1,15 @@
...
-"Scan Card" = "Scan Card";
+/* An error message displayed when setting a default payment method fails. */
+"Default payment method was not changed. Please try again." = "Default payment method was not changed. Please try again.";
 
 /* Button title to open camera to scan credit/debit card */
 "Scan card" = "Scan card";
 
+/* Text for button to scan a credit card */
+"Scan Card" = "Scan Card";
+
```
porter-stripe, line 12: Kinda strange the diff looks like this but seems ~fine

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSheetDataSource.swift
```diff
@@ -181,6 +181,15 @@ extension CustomerSheetDataSource {
         }
     }
 
+    func setAsDefaultPaymentMethod(paymentMethodId: String, customerID: String) async throws -> STPCustomer? {
+        switch dataSource {
+        case .customerAdapter:
+            return nil
```
porter-stripe, line 187: Can you add a comment about why we do this? Maybe we should throw an error in this case.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
```diff
@@ -104,7 +104,9 @@ final class UpdatePaymentMethodViewController: UIViewController {
     }()
 
     private lazy var setAsDefaultCheckbox: CheckboxElement? = {
-        guard viewModel.canSetAsDefaultPM && PaymentSheet.supportedDefaultPaymentMethods.contains(where: {
+        guard viewModel.canSetAsDefaultPM,
```
porter-stripe, line 107: It seems like canSetAsDefaultPM should check isDefault so it's more of the single source of truth, otherwise it reads funny that we need to check something else if it's able to be same with the supportedDefaultPaymentMethods check

StripePayments/StripePayments/Source/API Bindings/STPAPIClient+Payments.swift
```diff
@@ -1400,6 +1400,40 @@ extension STPAPIClient {
             }
         })
     }
+
+    @_spi(STP) public func setAsDefaultPaymentMethod(
```
porter-stripe, line 1404: Let's add some doc comments here like we do for other API functions

porter-stripe (commented):
Looks good just a few comments!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSheetDataSource.swift
joyceqin-stripe: I've added a fatal error that explains that customer adapter does not support the set as default feature

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSheetDataSource.swift
```diff
@@ -181,10 +181,10 @@ extension CustomerSheetDataSource {
         }
     }
 
-    func setAsDefaultPaymentMethod(paymentMethodId: String, customerID: String) async throws -> STPCustomer? {
+    func setAsDefaultPaymentMethod(paymentMethodId: String, customerID: String) async throws -> STPCustomer {
         switch dataSource {
         case .customerAdapter:
-            return nil
+            fatalError("CustomerAdapter does not support the set as default payment method feature")
```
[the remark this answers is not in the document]
joyceqin-stripe: Just did, but in the process of doing so, the UpdatePaymentMethodViewModel's initializer became failable and I had to unwrap the optionals in all the instances it was used. I'm not sure that's the best way— is there a better way?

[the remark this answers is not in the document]
joyceqin-stripe: Just did

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewModel.swift
```diff
@@ -50,11 +50,12 @@ class UpdatePaymentMethodViewModel {
         case .SEPADebit:
             return .Localized.manage_sepa_debit
         default:
-            fatalError("Updating payment method has not been implemented for \(paymentMethod.type)")
+            stpAssertionFailure("Updating payment method has not been implemented for \(paymentMethod.type)")
+            return nil
```
porter-stripe, line 54: To improve ergonomics when using these I think we should just use an empty string rather than nil. Also let's use assertionFailure instead of stpAssertion, I updated my comment shortly after I had originally suggested stpAssertion. We want these to fire in merchant apps.

/// A wrapper that only calls `assertionFailure` when the `ENABLE_STPASSERTIONFAILURE` compiler flag is set.
/// Use this for assertions that should not trigger in merchant apps.
@inlinable @_spi(STP) public func stpAssertionFailure

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewModel.swift
```diff
@@ -63,13 +64,15 @@ class UpdatePaymentMethodViewModel {
...
         }
     }()
 
-    init(paymentMethod: STPPaymentMethod, appearance: PaymentSheet.Appearance, hostedSurface: HostedSurface, cardBrandFilter: CardBrandFilter = .default, canRemove: Bool, isCBCEligible: Bool, allowsSetAsDefaultPM: Bool = false, isDefault: Bool = false) {
+    init?(paymentMethod: STPPaymentMethod, appearance: PaymentSheet.Appearance, hostedSurface: HostedSurface, cardBrandFilter: CardBrandFilter = .default, canRemove: Bool, isCBCEligible: Bool, allowsSetAsDefaultPM: Bool = false, isDefault: Bool = false) {
         guard PaymentSheet.supportedSavedPaymentMethods.contains(paymentMethod.type) else {
-            fatalError("Unsupported payment type \(paymentMethod.type) in UpdatePaymentMethodViewModel")
+            stpAssertionFailure("Unsupported payment type \(paymentMethod.type) in UpdatePaymentMethodViewModel")
+            return nil
```
porter-stripe: Can you find a way to make this not failable? The `guard let updateViewModel else { return }` everywhere seem smelly. Make just make this some kind of empty/default view model.

joyceqin-stripe: Made it a default view model

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewModel.swift
```diff
@@ -64,12 +64,24 @@ class UpdatePaymentMethodViewModel {
...
+        guard PaymentSheet.supportedSavedPaymentMethods.contains(paymentMethod.type) else {
+            assertionFailure("Unsupported payment type \(paymentMethod.type) in UpdatePaymentMethodViewModel")
+            self.paymentMethod = STPPaymentMethod(stripeId: "", type: .unknown)
+            self.appearance = PaymentSheet.Appearance()
+            self.hostedSurface = .paymentSheet
+            self.cardBrandFilter = .default
+            self.canRemove = false
+            self.isCBCEligible = false
+            self.canSetAsDefaultPM = false
+            self.isDefault = false
```
porter-stripe, line 82: Hmm, any reason we can't just fall through after the assert?
```swift
    init(paymentMethod: STPPaymentMethod, appearance: PaymentSheet.Appearance, hostedSurface: HostedSurface, cardBrandFilter: CardBrandFilter = .default, canRemove: Bool, isCBCEligible: Bool, allowsSetAsDefaultPM: Bool = false, isDefault: Bool = false) {
        if !PaymentSheet.supportedSavedPaymentMethods.contains(paymentMethod.type) {
            assertionFailure("Unsupported payment type \(paymentMethod.type) in UpdatePaymentMethodViewModel")
        }
        self.paymentMethod = paymentMethod
        self.appearance = appearance
        self.hostedSurface = hostedSurface
        self.cardBrandFilter = cardBrandFilter
        self.canRemove = canRemove
        self.isCBCEligible = isCBCEligible
        self.canSetAsDefaultPM = allowsSetAsDefaultPM && !isDefault
        self.isDefault = isDefault
    }
```

joyceqin-stripe: I suppose not

porter-stripe (approved):

PR #4587, opened 2025-02-19
Fix embedded nil payment option bug

yuki-stripe:
## Summary
Fixes the following embedded bug:
1. Enter card details, hit continue
2. Go back into card form, make the form invalid, and close the sheet
The problems:
• We don't inform the merchant that the payment option is nil.
• The "card" row button remains selected. It should be cleared out.

Also:
- Adds a localized error string when you attempt confirmation when there's no valid payment option, vend that for FC and Embedded. 
- Fixes incorrect usage of `stpAssertion` ("/// Use this for assertions that **should not trigger in merchant apps**.")

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-3078

## Testing
Added test to UI Test and also removed it redundantly testing US Bank Account (already tested by `testUSBankAccount`) and saved PMs (already tested by `testConfirmationWithUserButton_savedPaymentMethod`)

## Changelog
See changelog

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Categories/String+Localized.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Categories/String+Localized.swift
@@ -468,7 +468,7 @@ extension String.Localized {
        )
     }
 
-    @_spi(STP) public static var set_as_default_payment_method: String {
+    public static var set_as_default_payment_method: String {
         STPLocalizedString(
             "Set as default payment method",
             "Label of a checkbox that when checked makes a payment method as the default one."
@@ -479,6 +479,13 @@ extension String.Localized {
         STPLocalizedString(
             "Change",
             "Label for a button that lets you change the details of a payment method"
+       )
+    }
+
+    static var please_choose_a_valid_payment_method: String {
+        STPLocalizedString(
+            "Please choose a valid payment method.",
+            "Error message that's displayed when you try to confirm a payment without a valid payment method"
        )
     }
 }
--- a/StripePaymentSheet/StripePaymentSheet/Source/Categories/String+Localized.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Categories/String+Localized.swift
@@ -468,7 +468,7 @@ extension String.Localized {
        )
     }
 
-    public static var set_as_default_payment_method: String {
+    static var set_as_default_payment_method: String {
         STPLocalizedString(
             "Set as default payment method",
             "Label of a checkbox that when checked makes a payment method as the default one."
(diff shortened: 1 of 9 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
```diff
@@ -36,7 +36,7 @@ class EmbeddedPaymentMethodsView: UIView {
     private let appearance: PaymentSheet.Appearance
     private let rowButtonAppearance: PaymentSheet.Appearance
     private let customer: PaymentSheet.CustomerConfiguration?
-    private var previousSelectedRowButton: RowButton? {
+    var previousSelectedRowButton: RowButton? {
```
porter-stripe, line 39: Can this be private(set)?

StripePaymentSheet/StripePaymentSheet/Source/Categories/String+Localized.swift
```diff
@@ -468,7 +468,7 @@ extension String.Localized {
        )
     }
 
-    @_spi(STP) public static var set_as_default_payment_method: String {
+    public static var set_as_default_payment_method: String {
```
porter-stripe, line 471: Is this really supposed to be public? I'm wondering why the public API check isn't picking this up.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
```diff
@@ -209,13 +209,12 @@ public final class EmbeddedPaymentElement {
         analyticsHelper.log(event: .mcConfirmEmbedded)
         guard let presentingViewController else {
             let errorMessage = "Presenting view controller is nil. Please set EmbeddedPaymentElement.presentingViewController."
-            stpAssertionFailure(errorMessage)
+            assertionFailure(errorMessage)
```
porter-stripe, line 212: Why not stpAssertionFailure? it calls assertionFailure under the hood

StripePaymentSheet/StripePaymentSheet/Source/Categories/String+Localized.swift
yuki-stripe: Oh whoops I meant to delete `public` as well.  IDK why that didn't get flagged!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
yuki-stripe: https://github.com/stripe/stripe-ios/blob/8a549e7c/StripeCore/StripeCore/Source/Helpers/STPAssert.swift#L20-L21

```
/// A wrapper that only calls `assertionFailure` when the `ENABLE_STPASSERTIONFAILURE` compiler flag is set.
/// Use this for assertions that should not trigger in merchant apps.
```

porter-stripe (approved):

porter-stripe (approved):

PR #4589, opened 2025-02-19
Refactor CBC to use updateParams

wooj-stripe:
## Summary
* Refactors CBC so that it uses updateParams instead of using `hasChangedCardBrand`.

* Removed `openCardBrandDropdown` and `closeCardBrandDropdown` as this seems like too granular of analytics that I don't think we need -- instead it just logs cardBrandSelected, when the card brand changes.

Note: While removing implementing the cardBrandSelected in `CardSectionElement`, i went down the path of changing all elements to `lazy var xxx`, but felt this was too high risk of a change. This WAS needed to debounce the number of metrics we were sending, but found a way to do this with PaymentSheetAnalyticsHelper instead.

## Motivation
Refactoring code to make update payment method MM/YY, name on card, eaiser.

## Testing
Relying on existing tests


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/CardSection/CardSectionElement.swift
```diff
@@ -146,22 +147,6 @@ final class CardSectionElement: ContainerElement {
         self.preferredNetworks = preferredNetworks
         self.lastPanElementValidationState = panElement.validationState
         cardSection.delegate = self
-
-        // Hook up CBC analytics after self is initialized
-        self.cardBrandDropDown?.didPresent = { [weak self] in
```
porter-stripe, line 151: Is didPresent used anywhere now? We may be able to remove it, same w/ didTapClose.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewModel.swift
```diff
@@ -80,4 +76,26 @@ class UpdatePaymentMethodViewModel {
...
+    }
+    func hasChangedFields(original: STPPaymentMethodCard, updated: STPPaymentMethodCardParams) -> Bool {
+        let updatedBrand = original.preferredDisplayBrand != updated.networks?.preferred?.toCardBrand
+        return updatedBrand
+    }
+}
+
+extension UpdatePaymentMethodViewModel {
+    enum UpdatePaymentMethodOptions {
+        case card(paymentMethodCardParams: STPPaymentMethodCardParams)
```
porter-stripe, line 99: @joyceqin-stripe does this look ok with how we update banks and SEPA?

porter-stripe (commented):
looks good!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/SavedPaymentMethodFormFactory/SavedPaymentMethodFormFactory+Card.swift
```diff
@@ -13,40 +13,45 @@ import Foundation
...
+                    STPCard.brand(from: $0)
+                }).filter { self.viewModel.cardBrandFilter.isAccepted(cardBrand: $0) } ?? []
+                let cardBrand = cardBrands[field.selectedIndex]
+                let preferredNetworkAPIValue = STPCardBrandUtilities.apiValue(from: cardBrand)
+
+                params.paymentMethodParams.card?.networks = .init(preferred: preferredNetworkAPIValue)
+
+                if preferredNetworkAPIValue != self.lastCardBrandLogSelectedEventSent {
+                    STPAnalyticsClient.sharedClient.logPaymentSheetEvent(event: self.viewModel.hostedSurface.analyticEvent(for: .cardBrandSelected),
+                                                                         params: ["selected_card_brand": preferredNetworkAPIValue])
```
joyceqin-stripe, line 46: Do we not want "cbc_event_source": "edit" in the analytics params anymore? "cbc_event_source": "add" was still included in logCardBrandSelected

joyceqin-stripe (approved):
Lgtm

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewModel.swift
joyceqin-stripe: We don't currently support updating banks or SEPA, but this should work fine with setting default

wooj-stripe: The idea is that we'd add additional cases for other payment methods as we add support for updating them.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewModel.swift
```diff
@@ -80,4 +76,26 @@ class UpdatePaymentMethodViewModel {
...
+           let cardParams = params.paymentMethodParams.card,
+           let originalPaymentMethodCard = paymentMethod.card,
+           hasChangedFields(original: originalPaymentMethodCard, updated: cardParams) {
+            return .card(paymentMethodCardParams: cardParams)
+        }
+        return nil
+    }
+    func hasChangedFields(original: STPPaymentMethodCard, updated: STPPaymentMethodCardParams) -> Bool {
+        let updatedBrand = original.preferredDisplayBrand != updated.networks?.preferred?.toCardBrand
+        return updatedBrand
```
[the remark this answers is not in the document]
wooj-stripe: I was going to keep this here, because I am going to add support for detecting changes for MM/YY, name [on card], etc.."

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/SavedPaymentMethodFormFactory/SavedPaymentMethodFormFactory+Card.swift
wooj-stripe: Ah! great catch, i think i missed adding `"cbc_event_source": "edit"`. Will add that!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/CardSection/CardSectionElement.swift
wooj-stripe: Removed!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/SavedPaymentMethodFormFactory/SavedPaymentMethodFormFactory.swift
```diff
@@ -12,13 +12,14 @@ import Foundation
 import UIKit
 
 protocol SavedPaymentMethodFormFactoryDelegate: AnyObject {
-    func didUpdate(_: Element, didUpdateCardBrand: Bool)
+    func didUpdate(_: Element)
 }
 
 class SavedPaymentMethodFormFactory {
```
yuki-stripe, line 18: This class seems fishy. It's called a "Factory" which implies it's responsibility is to make instances of something (a saved card form?), but it
- owns the form it creates
- is the delegate of the form it creates
- doesn't even have a public method to create instances of a saved card form?

At the minimum it seems like this should be renamed to some sort of "Controller".

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/SavedPaymentMethodFormFactory/SavedPaymentMethodFormFactory+Card.swift
```diff
@@ -13,40 +13,46 @@ import Foundation
...
+                let cardBrand = cardBrands[field.selectedIndex]
+                let preferredNetworkAPIValue = STPCardBrandUtilities.apiValue(from: cardBrand)
+
+                params.paymentMethodParams.card?.networks = .init(preferred: preferredNetworkAPIValue)
+
+                if preferredNetworkAPIValue != self.lastCardBrandLogSelectedEventSent {
+                    STPAnalyticsClient.sharedClient.logPaymentSheetEvent(event: self.viewModel.hostedSurface.analyticEvent(for: .cardBrandSelected),
+                                                                         params: ["selected_card_brand": preferredNetworkAPIValue,
+                                                                                  "cbc_event_source": "edit", ])
+                    self.lastCardBrandLogSelectedEventSent = preferredNetworkAPIValue
```
yuki-stripe, line 48: Is there a retain cycle here? The Factory class owns the card form, which owns this closure, which owns `self` (the factory).

Do we need to do this work in the closure at all?  Is this not work we can do in `didUpdate`?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
```diff
@@ -97,10 +98,14 @@ final class UpdatePaymentMethodViewController: UIViewController {
         return button
     }()
 
+    private lazy var formFactory: SavedPaymentMethodFormFactory = {
+        let factory = SavedPaymentMethodFormFactory(viewModel: viewModel)
```
yuki-stripe, line 102: I'm confused by the modeling here. UpdatePMVC owns a ViewModel and a Factory. The factory _also_ owns the ViewModel. How should we think about this?  Of the 3 entities here, who is responsible for what?

porter-stripe: All the factory needs it the paymentMethod.type from the ViewModel so that definitely can be scoped down in terms of owning the view model.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/SavedPaymentMethodFormFactory/SavedPaymentMethodFormFactory.swift
porter-stripe: Created a task here: https://jira.corp.stripe.com/browse/MOBILESDK-3160

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/SavedPaymentMethodFormFactory/SavedPaymentMethodFormFactory+Card.swift
wooj-stripe: I booted up allocations and i'm not seeing any persistent allocations for SavedPaymentMethodFormFactory nor UpdatePaymentMethodViewModel.

Seems like the work should be done in here because it's updating the `params`.

In any case, seems like the factory should just create the instance and not own it. Will update.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/SavedPaymentMethodFormFactory/SavedPaymentMethodFormFactory.swift
wooj-stripe: I will handle this as part of this PR.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
wooj-stripe: It actually needs more than this, as sepa and us bank account need more. I will post a new revision shortly which turns the Factory to a class with one static function which creates PaymentMethodElements and has zero state.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/SavedPaymentMethodFormFactory/SavedPaymentMethodFormFactory+Card.swift
wooj-stripe: Made this class effectively a container for static methods, so this comment now seems out of date. Thanks for calling this out!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewModel.swift
```diff
@@ -82,4 +79,37 @@ class UpdatePaymentMethodViewModel {
...
+
+        if let params = paymentMethodElement.updateParams(params: confirmParams),
+           let cardParams = params.paymentMethodParams.card,
+           let originalPaymentMethodCard = paymentMethod.card,
+           hasChangedFields(original: originalPaymentMethodCard, updated: cardParams) {
+            return .card(paymentMethodCardParams: cardParams)
+        }
+        return nil
+    }
+    func hasChangedFields(original: STPPaymentMethodCard, updated: STPPaymentMethodCardParams) -> Bool {
```
yuki-stripe, line 94: This method name implies it merely calculates something (whether or not the fields have changed), which is a great single purpose method. I'm not sure it should also send a 'card brand selected' analytic - that seems like an unrelated responsibility. Why are these two distinct things combined?

I'd expect the function that handles 'card brand was selected' (something that handles the dropdown value being changed) to send the 'card brand selected' analytic, is that possible?

wooj-stripe: Agree. I initially had this in the factory, but since that was changed to a static function, I needed some way to keep track of what brand we've logged already (`lastCardBrandLogSelectedEventSent`).

In haste, I moved this functionality over, but you're totally right. So, to remedy this, I've added a function to the viewModel to take care of sending this metric.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
yuki-stripe: What's the purpose/responsibility of the ViewModel?  What belongs in there vs. UpdatePMVC?

wooj-stripe: I've renamed ViewModel to a struct of Configuration -- they are read-only/computed values that are passed into the UpdatePaymentMethodViewController.  Any state can be managed by the ViewController directly.

One issue i ran into was with setting the initial error state -- i've introduced a callback in the factory, as I couldn't figure out an elegant way to enumerate through each text element of the form from the ViewController.

PR #4598, opened 2025-02-24
Replace `las_client_secret` in logs with `las_id`

mats-stripe:
## Summary

Instead of logging the client secret, we should now log the link account session client secret. This replaces those parameters, passes along the link account session ID when it is available (i.e. after we have received the manifest).

## Motivation

This is in the spirit of least-privilege.

## Testing

N/a

## Changelog

N/a

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/FinancialConnectionsSheet.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/FinancialConnectionsSheet.swift
@@ -111,7 +111,7 @@ final public class FinancialConnectionsSheet {
     public let configuration: FinancialConnectionsSheet.Configuration
 
     /// Completion block called when the sheet is closed or fails to open
-    private var completion: ((HostControllerResult) -> Void)?
+    private var completion: ((HostControllerResult, String?) -> Void)?
 
     private var hostController: HostController?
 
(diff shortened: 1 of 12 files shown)


StripeFinancialConnections/StripeFinancialConnections/Source/FinancialConnectionsSheet.swift
```diff
@@ -111,7 +111,7 @@ final public class FinancialConnectionsSheet {
     public let configuration: FinancialConnectionsSheet.Configuration
 
     /// Completion block called when the sheet is closed or fails to open
-    private var completion: ((HostControllerResult) -> Void)?
+    private var completion: ((HostControllerResult, String?) -> Void)?
```
tillh-stripe, line 114: Should these be combined in a single struct?

tillh-stripe (dismissed):

StripeFinancialConnections/StripeFinancialConnections/Source/FinancialConnectionsSheet.swift
mats-stripe: Went with a named type `typealias HostControllerOutcome = (result: HostControllerResult, sessionId: String?)` here instead of a Struct as to not introduce a third result type in `FinancialConnectionsSheet`. https://github.com/stripe/stripe-ios/pull/4598/commits/4be91484d90650fb6fd7e216457dddf77f68a5a6

tillh-stripe (approved):

PR #4599, opened 2025-02-25
Error on confirm when using set as default feature + SSC

joyceqin-stripe:
## Summary

The set as default feature does not support deferred server-side confirmation, so we crash the app in test mode if a merchant tries to confirm with deferred ssc and set as default.
## Motivation

[MOBILESDK-3050](https://jira.corp.stripe.com/browse/MOBILESDK-3050)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+DeferredAPI.swift
```diff
@@ -167,4 +169,9 @@ extension PaymentSheet {
         }
         return paymentUserAgentValues
     }
+
+    private static func setAsDefaultErrorOnSSC(testMode: Bool, allowsSetAsDefaultPM: Bool) {
```
yuki-stripe, line 173: this name confuses me. The verb (what this method does) is "set as default", but I think this method validates something and should be called, like, `validateSomethingSomething`?

yuki-stripe: Or just inline this, not sure the method adds much value.  I'd also just combine the implementation into a single assert().

yuki-stripe (dismissed):

yuki-stripe (approved):

PR #4602, opened 2025-02-26
Update FC Example app to handle Instant Debits results

mats-stripe:
## Summary

This updates our example app to successfully handle the completion of the Instant Debits flow. We were previously treating it as a failure.

## Motivation

💅 ✨ 

## Testing

| Before | After |
|--------|--------|
| Simulator Screenshot - iPhone 16 - 2025-02-26 at 13 15 30 | Simulator Screenshot - iPhone 16 - 2025-02-26 at 13 16 17 |

## Changelog

N/a


tillh-stripe (approved):
❤️

PR #4603, opened 2025-02-26
Update US Bank Account payment method parsing

tillh-stripe:
## Summary

This pull request fixes an issue in the parsing logic of `STPPaymentMethodUSBankAccount`, which used the `linked_account` key instead of `financial_connections_account`.

We’re changing the key behind the scenes, but keeping the variable name to avoid a breaking change. We’ll rename the field in the next major version bump.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 * [Fixed] Fixed an issue in Embedded Payment Element (private beta) where the row could be selected even though there's no valid payment option.
 * [Added] Support for Crypto in PaymentSheet
 
+### Payments
+* [Fixed] Fixed an issue that caused `STPPaymentMethodUSBankAccount.linkedAccount` to be `nil` for some merchants.
+
 ## 24.6.0 2025-02-18
 * [Fixed] The SDK and example apps now compile in Xcode 16.2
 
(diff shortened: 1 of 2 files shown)


mats-stripe (approved):
Non-blocking: Thoughts on deprecating the `linkedAccount` property in this PR in anticipation of the upcoming breaking rename? Something like this should work:

```swift
@available(*, deprecated, message: "Use `financialConnectionsAccount` instead.", renamed: "financialConnectionsAccount")
public var linkedAccount: String {
  get { financialConnectionsAccount }
  set { financialConnectionsAccount = newValue }
}

public let financialConnectionsAccount: String
```

And in this scenario, you'd be assigning `response["financial_connections_account"]` to the `financialConnectionsAccount` property.

tillh-stripe:
@mats-stripe We decided not to do that, since lots of developers treat warnings as errors. But we will rename the field in the upcoming major version bump.

CHANGELOG.md
```diff
@@ -3,6 +3,9 @@
 * [Fixed] Fixed an issue in Embedded Payment Element (private beta) where the row could be selected even though there's no valid payment option.
 * [Added] Support for Crypto in PaymentSheet
 
+### Payments
+* [Fixed] Fixed an issue that caused `STPPaymentMethodUSBankAccount.linkedAccount` to be `nil` for some merchants.
```
mats-stripe, line 7: Just for my understanding, what is the context of "for some merchants" here? Is this issue only happening for merchants on certain SDK versions?

mats-stripe (approved):

CHANGELOG.md
tillh-stripe: Discussed in Slack.

PR #4608, opened 2025-02-27
Refactors TextFieldElement hasDisabledAppearance to allow for individual control of each field

wooj-stripe:
## Summary
Pre-work required for updating a payment method

## Motivation
We were previously setting the disabled appearance to the entire section, but we need to be able to control individual fields.

## Testing
Manually and relying on existing snapshot tests -- though, there are some severe limitations with snapshot tests. Darkmode tests no longer show the mask being applied to each field, unfortunately.


StripeUICore/StripeUICore/Source/Elements/TextField/TextFieldElementConfiguration.swift
```diff
@@ -40,6 +40,11 @@ import UIKit
      */
     var isEditable: Bool { get }
 
+    /**
+     * Adds a visible mask over the textFIeldView often used when isEditable == false
```
wooj-stripe, line 44: Need to update textFIeldView to textFieldView

StripeUICore/StripeUICore/Source/Elements/TextField/TextFieldElementConfiguration.swift
```diff
@@ -40,6 +40,11 @@ import UIKit
      */
     var isEditable: Bool { get }
 
+    /**
+     * Adds a visible mask over the textFieldView often used when isEditable == false
+     */
+
+    var hasDisabledAppearance: Bool { get }
```
yuki-stripe, line 47: What's the difference between 'disabled' and 'not editable' (isEditable = false)?

wooj-stripe: `disabledAppearance = true` -- Adds a grey transparent background signaling to a user they are unable to tap the field to edit.
`isEditable = false`  -- Doesn't allow a user to tap on a field to edit it.

The reason for this new field (rather than just apply a disabled appearance to `isEditable==false`) is due to the pan field

If the card is CBC enabled, the pan field will be where `isEditable=false && disabledAppearance=false` signaling they can update the CardBrand.

If the card is not cbc enabled, the pan field will be set to `isEditable = false && disabledAppearance=true` signaling there is nothing that can be updated.

yuki-stripe: It sounds like you're saying disabledAppearance controls whether or not the Card Brand is editable, is that right?


@porter-stripe Looking at the code, I'm not seeing why we need to use `isEditable` - is it not the same as setting `isUserInteractionEnabled = false`, which we can do by calling `sendEventToSubviews(.shouldDisableUserInteraction)`

wooj-stripe: > It sounds like you're saying disabledAppearance controls whether or not the Card Brand is editable, is that right? 

Yes, but we'll also apply disabledAppearance to MM/YY when updating cards.  E.g. when users are using ephemeral keys, disabledAppearance will equal true for MM/YY and disabledAppearance will equal false when using CustomerSessions.

StripeUICore/StripeUICore/Source/Elements/TextField/TextFieldView.swift
```diff
@@ -24,6 +24,13 @@ protocol TextFieldViewDelegate: AnyObject {
 class TextFieldView: UIView {
     weak var delegate: TextFieldViewDelegate?
     private lazy var toolbar = DoneButtonToolbar(delegate: self, theme: viewModel.theme)
+
+    lazy var transparentMaskView: UIView = {
```
joyceqin-stripe, line 28: Why does the resulting color differ so much in dark mode? It feels like in dark mode it doesn't look clear that the fields are disabled

joyceqin-stripe: When I set the Last4Configuration hasDisabledAppearance to false, I can't see the difference between it and the other fields in dark mode

wooj-stripe: Agree it is very subtle in dark mode (and also can vary depending on how the user configures the appearance api). The mask is an approximation based on the current component color.  It might be easier to see if you change the component color using the appearance api.  

It's possible that we revisit this in the future, but this pr is more or less a refactor which should preserve existing behavior.  Is the mask more prominent in `master` w/ dark mode?

StripeUICore/StripeUICore/Source/Elements/TextField/TextFieldElementConfiguration.swift
porter-stripe: Yeah I think `sendEventToSubviews(.shouldDisableUserInteraction)` can achieve the same, but I don't find it particularly intuitive/obvious to use. I don't feel strongly though, peaceful w/ removing isEditable in favor of  sendEventToSubviews.

Agree the name is confusing and easily conflated w/ disabled, I think we were going for some general idea like this when naming it: https://developer.apple.com/documentation/uikit/uitextview/iseditable

yuki-stripe: Is there a typo in this sentence:
> isEditable=false && disabledAppearance=false signaling they can update the CardBrand.

isEditable = false, but they can update the card brand? That's not making sense to me

wooj-stripe: > isEditable = false, but they can update the card brand? That's not making sense to me
Sorry, updated!

yuki-stripe: I'm still confused. In the updated example, you can just apply the disabled appearance when isEditable == false .  

I'm guessing there are 3 possible states?
1. Editable
2. Not editable
3. Not editable and disabled

Is that right?

wooj-stripe: The pan field is tricky since it contains the pan AND the controls for CardBrandChoice.

For the Pan field, when there is a CBC-Enabled card, the state will be Not Editable (so that users can't change the pan) & disabled appearance == false (so that users know they CAN update the card brand)

When there is a regular, non-CBC Enabled card, the state of this field will still be non-editable, but disabled appearance == true (because users can't tap the CBC Control).

In all other cases when isEditable == false (CVC, or something non-editable), disabledAppearance should equal true.

So, to get back to your original question, I think there are four states we need to support:
Editable = false && disabledAppearance = false (example: CBC enabled card -- pan field)
Editable = false && disabledAppearance = true (example: Non-cbc enabled card -- pan field)
Editable = true && disabledAppearance = false (example: all fields when add carding a card)
Editable = false && disabledAppearance = false (example: CVC on update page)

wooj-stripe: The pan field is tricky since it contains the pan AND the controls for CardBrandChoice.

For the Pan field, when there is a CBC-Enabled card, the state will be Not Editable (so that users can't change the pan) & disabled appearance == false (so that users know they CAN update the card brand)

When there is a regular, non-CBC Enabled card, the state of this field will still be non-editable, but disabled appearance == true (because users can't tap the CBC Control).

In all other cases when isEditable == false (CVC, or something non-editable), disabledAppearance should equal true.

So, to get back to your original question, I think there are 3 states we need to support:
Editable = false && disabledAppearance = false (example: CBC enabled card -- pan field)
Editable = false && disabledAppearance = true (example: Non-cbc enabled card -- pan field, also CVC on update page)
Editable = true && disabledAppearance = false (example: all fields when add carding a card)

This is not a case:
Editable = true && disabledAppearance = true 
This is because if we can edit a field, we shouldn't be applying a disabled appearance to it.

yuki-stripe: I _think_ this is a symptom of the bigger tech debt around CardBrandChoice. I have a ticket here that touches on it https://jira.corp.stripe.com/browse/MOBILESDK-3088

Talked live - peaceful with making an enum here so it's less confusing.

yuki-stripe (approved):

PR #4610, opened 2025-02-28
Update changelog to include connect updates.

nschris-stripe:
^


CHANGELOG.md
```diff
@@ -1,4 +1,8 @@
 ## X.Y.Z 2025-XX-YY
+### Connect (Private Preview)
+* [Changed] The SDK now manages the onboarding component's presentation.
+* [Changed] The payouts component is no longer available.
+
 ### PaymentSheet
 * [Fixed] Fixed an issue in Embedded Payment Element (private beta) where the row could be selected even though there's no valid payment option.
```
mats-stripe, line 7: While we're here, would it make sense to update this line to also say `private preview` instead of `private beta`? IIRC we're using the term `preview` instead of `beta now. (c.c. @yuki-stripe)

mats-stripe (dismissed):
LGTM, just one question.

CHANGELOG.md
nschris-stripe: Updated

mats-stripe (approved):

PR #4611, opened 2025-02-28
Proposing to separate local vs backend stored default

wooj-stripe:

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerAdapter/CustomerPaymentOption.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerAdapter/CustomerPaymentOption.swift
@@ -56,18 +56,22 @@ public enum CustomerPaymentOption: Equatable {
     /// - Parameter elementsSession: the ElementsSession.
     /// - Parameter surface: `.paymentSheet` or `.customerSheet`
     /// - Returns: Selected payment method.
-    @_spi(STP) public static func selectedPaymentMethod(for customerID: String?, elementsSession: STPElementsSession?, surface: HostedSurface?) -> CustomerPaymentOption? {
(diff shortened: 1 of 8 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerAdapter/CustomerPaymentOption.swift
```diff
@@ -56,18 +56,22 @@ public enum CustomerPaymentOption: Equatable {
...
+        case .paymentSheet:
+            if let paymentMethod = elementsSession.customer?.getDefaultOrFirstPaymentMethod() {
+                return CustomerPaymentOption.stripeId(paymentMethod.stripeId)
+            }
+        case .customerSheet:
+            if let paymentMethod = elementsSession.customer?.getDefaultPaymentMethod() {
+                return CustomerPaymentOption.stripeId(paymentMethod.stripeId)
+            }
+        }
+        return nil
```
wooj-stripe, line 71: This should probably not be part of this class.

wooj-stripe: I think i realize what the problem is -- if there is no backend default, then it defaults to the local store.

joyceqin-stripe: It doesn't default to local store— local store is only used when the set as default feature is disabled. When there's no backend default, in PaymentSheet, we use the first payment method in the list of saved payment methods. In CustomerSheet, we don't select anything.

joyceqin-stripe (changes requested):
I'm still not convinced that separating it is necessary, but I'm fine going this route instead. It seems simpler to me to just have the selectedPaymentMethod function than to have all the if statements, but I'm not dying on this hill

PR #4612, opened 2025-02-28
Allow updating of mm/yy behind client side flag

wooj-stripe:
## Summary
Allow updating of mm/yy behind client side flag

## Motivation
Update PM Project

## Testing
Added snapshot tests and e2e tests.

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
+++ b/Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
@@ -1640,7 +1640,7 @@ class PaymentSheetCustomerSessionDedupeUITests: PaymentSheetUITestCase {
             XCTFail("Unable to get expiraiton field")
             return
         }
-        XCTAssertEqual(expirationDate, "02/32")
+        XCTAssertEqual(expirationDate, "03/32")
     }
 }
 


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController+Configuration.swift
```diff
@@ -50,6 +51,9 @@ extension UpdatePaymentMethodViewController {
         var footnote: String? {
             switch paymentMethod.type {
             case .card:
+                if canUpdate {
```
porter-stripe, line 54: Nit: Guard feels better here

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController+Configuration.swift
```diff
@@ -16,12 +16,13 @@ extension UpdatePaymentMethodViewController {
...
         let canRemove: Bool
+        let canUpdate: Bool
         let isCBCEligible: Bool
         let canSetAsDefaultPM: Bool
         let isDefault: Bool
 
-        var canEdit: Bool {
-            return canUpdateCardBrand || canSetAsDefaultPM
+        var shouldShowSaveButton: Bool {
+            return canUpdateCardBrand || canSetAsDefaultPM || canUpdate
```
porter-stripe, line 25: Why are canUpdateCardBrand and canUpdate separate? They both make the same API call.

porter-stripe (commented):
Few comments nothing blocking

Tests/ReferenceImages_64/StripePaymentSheetTests.UpdatePaymentMethodViewControllerSnapshotTests/test_UpdatePaymentMethodViewControllerCanUpdateSetAsDefaultCard@3x.png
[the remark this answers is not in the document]
wooj-stripe: It's a limitation in the screenshot framework :(

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController+Configuration.swift
wooj-stripe: True, updating a card brand and updating mm/yy use the same api call, but not all cards are eligible for CBC.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController+Configuration.swift
wooj-stripe: I think i disagree.  A guard would result in:
```
guard !canUpdate else {
   return nil
}
return canUpdateCardBrand ? .Localized.only_card_brand_can_be_changed : .Localized.card_details_cannot_be_changed

```

The way this reads feels awkward because you are ensuring that you can not update, and then immediately read "can update card brand".

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController+Configuration.swift
joyceqin-stripe: I think it can be a little confusing to see canUpdate and canUpdateCardBrand, it's hard to understand the difference at a first glance.

wooj-stripe: I thought about this, but it seemed to align with `canRemove`, so moved forward with that. Maybe we change `canUpdateCardBrand` to `isCardBrandUpdatable` -- would that be clearer, or any other ideas?

porter-stripe: I see this makes sense now, I was conflating `canUpdate` with "has made changes that are updatable" not what it actually is (a flag from elements/session).

porter-stripe (approved):

PR #4614, opened 2025-03-01
Add alert for when we expect a 19 digit PAN, but a 16 digit one is entered

davidme-stripe:
## Summary
Add a (very specific) alert for the situation where a user enters a 16 digit PAN and switches fields, but we expected a >16-digit PAN. This will help us debug potential issues with our BIN data.

## Motivation
MOBILESDK-3217

## Testing
Added a test.

## Changelog
Not user-facing.


StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/TextFieldElement+CardTest.swift
```diff
@@ -336,6 +337,65 @@ class TextFieldElementCardTest: STPNetworkStubbingTestCase {
...
+        // A 16 digit card number that should be 19 digits:
+        let unionPay19_but_16_digits_entered = "6235510000000002"
+
+        // Cache the UnionPay BIN range first:
+        let allRetrievalsAreComplete = expectation(description: "Fetch BIN Range")
+        (textFieldElement.configuration as! TextFieldElement.PANConfiguration).binController.retrieveBINRanges(forPrefix: unionPay19_but_16_digits_entered) { _ in
+            allRetrievalsAreComplete.fulfill()
+        }
+        // Wait for the fetch to complete
+        waitForExpectations(timeout: 10, handler: nil)
```
wooj-stripe, line 378: Since we have two different expectations that should happen at two different times of the test, i think we should change the first to:
```
        wait(for: [allRetrievalsAreComplete], timeout: 10)
```

and the second to:
```
        wait(for: [logExpectation], timeout: 10)
```

StripeCore/StripeCore/Source/Analytics/STPAnalyticEvent.swift
```diff
@@ -44,6 +44,7 @@ import Foundation
     case cardMetadataLoadedTooSlow = "stripeios.card_metadata_loaded_too_slow"
     case cardMetadataResponseFailure = "stripeios.card_metadata_load_failure"
     case cardMetadataMissingRange = "stripeios.card_metadata_missing_range"
+    case cardMetadataExpectedExtraDigitsButUserEntered16 = "stripeios.card_metadata.expected_extra_digits_but_user_entered_16"
```
wooj-stripe, line 47: I don't feel strongly about it, but if we are being verbose, we could change this to..
card_metadata.expected_extra_digits_but_user_finished_entering_16

(meaning that they actually entered 16 and were done typing)

wooj-stripe (dismissed):

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/TextFieldElement+CardTest.swift
davidme-stripe: Agreed!

StripeCore/StripeCore/Source/Analytics/STPAnalyticEvent.swift
davidme-stripe: Good point, though that's confusing in a similar way — it could be interpreted as "the user finished entering 16 digits" without implying that they are not going to type any more. 😅 

If we want to be *really* verbose, maybe `expected_extra_digits_but_user_entered_16_then_switched_fields`?

PR #4622, opened 2025-03-03
Allow commits for release due to false positive on PII

wooj-stripe:
## Summary
Updates the release script to add `GIT_VALID_PII_OVERRIDE` to the message.  Unable to figure out how to use `--no-verify` on push, as the script breaks on commit.

## Motivation
Friction during releasing

## Testing
Manually tested during last release

### DIFF
--- a/ci_scripts/push_spm_mirror.rb
+++ b/ci_scripts/push_spm_mirror.rb
@@ -50,7 +50,7 @@ Dir.mktmpdir do |tmp_dir|
 
   Dir.chdir(tmp_dir + '/stripe-ios-spm') do
     run_command('git add .')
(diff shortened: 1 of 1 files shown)


davidme-stripe:
Can you add `--no-verify` to the `git commit` command? It should work, per the man page: https://git-scm.com/docs/git-commit

wooj-stripe:
> Can you add `--no-verify` to the `git commit` command? It should work, per the man page: https://git-scm.com/docs/git-commit

Sure, updated the PR, but didn't test it. I guess it's better than what we currently have, so i'm peaceful with it.

mats-stripe (approved):

PR #4623, opened 2025-03-04
Ensure analytics sent

joyceqin-stripe:
## Summary

Added analytics and params based on the [eng brief ](https://docs.google.com/document/d/19OB4Hlfqo-VBKu1ZxFT53hCs9OTUHgVHvdOdFRoSDI8/edit?tab=t.0#heading=h.e0xfe5cqe659)
Set as default analytics are now separate from update card brand analytics
## Testing

In a UI test, verify analytics sent


StripeCore/StripeCore/Source/Analytics/STPAnalyticEvent.swift
```diff
@@ -231,19 +231,23 @@ import Foundation
     case paymentSheetDisplayCardBrandDropdownIndicator = "mc_display_cbc_dropdown"
     case paymentSheetCardBrandSelected = "mc_cbc_selected"
     case paymentSheetOpenEditScreen = "mc_open_edit_screen"
-    case paymentSheetUpdateCard = "mc_update_card"
-    case paymentSheetUpdateCardFailed = "mc_update_card_failed"
+    case paymentSheetUpdateCardBrand = "mc_update_card"
+    case paymentSheetUpdateCardBrandFailed = "mc_update_card_failed"
```
porter-stripe, line 235: Why change the name? The raw string still is just "update_card"

StripeCore/StripeCore/Source/Analytics/STPAnalyticEvent.swift
```diff
@@ -231,19 +231,23 @@ import Foundation
...
     case paymentSheetDisallowedCardBrand = "mc_disallowed_card_brand"
 
     // MARK: - CustomerSheet card brand choice
     case customerSheetDisplayCardBrandDropdownIndicator = "cs_display_cbc_dropdown"
     case customerSheetCardBrandSelected = "cs_cbc_selected"
     case customerSheetOpenEditScreen = "cs_open_edit_screen"
-    case customerSheetUpdateCard = "cs_update_card"
-    case customerSheetUpdateCardFailed = "cs_update_card_failed"
+    case customerSheetUpdateCardBrand = "cs_update_card"
+    case customerSheetUpdateCardBrandFailed = "cs_update_card_failed"
```
porter-stripe, line 244: Same question here on the rename.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetAnalyticsHelperTest.swift
```diff
@@ -171,6 +171,125 @@ final class PaymentSheetAnalyticsHelperTest: XCTestCase {
         }
     }
 
+    func testLogLoadSucceededSetAsDefaultEnabled() {
```
porter-stripe, line 174: Do we need a whole new test for this? It seems mostly copy pasted from testLogLoadSucceeded. Can we just update that test?

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetAnalyticsHelperTest.swift
```diff
@@ -171,6 +171,125 @@ final class PaymentSheetAnalyticsHelperTest: XCTestCase {
...
+            XCTAssertEqual(loadSucceededPayload["selected_lpm"] as? String, "none")
+            XCTAssertEqual(loadSucceededPayload["intent_type"] as? String, "payment_intent")
+            XCTAssertEqual(loadSucceededPayload["ordered_lpms"] as? String, "card")
+            XCTAssertEqual(loadSucceededPayload["integration_shape"] as? String, shapeString)
+            XCTAssertEqual(loadSucceededPayload["set_as_default_enabled"] as? Bool, true)
+            XCTAssertEqual(loadSucceededPayload["has_default_payment_method"] as? Bool, false)
+        }
+    }
+
+    func testLogLoadSucceededSetAsDefaultEnabledHasDefaultPaymentMethod() {
```
porter-stripe, line 211: Same, can we just update testLogLoadSucceeded?

StripePaymentSheet/StripePaymentSheet/Source/Analytics/PaymentSheetAnalyticsHelper.swift
```diff
@@ -133,6 +133,13 @@ final class PaymentSheetAnalyticsHelper {
             let linkMode: String = elementsSession.linkPassthroughModeEnabled ? "passthrough" : "payment_method_mode"
             params["link_mode"] = linkMode
         }
+        if elementsSession.customer?.customerSession != nil {
```
porter-stripe, line 136: @wooj-stripe how do we handle customer session analytics today? Why don't we just log the whole thing or have it define an "analyticPayload", similar to PaymentSheet.Configuration.

StripePaymentSheet/StripePaymentSheet/Source/Analytics/PaymentSheetAnalyticsHelper.swift
```diff
@@ -133,6 +133,13 @@ final class PaymentSheetAnalyticsHelper {
...
+            let setAsDefaultEnabled = elementsSession.paymentMethodSetAsDefaultForPaymentSheet
+            params["set_as_default_enabled"] = setAsDefaultEnabled
+            if setAsDefaultEnabled {
+                params["has_default_payment_method"] = elementsSession.customer?.defaultPaymentMethod != nil
```
porter-stripe, line 140: Do we only want to send these values if customer session is being used? Or is it valid to send them if it's not being used? (they'd just be false)

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
```diff
@@ -218,25 +218,31 @@ final class UpdatePaymentMethodViewController: UIViewController {
...
+                STPAnalyticsClient.sharedClient.logPaymentSheetEvent(event: configuration.hostedSurface.analyticEvent(for: .setDefaultPaymentMethod),
+                                                                     params: ["payment_method_type": configuration.paymentMethod.type.identifier])
+            }
+        } catch let error as NSError {
             updateButton.update(state: .enabled)
             latestError = error
-            STPAnalyticsClient.sharedClient.logPaymentSheetEvent(event: configuration.hostedSurface.analyticEvent(for: .updateCardFailed),
-                                                                 error: error,
-                                                                 params: analyticsParams)
+            if case NSError.stp_cardBrandNotUpdatedError() = error,
```
joyceqin-stripe, line 235: I'm not convinced this works unless in the API I do continuation.resume(with: .failure(NSError.stp_cardBrandNotUpdatedError())) instead of continuation.resume(with: .failure(error ?? NSError.stp_cardBrandNotUpdatedError()))— any ideas of how better to do this?

StripePayments/StripePayments/Source/API Bindings/STPAPIClient+Payments.swift
```diff
@@ -887,7 +887,11 @@ extension STPAPIClient {
                 if let paymentMethod = paymentMethod {
                     continuation.resume(with: .success(paymentMethod))
                 } else {
-                    continuation.resume(with: .failure(error ?? NSError.stp_genericFailedToParseResponseError()))
+                    if paymentMethodUpdateParams.card?.networks?.preferred != nil {
+                        continuation.resume(with: .failure(error ?? NSError.stp_cardBrandNotUpdatedError()))
```
porter-stripe, line 891: If we want a specific error about updating the payment method let's just add a generic payment method update failed error.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
porter-stripe, line 235: Why do we need a specific error for just updating brand failed? How will this scale if you modify multiple things and the update fails? What event will we send and what error will we throw? I think we should replace stp_cardBrandNotUpdatedError with a generic payment method update failed event.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
```diff
@@ -218,25 +218,31 @@ final class UpdatePaymentMethodViewController: UIViewController {
...
+               case .card(let paymentMethodCardParams) = updatePaymentMethodOptions,
+               let selectedCardBrand = paymentMethodCardParams.networks?.preferred {
+                STPAnalyticsClient.sharedClient.logPaymentSheetEvent(event: configuration.hostedSurface.analyticEvent(for: .updateCardBrandFailed),
+                                                                     params: ["selected_card_brand": selectedCardBrand])
+            }
+            if case NSError.stp_defaultPaymentMethodNotUpdatedError() = error,
+               setAsDefaultValue ?? false {
```
porter-stripe, line 242: Seems like we should always log this error analytic regardless of the value of "setAsDefaultValue" if stp_defaultPaymentMethodNotUpdatedError is thrown.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
```diff
@@ -377,8 +377,12 @@ extension PaymentSheetFormFactory {
             isSelectedByDefault: false,
             didToggle: didToggle
         )
-        return PaymentMethodElementWrapper(element) { defaultCheckbox, params in
-            params.setAsDefaultPM = defaultCheckbox.isSelected
+        return PaymentMethodElementWrapper(element) { checkbox, params in
+            if checkbox.checkboxButton.isHidden {
```
porter-stripe, line 381: Is there ever a time the checkbox becomes "unhidden" or is it always either hidden or shown?

StripePayments/StripePayments/Source/API Bindings/STPAPIClient+Payments.swift
porter-stripe, line 891: I think this should be "continuation.resume(throwing:".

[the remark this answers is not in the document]
porter-stripe: To do a quick smoke test you can modify the requests to make them fail somehow, maybe an extra parameter or something.

StripePayments/StripePayments/Source/API Bindings/STPAPIClient+Payments.swift
```diff
@@ -887,7 +887,11 @@ extension STPAPIClient {
                 if let paymentMethod = paymentMethod {
                     continuation.resume(with: .success(paymentMethod))
                 } else {
-                    continuation.resume(with: .failure(error ?? NSError.stp_genericFailedToParseResponseError()))
+                    if paymentMethodUpdateParams.card?.networks?.preferred != nil {
```
porter-stripe, line 890: Hm, something feels off that this function in the API client is trying to infer what kind of update was performed. It shouldn't care, it just takes the params and tries to update. Maybe this should be handled in the UpdatePaymentMethodViewController.

StripePaymentSheet/StripePaymentSheet/Source/Analytics/PaymentSheetAnalyticsHelper.swift
joyceqin-stripe: According to the eng brief, we only send if customer session is used

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
joyceqin-stripe: I was basing it off of this doc about error cases with updating card brand and default: https://docs.google.com/document/d/1sbqluGazKQaZsPcx4zWiPwDNhD7SDcUhYWOXTv-08nU/edit?tab=t.0

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
joyceqin-stripe: What do you mean by "unhidden"? It starts off hidden, it is shown if the save checkbox is checked, and it gets hidden again if the save checkbox is unchecked.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetAnalyticsHelperTest.swift
joyceqin-stripe: The elements session loaded is very different from testLogLoadSucceeded— the elements session in testLogLoadSucceeded doesn't enable customer session features. I suppose I could update testLogLoadSucceeded to enable customer session features.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetAnalyticsHelperTest.swift
joyceqin-stripe: In this test, not only are customer session features and set as default enabled, but also its elements session has a default payment method. The previous one has set as default enabled but no default payment method.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetAnalyticsHelperTest.swift
porter-stripe: Yeah let's update testLogLoadSucceeded to have custom session features.

StripePayments/StripePayments/Source/API Bindings/STPAPIClient+Payments.swift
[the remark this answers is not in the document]
joyceqin-stripe: It looks like when I do that, the error that returns is "There was an unexpected error -- try again in a few seconds". It doesn't seem right to overwrite the error, but how would we show a custom error?

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetAnalyticsHelperTest.swift
porter-stripe: I think it would be sufficient to only test when they have a default payment method, so we could consolidate these two new tests into updating the testLogLoadSucceeded.

Probably don't need to test the case of no default b/c we have other tests that assert on the size of the payload.

StripePayments/StripePayments/Source/API Bindings/STPAPIClient+Payments.swift
```diff
@@ -1439,7 +1439,7 @@ extension STPAPIClient {
         try await withCheckedThrowingContinuation({ continuation in
             self.setAsDefaultPaymentMethod(paymentMethodID, for: customerID, using: ephemeralKey) { customer, error in
                 guard let customer = customer else {
-                    continuation.resume(throwing: error ?? NSError.stp_defaultPaymentMethodNotChangedError())
+                    continuation.resume(throwing: error ?? NSError.stp_defaultPaymentMethodNotUpdatedError())
```
joyceqin-stripe, line 1442: Should I just do continuation.resume(throwing: NSError.stp_defaultPaymentMethodNotUpdatedError()) here? Is there any reason to keep the original error?

StripePayments/StripePayments/Source/API Bindings/STPAPIClient+Payments.swift
joyceqin-stripe: I've moved it into the UpdatePaymentMethodViewControllerDelegates

joyceqin-stripe: Okay actually I encountered complications due to concurrency, and I'm not sure how to check if both calls failed or just one. I'm using actors to collect the errors from the taskgroups, but it feels like a wrong approach? Not sure

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -236,19 +236,46 @@ extension EmbeddedPaymentElement: UpdatePaymentMethodViewControllerDelegate {
 
     func didUpdate(viewController: UpdatePaymentMethodViewController,
                    paymentMethod: StripePayments.STPPaymentMethod) async throws {
-        try await withThrowingTaskGroup(of: Void.self) { group in
+        actor ErrorCollector {
```
porter-stripe, line 239: This seems to be gettin really complex, and I can see why it's tricky! I think there are two possible valid approaches here.

One: Make updateCardBrand, updateDefault, and UpdatePaymentMethodViewControllerDelegate.didUpdate return a `Result` instead of throwing an error. https://developer.apple.com/documentation/swift/result

Here is a rough example: 

```swift
func func didUpdate(...) async -> Result<Void, [Error]> {
    var errors: [Error] = []
    
    let firstResult = await updateCardBrand()
    let secondResult = await updateDefault()

    if case .failure(let error) = firstResult {
        errors.append(error)
    }
    
    if case .failure(let error) = secondResult {
        errors.append(error)
    }
    
    if errors.isEmpty {
        return .success(())
    } else {
        return .failure(errors) // Handle multiple or one error in UpdatePaymentMethodViewControllerDelegate.didUpdate so the errors can all be handled in one spot
    }
}
```

Another approach may be something like:

```swift
func didUpdate(...) async throws {
    let (firstError, secondError) = await (
        Task {
            do {
                try await updateCardBrand()
                return nil
            } catch {
                return error
            }
        }.value,
        Task {
            do {
                try await updateDefault()
                return nil
            } catch {
                return error
            }
        }.value
    )
    
    let errors = [firstError, secondError].compactMap { $0 }
    
    // Handle errors
}
```

My preference is for the first approach since it let's us
1. Handle all the errors/analytics in a central spot in UpdatePaymentMethodViewController
2. Will probably let us remove the TaskGroups in other didUpdate(...) implementations

Let me know what you think! If we go route one, we should probably do that refactor work in a separate PR to this one and land it before this one.

StripePaymentSheet/StripePaymentSheet/Source/Analytics/PaymentSheetAnalyticsHelper.swift
wooj-stripe: Currently we only send `customer_access_provider` which describes `ek` vs `cs` - but agree -- we can send the deserialized values of each feature as well!

StripePaymentSheet/StripePaymentSheet/Source/Analytics/STPAnalyticsClient+CustomerSheet.swift
```diff
@@ -15,15 +15,25 @@ extension STPAnalyticsClient {
     }
 
     // PM selection & Confirmation
-    func logCSSelectPaymentMethodScreenConfirmedSavedPMSuccess(type: String?) {
+    func logCSSelectPaymentMethodScreenConfirmedSavedPMSuccess(type: String?, syncDefaultEnabled: Bool? = nil) {
         let paymentMethodType = type ?? "unknown"
+        var params: [String: Any] = [:]
+        params["payment_method_type"] = paymentMethodType
```
porter-stripe, line 21: Nit: Can put this in one line `var params: [String: Any] = ["payment_method_type": paymentMethodType]`, same with below.

StripePaymentSheet/StripePaymentSheet/Source/Analytics/STPAnalyticsClient+CustomerSheet.swift
```diff
@@ -15,15 +15,25 @@ extension STPAnalyticsClient {
     }
 
     // PM selection & Confirmation
-    func logCSSelectPaymentMethodScreenConfirmedSavedPMSuccess(type: String?) {
+    func logCSSelectPaymentMethodScreenConfirmedSavedPMSuccess(type: String?, syncDefaultEnabled: Bool? = nil) {
```
porter-stripe, line 18: If possible, I think it's generally best to avoid optional Bools, it is typically a code smell. It creates an additional state, that introduces ambiguity. 

An optional boolean has three possible states:

true
false
nil
This creates ambiguity in your code logic. What does nil actually mean in your context? Is it "unknown," "not applicable," or something else?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSavedPaymentMethodsViewController.swift
```diff
@@ -430,12 +429,16 @@ class CustomerSavedPaymentMethodsViewController: UIViewController {
                 case .saved(let paymentMethod, _):
                     let paymentOptionSelection = CustomerSheet.PaymentOptionSelection.paymentMethod(paymentMethod)
                     let type = STPPaymentMethod.string(from: paymentMethod.type)
+                    var syncDefaultEnabled: Bool?
```
porter-stripe, line 432: Going w/ my comment above, let's just make it default to false

porter-stripe: Just an idea:
```swift
let syncDefaultEnabled: Bool = {
    if case .customerSession = self.customerSheetDataSource.dataSource {
        return self.paymentMethodSyncDefault
    }
        return false
}()
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/CardSection/HostedSurface.swift
```diff
@@ -48,6 +52,10 @@ import Foundation
             return .customerSheetUpdateCardFailed
         case (.closeEditScreen, .customerSheet):
             return .customerSheetClosesEditScreen
+        case (.setDefaultPaymentMethod, .customerSheet):
+            return STPAnalyticEvent.unexpectedCustomerSheetError
+        case (.setDefaultPaymentMethodFailed, .customerSheet):
+            return STPAnalyticEvent.unexpectedCustomerSheetError
```
porter-stripe, line 58: I see we have paymentSheetSetDefaultPaymentMethod and paymentSheetSetDefaultPaymentMethodFailed. Why do we have PaymentSheet events for these failures but not CustomerSheet and instead use an unexpected generic error?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
```diff
@@ -235,14 +235,31 @@ final class UpdatePaymentMethodViewController: UIViewController {
...
         switch updatePaymentMethodResult {
         case .success:
-            STPAnalyticsClient.sharedClient.logPaymentSheetEvent(event: configuration.hostedSurface.analyticEvent(for: .updateCard),
-                                                                 params: analyticsParams)
+            if case .card(let paymentMethodCardParams) = updatePaymentMethodOptions,
+               let selectedCardBrand = paymentMethodCardParams.networks?.preferred {
+                STPAnalyticsClient.sharedClient.logPaymentSheetEvent(event: configuration.hostedSurface.analyticEvent(for: .updateCard),
+                                                                     params: ["selected_card_brand": selectedCardBrand])
+            }
+            if setAsDefaultValue == true {
```
porter-stripe, line 243: Nit: Should usually avoid using `==` when comparing a boolean, so just `if setAsDefaultValue`. Also could we give setAsDefaultValue a better name? It's a little vague and not consistent with Swift API design guidelines.

> Uses of Boolean methods and properties should read as assertions about the receiver when the use is nonmutating, e.g. x.isEmpty, line1.intersects(line2). 
https://www.swift.org/documentation/api-design-guidelines/

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
```diff
@@ -235,14 +235,31 @@ final class UpdatePaymentMethodViewController: UIViewController {
...
+                STPAnalyticsClient.sharedClient.logPaymentSheetEvent(event: configuration.hostedSurface.analyticEvent(for: .setDefaultPaymentMethod),
+                                                                     params: ["payment_method_type": configuration.paymentMethod.type.identifier])
+            }
         case .failure(let errors):
             updateButton.update(state: .enabled)
             latestError = errors.count == 1 ? errors[0] : NSError.stp_genericErrorOccurredError()
-            STPAnalyticsClient.sharedClient.logPaymentSheetEvent(event: configuration.hostedSurface.analyticEvent(for: .updateCardFailed),
-                                                                 error: latestError,
-                                                                 params: analyticsParams)
+            if errors.contains(where: { error in error as NSError == NSError.stp_cardBrandNotUpdatedError() }) {
```
porter-stripe, line 250: Super super nit, can we put parenthesis around the casting, I think it will read easier. Same w/ below.
`error in (error as NSError) == NSError.stp_cardBrandNotUpdatedError()`

Another super nit: I find it hard to read b/c there are so many instances of the same word "error", it is a little hard for me to follow. An alternative way to write this is:

`if errors.contains(where: { ($0 as NSError) == NSError.stp_cardBrandNotUpdatedError() })`

StripePaymentSheet/StripePaymentSheet/Source/Analytics/STPAnalyticsClient+CustomerSheet.swift
joyceqin-stripe: The nil case in this Optional Bool represents that CustomerSession is not in use. We don't want it to default to false because this metric can be used to track feature adoption— we don't want to include those not using CustomerSession. This will allow us to see, of those using CustomerSession, how many have enabled the feature and how many have not.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/CardSection/HostedSurface.swift
joyceqin-stripe: On CustomerSheet, syncing default coincides with its confirm event. We are not sending an event for setDefaultPaymentMethod on the surface CustomerSheet, so these events never actually happen. We send this data on the CustomerSheet confirm event instead.

porter-stripe (approved):

PR #4627, opened 2025-03-04
Add draft CPM API

porter-stripe:
## Summary
- Adds a first draft of the CPM public API
- This API is not final and still needs API review, but adding this so we can easily test and make progress on the internal implementation of CPMs while in API review.

## Motivation
- Start on CPMs

## Testing
N/A

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
@@ -446,6 +446,7 @@ extension EmbeddedPaymentElement {
     public typealias Address = PaymentSheet.Address
     public typealias BillingDetailsCollectionConfiguration = PaymentSheet.BillingDetailsCollectionConfiguration
     public typealias ExternalPaymentMethodConfiguration = PaymentSheet.ExternalPaymentMethodConfiguration
+    public typealias CustomPaymentMethodConfiguration = PaymentSheet.CustomPaymentMethodConfiguration
 }
 
 // MARK: - EmbeddedPaymentElement.PaymentOptionDisplayData
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
@@ -183,6 +183,9 @@ extension PaymentSheet {
         /// Configuration for external payment methods.
         public var externalPaymentMethodConfiguration: ExternalPaymentMethodConfiguration?
 
+        /// Configuration for custom payment methods.
+        @_spi(CustomPaymentMethodsBeta) public var customPaymentMethodConfiguration: CustomPaymentMethodConfiguration?
+
         /// By default, PaymentSheet will use a dynamic ordering that optimizes payment method display for the customer.
         /// You can override the default order in which payment methods are displayed in PaymentSheet with a list of payment method types.
         /// See https://stripe.com/docs/api/payment_methods/object#payment_method_object-type for the list of valid types.  You may also pass external payment methods.
@@ -507,6 +510,59 @@ extension PaymentSheet {
         /// Your implementation should complete the payment and call the `completion` parameter with the result.
         /// - Note: This is always called on the main thread.
         public var externalPaymentMethodConfirmHandler: ExternalPaymentMethodConfirmHandler
+    }
+
+    /// Configuration for custom payment methods
+    @_spi(CustomPaymentMethodsBeta) public struct CustomPaymentMethodConfiguration {
+
+        /// Defines a custom payment method type that can be displayed in PaymentSheet
+        public struct CustomPaymentMethod {
+
+            /// The unique identifier for this custom payment method type in the format of "cmpt_..."
+            /// Obtained from the Stripe Dashboard at https://dashboard.stripe.com/settings/custom_payment_methods
+            let id: String
+
+            /// Optional subcopy text to be displayed below the custom payment method's display name.
+            let subcopy: String?
+
+            /// When true, PaymentSheet will collect billing details for this custom payment method type
+            /// in accordance with the `billingDetailsCollectionConfiguration` settings.
+            /// This has no effect if `billingDetailsCollectionConfiguration` is not configured.
+            var shouldCollectBillingDetails = false
+
+            /// Initializes an `CustomPaymentMethod`
+            /// - Parameters:
+            ///   - id: The unique identifier for this custom payment method type in the format of "cmpt_..."
+            ///   - subcopy: Optional subcopy text to be displayed below the custom payment method's display name.
+            public init(id: String, subcopy: String? = nil) {
+                self.id = id
+                self.subcopy = subcopy
+            }
+        }
+
+        /// Initializes an `CustomPaymentMethodConfiguration`
+        /// - Parameter customPaymentMethods: A list of custom payment methods to display in PaymentSheet.
+        /// - Parameter customPaymentMethodConfirmHandler: A handler called when the customer confirms the payment using a custom payment method.
+        public init(customPaymentMethods: [CustomPaymentMethod], customPaymentMethodConfirmHandler: @escaping PaymentSheet.CustomPaymentMethodConfiguration.CustomPaymentMethodConfirmHandler) {
+            self.customPaymentMethods = customPaymentMethods
+            self.customPaymentMethodConfirmHandler = customPaymentMethodConfirmHandler
+        }
+
+        /// A list of custom payment methods to display in PaymentSheet.
+        public var customPaymentMethods: [CustomPaymentMethod] = []
+
+        /// - Parameter customPaymentMethodType: The custom payment method to confirm payment with e.g., "cmpt_xxx"
+        /// - Parameter billingDetails: An object containing any billing details you've configured PaymentSheet to collect.
+        /// - Returns: The result of the attempt to confirm payment using the given custom payment method.
+        public typealias CustomPaymentMethodConfirmHandler = (
+            _ customPaymentMethodType: String,
+            _ billingDetails: STPPaymentMethodBillingDetails
+        ) async -> PaymentSheetResult
+
+        /// This handler is called when the customer confirms the payment using an custom payment method.
+        /// Your implementation should complete the payment and return the result.
+        /// - Note: This is always called on the main thread.
+        public var customPaymentMethodConfirmHandler: CustomPaymentMethodConfirmHandler
     }
 }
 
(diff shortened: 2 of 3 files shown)


joyceqin-stripe (commented):
It looks fine to me, but I don't really have much context about custom payment methods— how does it differ from external payment methods or lpms? Also, I'm not sure what a subcopy is or how it differs from like a subtitle

porter-stripe:
@joyceqin-stripe good point! I should have added some docs for context. Here they are:
https://docs.google.com/document/d/1tvTZrjKc25omcol5gOKvovmSV-PTVBK8KBXRTXLYQZE/edit?tab=t.0#heading=h.hoekn8ut09uh

https://docs.google.com/document/d/1Ilxh1sTM34jWGP_u-5Ig9odt8iFnH9jnwcdoEMcZw1w/edit?tab=t.0#heading=h.cz1xkpga7giy

These will be helpful in understanding future code reviews.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
```diff
@@ -508,6 +511,59 @@ extension PaymentSheet {
         /// - Note: This is always called on the main thread.
         public var externalPaymentMethodConfirmHandler: ExternalPaymentMethodConfirmHandler
     }
+
+    /// Configuration for custom payment methods
+    @_spi(CustomPaymentMethodsBeta) public struct CustomPaymentMethodConfiguration {
+
+        /// Defines a custom payment method type that can be displayed in PaymentSheet
+        public struct CustomPaymentMethod {
```
wooj-stripe, line 519: I think the name should be `CustomPaymentMethodType` since we typically refer to instances of a "payment method type" as a "payment method"

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
```diff
@@ -508,6 +511,59 @@ extension PaymentSheet {
...
+
+            /// The unique identifier for this custom payment method type in the format of "cmpt_..."
+            /// Obtained from the Stripe Dashboard at https://dashboard.stripe.com/settings/custom_payment_methods
+            let id: String
+
+            /// Optional subcopy text to be displayed below the custom payment method's display name.
+            let subcopy: String?
+
```
wooj-stripe, line 527: I'm guessing the display name comes from the backend, so I'm curious why the sub copy isn't modeled there as well?  Also, is it awkward that they will control the text, but we control the font/layout, etc. ?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
```diff
@@ -508,6 +511,59 @@ extension PaymentSheet {
...
+            /// When true, PaymentSheet will collect billing details for this custom payment method type
+            /// in accordance with the `billingDetailsCollectionConfiguration` settings.
+            /// This has no effect if `billingDetailsCollectionConfiguration` is not configured.
+            var shouldCollectBillingDetails = false
```
wooj-stripe, line 531: Instead of a bool, maybe:
```
        public var billingDetailsCollectionConfiguration = BillingDetailsCollectionConfiguration(name: .never, phone: .never, email: .never, address: .never)
```

That way it defaults to never, but, users can turn on billing address for cards but turn this off for this specific custom payment method?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
```diff
@@ -508,6 +511,59 @@ extension PaymentSheet {
...
+            public init(id: String, subcopy: String? = nil) {
+                self.id = id
+                self.subcopy = subcopy
+            }
+        }
+
+        /// Initializes an `CustomPaymentMethodConfiguration`
+        /// - Parameter customPaymentMethods: A list of custom payment methods to display in PaymentSheet.
+        /// - Parameter customPaymentMethodConfirmHandler: A handler called when the customer confirms the payment using a custom payment method.
+        public init(customPaymentMethods: [CustomPaymentMethod], customPaymentMethodConfirmHandler: @escaping PaymentSheet.CustomPaymentMethodConfiguration.CustomPaymentMethodConfirmHandler) {
```
wooj-stripe, line 546: if we change `CustomPaymentMethod` to `CustomPaymentMethodType` then the argument should change too.

wooj-stripe: One confirmHandler for all customPaymentMethods --  If we expect all customHandlers to effectively be the same thing, this seems like the right call, but if merchants are expected to write very specific code for each payment method, maybe we add this to the CustomPaymentMethod[Type] struct

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
```diff
@@ -508,6 +511,59 @@ extension PaymentSheet {
...
+            self.customPaymentMethods = customPaymentMethods
+            self.customPaymentMethodConfirmHandler = customPaymentMethodConfirmHandler
+        }
+
+        /// A list of custom payment methods to display in PaymentSheet.
+        public var customPaymentMethods: [CustomPaymentMethod] = []
+
```
wooj-stripe, line 553: Seems like an ordered list is what we want, but we probably want to (somewhere) assert on uniqueness of the arguments passed in.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
```diff
@@ -508,6 +511,59 @@ extension PaymentSheet {
...
+        /// - Parameter customPaymentMethodType: The custom payment method to confirm payment with e.g., "cmpt_xxx"
+        /// - Parameter billingDetails: An object containing any billing details you've configured PaymentSheet to collect.
+        /// - Returns: The result of the attempt to confirm payment using the given custom payment method.
+        public typealias CustomPaymentMethodConfirmHandler = (
+            _ customPaymentMethodType: String,
```
wooj-stripe, line 558: Why wouldn't we just pass the `CustomPaymentMethod` here?  Also the name `customPaymentMethodType` is confusing here since `CustomPaymentMethod.id` holds the same value.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
```diff
@@ -508,6 +511,59 @@ extension PaymentSheet {
...
+            _ billingDetails: STPPaymentMethodBillingDetails
+        ) async -> PaymentSheetResult
```
wooj-stripe, line 560: should we typealias PaymentSheetResult? (e.g. PaymentElementResult)

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
porter-stripe: Subcopy may need to be dynamic, e.g. a BNPL "Pay in X payments of Y"

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
porter-stripe: Good point. We'll filter out dupes eventually just not in this PR

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
porter-stripe: Yeah, agree!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
porter-stripe: I kinda prefer the bool, I'll let this get hashed out in API review

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
porter-stripe: Merchants can inspect the ID of the CPM to determine their flow. This matches EPMs, but fair point!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
porter-stripe: Yeah I think in the future that this is a good idea, just not specific to CPMs.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
porter-stripe: Fair! Updated to be the entire CPM

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
wooj-stripe: Assuming this will be a client side parameter, subcopy is feels vague as it has no attachment to the position of  display name -- maybe this is intentional as it gives us freedom to move this text around?  I would have expected the name to be something like `subDisplayNameCopy` or `subDisplayNameText`.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
```diff
@@ -446,6 +446,7 @@ extension EmbeddedPaymentElement {
     public typealias Address = PaymentSheet.Address
     public typealias BillingDetailsCollectionConfiguration = PaymentSheet.BillingDetailsCollectionConfiguration
     public typealias ExternalPaymentMethodConfiguration = PaymentSheet.ExternalPaymentMethodConfiguration
+    public typealias CustomPaymentMethodConfiguration = PaymentSheet.CustomPaymentMethodConfiguration
```
wooj-stripe, line 449: Does this need to be behind `@_spi(CustomPaymentMethodsBeta)`

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
porter-stripe: Subcopy is currently the same as the web API parameter name, happy to discuss specific naming and things in an API review if you want to be involved.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
porter-stripe: Good thought, not needed!

wooj-stripe (approved):

PR #4631, opened 2025-03-05
Delete unused Basic Integration analytics

yuki-stripe:
## Motivation
https://jira.corp.stripe.com/browse/RUN_MOBILESDK-3880

## Testing
n/a

## Changelog
not user facing


sfriedman-stripe (approved):
LGTM, but did Periphery not catch this? cc @porter-stripe @joyceqin-stripe

porter-stripe:
@sfriedman-stripe It looks like this is a limitation of Periphery. These enums are case's on part of an enum which have an associated value (String). According to Periphery's docs these cannot be reliably analyzed to determine if they are used or not. You can see an example of why not in their docs.

[biPaymentCompleteSavedPMSuccess](https://github.com/peripheryapp/periphery?tab=readme-ov-file#enumerations)

PR #4633, opened 2025-03-05
Add CPM toggle to playground

porter-stripe:
## Summary
- Adds a toggle to the playground to let us test CPMs
- This currently doesn't do anything

## Motivation
- CPMs!

## Testing
- Manual

## Changelog
N/A

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
@@ -229,6 +229,10 @@ struct PaymentSheetTestPlayground: View {
             if playgroundController.settings.merchantCountryCode.rawValue != newCountry.rawValue {
                 playgroundController.settings.customerMode = .guest
             }
+            // Disable CPMs if we switch to non-US merchant
+            if newCountry != .US {
+                playgroundController.settings.customPaymentMethods = .off
+            }
             playgroundController.settings.merchantCountryCode = newCountry
         }
     }
--- a/Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
@@ -44,6 +44,10 @@ struct PaymentSheetTestPlayground: View {
         }
         SettingView(setting: $playgroundController.settings.userOverrideCountry)
         SettingView(setting: $playgroundController.settings.externalPaymentMethods)
+        // The hardcoded CPM id is only available on our US merchant
+        if playgroundController.settings.merchantCountryCode == .US {
+            SettingView(setting: $playgroundController.settings.customPaymentMethods)
+        }
         SettingView(setting: $playgroundController.settings.preferredNetworksEnabled)
         SettingView(setting: $playgroundController.settings.cardBrandAcceptance)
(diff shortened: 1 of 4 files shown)


Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
```diff
@@ -368,6 +370,29 @@ class PlaygroundController: ObservableObject {
         }
     }
 
+    var customPaymentMethodConfiguration: PaymentSheet.CustomPaymentMethodConfiguration? {
+        switch settings.customPaymentMethods {
+        case .on:
+            // Obtained from https://dashboard.stripe.com/settings/custom_payment_methods
+            let customPaymentMethodType = PaymentSheet.CustomPaymentMethodConfiguration.CustomPaymentMethodType(id: "cpmt_1QpId5Lu5o3P18ZpLwSqMXws",
```
wooj-stripe, line 377: Maybe we have this vended from our merchant backend, and we add a parameter so we can test 1 or many custom payment methods? (e.g. the setting for customPaymentMethods is a value between 0-3 rather than a boolean)

Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
```diff
@@ -44,6 +44,10 @@ struct PaymentSheetTestPlayground: View {
         }
         SettingView(setting: $playgroundController.settings.userOverrideCountry)
         SettingView(setting: $playgroundController.settings.externalPaymentMethods)
+        // The hardcoded CPM id is only available on our US merchant
+        if playgroundController.settings.merchantCountryCode == .US {
+            SettingView(setting: $playgroundController.settings.customPaymentMethods)
+        }
```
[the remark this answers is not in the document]
porter-stripe: Yeah it could! It would no-op internally, but better to not set it, updated this.

Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
porter-stripe: Hm, it feels needlessly complex tbh. Even if we needed multiple CPMs I'd rather have them defined client-side, this is more consistent with how merchants will integrate as well. Although nothing stopping them from using a backend to vend them as you suggested.

Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
```diff
@@ -368,6 +370,29 @@ class PlaygroundController: ObservableObject {
...
+            return nil
+        }
+    }
+
+    func handleCustomPaymentMethod(
+        _ customPaymentMethodType: PaymentSheet.CustomPaymentMethodConfiguration.CustomPaymentMethodType,
+        _ billingDetails: STPPaymentMethodBillingDetails
+    ) async -> PaymentSheetResult {
+        return await withCheckedContinuation { continuation in
+            handleExternalPaymentMethod(type: customPaymentMethodType.id, billingDetails: billingDetails) { result in
```
joyceqin-stripe, line 390: Will handling custom payment methods always be the same as handling external payment methods? Or will it eventually differ

porter-stripe: Handling confirmation of CPMs and EPMs will be the same. The merchant is responsible for handling the confirmation, then returning back to us the final result.

PR #4636, opened 2025-03-06
Refactor update payment method to return Result

joyceqin-stripe:
## Summary

Refactor UpdatePaymentMethodViewControllerDelegate's didUpdate to return a Result<Void, [Error]> instead of throwing 
Helper functions updateCardBrand and updateDefault return Result<Void, Error>
## Motivation

Cleaner error handling for updating payment method
## Testing

Manual— sent invalid requests for updating card brand/default and checked the error message

### DIFF
--- a/StripeCore/StripeCore/Source/Categories/NSError+Stripe.swift
+++ b/StripeCore/StripeCore/Source/Categories/NSError+Stripe.swift
@@ -60,13 +60,35 @@ extension NSError {
         )
     }
 
-    @objc @_spi(STP) public class func stp_defaultPaymentMethodNotChangedError() -> NSError {
+    @objc @_spi(STP) public class func stp_cardBrandNotUpdatedError() -> NSError {
         let userInfo = [
-            NSLocalizedDescriptionKey: self.stp_defaultPaymentMethodNotChangedMessage()
+            NSLocalizedDescriptionKey: self.stp_cardBrandNotUpdatedMessage()
         ]
         return NSError(
             domain: STPError.stripeDomain,
-            code: STPErrorCode.invalidRequestError.rawValue,
+            code: STPErrorCode.apiError.rawValue,
+            userInfo: userInfo
+        )
+    }
+
+    @objc @_spi(STP) public class func stp_defaultPaymentMethodNotUpdatedError() -> NSError {
+        let userInfo = [
+            NSLocalizedDescriptionKey: self.stp_defaultPaymentMethodNotUpdatedMessage()
+        ]
+        return NSError(
+            domain: STPError.stripeDomain,
+            code: STPErrorCode.apiError.rawValue,
+            userInfo: userInfo
+        )
+    }
+
+    @objc @_spi(STP) public class func stp_genericErrorOccurredError() -> NSError {
+        let userInfo = [
+            NSLocalizedDescriptionKey: self.stp_genericErrorOccurredMessage()
+        ]
+        return NSError(
+            domain: STPError.stripeDomain,
+            code: STPErrorCode.apiError.rawValue,
             userInfo: userInfo
         )
     }
@@ -144,10 +166,24 @@ extension NSError {
         )
     }
 
-    @objc @_spi(STP) public class func stp_defaultPaymentMethodNotChangedMessage() -> String {
+    @objc @_spi(STP) public class func stp_cardBrandNotUpdatedMessage() -> String {
         return STPLocalizedString(
-            "Default payment method was not changed. Please try again.",
+            "Card brand was not updated. Please try again.",
+            "An error message displayed when updating a card brand fails."
+        )
+    }
+
+    @objc @_spi(STP) public class func stp_defaultPaymentMethodNotUpdatedMessage() -> String {
+        return STPLocalizedString(
+            "Default payment method was not updated. Please try again.",
             "An error message displayed when setting a default payment method fails."
+        )
+    }
+
+    @objc @_spi(STP) public class func stp_genericErrorOccurredMessage() -> String {
+        return STPLocalizedString(
+            "An error occurred. Please try again.",
+            "A generic error message displayed when an error occurs."
         )
     }
 }
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
@@ -235,42 +235,66 @@ extension EmbeddedPaymentElement: UpdatePaymentMethodViewControllerDelegate {
     }
 
     func didUpdate(viewController: UpdatePaymentMethodViewController,
-                   paymentMethod: StripePayments.STPPaymentMethod) async throws {
-        try await withThrowingTaskGroup(of: Void.self) { group in
-            if let updateParams = viewController.updateParams,
-               case .card(let paymentMethodCardParams) = updateParams {
-                group.addTask {
-                    try await self.updateCardBrand(paymentMethod: paymentMethod, updateParams: STPPaymentMethodUpdateParams(card: paymentMethodCardParams, billingDetails: nil))
-                }
-            }
-            if viewController.setAsDefaultValue ?? false {
-                group.addTask {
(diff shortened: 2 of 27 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -235,42 +235,63 @@ extension EmbeddedPaymentElement: UpdatePaymentMethodViewControllerDelegate {
...
-        // Update savedPaymentMethods
-        if let row = self.savedPaymentMethods.firstIndex(where: { $0.stripeId == updatedPaymentMethod.stripeId }) {
-            self.savedPaymentMethods[row] = updatedPaymentMethod
+            // Update savedPaymentMethods
+            if let row = self.savedPaymentMethods.firstIndex(where: { $0.stripeId == updatedPaymentMethod.stripeId }) {
+                self.savedPaymentMethods[row] = updatedPaymentMethod
+            }
+            return .success(())
+        } catch {
+            return .failure(NSError.stp_cardBrandNotUpdatedError())
```
porter-stripe, line 283: Same here, just `.failure(error)`. Theres a few cases where we can do this, I'll stop flagging them individually.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
```diff
@@ -15,10 +15,15 @@ import UIKit
...
     func didUpdate(viewController: UpdatePaymentMethodViewController,
-                   paymentMethod: STPPaymentMethod) async throws
+                   paymentMethod: STPPaymentMethod) async -> UpdatePaymentMethodResult
     func shouldCloseSheet(_: UpdatePaymentMethodViewController)
 }
 
+enum UpdatePaymentMethodResult {
+    case success
+    case failure([Error])
+}
```
porter-stripe, line 25: I don't think we need to define our own Result, let's re-use the Swift built in one.  Should be able to do `Result<Void, [Error]>`.

https://developer.apple.com/documentation/swift/result

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
```diff
@@ -227,11 +232,18 @@ final class UpdatePaymentMethodViewController: UIViewController {
...
             STPAnalyticsClient.sharedClient.logPaymentSheetEvent(event: configuration.hostedSurface.analyticEvent(for: .updateCard),
                                                                  params: analyticsParams)
-        } catch {
+        case .failure(let errors):
+            var error: Error
+            if errors.count > 1 {
+                error = NSError.stp_genericErrorOccurredError()
+            } else {
+                error = errors[0]
+            }
```
porter-stripe, line 246: Just an idea:

```swift
    let displayError: Error = {
        guard !errors.isEmpty else {
            // TODO add comment about this case
            return NSError.stp_genericErrorOccurredError()
        }
        
        return errors.count == 1 ? errors[0] : NSError.stp_genericErrorOccurredError()
    }()
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetVerticalViewController.swift
```diff
@@ -868,40 +868,61 @@ extension PaymentSheetVerticalViewController: UpdatePaymentMethodViewControllerD
...
+            errors.append(error)
+        }
+
+        if errors.isEmpty {
+            // Update UI
+            regenerateUI()
+            _ = viewController.bottomSheetController?.popContentViewController()
+            return .success
+        } else {
+            return .failure(errors)
```
porter-stripe, line 898: I think a nice way to write this so you don't need to define optional results would be something like below. We could also apply this to other spots we do similar operations.

```swift
func didUpdate(viewController: UpdatePaymentMethodViewController,
               paymentMethod: STPPaymentMethod) async -> Result<Void, [Error]> {
    var errors: [Error] = []

    // Perform card brand update if needed
    if let updateParams = viewController.updateParams,
       case .card(let paymentMethodCardParams) = updateParams {
        let cardParams = STPPaymentMethodUpdateParams(card: paymentMethodCardParams, billingDetails: nil)
        if case .failure(let error) = await updateCardBrand(paymentMethod: paymentMethod, updateParams: cardParams) {
            errors.append(error)
        }
    }
    
    // Update default payment method if needed
    if viewController.setAsDefaultValue == true {
        if case .failure(let error) = await updateDefault(paymentMethod: paymentMethod) {
            errors.append(error)
        }
    }

    // Early return for errors using guard
    guard errors.isEmpty else {
        return .failure(errors)
    }
    
    // Success
    regenerateUI()
    _ = viewController.bottomSheetController?.popContentViewController()
    return .success(())
}
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
porter-stripe: On second thought, maybe we want that stp_cardBrandNotUpdatedError or specific error, maybe we keep as is. What do you think?

porter-stripe (commented):
nice!

StripeCore/StripeCore/Source/Categories/NSError+Stripe.swift
```diff
@@ -60,13 +60,35 @@ extension NSError {
...
+            NSLocalizedDescriptionKey: self.stp_defaultPaymentMethodNotUpdatedMessage()
+        ]
+        return NSError(
+            domain: STPError.stripeDomain,
+            code: STPErrorCode.apiError.rawValue,
+            userInfo: userInfo
+        )
+    }
+
+    @objc @_spi(STP) public class func stp_genericErrorOccurredError() -> NSError {
```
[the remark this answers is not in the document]
joyceqin-stripe: I don't know, the failed to parse error is "The response from Stripe failed to get parsed into valid JSON." That seems pretty different to me, and Bella previously said that she worked with JJ for the exact string to use.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
joyceqin-stripe: It seems to me that we should use the stp_cardBrandNotUpdatedError— this is the error that should be shown on the Update Payment Method screen if something goes wrong while updating the card brand

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
joyceqin-stripe: I tried doing `Result<Void, Error>`, but it wouldn't let me. This was the workaround I found.

StripeCore/StripeCore/Source/Categories/NSError+Stripe.swift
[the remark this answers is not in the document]
porter-stripe: The error that is actually shown to users is "There was an unexpected error -- try again in a few seconds" as far as I can tell. The JSON error is what is returned to merchants, and I think could be fairly accurate.

This new error message is: "An error occurred. Please try again."

I don't feel strongly here, but a new error with the message "An error occurred. Please try again." seems duplicative compared to just using "There was an unexpected error -- try again in a few seconds." I don't see the rationale.

Up to you to keep or not.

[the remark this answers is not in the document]
joyceqin-stripe: I don't feel strongly about it, I think there's no harm in keeping it though

porter-stripe (approved):
🚢

PR #4637, opened 2025-03-07
Send and decode CPMs from elements/sessions

porter-stripe:
## Summary
- We send CPM ids from merchants to elements/session then get them back and decode them
- A following PR will render these CPMs in MPE

## Motivation
- CPMs!

## Testing
- Manual
- New unit tests

## Changelog
N/A

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
@@ -374,7 +374,7 @@ class PlaygroundController: ObservableObject {
         switch settings.customPaymentMethods {
         case .on:
             // Obtained from https://dashboard.stripe.com/settings/custom_payment_methods
-            let customPaymentMethodType = PaymentSheet.CustomPaymentMethodConfiguration.CustomPaymentMethodType(id: "cpmt_1QpId5Lu5o3P18ZpLwSqMXws",
+            let customPaymentMethodType = PaymentSheet.CustomPaymentMethodConfiguration.CustomPaymentMethodType(id: "cpmt_1QpIMNLu5o3P18Zpwln1Sm6I",
                                                                                                                 subcopy: "Pay with BufoPay")
             return .init(customPaymentMethodTypes: [customPaymentMethodType], customPaymentMethodConfirmHandler: handleCustomPaymentMethod(_:_:))
         case .off:
--- a/StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/CustomPaymentMethod.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/CustomPaymentMethod.swift
@@ -13,15 +13,19 @@ struct CustomPaymentMethod: Decodable {
     /// The type of the external payment method. e.g. `"external_foopay"`
     /// These match the strings specified by the merchant in `ExternalPaymentMethodConfiguration`.
     let displayName: String?
-    /// A localized label for the payment method e.g. "FooPay"
+    
+    /// The type (id) of the external payment method. e.g. `"cpmt_..."`
+    /// These match the ids specified by the merchant in `CustomPaymentMethodConfiguration`.
     let type: String
-
+    
+    /// URL of a 48x pixel tall, variable width PNG representing the payment method.
     let logoUrl: URL?
     
+    /// If there was an error fetching this custom payment method this will be populated with the error
     let error: String?
 
-    /// Helper method to decode the `v1/elements/sessions` response's `external_payment_methods_data` hash.
-    /// - Parameter response: The value of the `external_payment_methods_data` key in the `v1/elements/sessions` response.
+    /// Helper method to decode the `v1/elements/sessions` response's `custom_payment_methods_data` hash.
+    /// - Parameter response: The value of the `custom_payment_methods_data` key in the `v1/elements/sessions` response.
     public static func decoded(fromAPIResponse response: [[AnyHashable: Any]]?) -> [CustomPaymentMethod]? {
         guard let response else {
             return nil
--- a/StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/CustomPaymentMethod.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/CustomPaymentMethod.swift
@@ -13,7 +13,7 @@ struct CustomPaymentMethod: Decodable {
     /// The display name of this custom payment method as defined in the Stripe dashboard
     let displayName: String?
 
-    /// The type (id) of the external payment method. e.g. `"cpmt_..."`
+    /// The type (id) of the custom payment method. e.g. `"cpmt_..."`
     /// These match the ids specified by the merchant in `CustomPaymentMethodConfiguration`.
     let type: String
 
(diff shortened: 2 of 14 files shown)


StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/CustomPaymentMethod.swift
```diff
@@ -0,0 +1,44 @@
...
+//  StripePaymentSheet
+//
+//  Created by Nick Porter on 3/5/25.
+//
+
+import Foundation
+@_spi(STP) import StripeCore
+@_spi(STP) import StripePayments
+
+struct CustomPaymentMethod: Decodable {
```
porter-stripe, line 12: Noting that all these properties (besides type) are intentionally optional, see an example response here for why: https://github.com/stripe/stripe-ios/pull/4637/files#diff-c71d9097e968f7c5fd01a4cb9425c80f53998412b3026808eca160f1ca08b35fR139

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/CustomPaymentMethod.swift
```diff
@@ -0,0 +1,45 @@
...
+
+    /// The type (id) of the custom payment method. e.g. `"cpmt_..."`
+    /// These match the ids specified by the merchant in `CustomPaymentMethodConfiguration`.
+    let type: String
+
+    /// URL of a 48x pixel tall, variable width PNG representing the payment method.
+    let logoUrl: URL?
+
+    /// If true, this custom payment method was created using a preset in the Stripe dashboard
+    let isPreset: Bool?
```
joyceqin-stripe, line 25: What does it mean for isPreset to be nil vs false?

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/STPElementsSession.swift
```diff
@@ -169,6 +175,23 @@ extension STPElementsSession: STPAPIResponseDecodable {
             return epms
         }()
 
+        let customPaymentMethods: [CustomPaymentMethod] = {
+            let customPaymentMethodDataKey = "custom_payment_method_data"
+            guard response[customPaymentMethodDataKey] != nil, !(response[customPaymentMethodDataKey] is NSNull) else {
```
joyceqin-stripe, line 180: Why do we need to check both `response[customPaymentMethodDataKey] != nil` and `!(response[customPaymentMethodDataKey] is NSNull)`?

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/CustomPaymentMethod.swift
porter-stripe: Good question, I touched on it in this comment: https://github.com/stripe/stripe-ios/pull/4637#discussion_r1985268417

For more of an explanation though, this is optional based on the behavior of the backend. There are instances where the backend may not return us "is_preset" in the response so our model client-side needs to handle that. The case where isPreset is nil, is the case where there is no value specified by the backend. The goal of `CustomPaymentMethod` is to match the backend definition/behavior.

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/STPElementsSession.swift
porter-stripe: If you look higher up in this function you will see we do this for external payment methods and customer. The reason is b/c:

1. nil check: This checks if the key doesn't exist in the response dictionary at all.
2. NSNull check: This checks if the key exists but its value is explicitly set to NSNull (null)

Here is an example:

Case 1: Key doesn't exist (nil check)
```json
{
  "other_data": "some value",
  "payment_methods": ["card", "paypal"]
  // "custom_payment_method_data" key is completely absent
}
```

Case 2:
```json
{
  "other_data": "some value",
  "payment_methods": ["card", "paypal"],
  "custom_payment_method_data": null
}
```

We need to be able to handle both scenarios just to be safe.

joyceqin-stripe (approved):
Looks fine to me

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/STPElementsSessionTest.swift
```diff
@@ -113,6 +113,77 @@ class STPElementsSessionTest: XCTestCase {
...
+                "logo_url": "stripe.com",
+                "display_name": "BufoPay (test)",
+                "type": "cpmt_123",
+                "error": NSNull(),
+                "is_preset": false,
+            ],
+            [
+                "logo_url": NSNull(),
+                "display_name": NSNull(),
+                "type": "cpmt_invalid",
```
wooj-stripe, line 150: Not your fault, but i think this object was modeled awkwardly.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/STPElementsSessionTest.swift
```diff
@@ -113,6 +113,77 @@ class STPElementsSessionTest: XCTestCase {
...
+        }))
+    }
+
+    func testSuccessCPMParsing() {
+        // If STPElementsSession decodes a dict...
+        var elementsSessionJson = STPTestUtils.jsonNamed("ElementsSession")!
+
+        let customPaymentMethodsArray: [[String: Any]] = [
+            [
+                "logo_url": "stripe.com",
```
wooj-stripe, line 141: should logo_url be a fully qualified url: etc:  https://  -- i'd expect this fail?

Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
```diff
@@ -374,7 +374,7 @@ class PlaygroundController: ObservableObject {
         switch settings.customPaymentMethods {
         case .on:
             // Obtained from https://dashboard.stripe.com/settings/custom_payment_methods
-            let customPaymentMethodType = PaymentSheet.CustomPaymentMethodConfiguration.CustomPaymentMethodType(id: "cpmt_1QpId5Lu5o3P18ZpLwSqMXws",
+            let customPaymentMethodType = PaymentSheet.CustomPaymentMethodConfiguration.CustomPaymentMethodType(id: "cpmt_1QpIMNLu5o3P18Zpwln1Sm6I",
```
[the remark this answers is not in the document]
porter-stripe: Yep this code will come from the merchants app, they could vend it from the backend, but totally up to them. Our integration guide will show them to just hardcode in the app code for simplicity, same with how we do it for EPMs. https://docs.stripe.com/payments/mobile/external-payment-methods?platform=ios

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/STPElementsSessionTest.swift
porter-stripe: Yeah I agree! This object will be "translated" into something that is displayable in MPE, so we'll have a better model of this in a future PR, but for this type I want it to match the backend behavior.

Totally agree the model of the object in the error state is awkward.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/STPElementsSessionTest.swift
porter-stripe: Added https!

wooj-stripe (approved):

PR #4641, opened 2025-03-07
show disabled checkbox when it is the default pm

joyceqin-stripe:
## Summary

Always display the default checkbox on the update payment method screen. Disable it if it is already the default and change the display text to "Default payment method"
## Motivation

https://jira.corp.stripe.com/browse/MOBILESDK-3260
## Testing

Updated UI tests, snapshot tests


porter-stripe (approved):
🚢

PR #4643, opened 2025-03-10
Hide default checkbox and auto-set if first

joyceqin-stripe:
## Summary

If a customer has no saved payment methods and the default feature is enabled, hide the default checkbox and automatically set new card or US bank account as the default.
## Motivation

Bug bash suggestion
## Testing

Updated UI test

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
+++ b/Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
@@ -2816,10 +2816,8 @@ class PaymentSheetDefaultSPMUITests: PaymentSheetUITestCase {
         var saveThisCardToggle = app.switches["Save payment details to Example, Inc. for future purchases"]
         saveThisCardToggle.tap()
         XCTAssertTrue(saveThisCardToggle.isSelected)
-        // toggle set this card as default
-        var setDefaultToggle = app.switches["Set as default payment method"]
-        setDefaultToggle.tap()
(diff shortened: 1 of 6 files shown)


Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
```diff
@@ -2854,8 +2852,7 @@ class PaymentSheetDefaultSPMUITests: PaymentSheetUITestCase {
         saveThisCardToggle.tap()
         XCTAssertTrue(saveThisCardToggle.isSelected)
         // do not set this card as default
-        setDefaultToggle = app.switches["Set as default payment method"]
-        XCTAssertFalse(setDefaultToggle.isSelected)
+        XCTAssertFalse(app.switches["Set as default payment method"].isSelected)
```
porter-stripe, line 2855: Do we need this? Is the "Set as default payment method" even visible?

porter-stripe (approved):
Just a non blocking question

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
joyceqin-stripe: We previously saved a card, so the "Set as default payment method" checkbox is shown when we try adding another card. Since the new card wouldn't be the only saved payment method, it wouldn't automatically be set as the default.

wooj-stripe (approved):

PR #4649, opened 2025-03-10
End editing when updating a card

wooj-stripe:
## Summary
Calls end editing before attempting update a card.

## Motivation
Errors were sometimes not being displayed because the field was still editing. 

## Testing
Manual testing

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
@@ -230,7 +230,7 @@ final class UpdatePaymentMethodViewController: UIViewController {
         guard let delegate else {
             return
         }
-        // Ensure endEditing is called prior to isUserInteractionEnabled
+        // Ensure endEditing(true) is called prior to setting isUserInteractionEnabled
         view.endEditing(true)
         view.isUserInteractionEnabled = false
         updateButton.update(state: .spinnerWithInteractionDisabled)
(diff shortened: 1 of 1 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
```diff
@@ -226,8 +224,14 @@ final class UpdatePaymentMethodViewController: UIViewController {
 
         present(alertController, animated: true, completion: nil)
     }
-
-    private func update(updatePaymentMethodOptions: UpdatePaymentMethodOptions?) async {
+    private func update(updatePaymentMethodOptions: UpdatePaymentMethodOptions?) {
+        // End editing otherwise, if the textfield is still selected, errors won't be rendered on screen
+        view.endEditing(true)
```
joyceqin-stripe, line 229: Is there a reason we need to separate _update into a separate function? Couldn't we call view.endEditing(true) in the same place where we set view.isUserInteractionEnabled = false?

wooj-stripe: I tried to, but call to `endEditing` in a Task (off the main thread) seems to be ignored.  Alternatively I could have put this in the button closure, but it seemed kind of important and if we decide to call update() somewhere else in the codebase, it would easily be missed.

porter-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
joyceqin-stripe: I tried it out, and it looks to me like if you put `view.endEditing(true)` _before_ `view.isUserInteractionEnabled = false`, it works, but not if you put it after

joyceqin-stripe (commented):
Try putting `view.endEditing(true)` on line 237 and see if that works— I don't think we need to separate it into another function

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
wooj-stripe: NICE FIND!

joyceqin-stripe (approved):

PR #4660, opened 2025-03-12
Make circular edit buttons primary color instead of grey, make whole pm tappable in horizontal mode

joyceqin-stripe:
## Summary

Made circular edit buttons blue to indicate tap-ability and made cells tappable
## Motivation

[MOBILESDK-3358](https://jira.corp.stripe.com/browse/MOBILESDK-3358)
## Testing

- Manually
- Updated snapshot

## Changelog

N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentMethodCollectionView.swift
```diff
@@ -299,6 +299,21 @@ extension SavedPaymentMethodCollectionView {
             }
         }
 
+        private func setupTapGestureIfNecessary() {
+            // Check if the gesture has already been set
+            if tapGestureRecognizer == nil {
+                tapGestureRecognizer = UITapGestureRecognizer(target: self, action: #selector(didSelectAccessory))
+                self.addGestureRecognizer(tapGestureRecognizer!)
```
wooj-stripe, line 306: We should avoid using `!` in non-test code

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentMethodCollectionView.swift
```diff
@@ -299,6 +299,21 @@ extension SavedPaymentMethodCollectionView {
...
+            }
+        }
+
+        private func removeTapGestureIfNecessary() {
+            if let gesture = tapGestureRecognizer {
+                self.removeGestureRecognizer(gesture)
+                tapGestureRecognizer = nil
+            }
+        }
```
wooj-stripe, line 315: Instead of adding/removing a tap gesture recognizer, did we consider something like:

1. in `init()` add a gesture recognizer to the view:
```
            self.addGestureRecognizer(UITapGestureRecognizer(target: self, action: #selector(didTap)))
```

2. implement didTap() to only allow users to go to the update page if we're in 'remove/editing' mode:
```
        @objc
        private func didTap() {
            if isRemovingPaymentMethods { // probably needs a better name :P
                delegate?.paymentOptionCellDidSelectEdit(self)
             }
         }
```

wooj-stripe: Another approach (which, should work?) would be to hook into the ViewController's UICollectionViewDelegate  (collectionView: shouldSelectItemAt)

joyceqin-stripe: I can try the UICollectionViewDelegate approach, but adding the tap gesture on init without removing it doesn't work— it kind of takes over, and any taps outside of editing mode don't register properly, e.g. tapping another pm to select it won't select it

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSavedPaymentMethodsCollectionViewController.swift
```diff
@@ -409,6 +409,10 @@ extension CustomerSavedPaymentMethodsCollectionViewController: UICollectionViewD
         -> Bool
     {
         guard !self.collectionView.isRemovingPaymentMethods else {
+            guard let cell = collectionView.cellForItem(at: indexPath) as? SavedPaymentMethodCollectionView.PaymentOptionCell else {
+                return false
+            }
+            cell.didSelectAccessory()
             return false
         }
```
wooj-stripe, line 417: 1. Your logic is totally correct but, and this is a stylistic comment but reading two `guard`s in a row feel a bit awkward. Guards force a an `else..return` so we typically use them to indicate a pre-condition or a short-circuit to force stop executing the function -- one guard feels like it's enough since the compiler will enforce ending the function in the else clause.

2. Seems like we don't have to change didSelectAccessory() from private to internal -- we can call `paymentOptionCellDidSelectEdit` directly.

This results in the following (and would apply to the payment sheet case as well)

```
        guard !self.collectionView.isRemovingPaymentMethods else {
            if let cell = collectionView.cellForItem(at: indexPath) as? SavedPaymentMethodCollectionView.PaymentOptionCell {
                paymentOptionCellDidSelectEdit(cell)
            }
            return false
        }
```

wooj-stripe (approved):

PR #4661, opened 2025-03-12
Don't select form payment option if no previous selection

porter-stripe:
## Summary
- Now anytime we close the form we revert back to the previous selection, if any.
- Fixed a bug where we were updating the payment option to the merchant when a form was presented (before the user tapped Continue in the form).

## Motivation
- https://jira.corp.stripe.com/browse/MOBILESDK-3228

## Testing
- Manual
- New unit test

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -401,12 +402,9 @@ extension EmbeddedPaymentElement: EmbeddedFormViewControllerDelegate {
...
                 // Go back to the previous selection if there was one
                 embeddedPaymentMethodsView.resetSelectionToLastSelection()
-            } else {
-                // If there wasn't a previous selection, deselect if the selection isn't valid.
-                let isCurrentSelectionValid = embeddedFormViewController.selectedPaymentOption != nil
-                if !isCurrentSelectionValid {
-                    embeddedPaymentMethodsView.resetSelection()
-                }
+            } else if paymentOption != lastUpdatedPaymentOption {
+                // Reset selection if no previous selection exists and the closing form's payment option differs from last updated
```
yuki-stripe, line 406: This comment is too much of a direct translation of the code IMO and not useful enough. Can we say what we're trying to do here at a higher level?

For example, the code above:
```
            if let lastSelection, lastSelection != currentlySelectedType {
                // Go back to the previous selection if there was one
```
This tells you what we're trying to accomplish and would hold true regardless of what the actual code looks like.

If the comment was a direct translation of the code it'd say something like:
`// Reset selection to last selection if there's a last selection and the last selection differs from the currently 
selected type.`

which isn't so helpful.

porter-stripe: updated!

Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
```diff
@@ -645,10 +645,11 @@ class EmbeddedUITests: PaymentSheetUITestCase {
...
         app.buttons["Close"].waitForExistenceAndTap()
-        // ...should de-select the row and clear the payment option
-        XCTAssertFalse(app.staticTexts["Payment method"].waitForExistence(timeout: 1))
-        XCTAssertFalse(app.buttons["New card"].isSelected)
-        XCTAssertFalse(app.buttons["Checkout"].isEnabled)
+        // ...should re-select the last payment option (card)
+        XCTAssertTrue(app.staticTexts["Payment method"].waitForExistence(timeout: 1))
+        XCTAssertEqual(app.staticTexts["Payment method"].label, "•••• 4444")
+        XCTAssertTrue(app.buttons["New card"].isSelected)
+        XCTAssertTrue(app.buttons["Checkout"].isEnabled)
```
yuki-stripe, line 652: This is a bug!  We've broken our contract with the merchant - paymentOption is nil now, but we didn't tell them it updated. Their checkout button is enabled, but if they `confirm`, they'll get an assert.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentElementTest.swift
```diff
@@ -563,6 +613,48 @@ class EmbeddedPaymentElementTest: XCTestCase {
...
+        XCTAssertTrue(delegateDidUpdatePaymentOptionCalled, "Delegate should be updated immediately for payment methods without forms")
+        XCTAssertEqual(sut.paymentOption?.label, "Cash App Pay")
+
+        // Reset delegate tracking
+        delegateDidUpdatePaymentOptionCalled = false
+
+        // Select Card (which has a form to collect input)...
+        sut.embeddedPaymentMethodsView.didTap(rowButton: sut.embeddedPaymentMethodsView.getRowButton(accessibilityIdentifier: "Card"))
+
+        // ...the delegate should not be called yet since the form needs to be filled
```
yuki-stripe, line 640:
```suggestion
        // ...the delegate should not be called yet since the payment option is in an indeterminate state while the form is open
```

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentElementTest.swift
```diff
@@ -563,6 +613,48 @@ class EmbeddedPaymentElementTest: XCTestCase {
...
+        XCTAssertFalse(delegateDidUpdatePaymentOptionCalled, "Delegate should not be updated immediately for payment methods with forms")
+        XCTAssertNil(sut.paymentOption, "Payment option should be nil until card form is filled")
```
yuki-stripe, line 642:
```suggestion
        XCTAssertNil(sut.paymentOption, "Payment option should be nil until form is canceled or completed")
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -394,35 +395,20 @@ extension EmbeddedPaymentElement: EmbeddedFormViewControllerDelegate {
...
-                // Go back to the previous selection if there was one
-                embeddedPaymentMethodsView.resetSelectionToLastSelection()
-            } else {
-                // If there wasn't a previous selection, deselect if the selection isn't valid.
-                let isCurrentSelectionValid = embeddedFormViewController.selectedPaymentOption != nil
-                if !isCurrentSelectionValid {
-                    embeddedPaymentMethodsView.resetSelection()
-                }
-            }
+        // Go back to the previous selection if there was one
```
yuki-stripe, line 398: I ticketed up the work to restore the previous form when you cancel https://jira.corp.stripe.com/browse/MOBILESDK-3361

Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
porter-stripe: Ahh, you're right, good catch, thanks! You're right below, we will need to regenerate the form VC (and maybe even restore the cache to it's previous state?)

I made a change so that if we dismiss the form and it is of the same type we previously had selected we
1. Clear the selecting entirely if the payment option in the form has changed and user hits cancel
2. If the payment option hasn't changed go back to what we have previously selected. We can do this b/c the form matches the payment option and don't need to regenerate the form.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -394,35 +395,32 @@ extension EmbeddedPaymentElement: EmbeddedFormViewControllerDelegate {
...
-                    embeddedPaymentMethodsView.resetSelection()
-                }
-            }
+        let lastSelection = embeddedPaymentMethodsView.previousSelectedRowButton?.type
+        let currentlySelectedType = embeddedPaymentMethodsView.selectedRowButton?.type
+
+        // If the user re-selects a valid payment option w/ form, then modifies it, then hits close, we clear selection
+        // Ideally we would revert back to the valid payment option that existed when the form was presented rather than totally clear selection
+        // To restore to the previous payment option we need to restore the previous form VC that contained the previous payment option
+        // TODO (https://jira.corp.stripe.com/browse/MOBILESDK-3361): Consider restoring the form VC and form cache to revert to the last valid payment option
```
yuki-stripe, line 404: Good comment!

yuki-stripe (approved):

PR #4670, opened 2025-03-17
Try to rename file icon_x_standalone

porter-stripe:
## Summary
- Some users report issues when loading the SDK via cocoapods w.r.t to this image
- We thought we fixed this issue, but still can't reproduce
- The only difference from this image and other is that the image name and the name of the file didn't match. This shouldn't be an issue but worth trying.

## Motivation
https://jira.corp.stripe.com/browse/RUN_MOBILESDK-3971

## Testing
- N/A

## Changelog
N/A


davidme-stripe (approved):

davidme-stripe:
Hrm, this is probably worth trying, but I think "icon_x_standalone" is just the first image to load. We've seen this issue with Asset Catalogs and React Native before, but haven't received any reports of it in a long time, so I thought it was fixed…

PR #4675, opened 2025-03-17
LPM rendering events

joyceqin-stripe:
## Summary

Added LPM render event to determine what the user sees without overflow.

For horizontal mode, the event has params "visible_lpms" and "hidden_lpms" corresponding to the visible cells in the carousel and wallet headers on render and the rest of the lpms that are not visible in the carousel.
For vertical mode and embedded, the event has params "visible_lpms" and "hidden_lpms", but there are no hidden lpms, as the user scrolls through all of the lpms in order to pay.

Emitted on viewDidAppear for horizontal and vertical, emitted on layout subviews for embedded.

LPM default selected (which LPM is default selected if any) can be found on mc_form_shown - "selected_lpm"
LPM selected (which LPMs did buyer select if any) can be found on mc_carousel_payment_method_tapped - "selected_lpm"
## Motivation

[MOBILESDK-3221](https://jira.corp.stripe.com/browse/MOBILESDK-3221)
## Testing

- Unit test log
- UI tests that verify analytics

## Changelog

N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetVerticalViewController.swift
```diff
@@ -470,9 +470,23 @@ class PaymentSheetVerticalViewController: UIViewController, FlowControllerViewCo
...
+        logRenderLPMs()
         isLinkWalletButtonSelected = false
     }
 
+    private func logRenderLPMs() {
+        // The user has to scroll through all the payment method options before checking out, so all of the lpms are visible
+        var visibleLPMs: [String] = paymentMethodListViewController?.rowButtons.filter { !$0.type.isSaved }.compactMap { $0.type.analyticsIdentifier } ?? []
+        // Add wallet LPMs
+        if PaymentSheet.isApplePayEnabled(elementsSession: elementsSession, configuration: configuration) && !shouldShowApplePayInList {
+            visibleLPMs.append("apple_pay")
```
davidme-stripe, line 482: Maybe we could use the `RowButtonType.applePay.analyticsIdentifier` here? Not totally sure, as we don't have a similar value to reference in Horizontal mode.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetVerticalViewController.swift
```diff
@@ -470,9 +470,23 @@ class PaymentSheetVerticalViewController: UIViewController, FlowControllerViewCo
...
+        }
+        if PaymentSheet.isLinkEnabled(elementsSession: elementsSession, configuration: configuration) && !shouldShowLinkInList {
+            visibleLPMs.append("link")
```
davidme-stripe, line 485: Same here, but with Link

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetFlowControllerViewController.swift
```diff
@@ -217,12 +217,18 @@ class PaymentSheetFlowControllerViewController: UIViewController, FlowController
             cbcEligible: elementsSession.isCardBrandChoiceEligible,
             analyticsHelper: analyticsHelper
         )
+
+        // if Link is the only payment option in FlowController, we add it as a wallet header
+        // otherwise, ApplePay and Link don't show up in the LPM options when adding a new payment method
+        let isLinkOnlyMode: Bool = !isApplePayEnabled && isLinkEnabled && !savedPaymentOptionsViewController.hasOptionsExcludingAdd
```
davidme-stripe, line 223: I'm wondering if we could share this logic with the existing `linkOnlyMode` bool on line 64.

One way to do this might be to change it from a `var` to a `let private linkOnlyMode: Bool` and then initialize it here in the init:
```
self.linkOnlyMode = !isApplePayEnabled && isLinkEnabled && !savedPaymentOptionsViewController.hasOptionsExcludingAdd
```

But we'd need to make sure `savedPaymentOptionsViewController.hasOptionsExcludingAdd` won't change later. And it seems like it might, as the viewModels of SPOVC could change as the user adds/removes PM options. But we already don't account for that (as we check `linkOnlyMode` in `viewDidLoad()`), so maybe that's fine.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/AddPaymentMethodViewController.swift
```diff
@@ -138,9 +141,21 @@ class AddPaymentMethodViewController: UIViewController {
 
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
+        logRenderLPMs()
         delegate?.didUpdate(self)
     }
 
+    private func logRenderLPMs() {
+        // These are the cells that are visible without scrolling in the horizontal carousel
```
davidme-stripe, line 149: How does this work in the PayWithLinkViewController, or if no LPMs are available? Will we just send an empty list + the wallet buttons if available? Would we send 'card' even if the buttons aren't shown?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetFlowControllerViewController.swift
joyceqin-stripe: I think I've found a bug because we don't check linkOnlyMode enough. If you start with link enabled, Apple Pay disabled, and 1 saved pm, then on viewDidLoad, linkOnlyMode is false. But if you remove that last saved pm, linkOnlyMode is true, but since we don't check that value, we get behavior where we're not automatically directed to the add screen. When we select the add button, we correctly determine that the wallet header should be shown. However, our wallet header view doesn't have any wallet options in it, so the Link header doesn't show up

https://github.com/user-attachments/assets/a8fe46a9-7201-4645-b848-bfb98b2f5ad1

I'll make a pr sometime to fix this so we automatically switch to .addingNew when we don't have other pms besides Link + display the wallet view correctly.

So I think we should keep linkOnlyMode as a var, and I think walletHeaders in the Add VC should be a var that we update to reflect the value of linkOnlyMode prior to switching the mode to .addingNew for the analytics. Wdyt?

davidme-stripe: Great find! Adding a walletHeaders var sounds good to me. You could also have it go the other way, where the AddVCDelegate adds a delegate method to fetch the walletHeaders, and then AddVC can fetch the current value from its delegate.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/AddPaymentMethodViewController.swift
joyceqin-stripe: I didn't consider this case! Would we want the PayWithLinkViewController to log anything, or do we only care about the lpms rendered in PaymentSheet?

If no lpms are available but the card form is shown, I think we should log card + the wallet buttons if available. The code currently does not handle this case.

davidme-stripe: That makes sense! Technically we're not showing any buttons if `card` is the only available PM, but I think it's fair to log `card` in that case

davidme-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
```diff
@@ -214,10 +214,24 @@ class EmbeddedPaymentMethodsView: UIView {
         fatalError("init(coder:) has not been implemented")
     }
 
+    private func logRenderLPMs() {
+        // The user has to scroll through all the payment method options before checking out, so all of the lpms are visible
```
yuki-stripe, line 218: This isn't necessarily true @joyceqin-stripe @davidme-stripe. Some of our merchants are rendering Embedded in a separate screen from checkout where tapping a row _immediately_ goes back to the checkout screen or immediately confirms.  

If a PM were rendered off screen in this case, the customer can complete checkout without seeing it.

PR #4676, opened 2025-03-17
Fix logic for when editing is available for update

wooj-stripe:
## Summary
Allows users who have integrated with CustomerSessions to update cards if they have paymentmethodRemove || paymentmethodDefault set to disabled.

## Motivation
Currently updating cards is only available if you enable paymentMethodDefault or paymentmethodRemove.

## Testing
updated existing tests and manually tested

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSavedPaymentMethodsCollectionViewController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSavedPaymentMethodsCollectionViewController.swift
@@ -95,11 +95,11 @@ class CustomerSavedPaymentMethodsCollectionViewController: UIViewController {
             return false
         case 1:
             // If there's exactly one PM, customer can only edit if configuration allows removal or if that single PM is editable
-            return (configuration.paymentMethodRemove && configuration.allowsRemovalOfLastSavedPaymentMethod) || viewModels.contains(where: {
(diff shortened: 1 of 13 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSavedPaymentMethodsCollectionViewController.swift
```diff
@@ -95,11 +95,11 @@ class CustomerSavedPaymentMethodsCollectionViewController: UIViewController {
             return false
         case 1:
             // If there's exactly one PM, customer can only edit if configuration allows removal or if that single PM is editable
-            return (configuration.paymentMethodRemove && configuration.allowsRemovalOfLastSavedPaymentMethod) || viewModels.contains(where: {
+            return (configuration.paymentMethodRemove && configuration.allowsRemovalOfLastSavedPaymentMethod) || configuration.paymentMethodUpdate || configuration.paymentMethodSyncDefault || viewModels.contains(where: {
                 $0.isCoBrandedCard && cbcEligible
             })
         default:
-            return configuration.paymentMethodRemove || viewModels.contains(where: {
+            return configuration.paymentMethodRemove || configuration.paymentMethodUpdate || configuration.paymentMethodSyncDefault || viewModels.contains(where: {
```
wooj-stripe, line 102: Looks like we forgot to add `paymentMethodSyncDefault` here, so adding it in addition to `paymentMethodUpdate`

wooj-stripe: Ah! Right!!! Oops.  Will change that back! Thanks for catching.

porter-stripe (approved):
🚢

PR #4686, opened 2025-03-19
Move view controller helpers to Stripe Core.

nschris-stripe:
## Summary
This moves some ViewController helpers used for SwiftUI presentation into StripeCore. I am doing this because we are actively considering using a similar approach to the payment bottom sheet view modifier for the Connect SDK. There is one minor change I made to make findViewController more accurate, called out in a comment.

## Demo

https://github.com/user-attachments/assets/249449e0-cdbf-47b1-8e4f-12017cf83a30

### DIFF
--- a/StripeUICore/StripeUICore/Source/Categories/UIResponder+Stripe.swift
+++ b/StripeUICore/StripeUICore/Source/Categories/UIResponder+Stripe.swift
@@ -1,0 +1,22 @@
+//
+//  UIResponder+Stripe.swift
+//  StripeUICore
+//
+//  Created by Chris Mays on 3/19/25.
+//
+
+import UIKit
+
+@_spi(STP) public extension UIResponder {
+    @available(iOSApplicationExtension, unavailable)
+    @objc func findViewController() -> UIViewController? {
+        if let nextResponder = self.next as? UIViewController {
+            return nextResponder
+        } else if let nextResponder = self.next {
+            return nextResponder.findViewController()
+        } else {
+            // Can't find a view, attempt to grab the top most view controller
+            return UIViewController.topMostViewController()
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


StripeCore/StripeCore/Source/Categories/UIResponder+Stripe.swift
```diff
@@ -0,0 +1,21 @@
...
+//  Created by Chris Mays on 3/19/25.
+//
+
+import UIKit
+
+extension UIResponder {
+    @objc @_spi(STP) public func stp_findViewController() -> UIViewController? {
+        if let nextResponder = self.next as? UIViewController {
+            return nextResponder
+        } else if let nextResponder = self.next {
```
nschris-stripe, line 14: Previously the function assumed the only things in the responder chain were UIViewController or UIView. That is not a requirement of the responder chain, so I made this slightly more generic.

Previously looked like this:
```
func findViewController(for uiView: UIView) -> UIViewController? {
    if let nextResponder = uiView.next as? UIViewController {
        return nextResponder
    } else if let nextResponder = uiView.next as? UIView {
        return findViewController(for: nextResponder)
    } else {
        // Can't find a view, attempt to grab the top most view controller
        return topMostViewController()
    }
}
 ```

File: StripeUICore/StripeUICore/Source/Categories/UIResponder+Stripe.swift [hunk 2]
```diff
@@ -0,0 +1,22 @@
...
+//  StripeUICore
+//
+//  Created by Chris Mays on 3/19/25.
+//
+
+import UIKit
+
+@_spi(STP) public extension UIResponder {
+    @available(iOSApplicationExtension, unavailable)
+    @objc func findViewController() -> UIViewController? {
```
nschris-stripe, line 12: Previously the function assumed the only things in the responder chain were UIViewController or UIView. That is not a requirement of the responder chain, so I made this slightly more generic.

Previously looked like this:
```
func findViewController(for uiView: UIView) -> UIViewController? {
    if let nextResponder = uiView.next as? UIViewController {
        return nextResponder
    } else if let nextResponder = uiView.next as? UIView {
        return findViewController(for: nextResponder)
    } else {
        // Can't find a view, attempt to grab the top most view controller
        return topMostViewController()
    }
}
 ```

porter-stripe (approved):

PR #4689, opened 2025-03-19
[Connect] Update nav bar appearance setting to support dynamic type.

nschris-stripe:
## Summary
This updates the nav bar title to support dynamic type when a custom font is set and also sets default colors when colors were not set. This prevents the nav bar from changing with dark mode if the appearance configuration was not setup to support dark mode.

## Demo

https://github.com/user-attachments/assets/4cb0e052-5b4d-4558-866b-5e982b463d0e

### DIFF
--- a/StripeConnect/StripeConnect/Source/Components/AccountOnboardingController.swift
+++ b/StripeConnect/StripeConnect/Source/Components/AccountOnboardingController.swift
@@ -130,8 +130,14 @@ public final class AccountOnboardingController {
         navController.navigationBar.standardAppearance = appearance
         navController.navigationBar.scrollEdgeAppearance = appearance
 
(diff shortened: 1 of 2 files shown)


StripeConnect/StripeConnect/Source/Components/AccountOnboardingController.swift
```diff
@@ -130,8 +130,14 @@ public final class AccountOnboardingController {
         navController.navigationBar.standardAppearance = appearance
         navController.navigationBar.scrollEdgeAppearance = appearance
 
-        let closeButton = UIBarButtonItem(barButtonSystemItem: .close, target: self, action: #selector(closeButtonTapped(_:)))
+        let closeButton = UIBarButtonItem(
```
nschris-stripe, line 133: Updating the close button because the system close button does not support tint colors.

mats-stripe (approved):

PR #4692, opened 2025-03-20
Support BDCC in native Link

tillh-stripe:
## Summary

This pull request allows Link to be enabled even if the merchant is providing a custom billing details collection configuration (BDCC).

We pass the BDCC along to the native Link experience and collect the required billing details there:
- If a consumer pays with a new payment method, we collect the fields right there.
- If a consumer pays with an existing payment method, we evaluate whether they satisfy the BDCC. We try to prefill empty fields with any information we have through the Link account, such as the phone number. If not, we present a screen for them to “confirm payment details”.

## Testing

End-to-end tests for new and existing consumers, in both payment method mode and passthrough mode.

## Changelog

[Changed] Link is now available as a payment method even if you collect billing details via `billingDetailsCollectionConfiguration`.


StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/STPAPIClient+Link.swift
```diff
@@ -385,8 +389,10 @@ extension STPAPIClient {
         ]
 
         if let details = updateParams.details, case .card(let expiryDate, let billingDetails) = details {
-            parameters["exp_month"] = expiryDate.month
-            parameters["exp_year"] = expiryDate.year
+            if let expiryDate {
```
davidme-stripe, line 392: Does the update call work if expiryDate is nil?

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Elements/LinkCardEditElement.swift
```diff
@@ -111,14 +125,21 @@ final class LinkCardEditElement: Element {
...
-            cardBrandProvider: { [weak self] in
-                self?.paymentMethod.cardDetails?.stpBrand ?? .unknown
-            }
-        ),
-        theme: theme
-    )
+    private lazy var cvcElement: TextFieldElement = {
+        let cvcElement = TextFieldElement(
+            configuration: TextFieldElement.CVCConfiguration(
+                // If we're using a placeholder, we just fill up the view with zeros.
```
davidme-stripe, line 131: We'll render it as "•••" though, right?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+PaymentMethodAvailability.swift
```diff
@@ -81,6 +80,14 @@ extension PaymentSheet {
     internal static var supportedLinkPaymentMethods: [STPPaymentMethodType] = []
 }
 
+private extension STPElementsSession {
+    func isCompatible(with configuration: PaymentElementConfiguration) -> Bool {
+        // We can't collect billing details if we're in the web flow, so turn Link off for those cases.
+        let nativeLink = deviceCanUseNativeLink(elementsSession: self, configuration: configuration)
```
davidme-stripe, line 86: Kinda wondering what happens if the device fails the attestation check after this — I almost wonder if we should just hide Link entirely if attestation fails instead of falling back to the web flow.

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Elements/LinkCardEditElement.swift
tillh-stripe: Yep, but with this ugly workaround 😕 I just noticed that `CensoredCVCConfiguration` exists, which gives us that functionality for free, but it makes the look different from the other disabled fields. What do you think about this?

Simulator Screenshot - iPhone 16 Pro - 2025-03-28 at 11 26 10

davidme-stripe (approved):

davidme-stripe (approved):

PR #4698, opened 2025-03-21
[Connect] Fix bug where delegate callbacks would not fire on dismiss.

nschris-stripe:
## Summary
This fixes an issue where the accountOnboardingDidExit callback would not fire on dismiss. I have also added UI Tests to cover both the accountOnboardingDidExit and didFailLoadWithError to ensure no further regressions.

## Demo UI Tests

https://github.com/user-attachments/assets/3cfeb01c-0441-42e7-bea7-2513139783e7

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
-## x.x.x x-x-x
+## 24.9.0 2025-03-24
+### Connect (Private Preview)
+* [Fixed] Fixed an issue where Account Onboarding exit events were not being properly emitted.
+
 ### PaymentSheet
 * [Fixed] Embedded Payment Element (private beta) possibly showing empty forms after calling `update()`.
 
--- a/Example/StripeConnectExample/StripeConnectExample/Helpers/ToastManager.swift
+++ b/Example/StripeConnectExample/StripeConnectExample/Helpers/ToastManager.swift
@@ -1,0 +1,97 @@
+//
+//  ToastManager.swift
+//  StripeConnect Example
+//
+//  Created by Chris Mays on 3/20/25.
+//
+
+import SwiftUI
+import UIKit
+
+struct ToastMessage: Identifiable, Equatable {
+    let id = UUID()
+    let message: String
+    let createdAt = Date()
+}
+
+class ToastManager: ObservableObject {
+    static let shared = ToastManager()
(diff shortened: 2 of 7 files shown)


Example/StripeConnectExample/StripeConnectExample/Helpers/ToastManager.swift
```diff
@@ -0,0 +1,97 @@
...
+import SwiftUI
+import UIKit
+
+struct ToastMessage: Identifiable, Equatable {
+    let id = UUID()
+    let message: String
+    let createdAt = Date()
+}
+
+class ToastManager: ObservableObject {
```
nschris-stripe, line 17: Toast manager was mostly created so that UI tests have something to key off of when delegate methods are called, but it is also very useful for manually testing that the delegate callbacks work.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 24.9.0 2025-03-24
+### Connect (Private Preview)
+* [Fixed] Fixed an issue in Account Onboarding where accountOnboardingDidExit did not get called.
```
mats-stripe, line 3: Maybe something a little less specific for the release note? Not sure what the general thoughts are on this though. 

```
* [Fixed] Fixed an issue where Account Onboarding exit events were not being properly emitted.
```

Example/StripeConnectExample/StripeConnectExample/Helpers/ToastManager.swift
```diff
@@ -0,0 +1,97 @@
...
+    static let shared = ToastManager()
+    @Published var toasts: [ToastMessage] = []
+    private let displayDuration: TimeInterval = 20.0
```
mats-stripe, line 20: 20 seconds feels quite long for toasts 😅 I would expect 3-5s

mats-stripe (approved):

Example/StripeConnectExample/StripeConnectExample/Helpers/ToastManager.swift
nschris-stripe: Yeah I agree, I have it this way to make it so the UITests are less flakey. I think I can tweak this over time or change the value based on whether tests are running.

mats-stripe (approved):

PR #4700, opened 2025-03-21
Fix embedded not showing forms after updates

porter-stripe:
## Summary
- Fixed a bug where sometimes after updating form would show up as empty.
- This was caused by forms attempting to be added to different/multiple view controller hierarchies, since instances are shared through the cache.

## Motivation
- Bug fix

## Testing
- Updated UI test to catch this

## Changelog
See diff

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## x.x.x x-x-x
+### PaymentSheet
+* [Fixed] Embedded Payment Element possibly showing empty forms after calling `update()`.
+
 ## 24.8.0 2025-03-17
 ### PaymentSheet, CustomerSheet
 * [Added] Support for default payment methods in private beta.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## x.x.x x-x-x
 ### PaymentSheet
-* [Fixed] Embedded Payment Element possibly showing empty forms after calling `update()`.
+* [Fixed] Embedded Payment Element (private beta) possibly showing empty forms after calling `update()`.
 
 ## 24.8.0 2025-03-17
 ### PaymentSheet, CustomerSheet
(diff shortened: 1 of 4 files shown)


porter-stripe:
@yuki-stripe I punted on the assert. I found it pretty tricky to correctly assert this, since there are cases in our code where it's somewhat valid for the form to already be part of another view before we add it to a different one. We might need to rethink some other aspects of our approach to prevent this in the future. However, I did add a test to cover this scenario.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## x.x.x x-x-x
+### PaymentSheet
+* [Fixed] Embedded Payment Element (private beta) possibly showing empty forms after calling `update()`.
```
```suggestion
* [Fixed] Embedded Payment Element (private beta) possibly showing empty forms after calling `update()` and switching between intent configuration modes.
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -205,7 +205,7 @@ extension EmbeddedPaymentElement: EmbeddedPaymentMethodsViewDelegate {
             paymentMethodType: paymentMethodType,
             previousPaymentOption: nil,
             analyticsHelper: analyticsHelper,
-            formCache: formCache,
+            formCache: .init(),  // Use a fresh form cache to ensure forms aren't re-added to a different view controller's hierarchy
```
yuki-stripe, line 208: WDYT about not creating the Form VC in the implementation of this function entirely? I think I dismissed it earlier thinking it would be annoying to pass in all the params to the form factory, but it's not so bad:

```
        // Show an animation on the label if the payment method shows a form
        if let form = self.formCache[paymentMethodType] {
            return form.collectsUserInput
        } else {
            let form = PaymentSheetFormFactory(
                intent: intent,
                configuration: .paymentSheet(configuration),
                paymentMethod: paymentMethodType,
                previousCustomerInput: nil,
                linkAccount: LinkAccountContext.shared.account,
                accountService: LinkAccountService(apiClient: configuration.apiClient, elementsSession: elementsSession),
            ).make()
            return form.collectsUserInput
        }
```

yuki-stripe (approved):

PR #4712, opened 2025-03-25
[FCLite] Make FC Lite available in MPE

mats-stripe:
## Summary

This integrates FC Lite into MPE. The main idea here is we first check if the full SDK is available, use that if so, and otherwise fallback on FC Lite.

> [!IMPORTANT]  
>This PR does not release FC Lite. There is a guardrail in place (`fcLiteFeatureEnabled`) that prevents it from being accessed, and this is only currently enabled by our example apps.

## Testing

Manual testing done! Here's the Payment Sheet example app: 

https://github.com/user-attachments/assets/d9c7f0ef-fcff-48ea-b5fd-ed89b64f36fb

And the Financial Connections example app:

https://github.com/user-attachments/assets/e47df2f8-362e-49d0-a7db-0b5f401d8c8b

## Changelog

N/a

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
@@ -491,6 +491,10 @@ class PlaygroundController: ObservableObject {
             // Hack to enable incentives in Instant Debits
             let enableInstantDebitsIncentives = newValue.instantDebitsIncentives == .on
             UserDefaults.standard.set(enableInstantDebitsIncentives, forKey: "FINANCIAL_CONNECTIONS_INSTANT_DEBITS_INCENTIVES")
+
+            let enableFcLite = newValue.fcLiteEnabled == .on
+            FinancialConnectionsSDKAvailability.fcLiteFeatureEnabled = enableFcLite
+            FinancialConnectionsSDKAvailability.shouldPreferFCLite = enableFcLite
         }.store(in: &subscribers)
 
         // Listen for analytics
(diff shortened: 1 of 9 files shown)


Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
```diff
@@ -884,7 +888,7 @@ extension PlaygroundController {
             else {
                 if let data = data,
                    (response as? HTTPURLResponse)?.statusCode == 400 {
-                    let errorMessage = String(data: data, encoding: .utf8)!
+                    let errorMessage = String(decoding: data, as: UTF8.self)
```
mats-stripe, line 891: Drive-by fix. Our linter was yelling about this line

StripePaymentSheet/StripePaymentSheet/Source/Internal/FC Lite/FCLiteImplementation.swift
```diff
@@ -0,0 +1,33 @@
...
+//  FCLiteImplementation.swift
+//  StripePaymentSheet
+//
+//  Created by Mat Schmid on 2025-03-25.
+//
+
+@_spi(STP) import StripeCore
+import UIKit
+
+/// NOTE: If you change the name of this class, make sure to also change it `FinancialConnectionsSDKAvailability` file.
```
tillh-stripe, line 11:
```suggestion
/// NOTE: If you change the name of this class, make sure to also change it in the `FinancialConnectionsSDKAvailability` file.
```

tillh-stripe, line 11: Can we add a test for this, where we test that `StripePaymentSheet.FCLiteImplementation` is available?

mats-stripe: Done in https://github.com/stripe/stripe-ios/pull/4712/commits/58d8ed923d160e186bc96bb604dc401df30209b4

tillh-stripe (approved):

PR #4714, opened 2025-03-25
Fix CardScan animation

davidme-stripe:
## Summary
This "Scan Card" animation has been bugging me: It overlaps the other fields as it animates out.

* To fix this, I pinned it to the top of a container view, unpinned the bottom, and enabled clipsToBounds. It'll now stay stuck to the top of the container while the container's size animates down to 0.
* I also moved the disabling of the captureSession to *after* the animation kicks off to make closing the view a little more responsive, as disconnecting the camera input blocks the main thread for a bit.
* Also switched the animation to use a basic spring.

| Before | After |
| ------ | ----- |
| CleanShot 2025-03-25 at 14 41 10 | CleanShot 2025-03-25 at 14 39 29 |

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-1302

## Testing
Tested in simulator and on an iPhone 16 Pro

## Changelog
Not significant enough

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Helpers/STPCardScanner.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Helpers/STPCardScanner.swift
@@ -461,7 +461,7 @@ class STPCardScanner: NSObject, AVCaptureVideoDataOutputSampleBufferDelegate {
         if let startTime = startTime {
             duration = Date().timeIntervalSince(startTime)
         }
-        
+
         DispatchQueue.main.async(execute: {
             if params == nil {
                 STPAnalyticsClient.sharedClient.logCardScanCancelled(withDuration: duration ?? 0.0)
@@ -474,7 +474,7 @@ class STPCardScanner: NSObject, AVCaptureVideoDataOutputSampleBufferDelegate {
                 self.cameraView?.captureSession = nil
             }
         })
-        
+
         isScanning = false
         captureDevice?.unlockForConfiguration()
         captureSession?.stopRunning()
--- a/StripePaymentSheet/StripePaymentSheet/Source/Helpers/STPCardScanner.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Helpers/STPCardScanner.swift
@@ -461,21 +461,23 @@ class STPCardScanner: NSObject, AVCaptureVideoDataOutputSampleBufferDelegate {
         if let startTime = startTime {
             duration = Date().timeIntervalSince(startTime)
         }
-        isScanning = false
-        captureDevice?.unlockForConfiguration()
-        captureSession?.stopRunning()
-
+        
         DispatchQueue.main.async(execute: {
             if params == nil {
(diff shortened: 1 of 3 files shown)


sfriedman-stripe:
Much better! For bonus points: Maybe it's just my eye, but it feels like there's still a bit of a "flash" when CardScan closes. I don't know how to easily slow the GIF down, but speculating: are we possibly fading out the view as we collapse it? Maybe we should leave the opacity alone?

davidme-stripe:
Good call: It's partially the alpha animation, but also that some of the layer properties don't animate correctly with setHidden. We can throw a snapshot view over it to hide that. Still looks a bit off when the black camera border folds into the bottom of the expiration date / CVC form:

CleanShot 2025-03-26 at 21 35 08

But if we use a snapshot view *and* keep the alpha animation, it's a bit better:

CleanShot 2025-03-26 at 21 40 00

sfriedman-stripe:
I'm probably partial to the first one that doesn't touch the alpha at all because it looks like less is happening at once, but I leave it up to you!

davidme-stripe:
Makes sense, I've been staring at it for too long at this point

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/CardScanningView.swift
```diff
@@ -163,6 +169,22 @@ class CardScanningView: UIView, STPCardScannerDelegate {
         self.stop()
     }
 
+    var snapshotView: UIView?
+
+    // The shape layers don't animate cleanly during setHidden,
+    // so let's use a snapshot view instead.
+    func prepDismissAnimation() {
+        if let snapshot = self.containerView.snapshotView(afterScreenUpdates: true) {
```
wooj-stripe, line 177: Whoa!! nice!

StripePaymentSheet/StripePaymentSheet/Source/Helpers/STPCardScanner.swift
```diff
@@ -471,10 +471,11 @@ class STPCardScanner: NSObject, AVCaptureVideoDataOutputSampleBufferDelegate {
...
                 STPAnalyticsClient.sharedClient.logCardScanSucceeded(withDuration: duration ?? 0.0)
             }
-            self.feedbackGenerator = nil
-
-            self.cameraView?.captureSession = nil
             self.delegate?.cardScanner(self, didFinishWith: params, error: error)
+            self.feedbackGenerator = nil
+            DispatchQueue.main.async {
+                self.cameraView?.captureSession = nil
+            }
```
wooj-stripe, line 478: I'm curious why this was needed

wooj-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/Helpers/STPCardScanner.swift
davidme-stripe: Mentioned above:
> I also moved the disabling of the captureSession to after the animation kicks off to make closing the view a little more responsive, as disconnecting the camera input blocks the main thread for a bit.

PR #4719, opened 2025-03-25
Improve expiration date scanning

davidme-stripe:
## Summary
Due to improvements in iOS' OCR and changes to modern cards, we're often seeing strings like `Valid Thru 01/30 Security Code 123`, which are not parsed by the existing logic. To catch these, we'll add another check to look for dates in a "00/00" format.

Before: 5/14 test cards successfully populated an expiration date on scan
After: 14/14 test cards succeeded. Scanning was marginally faster, as we don't have to wait for the timeout

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-3369

## Testing
Tested on iPhone 16 Pro, iOS 18.3.2

## Changelog
Adding


StripePaymentsUI/StripePaymentsUI/Source/Helpers/STPStringUtils.swift
```diff
@@ -66,4 +66,28 @@ import Foundation
         ).first
         return result != nil && (result?.numberOfRanges ?? 0) > 0
     }
+
+    static let slashFormattedExpirationDateRegex: NSRegularExpression? = {
+        return try? NSRegularExpression(pattern: #"\b(\d{2})[/\.](\d{2,4})\b"#)
```
wooj-stripe, line 71: `[/\.]`
this means there is a `/` or a `.` ?  if so, seems like we should have a test for this?

StripePaymentsUI/StripePaymentsUI/Source/Helpers/STPStringUtils.swift
```diff
@@ -66,4 +66,28 @@ import Foundation
...
+    }()
+
+    /// Returns a sanitized expiration date (in "0101" format) from a string.
+    /// This differs from the existing expiration date parser, as it only looks for dates formatted by slashes.
+    /// This is only intended for OCR use, as we'll often get messy strings like "Exp 01/20 Verification Code 123"
+    @_spi(STP) public class func sanitizedExpirationDateFromOCRString(_ string: String) -> String? {
+        guard let regex = slashFormattedExpirationDateRegex,
+              let match = regex.firstMatch(in: string, range: NSRange(string.startIndex..., in: string)),
+              match.numberOfRanges >= 3,
```
wooj-stripe, line 80: why >= 3? shouldn't it be == 3?

Stripe/StripeiOSTests/STPStringUtilsTest.swift
```diff
@@ -139,4 +139,12 @@ class STPStringUtilsTest: XCTestCase {
         XCTAssertEqual(STPStringUtils.expirationDateString(from: " "), " ")
         XCTAssertEqual(STPStringUtils.expirationDateString(from: "12/"), "12/")
     }
+
+    func testSanitizedExpirationDateFromOCRString() {
+        XCTAssertEqual(STPStringUtils.sanitizedExpirationDateFromOCRString("12/1995"), "1295")
```
wooj-stripe, line 144: Maybe would be good to add some cases that return nil. e.g. `aa/bb`

StripePaymentsUI/StripePaymentsUI/Source/Helpers/STPStringUtils.swift
davidme-stripe: I think it's okay if we match more dates and pick the first one, but I'll add a test for this

StripePaymentsUI/StripePaymentsUI/Source/Helpers/STPStringUtils.swift
davidme-stripe: Eh, I think we should only do `/` for now, I can't actually find any cards where we'd use a `.`. It's just more specific than the old one (which allows for any 1-3 non-digit characters) in between

wooj-stripe (approved):

PR #4720, opened 2025-03-25
Animation jank when US bank account is selected and then deleted

joyceqin-stripe:
## Summary

Wait for collectionView to finish updating its layout before dismissing the update screen
## Motivation

[MOBILESDK-3289](https://jira.corp.stripe.com/browse/MOBILESDK-3289)
## Testing

- Manually

Remove a selected US bank account in horizontal mode (most noticeable in FlowController)
Before:

https://github.com/user-attachments/assets/b20e90b4-681f-4324-9deb-cfadf01416f4

After:

https://github.com/user-attachments/assets/24de5268-252e-4c4c-aba9-04c48fb0d518

## Changelog

N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentOptionsViewController.swift
```diff
@@ -668,8 +669,9 @@ extension SavedPaymentOptionsViewController: UpdatePaymentMethodViewControllerDe
         }
         // if it isn't the last saved pm, waiting for update pm screen dismissal results in a weird flash, so we do it like this
```
wooj-stripe, line 670: Is this comment still relevant?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentOptionsViewController.swift
```diff
@@ -610,7 +610,7 @@ extension SavedPaymentOptionsViewController: PaymentOptionCellDelegate {
         self.bottomSheetController?.pushContentViewController(editVc)
     }
 
-    private func removePaymentMethod(_ paymentMethod: STPPaymentMethod) {
+    private func removePaymentMethod(_ paymentMethod: STPPaymentMethod, completion: (() -> Void)? = nil) {
```
wooj-stripe, line 613: It's possible to return early, in which case, the completion block won't execute. Maybe we can turn this function into `async throws` and use `withCheckedThrowingContinuation`? If the function is unable to continue, we use ` continuation.resume(throwing:`, and in the event we want to succeed, we use ` continuation.resume()`

From the caller's perspective, it looks like:
```
            Task {
                try await removePaymentMethod(paymentMethod)
                 _ = self.bottomSheetController?.popContentViewController()
             }
```
Seems clearer this way in that if payment method removal fails, then it's clear to the caller that we won't be popping the content view controller.

wdyt?

joyceqin-stripe: I agree that seems clearer, but removing/detaching payment methods isn't async and doesn't throw— would we want to change that? Also, if it returns early, wouldn't we not want to pop the content view controller anyways?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentOptionsViewController.swift
joyceqin-stripe: Yeah, still relevant. It used to be that the whole didRemove func was just removePaymentMethod(paymentMethod)
_ = viewController.bottomSheetController?.popContentViewController()
But there was animation jank when removing the last pm because it would try to switch to the add screen and dismiss the update screen around the same time, so the fix was to wait for the update screen to finish dismissing and then switch. But dismissing the update screen before removing led to a weird flash, so we needed different order of operations for last vs not last.

For this particular issue, the jank comes from the update screen being dismissed before the UI is finished updating (hiding the mandate, button, etc). So we want to wait to dismiss the update screen to give the UI time to update so the user doesn't see it

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentOptionsViewController.swift
wooj-stripe: Ah, right! -- i suppose we do want to pop even if we can't remove the view model, oops!  Yea, i can't think of a better way here -- completion seems fine -- lets just make sure it to call it in the guard block before returning.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentOptionsViewController.swift
wooj-stripe: Ideally, we'd fix the actual jank, but can understand fixing it this way in the short term. Maybe we open a ticket against ourselves to come up with a better fix later? wdyt?

joyceqin-stripe: What would be fixing the actual jank, like making the mandate and button disappear more smoothly?

wooj-stripe (approved):

PR #4721, opened 2025-03-26
Allowing JP and CN characters in name field

wooj-stripe:
## Summary
Setting a UITextField's attributedText field apparently cancels the ability to type most hiragana and katakana characters.  For example, it's impossible to any hiragana represented as two(or more) roman characters: "ko"(こ) , "ni"　（に）, "chi"　（ち）, "wa"　（わ） but possible to input single roman characters that represent "a" (あ）, "i'（い）, "e"　（え）,  "u"（う）"o"（お).

Note that this is also possible with Hiragana keyboard (Shaped like a T9).  Attempt to type "ta" (た)  and then you'll realize you're unable to access the modifiers which make it:  "da" (だ).

## Motivation
Unable to input characters.

## Testing
Manually. The testing story here isn't great.  Unit tests don't easy/useful... I can try to write some ui tests (but would require setting the emulator's keyboard to JP), which i'm not quite sure how to do right now.

### DIFF
--- a/StripeUICore/StripeUICore/Source/Elements/TextField/TextFieldView.swift
+++ b/StripeUICore/StripeUICore/Source/Elements/TextField/TextFieldView.swift
@@ -193,7 +193,14 @@ class TextFieldView: UIView {
         // the same relative position in case attributedText adds more characters
         let cursorOffsetFromEnd = textField.selectedTextRange.map { textField.offset(from: textField.endOfDocument, to: $0.end) }
 
-        textField.attributedText = viewModel.attributedText
+        // Don't mess with attributed text if the IME is currently in progress (Japanese/Chinese/Hindi characters)
+        // Note: Setting textField.attributedText cancels the IME
+        if textField.markedTextRange != nil {
+            textField.text = viewModel.attributedText.string
+        } else {
+            textField.attributedText = viewModel.attributedText
+        }
+
         if let cursorOffsetFromEnd = cursorOffsetFromEnd,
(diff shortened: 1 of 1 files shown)


StripeUICore/StripeUICore/Source/Elements/TextField/TextFieldElement.swift
```diff
@@ -87,9 +88,11 @@ import UIKit
                 return String(format: localized, configuration.label)
             }
         }()
+
         return ViewModel(
             placeholder: placeholder,
             accessibilityLabel: configuration.accessibilityLabel,
+            shouldUseAttributedText: configuration.shouldUseDisplayText,
```
joyceqin-stripe, line 95: It feels confusing to me to have shouldUseAttributedText and shouldUseDisplayText, especially when you set the value for shouldUseAttributedText to be configuration.shouldUseDisplayText. What's the difference between attributed text and display text?

wooj-stripe: Great call -- I wrestled with this as well, but ultimately went with "shouldUseDisplayText" because we are passing `configuration.makeDisplayText(for: text)` in as parameter for `attributedText:`.

> What's the difference...
Based on my read of the code, their values are effectively the same.  From the perspective of the ViewModel, it is just attributedText (generic), but from the perspective of the the TextFieldElement it's the displayText.  Not sure how the code got this way, but could easily see attributedText getting renamed to displayText.

davidme-stripe (approved):

PR #4735, opened 2025-03-27
Replace optional NSNumber with Bool

porter-stripe:
## Motivation
- v25

## Testing
- CI


StripePayments/StripePayments/Source/API Bindings/Models/PaymentIntents/STPPaymentIntentParams.swift
```diff
@@ -79,11 +79,18 @@ public class STPPaymentIntentParams: NSObject {
...
+        get { return NSNumber(value: savePaymentMethod) }
+        set { savePaymentMethod = newValue?.boolValue ?? false }
+    }
 
-    /// `@YES` to set this PaymentIntent’s PaymentMethod as the associated Customer's default
-    /// This should be a boolean NSNumber, so that it can be `nil`
-    @objc @_spi(STP) public var setAsDefaultPM: NSNumber?
+    /// `true` to save this PaymentIntent’s PaymentMethod or Source to the associated Customer,
+    /// if the PaymentMethod/Source is not already attached.
+    public var savePaymentMethod: Bool = false
```
yuki-stripe: This should be nullable right? I'm also trying to figure out if we should just remove/deprecate this, I don't see it in the public API anymore and I think it's superseded by setup_future_usage.

porter-stripe: Yeah agree this should be nullable.

StripePayments/StripePayments/Source/API Bindings/Models/PaymentIntents/STPPaymentIntent.swift
```diff
@@ -118,7 +118,13 @@ public class STPPaymentIntent: NSObject {
     @objc public let status: STPPaymentIntentStatus
 
     /// The list of payment method types (e.g. `[NSNumber(value: STPPaymentMethodType.card.rawValue)]`) that this PaymentIntent is allowed to use.
-    @objc public let paymentMethodTypes: [NSNumber]
+    @objc(paymentMethodTypes)
+    public var paymentMethodTypes_objc: [NSNumber] {
```
yuki-stripe, line 122: Let's also add 
`@available(swift, obsoleted: 1.0, renamed: "paymentMethodTypes")`

to here and others.

porter-stripe: Not sure I agree, the only people who will change their integration to use suffixed "_objc" properties already know they need it? So we would just be telling objc users to not use it when they've already chosen to use it b/c they need it.

yuki-stripe: The `swift` in the `@available` means it the attribute only applies to Swift, so it won't affect objc users.

porter-stripe: Ah I see. I'm having difficulty doing this, not sure it's feasible for us. 

Example:

```swift
// STPPaymentIntentParams.swift

    @available(swift, obsoleted: 1.0, renamed: "savePaymentMethod")
    @objc(savePaymentMethod)
    public var savePaymentMethod_objc: NSNumber? {
        get {
            guard let savePaymentMethod else { return nil }
            return NSNumber(value: savePaymentMethod)
        }
        set { savePaymentMethod = newValue?.boolValue }
    }

    public var savePaymentMethod: Bool?

    @objc
    public class func propertyNamesToFormFieldNamesMapping() -> [String: String] {
        return [
            NSStringFromSelector(#selector(getter: savePaymentMethod_objc)): "save_payment_method",
        ]
    }

// the function errors with 'savePaymentMethod_objc' has been renamed to 'savePaymentMethod'
// we can't use savePaymentMethod directly b/c we need @objc compatibility 
```

yuki-stripe: Ah that's annoying, we need to add a `@objc internal var _savePaymentMethod: NSNumber?` similar to  `setupFutureUsageRawString` in this file.

But actually looking into this property more, this was silently removed I think before PI went to GA. You can see it removed in [2020 in stripe-java](https://github.com/stripe/stripe-java/commit/695199502d907a322cad0178ff31aae6995f717c#diff-fe44d0dd4819e2def9c845a77b02bd987eda62960f643f6c5a527a6d884fb9f8L109) for example. It looks like there are still [users](https://hubble.corp.stripe.com/saved-queries/60fbf427-cfe2-41c8-83e8-07b3042c0d02?filter-stripe_client_user_agent=&filter-api_method=&filter-bindings_language=obje), oddly, so let's just mark it as deprecated for now and recommend setupFutureUsage instead.

porter-stripe: Adding deprecations with `@available(*, deprecated, message: "Use setupFutureUsage instead")` make our own builds fail :( 

'savePaymentMethod' is deprecated: Use setupFutureUsage instead

I'm voting to just leave this property as-is, unless we want to just totally remove it.

yuki-stripe: Ok, let's leave alone for now!

StripePayments/StripePayments/Source/API Bindings/Models/PaymentIntents/STPPaymentIntentParams.swift
```diff
@@ -79,7 +79,18 @@ public class STPPaymentIntentParams: NSObject {
...
+        get {
+            guard let savePaymentMethod else { return nil }
+            return NSNumber(value: savePaymentMethod)
+        }
+        set { savePaymentMethod = newValue?.boolValue }
+    }
+
+    /// `true` to save this PaymentIntent’s PaymentMethod or Source to the associated Customer,
+    /// if the PaymentMethod/Source is not already attached.
+    public var savePaymentMethod: Bool?
```
yuki-stripe: I would just undo this, let's not touch `savePaymentMethod` here

CHANGELOG_v25_draft.md
```diff
@@ -28,6 +28,10 @@ This major version introduces many small breaking changes. Please see [MIGRATING
 * [Changed] Replaces the `ApplePayContextDelegate.didCreatePaymentMethod` method with an async version.
 
 ### Payments
+* [Changed] `STPPaymentIntent.paymentMethodTypes` and `STPSetupIntent.paymentMethodTypes` now return `[STPPaymentMethodType]` instead of `[NSNumber]`. Objective-C callers can use the `paymentMethodTypes_objc` property to access the legacy `[NSNumber]` representation.
+* [Changed] `STPPaymentIntentParams.savePaymentMethod` now uses `Bool?` instead of `NSNumber?` for better Swift ergonomics. Objective-C callers can continue using the `savePaymentMethod_objc` property which returns `NSNumber?`.
+* [Changed] `STPSetupIntentConfirmParams.useStripeSDK` now uses `Bool?` instead of `NSNumber?` for better Swift ergonomics. Objective-C callers can continue using the `useStripeSDK_objc` property which returns `NSNumber?`.
```
yuki-stripe:
```suggestion
* [Changed] `STPPaymentIntent.paymentMethodTypes` and `STPSetupIntent.paymentMethodTypes` now return `[STPPaymentMethodType]` instead of `[NSNumber]` in Swift for better ergonomics.
* [Changed] `STPSetupIntentConfirmParams.useStripeSDK` now uses `Bool?` instead of `NSNumber?` in Swift for better ergonomics.
```

> Objective-C callers can use the `*_objc` property to access the legacy `[NSNumber]` representation

I don't think this is true, obj-c users don't see the `_obc` property because it's renamed in objc to `paymentMethodTypes`. Nothing changes in this PR for them.

yuki-stripe: We should probably add `STPSetupIntentConfirmParams.useStripeSDK` too but I can do that in a follow-up PR

PR #4737, opened 2025-03-27
Edit/Done buttons primary color instead of grey

joyceqin-stripe:
## Summary

Made the navbar's additionalButton the primary color by default instead of the icon color
## Motivation

https://jira.corp.stripe.com/browse/MOBILESDK-3275
## Testing

- Snapshot tests

## Changelog

N/A


porter-stripe (approved):
LGTM, what was the outcome of the secondary action color? Are we gonna add it?

joyceqin-stripe:
> LGTM, what was the outcome of the secondary action color? Are we gonna add it?

I don't know if we're going to add it, but I talked to JJ yesterday and she said that for now at least, we'll make the buttons primary. I think we may revisit the matter of the secondary action color, since this was the use case for which secondary action was intended. I don't remember why we didn't end up implementing it, but I remember there was a reason.

PR #4742, opened 2025-03-28
Increase area of back button tap targeting

wooj-stripe:
## Summary
Increases the tappable area of the back button

## Motivation
back button is too hard to tap!

## Testing
Manual testing and relying on all the snapshot tests in `PaymentSheetVerticalViewControllerSnapshotTest`

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/SheetNavigationBar.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/SheetNavigationBar.swift
@@ -28,6 +28,7 @@ class SheetNavigationBar: UIView {
         stack.alignment = .center
         return stack
     }()
+    // Used for allowing larger tap area to the left of closeButtonLeft
     fileprivate lazy var dummyView: UIView = {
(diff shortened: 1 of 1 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/SheetNavigationBar.swift
```diff
@@ -22,10 +22,16 @@ class SheetNavigationBar: UIView {
...
+        let stack = UIStackView(arrangedSubviews: [dummyView, closeButtonLeft, backButton, testModeView])
         stack.spacing = PaymentSheetUI.defaultPadding
+        stack.setCustomSpacing(PaymentSheetUI.navBarPadding, after: dummyView)
+        stack.alignment = .center
         return stack
     }()
+    fileprivate lazy var dummyView: UIView = {
+        let dummyView = UIView(frame: .zero)
+        return dummyView
+    }()
```
porter-stripe, line 34: Looks good, but if someone see's this in the future they might not understand what this dummyView does. Can we add a comment?

porter-stripe (approved):

PR #4752, opened 2025-03-31
Add RowSelectionBehavior API

porter-stripe:
## Summary
- Add RowSelectionBehavior API
- Throws an error if a merchant is using immediateAction and confirm while using Apple Pay or Customer. 

## Motivation
- https://docs.google.com/document/d/1_LyLbnJV0KnHPp739xXUYAFr9bdjRvxp-Mh20Y2ZmWM/edit?tab=t.0

## Testing
- Added unit tests

## Changelog
N/A (will mention this API in the changelog once it is totally implemented)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -151,6 +151,9 @@ extension EmbeddedPaymentElement: EmbeddedPaymentMethodsViewDelegate {
     func embeddedPaymentMethodsViewDidTapPaymentMethodRow() {
         guard let selectedFormViewController else {
             // If the current selection has no form VC, there's nothing to do
```
wooj-stripe, line 153: We should update this comment. Specifically we should mention why there is no selectedFormViewController when in immediateAction mode

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -438,6 +441,9 @@ extension EmbeddedPaymentElement: EmbeddedFormViewControllerDelegate {
             }
         }
         embeddedFormViewController.dismiss(animated: true)
+        if case .immediateAction(let didSelectPaymentOption) = configuration.rowSelectionBehavior {
+            didSelectPaymentOption()
+        }
```
wooj-stripe, line 446: Should this work be done in the completion of dismissing?

Also, just to be clear, didSelectPaymentOption() always gets fired while `informDelegateIfPaymentOptionUpdated` works ONLY when the payment method changes, correct?  Feels awkward to sometimes communicate paymentOption via a callback sometimes, and sometimes through a delegate, but i'm guessing it was already discussed during api review?

porter-stripe: re: Should this work be done in the completion of dismissing?

That's an interesting idea, I don't feel too strongly but my general sense is that as soon as the user taps "Continue" we should invoke the callback. This is also consistent with how we inform the delegate, so I'm inclined to leave as-is.

re: but i'm guessing it was already discussed during api review?

Yep, there were some iterations on this, but landed on this approach.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
```diff
@@ -67,6 +67,14 @@ public final class EmbeddedPaymentElement {
         intentConfiguration: IntentConfiguration,
         configuration: Configuration
     ) async throws -> EmbeddedPaymentElement {
+        if case .immediateAction = configuration.rowSelectionBehavior,
+           case .confirm = configuration.formSheetAction,
+            configuration.applePay != nil || configuration.customer != nil {
```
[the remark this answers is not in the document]
porter-stripe: Thanks for the suggestion! This suggested expression is a little bit different than what is in the PR. The suggestion will throw an error if the merchant is NOT using `confirm` w/ `immediateAction` (among other things).

I did try to make it easier to read by moving into a function as suggested, adding a comment, and some parenthesis. Lmk what you think.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
porter-stripe: Ah yeah, updated!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -543,6 +549,17 @@ extension EmbeddedPaymentElement {
             stpAssertionFailure("3DS2 was triggered unexpectedly")
         })
     }
+
+    static func validateRowSelectionConfiguration(configuration: Configuration) throws {
+        // Fail init if the merchant is using immediateAction and confirm form sheet action along w/ either a Customer or Apple Pay configuration
+        if case .immediateAction = configuration.rowSelectionBehavior,
+           case .confirm = configuration.formSheetAction,
+            configuration.applePay != nil || configuration.customer != nil {
```
wooj-stripe, line 557: how about:
```
        if case .immediateAction = configuration.rowSelectionBehavior {
            guard case .continue = configuration.formSheetAction,
                  configuration.customer == nil,
                  configuration.applePay == nil else {
                throw PaymentSheetError.integrationError(nonPIIDebugDescription: "Using .immediateAction with .confirm form sheet action is not supported when Apple Pay or a customer configuration is provided. Use .default row selection behavior or disable Apple Pay and saved payment methods.")
            }
    ```

File: StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift (hunk 3 above)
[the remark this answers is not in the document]
Comment (wooj-stripe) [reply StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift]: Sorry, was a typo.. i intended that case to be `.continue`

File: StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift (hunk 4 above)
Comment (porter-stripe) [reply StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift:557]: I think this has a similar issue, but in reverse now. Merchants are allowed to use the `confirm` form sheet action, so long as they aren't using a Customer or Apple Pay.

File: StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentElementTest.swift [hunk 5]
```diff
@@ -657,7 +705,118 @@ class EmbeddedPaymentElementTest: XCTestCase {
         // ...the delegate should be called after the form is submitted
         XCTAssertTrue(delegateDidUpdatePaymentOptionCalled, "Delegate should be updated after card form is completed")
         XCTAssertEqual(sut.paymentOption?.label, "•••• 4242")
+
+        await fulfillment(of: [rowSelectionBehaviorExpectation])
+    }
+
+    func testCreateFails_whenImmediateActionWithConfirmAndApplePay() async throws {
+        // Given a config that has rowSelectionBehavior = immediateAction, formSheetAction = .confirm, and Apple Pay
+        var config = configuration
```
wooj-stripe, line 714: i'm confused here, because the computed variable `configuration` is:
```
    lazy var configuration: EmbeddedPaymentElement.Configuration = {
        var config = EmbeddedPaymentElement.Configuration._testValue_MostPermissive(isApplePayEnabled: false)
        config.apiClient = STPAPIClient(publishableKey: STPTestingDefaultPublishableKey)
        return config
    }()
```

but when i run it, apple pay configuration is set

porter-stripe: ah, `config` starts as a copy of `self.configuration`, but below on 719 is updated to add apple pay

wooj-stripe (approved):

PR #4766, opened 2025-04-01
[Connect] Fix loading indicator jump when transitioning from native loading to web loading.

nschris-stripe:
## Summary
This is an adaptation of the ActivityIndictor provided in StripeUICore. The main differences are:
- Design tweaks to match what connect embed displays
- Timing function change to precisely sync with the web component for seamless transitions.

### Before

https://github.com/user-attachments/assets/e9231764-fb3d-4ec7-9412-2a536dbbd0c6

### After

https://github.com/user-attachments/assets/64906477-98ad-46f4-8739-fa8c50e824f5


mats-stripe:
The `After` video looks soooo good! 👏 Just curious, will this all still lineup no matter the device / screen size?

nschris-stripe:
It should still line up. There is still some work to be done on web to ensure it matches with all text sizes etc but we agreed on a general contract for how it should be displayed.

mats-stripe (approved):
Amazing work!

PR #4769, opened 2025-04-02
Clear payment option for immediateAction

porter-stripe:
## Summary
- When rowSelectionBehavior = immediateAction and action = confirm, we also:
   - Call clearPaymentOption after successful loads/updates, and after confirmation cancels or fails.

## Motivation
- https://docs.google.com/document/d/1_LyLbnJV0KnHPp739xXUYAFr9bdjRvxp-Mh20Y2ZmWM/edit?tab=t.0#heading=h.o12ytfel7pea

## Testing
- Manual
- Unit tests

## Changelog
See diff

### DIFF
--- a/StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentElementTest.swift
+++ b/StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentElementTest.swift
@@ -738,6 +738,8 @@ class EmbeddedPaymentElementTest: XCTestCase {
         }
     }
 
+    // MARK: Immediate action tests
+
     func testCreateFails_whenImmediateActionWithConfirmAndCustomer() async throws {
         // Given a config that has rowSelectionBehavior = immediateAction, formSheetAction = .confirm, and a customer configuration
         var config = configuration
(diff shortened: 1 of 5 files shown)


StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentElementTest.swift
```diff
@@ -817,6 +824,88 @@ class EmbeddedPaymentElementTest: XCTestCase {
...
+        sut.selectedFormViewController?.didTapPrimaryButton()
+        XCTAssertNotNil(sut.paymentOption, "Payment option should be set after filling out the card form.")
+
+        // When we call confirm() knowing it will fail...
+        let confirmResult = await sut.confirm()
+
+        // The result should be .failed and that triggers clearPaymentOptionIfNeeded()
+        XCTAssertTrue(confirmResult.isCanceledOrFailed)
+        XCTAssertNil(sut.paymentOption,
+            "Payment option should have been cleared after a canceled/failed confirmation.")
```
wooj-stripe, line 902: Ideally, we'd have two more tests:
1. one for testing cancel which tests that sut.paymentOption is nil
2. one for testing the case where intentCreationCallback() returns successful w/ a pi, and paymentOption isn't nil.

porter-stripe: For #1 I can add this~

Unless I'm misunderstanding, for #2 we have: https://github.com/stripe/stripe-ios/blob/master/StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentElementTest.swift#L276

wooj-stripe (approved):

PR #4782, opened 2025-04-07
Add CPM invalid metric

porter-stripe:
## Summary
- Fire a metric if a merchant passing in an invalid CPM

## Motivation
- CPM metrics

## Testing
- Manual
- Difficult to write tests for this b/c of the assertionFailure right below it

## Changelog
N/A


StripeCore/StripeCore/Source/Analytics/STPAnalyticEvent.swift
```diff
@@ -288,4 +288,7 @@ import Foundation
     case assertionSucceeded = "stripeios.attest.assertion.succeeded"
     case resetKeyForAssertionError = "stripeios.attest.reset_key_for_assertion_error"
     case resetKeyForAttestationError = "stripeios.attest.reset_key_for_attestation_error"
+
+    // MARK: - Custom Payment Methods
+    case paymentSheetInvalidCPM = "mc_invalid_cpm"
```
wooj-stripe, line 293: Seems like we should have a more generic name that doesn't include "paymentSheet".

wooj-stripe (approved):

PR #4789, opened 2025-04-09
Handle QR code + web view LPMs for embedded

porter-stripe:
## Summary
- Some LPMs like PromptPay/PayNow need to present a PollingViewController after web view has dismissed to check if the payment has been completed. 
- This was an oversight and not implemented on the AuthenticationContext for embedded
- The implementation waits for the SFSafariViewController to dismiss before presenting the polling VC and also keeps a flag in sync for if we need to present the polling VC when it is dismissed to prevent accidental presentations.
- Polling VC is responsible for completing the payment

https://github.com/user-attachments/assets/b9859f74-524c-4282-a852-8a775486939e

## Motivation
- https://stripe.slack.com/archives/C01CY476ESJ/p1744164704749489?thread_ts=1743523239.980129&cid=C01CY476ESJ

## Testing
- Manual
- New UI test

## Changelog
See diff

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
@@ -623,16 +623,13 @@ extension STPAuthenticationContextWrapper: PaymentSheetAuthenticationContext {
         shouldPresentPollingVC = true
     }
 
-    func authenticationContextWillDismiss(_ viewController: UIViewController) {
+    func authenticationContextDidDismiss(_ viewController: UIViewController) {
         // The following code should only be executed if we are dismissing a SFSafariViewController
         guard viewController is SFSafariViewController else { return }
 
         if let pollingViewController = self.pollingVC {
-            // Delay presentation to allow the SFSafariViewController to dismiss completely
-            DispatchQueue.main.asyncAfter(deadline: .now() + 0.2) {
-                self._presentingViewController.present(pollingViewController, animated: true)
-                self.shouldPresentPollingVC = false
-            }
+            self._presentingViewController.present(pollingViewController, animated: true)
+            self.shouldPresentPollingVC = false
         }
     }
 
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
@@ -627,7 +627,7 @@ extension STPAuthenticationContextWrapper: PaymentSheetAuthenticationContext {
         // The following code should only be executed if we are dismissing a SFSafariViewController
         guard viewController is SFSafariViewController else { return }
 
-        if let pollingViewController = self.pollingVC {
+        if let pollingViewController = self.pollingVC, shouldPresentPollingVC {
             self._presentingViewController.present(pollingViewController, animated: true)
             self.shouldPresentPollingVC = false
         }
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
@@ -4,6 +4,7 @@
 //
 //  Created by Yuki Tokuhiro on 10/10/24.
 //
+import SafariServices
 @_spi(STP) import StripeCore
 @_spi(STP) import StripePayments
 @_spi(STP) import StripePaymentsUI
@@ -581,9 +582,14 @@ extension EmbeddedPaymentElement {
 
 final class STPAuthenticationContextWrapper: UIViewController {
     let _presentingViewController: UIViewController
+    let appearance: PaymentSheet.Appearance
 
-    init(presentingViewController: UIViewController) {
+    private var pollingVC: PollingViewController?
+    private var shouldPresentPollingVC = false
+
+    init(presentingViewController: UIViewController, appearance: PaymentSheet.Appearance) {
         self._presentingViewController = presentingViewController
+        self.appearance = appearance
         super.init(nibName: nil, bundle: nil)
     }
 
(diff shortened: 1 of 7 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -592,7 +598,44 @@ final class STPAuthenticationContextWrapper: UIViewController {
...
+
+    func authenticationContextWillDismiss(_ viewController: UIViewController) {
+        // The following code should only be executed if we are dismissing a SFSafariViewController
+        guard viewController is SFSafariViewController else { return }
+
+        if let pollingViewController = self.pollingVC {
+            // Delay presentation to allow the SFSafariViewController to dismiss completely
+            DispatchQueue.main.asyncAfter(deadline: .now() + 0.2) {
+                self._presentingViewController.present(pollingViewController, animated: true)
+                self.shouldPresentPollingVC = false
```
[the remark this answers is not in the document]
porter-stripe: Ah thanks fixed!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -592,7 +598,44 @@ final class STPAuthenticationContextWrapper: UIViewController {
...
+        shouldPresentPollingVC = true
+    }
+
+    func authenticationContextWillDismiss(_ viewController: UIViewController) {
+        // The following code should only be executed if we are dismissing a SFSafariViewController
+        guard viewController is SFSafariViewController else { return }
+
+        if let pollingViewController = self.pollingVC {
+            // Delay presentation to allow the SFSafariViewController to dismiss completely
+            DispatchQueue.main.asyncAfter(deadline: .now() + 0.2) {
```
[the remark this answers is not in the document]
porter-stripe: Yeah, thanks for probing. Added authenticationContextDidDismiss, and am now using that instead!

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
```diff
@@ -1719,6 +1719,8 @@ public class STPPaymentHandler: NSObject {
                                 // This isn't great, but UIViewController is non-nil in the protocol. Maybe it's better to still call it, even if the VC isn't useful?
                                 context.authenticationContextWillDismiss?(UIViewController())
                             }
+                            // This isn't great, but UIViewController is non-nil in the protocol. Maybe it's better to still call it, even if the VC isn't useful?
+                            self.callContextWillDismissIfNeeded(context, UIViewController())
```
wooj-stripe, line 1723: 1. I don't understand the need for instantiating `UIViewController()` why not just pass in nil ?
I'm also not sure why we replicated the comment?

2. Calling `callContextWillDismissIfNeeded`  here doesn't seem correct because the body of the function calls ` authenticationContextDidDismiss`, but right before calling this function's we've just executed `authenticationContextWillDismiss`. I'm not sure how we can be certain that the dismissal happens exactly between the execution of these lines.

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
```diff
@@ -2177,6 +2179,17 @@ public class STPPaymentHandler: NSObject {
         }
         return STPPaymentHandlerError(code: errorCode, loggingSafeUserInfo: userInfo) as NSError
     }
+
+    func callContextWillDismissIfNeeded(_ context: (any STPAuthenticationContext)?, _ viewController: UIViewController?) {
```
wooj-stripe, line 2183: Took me a while to wrap my head around the naming of this function, but if i'm correct -it literally executing context.willDismiss.. but.. seems to execute `authenticationContextDidDismiss()`

porter-stripe: It's just a typo, thanks for catching.

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
porter-stripe: 1. I just copied over the existing functionality of the willDismiss as it's nearly identical in API shape (passing the view controller). Passing in nil will just make callContextDidDismissIfNeeded no-op, so I don't think that's an option and passing nil back to the delegate doesn't make sense IMO. We could update the newly added protocol function to accept an optional UIViewController, but that inconsistency with the existing `authenticationContextWillDismiss` would seem awkward, so I leaned towards keeping the existing pattern. 

2. AFAIK, the dismiss has already happened once the `completionHandler` for the `ASWebAuthenticationSession` is invoked, so I believe it's valid here.

porter-stripe: I'm guessing we pass `UIViewController()`for the existing protocol function b/c the protocol was defined before we added this functionality. Then we got here and realized we can't pass nil unless we make a change to the public API, so we just pass in a dummy VC. Not great, but I'm not proposing to modify the public API signature.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -592,7 +598,41 @@ final class STPAuthenticationContextWrapper: UIViewController {
...
+
+    func presentPollingVCForAction(action: StripePayments.STPPaymentHandlerPaymentIntentActionParams, type: StripePayments.STPPaymentMethodType, safariViewController: SFSafariViewController?) {
+        // Initialize the polling view controller and flag it for presentation
+        self.pollingVC = PollingViewController(currentAction: action, viewModel: PollingViewModel(paymentMethodType: type),
+                                                      appearance: self.appearance, safariViewController: safariViewController)
+        shouldPresentPollingVC = true
+    }
+
+    func authenticationContextDidDismiss(_ viewController: UIViewController) {
+        // The following code should only be executed if we are dismissing a SFSafariViewController
```
wooj-stripe: "if SFSafariViewController is fully dismissed" sounds better.

wooj-stripe (approved):

PR #4791, opened 2025-04-09
Dismiss keyboard when removing payment method

wooj-stripe:
## Summary
Dismisses keyboard when a user hits the remove button to avoid animation jank

## Motivation
Has animation jank if keyboard is up.

## Testing
Manual

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
@@ -218,6 +218,12 @@ final class UpdatePaymentMethodViewController: UIViewController {
     }
 
     @objc private func removePaymentMethod() {
+        for element in self.paymentMethodForm.getAllUnwrappedSubElements() {
+            if let textElement = element as? TextFieldElement {
+                textElement.endEditing(true, continueToNextField: false)
+            }
+        }
+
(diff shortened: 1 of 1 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/UpdatePaymentMethodViewController.swift
```diff
@@ -218,6 +218,12 @@ final class UpdatePaymentMethodViewController: UIViewController {
     }
 
     @objc private func removePaymentMethod() {
+        for element in self.paymentMethodForm.getAllUnwrappedSubElements() {
+            if let textElement = element as? TextFieldElement {
+                textElement.endEditing(true, continueToNextField: false)
+            }
+        }
+
```
porter-stripe, line 226: What's the scenario where the keyboard is presented when a payment method is removed? Shouldn't the taps required to remove the payment method dismiss the keyboard?

wooj-stripe: Ah, good question -- a user can tap any text field (which presents the keyboard), then the user decides they no longer want to edit that field and attempts to delete the payment method while the keyboard is still showing.  This will introduce some rendering jank because UIAlertView dismisses keyboard without re-drawing the screen -- so i've remedied this by forcibly dismissing the keyboard before this happens.

porter-stripe (approved):

PR #4793, opened 2025-04-10
Fix Link selection issue in FlowController

tillh-stripe:
## Summary

This pull request fixes an issue in FlowController where selecting Link after having previously selected an LPM wouldn't result in the payment method being changed.

**Before**: Tapping `Pay with Link` after having filled out the card form doesn't do anything. It also discards the selection if no new selection is made.

https://github.com/user-attachments/assets/c5fff708-25e3-49f5-8ceb-9b3f383c39f5

**After**: Tapping `Pay with Link` after having filled out the card form changes the payment method to `Link`. It also correctly preserves the selection if no new selection is made.

https://github.com/user-attachments/assets/8e800e89-682d-433e-bbad-35ddcc18de0b

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetFlowControllerViewController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetFlowControllerViewController.swift
@@ -27,13 +27,13 @@ class PaymentSheetFlowControllerViewController: UIViewController, FlowController
     var selectedPaymentOption: PaymentOption? {
         switch mode {
         case .addingNew:
-            if let paymentOption = addPaymentMethodViewController.paymentOption {
+            if isHackyLinkButtonSelected {
+                return .link(option: .wallet)
+            } else if let paymentOption = addPaymentMethodViewController.paymentOption {
                 return paymentOption
             } else if let paymentOption = savedPaymentOptionsViewController.selectedPaymentOption {
                 // If no valid payment option from adding, fallback on any saved payment method
                 return paymentOption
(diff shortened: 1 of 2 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetFlowControllerViewController.swift
```diff
@@ -27,13 +27,13 @@ class PaymentSheetFlowControllerViewController: UIViewController, FlowController
     var selectedPaymentOption: PaymentOption? {
         switch mode {
         case .addingNew:
-            if let paymentOption = addPaymentMethodViewController.paymentOption {
+            if isHackyLinkButtonSelected {
```
tillh-stripe, line 30: We now check whether the Link button was pressed first. Let me know if there's any concern around that.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetFlowControllerViewController.swift
```diff
@@ -432,6 +432,9 @@ class PaymentSheetFlowControllerViewController: UIViewController, FlowController
 
     @objc
     private func didTapContinueButton() {
+        // The user is continuing with an LPM, so we un-select Link
+        isHackyLinkButtonSelected = false
```
tillh-stripe, line 436: We reset the Link selection if the user continues with an LPM. Just dismissing the sheet will no longer set `isHackyLinkButtonSelected` to false, meaning that its selection will persist. This is in line with how LPMs work today.

porter-stripe (approved):

PR #4794, opened 2025-04-10
[Connect] Add fetch app info event.

nschris-stripe:
## Summary
Adds the fetchAppInfo event to stripe connect.

## Demo
Validated connect embed was calling the message handler and added tests:
Screenshot 2025-04-10 at 4 34 40 PM

### DIFF
--- a/StripeConnect/StripeConnect/Source/Internal/Webview/ConnectComponentWebViewController.swift
+++ b/StripeConnect/StripeConnect/Source/Internal/Webview/ConnectComponentWebViewController.swift
@@ -50,6 +50,8 @@ class ConnectComponentWebViewController: ConnectWebViewController {
 
     let componentType: ComponentType
 
+    let bundleIdProvider: () -> String?
+
     init<InitProps: Encodable>(
         componentManager: EmbeddedComponentManager,
         componentType: ComponentType,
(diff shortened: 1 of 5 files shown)


StripeConnect/StripeConnect/Source/Internal/Webview/ConnectComponentWebViewController.swift
```diff
@@ -335,6 +342,9 @@ private extension ConnectComponentWebViewController {
                          appearance: .init(appearance: componentManager.appearance, traitCollection: self.traitCollection),
                          fonts: componentManager.fonts.map({ .init(customFontSource: $0) }))
         }))
+        addMessageHandler(FetchAppInfoMessageHandler.init(didReceiveMessage: { [weak self] _ in
+            return .init(applicationId: self?.bundleIdProvider() ?? "")
+        }))
```
mats-stripe, line 347: Should we instead only add this message handler if we have a bundle ID? 
```suggestion
        if let bundleId = bundleIdProvider() {
            addMessageHandler(FetchAppInfoMessageHandler.init(didReceiveMessage: { 
                return .init(applicationId: bundleId)
            }))
        }
```

mats-stripe (approved):
LGTM, one non-blocking suggestion

PR #4785, opened 2025-04-08
Respect `configuration.style` in native Link UI

tillh-stripe:
## Summary

This pull request fixes an issue where the verification dialog as well as the OTP field in the verification screen didn't response the `userInterfaceStyle` of MPE.

## Motivation

[LINK_MOBILE-117](https://jira.corp.stripe.com/browse/LINK_MOBILE-117)

PR #4796, opened 2025-04-11
Fix another theming issue with Link verification fields

tillh-stripe:
## Summary

This pull request fixes a new theming issue in the Link verification view, where the individual text fields would render in light mode even if the device was in dark mode.

This was likely introduced in https://github.com/stripe/stripe-ios/pull/4785, where I attempted to correctly render the text fields if merchants set `alwaysLight` or `alwaysDark` on MPE.

### DIFF
--- a/StripeUICore/StripeUICore/Source/Controls/OneTimeCodeTextField.swift
+++ b/StripeUICore/StripeUICore/Source/Controls/OneTimeCodeTextField.swift
@@ -808,7 +808,14 @@ private extension OneTimeCodeTextField {
         }
 
         private func updateColors() {
-            let traitCollection = UITraitCollection(userInterfaceStyle: configuration.userInterfaceStyle)
+            let traitCollection = switch configuration.userInterfaceStyle {
+            case .light, .dark:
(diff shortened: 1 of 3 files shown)


StripeUICore/StripeUICore/Source/Controls/OneTimeCodeTextField.swift
```diff
@@ -808,7 +808,14 @@ private extension OneTimeCodeTextField {
...
         private func updateColors() {
-            let traitCollection = UITraitCollection(userInterfaceStyle: configuration.userInterfaceStyle)
+            let traitCollection = switch configuration.userInterfaceStyle {
+            case .light, .dark:
+                UITraitCollection(userInterfaceStyle: configuration.userInterfaceStyle)
+            case .unspecified:
+                UIScreen.main.traitCollection
+            @unknown default:
+                UIScreen.main.traitCollection
+            }
```
nschris-stripe, line 818: Just a thought: Can you use self.traitCollection directly here? If the userInterface is getting overridden in the hierarchy then it should apply here.

nschris-stripe (approved):
Looks good!

PR #4802, opened 2025-04-15
Updates tap targeting for viewMore

wooj-stripe:
## Summary
Increases the tappable area for viewing more saved payment methods for all three types of rows -- floating button, checkmark, and radio

For floating button and radio, i've extended the view to the edges.
For checkmark, I've made the label of the digits above "view more" to be tappable.

## Motivation
Tap targeting is too small.

## Testing
Manual testing, and relying on existing screenshot tests to ensure I have not broken any layout issues.

|    | before | After |
| -------- | ------- | ------- |
| floating  | before_floating |   fb_after |
| radio | before_embed_radio  | after_embed_radio | 
| check    | before_embed_check   | after_embed_check (note, the blue portion in this image is tappable!) |


wooj-stripe:
This approach had too many flaws based on the accessory view always being added and the stack view automagically handling width of components.

PR #4813, opened 2025-04-17
Complete cancelled FCLite sessions

mats-stripe:
## Summary

When the FC Lite flow is dismissed by the user, we still attempt to get a session object. In this scenario, we should call the `complete` endpoint ourselves instead of the `session_receipt` endpoint. When the flow is cancelled from within the webview (i.e. tapping the `X`), the web flow calls the `complete` endpoint, so we just get the session receipt.

## Motivation

Fixes bugs when swiping to dismiss.

## Testing

Before: 

https://github.com/user-attachments/assets/8e2500f6-2ebd-489b-9676-8f2d8cf3dc23

After:

https://github.com/user-attachments/assets/e1da957c-3a77-4be3-9e55-f23a6cb544da

Dismissing the flow during various points in the flow:

https://github.com/user-attachments/assets/42c4597a-f6e4-4ff5-8454-809079dd177e

## Changelog

N/a

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Internal/FC Lite/Controllers/FCLiteAuthFlowViewController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Internal/FC Lite/Controllers/FCLiteAuthFlowViewController.swift
@@ -13,8 +13,14 @@ import UIKit
 
 class FCLiteAuthFlowViewController: UIViewController {
     enum WebFlowResult {
+        enum CancellationType {
+            case none
+            case cancelledWithinWebview
+            case cancelledOutsideWebView
+        }
+        
         case success(returnUrl: URL)
-        case cancelled
+        case cancelled(CancellationType)
         case failure(Error)
     }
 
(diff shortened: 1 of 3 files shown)


StripePaymentSheet/StripePaymentSheet/Source/Internal/FC Lite/Controllers/FCLiteAuthFlowViewController.swift
```diff
@@ -13,8 +13,14 @@ import UIKit
 
 class FCLiteAuthFlowViewController: UIViewController {
     enum WebFlowResult {
+        enum CancellationType {
+            case none
```
tillh-stripe, line 17: Why do we need this case?

tillh-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/Internal/FC Lite/Controllers/FCLiteAuthFlowViewController.swift
mats-stripe: Whoops, removed in https://github.com/stripe/stripe-ios/pull/4813/commits/2661a9b4980a0a85960d2962000007d6840504f9

tillh-stripe (approved):

PR #4815, opened 2025-04-17
Add BDCC flag for CPMs in playground

porter-stripe:
## Summary
- Adds a flag to enable BDCC for CPMs in the playground
- Tries in a best effort manner to improve the mode_mismatch error message. It's not helpful.

## Motivation
- https://docs.google.com/document/d/1UNuVxLk2zJBbx6AEqWQ8IPuysv3agmXRgtxefc6Mbnk/edit?tab=t.0

## Testing
- Manual

## Changelog
N/A

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
@@ -432,7 +432,7 @@ class PlaygroundController: ObservableObject {
                 fatalError()
             }
         } else if self.settings.uiStyle == .paymentSheet {
-            self.rootViewController.presentedViewController?.presentedViewController?.present(alert, animated: true)
+            self.rootViewController.presentedViewController?.present(alert, animated: true)
         } else {
             self.rootViewController.present(alert, animated: true)
         }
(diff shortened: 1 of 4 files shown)


Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
```diff
@@ -418,7 +421,17 @@ class PlaygroundController: ObservableObject {
             let exampleError = NSError(domain: "", code: 0, userInfo: [NSLocalizedDescriptionKey: "Something went wrong!"])
             completion(.failed(error: exampleError))
         })
-        if self.settings.uiStyle == .paymentSheet || self.settings.uiStyle == .embedded {
+
+        if self.settings.uiStyle == .embedded {
+            switch embeddedConfiguration.formSheetAction {
+            case .confirm:
+                self.rootViewController.presentedViewController?.presentedViewController?.present(alert, animated: true)
```
joyceqin-stripe, line 428: What's the purpose of this change? When I test with .embedded .confirm with this, the playground gets stuck in this loading state, but it works fine with .embedded .continue.
Screenshot 2025-04-17 at 1 14 01 PM

porter-stripe: Thanks for the catch. The reason for the change is b/c we need to make sure we are always presenting on the currently presented view controller (the top). If we don't, sometimes we don't get the alert to show up like you encountered. With embedded we have an additional VC that is presented which prompted the change.

I've made finding the top VC generic so it should always work now.

joyceqin-stripe (approved):

PR #4818, opened 2025-04-17
Release FC Lite

mats-stripe:
## Summary

Removes the client side guardrails to release FC Lite. This flow will be used when the `StripeFinancialConnections` SDK is not available.

> [!NOTE]  
> FC Lite access can be controlled remotely via `elements_disable_fc_lite`.

## Motivation

Bank payments 📈 

## Testing

Manual testing, bug bashes, unit tests, and E2E tests.

## Changelog

```
[Added] Bank payments are now available in the PaymentSheet without requiring a dependency on `StripeFinancialConnections`.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## x.x.x yyyy-yy-yy
 ### PaymentSheet
+* [Added] Bank payments are now available in the PaymentSheet without requiring a dependency on `StripeFinancialConnections`.
 * [Added] Ability to update saved cards when using CustomerSessions (private preview)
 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## x.x.x yyyy-yy-yy
 ### PaymentSheet
+* [Added] Bank payments are now available in the PaymentSheet without requiring a dependency on `StripeFinancialConnections`.
```
tillh-stripe, line 3: Should we include that it's via a web-based experience?

tillh-stripe (approved):

PR #4822, opened 2025-04-21
Revert "Fix compilation on xcode163 (#4790)"

wooj-stripe:
This reverts commit 00597abd23d37b4d554119c9eb8d72236c6cd191.

## Summary
Seems to break our macOS builds in xcode cloud

## Motivation
Reverting change

## Testing
None


mats-stripe:
Do we need the `#if target ...` logic here at all, or could we always set `scrollIndicatorInsets` to `verticalScrollIndicatorInsets`? My attempt to fix it is here, which seems to compile fine on MacOS: https://github.com/stripe/stripe-ios/pull/4817 (can close in favor of this PR though!)

PR #4830, opened 2025-04-22
Add a flag and read from PMO SFU in PaymentSheetFormFactory

joyceqin-stripe:
## Summary

Renamed PaymentMethodOptionsSetupFutureUsageBeta to PaymentMethodOptionsSetupFutureUsagePreview, added a shouldReadPaymentMethodOptionsSetupFutureUsage to configurations and checks PMO SFU in PaymentSheetFormFactory. This controls showing/hiding the save checkbox and mandates.
## Motivation

PMO SFU project
[MOBILESDK-3406](https://jira.corp.stripe.com/browse/MOBILESDK-3406)
[MOBILESDK-3408](https://jira.corp.stripe.com/browse/MOBILESDK-3408)
## Testing

Unit tests in PaymentSheetFormFactoryTests that check the visibility of the save checkbox and mandates and the value of isSettingUp
## Changelog

N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElementConfiguration.swift
```diff
@@ -137,6 +137,9 @@ extension EmbeddedPaymentElement {
         /// Note: Card brand filtering is not currently supported by Link.
         public var cardBrandAcceptance: PaymentSheet.CardBrandAcceptance = .all
 
+        /// If true, the payment method options setup future usage overrides the top-level setup future usage if set
+        @_spi(PaymentMethodOptionsSetupFutureUsagePreview) public var shouldReadPaymentMethodOptionsSetupFutureUsage: Bool = false
+
```
porter-stripe, line 142: Why do we need this flag if the API itself is gated w/ SPI? https://github.com/stripe/stripe-ios/blob/e1e970ba2d11dbd48de9439b3139e00d4eec7d1b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetIntentConfiguration.swift#L119

porter-stripe: How would I even be able to override the top-level setup future usage without using the gated API?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
```diff
@@ -102,6 +102,22 @@ class PaymentSheetFormFactory {
             let isAccountNotRegisteredOrMissing = linkAccount.flatMap({ !$0.isRegistered }) ?? true
             return isAccountNotRegisteredOrMissing && !UserDefaults.standard.customerHasUsedLink
         }()
+        let shouldReadPaymentMethodOptionsSetupFutureUsage: Bool = {
```
porter-stripe, line 105: Shouldn't we just read it if it's there, b/c that implies they are in the preview?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
```diff
@@ -102,6 +102,22 @@ class PaymentSheetFormFactory {
...
+                return false
+            }
+        }()
+        let paymentMethodType: STPPaymentMethodType = {
+            switch paymentMethod {
+            case .stripe(let paymentMethodType):
+                return paymentMethodType
+            default:
+                return .unknown
+            }
```
porter-stripe, line 119: I think we should explicitly switch on external and return it. I don't see why we'd return unknown here.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElementConfiguration.swift
joyceqin-stripe: The gated API is only for deferred intents. As it stands, merchants are already able to set payment method options setup future usage on payment intents— our SDK just hasn't respected it.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
joyceqin-stripe: It's already possible for an intent to have payment method options with setup future usage. Currently in our SDK, there isn't anything that actually reads the payment method options setup future usage values. It's possible for an intent to have payment method options with setup future usage without being in the preview.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
joyceqin-stripe: External can't convert to STPPaymentMethodType, and external payment methods aren't accepted for payment method options anyways. I could change it to return nil instead and make paymentMethodType optional and add guards in isSetupFutureUsageSet

joyceqin-stripe: I'm not sure that's better though

porter-stripe (approved):

PR #4833, opened 2025-04-23
Fix embedded not vending mandate sometimes

porter-stripe:
## Summary
- In some cases we were not vending the mandate text on the payment option when the merchant opted into displaying the mandate in their own UI.

## Motivation
- https://jira.corp.stripe.com/browse/MOBILESDK-3484
## Testing
- Manual
- New unit test

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
```diff
@@ -381,7 +381,8 @@ class EmbeddedPaymentMethodsView: UIView {
                 self.layoutIfNeeded()
             }
         case (true, false): // Hidden -> Hidden
-            break
+            // Update mandate text on the view so the payment option we vend to the merchant has the correct mandate
+            self.mandateView.attributedText = mandateText
```
porter-stripe, line 385: The payment option that embedded vends is a computed property and looks at mandateView.attributedText to vend the mandate. So even if we are not showing the mandate ourselves, we should at least update the string on the view so we can vend a mandate to the merchant.

wooj-stripe (approved):

PR #4839, opened 2025-04-24
Add card brand filtering for native link.

nschris-stripe:
## Summary
This introduces card brand filtering for native link in passthrough mode.

### Passthrough

https://github.com/user-attachments/assets/561da946-7084-4421-9c80-6f8e557a7bd4

https://github.com/user-attachments/assets/6cf39c6a-8c1c-4d16-9b5d-461a45b3acaf

https://github.com/user-attachments/assets/8cfde43c-79c0-44a9-a3a8-0594754b1f75

### PM

https://github.com/user-attachments/assets/3a147325-d993-4b3e-bde6-d54c78ea9b09


StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/PaymentDetails.swift
```diff
@@ -25,20 +25,22 @@ final class ConsumerPaymentDetails: Decodable {
     let billingEmailAddress: String?
     let nickname: String?
     var isDefault: Bool
+    var isSupported: Bool
```
tillh-stripe, line 28: I'm worried about storing local properties in the API model. Is there another place where we could keep track of this?

Here specifically, it allows me to pay with an unsupported payment method via the “Update card” screen, since the update response overrides this value.

https://github.com/user-attachments/assets/a25da7b8-4aee-4204-9fd0-22ba84cd931f

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Extensions/Intent+Link.swift
```diff
@@ -18,6 +18,10 @@ extension STPElementsSession {
         linkSettings?.passthroughModeEnabled ?? false
     }
 
+    var linkCardBrandFilteringEnabled: Bool {
+        linkPassthroughModeEnabled
+    }
```
tillh-stripe, line 23: Since this is only used in native Link, should we move this somewhere else so that it's scoped more tightly?

tillh-stripe (commented):
Looks great! There's one issue with storing the `isSupported` value in the API model, but otherwise this is 😗 👌

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Extensions/Intent+Link.swift
nschris-stripe: Possibly. I put it here because eventually we will want it for both, right?

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/PaymentDetails.swift
nschris-stripe: Ah, interesting! I think I have a few ideas here. I'll take a look.

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Extensions/Intent+Link.swift
nschris-stripe: Also, If you have ideas for a better place to put it, I'm happy to move it!

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/PaymentDetails.swift
nschris-stripe: Updated, I like this solution better, and starts the process of moving to a single source of truth.

tillh-stripe (commented):
Noticed one more issue: When updating a card of disallowed brand, it selects that card in the wallet screen and allows paying with it.

nschris-stripe:
> Noticed one more issue: When updating a card of disallowed brand, it selects that card in the wallet screen and allows paying with it.

Ah yes this is the issue I was discussing with you. It appears that when you update a card the view model and the view layer get out sync. In a follow up PR I was planning to refactor how this is structured to make it so that there is one source of truth rather than storing the currently selected card in 3 places and hoping they stay in sync. It's a bit out of scope for this change.

mats-stripe:
>When updating a card of disallowed brand, it selects that card in the wallet screen and allows paying with it.

This _feels_ like a blocker to me on merging this PR, but let me know if you disagree! Is it gonna require a big refactor to fix?

nschris-stripe:
I can probably patch it in this PR, and then make the real fix in the next.

mats-stripe (dismissed):
Looks good to me - did a bunch of manual testing!


https://github.com/user-attachments/assets/86f31542-2e50-4398-b4b7-063c3f0fd23e

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Components/PaymentMethodPicker/LinkPaymentMethodPicker.swift
```diff
@@ -44,10 +48,13 @@ final class LinkPaymentMethodPicker: UIView {
     weak var delegate: LinkPaymentMethodPickerDelegate?
     weak var dataSource: LinkPaymentMethodPickerDataSource?
 
-    var selectedIndex: Int = 0 {
-        didSet {
-            updateHeaderView()
-        }
+    var selectedIndex: Int {
+        dataSource?.selectedIndex ?? 0
```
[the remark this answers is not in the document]
nschris-stripe: It needs to be unless we move away from the standard datasource pattern of it being weakly held.

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Components/PaymentMethodPicker/LinkPaymentMethodPicker-Header.swift
```diff
@@ -141,6 +144,11 @@ extension LinkPaymentMethodPicker {
             fatalError("init(coder:) has not been implemented")
         }
 
+        func setSelectedPaymentMethod(selectedPaymentMethod: ConsumerPaymentDetails?, supported: Bool) {
+            self.collapsable = supported
+            self.selectedPaymentMethod = selectedPaymentMethod
```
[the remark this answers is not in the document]
nschris-stripe: It is still possible to have the payment method selected even if it's not supported. Technically the header should never show the selected payment method if it is not supported, but it is still better for consistency sake to have it update in case we ever do show the header in this case.

mats-stripe (approved):
LGTM!

tillh-stripe (approved):

PR #4840, opened 2025-04-24
Render Link payment methods in SPM screens

tillh-stripe:

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/ElementsCustomer.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/ElementsCustomer.swift
@@ -20,7 +20,7 @@ struct ElementsCustomer: Equatable, Hashable {
     public static func decoded(fromAPIResponse response: [AnyHashable: Any]?) -> ElementsCustomer? {
         // Required fields
         guard let response,
-              let paymentMethodsArray = response["payment_methods"] as? [[AnyHashable: Any]],
+              let paymentMethodsWithLinkDetailsArray = response["payment_methods_with_link_details"] as? [[AnyHashable: Any]],
               let customerSessionDict = response["customer_session"] as? [AnyHashable: Any],
               let customerSession = CustomerSession.decoded(fromAPIResponse: customerSessionDict)
         else {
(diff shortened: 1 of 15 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+PaymentMethodAvailability.swift
```diff
@@ -43,10 +43,10 @@ extension PaymentSheet {
     ]
 
     /// A list of `STPPaymentMethodType` that can be saved in PaymentSheet
-    static let supportedSavedPaymentMethods: [STPPaymentMethodType] = [.card, .USBankAccount, .SEPADebit]
+    static let supportedSavedPaymentMethods: [STPPaymentMethodType] = [.card, .USBankAccount, .SEPADebit, .link]
```
davidme-stripe, line 46: How does this work when CustomerSessions isn't set up (with the old ephemeral key flow) when we don't have the PM details? Do we just show the generic Link PMs, or are they not fetched at all?

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/ElementsCustomer.swift
```diff
@@ -20,24 +20,29 @@ struct ElementsCustomer: Equatable, Hashable {
     public static func decoded(fromAPIResponse response: [AnyHashable: Any]?) -> ElementsCustomer? {
         // Required fields
         guard let response,
-              let paymentMethodsArray = response["payment_methods"] as? [[AnyHashable: Any]],
+              let paymentMethodsWithLinkDetailsArray = response["payment_methods_with_link_details"] as? [[AnyHashable: Any]],
```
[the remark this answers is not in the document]
tillh-stripe: `payment_methods_with_link_details` contains the payment method and optional Link details. As far as I can tell, it contains everything that's also in `payment_methods`, but I'll verify that.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+PaymentMethodAvailability.swift
tillh-stripe: If a merchant is using ephemeral keys, then we call `/payment_methods` for all payment method types in `PaymentSheet.supportedSavedPaymentMethods`, which doesn't include Link.

PR #4846, opened 2025-04-25
[ir-xenon-tycoon] Delete cached model data before moving

martinmroz-stripe:
## Summary
This change:

1. Updates the downloader to use a unique UUID-based temporary URL for downloads.
2. When a model fails to load, proactively attempt to delete it.
3. When caching a model, delete the previous entry.

## Motivation
ir-xenon-tycoon is showing sporadic failures during identity verification reporting NSCocoaErrorDomain code 516 when attempting to start scanning. This error is caused when a file already exists at the destination trying to create or move. Based on the time at which it happens, it appears that this occurs during model loading. There are several places where we are moving into the new destination without checking if contents already exists -- and the file downloader is using the last path component to generate a temporary URL.

## Testing
Manually validated the flow with new installations and existing installations using the test app.

## Changelog
[TBD]


martinmroz-stripe:
@bric-stripe mind validating my thinking on this?

StripeCore/StripeCore/Source/Helpers/FileDownloader.swift
```diff
@@ -40,13 +40,17 @@ import Foundation
                 return promise.reject(with: NSError.stp_genericConnectionError())
             }
 
-            // Move the file to a temporary cache directory
-            let temporaryFileURL = FileManager.default.temporaryDirectory.appendingPathComponent(remoteURL.lastPathComponent)
+            // Move the file to a temporary cache directory after generating a unique name to avoid conflicts.
+            let fileManager = FileManager.default
+            let uniqueFileName = "\(UUID().uuidString)_" + remoteURL.lastPathComponent
```
bric-stripe, line 45: not familiar with all the places this is used but does anything depend on the `lastPathComponent` being stable? could create a new directory with the UUID instead if any concern with that

bric-stripe (commented):
looks good to me...one inline question about if we should be keeping the file name stable in case that's needed or shown to users anywhere ever 🤔 (not sure all the use cases)

kentwilliams-stripe (dismissed):
Thanks Martin!

StripeCore/StripeCore/Source/Helpers/FileDownloader.swift
martinmroz-stripe: It shouldn't this is just a file download API that wraps URLSession to download a file, and move it to a temporary location. The output file is then moved into a more permanent location.

davidme-stripe (approved):

PR #4857, opened 2025-04-30
Filtering PM types based on requirements checks PMO SFU

joyceqin-stripe:
## Summary

Added check to PM filtering that reads PMO SFU
## Motivation

PMO SFU project
## Testing

Unit tests for supportsAdding
In the playground, if you are in Payment+SFU mode and have PMO SFU for Affirm set to `none`, see Affirm shows up in the list of payment methods when you load the sheet
## Changelog

N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodType.swift
```diff
@@ -293,7 +293,7 @@ extension PaymentSheet {
             let requirements: [PaymentMethodTypeRequirement]
 
             // We have different requirements depending on whether or not the intent is setting up the payment method for future use
-            if intent.isSettingUp {
+            if configuration.shouldReadPaymentMethodOptionsSetupFutureUsage ? intent.isSetupFutureUsageSet(for: paymentMethod) : intent.isSettingUp {
```
porter-stripe, line 296: I recall a discussion where we thought we could remove this requirement b/c elements/session filters these out??

porter-stripe: Let's start a thread in the project channel to find out.

joyceqin-stripe: We can't remove this requirement, but I think we could remove unsupportedForSetup. Yes, the elements/session filters out things that don't support sfu and it considers pmo sfu, but it still holds true that some payment methods (specifically, sofort, iDEAL, and bancontact) have different requirements for when they're setting up vs when they're not. If delayed pms are not allowed and SFU is set, then sofort, iDEAL, and bancontact must be filtered out.

joyceqin-stripe: It feels cleaner to me to keep unsupportedForSetup just so that the default case doesn't have just an empty list or .unsupported which has a different meaning, but we could remove the case with the list of pms that don't support setup and let the default case catch all (which realistically should not execute since they're filtered out)

porter-stripe (approved):

PR #4862, opened 2025-05-01
Dismiss keyboard on scroll in payment sheet and address sheet

gbirch-stripe:
## Summary
Enable the onDrag keyboardDismissalMode for the scroll views in the payment sheet and address view controller.

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-1746

## Testing
https://github.com/user-attachments/assets/d76fc8d2-f79a-40ac-a7cf-45aab50e5777

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/AddressViewController/AddressViewController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/AddressViewController/AddressViewController.swift
@@ -343,16 +343,16 @@ extension AddressViewController {
 }
 
 // MARK: - ElementDelegate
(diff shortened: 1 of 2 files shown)


gbirch-stripe:
I think the API change check isn't relevant since we just had a do-nothing override of an already public method

joyceqin-stripe:
> I think the API change check isn't relevant since we just had a do-nothing override of an already public method, but someone please sanity check this

I think you're right, I don't think there should be any problems with removing it

joyceqin-stripe (approved):

davidme-stripe (approved):

Rightpartners:
I must pleased

PR #4882, opened 2025-05-08
Link pmo sfu

joyceqin-stripe:
## Summary

Link will display mandates correctly by reading from PMO SFU.

Note: in the playground, elementsSession.linkPassthroughModeEnabled is true except when APMS are off and Link is in PaymentMethod mode, for some reason, so you can only test setting Link as a payment method PMO SFU then. Otherwise, it's in passthrough mode, and it checks the underlying payment method's SFU.
## Motivation

Bug bash
## Testing

Unit tests
Manual testing
- Pay + Link PM mode, link:off_session, mandate showed up
- Pay + Link passthrough mode, link:off_session, mandate did not show up
- Pay + Link PM mode, link: none, card:off_session, mandate did not show up
- Pay + Link passthrough mode, link: none, card:off_session, mandate showed up
- Pay+SFU + Link PM mode, link:none, mandate did not show up
- Pay+SFU Link passthrough mode, link:none, mandate showed up
- Pay+SFU + Link PM mode, link: off_session, card:none, mandate showed up
- Pay + Link passthrough mode, link: off_session, card:none, mandate did not show up
## Changelog

N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
```diff
@@ -519,7 +519,7 @@ extension PaymentSheet {
                     switch result {
                     case .success:
                         STPAnalyticsClient.sharedClient.logLinkSignupComplete()
-                        let linkPaymentMethodType: STPPaymentMethodType = elementsSession.linkPassthroughModeEnabled ? intentConfirmParams.paymentMethodParams.type : .link
+                        let linkPaymentMethodType: STPPaymentMethodType = intentConfirmParams.paymentMethodParams.type
```
joyceqin-stripe, line 522: Upon further investigation, the .signUp case is only for inline link signup, and inline link signup is disabled when not in passthrough mode, so we don't need this check anymore

tillh-stripe: We will likely re-enable inline signup for payment method mode after making some changes. Could you undo this change?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
```diff
@@ -111,6 +111,9 @@ class PaymentSheetFormFactory {
             }
         }()
         let paymentMethodType: STPPaymentMethodType = {
+            if linkAccount != nil, !elementsSession.linkPassthroughModeEnabled {
+                return .link
+            }
```
tillh-stripe, line 116: Is this related to the PMM change above?

joyceqin-stripe: I realized that the PaymentSheetFormFactory within the AddVC used with Link was never checking PMO SFU for Link, as the payment method type passed into the form factory was card. This checks to see if it came from Link, so when it creates the form with Link in pm mode, it'll show mandates, etc properly.

tillh-stripe (approved):

PR #4888, opened 2025-05-12
Add Embedded MPE row selection behavior config to playground

gbirch-stripe:
## Summary
Adds the Embedded MPE row selection behavior config to the playground. This will unlock development and testing of features that rely on the immediateAction row selection behavior.

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-3542

## Testing
https://github.com/user-attachments/assets/d7e6e0f8-2b19-4f1b-a2bf-118893c0d85a

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlaygroundSettings.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlaygroundSettings.swift
@@ -586,6 +586,12 @@ struct PaymentSheetTestPlaygroundSettings: Codable, Equatable {
         case `continue`
     }
 
+    enum RowSelectionBehavior: String, PickerEnum {
+        static let enumName: String = "rowSelectionBehavior"
+        case `default`
+        case immediateAction
+    }
+
     enum CardBrandAcceptance: String, PickerEnum {
         static let enumName: String = "cardBrandAcceptance"
         case all
@@ -652,6 +658,7 @@ struct PaymentSheetTestPlaygroundSettings: Codable, Equatable {
     var collectAddress: BillingDetailsAddress
     var formSheetAction: FormSheetAction
     var embeddedViewDisplaysMandateText: DisplaysMandateTextEnabled
+    var rowSelectionBehavior: RowSelectionBehavior
     var cardBrandAcceptance: CardBrandAcceptance
 
     static func defaultValues() -> PaymentSheetTestPlaygroundSettings {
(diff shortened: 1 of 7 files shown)


Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
```diff
@@ -914,8 +918,8 @@ extension PlaygroundController {
                 let json = try? JSONSerialization.jsonObject(with: data, options: []) as? [String: Any]
             else {
                 if let data = data,
-                   (response as? HTTPURLResponse)?.statusCode == 400 {
-                    let errorMessage = String(decoding: data, as: UTF8.self)
+                   (response as? HTTPURLResponse)?.statusCode == 400,
+                   let errorMessage = String(data: data, encoding: .utf8) {
```
joyceqin-stripe, line 922: Could you explain this change?

mats-stripe: I've made similar changes in the past because our linter flags the original lines as an error IIRC

Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlaygroundSettings.swift
```diff
@@ -586,6 +586,12 @@ struct PaymentSheetTestPlaygroundSettings: Codable, Equatable {
         case `continue`
     }
 
+    enum RowSelectionBehavior: String, PickerEnum {
+        static let enumName: String = "rowSelectionBehavior"
+        case `default`
```
joyceqin-stripe, line 591: Nit: I'd prefer a different name than "default" that's more descriptive of the behavior

Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
gbirch-stripe: What mats said - that String initializer is not allowed by the linter. FYI to save you the research:
The key difference for our purposes (afaik) is that String(decoding:as:) has weird behavior with invalid data (potentially using unicode replacement characters) while String(data:encoding:) has an optional return type and returns `nil` if the input is invalid.

Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlaygroundSettings.swift
gbirch-stripe: I'm open to something like `selectOnTap`, but `default` is the literal name of the underlying enum that this setting changes, and my intuition is that it would be most clear to make the naming consistent.

joyceqin-stripe (approved):

PR #4890, opened 2025-05-12
Pmo sfu analytics

joyceqin-stripe:
## Summary

Send SFU value and a boolean for whether or not PMO SFU is set for analytics

No longer sending the dictionary of PMO SFU because on Hubble it ends up looking like "payment_method_options_setup_future_usage[cashapp]":"on_session", so you can't just query on payment_method_options_setup_future_usage. If we need to query on specific values, we can use /v1/elements/sessions req logs
## Motivation

PMO SFU project
https://jira.corp.stripe.com/browse/MOBILESDK-3531
## Testing

Unit test
## Changelog

N/A


StripePaymentSheet/StripePaymentSheet/Source/Analytics/PaymentSheetAnalyticsHelper.swift
```diff
@@ -432,6 +434,64 @@ final class PaymentSheetAnalyticsHelper {
         let analytic = PaymentSheetAnalytic(event: event, additionalParams: additionalParams)
         analyticsClient.log(analytic: analytic, apiClient: configuration.apiClient)
     }
+
+    private func setupFutureUsage(for intent: Intent?) -> String? {
```
porter-stripe, line 438: I think it may make sense for these functions to live in Intent.swift or a new file like Intent+Analytics.swift.

An idea like:

```swift
extension Intent {
    var setupFutureUsageString: String? {
        switch self {
        case .paymentIntent(let paymentIntent):
            return paymentIntent.setupFutureUsage.stringValue     // always present on a PI

        case .deferredIntent(let config)
                where case .payment(_, _, let usage?, _, _) = config.mode:
            return usage.rawValue                             

        default:
            return nil                                   
        }
    }
}
```

Then you can just:

additionalParams["setup_future_usage"] = intent.setupFutureUsageString

Can also do something similar for `paymentMethodOptionsSetupFutureUsage`

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetAnalyticsHelperTest.swift
```diff
@@ -186,7 +186,11 @@ final class PaymentSheetAnalyticsHelperTest: XCTestCase {
             // Load started -> succeeded
             sut.logLoadStarted()
             sut.logLoadSucceeded(
-                intent: ._testValue(),
+                intent: ._testPaymentIntent(
+                    paymentMethodTypes: [.card],
+                    setupFutureUsage: .offSession,
+                    paymentMethodOptionsSetupFutureUsage: [.card: "none"]
```
porter-stripe, line 192: Let's add some tests that capture more than just offSession, to make sure all the other string values are formatted properly.

joyceqin-stripe: We're now only sending a boolean flag for whether or not pmo sfu is set for payment_method_options_setup_future_usage, but I have added a test that checks both offSession and onSession for the top-level sfu

joyceqin-stripe: Here's the thread where the decision was made: https://stripe.slack.com/archives/C08FQ6WNYHW/p1747424627104249?thread_ts=1747421125.147339&cid=C08FQ6WNYHW

porter-stripe (approved):

PR #4893, opened 2025-05-13
Link move logout button.

nschris-stripe:
## Summary
Move the link sign out button to the email view.

## Demo

https://github.com/user-attachments/assets/6155cf0f-f4be-4764-9462-cae4170a8442


mats-stripe (commented):
Code changes LGTM, but we should confirm Android is doing the same before merging 🙏 

One thing I'm a little worried about is how limited horizontal space the email label now has. Based on the screenshot tests, it looks like we go from showing ~24 characters, to showing ~20 before truncating. Doesn't seem like a dig difference, but the average length for an email address is [22 characters](https://www.sigparser.com/developers/email-parsing/regex-validate-email-address#:~:text=Average%20email%20address%20length%20is%2022%20characters%20long). 

This isn't at all a blocker, but I'm curious if you / anyone has thoughts around this!

mats-stripe (approved):
Looks great!

PR #4894, opened 2025-05-13
Add missing payment sheet telemetry calls

gbirch-stripe:
## Summary
Add telemetry calls for:
- fetching telemetry params after load
- create payment method
- confirm payment intent

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-3163

## Testing
We currently disable telemetry on simulators and do not conduct unit tests on telemetry. I manually verified that telemetry parameters are attached to the create payment method and confirm payment intent API calls.

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift
@@ -149,6 +149,9 @@ final class PaymentSheetLoader {
                     analyticsHelper.startTimeMeasurement(.checkout)
                 }
 
+                // Initialize telemetry. Don't wait for this to finish to call completion.
+                STPTelemetryClient.shared.sendTelemetryData()
+
                 // Call completion
                 let loadResult = LoadResult(
                     intent: intent,
--- a/StripePayments/StripePayments/Source/API Bindings/STPAPIClient+Payments.swift
+++ b/StripePayments/StripePayments/Source/API Bindings/STPAPIClient+Payments.swift
@@ -921,7 +921,7 @@ extension STPAPIClient {
         )
 
         var params = [
-            "app": String(decoding: appData ?? Data(), as: UTF8.self),
+            "app": String(data: appData ?? Data(), encoding: .utf8) ?? "",
             "source": sourceID,
         ]
         if let returnURLString = returnURLString {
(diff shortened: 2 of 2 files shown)


Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetVerticalUITest.swift
```diff
@@ -117,7 +117,8 @@ class PaymentSheetVerticalUITests: PaymentSheetUITestCase {
         app.buttons["Continue"].tap() // For some reason, waitForExistenceAndTap() does not tap this!
         XCTAssertEqual(
             analyticsLog.map({ $0[string: "event"] }),
-            ["mc_load_started", "link.account_lookup.complete", "mc_load_succeeded", "mc_custom_init_customer_applepay", "mc_custom_sheet_newpm_show", "mc_lpms_render", "mc_custom_paymentoption_savedpm_select", "mc_lpms_render", "mc_confirm_button_tapped"]
+            // fraud detection telemetry should not be sent in tests, so it should report an API failure
+            ["mc_load_started", "link.account_lookup.complete", "mc_load_succeeded", "fraud_detection_data_repository.api_failure", "mc_custom_init_customer_applepay", "mc_custom_sheet_newpm_show", "mc_lpms_render", "mc_custom_paymentoption_savedpm_select", "mc_lpms_render", "mc_confirm_button_tapped"]
```
porter-stripe, line 121: Is there a way we can test that these fraud telemetry calls contain the correct payloads? E.g. that it contains the muid, guid, sid?

gbirch-stripe: The tail file updates require that muid, guid, and sid appear in the relevant API calls. They match any value since these IDs are randomly generated by the server. For example, from one of the updates:
```
X-Stripe-Mock-Request: allow_redisplay=unspecified&card\[cvc]=123&card\[exp_month]=1&card\[exp_year]=2040&card\[number]=4242424242424242&payment_user_agent=.*&type=card
X-Stripe-Mock-Request: allow_redisplay=unspecified&card\[cvc]=123&card\[exp_month]=1&card\[exp_year]=2040&card\[number]=4242424242424242&guid=.*&muid=.*&payment_user_agent=.*&sid=.*&type=card
```
The tests that have the `fraud_detection_data_repository.api_failure` analytics added are ones where the payment sheet is initialized and so the mirador call that happens at that time fails. At that point we don't yet have a guid, muid, or sid (that call fetches them for us), so just the fact that the call is attempted is what's important. If you think it would be good I could potentially look into mocking responses and adding tests specifically for this call on payment sheet initialization, I just didn't think it seemed necessary.

porter-stripe (approved):

PR #4895, opened 2025-05-13
Only support Link cards in SPM

tillh-stripe:
## Summary

This pull request updates `PaymentMethodWithLinkDetails.decodedObject()` to ignore Link payment methods that aren't cards, as we don't support those yet.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/PaymentMethodWithLinkDetails.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/PaymentMethodWithLinkDetails.swift
@@ -29,16 +29,9 @@ class PaymentMethodWithLinkDetails: NSObject, STPAPIResponseDecodable {
             return nil
         }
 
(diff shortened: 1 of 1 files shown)


StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/PaymentMethodWithLinkDetails.swift
```diff
@@ -29,9 +29,16 @@ class PaymentMethodWithLinkDetails: NSObject, STPAPIResponseDecodable {
             return nil
         }
 
+        let linkDetailsJson = response["link_payment_details"] as? [AnyHashable: Any]
```
davidme-stripe, line 32: Question: Why not do this check after attempting deserialization?

tillh-stripe: Oh right, I totally can. Have been following my [Android solution](https://github.com/stripe/stripe-android/pull/10799) too closely, where things are a little different.

davidme-stripe (approved):

PR #4897, opened 2025-05-14
Send customer ephemeral key to backend

tillh-stripe:
## Summary

This pull request adds a `legacy_customer_ephemeral_key` parameter to the Elements session request. This parameter helps us determine if inline signup can be enabled. The related backend work can be found [here](https://git.corp.stripe.com/stripe-internal/pay-server/pull/1084214).


gbirch-stripe:
Should a test be added to ensure that this key is attached to requests when appropriate?

PR #4900, opened 2025-05-15
Allow redisplay of Link payment methods that originate from inline signup

tillh-stripe:
## Summary

This pull request makes sure that Link payment methods have the correct `allow_redisplay` and `setup_future_usage` values set on them when added via Link inline signup.

## Motivation

Link PMs in SPM.

## Testing

Added a mock network test.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
@@ -438,9 +438,15 @@ extension PaymentSheet {
                     ConsumerPaymentDetails,
                     String?, // cvc
                     String?, // phone number
-                    Bool
-                ) -> Void = { linkAccount, paymentDetails, cvc, billingPhoneNumber, shouldSave in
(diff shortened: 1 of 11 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
```diff
@@ -529,6 +535,10 @@ extension PaymentSheet {
                     case .failure(let error as NSError):
                         STPAnalyticsClient.sharedClient.logLinkSignupFailure(error: error)
                         // Attempt to confirm directly with params as a fallback.
+                        intentConfirmParams.setAllowRedisplay(
+                            mobilePaymentElementFeatures: elementsSession.customerSessionMobilePaymentElementFeatures,
+                            isSettingUp: intent.isSetupFutureUsageSet(for: intentConfirmParams.paymentMethodParams.type)
+                        )
```
tillh-stripe, line 541: If we were unable to create the Link account, we should still respect the merchant's request to allow redisplay on this payment method.

davidme-stripe (approved):

PR #4917, opened 2025-05-20
Get WalletsButtonView order from elements/sessions

davidme-stripe:
## Summary
Get the WalletButtonsView list from elements/sessions.

## Motivation
Allowing this to be changed from the server.

## Testing
Added logic tests, updated snapshot test.

## Changelog
None yet


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/WalletButtonsView/WalletButtonsView.swift
```diff
@@ -12,40 +12,57 @@ import UIKit
...
+                // Also check PaymentSheet local availability logic
+                if PaymentSheet.isLinkEnabled(elementsSession: flowController.elementsSession, configuration: flowController.configuration) {
+                    wallets.append(.link)
+                }
+            case "apple_pay":
+                if PaymentSheet.isApplePayEnabled(elementsSession: flowController.elementsSession, configuration: flowController.configuration) {
+                    wallets.append(.applePay)
+                }
+            default:
+                break
```
wooj-stripe, line 36: If we introduced a new wallet type, seems like we would prefer to `continue` instead -- otherwise, it's possible to forget appending `.applepay` and/or `.link`

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/WalletButtonsView/WalletButtonsView.swift
```diff
@@ -137,7 +152,7 @@ struct WalletButtonsView_Previews: PreviewProvider {
 fileprivate extension PaymentSheet.FlowController {
     static func _mockFlowController() -> PaymentSheet.FlowController {
         let psConfig = PaymentSheet.Configuration()
-        let elementsSession = STPElementsSession(allResponseFields: [:], sessionID: "", orderedPaymentMethodTypes: [], unactivatedPaymentMethodTypes: [], countryCode: nil, merchantCountryCode: nil, linkSettings: nil, experimentsData: nil, flags: [:], paymentMethodSpecs: nil, cardBrandChoice: nil, isApplePayEnabled: true, externalPaymentMethods: [], customPaymentMethods: [], customer: nil)
+        let elementsSession = STPElementsSession(allResponseFields: [:], sessionID: "", orderedPaymentMethodTypes: [], orderedPaymentMethodTypesAndWallets: ["card", "link", "apple_pay"], unactivatedPaymentMethodTypes: [], countryCode: nil, merchantCountryCode: nil, linkSettings: nil, experimentsData: nil, flags: [:], paymentMethodSpecs: nil, cardBrandChoice: nil, isApplePayEnabled: true, externalPaymentMethods: [], customPaymentMethods: [], customer: nil)
```
wooj-stripe, line 155: I'm surprised adding "card" first works given we have a break; in the init?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/WalletButtonsView/WalletButtonsView.swift
davidme-stripe: I think the behavior is the same either way (it's in the `switch` not the `for` loop) but I can swap it out if it's less confusing

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/WalletButtonsView/WalletButtonsView.swift
davidme-stripe: The break is just for the switch statement, I think

wooj-stripe (approved):

PR #4923, opened 2025-05-21
Fix 3DS2 Whitelisting Encoding Issue

Twigz:
## Summary

Fixes an issue where we send a whitelisting value, even if no text is present. Some ACSs seem to validate that this value should *NOT* exist when the whitelisting text is not present.

## Motivation
An ACS reported failures when trying to validate challenges that did not contain whitelist text.

## Testing
Using the PaymentSheet Example to test cards that did not contain whitelist text to ensure that the field was no longer sent, and then validating that it was still sent when text did exist.


Stripe3DS2/Stripe3DS2/STDSChallengeResponseViewController.m
```diff
@@ -557,6 +557,14 @@ - (void)_resendButtonTapped:(UIButton *)sender {
...
 }
 
+- (nullable id<STDSChallengeResponseSelectionInfo>)whitelistResponse {
+    if (self.response.whitelistingInfoText == nil || self.response.whitelistingInfoText.length == 0) {
+        return nil;
+    } else {
+        return self.whitelistView.selectedResponse;
+    }
+}
+
```
sfriedman-stripe, line 567: Can we add a test for this logic?

Twigz: I added some by hooking into where the snapshots are being taken. I also had to update the public API to allow access, as currently it's not very accessible via the header. It's not amazing, but it does test the logic.

porter-stripe: @Twigz we would like to avoid making this public if possible. Instead can we remove the new public API from the header and user `performSelector:` instead?

Twigz: I believe so, just be aware if it changes that would crash (fortunately it would be in the tests)

porter-stripe: I think that's fine, thanks!

PR #4927, opened 2025-05-22
Hide FlowController wallet options when WalletButtonsView is enabled

davidme-stripe:
## Summary
* If a WalletButtonsView is in the view hierarchy, we will hide the Apple Pay / Link options from FC
* When it disappears, we'll show them again.
* Added a new `willUseWalletButtonsView` option to PaymentSheet.Configuration, which will disable setting the default to Apple Pay / Link (as they'll be shown in the WalletButtonsView).

Known limitations:
* This only works with FlowController in Vertical mode. In a future patch, we should support Embedded and Horizontal FlowController.

## Motivation
Configuring WalletButtonsView to work with FlowController.

## Testing
PS Example


Example/PaymentSheet Example/PaymentSheet Example/ExampleWalletButtonsView.swift
```diff
@@ -62,6 +91,7 @@ class ExampleWalletButtonsModel: ObservableObject {
                 configuration.customer = .init(
                     id: customerId, ephemeralKeySecret: customerEphemeralKeySecret)
                 configuration.returnURL = "payments-example://stripe-redirect"
+                configuration.willUseWalletButtonsView = true
```
wooj-stripe, line 94: Just a thought -- since `willUseWalletButtonsView` only applies to FlowController -- would it be better to pass this (and potentially other configs) through `PaymentSheet.FlowController.create()` instead of `PaymentSheet.Configuration" ?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
```diff
@@ -169,7 +169,8 @@ public class PaymentSheet {
                                 configuration: self.configuration,
                                 loadResult: loadResult,
                                 isFlowController: false,
-                                analyticsHelper: self.analyticsHelper
+                                analyticsHelper: self.analyticsHelper,
+                                walletButtonsShownExternally: false
```
wooj-stripe, line 173: I think we should either:
A. Remove passing in false since it defaults to false anyways
or
b. Remove the default in the init and leave this as `false`.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetVerticalViewController.swift
```diff
@@ -150,6 +151,7 @@ class PaymentSheetVerticalViewController: UIViewController, FlowControllerViewCo
         loadResult: PaymentSheetLoader.LoadResult,
         isFlowController: Bool,
         analyticsHelper: PaymentSheetAnalyticsHelper,
+        walletButtonsShownExternally: Bool = false,
```
wooj-stripe, line 154: Is there any reason to accept this in the init?  Seems like nobody is calling this with `true` ?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
```diff
@@ -503,6 +512,22 @@ extension PaymentSheet {
...
+                loadResult: self.viewController.loadResult,
+                analyticsHelper: analyticsHelper,
+                walletButtonsShownExternally: self.walletButtonsShownExternally,
+                previousPaymentOption: self._paymentOption
+            )
+            self.viewController.flowControllerDelegate = self
+
+            // Synchronously pre-load image into cache
+            // Accessing paymentOption has the side-effect of ensuring its `image` property is loaded (e.g. from the internet instead of disk) before we call the completion handler.
+            _ = self.paymentOption
```
wooj-stripe, line 528: There is no completion handler -- so we may load the image, but we may not report it back to anyone needing to display it?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetVerticalViewController.swift
davidme-stripe: We check it for `shouldShowApplePayInList` a few lines down. It's called by `FlowController.updateForWalletButtonsView()` with true when needed.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
davidme-stripe: Makes sense!

Example/PaymentSheet Example/PaymentSheet Example/ExampleWalletButtonsView.swift
davidme-stripe: I think it's a configuration option (even if it does only apply to FlowController), it seems odd to clutter up create with FC-specific options. It should support non-FC someday, and there are other options that don't apply to all PaymentSheets (e.g. `paymentMethodLayout` doesn't work for Embedded).

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetVerticalViewController.swift
wooj-stripe: Sorry, i was in the middle of writing this comment -- i meant to delete it. my bad!

Example/PaymentSheet Example/PaymentSheet Example/ExampleWalletButtonsView.swift
wooj-stripe: I was considering something like the following in case we wanted to separate out this specific use case for flow controller. Not sure if this it's warranted though.
```
FlowController.create(
    intentConfig:  PaymentSheet.IntentConfiguraiton,
    configuration: PaymentSheet.Configuration,
    walletConfiguration: PaymentSheet.WalletConfiguration,
    completion:
)
```

wooj-stripe (approved):

PR #4928, opened 2025-05-22
traitCollectionDidChange loading indicator

joyceqin-stripe:
## Summary

The loading indicator color used to be based off of the system mode (light or dark) instead of the component color, and it wouldn't change its color in response to an appearance change. Now we use traitCollectionDidChange to calculate the correct color and change it when necessary.
## Motivation

[MOBILESDK-3426](https://jira.corp.stripe.com/browse/MOBILESDK-3426)
## Testing

Before

https://github.com/user-attachments/assets/67209eed-8fb0-4711-bc4a-db3c257385f2

After

https://github.com/user-attachments/assets/f655440a-c0e6-4857-922f-4f4e4dcb9518

## Changelog

N/A


gbirch-stripe (approved):
lgtm

PR #4932, opened 2025-05-23
Add support for existing consumers in Financial Connections and FC Lite

mats-stripe:
## Summary

This lays the foundation for supporting existing Link consumers in Financial Connections (through the SDK and through FC Lite). 

The main changes here involve accepting a consumer session at the start of the flow, and using its publishable key for API requests throughout the flow. We also introduce a new flow result type, `linkedAccount`, which simply provides the account ID. This also involves updating our hosted auth flow URL builder to no longer append additional instant debits parameters to the URL. In this scenario we're using instant debits, but don't want the auth flow to create the payment method for us.

## Motivation

Part 1 of 2 of https://docs.google.com/document/d/1S7OyhUuqi7ESFO4rZDt71exHCyyZwGBWKQPr7m9LLJs/edit?usp=sharing

## Testing

Lots of manual testing will be done in the next PR, which plugs these changes to Link in MPE.

## Changelog

N/a


StripeFinancialConnections/StripeFinancialConnections/Source/FinancialConnectionsSheet.swift
```diff
@@ -122,6 +122,9 @@ final public class FinancialConnectionsSheet {
     /// Any additional Elements context useful for the Financial Connections SDK.
     @_spi(STP) public var elementsSessionContext: StripeCore.ElementsSessionContext?
 
+    /// A existing consumer, if avaialble.
```
tillh-stripe, line 125:
```suggestion
    /// An existing consumer, if available.
```

tillh-stripe (approved):

PR #4933, opened 2025-05-23
Support adding bank accounts in Link

mats-stripe:
## Summary

This wires up the changes introduced in the upstream branch for adding bank accounts to Link. It uses the Financial Connections SDK if available, and otherwise uses FC Lite. The general flow is;

1. Create a Link Account Session
2. Launch Financial Connections using the client secret we get in step 1
3. Create the payment details using the account ID we get in step 2
4. Refresh the Link payment method list

## Motivation

Part 2 of 2 of https://docs.google.com/document/d/1S7OyhUuqi7ESFO4rZDt71exHCyyZwGBWKQPr7m9LLJs/edit?usp=sharing

## Testing

Native SDK:

https://github.com/user-attachments/assets/6766aed1-21d2-4a5b-b963-52f0b0f0055f

Web SDK:

https://github.com/user-attachments/assets/13b17538-c6af-4c45-843c-26561e22da24

FC Lite

https://github.com/user-attachments/assets/3d63f0ec-9905-44e7-9635-9ce5c2d9887c

## Changelog

N/a

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController-WalletViewController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController-WalletViewController.swift
@@ -196,6 +196,7 @@ extension PayWithLinkViewController {
                 mandateView.setText(mandate)
             }
 
+            paymentPicker.reloadData()
             paymentPickerContainerView.toggleArrangedSubview(
                 mandateView,
                 shouldShow: viewModel.shouldShowMandate,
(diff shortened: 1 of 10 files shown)


StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController-WalletViewController.swift
```diff
@@ -645,37 +660,69 @@ extension PayWithLinkViewController.WalletViewController: LinkPaymentMethodPicke
...
+            let addBankAction = UIAlertAction(
+                title: STPLocalizedString(
+                    "Bank",
+                    "Label shown in the payment type picker describing a bank payment"
+                ),
+                style: .default
+            ) { [weak self] _ in
+                self?.addBankAccount()
             }
+            alertController.addAction(addBankAction)
```
tillh-stripe, line 678: We'll need to check if the account supports bank accounts. Web has some logic [here](https://livegrep.corp.stripe.com/view/stripe-internal/pay-server/lib/consumer/link_app/frontend/src/contexts/ConsumerAccountContext/FinancialConnections/flags.ts) for that, but we could (even should!) see if we can move it to the backend.

PR #4941, opened 2025-05-28
Appearance API formInsets

joyceqin-stripe:


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetAppearance.swift
```diff
@@ -48,6 +48,9 @@ public extension PaymentSheet {
         /// Describes the appearance of the Embedded Mobile Payment Element
         public var embeddedPaymentElement: EmbeddedPaymentElement = EmbeddedPaymentElement()
 
+        /// Describes the padding used for all forms
+        public var formInsets: UIEdgeInsets = UIEdgeInsets(top: PaymentSheetUI.defaultSheetMargins.top, left: PaymentSheetUI.defaultSheetMargins.leading, bottom: PaymentSheetUI.defaultSheetMargins.bottom, right: PaymentSheetUI.defaultSheetMargins.trailing)
+
```
gbirch-stripe, line 53: I guess this is more of an API question, but why does the API use a data type that uses left/right if we remap those to trailing/leading? It feels like this could be confusing in RTL scenarios. Should there be a comment on the API explaining this behavior?

joyceqin-stripe: I used UIEdgeInsets because Nick said he prefers UIEdgeInsets and I deferred to his knowledge. This API is not finalized, but I'm leaving for PTO tomorrow so Seth asked that I get my PRs reviewed as if the API will not change, so I think we can add a comment explaining this

joyceqin-stripe: Or we could detect if it's an RTL language or not and flip leading/trailing as necessary

joyceqin-stripe: Nick's main cases against NSDirectionalEdgeInsets are that the NS prefix is outdated, developers are more likely to be familiar with UIEdgeInsets, and it's unlikely we'll see any practical impact since 1. we currently don't support any RTL languages and 2. it would require someone to set different values for left and right. I can understand that logic, but Android's will use start/end which handles RTL, and it seems good to be prepared for the case of RTL anyways. I could really go either way, but I'm sort of leaning towards keeping it as UIEdgeInsets since the API has been reviewed as such already. WDYT?

gbirch-stripe (approved):

PR #4949, opened 2025-05-29
Add session id to `elements/session` call

joyceqin-stripe:
## Summary

Add "mobile_session_id" as a parameter in makeElementsSessionParams
## Motivation

https://jira.corp.stripe.com/browse/RUN_MOBILESDK-4026
## Testing

Make sure "mobile_session_id" matches, inspect `elements/session` request (e.g. https://dashboard.stripe.com/test/logs/req_ApNWux8upQlWyF), network stubs
## Changelog

N/A


StripeCore/StripeCore/Source/Analytics/AnalyticsHelper.swift
```diff
@@ -29,6 +29,10 @@ import Foundation
         sessionID = uuid.uuidString.lowercased()
     }
 
+    @_spi(STP) public func clearSessionID() {
```
gbirch-stripe: Can we make this internal? `StripeCore` is imported as `@testable` by `STPNetworkStubbingTestCase`, so it can access internal methods. This way, we don't have to expose a public API just for our own use in tests.

gbirch-stripe (approved):
nit: can omit `internal` since it will be internal by default

gbirch-stripe (approved):

PR #4966, opened 2025-06-03
Prevent Link dismissal more robustly

tillh-stripe:


StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController-UpdatePaymentViewController.swift
```diff
@@ -176,11 +176,17 @@ extension PayWithLinkViewController {
...
+            coordinator?.allowSheetDismissal(false)
+
             linkAccount.updatePaymentDetails(id: paymentMethod.stripeID, updateParams: updateParams) { [weak self] result in
                 guard let self else {
                     return
                 }
 
+                defer {
+                    self.coordinator?.allowSheetDismissal(true)
+                }
```
mats-stripe: Should we move this above the `guard` a few lines up?

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController.swift
```diff
@@ -561,12 +561,18 @@ extension PayWithLinkViewController: PayWithLinkCoordinating {
                 )
             )
         ) { [weak self] result, confirmationType in
-            self?.view.isUserInteractionEnabled = true
-            self?.context.isDismissible = true
+            if result.isCanceledOrFailed {
+                self?.allowSheetDismissal(true)
+            }
             completion(result, confirmationType)
```
mats-stripe: Just to confirm, we don't need to re-allow sheet dismissal here when the result is a success, because the sheet will dismiss itself, right?

mats-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController.swift
tillh-stripe: Yep, that's the idea. We could also move `allowSheetDismissal()` out of the confirm method and call it at the call-site, if that feels better?

mats-stripe: I think that'd be great!

mats-stripe (approved):

PR #4970, opened 2025-06-04
Show Link consumer email in row buttons

tillh-stripe:
## Summary

This pull request make the Link row button render the email address of a returning Link user as the subtitle, instead of rendering a generic message. We do this in both FlowController (in vertical mode) and Embedded.

## Testing

Snapshot tests.


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButton.swift
```diff
@@ -167,9 +167,13 @@ class RowButton: UIView, EventHandler {
         stpAssertionFailure("RowButton init not called from subclass, use RowButton.create() instead of RowButton(...).")
     }
 
-    func setSublabel(text: String?) {
+    func setSublabel(text: String?, animated: Bool = true) {
+        guard text != sublabel.text else {
+            return
+        }
```
tillh-stripe, line 173: Skipping animation if the text hasn't changed.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
```diff
@@ -243,6 +246,26 @@ class EmbeddedPaymentMethodsView: UIView {
...
+            if linkRowButton != nil {
+                initializeLinkAccountObserver()
+            }
+        } else {
+            LinkAccountContext.shared.removeObserver(self)
+        }
+    }
+
+    private func initializeLinkAccountObserver() {
+        LinkAccountContext.shared.addObserver(self, selector: #selector(onLinkAccountChange(_:)))
```
porter-stripe, line 267: Do we have anything preventing this from adding multiple observers? I think when we call this it will add multiple observers even if we already have one. Can we add a flag to not add an observer if we already have one?

porter-stripe:
```swift
    private var isObserverAdded = false
    private func addLinkAccountObserverIfNeeded() {
        guard !isObserverAdded else { return }
        LinkAccountContext.shared.addObserver(self, selector: #selector(onLinkAccountChange(_:)))
        isObserverAdded = true
    }
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
```diff
@@ -243,6 +246,26 @@ class EmbeddedPaymentMethodsView: UIView {
...
 
+    override func didMoveToWindow() {
+        super.didMoveToWindow()
+
+        if window != nil {
+            if linkRowButton != nil {
+                initializeLinkAccountObserver()
+            }
+        } else {
+            LinkAccountContext.shared.removeObserver(self)
```
porter-stripe, line 262: Let's also add

```swift
    deinit {
        LinkAccountContext.shared.removeObserver(self)
    }
```

To clean up and remove the observer when the view is destroyed.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentMethodsView.swift
tillh-stripe: That's what the code in `didMoveToWindow` is for. It adds the observer when the view is added to the window, and removes the observer when the view is removed from the window. That should make it so that at most one observer is added.

porter-stripe (approved):

PR #4983, opened 2025-06-09
Make payment option on FC published for SwiftUI

porter-stripe:

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
@@ -6,6 +6,7 @@
 //  Copyright © 2020 Stripe, Inc. All rights reserved.
 //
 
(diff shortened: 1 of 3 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
```diff
@@ -591,9 +590,36 @@ extension PaymentSheet {
...
-                previousPaymentOption: self._paymentOption
+                previousPaymentOption: self.internalPaymentOption
             )
             self.viewController.flowControllerDelegate = self
+            updatePaymentOption()
+        }
+
+        /// Updates the published paymentOption property based on the current state
+        internal func updatePaymentOption() {
+            if Thread.isMainThread {
```
davidme-stripe, line 601: I'm a little worried that this could be quirky depending on all the different things that update this, but I *think* it's okay. (Or at least it's just as reasonable as the old implementation)

PR #5005, opened 2025-06-13
Initial Shop Pay WKWebView bridge

davidme-stripe:
## Summary
Integrates a one-off ECE bridge for Shop Pay. We'll eventually extend this into a generic ECE bridge, but we'll start by hardcoding Shop Pay support.

* ECEViewController presents ECE in a WKWebView and bridges the ECE callbacks to the `ExpressCheckoutWebviewDelegate`.
* ShopPayECEPresenter uses ECEViewController to present Shop Pay, including bridging the ECE callbacks to the user's shipping address/rate handlers

This doesn't actually work end-to-end yet, as we haven't built out the new facilitated payment confirmation handler. It'll also need a *lot* of work if/when we enable it outside facilitated payment scenarios.

## Motivation
Enabling Shop Pay in WalletButtonsView

## Testing
Added integration tests, snapshot tests (fairly useless at the moment, but we'll style it soon), and unit tests for the ViewController and JS bridge.

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
```diff
@@ -475,7 +475,7 @@ extension PaymentSheet {
             ///     completion(updates)
             /// }
             /// ```
-            /// WARNING: If you do not call the completion handler, your app will hang until the Apple Pay sheet times out.
+            /// WARNING: If you do not call the completion handler, your app will hang until the Shop Pay sheet times out.
```
wooj-stripe, line 478: oops, thank you.

StripePaymentSheet/StripePaymentSheet/Source/Internal/ECE/ECEIndexHtml.swift
```diff
@@ -0,0 +1,322 @@
...
+            if (response.merchantDecision === "rejected") {
+              console.log(`Shipping rejected: ${response.error}`);
+              event.reject();
+              return;
+            }
+
+            // Update the total amount from the response
+            if (response.totalAmount) {
+              elements.update({ amount: response.totalAmount });
+            }
```
wooj-stripe: Is this what merchants typically have to do in their ECE integration? Maybe we should update our API so that merchants have to calculate the totalAmount, and then pass it in?

StripePaymentSheet/StripePaymentSheet/Source/Internal/ECE/ECEViewController.swift
```diff
@@ -0,0 +1,533 @@
...
+
+    private var webView: WKWebView!
+    private var popupWebView: WKWebView?
+
+    // Delegate for Express Checkout Element events
+    weak var expressCheckoutWebviewDelegate: ExpressCheckoutWebviewDelegate?
+
+    // We normally hide Express Checkout Element in WKWebViews (as many don't handle popups correctly)
+    // Fake Safari's UA to disable that behavior
+    static let FakeSafariUserAgent = "Mozilla/5.0 (iPhone; CPU iPhone OS 18_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/18.1.1 Mobile/15E148 Safari/604.1"
```
wooj-stripe, line 57: are there plans to fix this pre-launch?

StripePaymentSheet/StripePaymentSheet/Source/Internal/ECE/ECEIndexHtml.swift
```diff
@@ -0,0 +1,322 @@
...
+
+            // Use the response from native API
+            const { clientSecret, paymentIntentId } = response;
+
+            if (mode === "payment") {
+              const confirmedPI = await stripe.confirmPayment({
+                elements,
+                clientSecret,
+                confirmParams: {
+                  return_url: "https://amazon.com",
```
wooj-stripe: I'm guessing we need to update this return_url?

davidme-stripe: Ugh sorry, I deleted this but forgot to push

StripePaymentSheet/StripePaymentSheet/Source/Internal/ECE/ECEIndexHtml.swift
davidme-stripe: Hm, we should probably just get the amount from the Intent?

StripePaymentSheet/StripePaymentSheet/Source/Internal/ECE/ECEViewController.swift
davidme-stripe: It might be better to add a specific tag to the user agent instead of impersonating Safari, let's coordinate with ECE folks

StripePaymentSheet/StripePaymentSheet/Source/Internal/ECE/ShopPayECEPresenter.swift
```diff
@@ -0,0 +1,326 @@
...
+        }
+        if let phone = billingDetails["phone"] as? String {
+            paymentMethodParams.billingDetails?.phone = phone
+        }
+        if let name = billingDetails["name"] as? String {
+            paymentMethodParams.billingDetails?.name = name
+        }
+
+        // Create payment option
+        let confirmParams = IntentConfirmParams(type: .stripe(.unknown))
```
wooj-stripe, line 264: Do we know if `.unknown`, or do we need to pass in another value?

davidme-stripe: We'll want to replace this with `shop_pay` once it's a real payment method

StripePaymentSheet/StripePaymentSheet/Source/Internal/ECE/ECEIndexHtml.swift
wooj-stripe: The intent is unlikely to be created at this point -- afaik, this is only for UI purposes only.

davidme-stripe: I mean the `Intent`, so the deferred intent information in this case

StripePaymentSheet/StripePaymentSheet/Source/Internal/ECE/ECEBridgeTypes.swift
```diff
@@ -0,0 +1,294 @@
...
+/// Configuration for the payment interface from the click event
+struct ECEClickConfiguration: Codable {
+    /// Line items to display in the payment interface
+    let lineItems: [ECELineItem]
+
+    /// Available shipping rates
+    let shippingRates: [ECEShippingRate]?
+
+    /// Apple Pay specific options
+    let applePay: ECEApplePayOptions?
```
wooj-stripe, line 180: Not entirely sure why this is needed here.

PR #5014, opened 2025-06-17
Update payment option before dismissing sheet

porter-stripe:
## Summary
- Fix a race condition where the sheet was dismissed before the payment option was updated.

## Motivation
- Bug fix

## Testing
- Manual

## Changelog
N/A


tillh-stripe (approved):
Thanks for fixing this!

PR #5017, opened 2025-06-18
Fix tap targeting on walletbuttonsview shoppay button

wooj-stripe:
## Summary
Fixes that tap targeting of the ShopPay WalletButtonsView

## Motivation
Currently, you can only tap the icon

## Testing
Manual

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/WalletButtonsView/ShopPayButton.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/WalletButtonsView/ShopPayButton.swift
@@ -22,6 +22,7 @@ struct ShopPayButton: View {
                 .resizable()
                 .scaledToFit()
                 .frame(height: Constants.contentHeight)
+                .frame(minWidth: 180, maxWidth: .infinity)
         }
         .frame(maxWidth: .infinity)
         .frame(height: Constants.buttonHeight)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/WalletButtonsView/ShopPayButton.swift
```diff
@@ -22,6 +22,7 @@ struct ShopPayButton: View {
                 .resizable()
                 .scaledToFit()
                 .frame(height: Constants.contentHeight)
+                .frame(minWidth: 180, maxWidth: .infinity)
```
porter-stripe, line 25: I think a Modifier might be what we want.

```swift
        Button(action: action) {
            SwiftUI.Image(uiImage: Image.shoppay_logo_bw.makeImage(template: false))
                .resizable()
                .scaledToFit()
                .frame(height: Constants.contentHeight)
        }
        .frame(maxWidth: .infinity, minHeight: Constants.buttonHeight)
        .contentShape(Rectangle()) // Makes entire frame tappable
        .background(Color(Constants.backgroundColor))
        .foregroundColor(.black)
        .cornerRadius(Constants.cornerRadius)
```

wooj-stripe: I tried your suggestion:
```
        Button(action: action) {
            SwiftUI.Image(uiImage: Image.shoppay_logo_bw.makeImage(template: false))
                .resizable()
                .scaledToFit()
                .frame(height: Constants.contentHeight)
//                .frame(minWidth: 180, maxWidth: .infinity)
        }
        .frame(maxWidth: .infinity, minHeight: Constants.buttonHeight)
        .contentShape(Rectangle())
        .background(Color(Constants.backgroundColor))
        .foregroundColor(.black)
        .cornerRadius(Constants.cornerRadius)
```        

But tap targeting was still only on the icon.  I also tried adding `contentShape(Rectangle())` to the Image, but... that also did not work.

porter-stripe (approved):

PR #5018, opened 2025-06-18
Add ability to create shop_pay pm type

wooj-stripe:
## Summary
Adds plumbing for creating PM for on WalletButtonsView shopPay confirmation

## Motivation
Support ShopPay via WalletButtonsView

## Testing
Manually wrote some code which instrumented the creation of a PM. Inspected with object explorer.

### DIFF
--- a/StripePayments/StripePayments/Source/API Bindings/Models/PaymentMethods/STPPaymentMethodEnums.swift
+++ b/StripePayments/StripePayments/Source/API Bindings/Models/PaymentMethods/STPPaymentMethodEnums.swift
@@ -96,7 +96,7 @@ import Foundation
     case twint
     /// A Multibanco payment method
     case multibanco
-    /// A ShopPay payment mehtod
+    /// A ShopPay payment method
     @_spi(STP) case shopPay
     /// An unknown type.
     case unknown


StripePaymentSheet/StripePaymentSheet/Source/Internal/ECE/ShopPayECEPresenter.swift
```diff
@@ -260,13 +262,10 @@ extension ShopPayECEPresenter: ExpressCheckoutWebviewDelegate {
...
 
-        // Create payment option
-        let confirmParams = IntentConfirmParams(type: .stripe(.unknown))
-        confirmParams.paymentMethodParams.billingDetails = paymentMethodParams.billingDetails
-
-        // TODO: Create a payment method here from the data (using STPAPIClient) once the API is available
-        // For now, use a mock STPPaymentMethod instead
-        let paymentMethod = STPPaymentMethod(stripeId: "pm_123abc", type: .unknown)
+        // Create payment method
+        // TODO: If this fails, then return a PaymentSheetResult failed?
```
porter-stripe, line 266: What's stopping us from handling this TODO in this PR?

StripePayments/StripePayments/Source/API Bindings/Models/PaymentMethods/STPPaymentMethodEnums.swift
```diff
@@ -96,6 +96,8 @@ import Foundation
     case twint
     /// A Multibanco payment method
     case multibanco
+    /// A ShopPay payment mehtod
```
porter-stripe: typo in method

StripePaymentSheet/StripePaymentSheet/Source/Internal/ECE/ShopPayECEPresenter.swift
wooj-stripe: The wallet UI page doesn't load -- so it's difficult to know what the ux will be here, or how we even convey that an error has occured. It's possible that the correct behavior is to just throw, but it's also possible we convey this error some other way.

This code is expected to change more down the line (see line 280)

StripePayments/StripePayments/Source/API Bindings/Models/PaymentMethods/STPPaymentMethodEnums.swift
wooj-stripe: ty!

PR #5028, opened 2025-06-20
Extract FC Lite into its own module

mats-stripe:
## Summary

Extracts FC Lite from the `StripePaymentSheet` module into its own module `StripeFinancialConnectionsLite`. It only has one dependency, `StripeCore`.

> [!NOTE]  
>Let me know if I'm missing anything here that would be needed for a new SDK!

## Motivation

This will allow FC Lite be used outside of MPE.

## Testing

Things still work!

https://github.com/user-attachments/assets/4366ba73-07f0-4eb7-8396-61d36623ee1d

## Changelog

N/a


StripeFinancialConnectionsLite/README.md
```diff
@@ -0,0 +1,14 @@
...
+ |  ____|  / ____| | |     (_)| |         
+ | |__    | |      | |      _ | |_   ___  
+ |  __|   | |      | |     | || __| / _ \ 
+ | |      | |____  | |____ | || |_ |  __/ 
+ |_|       \_____| |______||_| \__| \___| 
+```
+
+## FinancialConnectionsLite
+
+FinancialConnectionsLite is a lightweight version of the `StripeFinancialConnections` SDK. It is a simple web wrapper that allows users to go through the Stripe pay-by-bank flow without requiring an additional dependency on the `StripeFinancialConnections` SDK. It is intended to be used in scenarios where the full SDK is not available.
```
carlosmuvi-stripe, line 12: minor - since FC flows are not necessarily ending up in payments we can rephrase this a bit 

```suggestion
FinancialConnectionsLite is a lightweight version of the `StripeFinancialConnections` SDK. It is a simple web wrapper that allows users to go through the Stripe bank account linking flows without requiring an additional dependency on the `StripeFinancialConnections` SDK. It is intended to be used in scenarios where the full SDK is not available.
```

StripeFinancialConnectionsLite/StripeFinancialConnectionsLite/Source/Controllers/FCLiteAuthFlowViewController.swift
```diff
@@ -0,0 +1,220 @@
...
+//  Created by Mat Schmid on 2025-03-19.
+//
+
+import AuthenticationServices
+import UIKit
+// `@preconcurrency` suppresses Sendable-related warnings from WebKit.
+@_spi(STP) import StripeCore
+@preconcurrency import WebKit
+
+class FCLiteAuthFlowViewController: UIViewController {
```
carlosmuvi-stripe, line 14: The idea here is to remove the `FCLiteAuthFlowViewController` in PaymentSheet right? I see other files being moved but this looks like a new addition duplicating the PS one.

StripeFinancialConnectionsLite/StripeFinancialConnectionsLite/Source/FC Lite/FinancialConnectionsLite.swift
carlosmuvi-stripe, line 1: ❤️

StripeFinancialConnectionsLite/StripeFinancialConnectionsLite/Source/Controllers/FCLiteAuthFlowViewController.swift
mats-stripe: I'm really not sure why some files are marked as "moved" and others seem like they're added 🤔 Everything seems moved in the directory structure

mats-stripe: Confirmed there isn't any FC Lite code left in `StripePaymentSheet`. Still not sure why it shows up like this in Github, but I think everything here is fine!

porter-stripe (approved):

PR #5042, opened 2025-06-25
Reduce shopPay jank

wooj-stripe:
## Summary
An attempt to remove jank
* Removes close button on the actual webview, replaced by a native button. Couldn't figure out how to hook into this... -- open to other ideas.
* Fixes memory leak by removing webview from controller
* Removes navigationController container to get rid of additional white space at the top

## Motivation
Experience before was a bit janky.

## Testing
Manually tested -- and will need to update unit tests, if needed


StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/ECE/ECEViewControllerSnapshotTests.swift
```diff
@@ -1,169 +0,0 @@
-//
-//  ECEViewControllerSnapshotTests.swift
-//  StripePaymentSheetTests
```
wooj-stripe, line 3: These tests didn't seem useful since they only rendered a blank screen w/ some navigation items (that have since been removed). We also don't own the screen that it is rendering so it will be difficult to maintain these. Removing these for now.

davidme-stripe: Sounds good, I was thinking we'd have the Shop Pay logo rendered on our end, but it makes sense to just have it in rendered by the page

StripePaymentSheet/StripePaymentSheet/Source/Internal/ECE/ECEViewController.swift
```diff
@@ -144,6 +150,39 @@ class ECEViewController: UIViewController {
         window.webkit.messageHandlers.ready.postMessage({
             type: 'bridgeReady'
         });
+
+        (function() {
+            function hideReturnButtons() {
+                const returnButtons = document.querySelectorAll('button[aria-label="Return to store"]');
```
davidme-stripe, line 156: Is this localized?

StripePaymentSheet/StripePaymentSheet/Source/Internal/ECE/ECEViewController.swift
```diff
@@ -144,6 +150,39 @@ class ECEViewController: UIViewController {
...
+                mutations.forEach(function() {
+                    hideReturnButtons();
+                });
+            });
+
+            // Start observing the document with the configured parameters
+            observer.observe(document.body, { childList: true, subtree: true });
+
+            // Also check periodically (as a fallback)
+            // setInterval(hideReturnButtons, 500);
```
davidme-stripe: Can we remove or uncomment this? Checking periodically makes sense to me

StripePaymentSheet/StripePaymentSheet/Source/Internal/ECE/ECEViewController.swift
wooj-stripe: Good question -- i've opened a thread with the owners of this for advice on how to handle this.

StripePaymentSheet/StripePaymentSheet/Source/Internal/ECE/ECEViewController.swift
wooj-stripe: Uncommenting seems to be fine -- but in my testing, it didn't seem needed, so i commented it out.    Will uncomment.

davidme-stripe: We can just take it out then

StripePaymentSheet/StripePaymentSheet/Source/Internal/ECE/ECEViewController.swift
wooj-stripe: ShopPay has a bug w/ the close button in that it isn't hooked up -- I think we can hook into this once it's fixed and remove the native button.

wooj-stripe:
REmoving dead code check as this param will be used later.

PR #5044, opened 2025-06-25
Add LPMs and Appearance Playground to WalletButtonsView Example

davidme-stripe:
## Summary
Adding an Appearance playground to Wallet Buttons View and fixing up the SPT test scenario (adding the other LPMs) for a bug bash.

## Testing
CI, tested in PS Example

## Changelog
None

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/ExampleWalletButtonsView.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/ExampleWalletButtonsView.swift
@@ -4,7 +4,7 @@
 //
 
 @_spi(STP) @_spi(SharedPaymentToken) import StripePayments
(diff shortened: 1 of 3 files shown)


Example/PaymentSheet Example/PaymentSheet Example/ExampleWalletButtonsView.swift
```diff
@@ -309,10 +325,11 @@ class ExampleWalletButtonsModel: ObservableObject {
                 configuration.customer = .init(id: customerId, customerSessionClientSecret: customerSessionClientSecret)
                 configuration.returnURL = "payments-example://stripe-redirect"
                 configuration.willUseWalletButtonsView = true
+                configuration.appearance = self?.appearance ?? PaymentSheet.Appearance()
```
wooj-stripe, line 328: super nit: instead of having this logic twice with `?? PaymentSheet.Appearance()`, probably should have just updated the init to set it..

Issue #2670, opened 2023-06-19
[BUG] Crash in `ImageScanningConcurrencyManager.__deallocating_deinit`


nalexn:
## Summary

Stripe identity verification is crashing for some of our users.

```swift
EXC_BREAKPOINT ImageScanningConcurrencyManager.__deallocating_deinit
BUG IN CLIENT OF LIBDISPATCH: Semaphore object deallocated while in use
```

Stack trace:
```swift
Crashed background thread:
0
libdispatch.dylib
__dispatch_semaphore_dispose$VARIANT$mp.cold.1
1
libdispatch.dylib
__dispatch_semaphore_dispose$VARIANT$mp
2
libdispatch.dylib
__dispatch_dispose$VARIANT$mp
3
myapp
<compiler-generated>
ImageScanningConcurrencyManager.__deallocating_deinit
4
libswiftCore.dylib
__swift_release_dealloc
5
libswiftCore.dylib
bool swift::HeapObjectSideTableEntry::decrementStrong
6
myapp
<compiler-generated>
closure #3 in ImageScanningConcurrencyManager.scanImage
7
myapp
<compiler-generated>
thunk for () -> ()
8
libdispatch.dylib
__dispatch_call_block_and_release
9
libdispatch.dylib
__dispatch_client_callout
10
libdispatch.dylib
__dispatch_continuation_pop$VARIANT$mp
11
libdispatch.dylib
__dispatch_async_redirect_invoke
12
libdispatch.dylib
__dispatch_root_queue_drain
13
libdispatch.dylib
__dispatch_worker_thread2
14
libsystem_pthread.dylib
__pthread_wqthread

---

Main thread:

0
libobjc.A.dylib
weak_entry_for_referent
1
libobjc.A.dylib
_weak_register_no_lock
2
libobjc.A.dylib
_objc_storeWeak
3
Foundation
-[NSConcreteMapTable assign:key:value:isNew:]
4
Foundation
-[NSConcreteMapTable setObject:forKey:]
5
AVFCapture
+[AVCaptureSession _startConfiguringProperty:forDevice:]
6
AVFCapture
-[AVCaptureSession _updateVideoHDREnabledForDevice:forceResetVideoHDRSuspended:]
7
AVFCapture
-[AVCaptureSession _updateDeviceActiveFormatsAndActiveConnections]
8
AVFCapture
-[AVCaptureSession _buildAndRunGraph:]
9
AVFCapture
[...]

PR #5055, opened 2025-06-26
Attempt to fix potential race condition in image scanning session's concurrency manager 

Pearapps:
## Summary

Tracks the balance of a semaphore in the scanning session concurrency code. If that is unbalanced by the time it goes out of memory, it will resolve it. I do not see a way this crash happens, but this should resolve it in any circumstance.

## Motivation

Fixes [#2670](https://github.com/stripe/stripe-ios/issues/2670)

## Testing

I went through the identity flow a handful of times, and cancelling mid flow as well during scanning.

### DIFF
--- a/StripeIdentity/StripeIdentity/Source/NativeComponents/Coordinators/ImageScanner/ImageScanningConcurrencyManager.swift
+++ b/StripeIdentity/StripeIdentity/Source/NativeComponents/Coordinators/ImageScanner/ImageScanningConcurrencyManager.swift
@@ -59,7 +59,7 @@ final class ImageScanningConcurrencyManager: ImageScanningConcurrencyManagerProt
 
     private var amountOfFutures = 0
     private var futureQueue: DispatchQueue = DispatchQueue(label: "com.stripe.identity.concurrent-image-scanner.futures")
-    
+
     private let analyticsClient: IdentityAnalyticsClient
     private let sheetController: VerificationSheetControllerProtocol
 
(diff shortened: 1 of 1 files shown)


StripeIdentity/StripeIdentity/Source/NativeComponents/Coordinators/ImageScanner/ImageScanningConcurrencyManager.swift
```diff
@@ -69,6 +72,14 @@ final class ImageScanningConcurrencyManager: ImageScanningConcurrencyManagerProt
         self.semaphore = DispatchSemaphore(value: maxConcurrentScans)
     }
 
+    deinit {
+        futureQueue.sync {
+            for _ in 0..<amountOfFutures {
+                self.semaphore.signal()
```
gbirch-stripe, line 78: Is it required that we `signal()` the exact right number of times? Is it possible to just `signal()` `maxConcurrentScans` times to ensure we don't deinitialize with a negative balance, or does a positive balance cause problems too?

Pearapps: Yeah, the API is designed for the amount of `wait`s and `signal`s to match

Pearapps: @gbirch-stripe Is this good to resolve?

joyceqin-stripe (approved):

PR #5058, opened 2025-06-26
Fix flaky testLoaderStartShowsLoadingIndicator

davidme-stripe:
This is a weird test that keeps flaking, and the if branch here doesn't make sense. Going to try `await Task.sleep`, as I'm wondering if the main thread is being blocked by `sleep()` which would prevent `isAnimating` from updating.

### DIFF
--- a/StripeConnect/StripeConnectTests/Internal/Webview/ConnectComponentWebViewControllerTests.swift
+++ b/StripeConnect/StripeConnectTests/Internal/Webview/ConnectComponentWebViewControllerTests.swift
@@ -223,7 +223,7 @@ class ConnectComponentWebViewControllerTests: XCTestCase {
         try await webVC.webView.evaluateOnLoaderStart(elementTagName: "payouts")
 
         // Wait for the animation state to settle after the async operation
-        await Task.sleep(nanoseconds: 100_000_000) // 100ms
+        try await Task.sleep(nanoseconds: 100_000_000) // 100ms
         XCTAssertFalse(webVC.activityIndicator.isAnimating)
     }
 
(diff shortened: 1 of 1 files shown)


porter-stripe (approved):

StripeConnect/StripeConnectTests/Internal/Webview/ConnectComponentWebViewControllerTests.swift
```diff
@@ -222,13 +222,9 @@ class ConnectComponentWebViewControllerTests: XCTestCase {
...
-        if !webVC.activityIndicator.isAnimating {
-            XCTAssertFalse(webVC.activityIndicator.isAnimating)
-        } else {
-            // If it is animating, wait a second before we check.
-            sleep(1)
-            XCTAssertFalse(webVC.activityIndicator.isAnimating)
-        }
+        // Wait for the animation state to settle after the async operation
+        await Task.sleep(nanoseconds: 100_000_000) // 100ms
+        XCTAssertFalse(webVC.activityIndicator.isAnimating)
```
wooj-stripe: Initially there was no if/else, and was only:
```
            XCTAssertFalse(webVC.activityIndicator.isAnimating)
```

I tried to de-flake it with the if/else statement you see there. Your proposed approach seems to introduce a less generous (100ms) sleep, which seems less likely to work?

davidme-stripe: I think we need to wait for the JS to finish executing so that it can tell the activity indicator to stop animating, 100ms after executing the JS on the loaded page seems like plenty of time for that to happen. (And it seems to work in this run.) But when we were calling sleep(), I think it was blocking the main thread, so it woke up and immediately checked `isAnimating` before the webview had a chance to tell the activity indicator to stop.

The if statement was also confusing to me, because we're checking if it's false and then immediately asserting that it's false?

wooj-stripe: Ah! Blocking the main thread -- maybe? in that case, yea.. lets try. try await

True.   Guess if it was false, there was no need to assert.

wooj-stripe (approved):

PR #5066, opened 2025-06-27
[flakebot] Fix race condition in VerificationSheetFlowControllerTest

davidme-stripe:
[testing a bot]

## Summary
Fixed segmentation faults in `testMissingDocFrontNoType()` and `testNextViewControllerDocumentWarmup()` caused by a race condition in the Future/Promise implementation.

### Root Cause
The race condition occurred because the `Future.result` property was not properly synchronized:
- `Promise` methods (`resolve`, `reject`) were directly assigning to `result` without queue protection
- `Future.observe()` was reading `result` from within `propertyAccessQueue` 
- Both threads could access the same `Result<AnyDocumentScanner, Error>` simultaneously during `objc_retain` operations

### Solution
- Added private `_result` backing storage
- Made `result` property getter/setter use `propertyAccessQueue.sync/async` for proper synchronization
- Simplified `report()` method since it's always called from queue context

### Test Plan
- [x] Run `VerificationSheetFlowControllerTest/testMissingDocFrontNoType()` 
- [x] Run `VerificationSheetFlowControllerTest/testNextViewControllerDocumentWarmup()`
- [ ] CI will verify no regressions across all Future/Promise usage

### DIFF
--- a/StripeCore/StripeCore/Source/Helpers/Async.swift
+++ b/StripeCore/StripeCore/Source/Helpers/Async.swift
@@ -36,11 +36,15 @@ import Foundation
 @_spi(STP) public class Future<Value> {
     public typealias Result = Swift.Result<Value, Error>
 
+    private var _result: Result?
     fileprivate var result: Result? {
-        // Observe whenever a result is assigned, and report it:
-        didSet {
+        get {
+            return propertyAccessQueue.sync { _result }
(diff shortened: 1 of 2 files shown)


davidme-stripe:
We should get rid of `Async.swift`, but this seems okay for now

PR #5076, opened 2025-06-27
Add more Shop Pay testing options

davidme-stripe:
Add more test options for the Shop Pay playground

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/ExampleWalletButtonsView.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/ExampleWalletButtonsView.swift
@@ -616,11 +616,7 @@ class ExampleWalletButtonsModel: ObservableObject {
         return config
     }
     func isValidShippingLocation(_ address: PaymentSheet.ShopPayConfiguration.PartialAddress) -> Bool {
-        // Check postal code restriction
-        if address.postalCode == "91911" {
-            return false
-        }
-        return true
+        return address.postalCode != "91911"
     }
 }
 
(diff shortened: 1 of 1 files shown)


Example/PaymentSheet Example/PaymentSheet Example/ExampleWalletButtonsView.swift
```diff
@@ -427,6 +479,13 @@ class ExampleWalletButtonsModel: ObservableObject {
     func onCompletion(result: PaymentSheetResult) {
         self.addDebugLog("PaymentSheet completion called with result: \(result)")
 
+        // Check if we should simulate payment failure for testing
+        if shopPayTestingOptions.simulatePaymentFailed {
```
wooj-stripe, line 487: Is never called in the shopPay flow because this is only hooked up to the 'confirm' button w/ the flow controller.  Seems like in `createPaymentIntentWithSPTTestBackend` -- any time we set  `self.paymentResult`, we should just call `onCompletion(result:....)`

wooj-stripe: Also on line 315, in the function `preparePaymentSheetWithOriginalBackend` where it says:
```
                        // Create the SPT on your backend here
```

Maybe lets call:
```
                        self?.onCompletion(result: .completed)
```

wooj-stripe: Proposing additional changes here (e.g. a reload button & success) to make success/failure cases in bug bashing more obvious.

https://github.com/stripe/stripe-ios/compare/add-shop-pay-testing-options...add-shop-pay-testing-options-fixHandleFail?expand=1

PR #5080, opened 2025-06-28
Update subtitle element padding

joyceqin-stripe:
## Summary

Update the subtitle element padding to the top instead of the bottom, [distance between title and subtitle element is 50% more than between subtitle element and form fields](https://stripe.slack.com/archives/CFA2HJ99A/p1751069508270919?thread_ts=1751064832.916989&cid=CFA2HJ99A)
[Confirmation from Lawrence](https://stripe.slack.com/archives/CFA2HJ99A/p1751302584166259?thread_ts=1751064832.916989&cid=CFA2HJ99A)
## Motivation

https://jira.corp.stripe.com/browse/MOBILESDK-3563

## Testing

Snapshots
## Changelog

N/A


StripeUICore/StripeUICore/Source/Elements/SubtitleElement.swift
```diff
@@ -24,7 +24,7 @@ import UIKit
         view.translatesAutoresizingMaskIntoConstraints = false
         containerView.addSubview(view)
 
-        let padding = UIEdgeInsets(top: 0, left: 0, bottom: 20, right: 0)
```
wooj-stripe, line 27: In PaymentMethodFormViewController, seems like we are already adding 4px of spacing:
```
stackView.spacing = form.getAllUnwrappedSubElements().contains(where: { $0 is SubtitleElement }) ? 4 : 24
```
By adding 14 here, are we saying there should really be 18px here?  Maybe we just update the formStackView to own the spacing between the element?

(but if it turns out to be really difficult, then this change is fine)

joyceqin-stripe: Yes, we actually want it to be 18px because the space between the subtitle element and the form fields is 12px. I considered updating it to 18px and removing the 4, but if we remove the padding from the subtitle element and let formStackView have ownership of the spacing, then passing in `isHorizontalMode` becomes moot. I figured making it 14 here and adding 4 would have essentially the same outcome, but if for whatever reason we wanted to change the padding later eventually, then it'd be easier (and we wouldn't have to reintroduce `isHorizontalMode`

wooj-stripe (approved):

PR #5083, opened 2025-06-30
Add new checkbox to AddressElement

porter-stripe:
## Summary
- Adds a "use billing as shipping" checkbox to address element.

https://docs.google.com/document/d/1bHMVR9eqzdYuAHyiZ7j4v62u4kwILpKxNghx5wqWxZU/edit?tab=t.0

## Motivation
🌲 

## Testing
- Manual
- New tests

## Changelog
N/A


Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
```diff
@@ -368,6 +368,7 @@ class PlaygroundController: ObservableObject {
             configuration.allowedCountries = ["US", "CA", "MX", "GB"]
         }
         configuration.additionalFields.checkboxLabel = "Save this address for future orders"
+        configuration.showUseBillingAddressCheckbox = true
```
wooj-stripe, line 371: seems like this should be configurable rather than hardcoded to true?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/AddressViewController/AddressViewController.swift
```diff
@@ -98,7 +98,18 @@ public class AddressViewController: UIViewController {
 
     // MARK: - Elements
     lazy var formElement: FormElement = {
-        let formElement = FormElement(elements: [addressSection, checkboxElement], theme: configuration.appearance.asElementsTheme)
+        var customSpacing: [(Element, CGFloat)] = []
+
+        // Add padding under the shipping equals billing checkbox if it exists
+        if let shippingCheckbox = shippingEqualsBillingCheckbox {
+            customSpacing.append((shippingCheckbox, PaymentSheetUI.defaultPadding))
+        }
```
wooj-stripe: Just double checking -- should we be using default padding or is there a configurable value in Appearance api?

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheet+AddressTests.swift
```diff
@@ -368,4 +371,68 @@ NZ
...
+
+        let checkbox = app.switches["Use billing address for shipping"]
+        XCTAssertTrue(checkbox.waitForExistence(timeout: 2.0))
+        XCTAssertTrue(checkbox.isSelected)
+
+        // 1. Manually toggle off -> on
+        checkbox.tap()
+        XCTAssertFalse(checkbox.isSelected)
+        checkbox.tap()
+        XCTAssertTrue(checkbox.isSelected)
```
wooj-stripe: As we tap these on and off -- seems like we should also validate that we are filling in the name/address/etc

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/AddressViewController/AddressViewController.swift
```diff
@@ -114,6 +125,26 @@ public class AddressViewController: UIViewController {
 
         return element
     }()
+
+    private var validDefaultValues: AddressViewController.Configuration.DefaultAddressDetails? {
+        guard configuration.defaultValues.address != .init() else { return nil }
```
wooj-stripe: Instead of:
```
configuration.defaultValues.address != .init()
```
Seems like we should add a descriptive extension on `address`, like -- `.isEmpty`?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/AddressViewController/AddressViewController.swift
```diff
@@ -114,6 +125,26 @@ public class AddressViewController: UIViewController {
...
+        guard !configuration.allowedCountries.isEmpty else { return configuration.defaultValues }
```
wooj-stripe, line 136: Why is it important that users set `allowedCountries` (but then defaults and returns?)

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/AddressViewController/AddressViewController.swift
```diff
@@ -114,6 +125,26 @@ public class AddressViewController: UIViewController {
 
         return element
     }()
+
+    private var validDefaultValues: AddressViewController.Configuration.DefaultAddressDetails? {
```
wooj-stripe: In general, i'm having trouble understanding what this function  -- maybe there are ways to make it clearer what we're trying to accomplish here?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/AddressViewController/AddressViewController.swift
porter-stripe: I was on the fence about this, technically we have "sectionSpacing" now, however, the default value (12) just doesn't look good w/ this checkbox. We could do some scale factor of it like appearance.sectionSpacing * 1.6 to get some better looking value. I think I'll do that.

Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
porter-stripe: For the sake of the keeping complexity on the playground down I'm going to leave this as hardcoded, it's also a niche feature that only shows up occasionally. We also don't allow configuring any of the other Address Element APIs (allowed countries, default address, checkbox label) in the playground, maybe we can add these in later in a new view.

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheet+AddressTests.swift
porter-stripe: Yeah good idea, updating this.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/AddressViewController/AddressViewController.swift
porter-stripe: Added comments to this variable to better explain when/why we allow certain default addresses and not others.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/AddressViewController/AddressViewController.swift
porter-stripe: Probably addressed w/ https://github.com/stripe/stripe-ios/pull/5083#discussion_r2178320907

PR #5090, opened 2025-06-30
Reinstate saved playground appearance

gbirch-stripe:
## Summary
Reinstating change removed in https://github.com/stripe/stripe-ios/pull/5057. Appearance settings now saved to user defaults. Better font encoding / decoding to prevent issue in which font would change to Times New Roman.

## Motivation
Ease of testing appearance settings in playground.

## Testing
Verified that no issue happened with the repro steps that previously reliably caused the bug (open payment sheet playground, open appearance, close without changing anything, back out to main menu, re-enter payment sheet playground, open payment sheet)

## Changelog
N/A


Example/PaymentSheet Example/PaymentSheet Example/Appearance+Codable.swift
```diff
@@ -0,0 +1,348 @@
...
+
+        try container.encode(textFieldInsets, forKey: .textFieldInsets)
+        try container.encode(formInsets, forKey: .formInsets)
+
+        // Font properties - will need to be expanded if more options are added to the playground
+        try container.encode(font.sizeScaleFactor, forKey: .fontSizeScaleFactor)
+        let fontBaseDescriptorData = try NSKeyedArchiver.archivedData(withRootObject: font.base.fontDescriptor, requiringSecureCoding: false)
+        try container.encode(fontBaseDescriptorData, forKey: .fontBaseDescriptor)
+        if let headline = font.custom.headline {
+            let headlineFontDescriptorData = try NSKeyedArchiver.archivedData(withRootObject: headline.fontDescriptor, requiringSecureCoding: false)
```
davidme-stripe, line 241: I generally wouldn't recommend this in production (especially without `requiringSecureCoding`!) but it seems fine here

PR #5105, opened 2025-07-02
Update example shipping options for shopPay

wooj-stripe:
## Summary
Updates wording to avoid confusion

## Motivation
too confusing

## Testing
manual

### DIFF
--- a/StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/ECE/ECEIntegrationTests.swift
+++ b/StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/ECE/ECEIntegrationTests.swift
@@ -49,7 +49,7 @@ class ECEIntegrationTests: XCTestCase {
                     id: "standard",
                     amount: 1000,
                     displayName: "Standard Shipping",
-                    deliveryEstimate: PaymentSheet.ShopPayConfiguration.DeliveryEstimate(
+                    deliveryEstimate: PaymentSheet.ShopPayConfiguration.DeliveryEstimate.structured(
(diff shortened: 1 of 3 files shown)


Example/PaymentSheet Example/PaymentSheet Example/ExampleWalletButtonsView.swift
```diff
@@ -509,11 +509,12 @@ class ExampleWalletButtonsModel: ObservableObject {
 
     var shopPayConfiguration: PaymentSheet.ShopPayConfiguration {
         let singleBusinessDay = PaymentSheet.ShopPayConfiguration.DeliveryEstimate.DeliveryEstimateUnit(value: 1, unit: .business_day)
```
davidme-stripe: Do we not support next day? Maybe we should have the example use the string-based delivery estimate? https://docs.stripe.com/js/elements_object/create_express_checkout_element#express_checkout_element_create-options-shippingRates-deliveryEstimate

wooj-stripe: Aie! thank you for calling this out -- this API needs to be re-worked!

wooj-stripe: Screenshot 2025-07-02 at 9 53 51 PM

Demonstrates all different cases -- the "NaN to Nan business days" is assumed to be a bug in the wallet UI's code. If we are good with the changes to the API -- i will update our docs!

davidme-stripe: Seems good! Let's flag NaN to NaN to the owning team

PR #5120, opened 2025-07-07
Always allow user interaction when presenting 3DS

mats-stripe:
## Summary

This fixes an issue where user interaction was disabled when presenting 3DS from the Pay with Link controller. In BottomSheetViewController we can force-enable user interaction when the 3DS controller is being presented.

## Motivation

https://stripe.slack.com/archives/C08G211PLER/p1751885395251659

## Testing

Before:

https://github.com/user-attachments/assets/87c75d62-7aed-456e-928d-ab33e17ea9b2

After:

https://github.com/user-attachments/assets/0e852f91-9099-4cb4-9cec-3a38e6e551f0

## Changelog

N/a

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/BottomSheetViewController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/BottomSheetViewController.swift
@@ -507,6 +507,15 @@ extension BottomSheetViewController: PaymentSheetAuthenticationContext {
         let threeDS2ViewController = BottomSheet3DS2ViewController(
             challengeViewController: authenticationViewController, appearance: appearance, isTestMode: isTestMode)
         threeDS2ViewController.delegate = self
+
+        // Re-enable user interaction when presenting the 3DS controller.
+        // This fixes an issue where the PayWithLinkViewController disables user interaction
+        // during payment confirmation but the 3DS controller needs to be interactable.
+        let wasUserInteractionEnabled = view.isUserInteractionEnabled
+        if !wasUserInteractionEnabled {
+            view.isUserInteractionEnabled = true
+        }
+
         pushContentViewController(threeDS2ViewController)
         // Remove a blur effect, if any
         self.removeBlurEffect(animated: true, completion: completion)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/BottomSheetViewController.swift
```diff
@@ -507,6 +507,15 @@ extension BottomSheetViewController: PaymentSheetAuthenticationContext {
...
             challengeViewController: authenticationViewController, appearance: appearance, isTestMode: isTestMode)
         threeDS2ViewController.delegate = self
+
+        // Re-enable user interaction when presenting the 3DS controller.
+        // This fixes an issue where the PayWithLinkViewController disables user interaction
+        // during payment confirmation but the 3DS controller needs to be interactable.
+        let wasUserInteractionEnabled = view.isUserInteractionEnabled
+        if !wasUserInteractionEnabled {
+            view.isUserInteractionEnabled = true
+        }
```
tillh-stripe: Should we have this logic in `PayWithLinkViewController.pushContentViewController()` instead?

mats-stripe: Done in https://github.com/stripe/stripe-ios/pull/5120/commits/6809fd2bb06419120749c63b419b4d684f31f6b6

mats-stripe:
Fixed back navigation in https://github.com/stripe/stripe-ios/pull/5120/commits/22190234ff0b654b36199b8368e50e93fe5eda1f


https://github.com/user-attachments/assets/a008c772-c1cb-4b6c-9a2e-34afde71b593

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController.swift
```diff
@@ -268,6 +268,13 @@ final class PayWithLinkViewController: BottomSheetViewController {
 
     override func pushContentViewController(_ contentViewController: any BottomSheetContentViewController) {
         super.pushContentViewController(contentViewController)
+
+        // Re-enable user interaction when presenting a new controller.
+        let wasUserInteractionEnabled = view.isUserInteractionEnabled
+        if !wasUserInteractionEnabled {
+            view.isUserInteractionEnabled = true
+        }
```
tillh-stripe, line 276: We _could_ limit this to 3DS view controllers, but I think we should be fine even with this implementation.

mats-stripe: I think this is OK

PR #5122, opened 2025-07-07
Update deserilization logic to allow deserilization independent between ps and cs

wooj-stripe:
## Summary
Make deserialization indepenent between ps and cs -- currently merchant is REQUIRED to be in both gates

## Motivation
Reduce operational overhead to onboard a merchant to both gates

## Testing
Added tests

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/CustomerSession.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/CustomerSession.swift
@@ -25,17 +25,15 @@ struct CustomerSession: Equatable, Hashable {
               let apiKey = response["api_key"] as? String,
               let apiKeyExpiry = response["api_key_expiry"] as? Int,
               let customer = response["customer"] as? String,
-              let componentsDict = response["components"] as? [AnyHashable: Any],
-              let mobilePaymentElementDict = componentsDict["mobile_payment_element"] as? [AnyHashable: Any],
-              let mobilePaymentElementEnabled = mobilePaymentElementDict["enabled"] as? Bool,
-              let customerSheetDict = componentsDict["customer_sheet"] as? [AnyHashable: Any],
-              let customerSheetEnabled = customerSheetDict["enabled"] as? Bool
+              let componentsDict = response["components"] as? [AnyHashable: Any]
         else {
             return nil
         }
 
         var mobilePaymentElementComponent: MobilePaymentElementComponent
-        if mobilePaymentElementEnabled {
+        if let mobilePaymentElementDict = componentsDict["mobile_payment_element"] as? [AnyHashable: Any],
+           let mobilePaymentElementEnabled = mobilePaymentElementDict["enabled"] as? Bool,
+           mobilePaymentElementEnabled {
             guard let mobilePaymentElementFeaturesDict = mobilePaymentElementDict["features"] as? [AnyHashable: Any],
                   let paymentMethodSave = mobilePaymentElementFeaturesDict["payment_method_save"] as? String,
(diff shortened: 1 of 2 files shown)


StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/STPElementsSessionTest.swift
```diff
@@ -504,6 +504,37 @@ class STPElementsSessionTest: XCTestCase {
...
+                                                                "customer_sheet": [
+                                                                    "enabled": true,
+                                                                    "features": [
+                                                                        "payment_method_remove": "enabled",
+                                                                        "payment_method_sync_default": "enabled",
+                                                                    ],
+                                                                ],
+                                                            ])
+
+        let allowsSetAsDefault = elementsSession.paymentMethodSyncDefaultForCustomerSheet
```
joyceqin-stripe: Extremely nit but name would more accurately be allowsSyncDefault

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/CustomerSession.swift
```diff
@@ -25,17 +25,15 @@ struct CustomerSession: Equatable, Hashable {
               let apiKey = response["api_key"] as? String,
               let apiKeyExpiry = response["api_key_expiry"] as? Int,
               let customer = response["customer"] as? String,
-              let componentsDict = response["components"] as? [AnyHashable: Any],
-              let mobilePaymentElementDict = componentsDict["mobile_payment_element"] as? [AnyHashable: Any],
-              let mobilePaymentElementEnabled = mobilePaymentElementDict["enabled"] as? Bool,
-              let customerSheetDict = componentsDict["customer_sheet"] as? [AnyHashable: Any],
-              let customerSheetEnabled = customerSheetDict["enabled"] as? Bool
+              let componentsDict = response["components"] as? [AnyHashable: Any]
```
joyceqin-stripe, line 28: Everything looks fine to me, but just wanted to check if there was ever a reason that it needed both mobile payment element and customer sheet previously?

joyceqin-stripe (dismissed):

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/CustomerSession.swift
wooj-stripe: Not that I remember. I think it was a copy paste error.  We probably had mobile_payment_element in there and made it non-optional because it was required, then when we added CustomerSheet, we incorrectly added it to the same spot.  Both `mobile_payment_element` and `customer_sheet` are required to be there (but only if you're in both gates).

joyceqin-stripe (approved):

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/STPElementsSessionTest.swift
```diff
@@ -504,6 +504,36 @@ class STPElementsSessionTest: XCTestCase {
...
+                                                                    "enabled": true,
+                                                                    "features": ["payment_method_save": "enabled",
+                                                                                 "payment_method_remove": "enabled",
+                                                                                 "payment_method_set_as_default": "enabled",
+                                                                                ],
+                                                                ],
+                                                            ])
+
+        let allowsSetAsDefault = elementsSession.paymentMethodSetAsDefaultForPaymentSheet
+        XCTAssertTrue(allowsSetAsDefault)
```
joyceqin-stripe, line 520: I guess this could also be a 1-line assert (no need for allowsSetAsDefault)

PR #5125, opened 2025-07-08
Scale error fonts in payment sheet

gbirch-stripe:


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
```diff
@@ -1081,6 +1081,8 @@ extension PaymentSheet.Appearance {
         fonts.sectionHeader = scaledFont(for: font.base.medium, style: .footnote, maximumPointSize: 18)
         fonts.caption = scaledFont(for: font.base.regular, style: .caption1, maximumPointSize: 20)
         fonts.footnote = scaledFont(for: font.base.regular, style: .footnote, maximumPointSize: 20)
+        fonts.error = scaledFont(for: font.base.regular, style: .caption2, maximumPointSize: 20)
+        fonts.smallFootnote = scaledFont(for: font.base.medium, style: .caption2, maximumPointSize: 18)
```
joyceqin-stripe, line 1085: Is there anywhere in particular the maximumPointSizes came from? Non-blocking and they seem like reasonable values but just wondering

gbirch-stripe: I was unable to find a definitive source for these numbers but decided to keep them in line with the rest of the UI

PR #5130, opened 2025-07-08
Use Link branding for PTM payment methods in SPM

tillh-stripe:
## Summary

This pull request extends the Link branding beyond payment methods of type `link` to payment methods that were created in passthrough mode.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/ElementsCustomer.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/ElementsCustomer.swift
@@ -57,9 +57,10 @@ struct ElementsCustomer: Equatable, Hashable {
             if enableLinkInSPM {
                 if let paymentMethodWithLinkDetails = PaymentMethodWithLinkDetails.decodedObject(fromAPIResponse: json) {
                     let paymentMethod = paymentMethodWithLinkDetails.paymentMethod
-                    paymentMethod.isLinkOrigin = paymentMethodWithLinkDetails.isLinkOrigin
                     if let linkDetails = paymentMethodWithLinkDetails.linkDetails {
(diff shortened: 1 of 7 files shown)


StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/ElementsCustomer.swift
```diff
@@ -59,6 +59,8 @@ struct ElementsCustomer: Equatable, Hashable {
                     let paymentMethod = paymentMethodWithLinkDetails.paymentMethod
                     if let linkDetails = paymentMethodWithLinkDetails.linkDetails {
                         paymentMethod.setLinkPaymentDetails(from: linkDetails)
+                    } else {
+                        paymentMethod.isLinkPassthroughMode = paymentMethodWithLinkDetails.isLinkOrigin
```
tillh-stripe, line 63: A payment method is only created in Link passthrough if `isLinkOrigin` is true but `linkDetails` is nil.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
```diff
@@ -779,7 +779,7 @@ extension PaymentOption {
         let hasLinkAccount = LinkAccountContext.shared.account?.isRegistered ?? false
         switch self {
         case .saved(let paymentMethod, _):
-            return paymentMethod.isLinkPaymentMethod && hasLinkAccount
+            return (paymentMethod.isLinkPaymentMethod || paymentMethod.isLinkPassthroughMode) && hasLinkAccount
```
tillh-stripe, line 782: Treat passthrough and payment method mode the same when it comes to the return-user experience.

davidme-stripe (dismissed):

PR #5131, opened 2025-07-08
Add configuration to control Link billing details collection

carlosmuvi-stripe:
## Summary
Added a new configuration option to control whether missing billing details should be collected for existing Link payment methods. This gives developers more control over the Link payment flow.

## Motivation
This change allows developers to skip the billing details collection step for existing Link payment methods when they don't need that information or want to streamline the payment flow. Previously, if a Link payment method was missing required billing details, the SDK would always prompt the user to fill them in.

## Testing
- [x] Code compiles successfully with `xcodebuild`
- [x] Created test scenarios for both `true` and `false` configuration values
- [x] Verified backward compatibility - defaults to `true` to maintain existing behavior
- [x] Tested that the logic properly checks the configuration setting in `PayWithLinkViewController`

## Changelog
- [Added] New `collectMissingBillingDetailsForExistingPaymentMethods` configuration option in `LinkConfiguration` to control billing details collection for existing Link payment methods.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController-WalletViewController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController-WalletViewController.swift
@@ -270,7 +270,7 @@ extension PayWithLinkViewController {
                     }
                 }
 
-                if isMissingRequestedBillingDetails(paymentDetails) {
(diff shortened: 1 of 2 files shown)


mats-stripe (dismissed):

mats-stripe:
I don't think we want to mention this in the changelog yet, right?

carlosmuvi-stripe:
> I don't think we want to mention this in the changelog yet, right?

we don't!

PR #3773, opened 2024-07-15
Re-enable refresh for CAP

porter-stripe:
## Summary
- Re-enables the refresh endpoint for Cash App Pay
- Disables polling
- Expands the payment method when using the refresh endpoint

## Motivation
https://docs.google.com/document/d/1Wq--3AO9CUEmUBRdktrJil-dgmKuCRTPBNQzgdYjm2o/edit#heading=h.g9hom2g12r6t

## Testing
- Manual in live and test mode
- Updated a few tests

## Changelog
See diff

PR #5133, opened 2025-07-08
Remove redundant E2E tests from PaymentSheetLPMUITest.swift

davidme-stripe:
## Summary
Removed redundant E2E tests from `PaymentSheetLPMUITest.swift`, as we have existing tests for these in `PaymentSheetLPMConfirmFlowTests.swift`. (In the few cases where we *don't* have existing integration tests, I've added some.) We don't need to end-to-end test all of these, especially redirect-based payment methods, as they all follow basically the same flow.

* testEPS (and added a new testEPSConfirmFlows in PaymentSheetLPMConfirmFlowTests.swift)
* testKlarnaPaymentMethod
* testAffirmPaymentMethod (and added a new testAffirmConfirmFlows in PaymentSheetLPMConfirmFlowTests.swift)
* testAmazonPayPaymentMethod 
* testAlmaPaymentMethod 
* testSunbitPaymentMethod
* testBilliePaymentMethod
* testSatispayPaymentMethod 
* testCryptoPaymentMethod
* testZipPaymentMethod (and added a new testZipConfirmFlows in PaymentSheetLPMConfirmFlowTests.swift)
* testCashAppPaymentMethod
* testCashAppPaymentMethod_setup
* testCashAppPaymentMethod_setupFutureUsage
* testAmazonPayPaymentMethod_setup
* testPayPalPaymentMethod_setup
* testRevolutPayPaymentMethod_setup
* testGrabPayPaymentMethod
* testSEPADebitPaymentMethod_PaymentSheet
* testAlipayPaymentMethod
* testPayNowPaymentMethod
* testPromptPayPaymentMethod
* testSwishPaymentMethod
* testBlikPaymentMethod
* testBacsDebit
* testMultibancoPaymentMethod
* testOXXOPaymentMethod
* testBoletoPaymentMethod

* Retained testUPIPaymentMethod, as it tests some custom UI behavior.


Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetLPMUITest.swift
```diff
@@ -8,417 +8,40 @@
...
-        let setupButton = app.buttons["Set up"]
-
-        // Select Cash App
-        tapPaymentMethod("Cash App Pay")
-
-        // Attempt set up
-        setupButton.tap()
-
-        // Close the webview, to simulate cancel
-        app.otherElements["TopBrowserBar"].buttons["Close"].waitForExistenceAndTap(timeout: 15)
```
yuki-stripe, line 309: @porter-stripe is there a reason to keep this cancel behavior UI test for cash app?

porter-stripe: IIRC, this was added due to an incident where CAP was broken in test mode and cancels were reporting successes. But there is nothing special about the CAP cancel behavior.

yuki-stripe: Sounds like a bug in the paymenthandler + polling code? Maybe we should keep this around then. And maybe try to write a more targeted test for whatever broke?

davidme-stripe: Looks like it was added here: https://github.com/stripe/stripe-ios/pull/3773#pullrequestreview-2178225433

I'm having trouble finding an incident for this. I wouldn't be opposed to keeping this one in as a test that (maybe more generically?) is exercising both cancel and success.

porter-stripe: Actually, it may have not been an incident. But actually when we adopted the refresh endpoint for CAP, they did not implement the cancel in test mode, which prompted us to add this test. I vote to keep this as it's the only LPM using the refresh endpoint at this point.

yuki-stripe: Let's also add a comment with this context/reason

davidme-stripe: Done!

yuki-stripe (approved):

PR #5134, opened 2025-07-09
Add intrinsicContentSize to CardFormView to allow presentation in SwiftUI to work properly

wooj-stripe:
## Summary
SwiftUI is having trouble rendering this view when surrounded by high priority views.  To remedy, returning the intrinsic content size seems to remedy this.  Not ideal as having to manually calculate it is difficult.  If there are other approaches I'm open to it, but given we only had to re-record a single snapshot test, i'm okay moving forward with this solution.

## Motivation
Surrounding the view with high priority views was breaking this view in swift UI

## Testing
Existing tests in:
* STPStackViewWithSeparatorSnapshotTests
* STPFormViewSnapshotTests
* STPCardFormViewSnapshotTests
Manually tested by using the UI Examples app and testing w/ surrounding between two high priority views.


StripePaymentsUI/StripePaymentsUI/Source/UI Components/STPFormView.swift
```diff
@@ -65,6 +65,22 @@ public class STPFormView: UIView, STPFormInputValidationObserver {
     static let cornerRadius: CGFloat = 6
     static let interSectionSpacing: CGFloat = 7
 
+    public override var intrinsicContentSize: CGSize {
```
porter-stripe, line 68: Any reason we can't just override this in the STPCardFormView.Representable? I think overriding this where we us constraints and intrinsicContentSize may not be correct but not sure.

porter-stripe (approved):

PR #5135, opened 2025-07-09
[Fixed] EmbeddedPaymentElement now errors when `configuration.rowSele…

yuki-stripe:
…ctionBehavior` is `.immediateAction` and `configuration.embeddedViewDisplaysMandateText` is `true` to ensure integrations display legal mandate text to the customer.

## Motivation
rowSelectionBehavior = immediateAction + embeddedViewDisplaysMandateText = false is a broken integration.

1. When formSheetAction = continue...
...if you tap a `a PM row with no form and a mandate` you'll never see the mandate that appears in embedded because you're immediately moved to a different screen.

2. When formSheetAction = confirm...
...if you tap `a PM row with no form and a mandate` you'll never see the mandate that appears in embedded because you will ~immediately (after a network request or two) see a webview to authenticate.

## Testing
See unit test

## Changelog
See changelog


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -569,6 +569,14 @@ extension EmbeddedPaymentElement {
     static func validateRowSelectionConfiguration(configuration: Configuration) throws {
         switch configuration.rowSelectionBehavior {
         case .immediateAction:
+            if configuration.embeddedViewDisplaysMandateText {
+                switch configuration.formSheetAction {
```
gbirch-stripe, line 573: What's the interaction with `formSheetAction` here? Are we expecting that there might be some future third `formSheetAction` with different behavior? It seems that right now it doesn't make a difference?

yuki-stripe: The error messages are slightly different

PR #5146, opened 2025-07-10
Adds New Module Structure for StripeCryptoOnramp

mliberatore:
## Summary
Introduces a new module that will contain Crypto Onramp functionality documented at [this internal tech spec](https://docs.google.com/document/d/1owKwHgD5s8u75EWDOny2bp6rlNZq1aI4NNUr43PKdVA/edit?tab=t.0#heading=h.5mbaom37vv77).

## Motivation
Please see the tech spec linked above and the links within for more information about the project. The motivation for this PR in particular is to start adding the module structure so we can begin the implementation work. We’ll likely eventually add an example app target to exercise new SDK functionality.

## Testing
1. Ensured that the project built and the placeholder test passed from within the new Xcode project.
2. Ensured that the project built and the placeholder test passed when selecting the StripeCryptoOnramp target from within the Workspace.
3. Compared the overall structure of the project to other SDKs in the workspace, ensuring we're following patterns and utilizing the appropriate build configurations.

CleanShot 2025-07-10 at 12 29 09
CleanShot 2025-07-10 at 12 29 25

## Changelog
N/A as the SDK has not yet been implemented, so nothing was added that is usable by clients.


Package.swift
```diff
@@ -161,6 +165,15 @@ let package = Package(
             name: "StripeConnect",
             dependencies: ["StripeCore", "StripeUICore", "StripeFinancialConnections"],
             path: "StripeConnect/StripeConnect"
+        ),
+        .target(
+            name: "StripeOnramp",
+            dependencies: ["StripeCore"],
```
mats-stripe, line 171: This module will eventually have more dependencies than this, but this is OK for now 👍

PR #5149, opened 2025-07-10
Add authentication-only API to LinkController

mats-stripe:
## Summary

This adds an API to the LinkController to allow authenticating a Link user. As part of this, we had to decouple the SignUp controller from the `PayWithLinkController`.  Here's a summary of the changes: 

- Creates the `LinkSignUpController`, which presents the `LinkSignUpViewController` in a bottom sheet, and allows verifying Link users in a `requiresVerification` state.
- Add an `presentForAuthentication` API to `LinkController`, which will lookup the Link account and either sign in or sign up the user.
- Adds an example screen to demo the `LinkController`.

## Motivation

👻

## Testing

Demo: 

https://github.com/user-attachments/assets/f425c191-f830-4842-acc1-553b0da58d83

## Changelog

N/a


StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/STPAPIClient+Link.swift
```diff
@@ -35,7 +35,7 @@ extension STPAPIClient {
             if doNotLogConsumerFunnelEvent {
                 parameters["do_not_log_consumer_funnel_event"] = true
             }
-            if let email, let emailSource {
+            if let email, !email.isEmpty, let emailSource {
```
mats-stripe, line 38: I'm getting an error when calling this API with an empty email: 

```
"You passed an empty string for 'email_address'. We assume empty values are an attempt to unset a parameter; however 'email_address' cannot be unset. You should remove 'email_address' from your request or supply a non-empty value."
```

Adding this check fixes that.

mats-stripe:
Good catch @mliberatore - fixed it in https://github.com/stripe/stripe-ios/pull/5149/commits/21e54466626517b92b9f4522a86a9e7fa08f3939. For now, I dismiss the keyboard before presenting the LinkController which fixes the animation jank. There's probably a deeper issue with our bottom sheet component and keyboard recognizers happening, but since this issue is also happens with our other bottom sheets (like MPE) it's something we can look into outside the scope of this PR 👍

mliberatore:
Sounds good, thanks for the patch :+1:

mats-stripe:
Closing in favor of https://github.com/stripe/stripe-ios/pull/5189 and https://github.com/stripe/stripe-ios/pull/5190

PR #5150, opened 2025-07-11
SwiftUI support for AddressElement

porter-stripe:
## Summary
- Adds SwiftUI APIs for AddressElement
- Adds an example view and tests

https://github.com/user-attachments/assets/330af63a-0ba5-4520-b143-89b9ad5b25c7

> Note: There is some animation jank after selecting an auto complete address, but this occurs on the UIKit integration as well. Something we should look into.

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-3335

## Testing
- Manual
- New UI tests
- New snapshot tests

## Changelog
See diff


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/AddressViewController/AddressViewController+SwiftUI.swift
```diff
@@ -0,0 +1,149 @@
...
+        init(address: Binding<AddressViewController.AddressDetails?>, dismiss: @escaping () -> Void) {
+            self.address = address
+            self.dismiss = dismiss
+        }
+
+        func addressViewControllerDidFinish(
+            _ addressViewController: AddressViewController,
+            with address: AddressViewController.AddressDetails?
+        ) {
+            self.address.wrappedValue = address
```
davidme-stripe, line 87: This doesn't seem to be called if I dismiss the sheet by swiping it down

davidme-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/AddressViewController/AddressViewController+SwiftUI.swift
porter-stripe: Will fix in a following PR, impacts UIKit too!

PR #5158, opened 2025-07-15
StripeCryptoOnramp: Additional Setup & Begins Public API

mliberatore:
## Summary
- Added remaining dependencies listed in the [Onramp tech spec under iOS > SDK Setup](https://docs.google.com/document/d/1owKwHgD5s8u75EWDOny2bp6rlNZq1aI4NNUr43PKdVA/edit?tab=t.0#heading=h.vqz32ccsplag) to reduce noise in future PRs. Will remove any if we end up not using them.
- Adds the very beginnings of the public API, which is still being finalized in some aspects.
- Addresses the inability to pass in an `STPAPIClient` to `LinkController`.

## Motivation

Necessary for the Crypto Onramp project.

Looking to keep moving forward and getting early feedback as I continue coming up to speed. I thought opening small changes like this would give a good opportunity for feedback, notes on approach, and the ability to ask others questions regarding conventions.

## Testing
No tests yet. I tried the API out manually by wiring up a button to the payment sheet example app. I navigated to another example first to establish a test `publishableKey`.

```swift
@_spi(CryptoOnrampSDKPreview) import StripeCryptoOnramp

// …

Button("CryptoOnramp isLinkUser") {
    Task {
        do {
            let coordinator = try await CryptoOnrampCoordinator.create(apiClient: .shared, appearance: .init())
            let result1 = try await coordinator.isLinkUser(email: "michael@lickability.net")
            print("Result 1: \(result1)")
            
            let result2 = try await coordinator.isLinkUser(email: "michael@lickability.com")
            print("Result 2: \(result2)")
            
            let result3 = try await coordinator.isLinkUser(email: "michael")
            print("Result 3: \(result3)")
        } catch {
            print("Error: \(error)")
        }
    }
}
```

Here are the raw logs, annotated with comments prefixed by `#`.
```
[Stripe SDK]: POST "/v1/consumers/mobile/sessions/lookup" 200 req_bvC98S4NDmgQ6D
LOG ANALYTICS: link.account_lookup.complete - [(key: "pay_var", value: "paymentsheet"), (key: "ocr_type", value: "none"), (key: "lookupResult", value: "found"), (key: "apple_pay_enabled", value: 1)]

# Expected true (I signed up with this address)
Result 1: true

LOG ANALYTICS: stripeios.attest.assertion.failed - [(key: "error_type", value: "StripeCore.StripeAttest.AttestationError"), (key: "error_code", value: "attestationNotSupported")]
[Stripe SDK]: POST "/v1/consumers/mobile/sessions/lookup" 200 req_Js3HRaWFYbqNak
LOG ANALYTICS: link.account_lookup.complete - [(key: "pay_var", value: "paymentsheet"), (key: "ocr_type", value: "none"), (key: "lookupResult", value: "notFound"), (key: "apple_pay_enabled", value: 1)]

# Expected false (I’ve never signed up)
Result 2: false

LOG ANALYTICS: stripeios.attest.assertion.failed - [(key: "error_type", value: "StripeCore.StripeAttest.AttestationError"), (key: "error_code", value: "attestationNotSupported")]
[Stripe SDK]: POST "/v1/consumers/mobile/sessions/lookup" 400 req_m7G0msGkl0sWPk
LOG ANALYTICS: link.account_lookup.failure - [(key: "request_id", value: "req_m7G0msGkl0sWPk"), (key: "pay_var", value: "paymentsheet"), (key: "ocr_type", value: "none"), (key: "error_type", value: "invalid_request_error"), (key: "error_code", value: "email_invalid"), (key: "apple_pay_enabled", value: 1)]

# Expected error (bad input)
Error: apiError(StripeCore.StripeAPIError(type: StripeCore.StripeAPIError.ErrorType.invalidRequestError, code: Optional("email_invalid"), docUrl: Optional(https://stripe.com/docs/error-codes/email-invalid), message: Optional("Invalid email address."), param: nil, statusCode: Optional(400), requestID: Optional("req_m7G0msGkl0sWPk"), _allResponseFieldsStorage: Optional(4 fields)))
```

## Changelog

N/A

### DIFF
--- a/StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
+++ b/StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
@@ -31,17 +31,23 @@ public protocol CryptoOnrampCoordinatorProtocol {
 public final class CryptoOnrampCoordinator: CryptoOnrampCoordinatorProtocol {
     private let linkController: LinkController
     private let apiClient: STPAPIClient
+    private let appearance: Appearance
 
-    private init(linkController: LinkController, apiClient: STPAPIClient = .shared) {
+    private init(linkController: LinkController, apiClient: STPAPIClient = .shared, appearance: Appearance) {
         self.linkController = linkController
         self.apiClient = apiClient
+        self.appearance = appearance
     }
 
     // MARK: - CryptoOnrampCoordinatorProtocol
 
-    public static func create(apiClient: STPAPIClient, appearance: Appearance) async throws -> CryptoOnrampCoordinator {
+    public static func create(apiClient: STPAPIClient = .shared, appearance: Appearance) async throws -> CryptoOnrampCoordinator {
         let linkController = try await LinkController.create(apiClient: apiClient, mode: .payment)
-        return CryptoOnrampCoordinator(linkController: linkController)
+        return CryptoOnrampCoordinator(
+            linkController: linkController,
+            apiClient: apiClient,
+            appearance: appearance
+        )
     }
 
     public func isLinkUser(email: String) async throws -> Bool {
--- a/StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
+++ b/StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
@@ -1,0 +1,10 @@
+//
+//  CryptoOnrampCoordinator.swift
+//  StripeCryptoOnramp
+//
+//  Created by Michael Liberatore on 7/14/25.
+//
+
+import Foundation
+
+
--- a/StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
+++ b/StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
@@ -26,7 +26,7 @@ public protocol CryptoOnrampCoordinatorProtocol {
     func isLinkUser(email: String) async throws -> Bool
 }
 
-///Coordinates headless Link user authentication and identity verification, leaving most of the UI to the client.
+/// Coordinates headless Link user authentication and identity verification, leaving most of the UI to the client.
 @_spi(CryptoOnrampSDKPreview)
 public final class CryptoOnrampCoordinator: CryptoOnrampCoordinatorProtocol {
     private let linkController: LinkController
--- a/StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
+++ b/StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
@@ -6,5 +6,45 @@
 //
 
 import Foundation
+@_spi(STP) import StripePaymentSheet
 
+/// Protocol describing a type that coordinates headless Link user authentication, identity verification, and payment, leaving most of the associated UI up to the client.
+@_spi(CryptoOnrampSDKPreview)
+public protocol CryptoOnrampCoordinatorProtocol {
+    
+    /// Creates a `CryptoOnrampCoordinator` to facilitate authentication, identity verification, and payment.
+    ///
+    /// - Parameter apiClient: The `STPAPIClient` instance for this coordinator. Defaults to `.shared`.
+    /// - Parameter appearance: Customizable appearance-related configuration for any Stripe-provided UI.
+    /// - Returns: A configured `CryptoOnrampCoordinator`.
+    static func create(apiClient: STPAPIClient, appearance: Appearance) async throws -> Self
+    
+    /// Determines whether the user with the specified email is already a Link user.
+    ///
+    /// - Parameter email: The email address of the user in question.
+    /// - Returns: Whether the email corresponds to an already registered Link user.
+    func isLinkUser(email: String) async throws -> Bool
+}
 
+///Coordinates headless Link user authentication and identity verification, leaving most of the UI to the client.
+@_spi(CryptoOnrampSDKPreview)
+public final class CryptoOnrampCoordinator: CryptoOnrampCoordinatorProtocol {
+    private let linkController: LinkController
+    private let apiClient: STPAPIClient
+    
+    private init(linkController: LinkController, apiClient: STPAPIClient = .shared) {
+        self.linkController = linkController
+        self.apiClient = apiClient
+    }
+    
+    // MARK: - CryptoOnrampCoordinatorProtocol
+    
+    public static func create(apiClient: STPAPIClient, appearance: Appearance) async throws -> CryptoOnrampCoordinator {
+        let linkController = try await LinkController.create(apiClient: apiClient, mode: .payment)
+        return CryptoOnrampCoordinator(linkController: linkController)
+    }
+    
+    public func isLinkUser(email: String) async throws -> Bool {
+        return try await linkController.lookupConsumer(with: email)
+    }
+}
--- a/StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
+++ b/StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
@@ -11,14 +11,14 @@ import Foundation
 /// Protocol describing a type that coordinates headless Link user authentication, identity verification, and payment, leaving most of the associated UI up to the client.
 @_spi(CryptoOnrampSDKPreview)
 public protocol CryptoOnrampCoordinatorProtocol {
-    
+
     /// Creates a `CryptoOnrampCoordinator` to facilitate authentication, identity verification, and payment.
     ///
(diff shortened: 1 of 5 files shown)


StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
```diff
@@ -0,0 +1,50 @@
+//
+//  CryptoOnrampCoordinator.swift
+//  StripeCryptoOnramp
+//
+//  Created by Michael Liberatore on 7/14/25.
+//
+
+import Foundation
+@_spi(STP) import StripePaymentSheet
+
```
mliberatore, line 10: For documentation, I typically write docs for anything `internal` or `public`, unless it corresponds to protocol conformance, which you can see below (since the protocol itself covers the documentation of those signatures). I couldn’t determine a workspace-wide pattern for where/when docs should be present, but I’m happy to follow any conventions here.

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
```diff
@@ -0,0 +1,50 @@
...
+    
+    /// Determines whether the user with the specified email is already a Link user.
+    ///
+    /// - Parameter email: The email address of the user in question.
+    /// - Returns: Whether the email corresponds to an already registered Link user.
+    func isLinkUser(email: String) async throws -> Bool
+}
+
+/// Coordinates headless Link user authentication and identity verification, leaving most of the UI to the client.
+@_spi(CryptoOnrampSDKPreview)
```
mliberatore, line 30: I’ve noticed in some cases in the workspace, these get added to the top-level type and _also_ publicly exposed members. Is that necessary, or is it sufficient to just annotate the containing type?

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
```diff
@@ -0,0 +1,50 @@
...
+    /// - Parameter apiClient: The `STPAPIClient` instance for this coordinator. Defaults to `.shared`.
+    /// - Parameter appearance: Customizable appearance-related configuration for any Stripe-provided UI.
+    /// - Returns: A configured `CryptoOnrampCoordinator`.
+    static func create(apiClient: STPAPIClient, appearance: Appearance) async throws -> Self
+    
+    /// Determines whether the user with the specified email is already a Link user.
+    ///
+    /// - Parameter email: The email address of the user in question.
+    /// - Returns: Whether the email corresponds to an already registered Link user.
+    func isLinkUser(email: String) async throws -> Bool
```
mliberatore, line 26: Some other SDKs in the workspace also expose completion-based and Future/Promise-based wrappers. Is this a convention we'd like to follow for the onramp project?

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
```diff
@@ -0,0 +1,50 @@
...
+    private init(linkController: LinkController, apiClient: STPAPIClient = .shared) {
+        self.linkController = linkController
+        self.apiClient = apiClient
+    }
+
+    // MARK: - CryptoOnrampCoordinatorProtocol
+
+    public static func create(apiClient: STPAPIClient, appearance: Appearance) async throws -> CryptoOnrampCoordinator {
+        let linkController = try await LinkController.create(apiClient: apiClient, mode: .payment)
+        return CryptoOnrampCoordinator(linkController: linkController)
```
mats-stripe: This is what I had in mind - WDYT?

```suggestion
        return CryptoOnrampCoordinator(
                linkController: linkController,
                apiClient: apiClient,
                appearance: appearance
        )
```

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
mats-stripe: I'm happy with only exposing Swift Concurrency based APIs in this SDK 👍

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
mats-stripe: What you described as your typical process works with us!

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
mats-stripe: Just to confirm my understanding; if the top-level type is annotated as `@_spi(CryptoOnrampSDKPreview) public`, and a nested field is annotated as `public`, that field will only be exposed to `@_spi(CryptoOnrampSDKPreview) import`, right?

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
```diff
@@ -0,0 +1,50 @@
...
+    private let apiClient: STPAPIClient
+
+    private init(linkController: LinkController, apiClient: STPAPIClient = .shared) {
+        self.linkController = linkController
+        self.apiClient = apiClient
+    }
+
+    // MARK: - CryptoOnrampCoordinatorProtocol
+
+    public static func create(apiClient: STPAPIClient, appearance: Appearance) async throws -> CryptoOnrampCoordinator {
```
mats-stripe:
```suggestion
    public static func create(apiClient: STPAPIClient = .shared, appearance: Appearance) async throws -> CryptoOnrampCoordinator {
```

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
```diff
@@ -0,0 +1,50 @@
...
+        let linkController = try await LinkController.create(apiClient: apiClient, mode: .payment)
+        return CryptoOnrampCoordinator(linkController: linkController)
+    }
+
+    public func isLinkUser(email: String) async throws -> Bool {
+        return try await linkController.lookupConsumer(with: email)
```
mats-stripe, line 54: 😍

mats-stripe (commented):
Looks great! Just suggested one small change around the `create` method.

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
mliberatore: Sure, happy to make that change. I wasn’t sure yet if we’d need to use the api client outside of the `create` method, but I did add a property for it, so I definitely should be passing that in so we don’t accidentally default it to `.shared` internally :+1:.

For appearance, I had left it out since we weren’t making use of it yet, but that's no problem to add now since we definitely will be using it!

I’ll push this change separately from the GitHub suggestion since I’ll need to add an `appearance` property to store, and modify `init`.

mats-stripe: Awesome - thank you!

>For appearance, I had left it out since we weren’t making use of it yet, but that's no problem to add now since we definitely will be using it!

Heads up that our Dead Code Detection GH Action might complain about this, but I think it's reasonable to add it now.

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
mliberatore: > and a nested field is annotated as `public`, that field will only be exposed to `@_spi(CryptoOnrampSDKPreview) import`, right?

That's my understanding, given that `import StripeCryptoOnramp` without the `@_spi(CryptoOnrampSDKPreview)` fails to compile when making a call to an unannotated `public` member, e.g.:

```swift
let coordinator = try await CryptoOnrampCoordinator.create(apiClient: .shared, appearance: .init())
```

where `create` is unannotated. This will error since the containing type is not exposed:
CleanShot 2025-07-15 at 16 24 50

If there's any other concerns, like the contained symbols somehow being found in built products, I’m happy to err on the side of caution and add the `@_spi(CryptoOnrampSDKPreview)` to all `public` declarations. Please let me know what you think!

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
mliberatore: Good to know!

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
mats-stripe: Thanks for checking! That confirms my understanding. We can stick with `@_spi(...) public` annotating the top-level class, and marking the inner fields as just `public` 👍

mats-stripe (approved):

PR #5159, opened 2025-07-15
Modernize root view in PaymentSheet Example

mats-stripe:
image

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetExampleAppRootView.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetExampleAppRootView.swift
@@ -6,51 +6,18 @@
 import StripePaymentSheet
 import SwiftUI
 
-@available(iOS 14.0, *)
+@available(iOS 15.0, *)
 struct PaymentSheetExampleAppRootView: View {
-
-    private var exampleDestinations: [NavigationDestination] {
-        [
-            .paymentSheet,
-            .paymentSheet_deferred,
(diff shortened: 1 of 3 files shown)


Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetExampleAppRootView.swift
```diff
@@ -91,6 +57,40 @@ struct PaymentSheetExampleAppRootView: View {
...
+                     .walletButtonsView_swiftUI,
+                     .addressCollection_swiftUI:
+                    result[.examples]?.append(destination)
+                case .customerSheet_playground,
+                     .paymentSheet_playground:
+                    result[.testPlaygrounds]?.append(destination)
+                }
+            }
+
+            return result
```
yuki-stripe, line 91: super nit - would it be simpler to do like:
```
return [
  .examples: [
    .paymentSheet,
    ...
  ],
  .testPlaygrounds: [
    ...
  ]
]
```
rather than construct it programatically

mats-stripe: I wanted to enforce the compiler requires new `NavigationDestination`s get added here. Your approach is cleaner, but doesn't get the compiler enforcement 😕 WDYT?

yuki-stripe: ah true, keep it as-is then.

PR #5168, opened 2025-07-16
Add billingAddress to AddressElement

porter-stripe:
## Summary
To gain parity with web we want to allow merchants to pass in a billingAddress into AddressElement alongside a shipping address. When a billing address is supplied by the merchant we will pre-fill it if no shipping address is also provided. We will also show a checkbox "Use billing address for shipping" that users can toggle to fill in the address section with their billing address.

## Motivation
- https://docs.google.com/document/d/1nLMw-zOGqhqyvjkAPdYQDRVPvCK17db6_s4ze7wHfaw/edit?tab=t.0

## Testing
- Updated test
- Manual

## Changelog
N/A


davidme-stripe (approved):

davidme-stripe:
One PQ nit on the behavior: It might be nicer if checking and unchecking "use billing address" didn't lose the user's last input.

PR #5169, opened 2025-07-16
Add ability to customize embedded payment element row button title font

yuki-stripe:
## Motivation
https://docs.google.com/document/d/1x8TbzCSFFZf_bZFoQoyAPRbgtV2ydU1e4veY-PS_Wd8/edit?tab=t.0

## Testing
see snapshot test

## Changelog
See changelog


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 ### PaymentSheet
 * [Fixed] EmbeddedPaymentElement now errors when `configuration.rowSelectionBehavior` is `.immediateAction` and `configuration.embeddedViewDisplaysMandateText` is `true` to ensure integrations display legal mandate text to the customer.
 * [Added] Added `Appearance.EmbeddedPaymentElement.Row.paymentMethodIconLayoutMargins` to customize the spacing around payment method icons in EmbeddedPaymentElement payment method rows.
+* [Added] Added `Appearance.EmbeddedPaymentElement.Row.titleFont` to customize the font of EmbeddedPaymentElement payment method row titles.
```
porter-stripe, line 6: Can we get these (paymentMethodIconLayoutMargins too) added to the AppearancePlaygroundView?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButton.swift
```diff
@@ -383,7 +384,11 @@ extension RowButton {
 
     static func makeRowButtonLabel(text: String, appearance: PaymentSheet.Appearance, isEmbedded: Bool) -> UILabel {
         let label = UILabel()
-        label.font = appearance.scaledFont(for: appearance.font.base.medium, style: .subheadline, maximumPointSize: 25)
+        if isEmbedded, let customFont = appearance.embeddedPaymentElement.row.titleFont {
+            label.font = customFont
+        } else {
+            label.font = appearance.scaledFont(for: appearance.font.base.medium, style: .subheadline, maximumPointSize: 25)
```
porter-stripe, line 390: Feels a bit off to me that merchants can't change the subtitle text or the action text.

Image

porter-stripe: It feels like this API should impact those labels, and we could add more granular control for those labels in the future. The current implementation would leave merchants in a state where they are forced to have mismatch fonts right next to each other unless they also wanted to change other fonts via Appearance (which seems like they don't want to if they want this API?)

yuki-stripe: Hmm I think it's unlikely the merchant uses different fonts the way you're describing. The only known use case for this API so far is to customize the *weight* of the title.

porter-stripe (approved):

porter-stripe (approved):

PR #5171, opened 2025-07-17
Flat with chevron -> Flat with disclosure, added disclosure view

yuki-stripe:
## Summary
* Renamed `PaymentSheet.Appearance.EmbeddedPaymentElement.Row.Style.flatWithChevron` to `PaymentSheet.Appearance.EmbeddedPaymentElement.Row.Style.flatWithDisclosure`
* Added `disclosureView` to customize the chevron icon

## Motivation
https://docs.google.com/document/d/1x8TbzCSFFZf_bZFoQoyAPRbgtV2ydU1e4veY-PS_Wd8/edit?tab=t.0

## Testing
See snapshot test
Manually tested playground, changed chevron color appearance

## Changelog
See changelog


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButtonFlatWithDisclosure.swift
```diff
@@ -10,15 +10,19 @@ import Foundation
...
+final class RowButtonFlatWithDisclosure: RowButton {
     // MARK: - Subviews
-    private lazy var chevronView: UIImageView = {
-        let chevronImageView = UIImageView(image: Image.icon_chevron_right.makeImage(template: true))
-
-        chevronImageView.tintColor = appearance.embeddedPaymentElement.row.flat.chevron.color
-        chevronImageView.contentMode = .scaleAspectFit
-        chevronImageView.translatesAutoresizingMaskIntoConstraints = false
-        return chevronImageView
+    private lazy var chevronView: UIView = {
```
gbirch-stripe: nit: rename this property to `disclosureView`?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButtonFlatWithDisclosure.swift
```diff
@@ -10,15 +10,19 @@ import Foundation
...
+        if let customView = appearance.embeddedPaymentElement.row.flat.chevron.disclosureView?() {
```
gbirch-stripe: I have some concerns about just dropping the custom view into our layout and constraint system. When I did some testing with this I found that the `horizontalStackView` that the `chevronView` is in is constrained to not be the full height of the row and thus restricts the `chevronView` from getting as large as it otherwise could. This could be an issue for the specific use case we're making these changes for. I had to slightly rearrange the layout to get this to work.

gbirch-stripe: Relatedly, is there any validation we want to add to this view? For example we could formalize assumptions we are making about it for layout purposes into validations.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/EmbeddedPaymentMethodsViewSnapshotTests.swift
```diff
@@ -878,9 +878,31 @@ class EmbeddedPaymentMethodsViewSnapshotTests: STPSnapshotTestCase {
         verify(embeddedView)
     }
 
-    func testEmbeddedPaymentMethodsView_flatWithChevron_savedPaymentMethod() {
+    func testEmbeddedPaymentMethodsView_flatWithDisclosure_customDisclosureView() {
```
gbirch-stripe, line 881: Would it be beneficial to add more snapshot tests given the open ended nature of this new config? For example ensure that it works with a fixed size view and/or image asset, a scalable vector image, etc.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButtonFlatWithDisclosure.swift
yuki-stripe: That's fair, I'll try restricting the max height and width of the view to reasonable limits?

I have 3 tests now
- Normal icon replacement
- Normal 'select button'
- Way too big select button

I also added the saved PM row to these snapshot tests, that revealed a bug as well.

yuki-stripe: LMK if there are other concerns. I don't think we need to be super guard-raily here though, like merchants can't possibly want to put _arbitrary_ size views here, there is surely a small set of custom views we'll ever see.

PR #5175, opened 2025-07-17
Minor tweaks for AE

porter-stripe:


davidme-stripe:
Can we add this to the CHANGELOG?

porter-stripe:
@davidme-stripe yep, done!

PR #5189, opened 2025-07-22
Update LinkController with new APIs and published LinkAccount

mats-stripe:
## Summary

Makes the following changes to `LinkController`:

1. Adds `registerNewLinkUser` API. Headlessly registers a new Link account.
2. Adds `presentForVerification` API. Presents the Link UI to verify an existing Link user.
3. Renames `present` to `collectPaymentMethod`.
4. Adds an `@Published` `linkAccount` property, and marks certain fields on the `PaymentSheetLinkAccount` object as internally public.

## Motivation

https://docs.google.com/document/d/1owKwHgD5s8u75EWDOny2bp6rlNZq1aI4NNUr43PKdVA/edit?usp=sharing

## Testing

Follow-up PR will include a new demo app to play around with the LinkController.

## Changelog

N/a

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkController.swift
@@ -28,10 +28,10 @@ import UIKit
         @_spi(STP) public let sublabel: String?
     }
 
-    @frozen @_spi(STP) public enum AuthenticationResult {
-        /// Authentication was completed successfully.
+    @frozen @_spi(STP) public enum VerificationResult {
+        /// Verification was completed successfully.
         case completed
-        /// Authentication was canceled by the user.
+        /// Verification was canceled by the user.
         case canceled
     }
 
@@ -211,11 +211,11 @@ import UIKit
     /// Presents the Link verification flow for an existing user which requires verification.
     /// `lookupConsumer` must be called before this.
     ///
-    /// - Parameter viewController: The view controller from which to present the authentication flow.
-    /// - Parameter completion: A closure that is called with the result of the authentication. It returns an `AuthenticationResult` if successful, or an error if the verification failed.
+    /// - Parameter viewController: The view controller from which to present the verification flow.
+    /// - Parameter completion: A closure that is called with the result of the verification. It returns a `VerificationResult` if successful, or an error if the verification failed.
     @_spi(STP) public func presentForVerification(
         from viewController: UIViewController,
-        completion: @escaping (Result<AuthenticationResult, Error>) -> Void
+        completion: @escaping (Result<VerificationResult, Error>) -> Void
     ) {
         guard let linkAccount, linkAccount.sessionState == .requiresVerification else {
             let error = IntegrationError.noActiveLinkConsumer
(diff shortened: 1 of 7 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkController.swift
```diff
@@ -153,12 +176,75 @@ import UIKit
...
+        guard let linkAccount else {
+            let error = IntegrationError.noActiveLinkConsumer
+            completion(.failure(error))
+            return
+        }
+        linkAccount.signUp(
+            with: phone,
+            legalName: fullName,
+            countryCode: country,
+            consentAction: .clicked_button_mobile_v1,
```
mats-stripe: I'm not sure which consent action makes the most sense to use here. This is for a scenario where a host app will be collecting the required fields (name, phone number) and registering new Link accounts (without any Stripe-owned UI be presented). Maybe a new consent action is needed? @wooj-stripe do you have any opinions?

mats-stripe: Here are all the existing options: https://stripe.sourcegraphcloud.com/stripe-internal/pay-server/-/blob/lib/consumer/api/consent_action_types.rb?L14-50

mats-stripe: Actually, I'm thinking that this `registerLinkUser` API should take in the consent action as a parameter. I can add a case for `entered_phone_number_email_clicked_signup_crypto_onramp` in out client side enum, which the Onramp SDK can pass here.

mats-stripe: Done in https://github.com/stripe/stripe-ios/pull/5189/commits/712b921b66723cb049754cb27bb4b7788e11745a

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkController.swift
```diff
@@ -28,6 +28,13 @@ import UIKit
         @_spi(STP) public let sublabel: String?
     }
 
+    @frozen @_spi(STP) public enum AuthenticationResult {
```
tillh-stripe: Should this be `VerificationResult`, since the method is called `presentForVerification`? Maybe it should be called `presentForAuthentication`?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkController.swift
```diff
@@ -153,12 +176,77 @@ import UIKit
         }
     }
 
+    /// Registers a new Link user with the provided details.
+    /// `lookupConsumer` must be called before this.
```
tillh-stripe, line 180: Can we avoid that this is necessary?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkController.swift
```diff
@@ -153,12 +176,77 @@ import UIKit
...
+            consentAction: consentAction,
+            completion: { [weak self] result in
+                LinkAccountContext.shared.account = self?.linkAccount
+                completion(result)
+            }
+        )
+    }
+
+    /// Presents the Link verification flow for an existing user which requires verification.
+    /// `lookupConsumer` must be called before this.
```
tillh-stripe, line 212: Same here!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkController.swift
mats-stripe: Sure, I can update the enum name 👍

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkController.swift
mats-stripe: How come? I think we should continue to enforce this as it ensures that the API is used in the intended sequence.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkController.swift
mats-stripe: Done in https://github.com/stripe/stripe-ios/pull/5189/commits/765bc49353443ee80444288a1d08d7a28db2a339

PR #4591, opened 2025-02-20
[Connect] Implement api changes for onboarding.

nschris-stripe:
## Summary
Changes include:
- Removing payouts from beta
- Migrating onboarding to workflows api.
- Updating example app to use new API.

https://docs.google.com/document/d/15cv9AFDCRU6Sxmi3IKAaWJjQKdMUH2HNIlL_fOxEX0c/edit?tab=t.0#heading=h.570u4leviueu

## Demo

https://github.com/user-attachments/assets/20bdcb98-1dbe-48b1-aaac-142f54f21785

PR #5191, opened 2025-07-22
Add payments + payouts experimentally to Connect Example App

torrance-stripe:
## Summary
- Adds back in view controller options to allow the payments + payouts component to be presented in a variety of ways (removed in https://github.com/stripe/stripe-ios/pull/4591/files) 
- Adds payments and payouts to the example app
- Fixes export functionality after it was unintentionally broke in https://github.com/stripe/stripe-ios/pull/4346 

https://github.com/user-attachments/assets/e4e6197d-d5b2-4644-9581-b6809686b20f

## Motivation
Manual testing purposes, not ready to be publicly integrated with yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
 * [Added] Added `Appearance.EmbeddedPaymentElement.Row.titleFont` to customize the font of EmbeddedPaymentElement payment method row titles.
 * [Changed] Renamed `PaymentSheet.Appearance.EmbeddedPaymentElement.Row.Style.flatWithChevron` to `PaymentSheet.Appearance.EmbeddedPaymentElement.Row.Style.flatWithDisclosure` and added an experimental `disclosureImage` property to customize the chevron icon displayed in EmbeddedPaymentElement payment method rows when the row style is flatWithDisclosure.
 * [Fixed] `EmbeddedPaymentElementDelegate` `didUpdateHeight` now accommodates cases where the embedded view can't change height until after `didUpdateHeight` is called, like inside a UITableViewCell.
+* [Fixed] Apple Pay now properly collects phone number and email address when marked as required in `PaymentSheet.Configuration.billingDetailsCollectionConfiguration`.
 
 ### AddressElement
 * [Added] SwiftUI support for AddressElement.
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -7,3 +7,4 @@
 */StripePaymentsUI @stripe/ios-payments-reviewers @stripe/ios-sdk-reviewers @stripe/ios-sdk-non-core-reviewers
 */StripeApplePay @stripe/ios-payments-reviewers @stripe/ios-sdk-reviewers @stripe/ios-sdk-non-core-reviewers
 */Stripe3DS2 @stripe/ios-payments-reviewers @stripe/ios-sdk-reviewers @stripe/ios-sdk-non-core-reviewers
+*/StripeConnect @stripe/ios-connect-reviewers @stripe/ios-sdk-reviewers @stripe/ios-sdk-non-core-reviewers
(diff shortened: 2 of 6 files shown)


Example/StripeConnectExample/StripeConnectExample/MainViewController.swift
```diff
@@ -123,6 +136,62 @@ class MainViewController: UITableViewController {
             onboardingConfig.delegate = self
             onboardingConfig.title = row.label
             onboardingConfig.present(from: self, animated: true)
+            return
```
guido-stripe, line 139: Instead of returning, would it make sense to create a generic code for handling all the three cases in the same way? Assuming that we don't want to force Onboarding in a specific way and only let the other work differently

Example/StripeConnectExample/StripeConnectExample/MainViewController.swift
```diff
@@ -214,3 +283,53 @@ extension MainViewController: AccountOnboardingControllerDelegate {
...
+extension MainViewController: PaymentsViewControllerDelegate {
+    func payments(_ payments: PaymentsViewController, didFailLoadWithError error: any Error) {
+        presentAlert(title: "Error loading payments", message: (error as NSError).debugDescription)
+    }
+}
+
+// MARK: - UINavigationControllerDelegate
+
+extension MainViewController: UINavigationControllerDelegate {
+    func navigationController(_ navigationController: UINavigationController, willShow viewController: UIViewController, animated: Bool) {
```
guido-stripe, line 306: What happens if this code is put in directly in the code above calling the `push` (similarly to the modal presentation)? I assume it won't work as expected, but just asking for making sure this is the way to have it, mainly because I find the code trickier to read (using a protocol/delegate), rather than doing things in the same code chunk

StripeConnect/StripeConnect/Source/Internal/Webview/ConnectWebViewController.swift
```diff
@@ -218,6 +218,18 @@ private extension ConnectWebViewController {
         // Remove reference so we can log if we inadvertently get simultaneous downloads
         self.downloadedFile = nil
     }
+
+    /// Checks if the given URL is a data export URL from Stripe's S3 bucket
+    func isDataExportURL(_ url: URL) -> Bool {
+        guard let host = url.host else { return false }
+        return host.contains("stripe-data-exports") ||
+               (host.contains("s3.amazonaws.com") && url.path.contains("stripe-data-exports"))
```
guido-stripe, line 226: I just wonder if this could be dangerous? Malicious actors can definitely have a `s3.amazonaws.com` domain and a path containing `stripe-data-exports`, although I'm not sure what kind of harm this could cause to the users.

StripeConnect/StripeConnect/Source/Internal/Webview/ConnectWebViewController.swift
```diff
@@ -218,6 +218,18 @@ private extension ConnectWebViewController {
...
+    }
+
+    /// Checks if the given host is an S3 domain
+    func isS3Host(_ host: String) -> Bool {
+        return host.contains("s3.amazonaws.com") || host.contains("stripe-data-exports")
```
guido-stripe, line 231: Same concerns as above

guido-stripe (commented):
@torrance-stripe overall changes look good, thanks! I left a few questions, although the only ones preventing immediate approval are the ones about security when checking the host/link. I'm pretty sure it should be safe but I prefer to have a confirmation. Thanks!

Example/StripeConnectExample/StripeConnectExample/MainViewController.swift
torrance-stripe: No, that's the intended functionality. You should read through the [API Review](https://docs.google.com/document/d/15cv9AFDCRU6Sxmi3IKAaWJjQKdMUH2HNIlL_fOxEX0c/edit?tab=t.0#heading=h.570u4leviueu) linked in https://github.com/stripe/stripe-ios/pull/4591 for more context!

Example/StripeConnectExample/StripeConnectExample/MainViewController.swift
torrance-stripe: I'm not sure how to do this, I'm just reverting the changes as they were removed in #4591

StripeConnect/StripeConnect/Source/Internal/Webview/ConnectWebViewController.swift
torrance-stripe: We have a CSP on the embedded components which will protect users on the web layer (which I won't link publicly 😄). I see this logic as more to determine when to show the native download card!

Example/StripeConnectExample/StripeConnectExample/MainViewController.swift
guido-stripe: OK, sounds good for now as this was mainly a nit!

Example/StripeConnectExample/StripeConnectExample/MainViewController.swift
guido-stripe: I see, thanks!

For reference:
> Previously the API provided a ViewController directly and it was on the integrator to handle presentation, dismissal and embedding it in a navigation controller. Now an AccountOnboardingController is returned which manages the presentation/dismissal. Initially we will only allow modal presentation.

guido-stripe (approved):
Let's go!

PR #5192, opened 2025-07-22
Xcode 15 deprecation cleanup

gbirch-stripe:
## Summary
Cleaning up some unneeded Xcode 15 support code

## Motivation
The joy of deleting code

## Testing
Builds

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
gbirch-stripe, line 1: I tinkered around with making the change in the TODO, but I don't think it makes sense at this point. There are custom implementations of the `STPAuthenticationContext` and `PaymentSheetAuthenticationContext` functions that coexist, and forcing a universal `UIViewController` conformance onto them would require some reworking if its even possible

PR #5193, opened 2025-07-22
Remove chevron from payment method change button for disclosure style

gbirch-stripe:
## Summary
Like the checkmark style, the disclosure row style should not include a chevron in the accessory (change/edit/view more) button.

## Motivation
Fix incorrect UI

## Testing
Updated screenshot tests, which already well cover this scenario

## Changelog
N/A


Example/PaymentSheet Example/PaymentSheet Example/EmbeddedPlaygroundViewController.swift
gbirch-stripe, line 1: Unrelated change. Fixing because it's been driving me nuts while debugging

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RightAccessoryButton.swift
```diff
@@ -166,20 +166,20 @@ extension RowButton.RightAccessoryButton {
...
                                        allowsPaymentMethodRemoval: Bool,
                                        allowsPaymentMethodUpdate: Bool,
-                                       isFlatCheckmarkOrDisclosureStyle: Bool = false) -> AccessoryType? {
+                                       omitChevron: Bool = false) -> AccessoryType? {
         guard savedPaymentMethodsCount > 0 else { return nil }
 
         // If we have more than 1 saved payment method always show the "View more" button
         if savedPaymentMethodsCount > 1 {
-            return isFlatCheckmarkOrDisclosureStyle ? .viewMore : .viewMoreChevron
+            return omitChevron ? .viewMore : .viewMoreChevron
```
porter-stripe, line 182: Maybe instead of passing in a bool, we should just do something like:

```swift
extension PaymentSheet.Appearance.EmbeddedPaymentElement.Row.Style {
    var viewMoreType: AccessoryType {
        switch self {
        case .flatWithCheckmark, .flatWithDisclosure:
            return .viewMore
        case .flatWithRadio, .floatingButton:
            return .viewMoreChevron
        }
    }
}
```

gbirch-stripe: This is also used in `EmbeddedPaymentMethodsView.swift:496`, where the options are `.change` or `.changeWithChevron`. I'm open to this but it means duplicating the logic for the two different situations

porter-stripe (approved):

PR #5201, opened 2025-07-23
Crypto Onramp SDK: Consumes Internal Customers API

mliberatore:
## Summary
Adds and tests a new API, `/v1/crypto/internal/customers`. It will be made use of in a future PR that adds more authentication APIs to `CryptoOnrampCoordinator`.

## Motivation
Required for the [Crypto Onramp SDK](https://docs.google.com/document/d/1owKwHgD5s8u75EWDOny2bp6rlNZq1aI4NNUr43PKdVA/edit?tab=t.0#heading=h.5mbaom37vv77).

## Testing
1. Ran the newly added tests.
2. Also added a temporary test method to `CryptoOnrampCoordinator` to exercise the functionality and ensure the request actually succeeded.

```swift
public func testGrantPartnerMerchantPermissions(presentingViewController: UIViewController) async throws {
    _ = try await lookupConsumer(with: "michael@lickability.net")
    _ = try await linkController.presentForVerification(from: presentingViewController)
    guard let linkAccount = await linkController.linkAccount else { return }
    let apiResult = try await apiClient.grantPartnerMerchantPermissions(with: linkAccount)
    print("API result: \(apiResult)")
}
```

```log
[Stripe SDK]: POST "/v1/crypto/internal/customers" 200 req_C19LrF4okYjiCi
API result: CustomerResponse(id: "crc_1RnjGqLu5o3P18Zp6ihdypaQ")
```

## Changelog
N/A

### DIFF
--- a/StripeCryptoOnramp/StripeCryptoOnramp/Source/Internal/API Bindings/CustomerRequest.swift
+++ b/StripeCryptoOnramp/StripeCryptoOnramp/Source/Internal/API Bindings/CustomerRequest.swift
@@ -7,8 +7,8 @@
 
 import Foundation
 
-/// Codable model passed to the `/v1/crypto/internal/customers` endpoint.
-struct CustomerRequest: Codable {
+/// Encodable model passed to the `/v1/crypto/internal/customers` endpoint.
+struct CustomerRequest: Encodable {
 
     /// Container for credentials required to make the request
     struct Credentials: Codable {


StripeCryptoOnramp/StripeCryptoOnrampTests/Unit/StripeCryptoOnrampTests.swift
```diff
@@ -1,15 +0,0 @@
-//
```
mliberatore, line 1: Removed as this was just a placeholder file until we started adding real tests.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/PayWithLinkButton.swift
```diff
@@ -467,7 +469,9 @@ private func makeAccountStub(email: String, isRegistered: Bool, lastPM: LinkPMDi
     return PayWithLinkButton.LinkAccountStub(
         email: email,
         redactedPhoneNumber: nil,
-        isRegistered: isRegistered
+        isRegistered: isRegistered,
+        sessionState: .verified,
+        consumerSessionClientSecret: nil
```
mliberatore, line 474: The changes in this file are simply to satisfy the new protocol requirements. This method creates `LinkAccountStub`s that are used in the SwiftUI previews below, for context, and the newly exposed properties are not used.

StripePaymentSheet/StripePaymentSheet/Source/Helpers/PaymentSheetLinkAccount.swift
```diff
@@ -11,10 +11,12 @@
...
-protocol PaymentSheetLinkAccountInfoProtocol {
-    var email: String { get }
-    var redactedPhoneNumber: String? { get }
-    var isRegistered: Bool { get }
+@_spi(STP) public protocol PaymentSheetLinkAccountInfoProtocol {
+    @_spi(STP) var email: String { get }
+    @_spi(STP) var redactedPhoneNumber: String? { get }
+    @_spi(STP) var isRegistered: Bool { get }
+    @_spi(STP) var sessionState: PaymentSheetLinkAccount.SessionState { get }
+    @_spi(STP) var consumerSessionClientSecret: String? { get }
```
mliberatore, line 19: Discussed in Slack, but this change was intended to allow the new API method to `throw` when the account is not in a verified state. The final result in this PR throws additionally when the `consumerSessionClientSecret` is unset.

The protocol itself is exposed so that it's trivial to replace the implementation within unit tests.

StripeCryptoOnramp/StripeCryptoOnramp/Source/Internal/API Bindings/CustomerRequest.swift
```diff
@@ -0,0 +1,34 @@
...
+//  CustomerRequest.swift
+//  StripeCryptoOnramp
+//
+//  Created by Michael Liberatore on 7/17/25.
+//
+
+import Foundation
+
+/// Codable model passed to the `/v1/crypto/internal/customers` endpoint.
+struct CustomerRequest: Codable {
```
mats-stripe: This is much nicer than what I usually do, which is using a `[String: Any]` blob in the API request 😅 

This is very very minor, but could the request object just be `Encodable` rather than `Codable`? Same with the response object being `Decodable` rather than `Codable`. Unless the `STPAPIClient` requires `Codable` objects - not sure!

StripePaymentSheet/StripePaymentSheet/Source/Helpers/PaymentSheetLinkAccount.swift
mats-stripe: Should these all be marked as `@_spi(STP) public` rather than just `@_spi(STP)`?

mats-stripe (commented):
Looks great - awesome test coverage! I left one very minor suggestion, and one possible code change might be needed

StripePaymentSheet/StripePaymentSheet/Source/Helpers/PaymentSheetLinkAccount.swift
mliberatore: They’ll match their containing declaration’s access level, so these properties are exposed as `public` by default. The compiler will enforce this, too.
Xcode 2025-07-24 09 06 39

mats-stripe: Ahh gotcha 👍

mats-stripe (approved):

StripeCryptoOnramp/StripeCryptoOnramp/Source/Internal/API Bindings/CustomerRequest.swift
mliberatore: Sure! I’ve kept `CustomerResponse` `Codable`, however, since we use encoding in the tests for convenience in mocking the response. Declaring `Encodable` conformance in an outside extension (i.e. in tests in our case) requires implementing `CodingKeys` and `encode(to:)` manually, so I figured the code savings there were worth the additional conformance, since the type is not `public`ly exposed in the end. Let me know if you’re okay with that.

mats-stripe: Yeah that's totally fine! Thanks for making the change

PR #5203, opened 2025-07-23
Crypto Onramp SDK: Additional Auth APIs

mliberatore:
## Summary
Adds two additional auth-related APIs as documented [here](https://docs.google.com/document/d/1fiRBeegAqMbaR7OUDUy0EA4idvHEDh_FcBE2DTz2vyo/edit?tab=t.0).

## Motivation
Required for the [Crypto Onramp SDK](https://docs.google.com/document/d/1owKwHgD5s8u75EWDOny2bp6rlNZq1aI4NNUr43PKdVA/edit?tab=t.0#heading=h.5mbaom37vv77).

## Testing
I don’t have an example app ready yet, so I wired up a few calls to an existing example UI and called the following from button actions.

```swift
do {
    let coordinator = try await CryptoOnrampCoordinator.create(apiClient: .shared, appearance: .init())
    try await coordinator.lookupConsumer(with: "michael@lickability.net")
    let authenticationResult = try await coordinator.presentForVerification(from: findViewController()!)
    print("Auth Result: \(authenticationResult)")
} catch {
    print("Error: \(error)")
}
```

**Success:**
```
[Stripe SDK]: POST "/v1/crypto/internal/customers" 200 req_exhqJjzpggKwz9
Auth Result: completed(customerId: "crc_1RnjGqLu5o3P18Zp6ihdypaQ")
```
**Canceled:**
```
Auth Result: canceled
```
**Re-ran with unregistered email:**
```
Error: noActiveLinkConsumer
```

```swift
do {
    let coordinator = try await CryptoOnrampCoordinator.create(apiClient: .shared, appearance: .init())
    try await coordinator.lookupConsumer(with: "mike@lickability.net")
    let id = try await coordinator.registerLinkUser(fullName: "Michael Liberatore", phone: "+13305555555", country: "US")
    print("Customer ID: \(id)")
} catch {
    print("Error: \(error)")
}
```

**Success**
```
[Stripe SDK]: POST "/v1/consumers/mobile/sign_up" 200 req_73mtNpnhfJl6WH
[Stripe SDK]: POST "/v1/crypto/internal/customers" 200 req_uO8TmAGvbadUvZ
Customer ID: crc_1RoA2DLu5o3P18ZpZte0on9V
```

**Re-ran with now-registered email:**
```
Error: An error occurred in PaymentSheet. Don't call sign up if not needed
```

## Changelog
N/A


StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/AuthenticationResult.swift
```diff
@@ -0,0 +1,17 @@
+//
+//  AuthenticationResult.swift
```
mliberatore: Note: I saw that the underlying `LinkController` APIs have switched to `VerificationResult` as a result of code review suggestions. I’m happy to make the same change here, but don’t want to stray too far from the APIs declared in the mobile API review document. Please let me know if I should update this and the associated docs throughout this PR to match!

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
```diff
@@ -33,6 +53,15 @@ public final class CryptoOnrampCoordinator: CryptoOnrampCoordinatorProtocol {
     private let apiClient: STPAPIClient
     private let appearance: Appearance
 
+    private var linkAccountInfo: PaymentSheetLinkAccountInfoProtocol {
+        get async throws {
```
mats-stripe, line 65: TIL `get async throws`

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
```diff
@@ -53,4 +82,25 @@ public final class CryptoOnrampCoordinator: CryptoOnrampCoordinatorProtocol {
     public func lookupConsumer(with email: String) async throws -> Bool {
         return try await linkController.lookupConsumer(with: email)
     }
+
+    public func registerLinkUser(fullName: String?, phone: String, country: String) async throws -> String {
+        try await linkController.registerLinkUser(
+            fullName: fullName,
+            phone: phone,
+            country: country,
+            consentAction: .entered_phone_number_email_clicked_signup_crypto_onramp
```
mats-stripe: Nice!

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
```diff
@@ -24,6 +25,25 @@ public protocol CryptoOnrampCoordinatorProtocol {
     /// - Parameter email: The email address to look up.
     /// - Returns: Returns `true` if the email is associated with an existing Link consumer, or `false` otherwise.
     func lookupConsumer(with email: String) async throws -> Bool
+
+    /// Registers a new Link user with the provided details.
+    /// `lookupConsumer` must be called before this.
+    ///
+    /// - Parameter fullName: The full name of the user.
+    /// - Parameter phone: The phone number of the user. Expected to be in E.164 format.
```
mats-stripe: Non-blocking: I now wonder if we should be more explicit about this phone number format, since API requests will fail otherwise. Couple ideas here, but I don't have strong opinions. I'll leave it up to you!

1. Tweak this comment a bit:

```suggestion
    /// - Parameter phone: The phone number of the user. Phone number must be in E.164 format (e.g., +12125551234), otherwise an error will be thrown.
```

2. Actually verify the format in our implementation before performing the API request, along the lines of:

```swift
guard isValidE164Format(phone) else {
    throw ValidationError.invalidPhoneFormat("Phone number must be in E.164 format (e.g., +12125551234)")
}
```

3. Detect API request errors specific to misformated phone numbers, and throw an `invalidPhoneFormat` error. The sign up API will return an error like this in that scenario:

```json
{
	"error": {
		"message": "There was an issue parsing the phone number ***********.",
		"type": "invalid_request_error"
	}
}
```

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/AuthenticationResult.swift
mats-stripe: Good catch! Sorry for the churn here, but I think it would make sense to rename this to `VerificationResult`. Updated the docs accordingly!

mliberatore: No worries at all, just wasn’t sure how finalized the public API was!

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
mliberatore: I went with 1 & 3 here. I left 2 out as I’d like to prevent a situation where the client would potentially not let something through that the server would parse just fine, for example if the format restrictions ever changed or loosened, and the two implementations diverged.

I ensured that the new error would be thrown in the case of an invalid phone number:

```swift
[Stripe SDK]: POST "/v1/consumers/mobile/sign_up" 400 req_Z2cZ7iF8Tc9bVB
Error: invalidPhoneFormat
```

and that other API errors were still properly thrown:

```swift
Error: An error occurred in PaymentSheet. Don't call sign up if not needed
```

mats-stripe: Amazing - thank you!

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
```diff
@@ -53,4 +90,37 @@ public final class CryptoOnrampCoordinator: CryptoOnrampCoordinatorProtocol {
...
+                phone: phone,
+                country: country,
+                consentAction: .entered_phone_number_email_clicked_signup_crypto_onramp
+            )
+        } catch {
+            if let stripeError = (error as? StripeError),
+               case let .apiError(stripeAPIError) = stripeError,
+               stripeAPIError.type == .invalidRequestError,
+               let message = stripeAPIError.message,
+               message.hasPrefix("There was an issue parsing the phone number") {
```
mats-stripe, line 107: I'm worried our error messages might be localized 😅 Let me look into this

mliberatore: Sounds good. If that ends up being the case, we can remove this logic and just stick with the documentation change, since the API error will still be thrown and contain the relevant error information that the SDK user could diagnose.

mats-stripe: Let's leave it as-is for now

PR #5208, opened 2025-07-24
Use the refresh endpoint for Klarna

davidme-stripe:
## Summary
Klarna now supports the refresh endpoint, start using it for Klarna PMs.

## Motivation
Improved Klarna UX

## Testing
Updated tests, existing CI, manually in PS Example

## Changelog
Not public yet


porter-stripe:
What's the benefit of using the refresh endpoint for Klarna? We don't do any polling so curious what the benefits are.

davidme-stripe:
@porter-stripe They want to get rid of the redirect bouncer, and the refresh endpoint will guarantee that they synchronously fetch the PI state from Klarna before returning it to us. (Currently they rely on the redirect bouncer to tell them to fetch the state.)

yuki-stripe (approved):

yuki-stripe (approved):

PR #5210, opened 2025-07-24
[Draft] Fixing copy around specific document types in document capture 

Pearapps:

### DIFF
--- a/StripeIdentity/StripeIdentity/Resources/Localizations/en.lproj/Localizable.strings
+++ b/StripeIdentity/StripeIdentity/Resources/Localizations/en.lproj/Localizable.strings
@@ -19,6 +19,9 @@
 /* Title of ID document scanning screen when scanning the back of an identity card */
 "Back of identity document" = "Back of identity document";
 
+/* Title of ID document scanning screen when scanning the back of either a driver's license, passport, or government issued photo id */
+"Back of %@" = "Back of %@";
+
 /* Error title displayed to the user when camera permissions have been denied
    Title displayed when requesting camera permissions */
 "Camera permission" = "Camera permission";
@@ -63,6 +66,9 @@
 
 /* Title of ID document scanning screen when scanning the front of an identity card */
 "Front of identity document" = "Front of identity document";
+
+/* Title of ID document scanning screen when scanning the front of either a driver's license, passport, or government issued photo id */
+"Front of %@" = "Front of %@";
 
 /* Title for document front warmup page */
 "Get ready to scan your photo ID" = "Get ready to scan your photo ID";
--- a/StripeIdentity/StripeIdentity/Source/API Bindings/Models/VerificationPage/VerificationPageStaticContentDocumentSelectPage.swift
+++ b/StripeIdentity/StripeIdentity/Source/API Bindings/Models/VerificationPage/VerificationPageStaticContentDocumentSelectPage.swift
@@ -15,6 +15,10 @@ extension StripeAPI {
         let buttonText: String
         let idDocumentTypeAllowlist: [String: String]
         let title: String
+        
+        var idDocumentTypeAllowlistKeys: [String] {
+            return Array(idDocumentTypeAllowlist.keys)
+        }
     }
 
 }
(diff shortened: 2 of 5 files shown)


StripeIdentity/StripeIdentity/Resources/Localizations/en.lproj/Localizable.strings
```diff
@@ -10,11 +10,14 @@
 /* Opens the app's settings in the Settings app */
 "App Settings" = "App Settings";
 
-/* Accessibility label when back identity card photo has successfully uploaded */
-"Back identity card photo successfully uploaded" = "Back identity card photo successfully uploaded";
+/* Accessibility label when back driver's license, passport, or government issued photo id photo has successfully uploaded */
+"Back %@ photo successfully uploaded" = "Back %@ photo successfully uploaded";
```
cjmisenas-stripe, line 14: How is this string different from the string on line 17? Which string should be used if for example, only `driving_license` and `passport` are [`allowed_types`](https://docs.stripe.com/api/identity/verification_sessions/create#create_identity_verification_session-options-document-allowed_types)

StripeIdentity/StripeIdentity/Resources/Localizations/en.lproj/Localizable.strings
```diff
@@ -10,11 +10,14 @@
...
 
-/* Description of back of identity card image */
-"Back of identity card" = "Back of identity card";
+/* Accessibility label when back identity document photo has successfully uploaded */
+"Back identity document photo successfully uploaded" = "Back identity document photo successfully uploaded";
+
+/* Title of ID document scanning screen when scanning the back of either a driver's license, passport, or government issued photo id */
+"Back of %@" = "Back of %@";
```
cjmisenas-stripe, line 20: By passing the document type, that means the document type string won't be localized right?

StripeIdentity/StripeIdentity/Resources/Localizations/en.lproj/Localizable.strings
Pearapps: Currently we only handle specific id type copy for if there is only 1 id type - otherwise it uses `identity document` if there's multiple types. I do see that the web supports various strings for each combination, which we could also handle here if we wanted to. I did not implement that at first because the issue didn't seem to have it as a requirement

StripeIdentity/StripeIdentity/Resources/Localizations/en.lproj/Localizable.strings
Pearapps: We are passing in a localized string here, so it is still localized

StripeIdentity/StripeIdentity/Resources/Localizations/en.lproj/Localizable.strings
Pearapps: I updated the PR to handled each combination!

StripeIdentity/StripeIdentity/Source/NativeComponents/DocumentSide.swift
```diff
@@ -7,8 +7,57 @@
...
+                return idDocument()
+            }
+        } else if localizedTypes.count == 3 {
+            // Handle all three types
+            return self == .front ? String.Localized.frontOfAllIdTypes : String.Localized.backOfAllIdTypes
+        } else if localizedTypes.count == 1, let type = localizedTypes.first {
+            // Handle single type
+            switch self {
+            case .front:
+                return String(format: STPLocalizedString("Front of %@", "Title of ID document scanning screen when scanning the front of either a driver's license, passport, or government issued photo id "), type)
```
joyceqin-stripe: Would prefer to have these STPLocalizedStrings consolidated in String+Localized

joyceqin-stripe (dismissed):

StripeIdentity/StripeIdentity/Source/NativeComponents/DocumentSide.swift
Pearapps: Would you like me to update that here, or as a follow up if possible? @joyceqin-stripe

Pearapps:
I believe I have addressed the comments from https://github.com/stripe/stripe-android/pull/11274, which also apply here. 

cc/ @cjmisenas-stripe

PR #5211, opened 2025-07-24
Add customizable embedded row subtitle font

yuki-stripe:
## Motivation
https://docs.google.com/document/d/1x8TbzCSFFZf_bZFoQoyAPRbgtV2ydU1e4veY-PS_Wd8/edit?tab=t.0#heading=h.m5oqcxpq5af6

## Testing
See snapshot tests

## Changelog
See changelog


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetAppearance.swift
```diff
@@ -329,6 +329,9 @@ public extension PaymentSheet.Appearance {
             /// - Note: If `nil`, uses a default font based on `appearance.font`
             public var titleFont: UIFont?
 
+            /// The font of the subtitle in a payment method row e.g. "Buy now or pay later with Klarna"
+            public var subtitleFont: UIFont?
```
porter-stripe, line 334: Should we add a note similar to what we do for `titleFont` when it's nil?
`/// - Note: If `nil`, uses a default font based on `appearance.font``

porter-stripe (approved):

porter-stripe (approved):

PR #5221, opened 2025-07-28
Crypto Onramp SDK: Example App

mliberatore:
## Summary
- Adds an example app to exercise usage of the `StripeCryptoOnramp` APIs.
- Does not yet incorporate Appearance-related APIs, which will come next. I wanted to keep that work separate so it's easier to review the SDK additions separately from all of the boilerplate of setting up the example.

The sample project’s deployment target is iOS 15, which matches StripeConnect Example, which has the latest deployment target of all of the examples.

## Motivation
To make manually testing existing and new public APIs easier, and to demonstrate correct usage of the SDK.

## Testing

Ran through the flow successfully with an unregistered email:

https://github.com/user-attachments/assets/1298f0de-b75c-4602-af49-89d09b1a3094

Ran through the flow successfully with a registered email:

https://github.com/user-attachments/assets/c317af5b-5717-43e5-879a-113f3902b73b

Ensured that error messages appeared when validation failed or API errors occurred:

Simulator Screenshot - iPhone 16 Pro - 2025-07-28 at 15 56 01
Simulator Screenshot - iPhone 16 Pro - 2025-07-28 at 15 56 26

## Changelog

N/A


Example/CryptoOnramp Example/CryptoOnramp Example/CryptoOnrampExampleView.swift
```diff
@@ -0,0 +1,168 @@
...
+                            .submitLabel(.go)
+                            .onSubmit {
+                                if !isNextButtonDisabled {
+                                    lookupConsumerAndContinue()
+                                }
+                            }
+                    }
+
+                    Button("Next") {
+                        lookupConsumerAndContinue()
```
mats-stripe, line 55: Is it possible to programmatically dismiss the keyboard here using `isEmailFieldFocused`? I noticed the same issue you describe in https://github.com/stripe/stripe-ios/pull/5149#pullrequestreview-3020931437

mats-stripe (dismissed):
LGTM! Just one question

Example/CryptoOnramp Example/CryptoOnramp Example/CryptoOnrampExampleView.swift
mliberatore: Great call, thanks–I only saw this occur a few times, but yes, after adding this in it seems to no longer occur :+1:

PR #5234, opened 2025-07-30
Crypto Onramp SDK: Appearance Customization

mliberatore:
## Summary
- Adds APIs to `StripeCryptoOnramp` to customize the appearance of certain Link UIs.
- Implements these changes for the verification view.
- Note that some customizations, such as the primary button appearance, are not yet present in UI presented via Crypto Onramp.

## Motivation
To allow clients of `StripeCryptoOnramp` to maintain some of their own branding when presenting Stripe-owned UI.

## Testing

Tested using the example project to ensure primary color and user interface style were being respected:

**Always Dark:**
Simulator Screenshot - iPhone 16 Pro - 2025-07-30 at 14 39 20 Simulator Screenshot - iPhone 16 Pro - 2025-07-30 at 14 39 23

**Always Light:**
Simulator Screenshot - iPhone 16 Pro - 2025-07-30 at 14 39 50Simulator Screenshot - iPhone 16 Pro - 2025-07-30 at 14 39 48

**Automatic:**
Simulator Screenshot - iPhone 16 Pro - 2025-07-30 at 14 40 12Simulator Screenshot - iPhone 16 Pro - 2025-07-30 at 14 40 15

## Changelog

WIP – @mats-stripe I might need some assistance here as this is the first change I’m making to an _existing_ framework, not just Crypto Onramp.


Stripe/StripeiOSTests/ConfirmButtonSnapshotTests.swift
```diff
@@ -200,6 +200,7 @@ class ConfirmButtonSnapshotTests: STPSnapshotTestCase {
 
         let confirmButton = ConfirmButton.makeLinkButton(
             callToAction: .continue,
+            appearance: PaymentSheet.Appearance.defaultLinkUIAppearance,
             didTap: {}
         )
         verify(confirmButton)
```
mliberatore, line 206: Just a general comment about this test. The lines above the fold…

```swift
// This should not have any effect when rendered in Link…
appearance.primaryButton.height = 30
```

…I don’t believe are testing what is actually intended. Since `appearance` is an instance of a `struct`, and it is a _copy_ of `PaymentSheet.Appearance.default`, even if we weren’t reconfiguring button height from within `makeLinkButton`, this wouldn’t have an effect, as we’re not modifying some globally mutable state that `makeLinkButton` would otherwise use.

A snapshot test in genreal is certainly fine, but if you’d like me to remove these lines above, I’m happy to do so.

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Extensions/ConfirmButton+Link.swift
```diff
@@ -12,12 +12,13 @@ extension ConfirmButton {
...
         callToAction: CallToActionType,
+        appearance: PaymentSheet.Appearance,
         compact: Bool = false,
         didTap: @escaping () -> Void
     ) -> ConfirmButton {
         let directionalLayoutMargins = compact ? LinkUI.compactButtonMargins : LinkUI.buttonMargins
 
-        var appearance = LinkUI.appearance
+        var appearance = appearance
         appearance.primaryButton.height = LinkUI.primaryButtonHeight(margins: directionalLayoutMargins)
```
mliberatore, line 22: Note that in a future PR, this may need to be updated if we want to allow customization via `LinkAppearance`. This is hardcoded call this method, but we’re not currently displaying UI from Crypto Onramp, so I didn’t want to make an unwanted change ahead of being able to test it.

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/LinkUI.swift
```diff
@@ -188,9 +188,9 @@ extension LinkUI {
 
 // MARK: - Appearance
 
-extension LinkUI {
+public extension PaymentSheet.Appearance {
 
-    static let appearance: PaymentSheet.Appearance = {
+    static let defaultLinkUIAppearance: PaymentSheet.Appearance = {
```
mliberatore, line 193: Originally we talked about moving away from a `static` property for this, but I think it still makes sense to have a default value to fall back to. This way, we can initialize `PaymentSheet.Appearance.linkUIAppearance` with what we currently provide as defaults.

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/LinkUI.swift
```diff
@@ -188,9 +188,9 @@ extension LinkUI {
 
 // MARK: - Appearance
 
-extension LinkUI {
+public extension PaymentSheet.Appearance {
```
mliberatore, line 191: This extension is exposed publicly because we need to apply changes to `defaultLinkUIAppearance` using a `LinkAppearance` instance from Crypto Onramp. This is done in the extension in `LinkAppearance.swift`. Passing `LinkAppearance` to `LinkController` would create a circular dependency on the two frameworks. An alternative would be to make `LinkAppearance` owned by `StripePaymentSheet`.

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/LinkUI.swift
```diff
@@ -199,6 +199,7 @@ extension LinkUI {
         // Text
         appearance.colors.text = .linkTextPrimary
         appearance.colors.textSecondary = .linkTextSecondary
+        appearance.colors.brandText = .linkTextBrand
```
mliberatore, line 202: Calling this out as this is the first usage of a newly added `brandText` in this PR. This was added because the "Resend Code" button on the link verification view was hardcoded to use a color not present in `PaymentSheet.Appearance`.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/WalletButtonsView/WalletButtonsView.swift
```diff
@@ -63,7 +63,7 @@ import WebKit
                     case .linkInlineVerification(let account):
                         LinkInlineVerificationView(
                             account: account,
-                            appearance: flowController.configuration.appearance,
+                            appearance: flowController.configuration.linkUIAppearance,
```
mliberatore: Note that unlike many other replacements in this PR, we are not moving from `LinkUI.apperance` to the configurable `linkUIAppearance`. Instead, we’re moving from the existing `PaymentSheet.Appearance` to the one configured for link.

It’s not clear to me whether this should’ve previously been using `LinkUI.appearance` or not. To minimize potential impact here, I’m happy to revert this line for now, until Crypto Onramp can test specific usage of `WalletButtonsView`’s created `LinkInlineVerificationView`. Please let me know, and I can make a note of it to come back to at that time.

StripeUICore/StripeUICore/Source/Controls/OneTimeCodeTextField.swift
```diff
@@ -838,7 +838,7 @@ private extension OneTimeCodeTextField {
             borderLayer.backgroundColor = backgroundColor.resolvedColor(with: traitCollection).cgColor
             borderLayer.borderColor = theme.colors.border.resolvedColor(with: traitCollection).cgColor
             caret.backgroundColor = label.textColor.resolvedColor(with: traitCollection).cgColor
-            focusRing.borderColor = tintColor.cgColor
+            focusRing.borderColor = tintColor.resolvedColor(with: traitCollection).cgColor
```
mliberatore, line 841: The values changed here were previously not respecting forced override of user interface style. This fixes the issue, which should be evident in the screenshots.

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController-NewPaymentViewController.swift
```diff
@@ -78,7 +79,11 @@ extension PayWithLinkViewController {
         private func makeConfiguration() -> PaymentElementConfiguration {
             var configuration = context.configuration
             configuration.linkPaymentMethodsOnly = true
-            configuration.appearance = LinkUI.appearance
+
+            // reconfigure `appearance` to use `linkUIAppearance` so all non-Link-specific
+            // UI gets appropriately customized for our Link use cases.
+            configuration.appearance = configuration.linkUIAppearance
+
```
mliberatore: I called this change out in slack as a bit of an awkward one here: https://lickability.slack.com/archives/C094P3BRBL1/p1753813784878079?thread_ts=1753798256.703019&cid=C094P3BRBL1

Stripe/StripeiOSTests/ConfirmButtonSnapshotTests.swift
mats-stripe: Nice catch, my guess is that this should be written as: 

```swift
var appearance = PaymentSheet.Appearance.defaultLinkUIAppearance
// This should not have any effect when rendered in Link
appearance.primaryButton.height = 30

let confirmButton = ConfirmButton.makeLinkButton(
    callToAction: .continue,
    appearance: appearance,
    didTap: {}
)
verify(confirmButton)
```

_That being said_, and as this tests suggests, we originally prevented the primary button in the LinkUI from having a dynamic height that could be changed with `appearance.primaryButton.height`, but I think we're gonna want to allow that now. We can add support for that in the next PR that implements the `LinkAppearance.PrimaryButtonConfiguration`!

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController-NewPaymentViewController.swift
mats-stripe: This seems reasonable to me 👍

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Extensions/ConfirmButton+Link.swift
mats-stripe: Exactly!

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/LinkUI.swift
mats-stripe: Absolutely - this makes sense!

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/LinkUI.swift
mats-stripe: Do either of the approaches seem more sensible to you? I lean towards keeping `LinkAppearance` in `StripeCryptoOnramp`, since we don't want to offer appearance customization of the Link UI elsewhere, but don't have ver strong opinions here

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/WalletButtonsView/WalletButtonsView.swift
mats-stripe: Just checked; we should stick with `flowController.configuration.appearance` here. The appearance here is used to style the OTP view, which should be based on the merchant's specified appearance (as the view will be embedded into their app)

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetAppearance.swift
```diff
@@ -70,6 +70,9 @@ public extension PaymentSheet {
         @_spi(AppearanceAPIAdditionsPreview)
         public var iconStyle: IconStyle = .filled
 
+        /// The interface style used by select Link UI.
+        public var linkUserInterfaceStyle: UIUserInterfaceStyle = .unspecified
```
mats-stripe: Instead of a new field `PaymentSheet.Appearance.linkUserInterfaceStyle`, I think we should set the `LinkAppearance.style` to `PaymentSheet.Configuration.style`. We can use those existing helper to configure the Link UI's view controllers accordingly

https://github.com/stripe/stripe-ios/blob/dcb47fe0aedfb8c9d4d6758560ef678c567afef3/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift#L18-L42

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetAppearance.swift
```diff
@@ -172,6 +175,9 @@ public extension PaymentSheet {
             /// The color used for input placeholder text
             public var componentPlaceholderText: UIColor = .secondaryLabel
 
+            /// The color used for brand-tinted text.
+            public var brandText: UIColor = .label
```
mats-stripe, line 176: For now, instead of adding this property, let's stick with using `primary` as the color for both primary buttons and brand-tinted text.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentElementConfiguration.swift
```diff
@@ -29,6 +29,7 @@ protocol PaymentElementConfiguration: PaymentMethodRequirementProvider {
     var defaultBillingDetails: PaymentSheet.BillingDetails { get set }
     var savePaymentMethodOptInBehavior: PaymentSheet.SavePaymentMethodOptInBehavior { get set }
     var appearance: PaymentSheet.Appearance { get set }
+    var linkUIAppearance: PaymentSheet.Appearance { get set }
```
mats-stripe: Is it possible for `linkUIAppearance` to strictly be scoped to `LinkController`? As in, wherever we present Link UI in LinkController, we could do something like `configuration.appearance = linkUIAppearance`, and only look at the `appearance` object in the Link UI? 

There might be something I'm missing here that would require us to have a separate instance of `PaymentSheet.Appearance`, so let me know!

mats-stripe (commented):
Apologies for the swarm of comments; I left some thoughts / suggestions that I'm happy to chat about!

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/LinkUI.swift
mliberatore: I lean the same way as you here, that `LinkAppearance` customization is a feature of `StripeCryptoOnramp`. My only hesitation was whether it was acceptable to your team that this type be exposed publicly to allow that. If that seems okay, then I’m happy to keep this for now.

That said, if Android and iOS need to stay in sync here, I believe Andrew and Carlos are going with the latter approach, that `LinkAppearance` lives in `StripePaymentSheet`. We can chat about that with everyone if that's a requirement to keep the implementations close!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentElementConfiguration.swift
mliberatore: I think I can make that work. Will revise and let you know if there's any concerns that pop up.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetAppearance.swift
mliberatore: Absolutely. I hadn’t seen this existing property on `PaymentSheet.Configuration`, but that works perfectly.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetAppearance.swift
mliberatore: Yes, I can do that. But please note that the existing `linkPlain` button style was hardcoding `.linkTextBrand` (`dynamic(light: brand600, dark: brand200)`), and in either case (link UI default appearance, or `PaymentSheet.Appearance.default`) `primary` is _not_ set to `.linkTextBrand` so those buttons _will_ change colors from what they are on `master`, regardless of whether we’re presenting via `LinkController`.

| Before | After |
|--------|-------|
| Before | After |

 And the opposite would be true, that if we assign `primary` to `linkTextBrand`, that'd effect any existing usage of `primary` throughout the UI.

That feels like an unwanted regression, so I’ll await further comment on this one just to make sure we're on the same page before proceeding.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentElementConfiguration.swift
mliberatore: Ok, done in [b0475bb](https://github.com/stripe/stripe-ios/pull/5234/commits/b0475bb5797c4d98c9f2e67e770d15f38b66c38d) if you’d like to see the diff. A few notes:

- Rather than storing `linkUIAppearance` and applying it when we're about to present UI, I set it on the `LinkController`’s `configuration.appearance`. If there's a specific reason to go with the suggestion of applying it at _presentation time_, happy to make the change.
- My concern with this approach is that any UI that was using the previous hardcoded `LinkUI.appearance` values (i.e. `master`, not this branch) will now be using `configuration.appearance`, since we needed to make those areas configurable for customization. For `LinkController` UI presentation, that's fine, since we're overriding and making use of what used to be `LinkUI.appearance`, and passing that through `configuration.appearance`. But if any of these existing UI components are used _outside_ of `LinkController`, they're no longer wired up to use the values from `LinkUI.appearance`, and would change as a result. If that's the case, we’d probably want to roll back a number of the changes that move away from hardcoded `LinkUI.appearance` calls to prevent unwanted changes.

mliberatore: ^ I’d probably expect some snapshot test failures as a result of this change, but I’ve pushed it here so it's easier to see and discuss.

mliberatore:
> Apologies for the swarm of comments

No worries at all! I fully expect this set of changes to require this kind of iteration.

mliberatore:
Closing in favor of a simplified approach in #5251

PR #5237, opened 2025-07-30
Update default metrics with latest feedback from design

davidme-stripe:
## Summary
Based on feedback from design, we're giving the forms a bit more breathing room and updating the spacing around the "X".

## Testing
Regenerated all 300(!) snapshot tests, inspected each with Kaleidoscope

## Changelog
Added


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/CardSection/CardSectionWithScannerView.swift
```diff
@@ -55,7 +55,7 @@ final class CardSectionWithScannerView: UIView {
 
         let stack = UIStackView(arrangedSubviews: [cardSectionTitleAndButton, cardSectionView, cardScanningView])
         stack.axis = .vertical
-        stack.spacing = ElementsUI.sectionSpacing
+        stack.spacing = 8
```
mats-stripe: Below this we're increasing the value of `sectionSpacing` (i.e. `ElementsUI.formSpacing`) from 12 -> 16, but in this instance we're decreasing the spacing to 8, and no longer using the semantic value? Is this all expected?

mats-stripe (dismissed):
LGTM, assuming the change I called out is intentional!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/CardSection/CardSectionWithScannerView.swift
davidme-stripe: I was confused by this too: [ElementsUI.sectionSpacing is 4](https://github.com/stripe/stripe-ios/blob/43836a8eaf5749de763020fda0b9ccdc297e6f2b/StripeUICore/StripeUICore/Source/Elements/ElementsUI.swift#L22), but [*ElementsAppearance*.sectionSpacing](https://github.com/stripe/stripe-ios/blob/43836a8eaf5749de763020fda0b9ccdc297e6f2b/StripeUICore/StripeUICore/Source/Elements/ElementsUI.swift#L96C16-L96C30) is ElementsUI.formSpacing, which is 16!

But you make a good point that it's confusing, I'm going to rename these.

mats-stripe: Ahh gotcha - thanks for clearing it up!

mats-stripe (approved):

PR #5240, opened 2025-07-31
Allow 3 attestations per day

davidme-stripe:
## Summary
We've found that this code is relatively stable now that we've rolled it out more widely. Instead of rate limiting to one attestation per 24 hour period, let's limit to 3 per day. (Using `calendar.isDate(now, inSameDayAs: firstAttempt)`)

This makes it easier to switch between production and development builds without being stuck without attestation for 24 hours.

## Testing
Updated existing tests.

## Changelog
N/A, private beta feature


gimenete-stripe (approved):
LGTM

mats-stripe (approved):

PR #5245, opened 2025-08-01
Allow skipping Link wallet for valid cards

tillh-stripe:
## Summary

This pull request adds the functionality to skip the Link wallet screen and immediately return a valid payment method and shipping address to the merchant if they're opening Link for payment method selection.

The functionality is currently gated behind `link_mobile_skip_wallet_in_flow_controller`.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController.swift
@@ -86,6 +86,7 @@ final class PayWithLinkViewController: BottomSheetViewController {
         let shouldFinishOnClose: Bool
         let shouldShowSecondaryCta: Bool
         let launchedFromFlowController: Bool
+        let canSkipWalletAfterVerification: Bool
         let initiallySelectedPaymentDetailsID: String?
         let callToAction: ConfirmButton.CallToActionType
         var lastAddedPaymentDetails: ConsumerPaymentDetails?
@@ -110,6 +111,7 @@ final class PayWithLinkViewController: BottomSheetViewController {
         ///   - shouldFinishOnClose: Whether or not Link should finish with `.canceled` result instead of returning to Payment Sheet when the close button is tapped.
         ///   - shouldShowSecondaryCta: Whether or not a secondary CTA to pay another way should be shown.
         ///   - launchedFromFlowController: Whether the flow was opened from `FlowController`.
+        ///   - canSkipWalletAfterVerification: Whether or not we should try to skip showing the wallet after verification.
         ///   - initiallySelectedPaymentDetailsID: The ID of an initially selected payment method. This is set when opened instead of FlowController.
         ///   - callToAction: A custom CTA to display on the confirm button. If `nil`, will display `intent`'s default CTA.
         ///   - analyticsHelper: An instance of `AnalyticsHelper` to use for logging.
@@ -121,6 +123,7 @@ final class PayWithLinkViewController: BottomSheetViewController {
             shouldFinishOnClose: Bool,
             shouldShowSecondaryCta: Bool = true,
             launchedFromFlowController: Bool = false,
+            canSkipWalletAfterVerification: Bool,
             initiallySelectedPaymentDetailsID: String?,
             callToAction: ConfirmButton.CallToActionType?,
             analyticsHelper: PaymentSheetAnalyticsHelper
@@ -132,6 +135,7 @@ final class PayWithLinkViewController: BottomSheetViewController {
             self.shouldFinishOnClose = shouldFinishOnClose
             self.shouldShowSecondaryCta = shouldShowSecondaryCta
             self.launchedFromFlowController = launchedFromFlowController
+            self.canSkipWalletAfterVerification = canSkipWalletAfterVerification
             self.initiallySelectedPaymentDetailsID = initiallySelectedPaymentDetailsID
             self.callToAction = callToAction ?? .makeDefaultTypeForLink(intent: intent)
             self.analyticsHelper = analyticsHelper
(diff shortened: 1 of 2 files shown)


StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/PaymentDetails.swift
```diff
@@ -81,6 +81,13 @@ extension ConsumerPaymentDetails {
 
         return true
     }
+
+    var isValidCard: Bool {
+        guard case let .card(cardDetails) = details else {
+            return false
+        }
+        return !cardDetails.hasExpired && !cardDetails.shouldRecollectCardCVC
```
mats-stripe, line 89: Is it still valid if the card requires 3D2S? I think it is, but wanted to double check

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController.swift
```diff
@@ -188,12 +191,14 @@ final class PayWithLinkViewController: BottomSheetViewController {
...
                 analyticsHelper: analyticsHelper
             ),
-            linkAccount: linkAccount
+            linkAccount: linkAccount,
+            canSkipWalletAfterVerification: canSkipWalletAfterVerification
         )
     }
 
-    private init(context: Context, linkAccount: PaymentSheetLinkAccount?) {
+    private init(context: Context, linkAccount: PaymentSheetLinkAccount?, canSkipWalletAfterVerification: Bool) {
```
mats-stripe: Could / should `canSkipWalletAfterVerification` live in the `Context` model? Especially since the context has an `elementsSession` already, so I think it could be computed as:

```swift
var canSkipWalletAfterVerification: Bool {
    elementsSession.canSkipLinkWallet
}
```

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Utils/LinkBillingDetailsValidator.swift
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    /// Validates that the `ConsumerPaymentDetails` has all required billing details. If not, it will attempt to fill in the missing details. If this is not possible,
+    /// it returns an `.incomplete` result with with partially filled billing details.
+    func validate(
+        _ consumerPaymentDetails: ConsumerPaymentDetails
+    ) async -> ValidationResult {
+        guard isMissingRequestedBillingDetails(consumerPaymentDetails) else {
+            // No additional billing details to collect.
+            let confirmationExtras = LinkConfirmationExtras(billingPhoneNumber: nil)
```
mats-stripe, line 36: Is it expected that phone is always nil here?

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Utils/LinkBillingDetailsValidator.swift
```diff
@@ -0,0 +1,120 @@
...
+
+        return !paymentDetailsAreSupported
+    }
+
+    private func updateBillingDetails(
+        paymentDetailsID: String,
+        billingAddress: BillingAddress?,
+        billingEmailAddress: String?
+    ) async throws -> ConsumerPaymentDetails {
+        let billingDetails = STPPaymentMethodBillingDetails(
```
mats-stripe, line 97: Should this also take a name and phone number?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/PayWithNativeLinkController.swift
```diff
@@ -83,7 +83,8 @@ final class PayWithNativeLinkController {
             from: presentingController,
             shouldOfferApplePay: shouldOfferApplePay,
             hidingUnderlyingBottomSheet: hidingUnderlyingBottomSheet,
-            shouldFinishOnClose: shouldFinishOnClose
+            shouldFinishOnClose: shouldFinishOnClose,
+            canSkipWalletAfterVerification: false
```
mats-stripe: Super minor: A code comment would be helpful to understand why this is hardcoded here 
```suggestion
            canSkipWalletAfterVerification: false // Only available for payment method selection
```

mats-stripe (commented):
Doing some manual testing still, but I have a few minor code comments / questions:

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/PaymentDetails.swift
tillh-stripe: Yes!

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController.swift
tillh-stripe: ~I initially had it there, but since it's only used in one screen, I removed it again. Using a computed property makes that more feasible though!~

It might not be this easy, because `PayWithLinkViewController` is sometimes launched with `canSkipLinkWallet` and sometimes without.

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Utils/LinkBillingDetailsValidator.swift
tillh-stripe: The billing details already contains a phone number, so we don't need to pass any value here.

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Utils/LinkBillingDetailsValidator.swift
tillh-stripe: Phone is pass via `LinkConfirmationExtras`, and name is part of the address.

PR #5251, opened 2025-08-04
Crypto Onramp SDK: Appearance Customization

mliberatore:
## Summary
- Adds APIs to `StripeCryptoOnramp` and `StripePaymentSheet` to customize the appearance of certain Link UIs.
- Implements these changes only for the verification view.
- Note that some customizations, such as the primary button appearance, are not yet present in UI presented via Crypto Onramp.

## Motivation
To allow clients of `StripeCryptoOnramp` to maintain some of their own branding when presenting Stripe-owned UI.

## Testing
Tested using the example project to ensure primary color (`systemPink` in these examples) and user interface style were being respected:
| `alwaysDark` (Light) | `alwaysDark` (Dark) |
|------|-------|
|  |  |

| `alwaysLight` (Light) | `alwaysLight` (Dark) |
|------|-------|
|  |  |

| `automatic` (Light) | `automatic` (Dark) |
|------|-------|
|  |  |

## Changelog
@mats-stripe I might need some assistance here as this is the first change I’m making to an _existing_ framework, not just Crypto Onramp.


StripeUICore/StripeUICore/Source/Controls/OneTimeCodeTextField.swift
```diff
@@ -838,7 +838,7 @@ private extension OneTimeCodeTextField {
             borderLayer.backgroundColor = backgroundColor.resolvedColor(with: traitCollection).cgColor
             borderLayer.borderColor = theme.colors.border.resolvedColor(with: traitCollection).cgColor
             caret.backgroundColor = label.textColor.resolvedColor(with: traitCollection).cgColor
-            focusRing.borderColor = tintColor.cgColor
+            focusRing.borderColor = tintColor.resolvedColor(with: traitCollection).cgColor
```
mliberatore, line 841: The values changed here were previously not respecting forced override of user interface style. This fixes the issue, which should be evident in the screenshots.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkAppearance.swift
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    /// The primary color used in the Link UI. Defaults to the Link brand color.
+    public let primaryColor: UIColor?
+
+    /// Configuration values for the primary button. Uses reasonable defaults if nothing is provided.
+    public let primaryButton: PrimaryButtonConfiguration?
+
+    /// Style options for colors in the Link UI.
+    public let style: PaymentSheet.UserInterfaceStyle
```
mliberatore, line 40: Note that since has moved to `StripePaymentSheet`, we have less of a need to define our own `style` type, and instead use `PaymentSheet.UserInterfaceStyle`. Calling this out, as it differs from the API in review.

mats-stripe: This is great!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkController.swift
```diff
@@ -118,10 +121,12 @@ import UIKit
     ///
     /// - Parameter apiClient: The `STPAPIClient` instance for this controller. Defaults to `.shared`.
     /// - Parameter mode: The mode in which the Link payment method controller should operate, either `payment` or `setup`.
+    /// - Parameter appearance: Link UI-specific appearance overrides. If not specified, `PaymentSheet.Configuration` defaults are used.
```
mats-stripe: Should this be: 
```suggestion
    /// - Parameter appearance: Link UI-specific appearance overrides. If not specified, `PaymentSheet.Appearance` defaults are used.
```

I guess it's both, since `PaymentSheet.UserInterfaceStyle` lives in the configuration object  🤷

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Verification/LinkVerificationViewController.swift
```diff
@@ -56,16 +58,20 @@ final class LinkVerificationViewController: UIViewController {
...
         self.linkAccount = linkAccount
+        self.appearance = appearance
         super.init(nibName: nil, bundle: nil)
 
         if mode.requiresModalPresentation {
             modalPresentationStyle = .custom
             transitioningDelegate = TransitioningDelegate.sharedDelegate
         }
+
+        appearance?.style.configure(self)
```
mats-stripe: We already apply the style to this view in `LinkVerificationController`: 

https://github.com/stripe/stripe-ios/blob/b06fc1fcadbccbd6acd252e243ef3e4d1aa5f473/StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Verification/LinkVerificationController.swift#L28

We also already do it for the rest of the Link UI, in `PayWithLinkViewController.viewDidLoad` and `BaseViewController.present`. That style value comes from the `PaymentSheet.Configuration`. 

I wonder if it's possible to consolidate these. Would it work if in `LinkController.create` we did something like:

```swift
configuration.style = appearance.style
```

and removed this change?

Example/CryptoOnramp Example/CryptoOnramp Example/RegistrationView.swift
```diff
@@ -11,6 +11,9 @@ import SwiftUI
 @_spi(CryptoOnrampSDKPreview)
 import StripeCryptoOnramp
 
+@_spi(STP)
+import StripePaymentSheet
```
mats-stripe, line 15: Is this import needed now?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkController.swift
mliberatore: Yeah, that was the intention, thanks!

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Verification/LinkVerificationViewController.swift
mliberatore: Great suggestion! Additionally, I hadn’t seen this configuration being done on `LinkVerificationController`, so thanks for calling that out.

Example/CryptoOnramp Example/CryptoOnramp Example/RegistrationView.swift
mliberatore: Yes. Although it's not evident in the diff, `RegistrationView`’s preview initializes a `CryptoOnrampCoordinator` and calls `init` on the `LinkAppearance` (formerly `Appearance`) it passes in. Since this type moved to `StripePaymentSheet`, this `import` is needed to use the publicly exposed initializer.

```
 'LinkAppearance' initializer is inaccessible due to '@_spi' protection level
```

I think the only way to remove the import would be to provide a default value for `LinkAppearance` in its `create` method. Let me know if that's something you’d like me to add to the public API.

mats-stripe: Ahh makes sense! This is fine!

PR #5252, opened 2025-08-04
Add PassiveCaptcha to STPElementsSession

joyceqin-stripe:
## Summary

Parse `passive_captcha` from `v1/elements/sessions` response using the `elements_enable_passive_captcha` flag
## Motivation

[MOBILESDK-3874](https://jira.corp.stripe.com/browse/MOBILESDK-3874)
## Testing

Decoding test
## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/PassiveCaptcha.swift
```diff
@@ -0,0 +1,39 @@
...
+/// PassiveCaptcha, delivered in the `v1/elements/sessions` response.
+/// - Seealso: https://git.corp.stripe.com/stripe-internal/pay-server/blob/master/lib/elements/api/resources/elements_passive_captcha_resource.rb
+struct PassiveCaptcha: Equatable, Hashable {
+
+    let siteKey: String
+    let rqData: String?
+
+    /// Helper method to decode the `v1/elements/sessions` response's `passive_captcha` hash.
+    /// - Parameter response: The value of the `passive_captcha` key in the `v1/elements/sessions` response.
+    public static func decoded(
```
gbirch-stripe, line 19: Can we use normal `Decodable` conformance? That would make this more standardized and intuitive to use

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/PassiveCaptcha.swift
```diff
@@ -0,0 +1,39 @@
...
+//  StripePaymentSheet
+//
+//  Created by Joyce Qin on 8/3/25.
+//
+
+import Foundation
+
+/// PassiveCaptcha, delivered in the `v1/elements/sessions` response.
+/// - Seealso: https://git.corp.stripe.com/stripe-internal/pay-server/blob/master/lib/elements/api/resources/elements_passive_captcha_resource.rb
+struct PassiveCaptcha: Decodable {
```
gbirch-stripe: Should have been more clear — meant implementing a custom `Decodable` conformance and using that to decode from the JSON. Looking at the context now, though, I see that what you did is the established pattern, and `Decodable` doesn't really make sense since it's going to be used in the context of already having this sub-dict with the data

joyceqin-stripe: Should I revert, then?

gbirch-stripe: Yes, otherwise it will use the default `Decodable` conformance, which won't work and could be confusing

PR #5257, opened 2025-08-05
Crypto Onramp SDK: KYC Collection API

mliberatore:
## Summary

Adds the public API for collecting KYC information. This is not yet wired up to the example app, so testing is done through unit tests and manual work to perform the request.

## Motivation

Required API for [the CryptoOnramp SDK](https://docs.google.com/document/d/1fiRBeegAqMbaR7OUDUy0EA4idvHEDh_FcBE2DTz2vyo/edit?tab=t.0#heading=h.570u4leviueu)

## Testing
1. Ran the newly added tests to ensure encoding / decoding were working as expected.
2. Triggered the API from the example app using test code to ensure the request was succeeding.
```swift
Task {
    do {
        try await coordinator.collectKYCInfo(
            info: .init(
                firstName: "Michael",
                lastName: "Liberatore",
                idNumber: "123456789",
                address: .init(
                    city: "New York",
                    country: "US",
                    line1: "123 Fake Street",
                    line2: "APT 2",
                    postalCode: "10108",
                    state: "New York"
                ),
                dateOfBirth: Date(timeIntervalSince1970: 0),
                birthCountry: "US",
                birthCity: "Pittsburgh"
            )
        )
    } catch {
        errorMessage = error.localizedDescription
    }
}
```

Finder 2025-08-05 11 12 19

## Changelog
N/A


StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/IdType.swift
```diff
@@ -0,0 +1,38 @@
+//
+//  IdType.swift
```
mliberatore, line 2: I noticed this type ended up getting removed from the mobile API review doc, and `idType` is now set to `"social_security_number"`. I had copied this information from prior, should I update to match the latest and hardcode the `idType` property instead?

StripeCryptoOnramp/StripeCryptoOnramp/Source/Internal/API Bindings/Credentials.swift
```diff
@@ -0,0 +1,21 @@
+//
+//  Credentials.swift
```
mliberatore, line 2: I moved this out `CustomerRequest` since now both of our requests make use of it.

StripeCryptoOnramp/StripeCryptoOnramp/Source/Internal/API Bindings/KYCDataCollectionRequest.swift
```diff
@@ -0,0 +1,79 @@
+//
+//  KYCDataCollectionRequest.swift
```
mliberatore, line 2: To keep `KycInfo` as a pure model representation of the data we collect, I added this `internal` wrapper type to handle the addition of `credentials` and to assist with `Date` to `mm/dd/yyyy` component conversion. 

A specifiable calendar is necessary in order to test the API in a stable way across machines in different timezones, rather than the tests falling back to using `Calendar.current`, creating a moving target. This behavior is explicitly tested in the added unit tests. A regular user will default to `Calendar.current`, so the date they input from any UI (e.g. `DatePicker`) will use their calendar by default, and the encoding will match so that fixed point in time (`Date`) serializes properly to the day / month / year in their time zone.

StripeCryptoOnramp/StripeCryptoOnramp/Source/Internal/API Bindings/KYCDataCollectionResponse.swift
```diff
@@ -0,0 +1,27 @@
...
+//  KYCDataCollectionResponse.swift
+//  StripeCryptoOnramp
+//
+//  Created by Michael Liberatore on 8/4/25.
+//
+
+import Foundation
+
+/// Codable model representing a response from the `/v1/crypto/internal/kyc_data_collection` endpoint.
+struct KYCDataCollectionResponse: Codable {
```
mliberatore, line 11: `Codable`, not `Decodable`, so that it's easier to generate the response in tests.

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/IdType.swift
mats-stripe: I think the latest on this is that we should only support SSN, so I don't think the full enum is needed (for now at least). 

We could either keep it dead simple and always hardcode `"id_type": "social_security_number"` in the API request, or keep this enum but only have this case. Something like this:

```swift
/// Represents possible types of customer identification.
@_spi(CryptoOnrampSDKPreview)
public enum IdType: String, Codable, CaseIterable {
    case socialSecurityNumber = "social_security_number"
}
```

Up to you!

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/KycInfo.swift
```diff
@@ -0,0 +1,58 @@
...
+    public let address: PaymentSheet.Address
+
+    /// The customer’s date of birth.
+    public let dateOfBirth: Date
+
+    /// The country in which the customer was born.
+    public let birthCountry: String?
+
+    /// The city in which the customer was born.
+    public let birthCity: String?
```
mats-stripe: We can get rid of these two fields, since they're optional in the API, and the UI won't be collecting this info for now. Updated the API review doc accordingly, and sorry for the churn 🙏

StripeCryptoOnramp/StripeCryptoOnramp/Source/Internal/API Bindings/KYCDataCollectionResponse.swift
```diff
@@ -0,0 +1,37 @@
...
+    let firstName: String
+
+    /// The user’s last name.
+    let lastName: String
+
+    /// A list of the user’s nationalities, if any were provided.
+    let nationalities: [String]
+
+    /// The country in which the user currently resides.
+    let residenceCountry: String
```
mats-stripe: All the fields in this model except for `personId` are optional on the [backend model](https://stripe.sourcegraphcloud.com/stripe-internal/pay-server/-/blob/lib/cryptocurrency/onramp/api/structs.rb?L43), we should do the same here to be safe: 
```suggestion
    /// The user’s first name.
    let firstName: String?

    /// The user’s last name.
    let lastName: String?

    /// A list of the user’s nationalities, if any were provided.
    let nationalities: [String]?

    /// The country in which the user currently resides.
    let residenceCountry: String?
```

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/KycInfo.swift
mliberatore: No problem

StripeCryptoOnramp/StripeCryptoOnrampTests/Unit/STPAPIClient+CryptoOnrampTests.swift
```diff
@@ -96,4 +122,126 @@ final class STPAPIClientCryptoOnrampTests: APIStubbedTestCase {
...
+            let parameters = String(data: httpBody, encoding: .utf8)?.parsedHTTPBodyDictionary ?? [:]
+
+            XCTAssertEqual(parameters.count, 14)
+            XCTAssertEqual(parameters["credentials[consumer_session_client_secret]"], Constant.requestSecret)
+            XCTAssertEqual(parameters["first_name"], "John")
+            XCTAssertEqual(parameters["last_name"], "Smith")
+            XCTAssertEqual(parameters["id_number"], "123456789")
+            XCTAssertEqual(parameters["id_type"], "social_security_number")
+            XCTAssertEqual(parameters["line1"], "123%20Fake%20Street")
+            XCTAssertEqual(parameters["line2"], "APT%202")
```
mats-stripe, line 147: Is this encoding expected? It probably is, but wanted to double check

mats-stripe (approved):
Nice, thank you!!

StripeCryptoOnramp/StripeCryptoOnrampTests/Unit/STPAPIClient+CryptoOnrampTests.swift
mliberatore: Yes, `STPAPIClient`’s `post` calls `let formData = URLEncoder.queryString(from: jsonDictionary).data(using: .utf8)` which eventually calls through to `addingPercentEncoding(withAllowedCharacters: URLQueryAllowed)` on each value.

PR #5258, opened 2025-08-05
Crypto Onramp SDK: Start Identity Verification API

mliberatore:
## Summary
- Adds the internal API to start the identity verification process.
- Exposes public API that calls this internal API. Does not yet display Identity SDK UI–that will come in a follow-up PR, at which time this will also be able to be exercised via the example app.

## Motivation
To implement the public API `promptForIdentityVerification` documented [here](https://docs.google.com/document/d/1fiRBeegAqMbaR7OUDUy0EA4idvHEDh_FcBE2DTz2vyo/edit?tab=t.0). 

## Testing
1. Ran the newly added tests.
2. Confirmed that calling the endpoint works successfully with test code after completing OTP verification:
```swift
Task {
    do {
        _ = try await coordinator.promptForIdentityVerification()
    } catch {
        print(error)
    }
}
```

## Changelog
N/A


StripeCryptoOnramp/StripeCryptoOnramp/Source/Internal/API Bindings/StartIdentityVerificationRequest.swift
```diff
@@ -0,0 +1,34 @@
...
+import Foundation
+
+/// Encodable model passed to the `/v1/crypto/internal/start_identity_verification` endpoint.
+struct StartIdentityVerificationRequest: Encodable {
+
+    /// Contains credentials required to make the request.
+    let credentials: Credentials
+
+    /// When `true`, the response will contain an `ephemeral_key` used to initialize the Identity SDK.
+    let isMobile: Bool
```
mats-stripe: This will always be `true`, right? Would it make sense to do: 

```suggestion
    let isMobile: Bool = true
```

mats-stripe (approved):
👌

mats-stripe (approved):

PR #5259, opened 2025-08-05
Rearrange email/phone fields into billing details

davidme-stripe:
## Summary
Merge the "Contact details" section into the "Billing address" section. If email/phone are present, we'll now name it the "Billing details" section.

## Motivation
Latest feedback from design, moves the phone number collection to a more logical place below the card details.

## Testing
Updated snapshot tests and E2E tests. Added some tests specific to this new combined section to ensure we can use defaults, enter details manually, and that the params always include these updated details.

## Changelog
Added


davidme-stripe:
Adding the "Ships Without Translations" tag: We want to ship these to the localization system, but they're not ready to use yet. For now, we'll use the old string and swap it out next week once translations are in.

porter-stripe (dismissed):
clean!

tillh-stripe (approved):

PR #5262, opened 2025-08-06
Passive hcaptcha

joyceqin-stripe:


Tests/installation_tests/cocoapods/with_frameworks_swift/Podfile
```diff
@@ -20,6 +20,16 @@ target 'CocoapodsTest' do
     installer.pods_project.build_configurations.each do |config|
       config.build_settings['GCC_TREAT_WARNINGS_AS_ERRORS'] = 'YES'
     end
+    
+    # Set APPLICATION_EXTENSION_API_ONLY for StripeCore when building for app extensions
```
joyceqin-stripe: I couldn't figure out any other way to get it to pass the install-tests, I'm open to any other suggestions. Moving HCaptcha into StripeCore caused an error: 'shared' is unavailable in application extensions for iOS: Use view controller based solutions where appropriate instead.

This is because HCaptcha's HCaptchaURLOpener and HCaptchaWebViewManager use UIApplication.shared, which is unavailable in application extensions. I added the @available(iOSApplicationExtension, unavailable) annotation to the HCaptcha classes that depend on them, but my function fetchPassiveHCaptchaToken relies on HCaptcha, and I needed to conditionally return nil for application extensions.

StripeCore/StripeCore/Source/Captcha/PassiveCaptcha.swift
```diff
@@ -0,0 +1,63 @@
...
+        )
+    }
+
+    @_spi(STP) public static func fetchPassiveHCaptchaToken(passiveCaptcha: PassiveCaptcha?, completion: @escaping (String?) -> Void) {
+        #if APPLICATION_EXTENSION_API_ONLY
+        // In app extension builds, HCaptcha functionality is not available
+        completion(nil)
+        #else
+        // In test environments, HCaptcha WebView loading may hang, so return nil immediately
+        if STPAnalyticsClient.isUnitOrUITest {
```
joyceqin-stripe: Initializing HCaptcha loads html, and HCaptcha validation can only happen after loading. But in unit tests, it hangs indefinitely.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
 * [Fixed] Fixed an issue that caused EmbeddedPaymentElement to log broken layout constraints.
 * [Fixed] The Apple Pay logo in `PaymentOptionDisplayData` no longer contains additional padding, bringing it in line with other payment method logos.
 * [Changed] When phone number or email address collection is enabled, these fields will now appear in the billing details section instead of appearing above the card field.
+* [Added] Passive HCaptcha on payment method tokenization API requests.
```
davidme-stripe: I don't need to changelog this (it's sort of an implementation detail of Radar)

StripeApplePay/StripeApplePay/Source/ApplePayContext/STPApplePayContext.swift
```diff
@@ -577,259 +579,261 @@ public class STPApplePayContext: NSObject, PKPaymentAuthorizationControllerDeleg
...
-                    self.authorizationController != nil
-                else {
-                    let errorMessage = "Failed on payment method completion"
-                    let errorAnalytic = ErrorAnalytic(event: .unexpectedApplePayError,
-                                                      error: InternalError.invalidState,
-                                                      additionalNonPIIParams: [
-                                                        "error_message": errorMessage,
-                                                      ])
-                    handleFinalState(.error, intentCreationError, errorAnalytic)
+        PassiveCaptcha.fetchPassiveHCaptchaToken(passiveCaptcha: passiveCaptcha) { hcaptchaToken in
```
davidme-stripe: Do we want to do this for Apple Pay? I think the sort of things hCaptcha prevents are less of a concern for Apple Pay transaction, so I'd be inclined to leave it out for latency reasons

StripeCore/StripeCore.xcodeproj/project.pbxproj
```diff
@@ -99,6 +99,21 @@
 		9DCBC08C182ED76A962961E7 /* NSURLComponents+Stripe.swift in Sources */ = {isa = PBXBuildFile; fileRef = D475CD59778FAC1028670AB5 /* NSURLComponents+Stripe.swift */; };
 		9FBA50345D53E82AA974F672 /* STPAPIClient+FileUpload.swift in Sources */ = {isa = PBXBuildFile; fileRef = F4FC2FC535DAF9B340F9EA75 /* STPAPIClient+FileUpload.swift */; };
 		A3E0F5DF2E3C8A0B00C9E5F7 /* STPClientAttributionMetadata.swift in Sources */ = {isa = PBXBuildFile; fileRef = A3E0F5DE2E3C8A0B00C9E5F7 /* STPClientAttributionMetadata.swift */; };
+		A3E0F6E52E43D98D00C9E5F7 /* String+Dict.swift in Sources */ = {isa = PBXBuildFile; fileRef = A3E0F6E32E43D98D00C9E5F7 /* String+Dict.swift */; };
```
davidme-stripe: (And if the above is true, we wouldn't have to move this into StripeCore!)

StripeCore/StripeCore/Source/Captcha/PassiveCaptcha.swift
davidme-stripe: I think we should add some kind of client-side timeout + analytic event for it timing out, considering how critical this is to the payment flow

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
```diff
@@ -181,82 +182,85 @@ extension PaymentSheet {
...
-                )
-                paymentHandler.confirmPayment(
-                    params,
-                    with: authenticationContext,
-                    completion: { actionStatus, paymentIntent, error in
-                        if let paymentIntent {
-                            setDefaultPaymentMethodIfNecessary(actionStatus: actionStatus, intent: .paymentIntent(paymentIntent), configuration: configuration, paymentMethodSetAsDefault: elementsSession.paymentMethodSetAsDefaultForPaymentSheet)
-                        }
-                        paymentHandlerCompletion(actionStatus, error)
+            PassiveCaptcha.fetchPassiveHCaptchaToken(passiveCaptcha: elementsSession.passiveCaptcha) { hcaptchaToken in
```
davidme-stripe: Can we fetch the passive HCaptcha token earlier? Like maybe when PaymentSheet initializes, and then we could have it ready when someone actually hits "Pay"?

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
```diff
@@ -0,0 +1,138 @@
...
+        self.testTimeout = testTimeout
+    }
+
+    @_spi(STP) public func start() {
+        guard let passiveCaptcha, validationTask == nil else { return }
+        let timeoutNs: UInt64 = {
+            if let testTimeout {
+                return testTimeout
+            }
+            return STPAnalyticsClient.isUnitOrUITest ? 0 : 6_000_000_000
```
davidme-stripe, line 58: 6 seconds is a long time, do we expect it to usually take this long?

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
```diff
@@ -0,0 +1,138 @@
...
+            } catch {
+                STPAnalyticsClient.sharedClient.logPassiveCaptchaError(error: error, siteKey: siteKey)
+                return nil
+            }
+        }
+    }
+
+    @_spi(STP) public func fetchToken() async -> String? {
+        if validationTask == nil {
+            start()
```
davidme-stripe, line 108: Could we have an analytic here too? That way we could be alerted if this is slowing down the payment. I'd hope that most of the time, the token would already be ready by the time we call `fetchToken()`

PR #5266, opened 2025-08-06
Don't add additional padding to Apple Pay logo

davidme-stripe:
## Summary
This logo contains additional padding, unlike every other image we vend. This is probably causing the Apple Pay button to appear unexpectedly small in user apps, so let's remove this padding.


gbirch-stripe:
The snapshot tests show that after this changes the logo became noticeably smaller in the payment method list. Is this intentional?

davidme-stripe:
Yes — it looked odd to me at first, but it was actually larger than all the other logos in the past. I do wonder if the outline makes it _optically_ smaller than the other logos, but some (like Giropay) have a similar outline, so I may have just been staring at it for too long. 
CleanShot 2025-08-06 at 08 50 48@2x (1)
CleanShot 2025-08-06 at 08 49 27@2x

porter-stripe (approved):

PR #5267, opened 2025-08-06
Call cancel in formSheetAction.confirm

porter-stripe:
## Summary
- Now when the user cancels the transaction we call the completion block with canceled 

## Motivation
- Bug fix

## Testing
Manual

## Changelog
See diff


yuki-stripe (approved):
nice test!

PR #5270, opened 2025-08-06
Temporarily revert opensCardScannerAutomatically impelementation

gbirch-stripe:
This reverts commit fa09071fbc01eada1c154263f19618e4266d0c01.

## Summary
This is a high risk feature, but the bug bash won't be until Tuesday, after the next SDK release. Waiting until after the release to add these changes.

## Motivation
N/A

## Testing
N/A

## Changelog
reverted changelog


porter-stripe:
Can we just put the public APIs under STP spi? Seems easier and less churny.

gbirch-stripe:
> Can we just put the public APIs under STP spi? Seems easier and less churny.

My concern is that this PR also changes the customer sheet form caching behavior (should have been a separate PR like you said) and changes the layout of the card form, even when the config is off. So I don't think that gating it with spi substantially lowers the risk

PR #5272, opened 2025-08-06
Crypto Onramp SDK: Identity Verification Flow

mliberatore:
## Summary
Implements the rest of `promptForIdentityVerification()`, including use of the Identity SDK to verify the user’s identity and return a result.

## Motivation
To implement the public API `promptForIdentityVerification` documented [here](https://docs.google.com/document/d/1fiRBeegAqMbaR7OUDUy0EA4idvHEDh_FcBE2DTz2vyo/edit?tab=t.0). 

## Testing
Added functionality to the example app to exercise the API. This video shows success, failure, and cancelation results. Note that this also works after registration, but the video only shows reaching this step after verification.

https://github.com/user-attachments/assets/a212d7c3-43b4-424d-a35b-accff9d873ee

## Changelog
N/A


Example/CryptoOnramp Example/CryptoOnramp Example/CryptoOnrampExampleView.swift
```diff
@@ -85,6 +88,9 @@ struct CryptoOnrampExampleView: View {
             .navigationBarTitleDisplayMode(.inline)
         }
         .onAppear {
+            guard coordinator == nil else {
+                return
+            }
```
mliberatore, line 93: I was experiencing an issue where this was getting triggered on dismissal of the identity sheet, effectively resetting verification status. Regardless, I didn’t mean for reinitialization to occur, so I prevent it.

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
```diff
@@ -53,8 +56,9 @@ public protocol CryptoOnrampCoordinatorProtocol {
 
     /// Creates an identity verification session and launches the verification flow.
     ///
+    /// - Parameter viewController: The view controller from which to present the verification flow.
     /// - Returns: An `IdentityVerificationResult` representing the outcome of the verification process.
-    func promptForIdentityVerification() async throws -> IdentityVerificationResult
+    func promptForIdentityVerification(from viewController: UIViewController) async throws -> IdentityVerificationResult
```
mliberatore, line 61: Just highlighting here that this is a change to the `public` API in case the mobile API review doc needs to be updated.

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
```diff
@@ -66,6 +70,9 @@ public final class CryptoOnrampCoordinator: CryptoOnrampCoordinatorProtocol {
 
         /// Phone number validation failed. Phone number should be in E.164 format (e.g., +12125551234).
         case invalidPhoneFormat
+
+        /// `ephemeralKey` is missing from the response after starting identity verification.
+        case missingEphemeralKey
```
mliberatore, line 81: This really shouldn’t ever happen, since we're hardcoding `isMobile` to `true`, and the API is documented to return the ephemeral key when that's the case. But since this key _is_ optional on the response, I’m handling it by `throw`ing if it's absent.

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
mats-stripe: Thank you - updated!

mats-stripe (dismissed):

PR #5273, opened 2025-08-06
Show cancel button in Link card form if launched from FlowController

tillh-stripe:


StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController-NewPaymentViewController.swift
```diff
@@ -50,6 +50,18 @@ extension PayWithLinkViewController {
             return button
         }()
 
+        private lazy var cancelButton: Button? = {
+            guard context.shouldShowSecondaryCta, context.launchedFromFlowController, linkAccount.isInSignupFlow else {
```
mats-stripe: Is there an alternative way to cancel this if it isn't launched from flow controller?  Trying to understand that condition

mats-stripe (dismissed):

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController-NewPaymentViewController.swift
tillh-stripe: If launched from PaymentSheet, we don't need this dedicated button, since the navigation bar's close button will return the user to PaymentSheet.

mats-stripe (approved):

PR #5279, opened 2025-08-07
Update account onboarding component to accept new `requirements` param

torrance-stripe:
## Summary
Updates the account onboarding component to accept a new `requirements` param. Allows the onboarding component to include only the requirements in the `only` field or to `exclude` requirements in the `exclude` field.

https://github.com/user-attachments/assets/04742195-6f42-4119-91c9-2614b87aa60c

You can see in the video that when the param isn't set, onboarding will collect all info. When it is set to `external_account`, it'll only ask for that users's bank account.

## Motivation
These changes were previously approved in this [API Review](https://jira.corp.stripe.com/browse/MOBILE_APIREVIEW-153). 

## Testing
Manually tested by updating the example app. Updated one CI test as well.

### DIFF
--- a/StripeConnect/StripeConnect/Source/Models/AccountCollectionOptions.swift
+++ b/StripeConnect/StripeConnect/Source/Models/AccountCollectionOptions.swift
@@ -27,7 +27,7 @@ public struct AccountCollectionOptions: Equatable, Codable {
         // Without this specific encoder, Swift will encode as `only: {_0: ["requirement.name"]}` as opposed to `only: ["requirement.name"]`
         public func encode(to encoder: Encoder) throws {
             var container = encoder.container(keyedBy: CodingKeys.self)
-            
+
             switch self {
             case .only(let array):
                 try container.encode(array, forKey: .only)


StripeConnect/StripeConnect/Source/Models/AccountCollectionOptions.swift
```diff
@@ -20,11 +20,35 @@ public struct AccountCollectionOptions: Equatable, Codable {
...
+    }
+
     /// Customizes collecting `currently_due` or `eventually_due` requirements
     public var fields: FieldOption = .currentlyDue
 
     /// Controls whether to include [future requirements](https://docs.stripe.com/api/accounts/object#account_object-future_requirements)
     public var futureRequirements: FutureRequirementOption = .omit
 
+    /// Specifies which requirements to collect or exclude
+    public var requirements: RequirementsOption?
```
torrance-stripe, line 51: I think adding this should be backwards compatible? I'm being extra paranoid as my mobile knowledge isn't great. It'd be great to confirm!

StripeConnect/StripeConnect/Source/Models/AccountCollectionOptions.swift
```diff
@@ -20,11 +20,35 @@ public struct AccountCollectionOptions: Equatable, Codable {
         case include
     }
 
+    public enum RequirementsOption: Equatable, Codable {
+        case only([String])
+        case exclude([String])
+
+        // Without this specific encoder, Swift will encode as `only: {_0: ["requirement.name"]}` as opposed to `only: ["requirement.name"]`
```
torrance-stripe, line 27: Not 100% sure this is the right solution, but this is what LLM spit out when I was debugging this problem

StripeConnect/StripeConnect/Source/Models/AccountCollectionOptions.swift
yuki-stripe: Yep, this has an implicit default value of nil and is backwards compatible.

StripeConnect/StripeConnect/Source/Models/AccountCollectionOptions.swift
yuki-stripe: This looks good!

StripeConnect/StripeConnect/Source/Models/AccountCollectionOptions.swift
```diff
@@ -20,11 +20,35 @@ public struct AccountCollectionOptions: Equatable, Codable {
         case include
     }
 
+    public enum RequirementsOption: Equatable, Codable {
+        case only([String])
+        case exclude([String])
```
yuki-stripe, line 25: It'd be great to add docstrings to these (and all the rest too for that matter - but that's not blocking feedback), that's the standard in our payments SDK e.g. [like this](https://github.com/stripe/stripe-ios/blob/af11b4298a40/StripePayments/StripePayments/Source/API%20Bindings/Models/PaymentMethods/STPPaymentMethod.swift). We often just copy and paste and/or link to the API.

As a user I'd be wondering what values I could pass in here.

torrance-stripe: I'll add it once we get closer to preview!

PR #5283, opened 2025-08-07
Crypto Onramp SDK: KYC Info Collection Example

mliberatore:
## Summary
- Adds an example view to exercise the KYC info collection API.
- Deduplicates titled fields across example views.
- Deduplicates SwiftUI preview wrappers used in the example views.

## Motivation
To ensure `CryptoOnrampCoordinator` public API usage is demonstrated in the example app.

## Testing
- Ensured existing views’ previews and functionality were correct.
- Used the example app to test the functionality of the KYC view, and its entry point.

https://github.com/user-attachments/assets/192ab5e0-4b44-4c03-9fba-cc9f2356c0ae

Note that the Date of Birth title label is misaligned in this demo video. I didn’t notice it until after I recorded it. I’ve since fixed this in 00295eeaaa3e1fdd624dd34e484e07709a167522. 

Tower 2025-08-07 15 02 20

## Changelog
N/A


Example/CryptoOnramp Example/CryptoOnramp Example/KYCInfoView.swift
```diff
@@ -0,0 +1,233 @@
...
+import SwiftUI
+
+@_spi(CryptoOnrampSDKPreview)
+import StripeCryptoOnramp
+
+@_spi(STP)
+import StripePaymentSheet
+
+/// A view used to collect KYC (Know Your Customer) data and exercise the `CryptoOnrampCoordinator’s` `collectKYCInfo(info:)` functionality.
+struct KYCInfoView: View {
```
mats-stripe, line 17: Not needed per se, but it might be nice to use our Address Elements here to collect the address: https://docs.stripe.com/elements/mobile/address-element?platform=ios

I believe the output is a `PaymentSheet.Address`, which is exactly what we need!

mats-stripe: Nvm I think the output is `AddressElement.AddressDetails`. I wonder if we should update our API to take that type instead 🤔 WDYT? https://docs.stripe.com/payments/mobile/collect-addresses?payment-ui=mobile&platform=plugins#retrieve-address-details

mats-stripe (dismissed):
LGTM! Just one suggestion

Example/CryptoOnramp Example/CryptoOnramp Example/KYCInfoView.swift
mliberatore: Oh, that's a really nice-looking component! I’m of two minds here though, since our SDK is meant to be mostly "headless" and use as minimal Stripe UI as possible, does demonstrating usage with Stripe-provided components in our example project encourage the wrong usage at all in your opinion? (Not a leading question, I’m just not sure myself).

> I wonder if we should update our API to take that type instead 🤔 WDYT?

Not that it can’t be adjusted, but I don’t love the relationship that gets made here, since `AddressElement.AddressDetails` is actually `AddressViewController.AddressDetails`. It feels a bit odd to pass a view controller’s subtype to a headless API to make a network request, even if that detail isn’t readily obvious because of the `typealias`. But that might be somewhat moot, cause we could also generalize it further down to `STPAddress` and expose the `stpAddress` extension on `AddressViewController.AddressDetails` and use that.

I’ll hold off for the moment to give you a chance to respond before I implement the changes.

mats-stripe: That's totally fair, good callout! The coupling isn't ideal. I'm not totally sure which address type we should use here. 

I don't want to bike-shed this too much, but something that might be nice is a new address model (womp-womp) that includes initializers from other address types (i.e. `STPAddress`, `PaymentSheet.Address`). I'm also OK to leave it as-is in this PR. WDYT?

mliberatore: I’ll pushed a change with the new `Address` type, following the optionality of the properties in `PaymentSheet.Address`. It seems like it might be warranted or at least convention given the number of Address types throughout the various SDKs. That, or a more comprehensive refactor to have them all use an underlying common address type, which is probably out of scope for what we're working towards. I could really go either way here, since we already have the `PaymentSheet` dependency and address type conversion is strictly property mapping. I think we'd need to update the mobile API review doc as well if we move forward. If not, I can easily revert the commit.

Lmk if you’d still like to explore using `AddressElement` in the example project.

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/Address.swift
```diff
@@ -0,0 +1,91 @@
...
+        self.country = country
+        self.line1 = line1
+        self.line2 = line2
+        self.postalCode = postalCode
+        self.state = state
+    }
+
+    /// Initializes an `Address` using an `STPAddress` instance
+    /// - Parameter address: The address from which to generate the new instance.
+    /// - Note: `address.name` is unused in the conversion.
```
mats-stripe, line 60: Phone and email are also unused, but I don't think it's too important to call this out.

mats-stripe (approved):
Love it - thank you!

PR #5286, opened 2025-08-07
Fix BDCC with bank account in Link

tillh-stripe:
## Summary

This pull request fixes an issue where confirming details for a bank account didn't work in Link.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/STPAPIClient+Link.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/STPAPIClient+Link.swift
@@ -466,6 +466,15 @@ extension STPAPIClient {
             }
         }
 
+        if let details = updateParams.details, case .bankAccount(let billingDetails) = details {
+            parameters["billing_address"] = billingDetails.consumersAPIParams
+
+            if let billingEmailAddress = billingDetails.email {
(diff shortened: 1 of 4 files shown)


StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/STPAPIClient+Link.swift
```diff
@@ -466,6 +466,15 @@ extension STPAPIClient {
             }
         }
 
+        if let details = updateParams.details, case .bankAccount(let billingDetails) = details {
```
mats-stripe, line 469: Minor: This and the above condition could be formed as a `switch`

```swift
if let details = updateParams.details {
    switch details {
    case .card(let expiryDate, let billingDetails, let preferredNetwork):
        ...
    case .bankAccount(let billingDetails):
        ...
    }
}
```

mats-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/STPAPIClient+Link.swift
tillh-stripe: Thought about this, but there is quite some code for the `.card` case. I'll try to clean this up later.

PR #5290, opened 2025-08-08
Add field validation when the disabled confirm button is tapped

davidme-stripe:
## Summary
When the user taps a disabled "Confirm" button and some fields are empty, show appropriate error strings.

CleanShot 2025-08-11 at 13 37 55

These aren't the final strings — once this lands, we'll add new "[Card number] is missing" localized strings to replace the incomplete/invalid ones for an empty field.

## Motivation
Users sometimes don't understand why the button is disabled. This provides them with hints on how to complete the form.

## Testing
Updated unit tests
Existing E2E UI tests

## Changelog
Added changelog entry


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/ConfirmButton.swift
```diff
@@ -235,6 +238,9 @@ class ConfirmButton: UIView {
     private func handleTap() {
         if case .enabled = state {
             didTap()
+        } else if case .disabled = state {
+            // When the disabled button is tapped, trigger validation error display
+            didTapWhenDisabled()
```
tillh-stripe, line 243: Might be helpful to execute some `.error` haptic feedback when running this method. What do you think?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/TextField/TextFieldElement+Card.swift
```diff
@@ -98,10 +98,10 @@ extension TextFieldElement {
             case invalidLuhn
             case disallowedBrand(brand: STPCardBrand)
 
-            func shouldDisplay(isUserEditing: Bool) -> Bool {
+            func shouldDisplay(isUserEditing: Bool, displayEmptyFields: Bool) -> Bool {
```
tillh-stripe, line 101: Should we update this method to also show errors on disabled button taps even if the user is still editing? In the video below, I tap the primary button after entering `1111` and it doesn't give an indication of why I can't proceed.

https://github.com/user-attachments/assets/a134075b-f1a5-45b0-90bf-bb5c587188d4

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/ConfirmButton.swift
davidme-stripe: Good idea!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/TextField/TextFieldElement+Card.swift
davidme-stripe: Yep, I'll try that

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedFormViewController.swift
```diff
@@ -382,6 +385,14 @@ class EmbeddedFormViewController: UIViewController {
 
         pay(with: selectedPaymentOption)
     }
+
+    @objc func didTapPrimaryButtonWhenDisabled() {
+        // When the disabled button is tapped, show validation errors on all form fields
+#if !os(visionOS)
+        UINotificationFeedbackGenerator().notificationOccurred(.error)
```
tillh-stripe, line 392: Do we need to call `prepare()` first? If so, might make sense to have a prepared reference to the feedback generator as a property.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetVerticalViewController.swift
```diff
@@ -698,6 +701,16 @@ class PaymentSheetVerticalViewController: UIViewController, FlowControllerViewCo
         pay(with: selectedPaymentOption)
     }
 
+    @objc func didTapPrimaryButtonWhenDisabled() {
+        // When the disabled button is tapped, show validation errors on all form fields
+#if !os(visionOS)
+        UINotificationFeedbackGenerator().notificationOccurred(.error)
```
tillh-stripe, line 707: Same here.

StripeUICore/StripeUICore/Source/Elements/TextField/TextFieldElement+Validation.swift
```diff
@@ -18,18 +18,18 @@ import Foundation
     /// If it doesn't suit your text field's needs, create a new enum instead of modifying this one!
     @frozen enum Error: TextFieldValidationError, Equatable {
         /// An empty text field differs from incomplete in that it never displays an error.
```
tillh-stripe: Nit: This comment needs to be updated now.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedFormViewController.swift
davidme-stripe: I'm not sure if there's a good time to call `prepare()` — the Taptic Engine will only stay active for a few seconds afterwards. Doing it when we initialize the form would probably cause it to just waste a little power and then fall asleep before the user taps the button.

PR #5296, opened 2025-08-11
Crypto Onramp SDK: Provide Merchant Image to Identity SDK

mliberatore:
## Summary
The Identity SDK requires a non-optional `brandLogo` image. This PR exposes the `merchantLogoUrl` and a new `async` `throwing` image download method on `DownloadManager` to accomplish the download while utilizing existing analytics error logging and image caching. 

The reason we need a separate download method is that the existing `public` one returns an image synchronously, and updates the `updateHandler` closure later, which is incompatible with our usage in providing a single, non-optional image to the Identity SDK on initialization.

## Motivation
To show the merchant image within the Identity SDK. See the screenshot below.

## Testing
1. Manually tested by using the Crypto Onramp example project, authenticating, tapping the Verify Identity button, and then tapping "Proceed" from the Identity test UI:

Simulator Screenshot - iPhone 16 Pro - 2025-08-11 at 11 31 17

2. Repeated the same test but forced `merchantLogoUrl` to return `nil` to test the fallback image:
Simulator Screenshot - iPhone 16 Pro - 2025-08-11 at 12 17 21
Simulator Screenshot - iPhone 16 Pro - 2025-08-11 at 12 17 26

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/Helpers/DownloadManager.swift
```diff
@@ -94,15 +109,7 @@ extension DownloadManager {
...
-        }
-    }
-
-    private func downloadImageAsync(url: URL, placeholder: UIImage, updateHandler: UpdateImageHandler) async {
-        let image = await downloadImage(url: url, placeholder: placeholder)
-        // Only invoke the `updateHandler` if the fetched image differs from the placeholder we already vended
-        if !image.isEqualToImage(image: placeholder) {
-            updateHandler(image)
+            throw error
         }
```
mliberatore, line 113: It's hard to follow in the diff, but the way the placeholder here used to work, is `downloadImage(url: url, placeholder: placeholder)` would return the `placeholder` in the case of an error caught in the `do`-`catch`. Then, in the deleted code above, the returned image is compared against the placeholder and `updateHandler` is called if it differs.

You’ll notice we no longer make this comparison since now, the renamed `downloadImageSkippingCacheRead` no longer takes / returns the placeholder, and throws in the case of error. So the `private` API here either returns the downloaded image or `throws`, and we never end up in a situation where we need to know whether the returned image is the placeholder or not.

PR #5304, opened 2025-08-11
Embedded - Don't show the change button for the disclosure style

gbirch-stripe:
## Summary
The disclosure style doesn't support a visual "selected" state. Therefore, we omit the change button to avoid visual confusion

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-3890

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -461,19 +449,23 @@ extension EmbeddedPaymentElement: EmbeddedFormViewControllerDelegate {
...
+        if shouldShowChangeButton {
+            embeddedPaymentMethodsView.selectedRowButton?.addChangeButton()
         }
-        // Show change button for new PMs that have a valid form
-        let shouldShowChangeButton: Bool = {
-            if case .new = type, selectedFormViewController != nil {
-               return true
-            }
-            return false
-        }()
```
yuki-stripe, line 474: Where did this logic go?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -461,19 +449,23 @@ extension EmbeddedPaymentElement: EmbeddedFormViewControllerDelegate {
...
+              case .new = type,
+              selectedFormViewController != nil else {
+            embeddedPaymentMethodsView.selectedRowChangeButtonState = (false, nil)
+            return
+        }
+
+        // Don't show the change button for the disclosure style
+        let shouldShowChangeButton = configuration.appearance.embeddedPaymentElement.row.style != .flatWithDisclosure
+        if shouldShowChangeButton {
+            embeddedPaymentMethodsView.selectedRowButton?.addChangeButton()
```
yuki-stripe, line 465: Since the tests pass and I think we have decent coverage, this probably all works, but I'm not yet understanding the change here. Like previously we were not unconditionally calling `addChangeButton()` were we? We only did that if `changeButtonState.shouldShowChangeButton == true`? 

I would think we shouldn't add the change button if we're hiding it.

yuki-stripe (commented):
The existing change button code (that I'm responsible for) is not so good, thanks/sorry for dealing with it!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
gbirch-stripe: Lines 456-457 in the new version

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
gbirch-stripe: Previously, whether or not we show the change button and whether or not we set a new sublabel were coupled. You can see in the old code that anytime `getChangeButtonState()` returns `shouldShowChangeButton = false` we treat it as if it had returned `sublabel = nil` as well. These now need to be uncoupled, though, since we now have a situation where we want to disable the change button but still want to set a new sublabel. 

We are still not unconditionally calling `addChangeButton()` — we only do it if `shouldShowChangeButton = true` and we don't exit out early from the `guard`

yuki-stripe: Ahh I didn't pay enough attention to the guard, thanks!

PR #5250, opened 2025-08-04
opensCardScannerAutomatically API implementation

gbirch-stripe:
## Summary
- Added opensCardScannerAutomatically API to PaymentSheet (including Flow Controller), CustomerSheet, and Embedded
- Changed CustomerSheet to preserve form data when reselected
- Added analytics collection to CustomerSheet playground
- Refactored logic/delegation around card scanner to make code clearer/more readable

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-3771
https://docs.google.com/document/d/1RwHDeiBK2UJN_aBkSk_axzI9jzmmH-dt6LqUEaYvUEA/edit?usp=sharing

## Testing
- UI tests for new config 
- UI tests for direct to card scan flow
- UI tests for CustomerSheet form input preservation

## Changelog
- [Added] `opensCardScannerAutomatically` API to Payment Sheet, Customer Sheet, and Embedded configurations
- [Changed] Customer Sheet caches form inputs

PR #5307, opened 2025-08-12
Restore direct to card scan

gbirch-stripe:
## Summary
- restores change made in https://github.com/stripe/stripe-ios/pull/5250 that was temporarily reverted in https://github.com/stripe/stripe-ios/pull/5270
- bug fix for card scanner being frozen with broken layout when opened and closed caused by duplicate function calls

## Motivation
N/A

## Testing
N/A

## Changelog
- [Added] `opensCardScannerAutomatically` API to Payment Sheet, Customer Sheet, and Embedded configurations
- [Changed] Customer Sheet caches form inputs


CHANGELOG.md
```diff
@@ -1,6 +1,11 @@
 ## x.x.x x-x-x
 
+### CustomerSheet
+* [Added] `opensCardScannerAutomatically` configuration
+* [Changed] Form inputs are cached
```
davidme-stripe: This doesn't make sense as a user-facing changelog entry — can we remove it? I'm also not sure why the form cache changes are related to the card scanner PR

gbirch-stripe: CustomerSheet form cacheing was related to the card scanner maintaining its state (open/closed), which is how it ended up in here. Agreed that it should have been a separate PR though

PR #5309, opened 2025-08-12
Allow filtering supported payment method types in LinkController

mats-stripe:
## Summary

This adds filtering the supported payment method types in `LinkController`. It launches right into the financial connections flow in the scenario where we've filtered to bank accounts, and no bank accounts are available. I also added toggles in the example view to test this out.

## Motivation

https://docs.google.com/document/d/1NL4a6Cbm5GdlWudaeL8dq8y0xZMvyIxv2CuaPbLYaXI/edit?usp=sharing

## Testing

https://github.com/user-attachments/assets/260c0cec-55b0-4656-b804-28a7e1eaf4e1

No existing bank accounts:

https://github.com/user-attachments/assets/0a6905e3-5f1d-4c0f-8a2e-472270a15a31

## Changelog

N/a

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController-WalletViewModel.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController-WalletViewModel.swift
@@ -174,10 +174,6 @@ extension PayWithLinkViewController {
         }
 
         var confirmButtonStatus: ConfirmButton.Status {
-            if selectedPaymentMethod == nil {
-                return .disabled
-            }
-
             if !selectedPaymentMethodIsSupported {
                 // Selected payment method not supported
                 return .disabled


StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController-WalletViewModel.swift
```diff
@@ -170,6 +170,10 @@ extension PayWithLinkViewController {
         }
 
         var confirmButtonStatus: ConfirmButton.Status {
+            if selectedPaymentMethod == nil {
+                return .disabled
+            }
```
tillh-stripe: The next check for `selectedPaymentMethodIsSupported` already covers this, no?

tillh-stripe (dismissed):

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController-WalletViewModel.swift
mats-stripe: Looks like it should yeah, not sure why I thought otherwise. I'll remove this change

PR #5316, opened 2025-08-13
Don't show loading indicator in Link OTP dialog

tillh-stripe:


StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Verification/LinkVerificationViewController.swift
```diff
@@ -124,17 +124,11 @@ final class LinkVerificationViewController: UIViewController {
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
 
-        activityIndicator.startAnimating()
```
mats-stripe, line 127: I think we can remove `activityIndicator` completely now

mats-stripe (commented):
Tested it out and it LGTM. One small change probably needed before merging:

PR #5319, opened 2025-08-13
Show an alert on broken constraints in PaymentSheet test app

yuki-stripe:
## Summary
In the PaymentSheet test app, we now show an alert if there's a broken autolayout constraint. There's also a way to ignore broken constraints, which I've used to ignore some existing ones that I couldn't easily fix.

## Motivation
Try to prevent our UI from having broken constraints.

## Testing
Manually tested - alert appears as expected, ignores ignored constraints as expected.

Here's an example of what it looks like when you see an alert:
https://github.com/user-attachments/assets/5d6fa6f4-d7a1-42fb-b200-70a0ef8a0d3b

## Changelog
Not user facing

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/AppDelegate+BrokenConstraints.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/AppDelegate+BrokenConstraints.swift
@@ -45,7 +45,7 @@ extension AppDelegate {
             return
         }
         // In other cases a broken constraint is generic, and we need to look at the hierarchy to avoid ignoring more than we intend
-        let recursiveDescription = window?.value(forKey: "recursiveDescription") as! String
+        let recursiveDescription = window?.value(forKey: "recursiveDescription") as? String ?? ""
         if
             recursiveDescription.contains("STP_Internal_LinkSheetNavigationBar") && constraint.debugDescription.contains("UISV-spanning-boundary") ||
             recursiveDescription.contains("STP_Internal_LinkSheetNavigationBar") && constraint.debugDescription.contains("UIButton") ||


Example/PaymentSheet Example/PaymentSheet Example/AppDelegate+BrokenConstraints.swift
```diff
@@ -0,0 +1,71 @@
...
+            "STP_Internal_LinkMoreInfoView", // https://jira.corp.stripe.com/browse/RUN_MOBILESDK-4562
+
+            "_UIRemoteKeyboardPlaceholderView", // Broken constraints in Apple's keyboard; unclear how it's our fault
+            "SystemInputAssistantView", // Same as ^
+        ]
+        guard !ignoredBrokenConstraints.contains(where: { constraint.debugDescription.contains($0) }) else {
+            return
+        }
+        // In other cases a broken constraint is generic, and we need to look at the hierarchy to avoid ignoring more than we intend
+        let recursiveDescription = window?.value(forKey: "recursiveDescription") as! String
```
wooj-stripe: any reason not to put this in the guard statement using `as? String` instead of the proposed `as! String`?

Example/PaymentSheet Example/PaymentSheet Example/AppDelegate+BrokenConstraints.swift
```diff
@@ -0,0 +1,71 @@
...
+//  PaymentSheet Example
+//
+//  Created by Yuki Tokuhiro on 8/13/25.
+//
+@_spi(STP) import StripeUICore
+import UIKit
+
+extension AppDelegate {
+    /// - Note: Only call this **once**!
+    func catchBrokenConstraints() {
```
wooj-stripe, line 12: should we add something (static var/mutex, singleton, etc) ensuring it's been called once?

Example/PaymentSheet Example/PaymentSheet Example/AppDelegate+BrokenConstraints.swift
```diff
@@ -0,0 +1,71 @@
...
+        }
+        let alert = UIAlertController(title: "Broken constraint!", message: "\(constraint)\nPlease fix it or file a bug!", preferredStyle: .alert)
+        alert.addAction(UIAlertAction(title: "OK", style: .default))
+        window?.rootViewController?.findTopMostPresentedViewController().present(alert, animated: true)
+    }
+}
+
+extension UIView {
+    @objc func willBreakConstraint(_ engine: Any, _ constraint: NSLayoutConstraint, _ conflict: Any) {
+        willBreakConstraint(engine, constraint, conflict) // swizzled, will call original impl instead
```
wooj-stripe, line 65: Seems like the comment should say "first" instead of "instead" -- since our function is being called fist, which then calls the original impl, and then posts a notification.

Example/PaymentSheet Example/PaymentSheet Example/AppDelegate+BrokenConstraints.swift
yuki-stripe: Ah I'll default it to an empty string if it's nil. I don't want to ignore broken constraints if this line stops working.

Example/PaymentSheet Example/PaymentSheet Example/AppDelegate+BrokenConstraints.swift
yuki-stripe: Hrm how would those options ensure it's only called once? I could keep track of how many times it's called and assert? Feels like maybe overkill 🤷

Example/PaymentSheet Example/PaymentSheet Example/AppDelegate+BrokenConstraints.swift
yuki-stripe: This line _appears_ to call the very method it's inside, leading to infinite recursion.  The comment explains that the method is swizzled, so calls to `willBreakConstraint` will **instead** call `engine:willBreakConstraint:dueToMutuallyExclusiveConstraints:` 🪄 .

yuki-stripe:
the public api bot is broken, just adding the label to ignore it.

wooj-stripe (approved):

PR #5325, opened 2025-08-14
Fix Link buggy bottom layout

yuki-stripe:
## Summary
TODO

## Testing
TODO

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController-WalletViewController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController-WalletViewController.swift
@@ -181,7 +181,7 @@ extension PayWithLinkViewController {
                 containerView.addArrangedSubview(cancelButton)
             }
 
-            contentView.addAndPinSubview(containerView)
+            contentView.addAndPinSubview(containerView, insets: .insets(bottom: 35))
 
             // If the initially selected payment method is not supported, we should automatically
             // expand the payment picker to hint the user to pick another payment method.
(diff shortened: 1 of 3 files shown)


StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController-WalletViewController.swift
```diff
@@ -181,7 +181,7 @@ extension PayWithLinkViewController {
                 containerView.addArrangedSubview(cancelButton)
             }
 
-            contentView.addAndPinSubview(containerView)
+            contentView.addAndPinSubview(containerView, insets: .insets(bottom: 35))
```
mats-stripe, line 184: Out of curiosity, where does 35 come from? Wonder if we can use a value in `LinkUI`, like `extraLargeContentSpacing`, which is 32

yuki-stripe: Purely to match the existing UI.  I'm happy to use `extraLargeContentSpacing` or whatever you'd like just lmk @mats-stripe

tillh-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController-WalletViewController.swift
mats-stripe: This is fine as-is!

mats-stripe (approved):

PR #5326, opened 2025-08-14
Allow filtering billing address countries

davidme-stripe:
## Summary
Add a new `allowedCountries` API for only allowing certain countries for billing addresses. When set, we'll also filter out any SPMs with the selected countries.

## Motivation
Some users want to only allow entering cards (or using saved cards) from specific countries.

## Testing
CI, manually in PaymentSheet Playground

## Changelog
Added


Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlaygroundSettings.swift
```diff
@@ -432,6 +432,40 @@ struct PaymentSheetTestPlaygroundSettings: Codable, Equatable {
...
+        var countries: Set<String> {
+            switch self {
+            case .all:
+                return []  // Empty set means all countries
+            case .usOnly:
+                return ["US"]
+            case .northAmerica:
+                return ["US", "CA", "MX"]
+            case .fiveCountries:
+                return ["US", "CA", "GB", "AU", "DE"]
```
wooj-stripe, line 452: I don't feel strongly here, but I was expecting EU5 after i read "five countries": "FR", "DE", "IT", "ES", "GB"

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory+UPI.swift
```diff
@@ -17,7 +17,7 @@ extension PaymentSheetFormFactory {
     func makeUPI() -> FormElement {
         let contactInformationElement = makeContactInformationSection(nameRequiredByPaymentMethod: false, emailRequiredByPaymentMethod: false, phoneRequiredByPaymentMethod: false)
         let billingAddressElement = configuration.billingDetailsCollectionConfiguration.address == .full
-            ? makeBillingAddressSection(countries: nil)
+            ? makeBillingAddressSection(countries: configuration.billingDetailsCollectionConfiguration.allowedCountries.isEmpty ? nil : Array(configuration.billingDetailsCollectionConfiguration.allowedCountries))
```
tillh-stripe: Feels like we could put this particular expression into an extension property like `billingDetailsCollectionConfiguration.allowedCountriesOrAll` (or some better name).

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/PaymentDetails.swift
```diff
@@ -79,9 +80,35 @@ extension ConsumerPaymentDetails {
...
+        guard !allowedCountries.isEmpty else {
+            // No filtering required
+            return true
+        }
+
+        switch details {
+        case .card:
+            // If the merchant is filtering, only allow cards with a billing country
+            if let country = billingAddress?.countryCode {
+                return allowedCountries.contains(country)
```
wooj-stripe, line 100: Seems like we should have some initial validation on init(?) so that we don't have to do case sensitive comparison (e.g. "us" vs "US")

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory+Card.swift
```diff
@@ -79,11 +79,14 @@ extension PaymentSheetFormFactory {
         let shouldIncludePhone = configuration.billingDetailsCollectionConfiguration.phone == .always
 
         let billingAddressSection: PaymentMethodElementWrapper<AddressSectionElement>? = {
+            let countries = configuration.billingDetailsCollectionConfiguration.allowedCountries.isEmpty
+                ? nil
+                : Array(configuration.billingDetailsCollectionConfiguration.allowedCountries)
```
wooj-stripe, line 84: This logic seems to be re-used many times -- seems like we should unify this logic somehow?

PR #5336, opened 2025-08-14
Tweak ordering in `WalletButtonsView`

tillh-stripe:
## Summary

This pull request updates the `WalletButtonsView` so that ordering is aligned between platforms.


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/WalletButtonsView/WalletButtonsView.swift
```diff
@@ -12,7 +12,6 @@ import WebKit
     enum ExpressType: Hashable {
         case applePay
         case link
-        case linkInlineVerification(PaymentSheetLinkAccount)
```
tillh-stripe, line 15: This is currently not used. Let's add it again if needed.

davidme-stripe (approved):

PR #5345, opened 2025-08-15
Load playground appearance from UserDefaults

gbirch-stripe:
## Summary
Small fix

## Motivation
Restore functionality

## Testing
N/A

## Changelog
N/A


wooj-stripe:
I tried to set the background, then loaded payment sheet, then hit back, reloaded the playground and my settings weren't preserved?

gbirch-stripe:
Issue was that SwiftUI was reinitializing the `PlaygroundController` and `CustomerSheetTestPlaygroundController` each time the pages were opened, but still with the same values fetched only once from the root view, resulting in stale values being used.

PR #5348, opened 2025-08-15
Fix dead code check

porter-stripe:
## Summary
- With recent Github deprecation on macOS 15 image we need to manually select an Xcode version


porter-stripe:
[ec5784e](https://github.com/stripe/stripe-ios/pull/5348/commits/ec5784ec643984e31d104be77eb58cf8f2556b78) showed a successful run of the dead code check

mats-stripe (approved):

PR #5351, opened 2025-08-18
Crypto Onramp SDK: Uses Custom DateOfBirth for KycInfo

mliberatore:
## Summary
Favors a date of birth to be specified in terms of day, month, year.

## Motivation
Discussion [here](https://lickability.slack.com/archives/C094P3BRBL1/p1755270675048949).

## Testing
1. Ran the updated test.
2. Ran through the KycInfo collection to ensure the date components from the UI were used properly:

CleanShot 2025-08-18 at 09 06 27

## Changelog
N/A

### DIFF
--- a/StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/KycInfo.swift
+++ b/StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/KycInfo.swift
@@ -12,6 +12,11 @@ import Foundation
 public struct KycInfo: Equatable {
 
     /// Represents a fixed date using simple components (day, month, year).
+    /// For example, March 31st, 1975 would be:
+    ///
+    /// ```
+    /// DateOfBirth(day: 31, month: 3, year: 1975)
+    /// ```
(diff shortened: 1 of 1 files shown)


StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/KycInfo.swift
```diff
@@ -11,6 +11,30 @@ import Foundation
 @_spi(CryptoOnrampSDKPreview)
 public struct KycInfo: Equatable {
 
+    /// Represents a fixed date using simple components (day, month, year).
+    public struct DateOfBirth: Encodable, Equatable {
```
mats-stripe: Super minor, but could we add a simple example? Does something like this make sense?

```suggestion
    /// Represents a fixed date using simple components (day, month, year).
    /// For example, March 31st, 1975 would be:
    ///
    /// ```
    /// DateOfBirth(day: 31, month: 3, year: 1975)
    /// ```
    public struct DateOfBirth: Encodable, Equatable {
```

Issue #5229, opened 2025-07-29
[BUG] iOS SDK Face Detector bug


cjmisenas-stripe:
## Summary

User reported they were able to capture document images but when trying to capture a selfie, they see the following error. We saw a rise in this specific error (specifically this [line](https://github.com/stripe/stripe-ios/blob/5541e9b3/StripeIdentity/StripeIdentity/Source/NativeComponents/Coordinators/VerificationSheetFlowController.swift#L692-L696)) on minor versions 23.29.* of the SDKs seemingly related to CoreML compatibility of the ML models. We’re also seeing lower submit rate for selfie verifications likely related to this issue. It looks like this appears to happen on first launch of this sheet from a particular app install, but not on future attempts.

<img width="1290" height="2796" alt="Image" src="https://github.com/user-attachments/assets/fa397392-1b9f-45f3-b341-d98666fb6df6" />

## SDK version

23.29.*

bcapps:
Looks like this was resolved here https://github.com/stripe/stripe-ios/pull/4846

PR #5363, opened 2025-08-19
[Identity] Use caches direcotry not temp directory for downloaded ML models in Identity

Pearapps:
## Summary

Rules out temporary directory being prematurely cleared leading to errors as described in [#5229](https://github.com/stripe/stripe-ios/issues/5229)

## Motivation

It is possible that the IO error we are seeing is because these models are not accessible, this is _possible_ in theory because of them being in the temporary directory. I do not know how likely that is, but this will rule it out.

## Testing

I ran through the playground example in the identity example app.

I tried an existing install, and new install -- using both caches directory code paths (although both tests on iOS 18+)

### DIFF
--- a/StripeIdentity/StripeIdentity/Source/NativeComponents/ML/IdentityMLModelLoader.swift
+++ b/StripeIdentity/StripeIdentity/Source/NativeComponents/ML/IdentityMLModelLoader.swift
@@ -89,14 +89,18 @@ final class IdentityMLModelLoader: IdentityMLModelLoaderProtocol {
         // Since the models are unlikely to be used after the user has finished
         // verifying their identity, cache them to a temp directory so the
         // system will delete them when it needs the space.
-        let tempDirectory = URL(
-            fileURLWithPath: NSTemporaryDirectory(),
-            isDirectory: true
-        )
+        
+        let cachesDirectory: URL
+        if #available(iOS 16.0, *) {
+            cachesDirectory = URL.cachesDirectory
+        } else {
+            let paths = FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask)
+            cachesDirectory = paths.first ?? URL(fileURLWithPath: NSTemporaryDirectory())
+        }
 
         // Create a name-spaced subdirectory inside the temp directory so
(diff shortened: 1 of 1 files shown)


StripeIdentity/StripeIdentity/Source/NativeComponents/ML/IdentityMLModelLoader.swift
```diff
@@ -89,14 +89,18 @@ final class IdentityMLModelLoader: IdentityMLModelLoaderProtocol {
...
-            fileURLWithPath: NSTemporaryDirectory(),
-            isDirectory: true
-        )
+
+        let cachesDirectory: URL
+        if #available(iOS 16.0, *) {
+            cachesDirectory = URL.cachesDirectory
+        } else {
+            let paths = FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask)
+            cachesDirectory = paths.first ?? URL(fileURLWithPath: NSTemporaryDirectory())
```
luisv-stripe: Is the `isDirectory: true` param ok to remove?

StripeIdentity/StripeIdentity/Source/NativeComponents/ML/IdentityMLModelLoader.swift
```diff
@@ -89,14 +89,18 @@ final class IdentityMLModelLoader: IdentityMLModelLoaderProtocol {
...
-        let tempDirectory = URL(
-            fileURLWithPath: NSTemporaryDirectory(),
-            isDirectory: true
-        )
+
+        let cachesDirectory: URL
+        if #available(iOS 16.0, *) {
+            cachesDirectory = URL.cachesDirectory
+        } else {
+            let paths = FileManager.default.urls(for: .cachesDirectory, in: .userDomainMask)
```
luisv-stripe, line 97: Were you able to test this code branch? Seems like you would need something that is < iOS 16.0

Pearapps: I did test that branch, but I did by commenting out the other code and running on iOS 18

PR #5378, opened 2025-08-20
Allow changing payment methods in the Onramp example app

mats-stripe:
## Summary

Allows changing payment methods after one has been selected in the Onramp example app. Once a payment token is created, you can no longer change payment methods.

## Motivation

💅 

## Testing

https://github.com/user-attachments/assets/7a578359-fd3c-4404-82d3-ca3819a7f63a

## Changelog

N/a


mliberatore (approved):
:+1:

PR #5393, opened 2025-08-22
Correct RowButton Floating button corner radius

wooj-stripe:
## Summary
Updates the way we calculate corner radius for corner radiuses for flowing button

## Motivation
Pushing the corner radius to 16 seems result in .... these buttons not matching the WalletHeaderButtons

## Testing
Added screenshot test

|Before | After|
| -------- | -------|
| image |  image |


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/ShadowedRoundedRectangleView.swift
```diff
@@ -30,6 +30,18 @@ class ShadowedRoundedRectangle: UIView {
         }
     }
 
+    var effectiveCornerRadius: CGFloat {
+        // Matches the formula used by `PKPaymentButton` for calculating
+        // the effective corner radius. The effective corner radius is snapped
+        // to half the button's height if the corner radius is
+        // greater or equals than approx. 1/3 of the height (`threshold`).
+        let threshold = 0.32214
```
joyceqin-stripe, line 38: Where did this come from? Could you provide a link?

wooj-stripe: https://github.com/stripe/stripe-ios/blob/a3dcff78ec7c054618d3d2ad2e06cf8ab5c835b7/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/PayWithLinkButton.swift#L350-L361

PR #5396, opened 2025-08-23
Initial iOS 26 changes behind config flag

wooj-stripe:
## Summary
Adds some inial changes for iOS26 behind a config flag

## Motivation
iOS 26 development on master.

## Testing
Will add some tests, and targets to run on iOS26.. i hope?

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
@@ -308,7 +308,7 @@ struct PaymentSheetTestPlayground: View {
         Binding<PaymentSheetTestPlaygroundSettings.EnableIOS26Changes> {
             return playgroundController.settings.enableIOS26Changes
         } set: { newValue in
-            NewDesignDetector.allowNewDesign = newValue == .on
+            LiquidGlassDetector.allowNewDesign = newValue == .on
             playgroundController.appearance = PaymentSheet.Appearance()
             playgroundController.settings.enableIOS26Changes = newValue
         }
--- a/StripePaymentSheet/StripePaymentSheet.xcodeproj/project.pbxproj
+++ b/StripePaymentSheet/StripePaymentSheet.xcodeproj/project.pbxproj
@@ -105,7 +105,7 @@
 		317B56812E42C00F005EAFB1 /* PaymentSheetFormFactory+CardEmailPhoneFieldsTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 317B56802E42C00F005EAFB1 /* PaymentSheetFormFactory+CardEmailPhoneFieldsTest.swift */; };
 		3190006D2DFD32BF00B84F6C /* ECEBridgeTypes.swift in Sources */ = {isa = PBXBuildFile; fileRef = 3190006C2DFD32BF00B84F6C /* ECEBridgeTypes.swift */; };
 		31AD3BE72B0C2D080080C800 /* UIApplication+StripePaymentSheet.swift in Sources */ = {isa = PBXBuildFile; fileRef = 31AD3BE62B0C2D080080C800 /* UIApplication+StripePaymentSheet.swift */; };
-		31AF62882E414CFD00C7C6A7 /* NewDesignDetector.swift in Sources */ = {isa = PBXBuildFile; fileRef = 31AF62872E414CFD00C7C6A7 /* NewDesignDetector.swift */; };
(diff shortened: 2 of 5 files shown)


wooj-stripe:
The Public interface, imho, is a false positive.

Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
```diff
@@ -1045,7 +1045,11 @@ extension PlaygroundController: STPAnalyticsClientDelegate {
 
 extension PlaygroundController {
     func serializeSettingsToNSUserDefaults() {
-        let settingsData = try! JSONEncoder().encode(settings)
+        // Never save changes for iOS26 since we have to set allowNewDesign based on a flag that isn't ready during boot time.
+        var settingsWithoutiOS26 = settings
+        settingsWithoutiOS26.enableIOS26Changes = .off
```
yuki-stripe, line 1050: is this still needed?  I'm not seeing why we couldn't persist this.

StripePaymentSheet/StripePaymentSheet/Source/Helpers/NewDesignDetector.swift
```diff
@@ -0,0 +1,29 @@
...
+//  StripePaymentSheet
+//
+//  Created by David Estes on 8/4/25.
+//
+
+import Foundation
+
+// Our SDK runs inside a user's app, we can't explicitly opt into or out of the new "Liquid Glass" design.
+// Instead, we'll do our best to detect which design to use, and adjust the default UI spacing and icons accordingly.
+@_spi(STP) public class NewDesignDetector {
```
yuki-stripe: nit: for this class and the properties, "NewDesign" is so generic, can we replace w/ LiquidGlass or something? So `LiquidGlassDetector`, `LiquidGlassDetector.isLiquidGlassEnabled` (or maybe just `LiquidGlassDetector.isEnabled`)

StripePaymentSheet/StripePaymentSheet/Source/Helpers/NewDesignDetector.swift
```diff
@@ -0,0 +1,29 @@
...
+    @_spi(STP) public static var allowNewDesign: Bool = false
+    @_spi(STP) public static var newDesignEnabled: Bool {
+        // If the app was built with Xcode 26 or later (which includes Swift compiler 6.2)...
+        #if compiler(>=6.2)
+        // And we're running on iOS 26 or later...
+        if #available(iOS 26.0, *) {
+            // And the app hasn't opted out of the new design...
+            if !(Bundle.main.infoDictionary?["UIDesignRequiresCompatibility"] as? Bool ?? false) && allowNewDesign {
```
yuki-stripe: What's the purpose of allowNewDesign? I thought it was maybe a test-only way to override `UIDesignRequiresCompatibility` for convenience but I don't think that's what this line would do exactly, maybe it should be like this?
```
let overrideUIDesignRequiresCompatibility: Bool?
...
let uiDesignRequiresCompatibility = Bundle.main.infoDictionary?["UIDesignRequiresCompatibility"] as? Bool ?? false
// override
if let overrideUIDesignRequiresCompatibility { return overrideUIDesignRequiresCompatibility }
else { return !uiDesignRequiresCompatibility }
```

Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
wooj-stripe: If this is saved to `.on`, playground shows that it is enabled, but the Playground is instantiated with .default (which doesn't properly set `NewDesignDetector::allowNewDesign`). As a workaround, i decided it would be easier to just always default this to .off.

StripePaymentSheet/StripePaymentSheet/Source/Helpers/NewDesignDetector.swift
wooj-stripe: Sure, we aren't adopting a ton of liquidglass items, but sure, lets go with `LiquidGlassDetector`

StripePaymentSheet/StripePaymentSheet/Source/Helpers/NewDesignDetector.swift
wooj-stripe: allowNewDesign stops changes from getting out to production, and is ONLY expected to be use for us in  the next couple of days.

If we didn't have this flag, and users don't opt out, and then we do a release, then users on iOS26 devices will get a half baked design for payment sheet.

yuki-stripe (approved):

PR #5404, opened 2025-08-25
Show Link reauthentication inline if applicable

tillh-stripe:
## Summary

This pull request updates how we re-authenticate Link consumers if their session has expired before re-opening the wallet.

Previously, we used the `retryingOnAuthError` method to have consumers verify their account when we encountered an expired session while loading the wallet. This caused a suboptimal UI (a loading dialog on top of the loading sheet) and could result in an endless loading state due to a race condition.

With this pull request, we no longer handle re-authentication for each API call separately. Instead, we catch any authentication-related errors in `loadAndPresentWallet` and navigate to the verification screen _after_ creating a new consumer session, which causes both payment details and shipping addresses to be fetched with the new consumer session client secret.


Tests/ReferenceImages_64/StripePaymentSheetTests.LinkVerificationViewSnapshotTests/testInlineLogin@3x.png
mats-stripe, line 1: I can't tell what changed here

PR #5406, opened 2025-08-25
opensCardScannerAutomatically API/implementation

gbirch-stripe:
## Summary
- implementation of opensCardScannerAutomatically API

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-3771
https://docs.google.com/document/d/1RwHDeiBK2UJN_aBkSk_axzI9jzmmH-dt6LqUEaYvUEA/edit?usp=sharing

## Testing
N/A

## Changelog
- [Added] opensCardScannerAutomatically API to Payment Sheet, Customer Sheet, and Embedded configurations


CHANGELOG.md
```diff
@@ -2,13 +2,15 @@
...
+* [Added] `opensCardScannerAutomatically` configuration
 
 ### PaymentSheet
 * [Added] Added new `presentPaymentOptions` APIs for UIKit and SwiftUI with a `didCancel` bool in the completion handler, making it possible to differentiate between a user closing FlowController and selecting a payment option. ([#5202](https://github.com/stripe/stripe-ios/pull/5202))
 * [Added] Added a new `allowedCountries` API to PaymentSheetConfiguration, enabling users to only allow billing addresses in certain countries.
 * [Fixed] Fixed Dynamic Actor Isolation crash in SwiftUI when using completion handlers with Swift 6 or Dynamic Actor Isolation feature flag enabled. Thanks @BrentMifsud! ([#5269](https://github.com/stripe/stripe-ios/issues/5269))
 * [Changed] The form will now highlight incomplete fields when the user taps a disabled "Confirm" button.
 * [Added] `STPError.localizedUserMessage(forErrorCode:)` will translates common Stripe API error codes to localized strings.
+* [Added] `opensCardScannerAutomatically` configuration
```
wooj-stripe: i suspect this is out of date now.

CLAUDE.md
```diff
@@ -71,6 +71,10 @@ xcodebuild test -scheme StripePaymentSheet -workspace Stripe.xcworkspace -destin
 - **Record network tests**: Use `AllStripeFrameworks-NetworkRecordMode` scheme (will fail while recording) and a specific test case
 - **Run network tests**: Use `AllStripeFrameworks` scheme to verify recorded network tests
 
+### UI Tests
+- **Run UI tests**: Use `PaymentSheet Example` scheme with target `PaymentSheetUITest`
+- **Run specific UI test class**: `-only-testing:PaymentSheetUITest/YourTestClassName`
+
```
wooj-stripe, line 77: Probably should be in a different commit?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/CardSection/CardSectionWithScannerView.swift
```diff
@@ -30,22 +30,36 @@ final class CardSectionWithScannerView: UIView {
...
         self.analyticsHelper = analyticsHelper
         self.linkAppearance = linkAppearance
         super.init(frame: .zero)
         installConstraints()
+
+        if opensCardScannerAutomatically {
+            cardScanButton.alpha = 0
+        } else {
+            cardScanningView.isHidden = true
+        }
```
wooj-stripe: This feels inconsistent because:
1. Bodies of if/else are modifying two different parameters -- one modifies cardScanButton, and the other modifies cardScanningView.
2. Inconsistent way of hiding elements -- in one call, we use isHidden, and in another, we set alpha == 0.

Separate comment, (not having to do with the inconsistent comment above):
We should avoid using isHidden. If something is hidden, and we set `isHidden` to true again, we tend to get very unreliable behavior.  This is why we've created `setHiddenIfNecessary`

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/CardSection/CardSectionWithScannerView.swift
```diff
@@ -89,9 +103,22 @@ final class CardSectionWithScannerView: UIView {
...
+        // If we leave the screen or an input field is focused, we close the scanner
         cardScanningView.stop()
         return super.resignFirstResponder()
     }
+
+    override func willMove(toWindow newWindow: UIWindow?) {
+        // We wait until we are added to the screen to start the scanner instead of at initialization
+        // If cardScanningView.start() is called when it is already started, nothing will happen
+        // The opensCardScannerAutomatically check is redudant since this should only apply in that case,
+        //    but it adds a bit of extra safety. This can be removed in the future.
```
wooj-stripe, line 115: When will we know that it can be removed in the future? Why not just remove it now?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/CardScanningView.swift
```diff
@@ -121,12 +121,6 @@ class CardScanningView: UIView, STPCardScannerDelegate {
         return button
     }()
 
-    private lazy var containerView: UIView = {
-        let containerView = UIView()
-        containerView.translatesAutoresizingMaskIntoConstraints = false
-        return containerView
-    }()
-
```
wooj-stripe, line 129: Curious as to the reason why we got rid of the containerView?

CLAUDE.md
gbirch-stripe: I can separate it out if you prefer. I've just been having Claude update its own rules for things it works on

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/CardSection/CardSectionWithScannerView.swift
gbirch-stripe: 1. I'm not really sure how to avoid this. We could add `cardScanningView.isHidden = false` to the first case and `cardScanButton.alpha = 1` to the second, but I think this is more confusing because those are in fact already the values.
2. The `cardScanButton` uses alpha because that's what needs to animate in and out, and `alpha = 0` disables user interaction for free as well, so it does everything it needs. `cardScanningView` needs to actually be removed for its animations, so we use that for it. 

In general with this PR I'm trying to leave the existing card scan implementation as-is to minimize risk, and `cardScanningView.isHidden = true` is how we were previously initializing it. This use also only runs at initialization, when `isHidden` is never already true

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/CardScanningView.swift
gbirch-stripe: Yes, the containerView was causing autolayout issues when the view initializes with the card scanner already open. It was originally there for convenience for the bottom constraints (so that we only have to lower the priority on one constraint), but we don't believe it was serving any deeper purpose

CHANGELOG.md
gbirch-stripe: fixed

CLAUDE.md
wooj-stripe: it's fine, lets leave it.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/CardSection/CardSectionWithScannerView.swift
wooj-stripe: I see.. hmm i think at the very least, we should change to using `setHiddenIfNecessary`

wooj-stripe (approved):

PR #5412, opened 2025-08-26
Update header for Link signup opt-in feature

tillh-stripe:
## Summary

This pull request adds a new string for the `signupOptIn` mode for Link inline signup.


tillh-stripe:
Skipping translations as this feature is limited to just one merchant who wants this change immediately without awaiting translations.

mats-stripe (approved):

PR #5418, opened 2025-08-27
Make payment method rows on vertical 64px high, and capsule like

wooj-stripe:
## Summary
Replaces look/feel of vertical payment buttons in paymentSheet/Embedded for iOS26

## Motivation
iOS26!

## Testing
Manual tested embedded floating button and vertical button w/ payment sheet.

Adding capsule style buttons:
|Before | After|
|--------|-------|
|image | image |


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButtonFloating.swift
```diff
@@ -9,13 +9,22 @@ import Foundation
...
 
+protocol SelectableRectangle: UIView {
+    var isSelected: Bool { get set }
+}
+
 /// A `RowButton` subclass that presents floating button style.
 final class RowButtonFloating: RowButton {
     // MARK: - Subviews
 
     /// The shadow view that manages corner radius and shadows and selection border
```
yuki-stripe:
```suggestion
    /// The view that manages corner radius and shadows and selection border
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButtonFloating.swift
```diff
@@ -113,10 +122,33 @@ final class RowButtonFloating: RowButton {
             horizontalStackView.bottomAnchor.constraint(equalTo: bottomAnchor, constant: -insets),
         ])
     }
+    func imageViewLeadingConstant() -> CGFloat {
```
yuki-stripe: nit: this seems like it should be a computed property like `let imageViewLeadingConstant: CGFloat = { .. }()` that way we know it's a constant?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButtonFloating.swift
```diff
@@ -113,10 +122,33 @@ final class RowButtonFloating: RowButton {
...
+        if isFlatWithCheckmarkOrChevronStyle && isDisplayingAccessoryView {
+            heightConstraint?.isActive = false
+            return
+        }
+
+        // If iOS 26+, use the specific height for floating buttons
+        if LiquidGlassDetector.isEnabled {
+            heightConstraint = heightAnchor.constraint(equalToConstant: 64.0)
+            heightConstraint?.isActive = true
+            return
```
yuki-stripe: Hmm row buttons need to have a dynamic height, see the `insets` property.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButtonFloating.swift
```diff
@@ -9,13 +9,22 @@ import Foundation
 @_spi(STP) import StripeUICore
 import UIKit
 
+protocol SelectableRectangle: UIView {
+    var isSelected: Bool { get set }
+}
```
yuki-stripe: nit move this to like `CapsuleRectangle.swift`?

Also not sure if it's any better but if we did this...
```
/// A shadowed rounded rect on < iOS 26 and a CapsuleRectangle on iOS 26+
class SelectableRoundedRectangle: UIView {
  var isSelected: Bool
 // Internally switch between the two
}
```
...it'd mean we only have to do the `if #available....` check here once rather than for each ShadowedRoundedRectangle user.

wooj-stripe: Initially the code started out this way, but code seemed weird given there was no shadows on capsules.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButtonFloating.swift
wooj-stripe: Great catch!!! -- updated the code to accommodate for this.

PR #5420, opened 2025-08-27
Allow updating phone number in onramp SDK

tillh-stripe:
## Summary

This pull request allows updating the Link account's phone number via the crypto onramp SDK.

### DIFF
--- a/Example/CryptoOnramp Example/CryptoOnramp Example/RegistrationView.swift
+++ b/Example/CryptoOnramp Example/CryptoOnramp Example/RegistrationView.swift
@@ -33,7 +33,6 @@ struct RegistrationView: View {
     @State private var showAuthenticatedView: Bool = false
     @State private var registrationCustomerId: String?
     @State private var isRegistrationComplete: Bool = false
-    @State private var didVerifyAfterRegistration: Bool = false
     @State private var showUpdatePhoneNumberSheet: Bool = false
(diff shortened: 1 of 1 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkController.swift
```diff
@@ -467,6 +467,52 @@ import UIKit
...
 
+    /// Updates the phone number for the current Link user.
+    ///
+    /// - Parameter phoneNumber: The phone number of the user. Phone number must be in E.164 format (e.g., +12125551234).
+    /// Throws if an authenticated Link user is not available, phone number format is invalid, or an API error occurs.
+    @_spi(STP) public func updatePhoneNumber(
+        to phoneNumber: String,
+        completion: @escaping (Result<Void, Error>) -> Void
+    ) {
+        guard let linkAccount = LinkAccountContext.shared.account, let consumerSessionClientSecret = linkAccount.currentSession?.clientSecret else {
```
mats-stripe, line 478: I wonder if we should also `guard linkAccount.sessionState == .requiresVerification` here also? If we've already gone through OTP the API will 403, so we might be able to short-circuit the call here. Up to you!

mats-stripe (approved):
LGTM! One suggestion

PR #5424, opened 2025-08-27
Fix card scanner layout bug

gbirch-stripe:
## Summary
Fixed a layout bug that sometimes presents with the following changes:
- don't take a snapshot when the scanner is off screen
- use afterScreenUpdates=false
Without these changes, taking a snapshot causes breakages to the parent stackview's automatic layout constraints, resulting in a layout bug. Either of these changes alone fixes it, but I'm adding both for redundancy / extra safety

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-3972

## Testing
Manual testing

## Changelog
N/A


gbirch-stripe:
Dead code detection and verify public interface don't apply since this is not being merged into `master`

davidme-stripe (approved):

PR #5425, opened 2025-08-27
Don't show Link in WalletButtonsView when disabled client-side

davidme-stripe:
## Summary
Don't show Link if it's disabled by the client-side Link config.

## Testing
In PS Example

## Changelog
N/A


wooj-stripe (approved):

tillh-stripe:
I don't think this is needed anymore following https://github.com/stripe/stripe-ios/pull/5435.

PR #5427, opened 2025-08-27
CryptoOnramp SDK: `performCheckout`  Adjustments

mliberatore:
## Summary
While working on the ReactNative SDK, we discovered that the Android implementation of `performCheckout` has a "canceled" result indicating that the user canceled the checkout (i.e. 3DS was dismissed).

This PR brings iOS in line with this, while also `throw`ing errors in the case failure instead of returning a `.failed` enum value.

## Motivation
To allow the ReactNative SDK implementation of CryptoOnramp to emit the same checkout results on both platforms.

## Testing
Verified that I was still able to make it through the checkout flow.
Simulator Screenshot - iPhone 16 Pro - 2025-08-27 at 16 42 20

## Changelog
N/A


Example/CryptoOnramp Example/CryptoOnramp Example/AuthenticatedView.swift
```diff
@@ -458,24 +458,31 @@ struct AuthenticatedView: View {
         errorMessage = nil
 
         Task {
```
mliberatore, line 460: > [!NOTE]
> Most of this diff (`AuthenticatedView.swift` and `CryptoOnrampCoordinator.swift`) is much more readable with "Hide whitespace" checked in GitHub

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
```diff
@@ -114,11 +114,12 @@ protocol CryptoOnrampCoordinatorProtocol {
     ///     The closure should return the onramp session client secret on success, or throw an Error on failure.
     ///     This closure may be called twice: once initially, and once more after handling any required authentication.
     /// - Returns: A `CheckoutResult` indicating whether the checkout succeeded or failed.
+    /// Throws an error if checkout fails.
```
mats-stripe:
```suggestion
    /// - Returns: A `CheckoutResult` indicating whether the checkout succeeded or was canceled.
    /// Throws if handling required actions fails, or an API error occurs.
```

mats-stripe (commented):
LGTM - just one docstring suggestion. Thanks for making the change!

mats-stripe (approved):

PR #5430, opened 2025-08-27
Make navigation bar compatible with liquid glass

wooj-stripe:
## Summary
* Updates navigation bar which emulates effect w/ liquid glass
* Increases size of nav bar
* Updates nav icons (close/back, edit/close)
* Updates content offset when swapping out content (e.g. tapping on card in flow controller vertical)

## Motivation
iOS26

## Testing
Manual testing -- will need to work on getting in snap shot tests after we validate design.

|before | after|
|-------| -----|
|image|image|


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetVerticalViewController.swift
```diff
@@ -839,7 +839,7 @@ extension PaymentSheetVerticalViewController: VerticalPaymentMethodListViewContr
                 // The payment method form collects user input, display it
                 self.paymentMethodFormViewController = pmFormVC
                 paymentMethodListContentOffsetPercentage = bottomSheetController?.contentOffsetPercentage
-                switchContentIfNecessary(to: pmFormVC, containerView: paymentContainerView, contentOffsetPercentage: 0)
+                switchContentIfNecessary(to: pmFormVC, containerView: paymentContainerView, contentOffsetPercentage: LiquidGlassDetector.isEnabled ? 1 : 0)
```
yuki-stripe: what's this do? I'm reading this as saying it sets the content offset to 100% ie scrolls to the bottom.

yuki-stripe: I might be misunderstanding it all but maybe we need to tweak the code that sets the content offset to account for the nav bar?

wooj-stripe: If we don't have this, then tapping on "add card"'s offset is hardcoded to zero -- Any recommendations on a better place to update the content offset would be appreciated!

wooj-stripe: Hmm I investigated this more -- We really shouldn't need this, since content inset of the scrollview is still 64, so i'm not sure why the contents of add card is getting hidden under the nav bar.

wooj-stripe: Changed the approach to move the contentContainerView down, rather than fussing with scrollOffsets and contentInsets!

PR #5437, opened 2025-08-28
Pass `allow_redisplay` along to FC SDK

tillh-stripe:
## Summary

This pull request makes us pass along the merchant-provided `allow_redisplay` value to the Financial Connections SDK. The Stripe.js implementation will be [updated](https://git.corp.stripe.com/stripe-internal/pay-server/pull/1180553) to then pass this value in the payment method creation call.

### DIFF
--- a/StripeCore/StripeCore/Source/Connections Bindings/HostedAuthUrlBuilder.swift
+++ b/StripeCore/StripeCore/Source/Connections Bindings/HostedAuthUrlBuilder.swift
@@ -85,6 +85,10 @@ import Foundation
 
         parameters.append("launched_by=ios_sdk")
 
+        if let allowRedisplay = elementsSessionContext?.allowRedisplay {
+            parameters.append("allow_redisplay=\(allowRedisplay)")
+        }
+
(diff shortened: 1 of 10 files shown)


StripeCore/StripeCore/Source/Connections Bindings/HostedAuthUrlBuilder.swift
```diff
@@ -85,6 +85,10 @@ import Foundation
 
         parameters.append("launched_by=ios_sdk")
 
+        if let allowRedisplay = elementsSessionContext?.allowRedisplay {
+            parameters.append("allow_redisplay=\(allowRedisplay)")
```
mats-stripe, line 89: This is also used when FC is launched in a webview by the SDK, so not just FC Lite. I'm guessing that's also expected / desired here, but just calling it out

tillh-stripe: Yep, that's desired!

PR #5438, opened 2025-08-28
Fix divider lines becoming 0 when borderWidth is 0

yuki-stripe:
## Summary
We have a rough edge in the appearance API when `appearance.borderWidth = 0` 

**Context**: 
borderWidth controls the thickness of the border of input fields among other things. We also make the input field divider line thickness match it. 

eg `borderWidth = 2`
image

**Problem**: 
When borderWidth is 0, the divider lines disappear:
 image

 

**Solution**: 
Make input field divider lines have a minimum thickness of 0.5.
CleanShot 2025-08-28 at 16 04 29@2x

## Motivation
Will help iOS 26 - we default to a minimum border width of 0 there.

## Testing
Manually tested but snapshot tests should cover this well.


davidme-stripe (dismissed):

davidme-stripe (approved):

porter-stripe (approved):
LGTM, wonder if we need a `componentDividerWidth`, like [`componentDivider`](https://github.com/stripe/stripe-ios/blob/master/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetAppearance.swift#L164)

PR #5443, opened 2025-08-29
[iOS 26] Fix keyboard input accessory view not overlapping content no…

yuki-stripe:
…w that it's transparent.

## Motivation
ios26 bug

## Testing
image


wooj-stripe:
is it just me, or ... it seems to be extra space at the bottom of the keyboard now?

wooj-stripe (approved):

Issue #5451, opened 2025-09-02
[BUG] [Identity] Live capture requirement not respected always for document capture


Pearapps:
The upload photo button still is shown even when live capture requirement is requested.

PR #5452, opened 2025-09-02
[Identity] Only show upload photo button in identity document capture if live capture is not required

Pearapps:
## Summary

Fixes [#5451](https://github.com/stripe/stripe-ios/issues/5451)

Enforce live capture by hiding the “Upload a Photo” fallback whenever requireLiveCapture is true. This removes the upload option in camera error and timeout states (leaving only “Try Again”), aligning the iOS UI with the verification session’s live capture requirement.

## Testing

Easiest way to test this is to timeout on the live capture screen.

Then you can see the error screen and verify that the upload photo button does not show up if it is not supposed.


StripeIdentity/StripeIdentity/Source/NativeComponents/ViewControllers/DocumentCaptureViewController.swift
```diff
@@ -194,33 +194,53 @@ final class DocumentCaptureViewController: IdentityFlowViewController {
...
-                .init(
-                    text: .Localized.upload_a_photo,
-                    didTap: { [weak self] in
-                        self?.transitionToFileUpload()
-                    }
-                ),
-            ]
+            if apiConfig.requireLiveCapture {
+                // Hide the upload button when live capture is required
+                return []
```
luisv-stripe, line 199: What does the screen look like in this scenario?

Pearapps: It just does not have the upload button.

Screenshot 2025-09-04 at 3 04 11 PM

luisv-stripe: In the error scenario where there are no buttons returned, does it just show an error screen and the user has to go back?

Pearapps: @luisv-stripe ah, yes.

luisv-stripe: In that case, can we ensure that there will always be a simple way for the user to exit? The top navigation buttons look like the standard iOS navigation. Are there any scenarios where those might be unavailable?

Pearapps: Those should definitely always be there

luisv-stripe (approved):

cjmisenas-stripe (approved):

PR #5455, opened 2025-09-02
Add iOS 26 snapshot tests

yuki-stripe:
## Summary
- Adds AllStripeFrameworks-iOS26 test scheme. This contains our iOS 26 tests.
- Adds `ci_scripts/generate_ios26_test_plan.rb`. This looks for tests and test classes marked `// @iOS26` and updates the test scheme to include them. It's run every time the test scheme is run. (warning: LLM generated)
- Adds FlowController snapshot iOS 26 tests. Tried to add PaymentSheet but they're flaking.

## Motivation
Test iOS 26 changes.

## Testing
I looked at the snapshot tests. I'll investigate the CVC recollection view in a future PR.

## Changelog
Not user facing


Stripe.xcworkspace/xcshareddata/xcschemes/AllStripeFrameworks-iOS26.xcscheme
```diff
@@ -0,0 +1,297 @@
...
+      buildConfiguration = "Debug"
+      selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
+      selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
+      shouldUseLaunchSchemeArgsEnv = "NO">
+      <PreActions>
+         <ExecutionAction
+            ActionType = "Xcode.IDEStandardExecutionActionsCore.ExecutionActionType.ShellScriptAction">
+            <ActionContent
+               title = "Run Script"
+               scriptText = "cd &quot;$SRCROOT/..&quot;&#10;echo &quot;Started at: $(date)&quot; | tee /tmp/generate_ios26_testplan.log&#10;ruby ci_scripts/generate_ios26_testplan.rb | tee -a /tmp/generate_ios26_testplan.log&#10;"
```
wooj-stripe, line 217: I'm not seeing where we actually read in this log file -- if we don't ever read it back in, maybe it would be better to use mktemp to create a temp file?
```
scriptText = "cd &quot;$SRCROOT/..&quot;&#10;LOG_FILE=$(mktemp)&#10;echo &quot;Log file: $LOG_FILE&quot;&#10;echo &quot;Started at: $(date)&quot; | tee $LOG_FILE&#10;ruby ci_scripts/generate_ios26_testplan.rb | tee -a $LOG_FILE&#10;echo &quot;Log saved to: $LOG_FILE&quot;&#10;"
```

ci_scripts/generate_ios26_testplan.rb
```diff
@@ -0,0 +1,245 @@
...
+require 'find'
+
+class IOS26TestPlanGeneratorV2
+  def initialize
+    @test_plan_file = File.join(File.dirname(__dir__), 'Stripe', 'AllStripeFrameworks-iOS26.xctestplan')
+    @tests_by_target = {}
+  end
+
+  def generate
+    log_info "Starting iOS 26 test plan generation (v2)..."
```
wooj-stripe: v2?

ci_scripts/generate_ios26_testplan.rb
```diff
@@ -0,0 +1,245 @@
...
+    log_info "Processing: #{file_path}"
+    
+    content = File.read(file_path, encoding: 'UTF-8')
+    lines = content.lines
+    
+    # Determine target name from file path
+    target_name = determine_test_target(file_path)
+    unless target_name
+      log_info "  ↳ Could not determine test target for #{file_path}"
+      return
```
wooj-stripe: Should we error out if we can't find the test target... and maybe log_error might be more appropriate?

ci_scripts/generate_ios26_testplan.rb
```diff
@@ -0,0 +1,245 @@
...
+        method_name = method_match[1]
+        # Get the class name for this method
+        class_name = find_class_name_for_method(lines, i)
+        if class_name
+          test_identifier = "#{class_name}/#{method_name}()"
+          log_info "    ↳ Method annotation: #{test_identifier}"
+          @tests_by_target[target_name] << test_identifier
+        else
+          log_warning "    ↳ Could not find class name for method #{method_name}"
+        end
```
wooj-stripe: What happens if we can't get the class name?

ci_scripts/generate_ios26_testplan.rb
```diff
@@ -0,0 +1,245 @@
...
+        break
+      
+      # If we hit something else that's not a comment or empty line, stop looking
+      else
+        log_info "    ↳ @iOS26 not followed by class or test method: #{line}"
+        break
+      end
```
wooj-stripe: If we hit this case, is it safe to ignore?

ci_scripts/generate_ios26_testplan.rb
```diff
@@ -0,0 +1,245 @@
...
+      # Read existing test plan
+      existing_plan = JSON.parse(File.read(@test_plan_file))
+      
+      # Update EVERY test target with selectedTests
+      existing_plan["testTargets"].each do |target|
+        target_name = target["target"]["name"]
+        
+        if @tests_by_target[target_name] && @tests_by_target[target_name].any?
+          # Sort and dedupe tests for this target
+          target["selectedTests"] = @tests_by_target[target_name].sort.uniq
```
wooj-stripe, line 189: I'm a little surprised that we would have the potential for duplicated tests -- is this needed?

Tests/ReferenceImages_64/StripePaymentSheetTests.PaymentSheetFlowControllerViewControllerSnapshotTests/testSavedScreen_SEPA_debit_iOS26@3x.png
wooj-stripe, line 1: I'm guessing we'll work on updating these screenshots as we fix more bugs w/ iOS26?

Stripe.xcworkspace/xcshareddata/xcschemes/AllStripeFrameworks-iOS26.xcscheme
yuki-stripe: The log file is to help debugging, I had trouble finding the output in xcode's logs. Happy to change this, what's the difference if we use mktemp? is it easy to know where the log file is?

ci_scripts/generate_ios26_testplan.rb
yuki-stripe: :) second claude attempt, fixed

ci_scripts/generate_ios26_testplan.rb
yuki-stripe: Yeah, good call. I'll make most of these failures error and log_error.

ci_scripts/generate_ios26_testplan.rb
yuki-stripe: I didn't notice this but I suppose it's reasonable to keep, like it's valid if you do:
```
// @iOS26
class FooTest { 
  // @iOS26
  func testFoo() {
```

Tests/ReferenceImages_64/StripePaymentSheetTests.PaymentSheetFlowControllerViewControllerSnapshotTests/testSavedScreen_SEPA_debit_iOS26@3x.png
yuki-stripe: Yeah when we fix this bug the test will fail in CI and we'll have to update the screenshot to make it pass.

ci_scripts/generate_ios26_testplan.rb
yuki-stripe: Will make this log_error and fail.

ci_scripts/generate_ios26_testplan.rb
yuki-stripe: Will make this log_error and fail.

Stripe.xcworkspace/xcshareddata/xcschemes/AllStripeFrameworks-iOS26.xcscheme
wooj-stripe: If we use mktemp, it puts it in somewhere like this:
```
/var/folders/hn/g470538158d2d2v81sj6sg080000gn/T/tmp.FOcvJmwLng
```
and.. i semi assume that some process on my mac will clean it up -- but then again -- anything under `/tmp` should get cleaned up as well. The file is easy to get a hold of if saved in a variable, otherwise, no, it's not easy to get a hold of.  I don't feel too strongly about this. feel free to ignore.

ci_scripts/generate_ios26_testplan.rb
wooj-stripe: Makes sense!

PR #5460, opened 2025-09-03
Add STPConfirmationTokenParams

porter-stripe:
## Summary
- Adds STPConfirmationTokenParams which are params that will be send to the ConfirmationToken create endpoint.
- https://docs.stripe.com/api/confirmation_tokens/test_create (note mandate_data is undocumented)

## Motivation
https://docs.google.com/document/d/1v8r8ARfsTcc61wUjgvPGNJhsDncTqxupTI5sHAkv9xQ/edit?tab=t.0

## Testing
- See unit tests

## Changelog
N/A


StripePayments/StripePayments/Source/API Bindings/Models/ConfirmationTokens/STPConfirmationTokenParams.swift
```diff
@@ -0,0 +1,95 @@
...
+
+    /// Shipping information to include with the ConfirmationToken.
+    @objc public var shipping: STPPaymentIntentShippingDetailsParams?
+
+    /// Details about the Mandate to create.
+    @objc public var mandateData: STPMandateDataParams?
+
+    /// `@YES` to set this ConfirmationToken's PaymentMethod as the associated Customer's default
+    /// This should be a boolean NSNumber, so that it can be `nil`
+    @objc @_spi(STP) public var setAsDefaultPM: NSNumber?
```
joyceqin-stripe, line 38: I know I implemented it as NSNumber? when I did default PMs (I was trying to follow the pattern from savePaymentMethod), but I feel like I remember someone telling me later that setAsDefaultPM could have been a Bool?

Is that the case here as well? Also, should I go back and refactor previous uses of setAsDefaultPM to be Bool? instead of NSNumber?

porter-stripe: The property is marked as `@objc` because `propertyNamesToFormFieldNamesMapping` requires it due to it's implementation. Since it's required to be bride-able to Obj-C, it needs to be a `NSNumber` because we cannot represent a Bool as optional in Obj-C. Maybe we could modify the implementation of `propertyNamesToFormFieldNamesMapping` to not require this but it's following the pattern used throughout the codebase so I lean toward keeping this pattern.

CleanShot 2025-09-03 at 12 05 05

PR #5470, opened 2025-09-03
[iOS26] Fix not respecting shadow when liquid glass enabled

yuki-stripe:
. instead of ignoring shadow, the default should simply be disabled for liquid glass.


Tests/ReferenceImages_64/StripePaymentSheetTests.PaymentSheetSnapshotTests/testPaymentSheetAppearance_iOS26@3x.png
yuki-stripe, line 1: The other problems in this screenshot - the border of the card cell and the border of the sections - is a test only issue, so I'm ignoring for now but I've added to the tracker as a P2.

Here's what it looks when actually run:
image

gbirch-stripe (approved):

PR #5482, opened 2025-09-04
Liquid Glass Remove Button

gbirch-stripe:
## Summary
Added iOS 26 Liquid Glass support to the RemoveButton. The new style is accomplished by setting a background color, imposing a translucent mask on top of the background, and then placing the button on top of that.

## Motivation
iOS 26 Liquid Glass

## Testing
Add new set of snapshot tests to test the button in both light and dark modes

## Changelog
N/A


gbirch-stripe:
Closed in favor of keeping the border for now

PR #5483, opened 2025-09-04
Liquid Glass Remove Button (with outline)

gbirch-stripe:
## Summary
Add iOS 26 Liquid Glass styling to remove button

## Motivation
iOS 26 Liquid Glass support

## Testing
Added remove button snapshot tests

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/RemoveButton.swift
```diff
@@ -36,7 +36,20 @@ class RemoveButton: UIButton {
         directionalLayoutMargins = NSDirectionalEdgeInsets(top: 10, leading: 16, bottom: 10, trailing: 16)
 
         let font = appearance.primaryButton.font ?? appearance.scaledFont(for: appearance.font.base.medium, style: .callout, maximumPointSize: 25)
-        if #available(iOS 15.0, *) {
+        if LiquidGlassDetector.isEnabled, #available(iOS 15.0, *) { // iOS 15 available check is redundant but makes compiler happy 🤠
```
wooj-stripe: seems like this should be:
#available(iOS 26.0, *) ?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/RemoveButton.swift
```diff
@@ -36,7 +36,20 @@ class RemoveButton: UIButton {
...
+            ios26_applyCapsuleCornerConfiguration()
+
+            var config = UIButton.Configuration.plain()
+            config.baseBackgroundColor = .clear
+            let strokeWidth = appearance.selectedBorderWidth ?? appearance.borderWidth * 1.5
+            // The UI doesn't work if we have no stroke width. Follows pattern from LiquidGlassRectangle
```
wooj-stripe: maybe more clear as:
```
// The UI breaks if stroke with <= 0, so default to 1.5
// Workaround:  Use layer.borderWidth because background.strokeWidth isn't compatible with .capsule()
```

gbirch-stripe: Good call, updated

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/RemoveButton.swift
gbirch-stripe: The check is there to make the compiler give us access to the button configuration API, and IMO using the version needed for the thing that it's there for is more readable / clear. Updated code comment for clarity

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/RemoveButton.swift
```diff
@@ -36,7 +36,20 @@ class RemoveButton: UIButton {
...
+            config.baseBackgroundColor = .clear
+            let strokeWidth = appearance.selectedBorderWidth ?? appearance.borderWidth * 1.5
+            // The UI breaks if stroke with <= 0, so default to 1.5. Follows pattern from LiquidGlassRectangle
+            // Workaround:  Use layer.borderWidth because background.strokeWidth isn't compatible with .capsule()
+            layer.borderWidth = strokeWidth > 0 ? strokeWidth : 1.5
+            layer.borderColor = appearance.colors.danger.cgColor
+            config.titleAlignment = .center
+            config.attributedTitle = AttributedString(title, attributes: AttributeContainer([.font: font, .foregroundColor: appearance.colors.danger]))
+            configuration = config
+        } else if #available(iOS 15.0, *) {
```
yuki-stripe: @joyceqin-stripe ooc why do we use UIButton.Configuration here?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/RemoveButton.swift
```diff
@@ -36,7 +36,20 @@ class RemoveButton: UIButton {
         directionalLayoutMargins = NSDirectionalEdgeInsets(top: 10, leading: 16, bottom: 10, trailing: 16)
 
         let font = appearance.primaryButton.font ?? appearance.scaledFont(for: appearance.font.base.medium, style: .callout, maximumPointSize: 25)
-        if #available(iOS 15.0, *) {
+        if LiquidGlassDetector.isEnabled, #available(iOS 15.0, *) { // iOS 15 available check is redundant but needed so compiler allows button config API 🤠
+            ios26_applyCapsuleCornerConfiguration()
```
yuki-stripe, line 48: Strangely the snapshot tests don't seem to show a capsule corner e.g.
CleanShot 2025-09-08 at 10 55 57@2x

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/RemoveButton.swift
joyceqin-stripe: Oh uh I think we didn't have any bordered, unfilled buttons back then and I looked up how to make a bordered button and the Apple dev doc came up. I don't think we specifically need UIButton.Configuration here

yuki-stripe: @joyceqin-stripe I see! Does it look different < iOS 15?

joyceqin-stripe: I don't think so, UIButton just doesn't support the Configuration < iOS 15

yuki-stripe: Ah ok. If we have two implementations that do the same thing w/ no difference/benefits, we should probably have one implementation and halve the code to read/maintain/test!

gbirch-stripe: Afaik UIButton.Configuration is the officially recommend way to customize buttons. I'm not aware of key functionality being deprecated (and thus forcing us to use the newer API), but given that it is best practice it seems sensible that we should try to adopt it

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/RemoveButton.swift
gbirch-stripe: Yes, I don't understand it either yet. I've verified that it's a snapshot-test-only issue

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/RemoveButton.swift
yuki-stripe: I disagree! Maybe we can chat more in the iOS sync. AFAICT the only benefit to UIButton.Configuration is that it is a nicer API. I think we lose more than we benefit if we still have to write/maintain/test it the old way too.

> You can configure and update a button with a [UIButton.Configuration](https://developer.apple.com/documentation/uikit/uibutton/configuration-swift.struct). A button configuration contains all the customization options available with other methods, such as [setTitle(_:for:)](https://developer.apple.com/documentation/uikit/uibutton/settitle(_:for:)), and can serve as a replacement for those methods. Alternatively, you can use a configuration in combination with these other methods and adopt new button behaviors and appearance without rewriting your customized [UIButton](https://developer.apple.com/documentation/uikit/uibutton) code.
https://developer.apple.com/documentation/uikit/uibutton/configuration-swift.struct

gbirch-stripe: Hmm yeah if the only benefit is the API being cleaner then I agree we should stick with a single universal solution and not have branching behavior.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/RemoveButton.swift
yuki-stripe: wow ugh!

yuki-stripe (approved):

PR #5488, opened 2025-09-05
Fix Financial Connections buttons on iOS 26

tillh-stripe:
## Summary

This pull request fixes minor issues on iOS 26 with the navigation bar buttons in the Financial Connections flow. No other iOS 26 design changes are planned for this flow.

| Before | After |
|--------|--------|
| Simulator Screenshot - iPhone 16 Pro - 2025-09-05 at 11 14 38 | Simulator Screenshot - iPhone 16 Pro - 2025-09-05 at 11 13 07 |

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/Helpers/Image.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/Helpers/Image.swift
@@ -35,17 +35,17 @@ enum Image: String, ImageMaker {
 
 extension UIImage {
     func applyFinancialConnectionsBackButtonEdgeInsets() -> UIImage {
-        if #unavailable(iOS 26.0) {
-            return withAlignmentRectInsets(UIEdgeInsets(top: 0, left: -13, bottom: -2, right: 0))
-        } else {
+        if FinancialConnectionsLiquidGlassDetector.isEnabled {
             return self
+        } else {
+            return withAlignmentRectInsets(UIEdgeInsets(top: 0, left: -13, bottom: -2, right: 0))
         }
     }
 }
 
 extension UIBarButtonItem {
     func applyFinancialConnectionsCloseButtonEdgeInsets() {
-        if #unavailable(iOS 26.0) {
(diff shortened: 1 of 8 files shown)


StripeFinancialConnections/StripeFinancialConnections/Source/Helpers/Image.swift
```diff
@@ -32,3 +32,21 @@ enum Image: String, ImageMaker {
...
     case bullet
 }
+
+extension UIImage {
+    func applyFinancialConnectionsBackButtonEdgeInsets() -> UIImage {
+        if #unavailable(iOS 26.0) {
+            return withAlignmentRectInsets(UIEdgeInsets(top: 0, left: -13, bottom: -2, right: 0))
+        } else {
+            return self
+        }
```
mats-stripe: Can we use this condition instead: 

```suggestion
        if LiquidGlassDetector.isEnabled {
            return self
        } else {
            return withAlignmentRectInsets(UIEdgeInsets(top: 0, left: -13, bottom: -2, right: 0))
        }
```

StripeFinancialConnections/StripeFinancialConnections/Source/Helpers/Image.swift
```diff
@@ -32,3 +32,21 @@ enum Image: String, ImageMaker {
...
+    }
+}
+
+extension UIBarButtonItem {
+    func applyFinancialConnectionsCloseButtonEdgeInsets() {
+        if #unavailable(iOS 26.0) {
```
mats-stripe: Same here:


```suggestion
        if !LiquidGlassDetector.isEnabled {
            imageInsets = UIEdgeInsets(top: 0, left: 0, bottom: 0, right: 5)
        }
```

StripeFinancialConnections/StripeFinancialConnections/Source/Helpers/Image.swift
tillh-stripe: I guess we need to use parts of it, but we won't have any opt-in from the merchant like MPE. I'll adjust!

PR #5490, opened 2025-09-05
[iOS 26] Fix how we make input fields larger for iOS 26.

yuki-stripe:
## Summary
Ends up fixing disabled backgrounds as well.

## Motivation
iOS 26

## Testing
- Adds SectionElementSnapshotTest and FormElementSnapshotTest 

## Changelog
Not user facing yet


StripeUICore/StripeUICore/Source/Helpers/StackViewWithSeparator.swift
```diff
@@ -40,14 +40,14 @@ import UIKit
...
-            layoutMargins = UIEdgeInsets(
-                top: spacing, left: spacing, bottom: spacing, right: spacing)
         }
     }
 
     public var borderWidth: CGFloat = 0 {
         didSet {
             backgroundView.layer.borderWidth = borderWidth
+            layoutMargins = UIEdgeInsets(
+                top: borderWidth, left: borderWidth, bottom: borderWidth, right: borderWidth)
```
yuki-stripe, line 50: forgot to move this in my other PR that added `borderWidth`.

StripeUICore/StripeUICore/Source/Elements/Section/SectionContainerView.swift
```diff
@@ -67,39 +68,6 @@ class SectionContainerView: UIView {
 
     override func layoutSubviews() {
         super.layoutSubviews()
-        // Set up each subviews border corners
```
davidme-stripe, line 70: Am I missing a file somewhere? Was all of this not necessary?

yuki-stripe: I was having trouble making this work on iOS 26 without slight visual weirdness where the subview corners were not exactly lined up with the border corner. 

This code's purpose is to get the subview content to not go outside the bounds of the SectionContainerView. That's supposed to be `clipsToBounds = true`'s job, so I implemented it that way instead here.

Why didn't we implement it that way in the first place? I probably couldn't figure out how to make it work the first time, because it was obnoxiously tricky:

The view hierarchy is like this:
```
- (1) SectionContainerView - controls shadows and background color
-- (2) StackViewWithSeparator (UIStackView subclass)
---- (3) TextFieldView - controls background color *when disabled*, otherwise it's clear.
---- (4) backgroundView - draws border radius, width
```

The responsibilities of who draws what is all mixed up - unless you round the corners of 1, 2, and 4, shadows or the disabled background or the border won't be rounded. And StackViewWithSeparator has a very particular and surprising implementation where it draws a border by insetting its views by the border width and un-hiding the background view.

PR #5497, opened 2025-09-08
Fix switching between iOS26->iOS18

wooj-stripe:
## Summary
An assert fires if you:

* Execute payment sheet w/ horizontal mode & iOS26 mode on
* Dismiss payment sheet.
* Turn iOS26 off
* Relaunch payment sheet

## Motivation
assertion failure.

## Testing
Manually

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/PaymentMethodTypeCollectionView.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/PaymentMethodTypeCollectionView.swift
@@ -222,10 +222,14 @@ extension PaymentMethodTypeCollectionView {
         // To repro issue: Launch PS with Appearance.liquidGlassDesignEnabled == true, then launch with
         //                 Appearance.liquidGlassDesignEnabled == false
         private lazy var _liquidGlassRectangle: SelectableRectangle = {
+            #if !os(visionOS)
             if #available(iOS 26.0, *) {
                 return LiquidGlassRectangle(appearance: appearance, isCapsule: false)
             }
             return _shadowedRoundedRectangle
+            #else
+            return _shadowedRoundedRectangle
+            #endif
         }()
         private lazy var _shadowedRoundedRectangle: SelectableRectangle = {
             return ShadowedRoundedRectangle(appearance: appearance)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/PaymentMethodTypeCollectionView.swift
```diff
@@ -216,18 +216,37 @@ extension PaymentMethodTypeCollectionView {
         private lazy var promoBadge: PromoBadgeView = {
             PromoBadgeView(appearance: appearance, tinyMode: true)
         }()
-        private lazy var selectableRectangle: SelectableRectangle = {
+
+        // Store liquidGlassRectangle and shadowedRoundedRectangle separately because
+        // instances of PaymentTypeCell can stick around when turning on/off Appearance.liquidGlassDesignEnabled
+        // To repro issue: Launch PS with Appearance.liquidGlassDesignEnabled == true, then launch with
+        //                 Appearance.liquidGlassDesignEnabled == false
```
yuki-stripe, line 223: Great comment! Instead of this can we just register a different reuse identifier?  I just tried it out and this fixes the assertion too.

```
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/PaymentMethodTypeCollectionView.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/PaymentMethodTypeCollectionView.swift
@@ -178,7 +178,9 @@ extension PaymentMethodTypeCollectionView: UICollectionViewDataSource, UICollect

 extension PaymentMethodTypeCollectionView {
     class PaymentTypeCell: UICollectionViewCell, EventHandler {
-        static let reuseIdentifier = "PaymentTypeCell"
+        static var reuseIdentifier: String {
+            return "PaymentTypeCell" + String(LiquidGlassDetector.isEnabled)
+        }
```

wooj-stripe: This was my first intuition and it did not work -- one sec, i will try again.

wooj-stripe: Yeah -- this approach doesn't work for me, and causes an assertion. Maybe we just go with the current approach?

yuki-stripe: I don't think the PR approach works b/c that cell only adds selectableRectangle once.  If selectableRectangle starts returning a different value, changes to it won't affect the view that's actually displayed.
The root issue is here:
```
        private static let sizingInstance = PaymentTypeCell(frame: .zero)
```

We can make this,`widthCache`, and `minWidth(for:appearance:)` instance vars instead.

yuki-stripe: https://github.com/stripe/stripe-ios/pull/5509

PR #5504, opened 2025-09-09
Manual entry button

joyceqin-stripe:
## Summary

Manual entry button floats over tableview
## Motivation

iOS 26 support
## Testing

Screenshot 2025-09-12 at 10 05 45 AM
Screenshot 2025-09-12 at 10 05 52 AM
Screenshot 2025-09-12 at 10 06 11 AM
Screenshot 2025-09-12 at 10 06 05 AM

## Changelog

N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Resources/Localizations/bg-BG.lproj/Localizable.strings
+++ b/StripePaymentSheet/StripePaymentSheet/Resources/Localizations/bg-BG.lproj/Localizable.strings
@@ -82,6 +82,8 @@
 
 "By saving your bank account for %@ you agree to authorize payments pursuant to <terms>these terms</terms>." = "Като запазвате банковата си сметка за %@, Вие се съгласявате да оторизирате плащания съгласно <terms>тези условия</terms>.";
 
+"By submitting your order to %@ and %@, you agree to save your information with %@ for future purchases, and agree to authorize payments pursuant to <terms>these terms</terms>." = "С изпращането на поръчката си до %1$@ и %2$@ се съгласявате да запазите информацията си в %3$@ за бъдещи покупки и се съгласявате да разрешите плащания съгласно <terms>тези условия</terms>.";
+
 "By submitting your order to %@ you agree to authorize payments pursuant to <terms>these terms</terms>." = "С изпращането на поръчката си до %@ Вие се съгласявате да разрешите плащания съгласно <terms>тези условия</terms>.";
 
 "CPF/CPNJ" = "CPF/CPNJ";
--- a/StripePaymentSheet/StripePaymentSheet/Resources/Localizations/ca-ES.lproj/Localizable.strings
+++ b/StripePaymentSheet/StripePaymentSheet/Resources/Localizations/ca-ES.lproj/Localizable.strings
@@ -82,6 +82,8 @@
 
 "By saving your bank account for %@ you agree to authorize payments pursuant to <terms>these terms</terms>." = "En desar el compte bancari per a %@, accepteu l'autorització de pagaments d'acord amb <terms>aquests termes</terms>.";
 
+"By submitting your order to %@ and %@, you agree to save your information with %@ for future purchases, and agree to authorize payments pursuant to <terms>these terms</terms>." = "Si envieu la comanda a %1$@ i %2$@, accepteu desar la vostra informació amb %3$@ per a compres futures i accepteu autoritzar els pagaments d'acord amb <terms>aquestes condicions</terms>.";
+
 "By submitting your order to %@ you agree to authorize payments pursuant to <terms>these terms</terms>." = "En enviar la comanda a %@, accepteu l'autorització de pagaments d'acord amb <terms>aquestes condicions</terms>.";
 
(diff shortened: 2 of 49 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/AutoComplete/AutoCompleteViewController.swift
```diff
@@ -138,14 +138,29 @@ class AutoCompleteViewController: UIViewController {
...
-        let stackView = UIStackView(arrangedSubviews: [formStackView, errorLabel, separatorView, tableView, manualEntryButton])
+        let buttonContainer = UIView()
+        buttonContainer.addSubview(manualEntryButton)
+        manualEntryButton.translatesAutoresizingMaskIntoConstraints = false
+        NSLayoutConstraint.activate([
+            manualEntryButton.leadingAnchor.constraint(equalTo: buttonContainer.leadingAnchor, constant:
+                                                            LiquidGlassDetector.isEnabled ? configuration.appearance.formInsets.leading : 0),
+            manualEntryButton.trailingAnchor.constraint(equalTo: buttonContainer.trailingAnchor, constant:
+                                                            LiquidGlassDetector.isEnabled ? -configuration.appearance.formInsets.trailing : 0),
+            manualEntryButton.topAnchor.constraint(equalTo: buttonContainer.topAnchor),
```
wooj-stripe: Should we be using formInsets here, or just hard code a specific offset?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/ManualEntryButton.swift
```diff
@@ -15,24 +15,38 @@ extension UIButton {
...
+            button.ios26_applyGlassConfiguration()
+            button.frame.size.height = appearance.primaryButton.height
+            let textColor = UIColor(dynamicProvider: { traitCollection in
+                if traitCollection.isDarkMode {
+                    return appearance.colors.background.contrastingColor
+                }
+
+                return appearance.colors.primary
+            })
+            button.configuration?.attributedTitle = AttributedString(.Localized.enter_address_manually, attributes: AttributeContainer([.font: font, .foregroundColor: textColor]))
```
wooj-stripe: Nit, move this to be the first line in the if-block since the else is updating titleLabel/tint/title/etc as the first thing in the else-block

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/AutoComplete/AutoCompleteViewController.swift
```diff
@@ -138,14 +138,29 @@ class AutoCompleteViewController: UIViewController {
...
+            manualEntryButton.bottomAnchor.constraint(equalTo: buttonContainer.bottomAnchor, constant: LiquidGlassDetector.isEnabled ? -8 : 0),
```
yuki-stripe: Use `buttonContainer.addAndPinSubview(.., insets: ...)` instead

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/AutoComplete/AutoCompleteViewController.swift
yuki-stripe: Probably formInsets so that it lines up with the nav bar buttons (which, it doesn't today, but we are going to change formInsets so that it does line everything up).

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/ManualEntryButton.swift
```diff
@@ -15,24 +15,38 @@ extension UIButton {
...
-            return appearance.colors.background.darken(by: 0.07)
-        })
+            button.backgroundColor = UIColor(dynamicProvider: { traitCollection in
+                if traitCollection.isDarkMode {
+                    return appearance.colors.componentBackground
+                }
 
+                return appearance.colors.background.darken(by: 0.07)
+            })
+        }
```
yuki-stripe: I think we might need to do something similar for our color, because the white button doesn't have good contrast if the background is white.

image

joyceqin-stripe: Should I use prominentClearGlass? It has better contrast when the background is white, but it also looks worse on the normal background. This is with the baseBackgroundColor set to the computed background color with alpha 0.5.
Screenshot 2025-09-11 at 10 15 55 AM
Screenshot 2025-09-11 at 10 16 04 AM
Screenshot 2025-09-11 at 10 16 24 AM

joyceqin-stripe: I also tried setting the background color with an alpha for the normal glass configuration, but it feels like the difference was negligible. Is there a better way to tint glass?

PR #5507, opened 2025-09-09
Opt-in API for iOS26 changes behind dev flag

wooj-stripe:


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/BottomSheetViewController.swift
```diff
@@ -106,6 +106,17 @@ class BottomSheetViewController: UIViewController, BottomSheetPresentable {
...
 
+    var isUsingGlassNavBar: Bool {
+        #if !os(visionOS)
+        guard #available(iOS 26.0, *) else {
+            return false
+        }
+        return LiquidGlassDetector.canRun && appearance.navigationBarStyle == .glass
+        #else
+        return false
+        #endif
```
yuki-stripe: This allows the possibility that `appearance.navigationBarStyle == .glass` while the bottom sheet thinks it's not using glass. Which one is right?

Can this just be `return appearance.navigationBarStyle == .glass`? We should have that be the single source of truth.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/SheetNavigationBar.swift
```diff
@@ -19,8 +19,18 @@ protocol SheetNavigationBarDelegate: AnyObject {
...
+        if LiquidGlassDetector.canRun,
+           #available(iOS 26.0, *),
+           appearance.navigationBarStyle == .glass {
+            return 76
+        } else {
+            return 52
+        }
+        #else
+        return 52
+        #endif
```
yuki-stripe: Same here.

Stripe/StripeiOSTests/ConfirmButtonSnapshotTests.swift
```diff
@@ -217,3 +217,23 @@ class ConfirmButtonSnapshotTests: STPSnapshotTestCase {
...
+
+extension PaymentSheet.Appearance {
+    mutating func applyLiquidGlassIfPossible() {
+        #if !os(visionOS)
+        if #available(iOS 26.0, *) {
+            self.applyLiquidGlass()
+        }
+        #endif
+    }
+    var liquidGlassIfPossible: PaymentSheet.Appearance {
```
yuki-stripe: i believe the standard ios naming pattern for this would be something more like`func applyingLiquidGlassIfPossible() -> PaymentSheet.Appearance`

Stripe/StripeiOSTests/ConfirmButtonSnapshotTests.swift
```diff
@@ -217,3 +217,23 @@ class ConfirmButtonSnapshotTests: STPSnapshotTestCase {
...
+        #if !os(visionOS)
+        if #available(iOS 26.0, *) {
```
yuki-stripe, line 232: Shouldn't have a duplicate implementation of the checks here; that's `applyLiquidGlassIfPossible`'s responsibility.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetAppearance.swift
```diff
@@ -296,6 +305,24 @@ public extension PaymentSheet {
     }
 }
 
+public extension PaymentSheet.Appearance {
+    /// Calling this function sets various properties (e.g. navigationBarStyle, borderWidth) to match iOS26 Liquid Glass
```
yuki-stripe, line 310: Should we call out this is in public preview?
```
- Note: This feature is in public preview while we gather feedback and is subject to change. Please use https://github.com/stripe/stripe-ios/issues to file feedback!
```
Something like this?

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/LinkUI.swift
```diff
@@ -204,6 +204,13 @@ extension LinkUI {
 
     static let appearance: PaymentSheet.Appearance = {
         var appearance = PaymentSheet.Appearance.default
+        // TODO: Remove LiquidGlassDetector.allowNewDesign
```
yuki-stripe, line 207: Interesting. So Link UI (not sure what this controls - everything Link related?) will be unconditionally Liquid Glass-y on iOS 26 etc, even when the merchant opts out in the plist or does not call `applyLiquidGlass`?

Example/PaymentSheet Example/PaymentSheet Example/CustomerSheetTestPlayground.swift
```diff
@@ -114,16 +115,41 @@ struct CustomerSheetTestPlayground: View {
...
+            return playgroundController.settings.liquidGlassNavigation
+        } set: { newValue in
+            guard #available(iOS 26.0, *) else {
+                return
+            }
+            playgroundController.appearance = PaymentSheet.Appearance()
+            playgroundController.appearance.navigationBarStyle = newValue == .on ? .glass : .plain
+            playgroundController.settings.liquidGlassNavigation = newValue
+        }
+    }
```
yuki-stripe: This should be in the appearance settings no?

Example/PaymentSheet Example/PaymentSheet Example/CustomerSheetTestPlayground.swift
```diff
@@ -114,16 +115,41 @@ struct CustomerSheetTestPlayground: View {
...
+
+    var liquidGlassBinding: Binding<CustomerSheetTestPlaygroundSettings.LiquidGlass> {
+        Binding<CustomerSheetTestPlaygroundSettings.LiquidGlass> {
+            return playgroundController.settings.liquidGlass
         } set: { newValue in
-            LiquidGlassDetector.allowNewDesign = newValue == .on
+            guard #available(iOS 26.0, *) else {
+                return
+            }
+            LiquidGlassDetector.allowNewDesign = false
```
yuki-stripe: Why does this unconditionally set to `false`? I'd expect this to toggle

Example/PaymentSheet Example/PaymentSheet Example/CustomerSheetTestPlayground.swift
```diff
@@ -114,16 +115,41 @@ struct CustomerSheetTestPlayground: View {
...
             playgroundController.appearance = PaymentSheet.Appearance()
```
yuki-stripe: I think this should be in the appearance settings and it should be a button. The toggle is not a good UI for this because it doesn't flip a value. It's a "reset appearance and apply liquid glass" button.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/BottomSheetViewController.swift
wooj-stripe: You're right -- the logic here is too complicated -- My thought process was that users could use `UIDesignRequiresCompatibility == true` . If a user does this, and sets a button's configuration to `.glass()` -- it has a no-op effect -- so, I was protecting against that case.

As a result, I think we should catch this earlier in the process e.g. when calling applyLiquidGlass -- if a user has set `UIDesignRequiresCompatibility` to `true` we skip execution of `applyLiquidGlass`.

yuki-stripe: 👍 I think asserting is better than silently failing, since we're saying it's an integration error.

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/LinkUI.swift
wooj-stripe: cc: @tillh-stripe

Example/PaymentSheet Example/PaymentSheet Example/CustomerSheetTestPlayground.swift
wooj-stripe: Ah! yes! we can!

Stripe/StripeiOSTests/ConfirmButtonSnapshotTests.swift
wooj-stripe: I couldn't find a better place to put this?  The other impl is in a different package.

Stripe/StripeiOSTests/ConfirmButtonSnapshotTests.swift
wooj-stripe: Hmmm our code would be, `.default.applyingLiquidGlassIfPossible`.  I don't feel strongly one way or another -- i'm fine with that.

Example/PaymentSheet Example/PaymentSheet Example/CustomerSheetTestPlayground.swift
wooj-stripe: I think I was working on an implementation where i was detecting whether or not applyLiquidGlass was executed or not -- but indeed, we can make this toggle.

Example/PaymentSheet Example/PaymentSheet Example/CustomerSheetTestPlayground.swift
wooj-stripe: sg -- i guess the reset appearance button would then reset everything to not opt-int into liquid glass.

Stripe/StripeiOSTests/ConfirmButtonSnapshotTests.swift
yuki-stripe: sorry what other implementation? `applyLiquidGlassIfPossible's ` is defined above this. I'm recommending you delete these two lines.

Stripe/StripeiOSTests/ConfirmButtonSnapshotTests.swift
yuki-stripe: https://www.swift.org/documentation/api-design-guidelines/#name-according-to-side-effects

```
/// Strips all the newlines from `self`
mutating func stripNewlines()

/// Returns a copy of `self` with all the newlines stripped.
func strippingNewlines() -> String
...
s.stripNewlines()
let oneLine = t.strippingNewlines()
...
```

Example/PaymentSheet Example/PaymentSheet Example/CustomerSheetTestPlayground.swift
yuki-stripe: The reset appearance button just does `appearance = PaymentSheet.Appearance()`

The 'apply liquid glass' button should really just do `appearance.applyLiquidGlass()`

Stripe/StripeiOSTests/ConfirmButtonSnapshotTests.swift
wooj-stripe: AH!! got it.... will update, but also this code is being duplicated in `STPFixtures+PaymentSheet.swift` due to it being in a different package.

Example/PaymentSheet Example/PaymentSheet Example/AppearancePlaygroundView.swift
```diff
@@ -439,6 +440,27 @@ struct AppearancePlaygroundView: View {
                 }
 
                 Button {
+                    appearance = PaymentSheet.Appearance()
+                    if #available(iOS 26.0, *) {
+                        appearance.applyLiquidGlass()
+                    }
+                    doneAction(appearance)
+                } label: {
+                    Text("Reset and apply Liquid Glass 🥃")
```
yuki-stripe: 🥃 ooc any particular reason we're resetting here rather than just applying liquid glass?

Example/PaymentSheet Example/PaymentSheet Example/AppearancePlaygroundView.swift
```diff
@@ -439,6 +440,27 @@ struct AppearancePlaygroundView: View {
                 }
 
                 Button {
+                    appearance = PaymentSheet.Appearance()
+                    if #available(iOS 26.0, *) {
```
yuki-stripe: Can we put the if outside this button so it doesn't show up < iOS 26?

Example/PaymentSheet Example/PaymentSheet Example/AppearancePlaygroundView.swift
```diff
@@ -439,6 +440,27 @@ struct AppearancePlaygroundView: View {
...
+                }
+
+                Button {
+                    appearance = PaymentSheet.Appearance()
+                    if #available(iOS 26.0, *) {
+                        appearance.navigationBarStyle = .glass
+                    }
+                    doneAction(appearance)
+                } label: {
+                    Text("Reset and apply Liquid Glass Nav 🧭")
```
yuki-stripe: This one should be a dropdown/toggle/switch - there are two states the navigationBarStyle can be in.

Stripe/StripeiOSTests/ConfirmButtonSnapshotTests.swift
yuki-stripe: I think something is getting lost in translation here. This code should look like:
```
    mutating func applyLiquidGlassIfPossible() {
        #if !os(visionOS)
        if #available(iOS 26.0, *) {
            self.applyLiquidGlass()
        }
        #endif
    }
    func applyingLiquidGlassIfPossible() -> PaymentSheet.Appearance { ... }
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetAppearance.swift
```diff
@@ -296,6 +305,26 @@ public extension PaymentSheet {
...
+    /// - Note: This feature is in public preview while we gather feedback and is subject to change. Please use https://github.com/stripe/stripe-ios/issues to file feedback!
+    @available(iOS 26.0, *)
+    @_spi(STP) mutating func applyLiquidGlass() {
+        assert(LiquidGlassDetector.isEnabledInMerchantApp, "Requirements for applying liquid glass are not met.")
```
yuki-stripe: The first thought as a user seeing this error message is probably "what are the requirements I'm not meeting?". Can we explain what went wrong / what they need to do to fix it?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetAppearance.swift
```diff
@@ -428,3 +457,19 @@ extension PaymentSheet.Appearance {
...
+extension PaymentSheet.Appearance.NavigationBarStyle {
+    var isGlass: Bool {
+        #if !os(visionOS)
+        guard #available(iOS 26.0, *) else {
+            return false
+        }
+        return self == .glass
+        #else
+        return false
+        #endif
```
yuki-stripe, line 472: Ideally we don't have a situation where `appearance.navigationBarStyle == .glass` while the navigation bar style is not, in fact, glass. That's possible with this code.

Can we make this var just return `self == .glass` (or remove it, it's not doing anything at that point) and then ensure the user can't set `navigationBarStyle = .glass` if, in fact, the navigation bar style can't be glass?

Example/PaymentSheet Example/PaymentSheet Example/AppearancePlaygroundView.swift
wooj-stripe: I guess i'd expect users to call apply liquid glass before applying any other customizations, so I figured resetting was appropriate -- e.g.:
```
let appearance = Appearance()
appearance.applyLiquidGlass()
apperance.colors..... = x
```

wooj-stripe: Added two options.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetAppearance.swift
wooj-stripe: paired offline, resolving.

PR #5511, opened 2025-09-09
Deprecate Promises, remove AsyncTests

gbirch-stripe:
## Summary
- Promises are not thread safe and can cause crashes and should be eliminated from the SDK. Marking them as deprecated to discourage people from adding new code with them
- AsyncTests are regularly failing not because of issues with the tests but because the Promises code they are testing is buggy. Removing these tests since they do not provide us with any information and slow down CI

## Motivation
Ensure that no new promise code is added to the SDK
Remove CI friction

## Testing
N/A

## Changelog
N/A


StripeCore/StripeCoreTests/Helpers/AsyncTests.swift
```diff
@@ -1,155 +0,0 @@
...
-        chainedFuture.observe { result in
-            XCTAssertEqual(result.successValue, 84)
-            expectation.fulfill()
-        }
-
-        wait(for: [expectation], timeout: 1.0)
-    }
-
-    func testFutureChainedFailure() {
-        let promise = Promise<Int>()
```
porter-stripe, line 55: Are promises still being used in the code base anywhere? Even though we don't want people to use them in the future and the tests are flakey I'm a little fud-ful of removing the tests.

gbirch-stripe: They're used in financial connections and identity, and I've notified both teams of this. The issue with keeping the tests around is that some of the scenarios they test are indeed broken, and the only reason they _usually_ pass is that we're not getting unlucky with the timing. So between failing all the time and only passing because they're not accurate, I'm not sure what the purpose of them is

porter-stripe: cc @mats-stripe can you review and see if this looks good as only FC and Identity use promises?

StripeCore/StripeCore/Source/Helpers/Async.swift
```diff
@@ -123,6 +124,7 @@ import Foundation
     }
 }
 
+// ⛔️ DEPRECATED: Promises are not fully thread safe and can cause crashes. Use Swift concurrency instead. ⛔️
 @_spi(STP) public class Promise<Value>: Future<Value> {
```
porter-stripe, line 128: The comments may get overlooked or not seen, one idea is to move this from STP to something like "DeprecatedPromises" or something so future users need to ack that they are deprecated if they start using them in a place where they aren't currently used.

gbirch-stripe: I tinkered with this but decided against it since the only reason I can imagine someone writing new Promise code is to interact with existing Promise code, which there's no way around right now due to lack of interoperability. So my feeling was that it was add friction without benefit, but I think I still have the commit somewhere so I'm curious to hear your thoughts

gbirch-stripe:
Tracking removal of Promise code:
Identity: https://github.com/stripe/stripe-ios/issues/5557
FC: https://jira.corp.stripe.com/browse/RUN_LINK_MOBILE-34

PR #5512, opened 2025-09-10
Adopt new Liquid Glass API in Link

tillh-stripe:
## Summary

This pull request updates the Link UI to work better with the recent changes to the Liquid Glass approach in MPE.


StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Components/Hint/LinkHintMessageView.swift
```diff
@@ -79,7 +78,12 @@ final class LinkHintMessageView: UIView {
             stackView.centerYAnchor.constraint(equalTo: centerYAnchor),
         ])
         backgroundColor = .linkSurfaceSecondary
-        layer.cornerRadius = Constants.cornerRadius
+
+        if let cornerRadius = LinkUI.appearance.cornerRadius {
+            layer.cornerRadius = cornerRadius
```
yuki-stripe, line 83: Not blocking but - why do we have both `LinkUI.appearance.cornerRadius` and `LinkUI.cornerRadius`? When are you supposed to use one vs the other?  Might be some room for improvement here

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/LinkUI.swift
```diff
@@ -203,16 +218,29 @@ extension LinkUI {
 
 extension LinkUI {
 
-    static let appearance: PaymentSheet.Appearance = {
+    static func applyLiquidGlassIfPossible(configuration: PaymentElementConfiguration) {
+        Self.useLiquidGlass = configuration.appearance.cornerRadius == nil
```
yuki-stripe: `cornerRadius == nil` means "default":
- below iOS 26, the corner radius is 6.0 (ie unchanged from the current default). 
- On iOS 26+,  multiple rounder corner configurations are used to match Liquid Glass design.
```suggestion
        Self.useLiquidGlass = configuration.appearance.cornerRadius == nil, LiquidGlassDetector.isEnabledInMerchantApp
```

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Components/NavigationBar/LinkSheetNavigationBar.swift
```diff
@@ -98,9 +98,8 @@ class LinkSheetNavigationBar: SheetNavigationBar {
 
     override func createBackButton() -> UIButton {
         let image = Image.icon_chevron_left_standalone.makeImage(template: true)
```
yuki-stripe, line 100: Ooh you know, we use system icons now for the liquid glass nav bar to exactly match the system. Maybe a follow- up.https://github.com/stripe/stripe-ios/pull/5544/files#diff-dab538710af695b014168dcd1a906083476596d0717dc3458de14748011beb74

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Components/Hint/LinkHintMessageView.swift
tillh-stripe: They both resolve to `LinkUI.cornerRadius`, but yes, there's no good answer. Now that we rely on something dynamic, we'll refactor this soon to avoid the static `LinkUI` approach.

yuki-stripe (approved):

PR #5514, opened 2025-09-10
Fix iOS26 animation w/ apple pay button

wooj-stripe:
## Summary
Animation bug with iOS26.

To repro:
* Returning customer
* Apple pay Enabld
* PaymentSheet.Complete w/ vertical mode
* Tap on "add card"

Observed behavior:
* Apple pay button grow/shrinks when switching out content between payment method view page and add card page

## Motivation
Animation bug with iOS26

## Testing
Manually on iOS26 and iOS18
* Switching between main page and add card page
* Switching between dark mode and light mode.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/WalletHeaderView.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/WalletHeaderView.swift
@@ -169,7 +169,8 @@ extension PaymentSheetViewController {
             addAndPinSubview(stackView)
         }
         private func updateStackView() {
-            // Update Apple Pay Button's color
+            // There's no way to change the color (PKPaymentButtonStyle) of the button,
+            // so swap between light and dark versions of Apple Pay button if needed
             let isBlackApplePayButton = appearance.colors.background.contrastingColor == .black
             if isBlackApplePayButton {
                 if let whiteApplePayButtonIndex = stackView.arrangedSubviews.firstIndex(of: applePayButtonWhite) {
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/WalletHeaderView.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/WalletHeaderView.swift
@@ -73,6 +73,18 @@ extension PaymentSheetViewController {
             return button
         }()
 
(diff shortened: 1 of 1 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/WalletHeaderView.swift
```diff
@@ -152,14 +163,37 @@ extension PaymentSheetViewController {
...
+                }
+            } else {
+                if let blackApplePayButonIndex = stackView.arrangedSubviews.firstIndex(of: applePayButtonBlack) {
+                    applePayButtonBlack.removeFromSuperview()
+                    stackView.removeArrangedSubview(applePayButtonBlack)
+                    stackView.insertArrangedSubview(applePayButtonWhite, at: blackApplePayButonIndex)
+                    if isApplePayLastButton {
+                        stackView.setCustomSpacing(Constants.labelSpacing, after: applePayButtonWhite)
+                    }
+                }
```
yuki-stripe, line 192: Seems like we could simplify by adding both white and black and then hide one or the other. But NBD, feel free to ship as-is.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/WalletHeaderView.swift
```diff
@@ -152,14 +163,37 @@ extension PaymentSheetViewController {
 
             if let lastButton = buttons.last {
                 stackView.setCustomSpacing(Constants.labelSpacing, after: lastButton)
+                isApplePayLastButton = lastButton == applePayButton
             }
 
             addAndPinSubview(stackView)
         }
+        private func updateStackView() {
+            // Update Apple Pay Button's color
```
yuki-stripe: Would be nice to comment why we do this - there's no way to change the color (`PKPaymentButtonStyle`) of the button, so we have to swap between light and dark versions.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/WalletHeaderView.swift
wooj-stripe: I suspect that if we hide/show we'd have to deal with the implicit spacing of a hidden view ... going to keep this approach.

yuki-stripe (approved):

PR #5518, opened 2025-09-10
Allow updating Shop Pay shipping rates

davidme-stripe:

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/ExampleWalletButtonsView.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/ExampleWalletButtonsView.swift
@@ -559,7 +559,7 @@ class ExampleWalletButtonsModel: ObservableObject {
                                 .init(name: "Silver Potato", amount: 345),
                                 .init(name: "Tax", amount: 200),
                                 .init(name: "Shipping", amount: selectedRate.amount), ],
-                    shippingRates: shippingRates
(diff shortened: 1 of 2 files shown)


StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/ECE/ShopPayECEPresenterTests.swift
```diff
@@ -703,4 +703,259 @@ class ShopPayECEPresenterTests: XCTestCase {
...
+        ]
+
+        // When - First update shipping address to get new rates
+        let firstResponse = try await sut.eceView(mockECEViewController, didReceiveShippingAddressChange: shippingAddress)
+        await fulfillment(of: [expectation], timeout: 1.0)
+
+        // Then - Check that new shipping rate is available
+        let firstShippingRates = firstResponse["shippingRates"] as? [[String: Any]]
+        XCTAssertEqual(firstShippingRates?.count, 3) // Original 2 + new 1
+        XCTAssertTrue(firstShippingRates?.contains { ($0["id"] as? String) == "overnight" } ?? false)
```
wooj-stripe, line 769: We're asserting shipping rates, but this funciton is named change in shipping address?

davidme-stripe: Yes, both the `shippingContactUpdateHandler` and `shippingMethodUpdateHandler` trigger an update of the cached rates, so there's one test to check if the change in address triggers a cache update, and then the next test checks if a change in rates triggers a cache update.

PR #5519, opened 2025-09-11
Improve polling logic

porter-stripe:
## Summary
- Improves retry logic to prevent an excessive polling duration when network timeouts occur. This logic was originally added for [3DS2 & SFSafariViewController](https://github.com/stripe-ios/stripe-ios-private/blob/private/StripePayments/StripePayments/PaymentHandler/STPPaymentHandler.swift#L1166) latency handling and later reused for LPM polling. This PR improves all retry instances in `STPPaymentHandler`.
- Implements the polling algorithm as specified here: https://docs.google.com/document/d/1hImjk9dQsrQ0UFq8eS_ujMMa8NnpzctSOz39dFHbU-E/edit?tab=t.0

## Motivation
- [MOBILESDK-3378](https://jira.corp.stripe.com/browse/MOBILESDK-3378)

## Testing
- Manual
- Existing tests
- New tests

## Changelog
- See diff


porter-stripe:
Pretty sure the dead code check is flagging a false positive.

StripePayments/StripePayments/Source/PaymentHandler/PollingBudget.swift
```diff
@@ -0,0 +1,92 @@
...
+        // During tests where we record network traffic, if we are hitting stubs we can make more network requests
+        // due to near-instant network responses, so we need to scale down the polling duration to account for this.
+        return isUsingNetworkStubs ? rawValue * 0.4 : rawValue
+    }
+
+    /// Detects if we're running in stub playback mode (HTTPStubs active but not recording)
+    private var isUsingNetworkStubs: Bool {
+        // Only scale duration when we're using stubs (HTTPStubs present and not recording the network STP_RECORD_NETWORK)
+        return NSClassFromString("HTTPStubs") != nil && ProcessInfo.processInfo.environment["STP_RECORD_NETWORK"] == nil
+    }
```
yuki-stripe: We should be rate-limiting requests to at most once a second - are we doing that?

StripePayments/StripePayments/Source/PaymentHandler/PollingBudget.swift
```diff
@@ -0,0 +1,92 @@
...
+/// Manages polling budgets for payment method transactions that require status polling.
+final class PollingBudget {
+    /// The timestamp when polling started (set when start() is called)
+    private var startDate: Date
+    /// The polling duration with automatic test environment adjustment
+    let duration: PollingDuration
+    /// Whether polling is currently allowed (within budget)
+    private(set) var canPoll: Bool = true
+
+    /// The elapsed time since polling started, or nil if polling hasn't started
```
yuki-stripe: Can't be nil.

StripePayments/StripePayments/Source/PaymentHandler/PollingBudget.swift
```diff
@@ -0,0 +1,92 @@
...
+        }
+    }
+
+    /// Helper function to determine the polling budget for any given payment method type
+    /// - Parameter paymentMethodType: The payment method type
+    /// - Returns: The polling budget if it exists for this payment method
+    static func networkTimeout(for paymentMethodType: STPPaymentMethodType) -> NSNumber? {
+        guard let pollingBudget = PollingBudget(startDate: Date(), paymentMethodType: paymentMethodType)?.duration.value else { return nil }
+        return NSNumber(value: pollingBudget)
+    }
```
yuki-stripe: I updated some comments in the doc after we last sync'd - I think this should be POLLING_DURATION - TIME_ELAPSED

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
```diff
@@ -1533,11 +1548,13 @@ public class STPPaymentHandler: NSObject {
                         if
                             let paymentMethod = paymentIntent.paymentMethod,
                             !STPPaymentHandler._isProcessingIntentSuccess(for: paymentMethod.type),
-                            paymentIntent.status == .processing && retryCount > 0
+                            paymentIntent.status == .processing,
+                            pollingBudget?.canPoll ?? true
                         {
-                            self._retryAfterDelay(retryCount: retryCount) {
+                            let processingPollingBudget = pollingBudget ?? PollingBudget(startDate: startDate, duration: 2)
```
yuki-stripe: i thought we decided on 1 second for this case?

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
```diff
@@ -1516,6 +1530,7 @@ public class STPPaymentHandler: NSObject {
             pingMarlinIfNecessary(
                 currentAction,
                 {
+                    let startDate = Date()
                     self.retrieveOrRefreshPaymentIntent(
```
yuki-stripe, line 1525: What about setting the network timeout to POLLING_DURATION - TIME_ELAPSED? Maybe that's too difficult to do w/o refactoring the code more?

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
```diff
@@ -1607,10 +1625,12 @@ public class STPPaymentHandler: NSObject {
                 currentAction.setupIntent = setupIntent
                 if let type = setupIntent.paymentMethod?.type,
                    !STPPaymentHandler._isProcessingIntentSuccess(for: type),
-                   setupIntent.status == .processing && retryCount > 0
+                   setupIntent.status == .processing,
+                   pollingBudget?.canPoll ?? true
                 {
-                    self._retryAfterDelay(retryCount: retryCount) {
-                        self._retrieveAndCheckIntentForCurrentAction(retryCount: retryCount - 1)
+                    let processingPollingBudget = pollingBudget ?? PollingBudget(startDate: startDate, duration: 5)
```
yuki-stripe: is this not the 2-second case above? Why 5 seconds?

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
```diff
@@ -1475,13 +1475,27 @@ public class STPPaymentHandler: NSObject {
...
+    ///   - delay: Time interval to wait before execution
+    ///   - pollingBudget: Budget tracker for polling attempts
+    ///   - block: Block to execute if budget allows
+    func pollAfterDelayIfBudgetAllows(delay: TimeInterval = 1, pollingBudget: PollingBudget, block: @escaping () -> Void) {
+        DispatchQueue.main.asyncAfter(deadline: .now() + delay) {
+            if pollingBudget.canPoll {
+                pollingBudget.recordPollAttempt()
+                block()
+            }
         }
```
yuki-stripe: Is it safe to not call the block in this method? I think that would mean this class never finishes

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
porter-stripe: Yep, should be 1!

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
porter-stripe: Ah they should match, now it is 1 second.

StripePayments/StripePayments/Source/PaymentHandler/PollingBudget.swift
porter-stripe: I considered that, but I felt like it would be somewhat risky! Setting a network timeout of less than 1 second concerns me, as it could be too short. Even a 1-second timeout is really quick... maybe we should ensure a minimum network timeout of 3 seconds?

```swift
    static func networkTimeout(for paymentMethodType: STPPaymentMethodType) -> NSNumber? {
        guard let pollingBudget = PollingBudget(startDate: Date(), paymentMethodType: paymentMethodType)?.duration.value else { return nil }
        return max(3, NSNumber(value: pollingBudget))
    }
```

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
porter-stripe: Totally doable with a small refactor, but I think it give me a bit of FUD that the network timeout can be so small, I think it's too aggressive of an approach.

StripePayments/StripePayments/Source/PaymentHandler/PollingBudget.swift
porter-stripe: Yeah we [are](https://github.com/stripe/stripe-ios/pull/5519/files#diff-6a81f3ee91853b490eedd97eac89e41e2a6e353447556afaa2cf3bacbb025a83R1484), even with that, the stubbed requests can get off about 1 more request than when actually hitting the network.

StripePayments/StripePayments/Source/PaymentHandler/PollingBudget.swift
yuki-stripe: What's the problem if the network request times out?

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
porter-stripe: Nope!! I made a mistake here, there can be a race condition....

Before calling the poll function, we always check `pollingBudget.canPoll`. If it's false, we execute another block of code to handle the current action as before. This pattern isn't perfect, but it's an artifact of the previous retry count approach, improving it would probably require a significant refactor.

Here's where the race condition happens: The first call to `pollingBudget.canPoll` returns true, so we proceed. But inside `pollAfterDelayIfBudgetAllows`, after the delay expires, we check `pollingBudget.canPoll` again. If it now returns false (because time ran out during the delay), we never execute the else block that's supposed to clean up the current action.

I'm going to remove the second check to `canPoll` inside of `pollAfterDelayIfBudgetAllows`. This could lead to us executing the retry block after the budget has been exhausted, but that should only happen once and is within the spec we laid out.

Example:
https://github.com/stripe/stripe-ios/blob/porter/improve-polling-2/StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift#L1592

StripePayments/StripePayments/Source/PaymentHandler/PollingBudget.swift
yuki-stripe: If the last request took >1 second, we shouldn't wait 1 second before firing a new request - we should send immediately.
If the last request took < 1 second, we need to wait 1 second - last_request_duration, not 1 second.

StripePayments/StripePayments/Source/PaymentHandler/PollingBudget.swift
porter-stripe: I think we should give the network a reasonable amount of time to respond. We could be timing out too early if we only give the network one second or less to return a response.

yuki-stripe: I think it's a trade-off between optimizing (lowering the duration of) the cancel case vs the success/failure case. I'd rather optimize the cancel case b/c I think waiting a couple of seconds in the cancel case is relatively much more frustrating than the success/failure case.

Like let's say we're at the 4 second mark when POLLING_DURATION = 5 seconds. Let's assume that the request needs 3 seconds to complete. Compare the cancel case:

1. We set the timeout to 1 second.
1. The network request times out after 1 second.
1. We make a final request at the 5 second mark.
2. We return .canceled at the 8 second mark.

vs.

1. We set the timeout to 3 seconds.
3. The network succeeds after 3 seconds
4. We make a final request at the 7 second mark.
5. We return .canceled at the 10 second mark.

---

Compare the success case:

1. We set the timeout to 1 second.
1. The network request times out after 1 second.
1. We make a final request at the 5 second mark.
2. We return success at the 8 second mark.

vs.

1. We set the timeout to 3 seconds.
3. The network succeeds after 3 seconds
5. We return success at the 7 second mark.

StripePayments/StripePayments/Source/PaymentHandler/PollingBudget.swift
porter-stripe: I see, I just pushed a commit to do this. With one caveat, for the first poll we always wait the full second, recording the network time of the first (non-retry) request in all the spots may get complex.

StripePayments/StripePayments/Source/PaymentHandler/PollingBudget.swift
porter-stripe: I think there's an edge case not being considered: we could be setting the network timeout to very small numbers like 0.1 seconds or something, and not even giving the network a chance to respond. You make a good point about optimizing for the cancel case. However, my concern is when we're very close to the polling duration limit, we might set timeouts so short (like 0.1-0.5 seconds) that we're essentially guaranteeing network failures, not due to actual network issues but simply because we didn't allow enough time for a round trip.

Maybe we could find a middle ground? For example, ensuring a max/min timeout of 1-2 seconds?

yuki-stripe: Yeah I'd also advocate for something like "if we are within 1 second of POLLING_DURATION, don't bother making a request, just wait until we hit POLLING_DURATION" if it's a bad idea to use network timeouts to achieve this.

porter-stripe: Added the behavior to set the network timeout to the duration remaining, or default to 3 seconds for the final poll.

yuki-stripe: I think final poll should be the default (60 seconds)!

porter-stripe: updated!

porter-stripe:
@yuki-stripe thanks for the feedback, should be ready for a look!

StripePayments/StripePayments/Source/PaymentHandler/PollingBudget.swift
yuki-stripe: > recording the network time of the first (non-retry) request

Is this not the start time?

StripePayments/StripePayments/Source/PaymentHandler/PollingBudget.swift
```diff
@@ -0,0 +1,113 @@
...
+    /// The elapsed time since polling started
+    private var elapsedTime: TimeInterval {
+        return Date().timeIntervalSince(startDate)
+    }
+
+    /// Dynamic network timeout for API requests based on remaining polling budget.
+    ///
+    /// This computed property optimizes network timeouts by adjusting them according to the remaining
+    /// time in the polling budget. It ensures that individual network requests don't exceed the total
+    /// time allocated for polling, while providing a minimum timeout for final polls (60 seconds).
```
yuki-stripe:
```suggestion
    /// time allocated for polling, while providing a timeout for final polls (60 seconds).
```

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
```diff
@@ -2047,10 +2067,9 @@ public class STPPaymentHandler: NSObject {
...
     }
 
-    static let maxChallengeRetries = 5
     func _markChallengeCompleted(
         withCompletion completion: @escaping STPBooleanSuccessBlock,
-        retryCount: Int = maxChallengeRetries
+        pollingBudget: PollingBudget? = nil
```
yuki-stripe: I think we should leave this code alone. The problem it's solving is quite different.

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
```diff
@@ -2140,11 +2160,13 @@ public class STPPaymentHandler: NSObject {
         } else {
             currentAction.apiClient.retrievePaymentIntent(withClientSecret: currentAction.paymentIntent.clientSecret,
                                                           expand: ["payment_method"],
+                                                          timeout: pollingBudget?.networkTimeout ?? 3, // Default to 3 seconds of network timeout when fetching an intent (first poll)
```
yuki-stripe: This is a bit dangerous; this method could be called anywhere in this class and "pollingBudget" only applies to a specific  part of the code. 

Also - why would we default to 3 seconds for the first poll?

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
```diff
@@ -1533,11 +1547,13 @@ public class STPPaymentHandler: NSObject {
                         if
                             let paymentMethod = paymentIntent.paymentMethod,
                             !STPPaymentHandler._isProcessingIntentSuccess(for: paymentMethod.type),
-                            paymentIntent.status == .processing && retryCount > 0
+                            paymentIntent.status == .processing,
+                            pollingBudget?.canPoll ?? true
                         {
-                            self._retryAfterDelay(retryCount: retryCount) {
+                            let processingPollingBudget = pollingBudget ?? PollingBudget(startDate: startDate, duration: 1)
```
yuki-stripe: Looking closer, we shouldn't use a 1 second timeout here. This is a different case than the LPM-cancel problem we're solving. In this case we're **sure** that the PaymentIntent will transition. The "polling budget" should be long, like 30 seconds.

StripePayments/StripePayments/Source/PaymentHandler/PollingBudget.swift
```diff
@@ -0,0 +1,113 @@
...
+        // During tests where we record network traffic, if we are hitting stubs we can make more network requests
+        // due to near-instant network responses, so we need to scale down the polling duration to account for this.
+        return isUsingNetworkStubs ? rawValue * 0.4 : rawValue
+    }
+
+    /// Detects if we're running in stub playback mode (HTTPStubs active but not recording)
+    private var isUsingNetworkStubs: Bool {
+        // Only scale duration when we're using stubs (HTTPStubs present and not recording the network STP_RECORD_NETWORK)
+        return NSClassFromString("HTTPStubs") != nil && ProcessInfo.processInfo.environment["STP_RECORD_NETWORK"] == nil
+    }
```
yuki-stripe: still not understanding why we need this, what happens if we delete this?

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
```diff
@@ -1475,13 +1475,25 @@ public class STPPaymentHandler: NSObject {
...
-        DispatchQueue.main.asyncAfter(deadline: .now() + delayTime) {
+    /// Schedules delayed execution of a retry block and records a poll attempt.
+    /// - Parameters:
+    ///   - delay: Time interval to wait before execution (used as target interval for optimization)
+    ///   - pollingBudget: Budget tracker for polling attempts
+    ///   - block: Block to execute if budget allows
+    func pollAfterDelay(delay: TimeInterval = 1, pollingBudget: PollingBudget, block: @escaping () -> Void) {
+        let optimizedDelay = pollingBudget.recommendedDelay(targetInterval: delay)
+        DispatchQueue.main.asyncAfter(deadline: .now() + optimizedDelay) {
+            pollingBudget.recordPollAttempt()
```
yuki-stripe: WDYT about moving this to the PollingBudget helper?  It'd make it impossible to accidentally not call `recordPollAttempt()`.

yuki-stripe: We should also delete `delay`.

StripePayments/StripePayments/Source/PaymentHandler/PollingBudget.swift
porter-stripe: Yeah, after this change we no longer need it, removed!
https://github.com/stripe/stripe-ios/pull/5519/commits/c9f4fe8b0dc1c0361f70f40a412175c1acba7348

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
porter-stripe: ack, updated!

StripePayments/StripePayments/Source/PaymentHandler/PollingBudget.swift
porter-stripe: oh good catch/idea, yes!

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
porter-stripe: Yeah that's a good idea, will do

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
porter-stripe: hmm, yeah I can get onboard with that.

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
```diff
@@ -1475,13 +1475,25 @@ public class STPPaymentHandler: NSObject {
...
         }
     }
 
-    func _retrieveAndCheckIntentForCurrentAction(currentAction: STPPaymentHandlerActionParams? = nil, retryCount: Int = maxChallengeRetries) {
+    /// Retrieves and checks the payment intent status for the current action.
+    /// If pollingBudget is nil, this is the first attempt and a new budget is created.
+    /// - Parameters:
+    ///   - currentAction: Action parameters to process, defaults to self.currentAction
+    ///   - pollingBudget: Existing polling budget, or nil for first attempt
+    func _retrieveAndCheckIntentForCurrentAction(currentAction: STPPaymentHandlerActionParams? = nil, pollingBudget: PollingBudget? = nil) {
```
yuki-stripe, line 1489: What happens if the network request times out? Do we enter into the error handling code and stop polling?

yuki-stripe: I understand the desire to refactor in place with the smallest changes possible.

But I wonder (really, haven't thought it through 100%) if it'd be easier / simpler
1. Don't make `_retrieveAndCheckIntentForCurrentAction` recursive
2. Rework the three users of `_retrieveAndCheckIntentForCurrentAction` to call something else that has the polling logic in a while loop.

porter-stripe: The behavior when the network timeout is unchanged from master. If the network request times out we retry if the budget allows, if it does not `canPoll` will be false in `_retrieveAndCheckIntentForCurrentAction`, and we will enter into the same error handling as before.

porter-stripe: The behavior when the network timeout is unchanged from master. If the network request times out we retry if the budget allows, if it does not canPoll will be false in `_retrieveAndCheckIntentForCurrentAction`, and we will enter into the same error handling as before.

The only difference in how we handle errors is that now we check `pollingBudget.canPoll` vs `retryCount > 0`

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
porter-stripe: Yeah I think defaulting to a timeout of 3 is dangerous, so removed this to just fallback to the default timeout of the networking client and refactored `pollingBudget` out of this function so it's just an optional timeout we can pass to the api client.

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
```diff
@@ -1527,11 +1527,21 @@ public class STPPaymentHandler: NSObject {
                         timeout: pollingBudget?.networkTimeout
                     ) { [self] paymentIntent, error in
                         guard let paymentIntent, error == nil else {
-                            let error = error ?? self._error(for: .unexpectedErrorCode, loggingSafeErrorMessage: "Missing PaymentIntent.")
-                            currentAction.complete(
-                                with: STPPaymentHandlerActionStatus.failed,
-                                error: error as NSError
-                            )
+                            // If we got an error retrieving the intent, retry if budget allows
+                            if let pollingBudget, pollingBudget.canPoll {
```
porter-stripe, line 1534: This will only retry _if_ we have a budget, which means it's a polling call, we won't retry if it's the first call. This is consistent w/ master retry logic, but we could do like `pollingBudget ?? PollingBudget(duration: someArbitraryDuration)` to retry first calls to.

yuki-stripe: I'm okay to call it good enough here but maybe we can add some more comments explaining, I'm hoping we can rewrite PaymentHandler next year and I don't want to lose track of this.

"This will only retry if we have a budget, which means it's a polling call, we won't retry if it's the first call. Ideally we retry on the first call too" or something

StripePayments/StripePayments/Source/PaymentHandler/PollingBudget.swift
```diff
@@ -0,0 +1,89 @@
...
+    func recommendedDelay(targetInterval: TimeInterval = 1.0) -> TimeInterval {
+        let timeSinceLastPoll = Date().timeIntervalSince(lastPollAttempt ?? startDate)
+        // If enough time has passed, poll immediately. Otherwise, wait for the remaining time.
+        return max(0, targetInterval - timeSinceLastPoll)
+    }
+
+    /// Executes a block after the recommended delay and automatically records the poll attempt
+    /// - Parameter block: Block to execute after the delay
+    func pollAfter(block: @escaping () -> Void) {
+        let optimizedDelay = recommendedDelay(targetInterval: 1.0)
```
yuki-stripe: maybe `stpAssert(canPoll)`?

StripePayments/StripePayments/Source/PaymentHandler/PollingBudget.swift
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    /// Calculates the recommended delay before the next poll
+    /// - Parameter targetInterval: The desired time interval between poll starts (default: 1.0 second)
+    /// - Returns: The time to wait before the next poll, optimized based on the last poll timing
+    func recommendedDelay(targetInterval: TimeInterval = 1.0) -> TimeInterval {
+        let timeSinceLastPoll = Date().timeIntervalSince(lastPollAttempt ?? startDate)
+        // If enough time has passed, poll immediately. Otherwise, wait for the remaining time.
+        return max(0, targetInterval - timeSinceLastPoll)
+    }
```
yuki-stripe: This stuff should probably be private and/or inlined (IMO `recommendedDelay` should be inlined), and the test can call `pollAfter` directly.

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
```diff
@@ -1635,10 +1667,10 @@ public class STPPaymentHandler: NSObject {
                             // If this is a web-based 3DS2 transaction that is still in requires_action, we may just need to refresh the SI a few more times.
                             // Also retry a few times for Cash App, the redirect flow is fast and sometimes the intent doesn't update quick enough
                             let shouldRetryForCard = paymentMethod.type == .card && setupIntent.nextAction?.type == .useStripeSDK
-                            if retryCount > 0, paymentMethod.type != .card || shouldRetryForCard, let pollingRequirement = paymentMethod.type.pollingRequirement {
-                                self._retryAfterDelay(retryCount: retryCount, delayTime: pollingRequirement.timeBetweenPollingAttempts) {
+                            if paymentMethod.type != .card || shouldRetryForCard, let pollingBudget = pollingBudget ?? .init(startDate: startDate, paymentMethodType: paymentMethod.type), pollingBudget.canPoll {
```
yuki-stripe: Should we still poll for cash app? here and for PIs.

porter-stripe: I don't know of any data that would indicate that we should poll for Cash App while it's using the refresh endpoint. AFAIK, the Cash App implementation of the refresh endpoint gives us the guarantee that the intent is in the correct status, I think it's other LPMs that can't make that guarantee.

StripePayments/StripePayments/Source/PaymentHandler/PollingBudget.swift
porter-stripe: Inlined recommendedDelay(...)!

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
yuki-stripe: Oh okay, let's update the comment then?

porter-stripe: Ah didn't see that comment, it's stale. Updated!

StripePayments/StripePayments/Source/PaymentHandler/PollingBudget.swift
porter-stripe: Also inlined recordPollAttempt

StripePayments/StripePayments/Source/PaymentHandler/PollingBudget.swift
```diff
@@ -6,6 +6,7 @@
 //
 
 import Foundation
+@_spi(STP) import StripeCore
 
 /// Manages polling budgets for payment method transactions that require status polling.
 final class PollingBudget {
```
yuki-stripe, line 12: This code is very un-threadsafe and will blow up if called concurrently :) I don't think STPPaymentHandler will ever do that though. We could maybe add assertMainThreads in some places but really this is just a note. Maybe something for the future refactor (🤞 ).

StripePayments/StripePayments/Source/PaymentHandler/PollingBudget.swift
```diff
@@ -0,0 +1,79 @@
...
+import Foundation
+@_spi(STP) import StripeCore
+
+/// Manages polling budgets for payment method transactions that require status polling.
+final class PollingBudget {
+    /// The timestamp when polling started (set when start() is called)
+    private var startDate: Date
+    /// The polling duration in seconds
+    private let duration: TimeInterval
+    /// Whether polling is currently allowed (within budget)
```
yuki-stripe:
```suggestion
    /// Whether polling is currently allowed
```

StripePayments/StripePayments/Source/PaymentHandler/PollingBudget.swift
```diff
@@ -0,0 +1,79 @@
...
+//
+//  Created by Nick Porter on 9/11/25.
+//
+
+import Foundation
+@_spi(STP) import StripeCore
+
+/// Manages polling budgets for payment method transactions that require status polling.
+final class PollingBudget {
+    /// The timestamp when polling started (set when start() is called)
```
yuki-stripe:
```suggestion
    /// The timestamp when polling started
```

StripePayments/StripePayments/Source/PaymentHandler/PollingBudget.swift
porter-stripe: Yep totally ack this! I think we make lots of assumptions in STPPaymentHandler about how it's used w.r.t to threading, like `inProgress` and `currentAction` are not thread safe either, and probably others. Maybe in the future STPPaymentHandler could be it's own actor.

yuki-stripe (approved):

PR #5521, opened 2025-09-11
Passive captcha test + playground switches

joyceqin-stripe:
## Summary

Added playground switches (oops).
Timeout doesn't happen on the prefetch— it happens on confirm-click.
Added tests and a duration field to the attach analytic
## Motivation

Bug bash

## Testing

Added tests

## Changelog

N/A


StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
```diff
@@ -49,26 +49,32 @@ private enum CaptchaResult {
...
     }
 
+    /// Test configuration - can be set by tests to override default behavior
+    private var testConfiguration: TestConfiguration?
+
+    public struct TestConfiguration {
+        let timeout: UInt64
+        let delayValidation: Bool
+
+        public init(timeout: UInt64, delayValidation: Bool = false) {
```
gbirch-stripe: what's the timeout unit? Could we either specify in the param name, or probably better use something like `TimeInterval` that's human understandable?

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
```diff
@@ -49,26 +49,32 @@ private enum CaptchaResult {
...
+        }
+    }
+
     private let passiveCaptcha: PassiveCaptcha?
     private var validationTask: Task<String?, Never>?
-    private let testTimeout: UInt64?
     private var isValidationComplete = false
 
-    public init(passiveCaptcha: PassiveCaptcha?, testTimeout: UInt64? = nil) {
+    public init(passiveCaptcha: PassiveCaptcha?, testConfiguration: TestConfiguration? = nil) {
```
gbirch-stripe: I don't love having test functionality coupled with the actual code, here in the init and in `start()`. Is there some way we could refactor this?

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
```diff
@@ -77,32 +83,20 @@ private enum CaptchaResult {
...
-                    group.addTask {
-                        try? await Task.sleep(nanoseconds: timeoutNs)
-                        return .error(Error.timeout)
-                    }
-                    // Wait for first completion and cancel remaining tasks
-                    let result = await group.next()
-                    group.cancelAll()
-                    return result
+                }
+                if let testConfiguration, testConfiguration.delayValidation {
```
gbirch-stripe: Why were we previously running the timeout and result tasks concurrently but now serially?

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
```diff
@@ -129,17 +120,43 @@ private enum CaptchaResult {
...
+            }
+            // Add timeout task
+            group.addTask {
+                try? await Task.sleep(nanoseconds: timeoutNs)
+                return nil
+            }
+            // Wait for first completion and cancel remaining tasks
+            let result = await group.next()
+            group.cancelAll()
+            if let result, result != nil {
```
gbirch-stripe: `result != nil` is redundant here, right?

joyceqin-stripe: result returns as a String??, and let result ends up being a String?

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
joyceqin-stripe: We're moving where the timeout happens. Previously, we would try to validate in the background for a few seconds and return nil if it timed out. Now, we're not activating the timeout until the customer has tapped the pay button, which gives the validation more time to return a token because the first HCaptcha load takes a long time

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
```diff
@@ -231,6 +231,9 @@ extension PaymentSheet {
         /// If true, the card form will instead initialize with the card scanner already open.
         public var opensCardScannerAutomatically: Bool = false
 
+        /// If true, an invisible challenge will be performed for human verification
+        @_spi(Captcha) public var enablePassiveCaptcha: Bool = false
```
yuki-stripe: what's a good place to learn more about this new property, i didn't see it in the eng brief

joyceqin-stripe: This should not have had its own thing, it can just be STP. This is just for playground purposes to enable/disable passive captcha

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetLoaderTest.swift
```diff
@@ -477,10 +477,12 @@ final class PaymentSheetLoaderTest: STPNetworkStubbingTestCase {
     }
 
     func testPassiveCaptchaDoesNotBlockLoad() async throws {
+        // OCS mobile test key from https://dashboard.hcaptcha.com/sites/edit/143aadb6-fb60-4ab6-b128-f7fe53426d4a
         let siteKey = "143aadb6-fb60-4ab6-b128-f7fe53426d4a"
```
gbirch-stripe: Is this safe to publicly share?

gbirch-stripe (dismissed):

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetLoaderTest.swift
joyceqin-stripe: Yes, in the actual code we use the key that's passed from the elements/sessions response. This key isn't actually used

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetLoaderTest.swift
```diff
@@ -475,4 +475,20 @@ final class PaymentSheetLoaderTest: STPNetworkStubbingTestCase {
...
+        // OCS mobile test key from https://dashboard.hcaptcha.com/sites/edit/143aadb6-fb60-4ab6-b128-f7fe53426d4a
+        let siteKey = "143aadb6-fb60-4ab6-b128-f7fe53426d4a"
+        let passiveCaptcha = PassiveCaptcha(siteKey: siteKey, rqdata: nil)
+        // even if validation takes a very long time, the payment sheet will load in a normal amount of time
+        let passiveCaptchaChallenge = PassiveCaptchaChallenge(passiveCaptcha: passiveCaptcha)
+        await passiveCaptchaChallenge.setTestConfiguration(PassiveCaptchaChallenge.TestConfiguration(timeout: 6, delay: 30))
+        await passiveCaptchaChallenge.start()
+        let expectation = XCTestExpectation(description: "Load")
+        let clientSecret = try await STPTestingAPIClient.shared.fetchPaymentIntent(types: ["card"])
+        PaymentSheetLoader.load(mode: .paymentIntentClientSecret(clientSecret), configuration: self.configuration, analyticsHelper: .init(integrationShape: .complete, configuration: configuration), integrationShape: .complete) { _ in
```
yuki-stripe: I'm not sure this test is really testing anything. The code that doesn't block load is not being tested here - that code is in PaymentSheet, FlowController, and Embedded (the line where they call `passiveCaptchaChallenge.start`).

I think to write _that_ test you'd want to stub out the captcha request to have a long delay (eg 60s) before returning, then initialize/load the integrations and assert they finish loading quickly.

TBH though given the simplicity of the code being tested - you're basically checking that you wrapped the `await` in a Task - I'm not super excited about spending time on it. Maybe just writing a stern comment ("Intentionally not blocking loading/initialization!") next to each is good enough.

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
```diff
@@ -49,26 +49,35 @@ private enum CaptchaResult {
...
+        public init(timeout: TimeInterval = 0) {
+            self.timeout = timeout
+        }
+    }
+
+    func setTestConfiguration(_ testConfiguration: TestConfiguration) {
+        if STPAnalyticsClient.isUnitOrUITest {
+            self.testConfiguration = testConfiguration
+        }
+    }
```
yuki-stripe: Nit: It'd be simpler if this was just `var timeout: TimeInterval = 6`. Feels like a valid settable property?

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
```diff
@@ -129,17 +120,43 @@ private enum CaptchaResult {
...
+        var isReady = false
+        return await withTaskGroup(of: String?.self) { group in
+            // Add hcaptcha task
+            group.addTask { [weak self] in
+                guard let self else { return nil }
+                if await validationTask == nil {
+                    await self.start()
+                } else {
+                    isReady = await isValidationComplete
+                }
```
yuki-stripe: Is the else case an expected state? I thought we need all users to be calling start ahead of fetchToken.

yuki-stripe: And if `validationTask == nil`, `isValidationComplete` is always false. What's the purpose of `isReady`?

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
```diff
@@ -129,17 +120,43 @@ private enum CaptchaResult {
         isValidationComplete = true
     }
 
-    @_spi(STP) public func fetchToken() async -> String? {
+    public func fetchToken() async -> String? {
         guard let siteKey = passiveCaptcha?.siteKey else { return nil }
```
yuki-stripe: This class doesn't make sense / have any purpose if `passiveCaptcha` is nil, right? If so it's probably cleaner and clearer to disallow the creation of it with a nil passiveCaptcha (ie make it non-nullable) rather than have all its methods account for this state.

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
```diff
@@ -129,17 +120,43 @@ private enum CaptchaResult {
...
+                return await validationTask?.value
+            }
+            // Add timeout task
+            group.addTask {
+                try? await Task.sleep(nanoseconds: timeoutNs)
+                return nil
+            }
+            // Wait for first completion and cancel remaining tasks
+            let result = await group.next()
+            group.cancelAll()
```
yuki-stripe: Just a note: this won't actually do anything - you'd have to check cancellation in the task and write code to 'cancel' remaining steps or whatever.

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
```diff
@@ -129,17 +120,43 @@ private enum CaptchaResult {
...
+            if let result, result != nil {
+                STPAnalyticsClient.sharedClient.logPassiveCaptchaAttach(siteKey: siteKey, isReady: isReady, duration: Date().timeIntervalSince(startTime) * 1000)
+                return result
+            } else {
+                STPAnalyticsClient.sharedClient.logPassiveCaptchaError(error: Error.timeout, siteKey: siteKey, duration: Date().timeIntervalSince(startTime) * 1000)
```
yuki-stripe: Are you sure this should be * 1000? TimeInterval is in seconds.

StripePayments/StripePaymentsTests/Captcha/PassiveCaptchaTests.swift
```diff
@@ -9,20 +9,59 @@
...
+    }
+
+    func testPassiveCaptchaIsReadyOnFetchToken() async throws {
+        let siteKey = "143aadb6-fb60-4ab6-b128-f7fe53426d4a"
+        let passiveCaptcha = PassiveCaptcha(siteKey: siteKey, rqdata: nil)
+        let passiveCaptchaChallenge = PassiveCaptchaChallenge(passiveCaptcha: passiveCaptcha)
+        await passiveCaptchaChallenge.setTestConfiguration(PassiveCaptchaChallenge.TestConfiguration(timeout: 6))
+        await passiveCaptchaChallenge.start()
+        // wait to make sure that the token will be ready by the time we call fetchToken
+        try await Task.sleep(nanoseconds: 6_000_000_000)
```
yuki-stripe, line 24: This should probably be a network recording test.

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
joyceqin-stripe: Realizing that I could just set `isReady` to `isValidationComplete` from the get-go. `isReady` is meant to indicate whether the validation is already done executing by the time fetchToken is called, so basically whether or not passive captcha is forcing us to wait.

I've also moved `start()` into the initializer so it gets called immediately and we don't need to explicitly call it in PaymentSheet, FlowController, Embedded, etc

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
joyceqin-stripe: The [eng brief](https://docs.google.com/document/d/1VDwjpPnKxxZo9liOC4k9ccrO-fLgFaplNAuplYm2HDY/edit?tab=t.sunyar73psj2#heading=h.iehlz0r4szg4) metrics say duration in ms

StripePayments/StripePaymentsTests/Captcha/PassiveCaptchaTests.swift
joyceqin-stripe: Couldn't figure out how to make it a network test

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
yuki-stripe: >  isReady is meant to indicate whether the validation is already done executing by the time fetchToken is called, so basically whether or not passive captcha is forcing us to wait.

Ah ok, makes sense. Seems fine to me, although wondering why you can't just solely look at duration to determine that.

Example/PaymentSheet Example/PaymentSheet Example/CustomerSheetTestPlaygroundController.swift
```diff
@@ -153,6 +153,7 @@ class CustomerSheetTestPlaygroundController: ObservableObject {
             configuration.cardBrandAcceptance = .allowed(brands: [.visa])
         }
         configuration.opensCardScannerAutomatically = settings.opensCardScannerAutomatically == .on
+        configuration.enablePassiveCaptcha = settings.enablePassiveCaptcha == .on
```
yuki-stripe, line 156: I can't remember - did we add this flag b/c UI tests were breaking when passive captcha was enabled?

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
```diff
@@ -129,20 +111,41 @@ private enum CaptchaResult {
...
+                validationTask?.cancel()
+            }
+            // Wait for first completion
+            let token: String? = await group.next().flatMap(\.self)
+            let siteKey = passiveCaptcha.siteKey
+            if let token {
+                STPAnalyticsClient.sharedClient.logPassiveCaptchaAttach(siteKey: siteKey, isReady: isReady, duration: Date().timeIntervalSince(startTime))
+                return token
+            } else {
+                STPAnalyticsClient.sharedClient.logPassiveCaptchaError(error: Error.timeout, siteKey: siteKey, duration: Date().timeIntervalSince(startTime))
```
yuki-stripe: Not necessarily a timeout error, right?

StripePayments/StripePaymentsTests/Captcha/PassiveCaptchaTests.swift
yuki-stripe: (We talked offline - can't find a way to get OHTTPStubs to hook into the captcha SDK's webview network stack)

I wonder if this will be flakey given it's hitting the live network :hm: - something to keep an eye on. We could build in some resiliency by retrying a few times if so.

Example/PaymentSheet Example/PaymentSheet Example/CustomerSheetTestPlaygroundController.swift
joyceqin-stripe: No, they weren't breaking. It was mostly to make it easier to test behavior with and without passive captcha, and the Android playground also has a switch, but if you think it would be better not to add this I'm peaceful with removing it.

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
joyceqin-stripe: You're right, a nil token could also mean an error from hcaptcha initialization or validation. I've updated the Task to return a CaptchaResult so we know whether it's nil due to timeout or otherwise, and we only want to log the timeout here (the previous errors would have been logged in `start`)

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
```diff
@@ -49,78 +49,63 @@ private enum CaptchaResult {
...
-                                hcaptcha.validate { result in
-                                    do {
-                                        let token = try result.dematerialize()
-                                        continuation.resume(returning: .success(token))
-                                    } catch {
-                                        continuation.resume(returning: .error(error))
-                                    }
-                                }
+                let result = await withCheckedContinuation { (continuation: CheckedContinuation<CaptchaResult, Never>) in
+                    // Prevent Swift Task continuation misuse
```
yuki-stripe: Can we also add - hcaptcha `validate` can unexpectedly be called multiple times.

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
```diff
@@ -50,7 +50,7 @@ private enum CaptchaResult {
     }
 
     private let passiveCaptcha: PassiveCaptcha
-    private var validationTask: Task<String?, Never>?
+    private var validationTask: Task<CaptchaResult?, Never>?
```
yuki-stripe: I think this might be more idiomatically a `Task<String, Swift.Error>` but let's figure that out in a follow-up PR.

yuki-stripe (approved):

yuki-stripe (approved):

PR #5527, opened 2025-09-11
Adjust when hcaptcha token fetched

joyceqin-stripe:
## Summary

Instead of fetching the token before calling confirm from PaymentSheet, FlowController. Embedded, CustomerSheet regardless of payment method and passing that around, we pass the PassiveCaptchaChallenge and call fetchToken only when it's needed.
Omit it for ApplePay and saved pms.
## Motivation

First captcha load takes several seconds. This disrupts the user experience for quick payment methods that don't require users to input payment details (like saved pms, ApplePay). As captcha is meant to guard against card testing, neither of these cases need passive captcha. Discussion on Slack [here](https://stripe.slack.com/archives/C08SZ67N0NM/p1757615895108549)

## Testing

Manual testing, CI. Tried to add network tests in StripePaymentSheetTests, but HCaptcha validation doesn't seem to work in test environments without a window (Webview can't load)
## Changelog

N/A


StripePayments/StripePayments/Source/API Bindings/Models/PaymentIntents/STPPaymentIntentParams.swift
```diff
@@ -295,7 +295,6 @@ extension STPPaymentIntentParams: NSCopying {
         copy.receiptEmail = receiptEmail
         copy.savePaymentMethod = savePaymentMethod
         copy.setAsDefaultPM = setAsDefaultPM
-        copy.radarOptions = radarOptions
```
joyceqin-stripe, line 304: Cleanup, found that copy.radarOptions = radarOptions is in here twice

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
```diff
@@ -190,83 +189,87 @@ extension PaymentSheet {
...
-                )
-                paymentHandler.confirmPayment(
-                    params,
-                    with: authenticationContext,
-                    completion: { actionStatus, paymentIntent, error in
-                        if let paymentIntent {
-                            setDefaultPaymentMethodIfNecessary(actionStatus: actionStatus, intent: .paymentIntent(paymentIntent), configuration: configuration, paymentMethodSetAsDefault: elementsSession.paymentMethodSetAsDefaultForPaymentSheet)
-                        }
-                        paymentHandlerCompletion(actionStatus, error)
+            Task {
```
davidme-stripe: Is this running on the main actor? Not sure if everything here is safe to run on a background thread

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
```diff
@@ -329,127 +330,135 @@ extension PaymentSheet {
...
-                    paymentIntentParams.radarOptions = radarOptions
-                    paymentIntentParams.mandateData = mandateData
-                    paymentHandler.confirmPayment(
-                        paymentIntentParams,
-                        with: authenticationContext,
-                        completion: { actionStatus, _, error in
-                            if actionStatus == .succeeded {
-                                linkAccount?.logout()
+                Task {
+                    let hcaptchaToken = await passiveCaptchaChallenge?.fetchToken()
```
davidme-stripe, line 397: Could this happen in the SPM flow as well? Having trouble following this code

davidme-stripe (approved):

gbirch-stripe (approved):

PR #5530, opened 2025-09-12
Fix divider visibility in `MultiElementRow`

tillh-stripe:
## Summary

This pull request fixes an issue where dividers between elements could stay visible in `MultiElementRow` even if the adjacent elements are set to be hidden.

~We now only store the actual views (and not the dividers) and re-compute the entire list of `arrangedSubviews` whenever an element is toggled.~

## Motivation

Polish.

## Testing

Updated snapshots.


StripeUICore/StripeUICore/Source/Elements/Section/SectionContainerView.swift
```diff
@@ -198,29 +198,32 @@ extension SectionContainerView: EventHandler {
 
 extension SectionContainerView {
     class MultiElementRowView: UIView {
+        private let theme: ElementsAppearance
+        private let views: [UIView]
+
+        private var multiElementStackView: UIStackView?
```
yuki-stripe: make this a let and non-optional or just delete it

yuki-stripe (changes requested):

StripeUICore/StripeUICore/Source/Elements/Section/SectionContainerView.swift
tillh-stripe: Removed!

mats-stripe (approved):

PR #5531, opened 2025-09-12
Onramp SDK analytics

mats-stripe:
## Summary

Implements analytics logging in the various StripeCryptoOnramp APIs. 

## Motivation

Analytics are described in this doc https://docs.google.com/document/d/1CY5IddahvJQIOw0FPTIz4SE4BLETnXjJDRE0oKcOioY/edit?usp=sharing

## Testing

N/a

## Changelog

N/a

### DIFF
--- a/StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
+++ b/StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
@@ -209,6 +209,7 @@ public final class CryptoOnrampCoordinator: NSObject, CryptoOnrampCoordinatorPro
                 analyticsClient: analyticsClient
             )
 
+            analyticsClient.elementsSessionId = await linkController.elementsSessionID
             analyticsClient.log(.sessionCreated)
             return coordinator
         } catch {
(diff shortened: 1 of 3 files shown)


StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
mats-stripe, line 1: I had to tap `Load Diff` to see the changes in this file

StripeCryptoOnramp/StripeCryptoOnramp/Source/Internal/CryptoOnrampAnalyticsClient.swift
```diff
@@ -0,0 +1,37 @@
...
+            clientId: "mobile-onramp-sdk",
+            origin: "stripe-onramp-ios"
+        )
+    ) {
+        self.analyticsClient = analyticsClient
+
+        if AnalyticsHelper.shared.sessionID == nil {
+            AnalyticsHelper.shared.generateSessionID()
+        }
+        additionalParameters["session_id"] = AnalyticsHelper.shared.sessionID ?? "N/a"
```
tillh-stripe: Why `N/a`? As far as I can see, we simply send `null` on Android.

mats-stripe: We can do null here if preferred, but that should never be hit based on the above check

tillh-stripe: Follow-up: Android was sending the Elements session ID, not the local session ID. Already fixed in latest commit.

PR #5534, opened 2025-09-12
Fallback to web auth flow if we can't authenticate the Link user natively

mats-stripe:
## Summary

Adds a fallback web-based Link authentication flow, used when we can't authenticate the Link user natively.

The general flow is:

1. Pass the known supported verification factors to the `/lookup` call
2. In the response, we will get back `mobile_fallback_webview_params`, which will tell us if we can or can't verify the user natively, based on the required verification factors for the current Link user
3. Using the URL provided in `mobile_fallback_webview_params.webview_open_url`, we launch a secure webview. This will authenticate the Link user
4. We observe for redirects to `link-popup://complete`, and dismiss the webview when detected
5. Call a new `/refresh` endpoint to get the latest consumer session back on the client.

## Motivation

https://docs.google.com/document/d/1Yxlz__5R4malxex123db3jyh8KnGaedBxLBUEEu2a0s/edit?usp=sharing

## Testing

> [!NOTE]  
>To test out the fallback flow yourself, set a random value (such as `["abc"]`) as the `supported_verification_types` in the mobile `/lookup` call [here](https://github.com/stripe/stripe-ios/pull/5534/files#diff-37b1d3022a0265c5ae1a48d7f4f2027426df1271f4c2b03bacb2ce6bc5b7e3b4R95)

https://github.com/user-attachments/assets/cc395110-9514-4f6a-a041-753766f6979c

## Changelog

N/a

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Verification/LinkVerificationController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Verification/LinkVerificationController.swift
@@ -66,8 +66,11 @@ final class LinkVerificationController {
     }
 
     private func presentMobileFallbackWebview(from presentingController: UIViewController, webviewUrl: URL) {
-        let webviewController = LinkVerificationWebFallbackController(authenticationUrl: webviewUrl)
-        webviewController.present(from: presentingController) { [weak self] result in
+        let webviewController = LinkVerificationWebFallbackController(
+            authenticationUrl: webviewUrl,
+            presentingWindow: presentingController.view.window
+        )
+        webviewController.present { [weak self] result in
             guard let self = self else { return }
 
             // If verification completed successfully, refresh the session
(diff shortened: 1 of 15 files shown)


StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Verification/LinkVerificationController.swift
```diff
@@ -43,7 +45,48 @@ final class LinkVerificationController {
...
+        let webviewController = LinkVerificationWebFallbackController(
+            authenticationUrl: webviewUrl,
+            presentingWindow: presentingController.view.window
+        )
+        webviewController.present { [weak self] result in
+            guard let self = self else { return }
+
+            // If verification completed successfully, refresh the session
+            if case .completed = result {
+                self.linkAccount.refresh { refreshResult in
```
tillh-stripe: Do we need another weak self here?

mats-stripe:
Some more testing, including dismissing the webview and reopening it, and completing a checkout:


https://github.com/user-attachments/assets/f845453e-6b01-4dc4-a70b-fb47b5fa5595

tillh-stripe (approved):

PR #5539, opened 2025-09-15
Update iOS26 default form insets

wooj-stripe:
## Summary
Aligns form and other components to 16px for liquid glass design.

## Motivation
iOS26 liquid glass design

## Testing
Snapshot tests


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/PaymentSheetUIKitAdditions.swift
```diff
@@ -37,6 +37,10 @@ enum PaymentSheetUI {
     static let delayBetweenSuccessAndDismissal: TimeInterval = 1.5
     static let minimumHitArea = CGSize(width: 44, height: 44)
 
+    // MARK: - Liquid Glass
+    static let liquidGlassSheetMargins: NSDirectionalEdgeInsets = .insets(
+        leading: 16, bottom: 40, trailing: 16)
+
```
yuki-stripe: nit: I think `applyLiquidGlass` is the right place for this to live, it is the single source of truth for defining 'liquid glass sheet margins' along with all the other liquid glass appearance values. We don't even use (or want to use) this in multiple places so i feel like we're just adding a layer of indirection for no benefit.

PR #5549, opened 2025-09-16
[iOS 26] Make cornerRadius nullable

yuki-stripe:
## Summary
- Made `appearance.cornerRadius` nullable.
- Lots of other fixes

## Motivation
https://docs.google.com/document/d/15KZBP1FVkXtV6E_H-sdZRE0YwrnIsNkj6x0vHi9cgtY/edit?tab=t.0

## Testing
Existing tests.

## Changelog
See changelog


Example/PaymentSheet Example/PaymentSheet Example/AppearancePlaygroundView.swift
```diff
@@ -82,8 +83,8 @@ struct AppearancePlaygroundView: View {
         )
 
         let cornerRadiusBinding = Binding(
-            get: { self.appearance.cornerRadius },
-            set: { self.appearance.cornerRadius = $0 }
+            get: { self.appearance.cornerRadius ?? -1 },
+            set: { self.appearance.cornerRadius = $0 < 0 ? nil : $0 }
```
yuki-stripe, line 89: using -1 to represent 'nil'
CleanShot 2025-09-16 at 09 09 09@2x

Example/PaymentSheet Example/PaymentSheet Example/AppearancePlaygroundView.swift
```diff
@@ -172,14 +173,26 @@ struct AppearancePlaygroundView: View {
         )
 
         let regularFontBinding = Binding(
-            get: { self.appearance.font.base.fontDescriptor.postscriptName },
-            set: { self.appearance.font.base = UIFont(name: $0, size: 12.0)! }
```
yuki-stripe, line 176: This font change and the others fix a bug with the playground that causes an exception: "Picker: the selection ".SFUI-Regular" is invalid and does not have an associated tag, this will give undefined results."

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetAppearance.swift
```diff
@@ -56,10 +61,10 @@ public extension PaymentSheet {
 
         /// The insets used for all input fields (e.g. textfields, dropdowns) in PaymentSheet.
         @_spi(AppearanceAPIAdditionsPreview)
-        public var textFieldInsets: NSDirectionalEdgeInsets = .insets(top: 4, leading: 11, bottom: 4, trailing: 11)
+        public var textFieldInsets: NSDirectionalEdgeInsets = NSDirectionalEdgeInsets(top: 4, leading: 11, bottom: 4, trailing: 11)
 
         /// Describes the padding used for all forms
-        public var formInsets: NSDirectionalEdgeInsets = PaymentSheetUI.defaultSheetMargins
+        public var formInsets: NSDirectionalEdgeInsets = NSDirectionalEdgeInsets(top: 0, leading: 20, bottom: 40, trailing: 20)
```
yuki-stripe, line 67: Avoids using non-public methods, makes it easier for readers to understand + makes Appearance the source of truth for default appearance values.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetAppearance.swift
```diff
@@ -454,25 +462,3 @@ public extension PaymentSheet.Appearance {
         case outlined
     }
 }
-
```
yuki-stripe, line 460: These were moved to PaymentSheetAppearance+Helpers.swift

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/WalletButtonsView/LinkInlineVerificationView.swift
```diff
@@ -82,7 +82,8 @@ struct LinkInlineVerificationView: View {
                         configuration: .init(
                             enableDigitGrouping: false,
                             font: LinkUI.font(forTextStyle: .title),
-                            itemCornerRadius: viewModel.appearance.cornerRadius,
+                            // TODO(iOS 26): Is this supposed to apply a Liquid Glass style?
+                            itemCornerRadius: viewModel.appearance.cornerRadius ?? 6.0,
```
yuki-stripe: cc @tillh-stripe - it looks like this hasn't been updated to use Liquid Glass corners. This PR would keep that behavior, so if you set `cornerRadius = nil` it uses the pre-iOS 26 default of 6.

StripeUICore/StripeUICore/Source/Elements/Form/FormView.swift
```diff
@@ -27,7 +27,8 @@ import UIKit
             stack.customBackgroundColor = viewModel.theme.colors.componentBackground
             stack.separatorColor = viewModel.theme.colors.divider
             stack.borderColor = viewModel.theme.colors.border
-            stack.borderCornerRadius = viewModel.theme.cornerRadius
+            // TODO(iOS 26): Does this need to respect corner radius = nil?
+            stack.borderCornerRadius = viewModel.theme.cornerRadius ?? 6.0
```
yuki-stripe, line 31: @tillh-stripe Link is the only user of this code (and it really shouldn't use FormView at all like this, but that's a separate discussion). I'm preserving the current behavior, where it doesn't ever set Liquid Glass corners.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/RemoveButton.swift
```diff
@@ -43,16 +43,10 @@ class RemoveButton: UIButton {
...
-            // Workaround:  Use layer.borderWidth because background.strokeWidth isn't compatible with .capsule()
-            layer.borderWidth = strokeWidth > 0 ? strokeWidth : 1.5
-        } else {
-            layer.borderWidth = strokeWidth
-            layer.cornerRadius = appearance.primaryButton.cornerRadius ?? appearance.cornerRadius
-        }
+        applyCornerRadius(appearance: appearance, ios26DefaultCornerStyle: .capsule, shouldUsePrimaryButtonCornerRadius: true)
+        // Workaround:  Use layer.borderWidth because background.strokeWidth isn't compatible with .capsule()
+        layer.borderWidth = 1.5
+
```
yuki-stripe: cc @gbirch-stripe - I don't think this should use `appearance` for its outline border width so I changed it to always be 1.5

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## 24.23.2 2025-09-16
 ## PaymentSheet
 * [Fixed] Fixed an issue with the Liquid Glass keyboard toolbar where tapping on the sheet immediately above the keyboard didn't have an effect.
+* [Changed] `appearance.cornerRadius` is now nullable. When `nil`, the corner radius is 6.0 pre-iOS 26 (ie unchanged from the current default). On iOS 26+, multiple rounder corner configurations are used to match Liquid Glass design.
```
wooj-stripe: Sentence 2 & 3, maybe:
```
Pre-iOS26 when `nil`, the corner radius is 6.0. On iOS 26+, multiple...
```

Example/PaymentSheet Example/PaymentSheet Example/AppearancePlaygroundView.swift
```diff
@@ -82,8 +83,8 @@ struct AppearancePlaygroundView: View {
         )
 
         let cornerRadiusBinding = Binding(
-            get: { self.appearance.cornerRadius },
-            set: { self.appearance.cornerRadius = $0 }
+            get: { self.appearance.cornerRadius ?? -1 },
```
wooj-stripe, line 88: Should probably add comment mentioning that `-1` is a sentinel value for `nil`

Example/PaymentSheet Example/PaymentSheet Example/AppearancePlaygroundView.swift
```diff
@@ -97,8 +98,8 @@ struct AppearancePlaygroundView: View {
         )
 
         let selectedBorderWidthBinding = Binding(
-            get: { appearance.selectedBorderWidth ?? appearance.borderWidth * 1.5 },
-            set: { self.appearance.selectedBorderWidth = $0 }
+            get: { appearance.selectedBorderWidth ?? -0.5 },
```
wooj-stripe, line 104: Maybe add a comment here documenting that -0.5 is a sentinel value used for `nil`

Example/PaymentSheet Example/PaymentSheet Example/AppearancePlaygroundView.swift
```diff
@@ -237,8 +250,10 @@ struct AppearancePlaygroundView: View {
         )
 
         let primaryButtonFontBinding = Binding(
-            get: { self.appearance.primaryButton.font?.fontDescriptor.postscriptName ?? UIFont.systemFont(ofSize: 16, weight: .medium).fontDescriptor.postscriptName },
-            set: { self.appearance.primaryButton.font = UIFont(name: $0, size: 16.0)! }
+            get: { self.appearance.primaryButton.font?.fontDescriptor.postscriptName ?? "System Default" },
```
wooj-stripe: Nit: since we've referenced this 3+ times -- maybe we just add:
```
    static let systemDefaultFont = "System Default"
```

Example/PaymentSheet Example/PaymentSheet Example/AppearancePlaygroundView.swift
```diff
@@ -400,11 +430,21 @@ struct AppearancePlaygroundView: View {
                         ColorPicker("textColor", selection: primaryButtonTextColorBinding)
                         ColorPicker("borderColor", selection: primaryButtonBorderColorBinding)
                         Stepper("borderWidth: \(Int(appearance.primaryButton.borderWidth))", value: primaryButtonCornerBorderWidth, in: 0...30)
-                        Stepper("cornerRadius: \(Int(appearance.primaryButton.cornerRadius ?? appearance.cornerRadius))",
-                                value: primaryButtonCornerRadiusBinding, in: 0...30)
+                        let cornerRadiusLabel: String = {
+                            if let cornerRadius = appearance.primaryButton.cornerRadius {
+                                String(format: "cornerRadius: %.1f", cornerRadius)
```
wooj-stripe: for consistency, this hsould probably be: `%.0f`

Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
```diff
@@ -15,6 +15,7 @@ import PassKit
 @_spi(STP) import StripeCore
 @_spi(STP) import StripePayments
 @_spi(CustomerSessionBetaAccess) @_spi(STP) @_spi(PaymentSheetSkipConfirmation) @_spi(ExperimentalAllowsRemovalOfLastSavedPaymentMethodAPI) @_spi(CustomPaymentMethodsBeta) @_spi(PaymentMethodOptionsSetupFutureUsagePreview) import StripePaymentSheet
+@_spi(STP) import StripeUICore
```
wooj-stripe: still needed?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/WalletHeaderView.swift
```diff
@@ -63,11 +63,12 @@ extension PaymentSheetViewController {
 
         private lazy var payWithLinkButton: PayWithLinkButton = {
             let button = PayWithLinkButton()
-            if LiquidGlassDetector.isEnabled {
+            if appearance.cornerRadius == nil, LiquidGlassDetector.isEnabledInMerchantApp {
                 button.ios26_applyCapsuleCornerConfiguration()
             } else {
-                button.cornerRadius = appearance.cornerRadius
+                button.cornerRadius = appearance.cornerRadius ?? 6
```
wooj-stripe: Should we have some centralized place where we have logic that does `appearance.cornerRadius ?? 6`?

StripeUICore/StripeUICore/Source/Helpers/LiquidGlassHelpers.swift
```diff
@@ -53,17 +50,30 @@ import UIKit
...
-        stpAssert(LiquidGlassDetector.isEnabled)
 #if compiler(>=6.2)
         if #available(iOS 26.0, *) {
             cornerConfiguration = .uniformCorners(radius: 26)
         }
 #endif
     }
+
+    /// Convenience method that returns whether or not `cornerConfiguration` was set.
+    /// You'd think this would be nil, but it's not nullable!
```
wooj-stripe: I'm unclear about this comment: "you'd think this would be nil..."

Tests/ReferenceImages_64/StripePaymentSheetTests.UpdatePaymentMethodViewControllerSnapshotTests/test_UpdatePaymentMethodViewControllerSEPADebitAppearance_iOS26@3x.png
wooj-stripe, line 1: Multiple of the update payment method snapshot tests seem to no longer have the capsule style fields -- is this expected?  I wouldn't have expected that we'd need to re-record these.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/WalletHeaderView.swift
yuki-stripe: We have a helper method that does this but `WalletHeaderView` and one other place I'm forgetting do non-standard things - in this case, sets `button.cornerRadius` instead of `button.layer.cornerRadius`. 

Open to ideas. I can pull out the 6 into a `static let defaultPreiOS26CornerRadius: CGFloat = 6` property?

Tests/ReferenceImages_64/StripePaymentSheetTests.UpdatePaymentMethodViewControllerSnapshotTests/test_UpdatePaymentMethodViewControllerSEPADebitAppearance_iOS26@3x.png
yuki-stripe: This custom appearance ("ms paint") has a corner radius of 0.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/WalletHeaderView.swift
wooj-stripe: Initially I was thinking we do something like this:
```
internal extension PaymentSheet.Appearance {
    var cornerRadiusOrDefault: CGFloat {
        cornerRadius ?? 6.0
    }
}
```
but, i think the intent is less clear.  `static let defaultPreiOS26CornerRadius: CGFloat = 6` is at least better.

StripeUICore/StripeUICore/Source/Helpers/LiquidGlassHelpers.swift
```diff
@@ -77,3 +88,30 @@ import UIKit
...
 }
+// MARK: - UIView helpers
+extension UIView {
+    @_spi(STP) public enum CornerStyle {
+        case capsule
+        case uniform
+    }
+
+    /// A convenience method that looks at `appearance.cornerRadius` and applies it or, if `nil`, handles what the default value should be.
+    @_spi(STP) public func applyCornerRadius(appearance: ElementsAppearance, ios26DefaultCornerStyle: CornerStyle = .uniform) {
```
wooj-stripe, line 105: I think this function would be clearer if we changed the signature to:
```
    @_spi(STP) public func applyCornerRadius(cornerRadius: CGFloat?, iOS26DefaultCornerStyle: CornerStyle = .uniform) {
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetAppearance+Helpers.swift
```diff
@@ -0,0 +1,44 @@
...
+    }
+    var isPlain: Bool {
+        return self == .plain
+    }
+}
+
+extension UIView {
+    /// A convenience method that looks at `appearance.cornerRadius` and applies it or, if `nil`, handles what the default value should be.
+    /// - Parameter shouldUsePrimaryButtonCornerRadius: If `true`, `appearance.primaryButton.cornerRadius` (if non-nil) takes precedence over `appearance.cornerRadius`.
+    func applyCornerRadius(appearance: PaymentSheet.Appearance, ios26DefaultCornerStyle: CornerStyle = .uniform, shouldUsePrimaryButtonCornerRadius: Bool = false) {
```
wooj-stripe: Seems like we could just have a funciton like:
```
    func resolveCornerRadius(appearance: PaymentSheet.Appearance, shouldUsePrimaryButtonCornerRadius: Bool = false) -> CGFloat? {
        if shouldUsePrimaryButtonCornerRadius, let primaryButtonCornerRadius = appearance.primaryButton.cornerRadius {
            return primaryButtonCornerRadius
        }
        return nil
    }
```

Then call `applyCornerRadius(cornerRadius: CGFloat, iOS26DefaultCornerStyle: ...)` (which I proposed in another comment).

wdyt?

Example/PaymentSheet Example/PaymentSheet Example/AppearancePlaygroundView.swift
gbirch-stripe: Awesome, this was driving me nuts. Thanks for fixing

StripeUICore/StripeUICore/Source/Helpers/LiquidGlassHelpers.swift
gbirch-stripe, line 105: Making sure I understand this right. Before, Liquid Glass being enabled meant you always went with our LG corners, and the corner radius setting only applied to non-LG. But now, if you set the cornerRadius is uniformly applies across LG and non-LG devices?

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/ViewModels/LinkInlineSignupViewModel.swift
```diff
@@ -323,7 +323,7 @@ final class LinkInlineSignupViewModel {
     var containerCornerRadius: CGFloat {
         switch mode {
         case .checkbox:
-            return LiquidGlassDetector.isEnabled ? LinkUI.cornerRadius : configuration.appearance.cornerRadius
+            return configuration.appearance.cornerRadius ?? LinkUI.cornerRadius
```
gbirch-stripe, line 326: Maybe I'm misreading, but I think the behavior changed here? Before we were using the Link corner radius for the LG case and now we're not?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/WalletHeaderView.swift
```diff
@@ -198,7 +199,12 @@ extension PaymentSheetViewController {
             // The corner configuration API that powers ios26_applyCapsuleCornerConfiguration doesn't work on PKPaymentButton
             // Instead, we set the cornerRadius directly
             // TODO(gbirch): align Apple Pay button liquid glass styling with other elements
-            button.cornerRadius = LiquidGlassDetector.isEnabled ? 34 : appearance.cornerRadius
+            if appearance.cornerRadius == nil, LiquidGlassDetector.isEnabledInMerchantApp {
+                button.cornerRadius = 34
+            } else {
+                button.cornerRadius = appearance.cornerRadius ?? 6
```
gbirch-stripe: +1 to @wooj-stripe's comment about leaving `?? 6` everywhere. If we ever want to change this value it's going to be difficult, and also doing it this way obscures the intended behavior of the code (is this view supposed to be using the default corner radius, or is it specifically supposed to be 6 for another reason?). The constant idea sounds good to me

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetAppearance+Helpers.swift
```diff
@@ -0,0 +1,44 @@
...
+
+internal extension PaymentSheet.Appearance {
+    var topFormInsets: NSDirectionalEdgeInsets {
+        return .insets(top: formInsets.top, leading: formInsets.leading, trailing: formInsets.trailing)
+    }
+}
+
+extension PaymentSheet.Appearance.NavigationBarStyle {
+    var isGlass: Bool {
+        #if !os(visionOS)
```
gbirch-stripe, line 20: 🤮. Nice find figuring this behavior out!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/RemoveButton.swift
gbirch-stripe: `strokeWidth = appearance.selectedBorderWidth ?? appearance.borderWidth * 1.5` was the existing behavior before I updated anything for LG, so this is a change in behavior from what it has been since December, when it was intentionally changed to this in response to quality review (see https://github.com/stripe/stripe-ios/pull/4390, https://docs.google.com/document/d/1cZmqoIs5XsfPAKJKsoHmMeQDKAm8oz7lDJzTu-K8MS8/edit?tab=t.0#bookmark=id.8f9q5q4r1ss4).

Tests/ReferenceImages_64/StripePaymentSheetTests.PaymentMethodTypeCollectionViewCellSnapshotTests/test_withPromoBadge_customAppearance_iOS26@3x.png
gbirch-stripe, line 1: Looks like the liquid glass corners got removed here. Is this a snapshot test bug, an intentional change, or an issue?

Tests/ReferenceImages_64/StripePaymentSheetTests.SavedPaymentOptionsViewControllerSnapshotTests/test_all_saved_pms_and_apple_pay_and_link_custom_appearance_iOS26@3x.png
gbirch-stripe, line 1: Same here and a bunch below — LG corners removed

StripeUICore/StripeUICore/Source/Helpers/LiquidGlassHelpers.swift
yuki-stripe: Yes, corner radius is never ignored now. A value of `nil` applies various corner configurations on iOS 26+.

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/ViewModels/LinkInlineSignupViewModel.swift
yuki-stripe: Yes the behavior changed. I think what Link wants is to set its corner radiuses to 26 when `appearance.cornerRadius` is nil on iOS 26+ i.e. when the merchant has configured the appearance to use Liquid Glass corners on iOS 26.

cc @tillh-stripe

Tests/ReferenceImages_64/StripePaymentSheetTests.PaymentMethodTypeCollectionViewCellSnapshotTests/test_withPromoBadge_customAppearance_iOS26@3x.png
yuki-stripe: The corner radius in this test is set to 2 - the promo badge now respects it.

Tests/ReferenceImages_64/StripePaymentSheetTests.SavedPaymentOptionsViewControllerSnapshotTests/test_all_saved_pms_and_apple_pay_and_link_custom_appearance_iOS26@3x.png
yuki-stripe: Generally all of these 'custom appearance' snapshot tests have an explicitly set corner radius that we now respect

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetAppearance+Helpers.swift
yuki-stripe: I think this method is actually significantly worse compared to just inlining the code! It replaces super simple logic with a method call that you have to look up (hurts readability) for ~no benefit b/c the code is so simple and straightforward.
```
let cornerRadius = resolveCornerRadius(appearance: appearance)
vs.
let cornerRadius = appearance.cornerRadius
```
or
```
let cornerRadius = resolveCornerRadius(appearance: appearance, shouldUsePrimaryButtonCornerRadius: true)
vs.
let cornerRadius = appearance.primaryButton.cornerRadius ?? appearance.cornerRadius
```

StripeUICore/StripeUICore/Source/Helpers/LiquidGlassHelpers.swift
yuki-stripe: I think I can see how it's unclear. What about calling this something like `applyCornerRadiusOrConfiguration(for appearance ElementsAppearance, ios26DefaultCornerStyle: CornerStyle = .uniform)`

I don't want to decompose this method if that's what you're suggesting b/c then all callers would have to do like
```
let cornerRadius: CGFloat? = resolveCornerRadius(...)
applyCornerRadius(cornerRadius)
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetAppearance+Helpers.swift
wooj-stripe: chatted offline, ignore this comment.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/WalletHeaderView.swift
```diff
@@ -63,11 +63,12 @@ extension PaymentSheetViewController {
...
             let button = PayWithLinkButton()
-            if LiquidGlassDetector.isEnabled {
+            if appearance.cornerRadius == nil, LiquidGlassDetector.isEnabledInMerchantApp {
                 button.ios26_applyCapsuleCornerConfiguration()
             } else {
-                button.cornerRadius = appearance.cornerRadius
+                button.cornerRadius = appearance.cornerRadius ?? PaymentSheet.Appearance.defaultCornerRadius
             }
+            button.applyCornerRadiusOrConfiguration(for: appearance, ios26DefaultCornerStyle: .capsule)
             button.accessibilityIdentifier = "pay_with_link_button"
```
wooj-stripe: Seems like we can get rid of this if/else, and just call the line you have on line 71:
```
button.applyCornerRadiusOrConfiguration(for: appearance, ios26DefaultCornerStyle: .capsule)
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetAppearance.swift
```diff
@@ -310,12 +315,17 @@ public extension PaymentSheet {
...
     @available(iOS 26.0, *)
     @available(visionOS, unavailable)
     @_spi(STP) mutating func applyLiquidGlass() {
-        assert(LiquidGlassDetector.isEnabledInMerchantApp, "Requirements for this function are using at least Xcode26 and not be opted out using UIDesignRequiresCompatibility.")
+#if compiler(<6.2)
+        assertionFailure("applyLiquidGlass() requires Xcode 26.")
+#endif
+        if let optOutFlag = Bundle.main.infoDictionary?["UIDesignRequiresCompatibility"] as? Bool, optOutFlag {
+            assertionFailure("applyLiquidGlass() requires UIDesignRequiresCompatibility = NO")
+        }
```
wooj-stripe: Nice idea! Giving users specific messages about exactly what is wrong -- Maybe we update `LiquidGlassDetector` so that all the logic is in one area:
```
    @_spi(STP) public static var isEnabledInMerchantApp: Bool {
        return meetsCompilerRequirements && !hasOptedOut
    }
    @_spi(STP) public static var meetsCompilerRequirements: Bool {
        #if compiler(>=6.2)
        return true
        #else
        return false
        #endif
    }
    @_spi(STP) public static var hasOptedOut: Bool {
        if let optOutFlag = Bundle.main.infoDictionary?["UIDesignRequiresCompatibility"] as? Bool {
            return optOutFlag
        }
        return false
    }
```

Then in PaymentSheetAppearance, we:
```
        guard LiquidGlassDetector.meetsCompilerRequirements else {
            assertionFailure("applyLiquidGlass() requires Xcode 26.")
            return
        }
        guard !LiquidGlassDetector.hasOptedOut else {
            assertionFailure("applyLiquidGlass() requires UIDesignRequiresCompatibility = NO")
            return
        }
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/RemoveButton.swift
yuki-stripe: 👍 the feedback was "Remove button stroke should be 2px – appears 1px" -- not sure where the decision was to make it use `selectedBorderWidth`. cc @joyceqin-stripe - any ideas?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/WalletHeaderView.swift
yuki-stripe: Sadly PayWithLinkButton has a special `cornerRadius` property whereas the helper (and ~all other users) set `layer.cornerRadius`.

But good catch, I don't think we should be setting line 71 at all.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetAppearance.swift
yuki-stripe: Nice, good idea

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/New Payment Method Screen/WalletHeaderView.swift
wooj-stripe: fwiw -- it seems to work by calling `applyCornerRadiusOrConfiguration` (for the link button) but i'm fine with this since we were always setting `button.cornerRadius`. (Note: i agree that this logic is needed for the apple pay button).

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetAppearance.swift
```diff
@@ -310,12 +315,19 @@ public extension PaymentSheet {
...
     @_spi(STP) mutating func applyLiquidGlass() {
-        assert(LiquidGlassDetector.isEnabledInMerchantApp, "Requirements for this function are using at least Xcode26 and not be opted out using UIDesignRequiresCompatibility.")
+        guard LiquidGlassDetector.meetsCompilerRequirements else {
+             assertionFailure("applyLiquidGlass() requires Xcode 26.")
+             return
+         }
+         guard !LiquidGlassDetector.hasOptedOut else {
+             assertionFailure("applyLiquidGlass() requires UIDesignRequiresCompatibility = NO")
+             return
+         }
```
wooj-stripe, line 330: Spacing is off here, but i'm not gonna block on this.

wooj-stripe (approved):

PR #5550, opened 2025-09-16
Add ConfirmationTokens public API

porter-stripe:
## Summary
- Adds the public API definition for ConfirmationTokens as defined in the API review
- No implementation, just the stubs under spi protection.

## Motivation
- https://docs.google.com/document/d/1zL6KkOaOM2vVqJ3PBM_o2qHSNej1B7Dhx09Fu0y1jCs/edit?tab=t.0

## Testing
- CI 

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetIntentConfiguration.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetIntentConfiguration.swift
@@ -7,6 +7,7 @@
 
 import Foundation
 @_spi(STP) import StripeCore
+@_spi(ConfirmationTokensPublicPreview) import StripePayments
 
 public extension PaymentSheet {
     /// Contains information needed to render PaymentSheet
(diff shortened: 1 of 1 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetIntentConfiguration.swift
```diff
@@ -35,6 +36,19 @@ public extension PaymentSheet {
             _ intentCreationCallback: @escaping ((Result<String, Error>) -> Void)
         ) -> Void
 
+        /// Called when the customer confirms payment using confirmation tokens.
```
gbirch-stripe, line 39: nit: Style guide recommends against using Latin abbreviation including `e.g.` and `etc.` in documentation due to frequent misuse and lack of understandability to non-native english speakers. Seems like we're not very adherent in our API documentation, so maybe this isn't a concern

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetIntentConfiguration.swift
```diff
@@ -111,6 +125,36 @@ public extension PaymentSheet {
             validate()
         }
 
+        /// Creates a `PaymentSheet.IntentConfiguration` with a confirmation token handler
+        /// - Parameters:
+        ///   - mode: The mode of this intent, either payment or setup
+        ///   - paymentMethodTypes: The payment method types for the intent
+        ///   - onBehalfOf: The account (if any) for which the funds of the intent are intended
```
gbirch-stripe, line 132: Maybe there's something better we can use than `intended`, which IMO makes it sound uncertain? Could go either way on this

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetIntentConfiguration.swift
porter-stripe: Interesting callout, yeah we do this quite a lot but maybe shouldn't

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetIntentConfiguration.swift
porter-stripe: Going to leave as-is to keep consistent w/ the API review and with the other confirmHandler doc comment!

PR #5523, opened 2025-09-11
Crypto Onramp SDK: Begins Separating Steps in Example App Flow 

mliberatore:
## Summary
This work starts to split out pieces of `AuthenticatedView` into separate steps, only shown when necessary. I’ve added a "flow coordinator" to help manage the state and verification fetching, leaving less responsibility in the root view of the flow. This work begins by separating out:
- KYC info collection (note that this itself can certainly be further separated into discrete steps)
- Identity verification

Further work will continue separating steps into their own views to streamline the process from authentication to checkout, removing any guesswork from what's needed to continue the process.

To make this easier, I’ve bumped up the deployment target to iOS 16 to use `NavigationStack`. Further work may include removing steps from the back-stack (i.e. registration) when we know that step can’t be re-completed.

> [!NOTE]
> There's currently a known issue where new accounts are seeing rejected KYC statuses on our demo backend. It's being investigated separately from this work. https://lickability.slack.com/archives/C094P3BRBL1/p1757602663522789

## Motivation
👻 

## Testing

Ran through multiple flows with various accounts.

### Existing user with kyc + identity verification completed. [Root > Authenticated View]

https://github.com/user-attachments/assets/79f8b2ca-addb-43ad-8a62-777ff562687e

### Existing user with only identity verification completed. [Root > Identity Verification > Authenticated View]

https://github.com/user-attachments/assets/297f2c0c-6068-4793-8538-52f38dd314f6

### New user. [Root > Registration View > KYC > Identity Verification > Authenticated View]

https://github.com/user-attachments/assets/90d75bfc-973c-42a1-80af-829f8f0296ed

## Changelog
N/A

PR #5555, opened 2025-09-16
Crypto Onramp SDK: Moves Wallet Selection to Own Screen in Example App

mliberatore:
## Summary
In an ongoing effort started at #5523, this moves another component out of `AuthenticatedView` into its own post-auth step: Wallet Selection / Attachment.

| Empty State | Wallets Available |
| --- | --- |
|  |  |

## Motivation
To make the example app flow more similar to desired client experience.

## Testing

Made sure I could run through the flow, and not continue until I provided a wallet. Then, made sure I could complete checkout with the selected wallet in the next step.

https://github.com/user-attachments/assets/2b71939f-f3b3-4df0-860b-b40924a02b05

## Changelog
N/A


Example/CryptoOnramp Example/CryptoOnramp Example/CryptoOnrampExampleView.swift
```diff
@@ -16,8 +16,9 @@ import StripePaymentSheet
 
 /// The main content view of the example CryptoOnramp app.
 struct CryptoOnrampExampleView: View {
+    @StateObject private var flowCoordinator = CryptoOnrampFlowCoordinator()
```
mliberatore, line 19: This should’ve always been a `@StateObject` to work properly with `ObservableObject` / `@Published`. It wasn’t until adding the wallet selection view that we were getting lucky that other state was changing, allowing the `path` updates to propagate.

Example/CryptoOnramp Example/CryptoOnramp Example/CryptoOnrampFlowCoordinator.swift
```diff
@@ -17,39 +17,33 @@ final class CryptoOnrampFlowCoordinator: ObservableObject {
 
     /// Represents the possible steps in the flow.
     enum Route: Hashable {
-        case registration
+        case registration(email: String, oAuthScopes: [OAuthScopes])
         case kycInfo
         case identity
-        case authenticated
+        case wallets(customerId: String)
+        case authenticated(customerId: String, wallet: CustomerWalletsResponse.Wallet)
```
mliberatore, line 24: Rather than having the clients of this class deal with optional state, I‘ve migrated to using non-optional associated values.

Example/CryptoOnramp Example/CryptoOnramp Example/CryptoOnrampFlowCoordinator.swift
```diff
@@ -17,39 +17,33 @@ final class CryptoOnrampFlowCoordinator: ObservableObject {
...
     }
 
-    /// The coordinator responsible for interacting with the necessary APIs, from authentication to checkout.
-    let onrampCoordinator: CryptoOnrampCoordinator
-
     /// Indicates whether the global loading interface should be shown.
-    let isLoading: Binding<Bool>
+    var isLoading: Binding<Bool>?
```
mliberatore, line 28: Necessary only because we can’t use an optional instance of this class w/ `@StateObject`, and need the loading state from the environment in the example app view, which we only have after initialization.

Example/CryptoOnramp Example/CryptoOnramp Example/IdentityVerificationView.swift
```diff
@@ -31,7 +31,7 @@ struct IdentityVerificationView: View {
                     .font(.largeTitle)
                     .padding()
                     .background {
-                        Color.secondary.opacity(0.2)
+                        Color(.systemGroupedBackground)
```
mliberatore, line 34: This is to match the color I’m using for a similar view in `WalletSelectionView` (the background of the SF Symbol near the top).

mats-stripe (approved):

PR #5558, opened 2025-09-16
iOS 26 Add liquid glass corners to card scanner

gbirch-stripe:
## Summary
Add liquid glass corners to card scanner

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-4042
https://docs.google.com/spreadsheets/d/1cxo0iF_cOynuSN2bdncKT38tojCKnP1swBgZsjIjOJY/edit?gid=0#gid=0

## Testing
Snapshot test

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/CardScanningView.swift
```diff
@@ -203,13 +203,23 @@ class CardScanningView: UIView, STPCardScannerDelegate {
...
 
+        let closeButtonInset: CGFloat
+        // If Liquid Glass is enabled, we use rounder corners to match the appearance of the text fields and other elements
+        // The close button is pushed a bit further away from the edge to compensate
+        if LiquidGlassDetector.isEnabled {
+            ios26_applyDefaultCornerConfiguration()
+            closeButtonInset = 12
+        } else {
+            layer.cornerRadius = Self.cornerRadius
+            cameraView.layer.cornerRadius = Self.cornerRadius
```
wooj-stripe, line 221: This is interesting -- the original code doesn't respect the `appearance.cornerRadius` and uses a hardcoded value of 4.

In any case, we're removing `LiquidGlassDetector.isEnable` very soon, and will replacing with an approach proposed [here](https://github.com/stripe/stripe-ios/pull/5549). Assuming we pass the ElementsTheme through to this class's init, I suspect we'll end up with something like:
```
        var closeButtonInset: CGFloat = 12
        applyCornerRadius(appearance: theme) //"magical" function which determines capsule or not.
        if !didSetCornerConfiguration {
            layer.cornerRadius = Self.cornerRadius
            cameraView.layer.cornerRadius = Self.cornerRadius
            closeButtonInset = 8
        }
```

Code generally looks fine for now, but will we'll want to coordinate with @yuki-stripe .

gbirch-stripe: Got it. @yuki-stripe let me know how you want to order / integrate our changes

wooj-stripe (approved):

PR #5573, opened 2025-09-17
Remove mandate data from ConfirmationToken

porter-stripe:
## Summary
- Removes mandate data from the STPConfirmationToken based on the Slack thread below.

## Motivation
- https://stripe.slack.com/archives/C04RQ1RN10F/p1758133547925909

## Testing
- CI

## Changelog
N/A

### DIFF
--- a/StripePayments/StripePayments/Source/API Bindings/Models/ConfirmationTokens/STPConfirmationToken.swift
+++ b/StripePayments/StripePayments/Source/API Bindings/Models/ConfirmationTokens/STPConfirmationToken.swift
@@ -32,9 +32,6 @@ import Foundation
     /// `true` if the object exists in live mode or the value `false` if the object exists in test mode.
     private(set) public var liveMode = false
 
-    /// Data used for generating a Mandate.
-    private(set) var mandateData: STPMandateDataParams?
-
     /// ID of the PaymentIntent this token was used to confirm.
     private(set) public var paymentIntentId: String?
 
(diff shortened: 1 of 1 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetIntentConfiguration.swift
```diff
@@ -165,7 +165,7 @@ public extension PaymentSheet {
 
         /// Called when the customer confirms payment.
         /// See the documentation for `ConfirmHandler` for more details.
-        public var confirmHandler: ConfirmHandler
+        public var confirmHandler: ConfirmHandler?
```
joyceqin-stripe, line 168: Just verifying, this is being made optional because they could have either a ConfirmHandler or a ConfirmationTokenConfirmHandler, correct? Will there be assertions/safeguards to make sure that at least one of them is non-nil?

porter-stripe: Yep that's correct we will have an assert internally that will ensure only one of them is set. We have one similar assert but we'll add another one. Here is the existing one:
```swift
        @_spi(ConfirmationTokensPublicPreview) public init(mode: Mode,
                                                           paymentMethodTypes: [String]? = nil,
                                                           onBehalfOf: String? = nil,
                                                           paymentMethodConfigurationId: String? = nil,
                                                           confirmationTokenConfirmHandler: @escaping ConfirmationTokenConfirmHandler,
                                                           requireCVCRecollection: Bool = false) {
            // ...
            self.confirmHandler = { _, _, callback in
                // fail immediately, this should never be called
                stpAssertionFailure("Confirmation token configuration should use confirmationTokenConfirmHandler, not confirmHandler")
                let error = PaymentSheetError.intentConfigurationValidationFailed(message: "Internal Confirmation Token error. Please file an issue at https://github.com/stripe/stripe-ios.")
                callback(.failure(error))
            }

        }
```

PR #5578, opened 2025-09-18
Change to STPConfirmPaymentMethodOptions on CTs

porter-stripe:
## Summary
- Rather than using STPPaymentMethodOptions, the existing type of STPConfirmPaymentMethodOptions better matches the create confirmation tokens endpoint params.

## Motivation
- https://docs.stripe.com/api/confirmation_tokens/test_create

## Testing
- New/updated tests

## Changelog
N/A

### DIFF
--- a/StripeCryptoOnramp/StripeCryptoOnramp.xcodeproj/project.pbxproj
+++ b/StripeCryptoOnramp/StripeCryptoOnramp.xcodeproj/project.pbxproj
@@ -53,11 +53,11 @@
 		49A11F432E5A295B002ECED5 /* AuthorizeResult.swift in Sources */ = {isa = PBXBuildFile; fileRef = 49A11F422E5A295B002ECED5 /* AuthorizeResult.swift */; };
 		49B34BBB2E5D0ACC0067E44F /* Image.swift in Sources */ = {isa = PBXBuildFile; fileRef = 49B34BBA2E5D0ACC0067E44F /* Image.swift */; };
 		49B7F6882E57988D0053C83B /* OnrampSessionResponse.swift in Sources */ = {isa = PBXBuildFile; fileRef = 49B7F6872E57988D0053C83B /* OnrampSessionResponse.swift */; };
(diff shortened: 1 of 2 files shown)


StripePayments/StripePayments/Source/API Bindings/Models/ConfirmationTokens/STPConfirmationTokenParams.swift
```diff
@@ -20,7 +20,7 @@ import Foundation
     @objc public var paymentMethodData: STPPaymentMethodParams?
 
     /// Payment-method-specific configuration for this ConfirmationToken.
-    @objc public var paymentMethodOptions: STPPaymentMethodOptions?
+    @objc public var paymentMethodOptions: STPConfirmPaymentMethodOptions?
```
joyceqin-stripe, line 23: What makes STPConfirmPaymentMethodOptions better than STPPaymentMethodOptions? Looking at the Stripe API doc for creating a test token, it looks like the main payment method option is card.installments, which neither of them have. The two of them seem somewhat similar— just wondering what makes one better

porter-stripe: Neither class supports installments (the SDK currently doesn't support this, so it's intentional). But STPConfirmPaymentMethodOptions is better since it's modeled after the PaymentMethodOptions used in the intent confirm endpoint, which is what ConfirmationTokens pass to under-the-hood at confirm time. Additionally, STPConfirmPaymentMethodOptions has a better API for our internal usage - all properties are vars making them easy to work with to set params, unlike STPPaymentMethodOptions which only has an initializer requiring allResponseFields.

STPPaymentMethodOptions is better for decoding PaymentMethodOptions from the server, while STPConfirmPaymentMethodOptions is better for serializing params to the server.

joyceqin-stripe (approved):

PR #5580, opened 2025-09-18
[iOS 26] Rework readonly component background color

yuki-stripe:
## Summary
Reworked the readonly component to not depend on LiquidGlass.isEnabled and instead depend on the brightness of the background color. 

This also fixes an existing bug where the color is brightened/darkened according to the device dark mode setting rather than the brightness of the background - incorrect if e.g. the user has set a black background for light and dark mode.

## Motivation
iOS 26

## Testing
Existing snapshot tests

https://github.com/user-attachments/assets/c52dab35-630a-41dd-8314-b9d2c315edc3

https://github.com/user-attachments/assets/0a4d0a81-7404-49c7-9020-c9d28e2b4c95


StripeUICore/StripeUICore/Source/Categories/UIColor+StripeUICore.swift
```diff
@@ -146,19 +146,6 @@ import UIKit
         return self.withAlphaComponent(alpha * 0.4)
     }
 
-    /// Returns a translucent mask based on the brightness of the color
-    var translucentMaskColor: UIColor {
```
yuki-stripe, line 150: Moved out of UIColor+, it's specific to Appearance

StripeUICore/StripeUICore/Source/Elements/Section/SectionContainerView.swift
```diff
@@ -105,6 +105,9 @@ class SectionContainerView: UIView {
         } else {
             backgroundColor = .tertiarySystemGroupedBackground
         }
+        // Draw border - need to do it here so that the cgcolor gets updated on `traitCollectionDidChange`
+        layer.borderColor = theme.colors.border.cgColor
+        layer.borderWidth = theme.borderWidth
```
yuki-stripe, line 110: Fixes a subtle existing issue with the `translucentBackground` where it would overlap with the border

StripeUICore/StripeUICore/Source/Helpers/LiquidGlassHelpers.swift
```diff
@@ -18,6 +18,9 @@ import UIKit
     }()
 
     @_spi(STP) public static var isEnabled: Bool {
+        guard #available(iOS 26.0, *) else {
+            return false
+        }
```
yuki-stripe, line 23: Whoops - broke this an earlier PR!

Tests/ReferenceImages_64/StripePaymentSheetTests.UpdatePaymentMethodViewControllerSnapshotTests/test_UpdatePaymentMethodViewControllerCanUpdateAppearance_iOS26@3x.png
wooj-stripe, line 1: The CVC looks way more dark previously -- maybe we can play with the formula to darken a bit more?

yuki-stripe: I'm not sure how to make this change, I'd have to think about it more. The brightness of the background here is 1.0 so we'd need to use some other sort of input. Open to ideas.

StripeUICore/StripeUICore/Source/Elements/ElementsUI.swift
```diff
@@ -128,6 +128,25 @@ import UIKit
...
+                if resolvedColor.isBright {
+                    // The brighter the background color, the less we need to darken the color in order to ensure it looks disabled _and_ has enough contrast.
+                    // There's undoubtedly a better formula, but I just want:
+                    // 0.04 when brightness is 1
+                    // 0.09 when brightness is 0.96
+                    // and some linear interpolation between:
+                    let darkenFactor = -1.25 * backgroundColor.brightness + 1.29
+                    return resolvedColor.darken(by: darkenFactor)
+                } else {
+                    return resolvedColor.lighten(by: 0.01)
```
wooj-stripe, line 145: I would have expected there to be a lighten factor as well?

yuki-stripe: Open to ideas! I played around a bit but the current state doesn't look good IMO to begin with. I think we'd have to sit down with design and come up with a bunch of examples of background color + component color + readonly component color states to understand what we want to do.

wooj-stripe: Without really understanding the formula here, it does seem to look better at -1.15 * brightness + 1.29

Tests/ReferenceImages_64/StripePaymentSheetTests.UpdatePaymentMethodViewControllerSnapshotTests/test_UpdatePaymentMethodViewControllerCanUpdateAppearance_iOS26@3x.png
wooj-stripe: Without really understanding the formula here, it does seem to look better at `-1.15 * brightness + 1.29`

PR #5582, opened 2025-09-18
Refactor STPClientAttributionMetadata

joyceqin-stripe:


StripeApplePay/StripeApplePay/Source/PaymentsCore/API/Models/ClientAttributionMetadata.swift
```diff
@@ -25,22 +26,25 @@ import Foundation
...
     /// The intent creation flow for the merchant integration. Can be `standard` or `deferred`
-    @objc public var paymentIntentCreationFlow: String?
+    @objc public let paymentIntentCreationFlow: String?
     /// The payment method selection for the merchant integration. Can be `automatic` or `merchant_specified`
-    @objc public var paymentMethodSelectionFlow: String?
+    @objc public let paymentMethodSelectionFlow: String?
 
-    public init(elementsSessionConfigId: String? = nil,
+    public init(clientSessionId: String? = AnalyticsHelper.shared.sessionID,
+                elementsSessionConfigId: String? = nil,
```
wooj-stripe: i'm guessing you already tried, but is it possible to make these non-nilable? e.g.:
```
public init(clientSessionId: String,
                  elementsSessionConfigId: String
...
```

This isn't (currently) possible since `AnalyticsHelper.sessionId` is optional, but seems like we can make this non-optional by removing `clearSessionID()` and just calling `generateSessionID()` instead.

StripeApplePay/StripeApplePay/Source/PaymentsCore/API/Models/ClientAttributionMetadata.swift
```diff
@@ -25,22 +26,25 @@ import Foundation
...
                 paymentIntentCreationFlow: IntentCreationFlow? = nil,
                 paymentMethodSelectionFlow: PaymentMethodSelectionFlow? = nil) {
-        self.clientSessionId = AnalyticsHelper.shared.sessionID
+        assert(!(clientSessionId?.isEmpty ?? true), "Client session id must not be empty.")
+        assert(!(elementsSessionConfigId?.isEmpty ?? true), "Elements session config id must not be empty.")
```
wooj-stripe: These should probably be stpAssert since we want to assert on them, but merchants effectively have no way to fix this if they encounter an assert.

Also, have we considered firing an analytic/unexpected error if this happens. Doesn't have to be in this pr, but seems like we should.

StripeApplePay/StripeApplePay/Source/PaymentsCore/API/Models/ClientAttributionMetadata.swift
joyceqin-stripe: `clearSessionID()` is only ever called in tests. Session IDs are only generated in PaymentSheet, FlowController, Embedded, and CustomerSheet when they are created. While we can't guarantee that AnalyticsHelper.shared.sessionID will always be non-nil, we know that when we try to create client attribution metadata, it should be non-nil because we only want to send client attribution metadata from those surfaces. So we don't want to send client attribution metadata from things like the LinkController, which doesn't get called from PaymentSheet, FlowController, Embedded, or CustomerSheet.

StripeCore/StripeCore/Source/Analytics/STPAnalyticEvent.swift
```diff
@@ -313,4 +313,7 @@ import Foundation
     case passiveCaptchaSuccess = "elements.captcha.passive.success"
     case passiveCaptchaError = "elements.captcha.passive.error"
     case passiveCaptchaAttach = "elements.captcha.passive.attach"
+
+    // MARK: - ClientAttributionMetadata
+    case clientAttributionMetadataInitFailed = "clientAttributionMetadataCreationFailed.init.failed"
```
wooj-stripe: The value of this seems off:  maybe it should be:
`clientAttributionMetadata.init.failed` ?

StripeApplePay/StripeApplePay/Source/PaymentsCore/API/Models/ClientAttributionMetadata.swift
wooj-stripe: If clearSessionID is only called from tests, then it seems possible to get rid of clearSessionID() so that we can guarantee that it will always be non-nil.  I think that would make the code a lot easier to reason about.  Is it really hard to change this?

Stripe/StripeiOSTests/STPConfirmationTokenParamsTest.swift
```diff
@@ -234,7 +233,7 @@ class STPConfirmationTokenParamsTest: XCTestCase {
 
         XCTAssertNil(params.clientAttributionMetadata)
 
-        let clientMetadata = STPClientAttributionMetadata()
+        let clientMetadata = STPFormEncodableClientAttributionMetadata(clientSessionId: "client_session_id", elementsSessionConfigId: "elements_session_config_id")
```
wooj-stripe: Seems like we should use a string somewhat representable of clientSessionId as a value rather than the decoding key "client_session_id".

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/STPAPIClient+Link.swift
```diff
@@ -370,7 +370,7 @@ extension STPAPIClient {
         cvc: String?,
         expectedPaymentMethodType: String?,
         billingPhoneNumber: String?,
-        clientAttributionMetadata: STPClientAttributionMetadata,
+        clientAttributionMetadata: STPClientAttributionMetadata?,
```
wooj-stripe, line 373: nit: this is no longer considered a "refactor" as we are fixing an issue here.

It's a bit confusing when we should/shouldn't pass `clientAttributionMetadata` -- we should probably comment this at the call site of sharePaymentDetails comment out why we are passing 'nil' vs not nil.

StripePayments/StripePayments/Source/API Bindings/Models/Shared/STPFormEncodableClientAttributionMetadata.swift
```diff
@@ -1,15 +1,14 @@
...
-//  Created by Joyce Qin on 8/4/25.
+//  Created by Joyce Qin on 10/02/25.
 //
 
 import Foundation
 @_spi(STP) import StripeCore
 
-// MARK: - STPFormEncodable
-extension STPClientAttributionMetadata: STPFormEncodable {
+@_spi(STP) public class STPFormEncodableClientAttributionMetadata: STPClientAttributionMetadata, STPFormEncodable {
```
wooj-stripe, line 11: I thought that we wanted to avoid inheritance across modules that need to obj-c compatibility?

StripeApplePay/StripeApplePay/Source/PaymentsCore/API/Models/ClientAttributionMetadata.swift
joyceqin-stripe: Well the point is that the session ID is nil by default, and it's only non-nil when generated in PaymentSheet, FlowController, Embedded, and CustomerSheet. We can't guarantee it's non-nil

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/STPAPIClient+Link.swift
joyceqin-stripe: Yeah, we only want to send clientAttributionMetadata from PaymentSheet, FlowController, Embedded, and CustomerSheet. `sharePaymentDetails` is a Link call, and it can be called from `LinkController`, which doesn't generate a session ID and therefore shouldn't send clientAttributionMetadata. I can try adding comments

StripePayments/StripePayments/Source/API Bindings/Models/Shared/STPFormEncodableClientAttributionMetadata.swift
joyceqin-stripe: There is a chance that the conformance could be stripped, but this feels like a very niche case, and the STPFormEncodable methods were moved into the class definition instead of as an extension. I've added an alert for if it ever happens again, and I guess if that happened maybe we would do the approach where we have two copies of the ClientAttributionMetadata class. If you feel strongly about it, we can just do that approach from the get-go. It's safe, but any additional client attribution metadata fields would have to be added in two places instead of one.

PR #5588, opened 2025-09-18
Use background color defined in appearance API for embedded playground

wooj-stripe:
## Summary
Use background color defined in appearance API for embedded playground #5588

## Motivation
Remove dependence on LiquidGlassDetector.isEnabled


wooj-stripe:
Going to roll this into a larger pr.

PR #5591, opened 2025-09-19
Update `This is your default` view in Link UI

tillh-stripe:
## Summary

This pull request replaces the `This is your default` label in the Link UI with the more pleasant hint view.


Tests/ReferenceImages_64/StripePaymentSheetTests.LinkHintMessageViewSnapshotTests/testOutlinedStyle_iOS26@3x.png
tillh-stripe, line 1: I'm unsure why this doesn't render the border correctly. Here's what it looks like on device.

image

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Extensions/UIColor+Link.swift
```diff
@@ -51,6 +51,8 @@ extension UIColor {
     static let linkContentOnPrimaryButton: UIColor = UIColor(red: 0, green: 0.12, blue: 0.06, alpha: 1.0)
     /// Separator color on primary button
     static let linkSeparatorOnPrimaryButton: UIColor = brand400
+    /// Border color around the Link hint message view
+    static let linkHintMessageBorder: UIColor = .dynamic(light: neutral300, dark: neutral700)
```
mats-stripe: Shoud we just use `linkBorderDefault` above?

I will say that the outline in dark mode seems too bright to me. Maybe it should be the same color as the icon? WDYT?

jkelle-stripe: I agree with this, it's a bit too bright. We do want the user to know this is the default payment method, but in light mode, it has a good balance with the form fields above where you know those are the priority, but it seems over prioritized visually.

tillh-stripe: My mistake — my code change in the last commit accidentally switched to a brighter color for dark mode. I'll fix that!

tillh-stripe: Fixed in 7f2442d74db2d6069854585f2849f3814fb51b24, including a change to better handle appearance changes.

jkelle-stripe: Looks great!!
image

jkelle-stripe (approved):
LGTM!

PR #5593, opened 2025-09-19
Respect `allowedCountries` in bank tab

tillh-stripe:
## Summary

This pull request ensures that the billing details collection configuration's `allowedCountries` are respected in the bank tab.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
@@ -213,7 +213,7 @@ class PaymentSheetFormFactory {
     func make() -> PaymentMethodElement {
         switch paymentMethod {
         case .instantDebits, .linkCardBrand:
-            return makeInstantDebits(countries: configuration.billingDetailsCollectionConfiguration.allowedCountriesArray)
+            return makeInstantDebits()
         case .external(let externalPaymentOption):
             return makeExternalPaymentMethodForm(subtitle: externalPaymentOption.displaySubtext,
                                                  disableBillingDetailCollection: externalPaymentOption.disableBillingDetailCollection)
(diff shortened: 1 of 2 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
```diff
@@ -213,7 +213,7 @@ class PaymentSheetFormFactory {
     func make() -> PaymentMethodElement {
         switch paymentMethod {
         case .instantDebits, .linkCardBrand:
-            return makeInstantDebits()
+            return makeInstantDebits(countries: configuration.billingDetailsCollectionConfiguration.allowedCountriesArray)
```
wooj-stripe: Is this correct? if it is empty -- should we default to the previous behavior, or... should we not even show instantDebits/linkCardBand ?

tillh-stripe: Same behavior as for cards: If `allowedCountries` is empty, then `allowedCountriesArray` is nil and we don't filter. As soon as the merchant passes in any value to `allowedCountries`, we only accept those.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
```diff
@@ -890,7 +890,7 @@ extension PaymentSheetFormFactory {
         return SubtitleElement(view: label, isHorizontalMode: configuration.isHorizontalMode)
     }
 
-    func makeInstantDebits(countries: [String]? = nil) -> PaymentMethodElement {
+    func makeInstantDebits(countries: [String]?) -> PaymentMethodElement {
```
wooj-stripe: technically we can make this:

`func makeInstantDebits() -> PaymentMethodElement {`

Then just reference `configuration.billingDetailsCollectionConfiguration.allowedCountriesArray` in `makeBillingAddressSection` on line 904.

This would make the code consistent with usbankaccount, blik, and upi.

PR #5603, opened 2025-09-19
Move ElementsAppearance to its own file and remove default theme from initializers

yuki-stripe:
## Summary
It's not safe to have view initializers like 
```
class SomeViewThatRespectsAppearance: UIView {
  init(appearance: Appearance = .default) { .. }
}
```
because that would allow you to drop the user provided appearance settings. Instead, users should be required to pass in an appearance, and explicitly make it the default if necessary.

This also fixes LinkInlineSignupView more info logo not using appearance.

## Motivation
Clean up after iOS 26 work

## Testing
Existing tests

## Changelog
Not user facing

### DIFF
--- a/StripeUICore/StripeUICore/Source/Controls/OneTimeCodeTextField.swift
+++ b/StripeUICore/StripeUICore/Source/Controls/OneTimeCodeTextField.swift
@@ -147,7 +147,7 @@ import UIKit
     // MARK: -
     public init(
         configuration: Configuration = Configuration(),
-        theme: ElementsAppearance,
+        theme: ElementsAppearance
     ) {
         self.configuration = configuration
         self.textStorage = TextStorage(capacity: configuration.numberOfDigits)
--- a/StripeUICore/StripeUICore/Source/Elements/ElementsUI.swift
+++ b/StripeUICore/StripeUICore/Source/Elements/ElementsUI.swift
@@ -7,7 +7,6 @@
 //
 
 import Foundation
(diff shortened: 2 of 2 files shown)


StripeUICore/StripeUICore.xcodeproj/project.pbxproj
```diff
@@ -91,6 +91,7 @@
 		B1B177526E04EE65A9D1C64A /* DynamicHeightContainerView.swift in Sources */ = {isa = PBXBuildFile; fileRef = B9E7CFED5747279A5976D7BA /* DynamicHeightContainerView.swift */; };
 		B6107E5E2D6E116417D22DD9 /* StripeUICoreBundleLocator.swift in Sources */ = {isa = PBXBuildFile; fileRef = E68363D1C6BB6468AC1DDA4C /* StripeUICoreBundleLocator.swift */; };
 		B65F5B032E60C5EB001425CC /* LiquidGlassHelpers.swift in Sources */ = {isa = PBXBuildFile; fileRef = B65F5B022E60C5EB001425CC /* LiquidGlassHelpers.swift */; };
+		B6F905562E7E0BC800942BA5 /* ElementsAppearance.swift in Sources */ = {isa = PBXBuildFile; fileRef = B6F905552E7E0BC500942BA5 /* ElementsAppearance.swift */; };
```
davidme-stripe, line 94: You can remove these changes, the xcodeproj will now pull in source files automatically

PR #5606, opened 2025-09-22
Add confirmation tokens to the playground

porter-stripe:
## Summary
- Adds confirmation token APIs to the playground to enable easy testing

## Motivation
- Confirmation Tokens

## Testing
- Manual

## Changelog
N/A

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
@@ -14,7 +14,7 @@ import Contacts
 import PassKit
 @_spi(STP) import StripeCore
 @_spi(STP) import StripePayments
-@_spi(CustomerSessionBetaAccess) @_spi(STP) @_spi(PaymentSheetSkipConfirmation) @_spi(ExperimentalAllowsRemovalOfLastSavedPaymentMethodAPI) @_spi(CustomPaymentMethodsBeta) @_spi(PaymentMethodOptionsSetupFutureUsagePreview) import StripePaymentSheet
+@_spi(CustomerSessionBetaAccess) @_spi(STP) @_spi(PaymentSheetSkipConfirmation) @_spi(ExperimentalAllowsRemovalOfLastSavedPaymentMethodAPI) @_spi(CustomPaymentMethodsBeta) @_spi(PaymentMethodOptionsSetupFutureUsagePreview) @_spi(ConfirmationTokensPublicPreview) import StripePaymentSheet
 import SwiftUI
 import UIKit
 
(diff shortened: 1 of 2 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+DeferredAPI.swift
```diff
@@ -10,7 +10,59 @@ import Foundation
...
+        allowsSetAsDefaultPM: Bool = false,
+        elementsSession: STPElementsSession?,
+        mandateData: STPMandateDataParams? = nil,
+        radarOptions: STPRadarOptions? = nil,
+        completion: @escaping (PaymentSheetResult, STPAnalyticsClient.DeferredIntentConfirmationType?) -> Void
+    ) {
+        // Assert that not both confirm handlers are populated
+        let hasConfirmHandler = intentConfig.confirmHandler != nil
+        let hasConfirmationTokenHandler = intentConfig.confirmationTokenConfirmHandler != nil
+        stpAssert(!(hasConfirmHandler && hasConfirmationTokenHandler), "Only one confirm handler should be set at a time")
```
joyceqin-stripe: Shouldn't we also assert that not both confirm handlers are nil? So assert that at least one of them is set

Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
```diff
@@ -995,20 +1030,67 @@ extension PlaygroundController {
...
+            if settings.integrationType == .deferred_csc_ct {
+                DispatchQueue.global(qos: .userInteractive).asyncAfter(deadline: .now() + 1) {
+                    intentCreationCallback(.success(self.clientSecret!))
+                }
+            }
+            return
+        case .deferred_ssc_ct:
+            break
+        default:
+            assertionFailure()
```
joyceqin-stripe: Should we add a message in this assertion failure similar to the one in confirmHandler?

Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
```diff
@@ -995,20 +1030,67 @@ extension PlaygroundController {
...
+        }
+
+        confirmHandlerInternal(
+            paymentMethodId: nil,
+            shouldSavePaymentMethod: false,
```
joyceqin-stripe: Why is it false for CTs?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+DeferredAPI.swift
porter-stripe: Yeah we can add one! I think it's pretty unlikely though: 

```swift
  var intentConfig = PaymentSheet.IntentConfiguration(
      mode: .payment(amount: 1099, currency: "USD"),
      confirmHandler: { _, _, callback in
          callback(.success("pi_test_client_secret"))
      }
  )

  // Setting handler to nil
  intentConfig.confirmHandler = nil
  // Pass to PaymentSheet
```

Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
porter-stripe: That's a little confusing, I've changed it to be nil. CTs don't need this flag.

PR #5613, opened 2025-09-22
Confirm Apple Pay with ConfirmationTokens

porter-stripe:
## Summary
- Adds support for confirming Apple Pay intents with ConfirmationTokens

## Motivation
- Confirmation tokens

## Testing
- Manual
- New unit tests
- E2E tests

## Changelog
N/A

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
+++ b/Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
@@ -1218,6 +1218,36 @@ class PaymentSheetDeferredServerSideUITests: PaymentSheetUITestCase {
         payWithApplePay()
     }
 
+    func testDeferredPaymentIntent_ApplePay_ConfirmationToken_ClientSideConfirmation() {
+        var settings = PaymentSheetTestPlaygroundSettings.defaultValues()
+        settings.layout = .horizontal
+        settings.integrationType = .deferred_csc_ct
+        settings.apmsEnabled = .off
+        loadPlayground(app, settings)
+
+        app.buttons["Present PaymentSheet"].tap()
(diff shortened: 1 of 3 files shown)


joyceqin-stripe (commented):
Sorry, I don't think I'm understanding ConfirmationTokens properly— could you help explain?
In the eng brief, it says the goal is to replace the deferred flow to use ConfirmationTokens rather than a payment method. When inspecting the Stripe requests, we still create a payment method and (for csc) we call confirm with an attached payment method. How do ConfirmationTokens fit into this flow? I can see that we create a confirmation token, and we use the confirmationTokenConfirmHandler to get the client secret, but do we not attach it anywhere? What does the ConfirmationToken do?

porter-stripe:
@joyceqin-stripe CTs will become our recommended integration path (replacing the current deferred flow in documentation, though the code remains for now). We may deprecate the existing deferred flow eventually.

On Apple Pay's unique handling:

Apple Pay always creates a payment method, even when using CTs. While this deviates from the CT pattern, I believe keeping it this way is the right choice because:
- We never save these payment methods anyway
- It keeps our implementation significantly simpler
- It's our existing, proven approach

We could refactor STPApplePayContext to conditionally create either a payment method (legacy flow) or a confirmation token (CT flow), but this added complexity isn't worth it.

https://github.com/stripe/stripe-ios/blob/master/StripeApplePay/StripeApplePay/Source/ApplePayContext/STPApplePayContext.swift#L580

joyceqin-stripe (approved):
We'll have to note for the data science team that client attribution metadata will be sent twice on ApplePay for the CT flow (once on payment method creation, once on CT creation), but that should be fine

porter-stripe:
@joyceqin-stripe if needed we can remove the CT creation metadata, just lmk

PR #5618, opened 2025-09-23
Fixes Link Wallet CVC Layout Issue

mliberatore:
## Summary
There was previously an issue that the keyboard was hiding the CVC view, which was fixed in #5546. The issue was re-reported, but with screenshots showing what appeared to be a collapsed CVC field. I was able to reproduce this by force-showing the software keyboard in the simulator, and entering 3 digits into the cvc field. This triggers an unexpected loss of focus of the cvc field, and additional layout issues that cause that same field to then expand and contract. When the keyboard is up after this, the field is almost completely collapsed:

| Keyboard is dismissed and field expands | When re-focused, field collapses almost completely | Field stays collapsed |
| --- | --- | --- |
|  |  |  |

These issues (automatic dismissal of the keyboard, expanding and collapsing of the cvc field) appear to be side effects of #5530. I’ve reverted the code changes, and implemented a new fix to maintain all dividers, and update their visibility according to the visibility of surrounding views.

## Motivation
To fix the newly caught layout issues with the CVC field.

## Testing
Ensured that the field remained focused even after entering 3 digits, and then once dismissing the keyboard, the field maintained its proper size:

| Keyboard, No Divider | No Keyboard, No Divider |
| --- | --- |
|  |  |

Forced the expiration date field to show and tested to ensure the divider was present:

| Keyboard, Divider | No Keyboard, Divider |
| --- | --- |
|  |  |

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController-WalletViewController.swift
```diff
@@ -120,23 +120,10 @@ extension PayWithLinkViewController {
...
-            var elements: [Element] = []
-            if viewModel.shouldRecollectCardExpiryDate {
-                elements.append(expiryDateElement)
-            }
-            if viewModel.shouldRecollectCardCVC {
-                elements.append(cvcElement)
-            }
-            return elements.isEmpty ? nil : elements
-        }
+        private lazy var cardDetailsRecollectionRow = SectionElement.MultiElementRow([expiryDateElement, cvcElement], theme: theme)
```
mliberatore, line 123: Stored in a separate property from `cardDetailsRecollectionSection` so it can be referenced directly without casting below to call `updateDividerVisibility()`.

StripeUICore/StripeUICore/Source/Elements/Section/SectionContainerView.swift
```diff
@@ -201,32 +201,66 @@ extension SectionContainerView: EventHandler {
 
 extension SectionContainerView {
     class MultiElementRowView: UIView {
+        private class DividerView: UIView {
```
mliberatore, line 204: I created a private `DividerView` type to more easily locate and reference within the arranged subviews when hiding/showing below.

StripeUICore/StripeUICore/Source/Elements/Section/SectionContainerView.swift
```diff
@@ -201,32 +201,66 @@ extension SectionContainerView: EventHandler {
...
+                widthConstraint.isActive = true
+                backgroundColor = color
+            }
+
+            required init?(coder: NSCoder) {
+                fatalError("init(coder:) has not been implemented")
+            }
+        }
+
+        private let stackView: UIStackView = {
```
mliberatore, line 217: Similarly, the `UIStackView` has been pulled out into a property to be directly referenced in `updateDividerVisibility()`.

mats-stripe (approved):
Changes LGTM. Manually tested it pretty thoroughly

PR #5620, opened 2025-09-23
Allow setting destination currency in crypto onramp example app

mats-stripe:
## Summary

A small change that allows users customize the destination currency in the crypto onramp example app

## Motivation

https://stripe.slack.com/archives/C08PAPXTZK6/p1758658765635829?thread_ts=1758594926.679789&cid=C08PAPXTZK6

## Testing

Manually tested, video in the slack thread above!

## Changelog

N/a

### DIFF
--- a/Example/CryptoOnramp Example/CryptoOnramp Example/PaymentView.swift
+++ b/Example/CryptoOnramp Example/CryptoOnramp Example/PaymentView.swift
@@ -332,6 +332,7 @@ struct PaymentView: View {
             actions: {
                 TextField("Currency (e.g. usdc)", text: $editingCurrencyText)
                     .textInputAutocapitalization(.never)
+                    .autocorrectionDisabled()
 
                 Button("Save") {
                     let trimmed = editingCurrencyText.trimmingCharacters(in: .whitespacesAndNewlines)


Example/CryptoOnramp Example/CryptoOnramp Example/PaymentView.swift
```diff
@@ -293,6 +328,29 @@ struct PaymentView: View {
                 Text(alert.message)
             }
         )
+        .alert(
+            "Edit Destination Currency",
+            isPresented: isPresentingEditCurrencyAlert,
+            actions: {
+                TextField("Currency (e.g. usdc)", text: $editingCurrencyText)
+                    .textInputAutocapitalization(.never)
```
mliberatore, line 334: I think we may want to turn off autocorrect as well. I was seeing `usdc` convert to `used` and `btc` convert to `etc` or `btw` in the simulator.

mats-stripe: Good catch!

mats-stripe: Done in https://github.com/stripe/stripe-ios/pull/5620/commits/25f16b7b17c3eabbdb20c4a90308f7fd4c36c3ac

PR #5624, opened 2025-09-24
Extend webview's scrollview through safe area

wooj-stripe:
## Summary
Extends the popup to the bottom of the screen

## Motivation
There seems to be a gap in the safe area if we don't do this.

## Testing
Manual

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Internal/ECE/ECEViewController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Internal/ECE/ECEViewController.swift
@@ -436,10 +436,10 @@ extension ECEViewController: WKUIDelegate {
 
         // Use the provided configuration directly (this fixes the configuration error)
         // Create popup with full screen bounds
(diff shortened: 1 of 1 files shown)


StripePaymentSheet/StripePaymentSheet/Source/Internal/ECE/ECEViewController.swift
```diff
@@ -450,10 +450,18 @@ extension ECEViewController: WKUIDelegate {
...
 
         // Add the popup webview as a fullscreen overlay
-        if let popupWebView = popupWebView {
+        if let popupWebView {
+            popupWebView.translatesAutoresizingMaskIntoConstraints = false
+            popupWebView.scrollView.contentInsetAdjustmentBehavior = .never
             view.addSubview(popupWebView)
-        }
 
+            NSLayoutConstraint.activate([
```
gbirch-stripe, line 458: nit: use `addAndPinSubview()` for cleanliness?

wooj-stripe:
Decided not to go down this route.

PR #5625, opened 2025-09-24
Prevent hcaptcha crash

joyceqin-stripe:
## Summary

Refactored PassiveCaptchaChallenge to use error throwing pattern instead of Result and refactored the functions. Instead of `start`, we have `fetchToken` that fetches the token or throws, and instead of the original `fetchToken`, we have `fetchTokenWithTimeout`, which awaits `fetchToken` with a timeout.
Added a lock with the nillableContinuation in `fetchToken` to prevent any possible Swift Task continuation misuse crash if it somehow tried to resume continuation from another thread.
## Motivation

assertionFailure: reported crashes due to a race condition that led to an assertionFailure due to validationTask not being initialized by the time await validationTask?.value was called. The refactor should fix this.
Swift Task continuation misuse crash: reported crashes due to attempted continuation resume.
https://jira.corp.stripe.com/browse/MOBILESDK-4057
## Testing

Manual, CI
## Changelog

N/A


StripePayments/StripePayments/Source/Captcha/HCaptchaWebViewManager.swift
```diff
@@ -78,28 +78,34 @@ internal class HCaptchaWebViewManager: NSObject {
...
-        return webview
-    }()
+         let webview = {
+             let debug = Log.minLevel == .debug
+             let webview = WKWebView(
+                 frame: CGRect(origin: CGPoint.zero, size: webViewInitSize),
+                 configuration: self.buildConfiguration()
+             )
+             webview.accessibilityIdentifier = "webview"
+             webview.accessibilityTraits = UIAccessibilityTraits.link
```
davidme-stripe: Why these traits? I don't think users can interact with this for passive captchas.

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
```diff
@@ -56,97 +55,109 @@ typealias CaptchaResult = Result<String, Error>
...
                                             host: "stripecdn.com")
                 STPAnalyticsClient.sharedClient.logPassiveCaptchaExecute(siteKey: siteKey)
-                let startTime = Date()
-                let result = await withCheckedContinuation { (continuation: CheckedContinuation<CaptchaResult, Never>) in
-                    // Prevent Swift Task continuation misuse - hcaptcha.validate can be called unexpectedly multiple times
-                    var nillableContinuation: CheckedContinuation<CaptchaResult, Never>? = continuation
-                    hcaptcha.didFinishLoading {
-                        hcaptcha.validate { result in
+                let result = try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<String, Error>) in
+                    // Prevent Swift Task continuation misuse with atomic flag and safer handling - hcaptcha.validate can be called unexpectedly multiple times
```
davidme-stripe: I usually wouldn't use an NSLock here: It's not immediately clear what `hcaptcha?.validate()` does, so we run the risk of deadlocks if (for example) something in `validate()` triggered another `didFinishLoading`.

Is there more info on why the continuation was being called multiple times? Is hCaptcha calling the `didFinishLoading` block multiple times?

StripePayments/StripePayments/Source/Captcha/HCaptchaWebViewManager.swift
```diff
@@ -137,6 +143,55 @@ internal class HCaptchaWebViewManager: NSObject {
...
+        if Thread.isMainThread {
+            Self.cleanup(webView: webViewToClean, observer: observerToRemove)
+        } else {
+            DispatchQueue.main.sync {
+                Self.cleanup(webView: webViewToClean, observer: observerToRemove)
+            }
+        }
+
+        // Clean up all closure references
+        completion = nil
```
davidme-stripe: I don't think setting any of these to `nil` in deinit is necessary, unless there's some side effect to these being set to nil.

StripePayments/StripePayments/Source/Captcha/HCaptchaWebViewManager.swift
```diff
@@ -137,6 +143,55 @@ internal class HCaptchaWebViewManager: NSObject {
...
+
+        // Capture references for cleanup
+        let webViewToClean = _webView
+        let observerToRemove = observer
+
+        // Clean up on the main thread
+        if Thread.isMainThread {
+            Self.cleanup(webView: webViewToClean, observer: observerToRemove)
+        } else {
+            DispatchQueue.main.sync {
```
davidme-stripe: It generally isn't a good idea to use DispatchQueue.main.sync: In this case, it could lead to a deadlock if this background thread is blocked or otherwise unable to dispatch to the main thread.

StripePayments/StripePayments/Source/Captcha/HCaptchaWebViewManager.swift
```diff
@@ -137,6 +143,55 @@ internal class HCaptchaWebViewManager: NSObject {
...
+        resultHandled = true
+
+        // Clear decoder reference
+        decoder = nil
+    }
+
+    private static func cleanup(webView: WKWebView?, observer: NSObjectProtocol?) {
+        // Remove observer
+        if let observer {
+            NotificationCenter.default.removeObserver(observer)
```
davidme-stripe: I think this (is the most important part of the deinit), though I'm not sure I understand what the observer is for. It's just for when a new UIWindow opens?

StripePayments/StripePayments/Source/Captcha/HCaptchaWebViewManager.swift
```diff
@@ -137,6 +143,55 @@ internal class HCaptchaWebViewManager: NSObject {
...
+        }
+
+        if let webView {
+            // Remove script message handlers to break retain cycles
+            webView.configuration.userContentController.removeScriptMessageHandler(forName: "hcaptcha")
```
davidme-stripe: Is `deinit` even called if there's a retain cycle that includes this class? Or is the retain cycle elsewhere?

StripePayments/StripePayments/Source/Captcha/HCaptchaWebViewManager.swift
joyceqin-stripe: This code is just copied from the lazy var webView— this was just to store a reference in _webView so we have a reference to clean up that won't trigger webView initialization if it wasn't already initialized

StripePayments/StripePayments/Source/Captcha/HCaptchaWebViewManager.swift
joyceqin-stripe: Yeah, it's for when a new window opens so that it can set up the webview

StripePayments/StripePayments/Source/Captcha/HCaptchaWebViewManager.swift
joyceqin-stripe: This is to remove the reference to the decoder, which is captured by the webview's usercontentcontroller, and the decoder is what decodes the messages that lead to callbacks getting called so we want to remove it

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
joyceqin-stripe: Do you want to remove the lock? I'm worried that the completion may be called from a different thread based on [when George had his crash](https://stripe.slack.com/archives/C02HQBW38JX/p1758567964569479), but we're also not sure that the branch George was on had the nillableContinuation safeguard.

~I believe that the didFinishLoading block was called again because both times we got the crash, it showed that the continuation tried to resume with a webview termination error, so I thought it had to do with webview load. One possibility is that the hcaptcha token expired, and the hcaptcha was reset, reloaded, and revalidated, and during the process, the webview terminated, leading to the continuation re-resume with the termination error. Or perhaps the webview terminated and HCaptcha tries to reload, but I haven't found code to support that. However, I have not been able to verify that this is what's actually happening, as I have not been able to reproduce it.~

EDIT: I think I found the issue. When the webview terminates, it calls completion with the error. I didn't realize this, but the completion block for validation gets set as the completion handler for everything, and that triggered the Swift Task continuation misuse issue. If we move the lock into the validate completion block, is that safe?

StripePayments/StripePayments/Source/Captcha/HCaptchaWebViewManager.swift
joyceqin-stripe: Would DispatchQueue.main.async work instead? Or should I just add a Task with @MainActor?

StripePayments/StripePayments/Source/Captcha/HCaptchaWebViewManager.swift
joyceqin-stripe: The idea is that if the user presented the sheet and then dismissed it, the callbacks wouldn't be triggered by the JavaScript, but I suppose removing the ScriptMessageHandler would prevent that anyways.

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
```diff
@@ -58,95 +57,106 @@ typealias CaptchaResult = Result<String, Error>
...
         self.passiveCaptcha = passiveCaptcha
-        Task { await start() } // Intentionally not blocking loading/initialization!
+        Task { try await fetchToken() } // Intentionally not blocking loading/initialization!
     }
 
-    private func start() {
-        guard validationTask == nil else { return }
+    deinit {
+        // Cancel any pending validation task
+        validationTask?.cancel()
```
yuki-stripe: This is perfectly fine but is it doing anything?

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
```diff
@@ -58,95 +57,106 @@ typealias CaptchaResult = Result<String, Error>
...
+            return try await withThrowingTaskGroup(of: String.self) { group in
+                let isReady = isValidationComplete
+                // Add hcaptcha task
+                group.addTask { [weak self] in
+                    guard let self else { throw PassiveCaptchaError.unexpected }
+                    do {
+                        return try await fetchToken()
+                    } catch {
+                        throw error
+                    }
```
yuki-stripe: I this should just be `return try await fetchToken()` right

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
```diff
@@ -58,95 +57,106 @@ typealias CaptchaResult = Result<String, Error>
...
-                // Only log error if due to timeout. Any other errors have already been logged in start()
-                if case Error.timeout = error {
-                    STPAnalyticsClient.sharedClient.logPassiveCaptchaError(error: error, siteKey: siteKey, duration: Date().timeIntervalSince(startTime))
+        let siteKey = passiveCaptcha.siteKey
+        do {
+            return try await withThrowingTaskGroup(of: String.self) { group in
+                let isReady = isValidationComplete
+                // Add hcaptcha task
+                group.addTask { [weak self] in
+                    guard let self else { throw PassiveCaptchaError.unexpected }
```
yuki-stripe: What's the idea behind throwing this particular error here?

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
```diff
@@ -58,95 +57,106 @@ typealias CaptchaResult = Result<String, Error>
...
+                }
+                // Add timeout task
+                group.addTask {
+                    try? await Task.sleep(nanoseconds: timeoutNs)
+                    throw PassiveCaptchaError.timeout
+                }
+                defer {
+                    group.cancelAll()
+                    validationTask?.cancel()
                 }
```
yuki-stripe, line 142: What's this for?

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
joyceqin-stripe: I can't just return or return nil, so I needed to throw something. Don't know if there's a different error I should be throwing

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
joyceqin-stripe: I've noticed that if I don't have that, the timeout doesn't work and it keeps blocking confirmation.

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
```diff
@@ -58,95 +57,95 @@ typealias CaptchaResult = Result<String, Error>
...
-                            } catch {
-                                nillableContinuation?.resume(returning: .failure(error))
-                                nillableContinuation = nil
+                let result = try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<String, Error>) in
+                    // Prevent Swift Task continuation misuse - the validate completion block can get called from multiple places
+                    var nillableContinuation: CheckedContinuation<String, Error>? = continuation
+
+                    // Weak reference to break retain cycles
+                    hcaptcha.didFinishLoading { [weak hcaptcha] in
+                        hcaptcha?.validate { result in
```
yuki-stripe: I looked up `didFinishLoading` and it says this - maybe this explains why we're seeing `validate` called multiple times?

```
     The closure may be called multiple times since the resources may also be loaded multiple times
     in case of error or reset. This may also be immediately called if the resources have already
     finished loading when you set the closure.
```

The HCaptcha code is *quite* confusing, so now I want to double check - how do we know this is the right way to use it? Do they have docs that say to do these two lines of code?

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
yuki-stripe: I see. I think I'd just capture self strongly here, because it's simpler and there's no retain cycle (self does not own the things that own this closure). Self will only be retained until the method returns.

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
```diff
@@ -58,95 +57,95 @@ typealias CaptchaResult = Result<String, Error>
...
     public init(passiveCaptcha: PassiveCaptcha) {
         self.passiveCaptcha = passiveCaptcha
-        Task { await start() } // Intentionally not blocking loading/initialization!
+        Task { try await fetchToken() } // Intentionally not blocking loading/initialization!
     }
 
-    private func start() {
-        guard validationTask == nil else { return }
+    private func fetchToken() async throws -> String {
+        if let validationTask {
```
yuki-stripe: Something that's making this code hard to read - why is a method named `fetchToken` returning the result of a 'validation task'?  Maybe this should be `fetchValidationToken` or `validate`?  

And while I'm asking questions - what is a validation token, anyways?

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
yuki-stripe: I'm finding this name confusing. "Passive Captcha" is the name of the process/action of passively checking that you're a human. This seems more like "data we need to initialize/start Passive Captcha", is that right? If so, can we name it to match? Or have I misunderstood something?

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
yuki-stripe: Woah. Can we look at that more? I don't see how this would fix an issue like that or how that bug could happen (you're syaing this method never returns?)

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
joyceqin-stripe: I think maybe the `didFinishLoading` closure isn't necessary anymore. Back in the early days when I was trying to get hcaptcha to work, I tried creating the hcaptcha and calling `validate` on it right away, but it never returned a token because the hcaptcha hadn't finished loading, which is why I added the `didFinishLoading` thing. But I previously didn't have it wrapped in an `await withCheckedContinuation`, so I think it didn't wait for the token to get populated. They have [this example](https://github.com/hCaptcha/HCaptcha-ios-sdk/blob/master/Example/Passive-Example/ContentView.swift#L42) in their SDK and it's wrapped in the await

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
joyceqin-stripe: It's just fetching the hcaptcha token— there's no special validation token. Calling `validate` fetches a token wrapped in an `HCaptchaResult`, and that's the token we want to return. I could rename the task to like `tokenTask` or something

joyceqin-stripe:
>Woah. Can we look at that more? I don't see how this would fix an issue like that or how that bug could happen (you're syaing this method never returns?)

I'm starting to think this timeout doesn't actually work.
It will return, but only after the task has completed, which defeats the purpose. From my understanding, the taskgroup doesn't complete until its children complete, even if we already know what we want to return, and since one of the child tasks awaits the validationTask's value, it still doesn't complete until validationTask has completed. The reason it looked like canceling the task did something is because I was using Task.sleep, which automatically checks for cancellation.

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
```diff
@@ -31,127 +31,145 @@ import Foundation
...
+
                         hcaptcha.validate { result in
-                            do {
-                                let token = try result.dematerialize()
-                                nillableContinuation?.resume(returning: .success(token))
-                                nillableContinuation = nil
-                            } catch {
-                                nillableContinuation?.resume(returning: .failure(error))
-                                nillableContinuation = nil
+                            Task { @MainActor in
```
yuki-stripe: Let's add a comment explaining why we put this in a MainActor Task.

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
```diff
@@ -31,127 +31,145 @@ import Foundation
...
-            case .none:
-                assertionFailure("No captcha result found!")
-                return nil
+                // Add timeout task
+                group.addTask {
+                    try await Task.sleep(nanoseconds: timeoutNs)
+                    throw PassiveCaptchaError.timeout
+                }
+                defer {
+                    tokenTask?.cancel()
```
yuki-stripe: Let's explain this too - why we need to cancel. Also can we just do this directly after grabbing the first result, that's a bit simpler than `defer`

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
```diff
@@ -31,127 +31,145 @@ import Foundation
...
+                let hcaptcha = try hcaptchaFactory.create(siteKey: siteKey, rqdata: rqdata)
                 STPAnalyticsClient.sharedClient.logPassiveCaptchaExecute(siteKey: siteKey)
-                let startTime = Date()
-                let result = await withCheckedContinuation { (continuation: CheckedContinuation<CaptchaResult, Never>) in
-                    // Prevent Swift Task continuation misuse - hcaptcha.validate can be called unexpectedly multiple times
-                    var nillableContinuation: CheckedContinuation<CaptchaResult, Never>? = continuation
-                    hcaptcha.didFinishLoading {
+                let result = try await withTaskCancellationHandler {
+                    try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<String, Error>) in
+                        // Prevent Swift Task continuation misuse - the validate completion block can get called from multiple places
```
yuki-stripe: WDYT about clarifying this comment to say that the validate block can be called multiple *times*? That's the real issue we're trying to solve with the nillable continuation.

StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
joyceqin-stripe: I kept `defer` because if `try await group.next()` throws (like in the case of a timeout), then `tokenTask?.cancel()` wouldn't be called. I could call `tokenTask?.cancel()`  in both blocks, but at that point it seems like it's fine to just keep it in `defer`, wdyt?

yuki-stripe: Ah makes sense, thanks.

PR #5626, opened 2025-09-24
Allow overriding country code in test mode

tillh-stripe:
## Summary

This pull request updates the usage of `userOverrideCountry` in the playground. Instead of only using it locally, we now send it to `elements/sessions`, which allows us consider the override country even in fields that are being computed in the backend.

See [here](https://livegrep.corp.stripe.com/view/stripe-internal/pay-server/lib/elements/command/retrieve_session_info.rb#L1306) for the implementation of `country_override` on the backend.

## Motivation

Easier testing for non-US folks.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/STPAPIClient+PaymentSheet.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/STPAPIClient+PaymentSheet.swift
@@ -13,16 +13,22 @@ import Foundation
 extension STPAPIClient {
     typealias STPIntentCompletionBlock = ((Result<Intent, Error>) -> Void)
 
-    func makeElementsSessionsParams(mode: PaymentSheet.InitializationMode,
-                                    epmConfiguration: PaymentSheet.ExternalPaymentMethodConfiguration?,
-                                    cpmConfiguration: PaymentSheet.CustomPaymentMethodConfiguration?,
-                                    clientDefaultPaymentMethod: String?,
-                                    customerAccessProvider: PaymentSheet.CustomerAccessProvider?) -> [String: Any] {
+    func makeElementsSessionsParams(
+        mode: PaymentSheet.InitializationMode,
+        epmConfiguration: PaymentSheet.ExternalPaymentMethodConfiguration?,
+        cpmConfiguration: PaymentSheet.CustomPaymentMethodConfiguration?,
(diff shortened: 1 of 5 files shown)


StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/STPAPIClient+PaymentSheet.swift
```diff
@@ -13,16 +13,22 @@ import Foundation
...
+        userOverrideCountry: String? = nil
+    ) -> [String: Any] {
         var parameters: [String: Any] = [
             "locale": Locale.current.toLanguageTag(),
             "external_payment_methods": epmConfiguration?.externalPaymentMethods.compactMap { $0.lowercased() } ?? [],
             "custom_payment_methods": cpmConfiguration?.customPaymentMethods.compactMap { $0.id } ?? [],
         ]
+        if let userOverrideCountry {
+            parameters["country_override"] = userOverrideCountry
+        }
```
tillh-stripe, line 31: This is only considered by the backend if we're in test mode.

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Utils/LinkURLGenerator.swift
```diff
@@ -60,7 +60,7 @@ class LinkURLGenerator {
         }
 
         // We only expect regionCode to be nil in rare situations with a buggy simulator. Use a default value we can detect server-side.
-        let customerCountryCode = configuration.defaultBillingDetails.address.country ?? Locale.current.stp_regionCode ?? elementsSession.countryCode(overrideCountry: configuration.userOverrideCountry) ?? "US"
+        let customerCountryCode = configuration.defaultBillingDetails.address.country ?? Locale.current.stp_regionCode ?? elementsSession.countryCode ?? "US"
```
tillh-stripe, line 63: No need to use the local value anymore; just use what comes from the Elements session.

jkelle-stripe:
@tillh-stripe Looks good to me! Before approving, do we need to address the dead code reply from GH actions? If not, I can approve

tillh-stripe:
@jkelle-stripe The comment is outdated. The [latest dead-code check](https://github.com/stripe/stripe-ios/actions/runs/18046084083/job/51357053882?pr=5626) was successful. We should figure out how the dead-code check can delete these outdated comments.

PR #5627, opened 2025-09-24
Improve state mismatch issues in StripeAttest

mats-stripe:
## Summary

This makes a few tweaks to the `StripeAttest` client to better handle client and server state mismatches. There are two scenarios this addresses:

1. Client thinks it's not attested, but server has attestation
    - Action: Client updates its local state to match server
2. Client thinks it's attested, but server needs attestation
    - Action: Client resets and re-attests

[I'm adding a feature flag](https://git.corp.stripe.com/stripe-internal/pay-server/pull/1208789) `link_mobile_attestation_state_sync_enabled` to control this new behavior, as we don't know yet if it is a risky change.

I'm also adding new analytic events for those two scenarios,

## Motivation

More details here https://docs.google.com/document/d/1NMQoDpCOv4dvYjMmwfrsezGhc---wKy2eFkeH_DBjAQ/edit?usp=sharing

## Testing

Unit tests added

## Changelog

N/a

### DIFF
--- a/StripeCore/StripeCore/Source/Attestation/StripeAttest.swift
+++ b/StripeCore/StripeCore/Source/Attestation/StripeAttest.swift
@@ -262,6 +262,10 @@ import UIKit
         if canSyncState && !challenge.initial_attestation_required && !successfullyAttested {
             // Server has attestation but client doesn't know - update client
             successfullyAttested = true
+            let event = GenericAnalytic(event: .stateMismatchNotAttestedLocally, params: [:])
+            if let apiClient {
+                STPAnalyticsClient.sharedClient.log(analytic: event, apiClient: apiClient)
+            }
         } else if challenge.initial_attestation_required && successfullyAttested {
             // Server needs attestation but client thinks it's done - reset client and retry
             if isRetry || !canSyncState {
@@ -271,7 +275,10 @@ import UIKit
             } else {
                 // Reset and retry
                 resetKey()
-                try await self.attest()
+                let event = GenericAnalytic(event: .stateMismatchNotAttestedRemotely, params: [:])
+                if let apiClient {
+                    STPAnalyticsClient.sharedClient.log(analytic: event, apiClient: apiClient)
+                }
                 return try await _assert(canSyncState: canSyncState, isRetry: true)
             }
         }
--- a/StripeCore/StripeCore/Source/Attestation/StripeAttest.swift
+++ b/StripeCore/StripeCore/Source/Attestation/StripeAttest.swift
@@ -74,6 +74,24 @@ import UIKit
         if let apiClient {
             STPAnalyticsClient.sharedClient.log(analytic: resetKeyAnalytic, apiClient: apiClient)
         }
+
+        // Check if this is a Link assertion error, which might indicate state issues
+        if StripeAttest.isLinkAssertionError(error: error) {
+            consecutiveStateErrors += 1
+
+            // If we've hit too many consecutive errors, clear all state
+            if consecutiveStateErrors >= Self.maxConsecutiveStateErrors {
+                clearAllAttestationState()
+                let maxErrorsAnalytic = GenericAnalytic(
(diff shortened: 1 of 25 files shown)


StripeCore/StripeCore/Source/Attestation/StripeAttest.swift
```diff
@@ -258,12 +258,22 @@ import UIKit
...
+        } else if challenge.initial_attestation_required && successfullyAttested {
+            // Server needs attestation but client thinks it's done - reset client and retry
+            if isRetry || !canSyncState {
+                // We already tried once - something is wrong
+                resetKey()
+                throw AttestationError.shouldAttestButKeyIsAlreadyAttested
+            } else {
+                // Reset and retry
+                resetKey()
+                try await self.attest()
```
mats-stripe: The one thing I wasn't totally sure of is if we should re-attest here, since the client has keys that the server  doesn't have. Curious what you think @davidme-stripe @tillh-stripe

StripePayments/StripePayments/Source/API Bindings/Models/Shared/LinkSettings.swift
```diff
@@ -106,6 +109,7 @@ import Foundation
         let linkDefaultOptIn = (response["link_default_opt_in"] as? String).flatMap { LinkDefaultOptIn(rawValue: $0) }
         let linkEnableDisplayableDefaultValuesInECE = response["link_enable_displayable_default_values_in_ece"] as? Bool ?? false
         let linkShowPreferDebitCardHint = response["link_show_prefer_debit_card_hint"] as? Bool ?? false
+        let attestationStateSyncEnabled = response["link_mobile_attestation_state_sync_enabled"] as? Bool
```
mats-stripe, line 112: PR for this is here https://git.corp.stripe.com/stripe-internal/pay-server/pull/1208789

StripeCore/StripeCore/Source/Attestation/StripeAttest.swift
```diff
@@ -258,12 +258,22 @@ import UIKit
...
-        // If the backend claims that attestation is required, but we already have an attested key,
-        // something has gone wrong.
-        if challenge.initial_attestation_required {
-            // Reset the key, we'll try again next time:
-            resetKey()
-            throw AttestationError.shouldAttestButKeyIsAlreadyAttested
+        // State alignment: sync client state with server
+        if canSyncState && !challenge.initial_attestation_required && !successfullyAttested {
+            // Server has attestation but client doesn't know - update client
+            successfullyAttested = true
```
tillh-stripe, line 264: This feels like a band-aid, since `attest` would set `successfullyAttested = true`. Do you strongly feel that we need it?

(If you think so, let's also instrument it, so that we know if this ever occurs.)

StripeCore/StripeCore/Source/Attestation/StripeAttest.swift
tillh-stripe: `resetKey()` resets `successfullyAttested` to false, so we can probably skip `attest()` and just call `_assert(isRetry: true)`.

mats-stripe: Oh duh! Will remove this line

StripeCore/StripeCore/Source/Attestation/StripeAttest.swift
mats-stripe: I think this is reasonable, as it will let the client assert when the server thinks it has successfully attested. No super strong opinions on where we go with this though.

I can add logs for the two cases here!

mats-stripe: Added logs here https://github.com/stripe/stripe-ios/pull/5627/commits/ec4645e8a3f862b3bd4c46f540cd6ad9cb76cc54

tillh-stripe (approved):

PR #5638, opened 2025-09-25
[Identity] Implementing `is_stripe` and `skip_success_page` functionality

Pearapps:
## Summary

This implements the functionality of the two new fields on the verification page. This also bumps the API version to 7.

These are the two functionality changes:

1. From the verification page mode, if the model has `is_stripe` set to true - the biometric consent screen will _not_ show the stripe logo, only the brand logo.

2. On that same model, if the `skip_success_page` is set to true - the final screen (success) of the identity flow is skipped and the flow just submits.

## Testing

I tested both these scenarios by replacing the boolean checks with the opposite values that I received from the API, and both scenarios behaved as intended.

---

The is_stripe change, where only the brand logo appears:

IMG_1959


luisv-stripe (approved):
LGTM

PR #5640, opened 2025-09-25
Create confirmation token params during confirmation

porter-stripe:
## Summary
- Creates the parameters to create confirmation tokens in the deferred flow, then vends that token back to the merchant.
- The remaining of the confirmation flow is left unimplemented as a TODO for a following PR.

## Motivation
- Confirmation tokens

## Testing
- New unit tests

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+ConfirmationTokens.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+ConfirmationTokens.swift
@@ -130,7 +130,9 @@ extension PaymentSheet {
                 }
 
                 // If no mandate data, fallback to STPPaymentIntentParams auto add functionality
-                confirmationTokenParams.mandateData = STPPaymentIntentParams.mandateDataIfRequired(for: paymentMethodType)
+                if confirmationTokenParams.mandateData == nil {
+                    confirmationTokenParams.mandateData = STPPaymentIntentParams.mandateDataIfRequired(for: paymentMethodType)
+                }
             case .setup:
                 // Setup intents always require mandate data for certain payment methods
                 if STPPaymentMethodType.requiresMandateDataForSetupIntent.contains(paymentMethodType) {
@@ -138,7 +140,9 @@ extension PaymentSheet {
                 }
 
                 // If no mandate data, fallback to STPSetupIntentConfirmParams auto add functionality
-                confirmationTokenParams.mandateData = STPSetupIntentConfirmParams.mandateDataIfRequired(for: paymentMethodType)
+                if confirmationTokenParams.mandateData == nil {
+                    confirmationTokenParams.mandateData = STPSetupIntentConfirmParams.mandateDataIfRequired(for: paymentMethodType)
+                }
             }
         }
 


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+ConfirmationTokens.swift
```diff
@@ -22,6 +36,276 @@ extension PaymentSheet {
...
+        Task { @MainActor in
+            // 1. Create the confirmation token params
+            let confirmationTokenParams = createConfirmationTokenParams(confirmType: confirmType,
+                                                                        configuration: configuration,
+                                                                        intentConfig: intentConfig,
+                                                                        elementsSession: elementsSession,
+                                                                        radarOptions: radarOptions)
+
+            let ephemeralKeySecret: String? = {
+                // Only needed when using existing saved payment methods, API will error if provided for now payment methods
```
yuki-stripe:
```suggestion
                // Only needed when using existing saved payment methods, API will error if provided for new payment methods
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+ConfirmationTokens.swift
```diff
@@ -22,6 +36,276 @@ extension PaymentSheet {
...
+                guard confirmationTokenParams.paymentMethod != nil else { return nil }
+                // Link saved payment methods don't require ephemeral keys, API will error if provided
+                guard !isSavedFromLink(from: confirmType) else { return nil }
+
+                return configuration.customer?.ephemeralKeySecretBasedOn(elementsSession: elementsSession)
```
yuki-stripe:
```suggestion
                return configuration.customer?.ephemeralKeySecret(basedOn: elementsSession)
```

Ultra nit: should be like this according to swift [api design guidelines](https://www.swift.org/documentation/api-design-guidelines/)
`func ephemeralKeySecret(basedOn elementsSession: STPElementsSession) { .. }`

yuki-stripe: Also - can we stpassert if customer is nil here?  That should never happen, right

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+ConfirmationTokens.swift
```diff
@@ -22,6 +36,276 @@ extension PaymentSheet {
...
+    ///   - elementsSession: Elements session for confirmation token creation
+    ///   - mandateData: Optional mandate data (auto-generated for required payment methods)
+    ///   - radarOptions: Optional fraud detection settings
+    /// - Returns: Configured confirmation token parameters ready for Stripe's API
+    static func createConfirmationTokenParams(
+        confirmType: ConfirmPaymentMethodType,
+        configuration: PaymentElementConfiguration,
+        intentConfig: PaymentSheet.IntentConfiguration,
+        allowsSetAsDefaultPM: Bool = false,
+        elementsSession: STPElementsSession?,
```
yuki-stripe: This is unused?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+ConfirmationTokens.swift
```diff
@@ -22,6 +36,276 @@ extension PaymentSheet {
...
+
+        // 1. Initialize confirmation token with basic configuration
+        let confirmationTokenParams = STPConfirmationTokenParams()
+        confirmationTokenParams.returnURL = configuration.returnURL
+        confirmationTokenParams.shipping = configuration.shippingDetails()?.paymentIntentShippingDetailsParams
+        confirmationTokenParams.clientContext = intentConfig.createClientContext(customerId: configuration.customer?.id)
+
+        // 2. Configure payment method details based on confirm type
+        configurePaymentMethodDetails(
+            confirmationTokenParams,
```
yuki-stripe: This "static func modifies the first parameter (`confirmationTokenParams`) based on the rest" pattern is ok but makes it unclear _what_ is being configured and isn't how we do this in our codebase. Feels very LLM-y. We would do something like `confirmationTokenParams.configureX`.

IMO though I don't think these separate methods are adding enough value relative to their cost (separate methods mean I have to look in there to understand what it's doing, makes me wonder how these methods avoid overwriting each others values, all the boilerplate parameters and docstrings in each method definition, etc). I prefer inline with comments. We're not even DRY'ing anything here.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+ConfirmationTokens.swift
porter-stripe: Agree on the API name, updated this!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+ConfirmationTokens.swift
porter-stripe: Good catch yes, not sure why the dead code check didn't catch this...

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+ConfirmationTokens.swift
porter-stripe: Fair point, I'll remove these and put the code in the higher level func `createConfirmationTokenParams`

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheet+ConfirmationTokenTests.swift
```diff
@@ -0,0 +1,407 @@
...
+        let confirmType = createTestSavedConfirmType()
+
+        let params = PaymentSheet.createConfirmationTokenParams(
+            confirmType: confirmType,
+            configuration: configuration,
+            intentConfig: intentConfig,
+            elementsSession: .emptyElementsSession
+        )
+
+        XCTAssertNil(params.mandateData)
```
joyceqin-stripe, line 405: Should we XCTAssert(params.setupFutureUsage, .none) here?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+ConfirmationTokens.swift
```diff
@@ -17,11 +17,177 @@ extension PaymentSheet {
...
+            let confirmationToken = try await configuration.apiClient.createConfirmationToken(with: confirmationTokenParams,
+                                                                                              ephemeralKeySecret: ephemeralKeySecret,
+                                                                                              additionalPaymentUserAgentValues: makeDeferredPaymentUserAgentValue(intentConfiguration: intentConfig))
+
+            // 3. Vend the ConfirmationToken and fetch the client secret from the merchant
+            _ = try await fetchIntentClientSecretFromMerchant(intentConfig: intentConfig,
+                                                              confirmationToken: confirmationToken)
+
+            // TODO(porter) Finish rest of CT confirmation
+            stpAssertionFailure("Confirmation Tokens not yet implemented.")
```
joyceqin-stripe, line 56: What's the rest of CT confirmation?

joyceqin-stripe (dismissed):

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheet+ConfirmationTokenTests.swift
porter-stripe: Good idea!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+ConfirmationTokens.swift
porter-stripe: Coming in a future PR!

joyceqin-stripe (approved):

PR #5647, opened 2025-09-30
Use merchant publishable key for Link requests

tillh-stripe:
## Summary

This pull request switches to using the merchant publishable key (together with the consumer session client secret) for consumer API requests. This aligns iOS with other platforms.


jkelle-stripe (approved):
Nice that this removes more than it adds!

PR #5648, opened 2025-10-01
Fix Constraint Conflicts on Zero Width Link Navigation Bar

jkelle-stripe:
## Summary
We were getting a constraint error where _UITemporaryLayoutWidth = 0 on the `LinkSheetNavigationBar`. We were trying to logic about how the title label should be laid out even when the width was 0. The `updateTitleConstraints` method gets called again after this when  the nav bar has width > 0 so now we wait until that time to execute our constraints. This should resolve some of the issues in [this](https://jira.corp.stripe.com/browse/LINK_MOBILE-199) Jira ticket.

## Motivation
Constraint issues in the logs and in the alerts.

## Testing
Snapshot testing


jkelle-stripe:
> Nice catch @jkelle-stripe, I'm curious if these snapshot tests have the iOS 26 tag? If they don't we could add the tag, but I think the snapshots need to also be recorded on the older iOS version

All set! I added the tag and rerecorded snapshots. There wasn't a change for either 16 or 26, but that's a good sign I'm not breaking anything here.

mats-stripe:
>There wasn't a change for either 16 or 26

I don't think the iOS 26 snapshots existed before this, but if nothing changed on the 16.4 tests then we're good 👍

mats-stripe (approved):

jkelle-stripe:
@mats-stripe I think this is ready for review. The snapshots look weird to me because of the transparency, but I think I'm doing the snapshotting correctly.

Tests/ReferenceImages_64/StripePaymentSheetTests.LinkNavigationBarSnapshotTests/testTruncatingTitle_iOS26@3x.png
mats-stripe, line 1: Text seems to be overlapping with the back button 👀

jkelle-stripe:
The constraint warning is now resolved and the UI looks like it looked previously. This was resolved with the `width > 0` check guarding the `layoutIfNeeded` call (which is what triggers the temporary width constraint when called before there is a width to the view). I fought with iOS 26 glass when the bar was in `.back` style with a truncating title because the glass button's width constraint would conflict with the title label's horizontal constraints. Setting the compression resistance to low resolved this issue.

PR #5649, opened 2025-10-01
Add a few misc. CT changes

porter-stripe:
## Summary
- Send is_confirmation_token in the analytics payload
- Remove the asserts these are already covered [here](https://github.com/stripe/stripe-ios/blob/master/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetIntentConfiguration.swift#L151)
- Send flag to the playground backend to determine if we are in the confirmation token flow 

## Motivation
- CTs

## Testing
- Manual

## Changelog
N/A

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
@@ -830,6 +830,7 @@ extension PlaygroundController {
             "link_mode": settings.linkEnabledMode.rawValue,
             "use_manual_confirmation": settings.integrationType == .deferred_mc,
             "require_cvc_recollection": settings.requireCVCRecollection == .on,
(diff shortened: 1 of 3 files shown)


joyceqin-stripe (commented):
Another misc change— I didn't catch this in my previous review but thought of it later, but in PaymentSheet+ConfirmationTokens.swift, for new payment methods, there's a line `confirmationTokenParams.paymentMethodData = paymentMethodParams` and a line `confirmationTokenParams.clientAttributionMetadata = paymentMethodParams.clientAttributionMetadata`.
I don't think we need `confirmationTokenParams.clientAttributionMetadata = paymentMethodParams.clientAttributionMetadata` because clientAttributionMetadata is already included under `confirmationTokenParams.paymentMethodData`

porter-stripe:
@joyceqin-stripe yeah I thought that too originally, however this matches web's behavior. https://admin.corp.stripe.com/request-log/req_n66otHcfk773BK

PR #5651, opened 2025-10-01
Crypto Onramp SDK: Prevents Back Navigation When Necessary in Example App

mliberatore:
## Summary

Originally, I wanted to drop path items from the back stack if certain steps couldn’t be completed (e.g. advancing back after completing registration, the pre-filled email address would no longer work to register, so drop that step) but this was causing navigation animation issues that required hacks to work around.

Instead, from certain steps, I disallow back navigation, except where the previous step (e.g. selecting a crypto wallet, selecting a payment method) could be repeated successfully and used in subsequent steps. Additionally, since the logout toolbar item could be applied using a similar pattern, I cleaned up the code around this to avoid duplication of that modifier.

## Motivation
To not allow the user to repeat a successful step that we know would cause an error.

## Testing
Ran through the flow with a new account to ensure the back button (and ability to swipe back) was only present when expected. I could only get as far as payment selection, as[ I’m receiving 500s from the demo backend](https://lickability.slack.com/archives/C094P3BRBL1/p1759340960103889) when creating an onramp session.

| Registration | KYC | Identity | Wallet | Payment |
| --- | --- | --- | --- | --- |
|  |  |  |  |  |

## Changelog
N/A

### DIFF
--- a/Example/CryptoOnramp Example/CryptoOnramp Example/CryptoOnrampExampleView.swift
+++ b/Example/CryptoOnramp Example/CryptoOnramp Example/CryptoOnrampExampleView.swift
@@ -106,59 +106,61 @@ struct CryptoOnrampExampleView: View {
             .navigationBarTitleDisplayMode(.inline)
             .navigationDestination(for: CryptoOnrampFlowCoordinator.Route.self) { route in
                 if let coordinator {
-                    switch route {
-                    case let .registration(email, scopes):
-                        RegistrationView(
-                            coordinator: coordinator,
-                            email: email,
-                            selectedScopes: scopes,
-                            livemode: livemode
-                        ) { customerId in
-                            flowCoordinator.advanceAfterRegistration(customerId: customerId)
-                        }
-                    case .kycInfo:
-                        KYCInfoView(coordinator: coordinator) {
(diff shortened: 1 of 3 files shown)


Example/CryptoOnramp Example/CryptoOnramp Example/CryptoOnrampExampleView.swift
```diff
@@ -106,59 +106,61 @@ struct CryptoOnrampExampleView: View {
             .navigationBarTitleDisplayMode(.inline)
             .navigationDestination(for: CryptoOnrampFlowCoordinator.Route.self) { route in
                 if let coordinator {
-                    switch route {
```
mliberatore, line 109: I’d recommend hiding whitespace changes to see more clearly what's changed here aside from indenting inside of the newly added `ZStack`. The `ZStack`’s purpose here is simply to group the views together to apply the two modifiers once instead of on each destination view. Intentionally [not using `Group`](https://chris.eidhof.nl/post/why-i-avoid-group/) for this.

CleanShot 2025-10-01 at 15 05 56

mats-stripe (approved):
Nice!

PR #5652, opened 2025-10-01
Fix Constraint Conflicts on New Payment Method View

jkelle-stripe:
## Summary
Remove conflicting constraints in new payment method.

## Motivation
Along side #5648, this PR should resolve [this](https://jira.corp.stripe.com/browse/LINK_MOBILE-199) JIRA ticket. 

We were receiving constraint issues that looked like this:

```
(
    "<NSLayoutConstraint:0x6000024addb0 H:|-(0)-[UIView:0x100da0110]   (active, names: '|':UIStackView:0x13e7b6030 )>",
    "<NSLayoutConstraint:0x6000024aead0 'UISV-canvas-connection' UILayoutGuide:0x600003b7e3e0'UIViewLayoutMarginsGuide'.leading == _UILayoutSpacer:0x600003dc30c0'UISV-alignment-spanner'.leading   (active)>",
    "<NSLayoutConstraint:0x6000024ae8a0 'UISV-spanning-boundary' _UILayoutSpacer:0x600003dc30c0'UISV-alignment-spanner'.leading <= UIView:0x100da0110.leading   (active)>",
    "<NSLayoutConstraint:0x6000024ae5d0 'UIView-leftMargin-guide-constraint' H:|-(20)-[UILayoutGuide:0x600003b7e3e0'UIViewLayoutMarginsGuide'](LTR)   (active, names: '|':UIStackView:0x13e7b6030 )>"
)
``` 

This is a conflict between the `PayWithLinkViewController.NewPaymentViewController.addPaymentMethodVC` and the stackView containing it, whose `stackView.isLayoutMarginsRelativeArrangement` was true. When this flag is true, the stack view will give padding on the margins for the containing elements. Later on in the view, we were setting the `addPaymentMethodVC`'s leading and trailing equal to the stackView's leading and trailing, which conflicts with the margin padding we were adding with `stackView.isLayoutMarginsRelativeArrangement = true`. The two in conflict are shown here:

offending constriants

The resolution of these constraints by the system was to remove the layout margin based constraints. This PR removes these layout margin constraints explicitly, since that's the design we were rendering after constraint resolution. This PR also adds snapshot tests for this view.

## Testing
Manual testing. These two screenshots show pre/post change and the stack views have the same width.
| Before | After |
| --- | --- |
|  |  |


jkelle-stripe:
Two questions left about this PR:

1. Are we ok with the difference in height in the add payment screen? This is caused by the removal of the offending layout margins that were causing the constraint conflicts horizontally. We could add them back only vertically, if the height difference is important.

~2. The comment I left about the iOS 26 checkbox clarity.~

Tests/ReferenceImages_64/StripePaymentSheetTests.PayWithLinkViewController_NewPaymentViewControllerSnapshotTests/testIsAddingFirstPaymentMethodAndReturnError__iOS26@3x.png
jkelle-stripe, line 1: This isn't related to my change, but I can't see the checkbox on this snapshot in iOS 26, whereas I can see on non-iOS 26 (above). Should we make this checkbox darker?

mats-stripe (approved):

PR #5658, opened 2025-10-03
Retry for network errors on first call

porter-stripe:
## Summary
- Completes a TODO that we should retry on the first fetch of an intent if we get back a network error to be more resilient. 

## Motivation
- Improved error handling, more graceful

## Testing
- Manual in live mode 

## Changelog
N/A

### DIFF
--- a/StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
+++ b/StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
@@ -1525,12 +1525,11 @@ public class STPPaymentHandler: NSObject {
                         timeout: pollingBudget?.networkTimeout
                     ) { [self] paymentIntent, error in
                         guard let paymentIntent, error == nil else {
-                            // If we got an error retrieving the intent, retry if budget allows.
-                            // Note: This will only retry if we have a pollingBudget, which means it's a polling call.
-                            // We won't retry if it's the first call (no pollingBudget). Ideally we should retry
-                            // on the first call too, but this is a limitation to address in a future rewrite.
-                            if let pollingBudget, pollingBudget.canPoll {
-                                pollingBudget.pollAfter {
+                            // Retry if polling budget allows. For the first call (no polling budget), create a minimal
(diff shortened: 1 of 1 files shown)


StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
```diff
@@ -1525,12 +1525,11 @@ public class STPPaymentHandler: NSObject {
...
-                            // Note: This will only retry if we have a pollingBudget, which means it's a polling call.
-                            // We won't retry if it's the first call (no pollingBudget). Ideally we should retry
-                            // on the first call too, but this is a limitation to address in a future rewrite.
-                            if let pollingBudget, pollingBudget.canPoll {
-                                pollingBudget.pollAfter {
+                            // Retry if polling budget allows. For the first call (no polling budget), create a minimal
+                            // budget to allow one retry. This handles transient network errors.
+                            let effectivePollingBudget = pollingBudget ?? PollingBudget(startDate: Date(), duration: 1)
+                            if effectivePollingBudget.canPoll {
+                                effectivePollingBudget.pollAfter {
```
wooj-stripe, line 1532: Code looks fine, but is this what we want to do? Are we seeing a pattern of failures in this case?

porter-stripe: Mainly resolving the comment which should be a TODO: https://github.com/stripe/stripe-ios/pull/5658/files#diff-6a81f3ee91853b490eedd97eac89e41e2a6e353447556afaa2cf3bacbb025a83L1530

Mostly just defensive rather than give up after 1 try.

porter-stripe: Also got a report here: https://stripe.slack.com/archives/C02TNKMM1HT/p1759507487356389?thread_ts=1758918760.012799&cid=C02TNKMM1HT

We rolled the flag back so no more impact but this resolves the issue.

PR #5669, opened 2025-10-06
Fix Xcode 15 Link New Payment Snapshot Tests

jkelle-stripe:
## Summary
The tests on Xcode 15 [failed](https://app.bitrise.io/app/b220f4ba85d134a7/pipelines/52a67dd3-86ca-4485-886a-cdcbe0c28ea7?tests_filter_status=all) when deploying v24.24.2. This PR fixes the build error with these tests by adding a missing `// @iOS26` tag to the tests.

## Motivation
Fix failing tests.

## Testing
- [ ] Run this locally on Xcode 15 (struggling to download this since Tahoe doesn't support)

### DIFF
--- a/StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Link/PayWithLinkViewController-NewPaymentViewControllerSnapshotTests.swift
+++ b/StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Link/PayWithLinkViewController-NewPaymentViewControllerSnapshotTests.swift
@@ -10,6 +10,7 @@ import StripeCoreTestUtils
 @testable@_spi(STP) import StripePaymentSheet
 @testable@_spi(STP) import StripeUICore
 
+// @iOS26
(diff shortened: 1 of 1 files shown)


mats-stripe (dismissed):
CI passed, but it didn't run the `legacy-tests-15` workflow. I think this is the correct fix though.

Stripe/AllStripeFrameworks-iOS26.xctestplan
```diff
@@ -33,10 +33,34 @@
     },
     {
       "selectedTests": [
+        "CVCReconfirmationViewControllerSnapshotTest/testCVCRecollectionScreen()",
```
jkelle-stripe, line 36: I'm slightly worried about the updates that were made to the test plan

jkelle-stripe: The `// @iOS26` tag is the source of truth, not the xctestplan, so no need to worry about the update here.

PR #5670, opened 2025-10-06
Fix STPApplePayContext error analytics

yuki-stripe:
## Summary
Rework STPApplePayContext analytic events to look like STPPaymentHandler and PaymentSheet - send a start and a finish event w/ relevant error info.

## Motivation
https://jira.corp.stripe.com/browse/RUN_MOBILESDK-4732

## Testing
See tests

## Changelog
Not user facing.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
@@ -252,7 +252,6 @@ class PaymentSheetFormFactory {
             }
 
             guard let spec = FormSpecProvider.shared.formSpec(for: paymentMethod.identifier) else {
-                stpAssertionFailure("Failed to get form spec for \(paymentMethod.identifier)!")
                 let errorAnalytic = ErrorAnalytic(event: .unexpectedPaymentSheetFormFactoryError, error: Error.missingFormSpec, additionalNonPIIParams: ["payment_method": paymentMethod.identifier])
                 analyticsHelper?.analyticsClient.log(analytic: errorAnalytic)
                 return FormElement(elements: [], theme: theme)
--- a/StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetFormFactoryTest.swift
+++ b/StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetFormFactoryTest.swift
@@ -1809,7 +1809,7 @@ class PaymentSheetFormFactoryTest: XCTestCase {
         )
         STPAssertTestUtil.shouldSuppressNextSTPAlert = true
         _ = factory.make()
-        XCTAssertEqual(STPAssertTestUtil.lastAssertMessage, "Failed to get form spec for card_present!")
+        XCTAssertTrue(STPAssertTestUtil.lastAssertMessage.contains("missingFormSpec"))
         let errorAnalytic = analyticsClient._testLogHistory.first!
         XCTAssertEqual(errorAnalytic["event"] as? String, STPAnalyticEvent.unexpectedPaymentSheetFormFactoryError.rawValue)
         XCTAssertEqual(errorAnalytic["payment_method"] as? String, "card_present")


StripeApplePay/StripeApplePay/Source/ApplePayContext/STPApplePayContext.swift
```diff
@@ -536,121 +557,73 @@ public class STPApplePayContext: NSObject, PKPaymentAuthorizationControllerDeleg
...
-                    let errorAnalytic = ErrorAnalytic(event: .unexpectedApplePayError,
-                                                      error: InternalError.invalidState,
-                                                      additionalNonPIIParams: [
-                                                        "error_message": errorMessage
-                                                      ])
-
-                    self.analyticsClient.log(analytic: errorAnalytic, apiClient: self.apiClient)
-                    handleFinalState(.error, nil, errorAnalytic)
-                }
+            guard !self.didFinish else {
```
yuki-stripe, line 593: Replaced the old `guard self.authorizationController != nil` behavior with explicit `guard !self.didFinish`

StripeApplePay/StripeApplePay/Source/ApplePayContext/STPApplePayContext.swift
```diff
@@ -262,8 +280,8 @@ public class STPApplePayContext: NSObject, PKPaymentAuthorizationControllerDeleg
     /// ApplePayContext passes this to the /confirm endpoint for PaymentIntents if it did not collect shipping details itself.
     /// :nodoc:
     @_spi(STP) public var shippingDetails: StripeAPI.ShippingDetails?
-    private weak var delegate: _stpinternal_STPApplePayContextDelegateBase?
-    @objc var authorizationController: PKPaymentAuthorizationController?
+    weak var delegate: _stpinternal_STPApplePayContextDelegateBase?
```
wooj-stripe, line 283: seems like this can continue to be private?

StripeApplePay/StripeApplePay/Source/ApplePayContext/STPApplePayContext.swift
```diff
@@ -359,16 +381,15 @@ public class STPApplePayContext: NSObject, PKPaymentAuthorizationControllerDeleg
...
-        if let authorizationController = authorizationController {
-            objc_setAssociatedObject(
-                authorizationController,
-                UnsafeRawPointer(&kApplePayContextAssociatedObjectKey),
-                nil,
-                .OBJC_ASSOCIATION_RETAIN_NONATOMIC
-            )
-        }
-        authorizationController = nil
+        objc_setAssociatedObject(
```
wooj-stripe, line 384: Before, if `_end()` was (erroneously) called multiple times, we wouldn't call `objc_setAssociatedObject` twice, but now that `authorizationController` no longer nullable, it's possible to call `objc_setAssociatedObject`. Is there any protection we can add to avoid this?

StripeApplePay/StripeApplePay/Source/ApplePayContext/STPApplePayContext.swift
```diff
@@ -536,121 +557,73 @@ public class STPApplePayContext: NSObject, PKPaymentAuthorizationControllerDeleg
...
-                if self.didCancelOrTimeoutWhilePending {
-                    self.authorizationController?.dismiss {
-                        DispatchQueue.main.async {
-                            self.callDidCompleteDelegate(status: .error, error: self.error)
-                            self._end()
-                        }
+        // Helpers to handle annoying logic around "Do I call completion block or dismiss + call delegate?"
+        // Helper 1: Handle failure
+        func handleFailure(error: Swift.Error?) {
+            self.error = error ?? NSError.stp_genericErrorOccurredError()
```
wooj-stripe, line 563: Defaulting to an error here feels a little awkward, i looked at a couple spots of where we're calling this function and it seems like there's always a error available?

StripeApplePay/StripeApplePay/Source/ApplePayContext/STPApplePayContext.swift
yuki-stripe: I check the delegate in the test.

StripeApplePay/StripeApplePay/Source/ApplePayContext/STPApplePayContext.swift
yuki-stripe: It's safe to call `objc_setAssociatedObject` multiple times, tried it out in a test without issue.

StripeApplePay/StripeApplePay/Source/ApplePayContext/STPApplePayContext.swift
yuki-stripe: Yeah I'm not sure how to cleanly get the non-optional error in all the guards like
```
                        guard case .success(let setupIntent) = result else {
                            handleFailure(error: result.error)
                            return
                        }
```

If I use a switch statement, I'd have to nest everything inside the `success` case. Any ideas? Otherwise the optional error might be the least bad option.

Really we should make this method **way** better by using async and throws here - that would solve this problem too.

StripeApplePay/StripeApplePay/Source/ApplePayContext/STPApplePayContext.swift
```diff
@@ -536,121 +557,73 @@ public class STPApplePayContext: NSObject, PKPaymentAuthorizationControllerDeleg
...
-                self.analyticsClient.log(analytic: analytic, apiClient: self.apiClient)
-                if self.didCancelOrTimeoutWhilePending {
-                    self.authorizationController?.dismiss {
-                        DispatchQueue.main.async {
-                            self.callDidCompleteDelegate(status: .success, error: nil)
-                            self._end()
-                        }
+            } else {
+                completion(PKPaymentAuthorizationStatus.failure, error)
+            }
```
wooj-stripe, line 574: We were previously logging an analytic each time handleFinalState was called, but now only do it when `didCancelOrTimeoutWhilePending == true`... and seems like we are no longer logging a significant number of analytics to instrument the code?  Feels like if we were to have another customer issue there will be less events available?

yuki-stripe: We send the finished analytic in `callDidCompleteDelegate`, which is always called at the end of the flow. 

If `didCancelOrTimeoutWhilePending == false`, the code in `paymentAuthorizationControllerDidFinish` does the work to dismiss the VC and call it.

StripeApplePay/StripeApplePay/Source/ApplePayContext/STPApplePayContext.swift
```diff
@@ -667,43 +640,23 @@ public class STPApplePayContext: NSObject, PKPaymentAuthorizationControllerDeleg
...
-                                        handleFinalState(.error, nil, errorAnalytic)
-                                    }
+                            ) { result in
+                                guard case .success(let setupIntent) = result else {
+                                    handleFailure(error: result.error)
+                                    return
+                                }
+                                guard setupIntent.status == .succeeded else {
+                                    // TODO: (MOBILESDK-467) Grab the actual error from the SetupIntent
+                                    handleFailure(error: Error.confirmationFailure)
```
wooj-stripe, line 650: Instead of waiting on MOBILESDK-467, which is unlikely to happen any time soon, maybe we attach the intentId to this error, and then use that to look up errors associated on confirming the intent?

yuki-stripe: We send `intent_id` in the finished analytic!

PR #5671, opened 2025-10-06
CryptoOnramp SDK: Exposes SDK Error Types

mliberatore:
## Summary
Exposes CryptoOnramp-specific error types. 

> [!NOTE]
> `CryptoOnrampCoordinator.Error` is already `public` as a nested type in a `public` extension, so that didn't need to change.  

## Motivation
To provide more specific error codes for certain APIs in the RN SDK. This allows us to match against native types rather than e.g. attempting to match error message strings.

## Testing
Simply built the app.

## Changelog
N/A

### DIFF
--- a/StripeCryptoOnramp/StripeCryptoOnramp/Source/Internal/API Bindings/STPAPIClient+CryptoOnramp.swift
+++ b/StripeCryptoOnramp/StripeCryptoOnramp/Source/Internal/API Bindings/STPAPIClient+CryptoOnramp.swift
@@ -13,7 +13,8 @@
 extension STPAPIClient {
 
     /// Errors that can occur that are specific to usage of crypto endpoints.
-    enum CryptoOnrampAPIError: LocalizedError {
+    @_spi(STP)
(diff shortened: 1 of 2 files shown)


StripeCryptoOnramp/StripeCryptoOnramp/Source/Internal/API Bindings/STPAPIClient+CryptoOnramp.swift
```diff
@@ -41,10 +43,6 @@ extension STPAPIClient {
 
         try validateSessionState(using: linkAccountInfo)
```
mliberatore, line 44:
```swift
private func validateSessionState(using linkAccountInfo: PaymentSheetLinkAccountInfoProtocol) throws {
    guard case .verified = linkAccountInfo.sessionState else {
        throw CryptoOnrampAPIError.linkAccountNotVerified
    }
}
```
I noticed this while making the change here, that the code below was redundant, and likely left over from a deduplication refactor when creating `validateSessionState(using:)`.

PR #5673, opened 2025-10-06
Handle hang when opening the PaymentSheet example playground

mats-stripe:
## Summary

There is a noticeable app hang when opening one of the PaymentSheet Example app playgrounds. This happens because we were synchronously fetching and decoding the settings and appearance objects from UserDefaults on initialization. 

## Motivation

Improve UX

## Testing

Before:

https://github.com/user-attachments/assets/95177d7e-66a1-4a61-ad4f-295b4d49a1c3

After:

https://github.com/user-attachments/assets/9293267f-6fdb-4208-979c-b248ecd506c5

## Changelog

N/a

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/CustomerSheetTestPlaygroundController.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/CustomerSheetTestPlaygroundController.swift
@@ -40,7 +40,6 @@ class CustomerSheetTestPlaygroundController: ObservableObject {
         self.settings = settings
         self.currentlyRenderedSettings = .defaultValues()
         $settings
-            .dropFirst()
             .sink { [weak self] newValue in
                 if let isLoading = self?.isLoading,
                    !isLoading,
--- a/Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
@@ -620,7 +620,7 @@ class PlaygroundController: ObservableObject {
         self.appearance = appearance
         self.currentlyRenderedSettings = .defaultValues()
 
-        $settings.dropFirst().sink { [weak self] newValue in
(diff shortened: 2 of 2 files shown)


Example/PaymentSheet Example/PaymentSheet Example/CustomerSheetTestPlayground.swift
```diff
@@ -28,9 +29,26 @@ struct CustomerSheetTestPlayground: View {
...
+                    Text("Loading playground...")
+                        .font(.subheadline)
+                        .foregroundColor(.secondary)
+                }
+                .frame(maxWidth: .infinity, maxHeight: .infinity)
+                .onAppear {
+                    DispatchQueue.main.async {
+                        isViewReady = true
+                    }
+                }
```
wooj-stripe, line 45: This doesn't seem to have any coordination with the actual task that is being kicked off? Instead of building this into a new screen, maybe we leverage the existing spinner that's part of the playground page?

mats-stripe: There's no direct coordination, but there doesn't need to be. The async task (~70ms) finishes before SwiftUI completes the layout work (~1500ms), so the playground is ready by the time the view renders.

The existing `isLoading` spinner doesn't help here because it only appears after SwiftUI has already laid out all playground elements. That is, the 1.5s hang is SwiftUI's layout pass, not the data loading. That spinner shows up too late to be meaningful here.

Here's the timeline (provided by Claude), which should help illustrate why this works:

```
User taps row
  ↓ (immediate)
Simple loading view renders in ~1ms  ← User sees this instantly!
  ↓ (parallel operations)
  ├─ Task.detached loads UserDefaults (~70ms) ← Happens in background
  └─ Next frame: isViewReady = true
     ↓
     Complex view starts building
     ↓ (~1500ms of layout work)
     onAppear() finally fires
```

PR #5675, opened 2025-10-06
 Fixed the dead code detection workflow to properly update comments when dead code is resolved in a PR.

yuki-stripe:
## Motivation
Annoying to have dead code checker comment when you fix the dead code.

## Testing
todo let's see

## Changelog
not user facing

### DIFF
--- a/.github/workflows/find-dead-code.yml
+++ b/.github/workflows/find-dead-code.yml
@@ -63,10 +63,9 @@ jobs:
 
       - uses: peter-evans/find-comment@v3
         id: find_comment
-        if: env.diff != ''
         with:
           issue-number: ${{ github.event.pull_request.number }}
-          body-includes: '🚨 New dead code detected'
+          body-includes: '[find-dead-code]'
 
       - uses: peter-evans/create-or-update-comment@v3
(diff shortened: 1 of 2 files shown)


yuki-stripe:
Ok it isn't updating the comment in this PR because it doesn't run if you haven't modified *any* swift files. But besides that edge case it should work as expected!

.github/workflows/find-dead-code.yml
```diff
@@ -7,7 +7,6 @@ on:
 
 jobs:
   dead-code-check:
-    if: ${{ !contains(github.event.pull_request.labels.*.name, 'skip dead code check') }}
```
porter-stripe, line 10: Why are we removing this? I think this is intentional to totally skip the dead code check (incase of build failures or something)

yuki-stripe: It used to skip the dead code check entirely if you have the 'skip dead code check' label. It had no relevance to build failures though.

Now if you have the label, it still runs the check (but doesn't fail the PR), that way the comment can tell say "Dead code detected **but ignored**".

PR #5676, opened 2025-10-07
Finish confirmation token confirm

porter-stripe:
## Summary
- Finish confirmation flow for confirmation tokens

## Motivation
- Confirmation tokens

## Testing
- Updates LPM confirm flow tests

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+ConfirmationTokens.swift
```diff
@@ -23,37 +23,111 @@ extension PaymentSheet {
...
 
-            // TODO(porter) Finish rest of CT confirmation
-            stpAssertionFailure("Confirmation Tokens not yet implemented.")
+                    // Check if it needs confirmation
+                    if [STPPaymentIntentStatus.requiresPaymentMethod, STPPaymentIntentStatus.requiresConfirmation].contains(paymentIntent.status) {
+                        // 5a. Client-side confirmation with confirmation token
+                        let paymentIntentParams = STPPaymentIntentParams(clientSecret: paymentIntent.clientSecret)
+                        paymentIntentParams.confirmationToken = confirmationToken.stripeId
+                        paymentIntentParams.returnURL = configuration.returnURL
+                        paymentIntentParams.radarOptions = radarOptions
```
joyceqin-stripe: Realizing that the way I implemented the radarOptions stuff was bad and confusing, I have a [draft pr](https://github.com/stripe/stripe-ios/pull/5685) to move radarOptions into the .saved confirm type. The radarOptions for new pms already lives in the paymentMethodParams, so the radarOptions param for new pms was always nil. Once radar options is moved into .saved, we can remove the radarOptions param and switch on the confirmType.

So then here we would switch on the confirmType, and for `.new(let paymentMethodParams, _, _)`, we would get `paymentIntentParams.radarOptions = paymentMethodParams.radarOptions`, and for `.saved(_, _, _, let radarOptions)` we would do `paymentIntentParams.radarOptions = radarOptions`

porter-stripe:
@joyceqin-stripe can we cover the changes you suggested in https://github.com/stripe/stripe-ios/pull/5685 after this merges?

joyceqin-stripe:
> @joyceqin-stripe can we cover the changes you suggested in #5685 after this merges?

Sure, but then where we're doing `paymentIntentParams.radarOptions = radarOptions` it would have to be something like this to be correct:
`switch confirmType {
    case .new(let paymentMethodParams, _, _, _, _):
        paymentIntentParams.radarOptions = paymentMethodParams.radarOptions
    case .saved:
        paymentIntentParams.radarOptions =radarOptions
 }`

PR #5677, opened 2025-10-07
Gbirch/9 25 card scan close signalling

gbirch-stripe:

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/CardSection/CardSectionWithScannerView.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/CardSection/CardSectionWithScannerView.swift
@@ -84,6 +84,11 @@ final class CardSectionWithScannerView: UIView {
         becomeFirstResponder()
     }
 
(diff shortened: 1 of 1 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/CardSection/CardSectionWithScannerView.swift
```diff
@@ -107,22 +112,19 @@ final class CardSectionWithScannerView: UIView {
...
-        //    but it adds a bit of extra safety. This can be removed in the future.
-        if newWindow != nil && !cardScanningView.isHidden && opensCardScannerAutomatically {
+        if newWindow != nil && !cardScanningView.isHidden {
             cardScanningView.startScanner()
             becomeFirstResponder()
         }
+
+        // If we are leaving the screen, we should stop (but not close) the scanner
+        if newWindow == nil {
+            cardScanningView.stopScanner()
```
davidme-stripe, line 125: Are there other situations where we'd want to stop the scanner but another field isn't editing?

PR #5681, opened 2025-10-08
CryptoOnramp SDK: Begins Seamless Sign-In

mliberatore:
## Summary
Adds a new API to `CryptoOnrampCoordinator`:

```swift
/// Authenticates the user with an encrypted Link auth token.
/// - Parameter linkAuthTokenClientSecret: An encrypted one-time-use auth token that, upon successful validation, leaves the Link account’s consumer session in an already-verified state, allowing the client to skip verification.
/// Throws if the user is the auth token is expired, has already been used, has been revoked, or an API error occurs.
func authenticateUserWithToken(_ linkAuthTokenClientSecret: String) async throws
```

This PR supports this addition with the necessary adjustments to `StripePaymentSheet` and `LinkController`, ultimately passing a new parameter (`link_auth_token_client_secret`) to the `/consumers/mobile/sessions/lookup`, where upon success, the resulting consumer session will be in an already-validated state.

## Motivation
[Seamless Sign-In Initiative](https://docs.google.com/document/d/1OEA4tnyLUpE4xoZiYQBYcn070M1WLVrgXupzDa3N_M0/edit?tab=t.0#heading=h.ezs4kyft654m) 

## Testing
I‘ve manually tested the API from the example app to ensure we get an expected error with a bogus token.

```
[Stripe SDK]: POST "/v1/consumers/mobile/sessions/lookup" 401 req_Tje8hHjb8V01Ug
Printing description of error:
▿ StripeError
  ▿ apiError : StripeAPIError
    - type : StripeCore.StripeAPIError.ErrorType.invalidRequestError
    ▿ code : Optional<String>
      - some : "link_auth_token_invalid"
    - docUrl : nil
    ▿ message : Optional<String>
      - some : "You must provide a valid link_auth_token_id."
    - param : nil
    ▿ statusCode : Optional<Int>
      - some : 401
    ▿ requestID : Optional<String>
      - some : "req_Tje8hHjb8V01Ug"
    ▿ _allResponseFieldsStorage : Optional<NonEncodableParameters>
      - some : 3 fields
```

### DIFF
--- a/Stripe/StripeiOSTests/LinkInlineSignupElementSnapshotTests.swift
+++ b/Stripe/StripeiOSTests/LinkInlineSignupElementSnapshotTests.swift
@@ -167,7 +167,6 @@ extension LinkInlineSignupElementSnapshotTests {
 
         func lookupLinkAuthToken(
             _ linkAuthTokenClientSecret: String,
-            doNotLogConsumerFunnelEvent: Bool,
             requestSurface: StripePaymentSheet.LinkRequestSurface,
             completion: @escaping (Result<StripePaymentSheet.PaymentSheetLinkAccount?, any Error>) -> Void
         ) {
--- a/Stripe/StripeiOSTests/LinkSignupViewModelTests.swift
+++ b/Stripe/StripeiOSTests/LinkSignupViewModelTests.swift
@@ -286,7 +286,6 @@ extension LinkInlineSignupViewModelTests {
 
         func lookupLinkAuthToken(
             _ linkAuthTokenClientSecret: String,
-            doNotLogConsumerFunnelEvent: Bool,
             requestSurface: StripePaymentSheet.LinkRequestSurface,
             completion: @escaping (Result<StripePaymentSheet.PaymentSheetLinkAccount?, any Error>) -> Void
         ) {


StripeCryptoOnramp/StripeCryptoOnramp/Source/Internal/CryptoOnrampAnalyticsEvent.swift
mats-stripe, line 1: Thanks for adding these!

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/STPAPIClient+Link.swift
```diff
@@ -25,6 +25,7 @@ extension STPAPIClient {
         canSyncAttestationState: Bool,
         doNotLogConsumerFunnelEvent: Bool,
         requestSurface: LinkRequestSurface = .default,
+        linkAuthTokenClientSecret: String? = nil,
```
mats-stripe: It's a bit more overheard, but instead of nesting this in `lookupConsumerSession`, could we have a separate API for this along the lines of `lookupLinkAuthToken`? We did this below for `lookupLinkAuthIntent`

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkController.swift
```diff
@@ -245,6 +245,33 @@ import UIKit
...
+            linkAccountService: linkAccountService,
+            requestSurface: requestSurface
+        ) { result in
+            switch result {
+            case .success(let linkAccount):
+                LinkAccountContext.shared.account = linkAccount
+                if linkAccount != nil {
+                    completion(.success(()))
+                } else {
+                    completion(.failure(PaymentSheetError.linkLookupNotFound(serverErrorMessage: "")))
```
mats-stripe, line 267: Just so I understand, the presence of a `linkAccount` means the auth token lookup succeeded. Otherwise, the token was invalid?

StripeCryptoOnramp/StripeCryptoOnramp/Source/Internal/CryptoOnrampAnalyticsEvent.swift
```diff
@@ -30,6 +31,7 @@ enum CryptoOnrampAnalyticsEvent {
     case linkPhoneNumberUpdated
     case linkUserAuthenticationStarted
     case linkUserAuthenticationCompleted
+    case linkUserAuthenticationWithTokenCompleted
```
mats-stripe, line 34: Just for consistency with the `linkUserAuthentication` events, would it make sense to also have a `Started` event for this API?

mliberatore: It seemed like non-Stripe-UI operations didn’t have "started" events. I’m not 100% sure if that's the distinction that was made for the analytics events for coordinator operations, but looking through them with @Twigz, that's the conclusion we came to.

Very very simple to add a `Started` variant to this, but just want to make sure we're following the convention set here. For example:
- `sessionCreated`
- `linkRegistrationCompleted`
- `linkAccountLookupCompleted`
- `linkPhoneNumberUpdated`
- `kycInfoSubmitted`
- `walletRegistered`
- `cryptoPaymentTokenCreated`
- `userLoggedOut`

all don't have a `Started` variant, where the following, which present Stripe UI, do:
- `linkUserAuthenticationStarted`
- `linkAuthorizationStarted`
- `identityVerificationStarted`
- `collectPaymentMethodStarted`
- `checkoutStarted` (might be an exception, since no Stripe UI is shown)

LMK what you think, please!

mats-stripe: Oh I forgot this API doesn't show any UI! Just `Completed` or `Error` is fine for here!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkController.swift
mliberatore: My current understanding from [the docs](https://docs.google.com/document/d/1KfdabfM6S289jnqZgaNjq6Sg-5jl28WD4OGA-h4CKHQ/edit?tab=t.0#heading=h.cj165yzgomwg) is that if the auth token is expired, used already, revoked, or not found, the API will error with 401 or 404.

This `else` case here simply handles the case where the common API could succeed, but with a response type of `.notFound` or `.noAvailableLookupParams`. While I don’t _think_ we’ll hit these cases given the nature of the API and what we're passing and expecting in response (superseded by the 401s or 404), this treats `.notFound` as a failure. `.noAvailableLookupParams` would also not be the result type, as we require the token in this new API, and that result only gets returned if it (along with email and emailSource) are also missing. Reaching this case should be impossible outside of programmer error, but handled nonetheless as a failure, since without an account, we've ultimately failed to retrieve a validated session.

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/STPAPIClient+Link.swift
mliberatore: Sure, I’ll make the change!

mliberatore: I’ve updated the naming to be consistent throughout the various layers using your suggestion.

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/STPAPIClient+Link.swift
```diff
@@ -58,6 +58,39 @@ extension STPAPIClient {
...
+            var parameters: [String: Any] = [
+                "request_surface": requestSurface.rawValue,
+                "session_id": sessionID,
+                "link_auth_token_client_secret": linkAuthTokenClientSecret
+            ]
+
+            parameters["customer_id"] = customerID
+
+            if doNotLogConsumerFunnelEvent {
+                parameters["do_not_log_consumer_funnel_event"] = true
```
mats-stripe: We can get rid of this parameter entirely for this API

mliberatore: Yes, updated!

PR #5683, opened 2025-10-09
Update logic for isEnabledInMerchantApp

wooj-stripe:
## Summary
We detected that when compiling with xcode26, but executing on <iOS26 target, the done button becomes unresponsive. But this applies to a broader set of problems where we may be executing code that was intended for liquid glass to execute on devices < iOS26.

## Motivation
Fix regression

## Testing
Manually

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## x.x.x yyyy-yy-yy
 ### PaymentSheet
-* [Fixed] Logic on detecting liquid glass executing. Link visual styling issues are present and `DoneButtonToolbar` was not tappable on devices using < iOS26 when compilling with Xcode26.
+* [Fixed] Fixed a bug where the`DoneButtonToolbar` was not tappable on devices < iOS 26 when compiled with Xcode26.
+* [Fixed] Fixed minor styling issues with Link UI on < iOS 26 when compiled with Xcode26.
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## x.x.x yyyy-yy-yy
+### PaymentSheet
+* [Fixed] Logic on detecting liquid glass executing, to avoid executing liquid glass specific code on <iOS26 devices.
```
yuki-stripe: This won't make sense to users - what user-facing thing did this fix? 

I would mention the done toolbar bug specifically. I think the other bugfixes amount to using the wrong styling in Link.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## x.x.x yyyy-yy-yy
+### PaymentSheet
+* [Fixed] Logic on detecting liquid glass executing. Link visual styling issues are present and `DoneButtonToolbar` was not tappable on devices using < iOS26 when compilling with Xcode26.
```
yuki-stripe:
```suggestion
* [Fixed] Fixed a bug where the`DoneButtonToolbar` was not tappable on < iOS 26 when compiled with Xcode 26.
* [Fixed] Fixed minor styling issues with Link UI on < iOS 26 when compiled with Xcode 26.
```

PR #5684, opened 2025-10-09
Add client_attribution_metadata to payment details and adjust for LinkController

joyceqin-stripe:
## Summary

Added `client_attribution_metadata` to the `/v1/consumers/payment_details` endpoint. Added a PaymentSheetAnalyticsHelper integration shape, .linkController, because LinkController is not a part of MPE and therefore should not send `client_attribution_metadata`. Added `client_attribution_metadata` to financial connections Link calls
## Motivation

`/v1/consumers/payment_details` has come into scope for Project Weave
## Testing

CI
## Changelog

N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+ConfirmationTokens.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+ConfirmationTokens.swift
@@ -35,11 +35,7 @@ extension PaymentSheet {
                     // Link saved payment methods don't require ephemeral keys, API will error if provided
                     guard !isSavedFromLink(from: confirmType) else { return nil }
 
-                    guard let customer = configuration.customer else {
-                        stpAssertionFailure("Customer should exist when using saved payment method")
-                        return nil
-                    }
-                    return customer.ephemeralKeySecret(basedOn: elementsSession)
+                    return configuration.customer?.ephemeralKeySecret(basedOn: elementsSession)
                 }()
 
                 // 2. Create the ConfirmationToken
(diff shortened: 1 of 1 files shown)


StripePaymentSheet/StripePaymentSheet/Source/Analytics/PaymentSheetAnalyticsHelper.swift
```diff
@@ -69,7 +81,7 @@ final class PaymentSheetAnalyticsHelper {
                 case (true, true):
                     return .mcInitCustomCustomerApplePay
                 }
-            case .complete:
+            case .complete, .linkController:
```
tillh-stripe, line 84: What's the reasoning for treating `linkController` like `complete`? Should we have dedicated LinkController events?

joyceqin-stripe: Perhaps. I treated it like `complete` because if you look in `LinkController.swift`, its original analytics helper was with the `complete` shape, so I was just trying to maintain status quo while being able to differentiate between Link requests that originated from MPE vs LinkController.

tillh-stripe: Got it. I'll make it a to-do for my team to set up separate events for `LinkController`.

StripePaymentSheet/StripePaymentSheet/Source/Internal/FC Lite/FinancialConnectionsLite.swift
```diff
@@ -20,6 +20,8 @@ import UIKit
     /// Any additional Elements context useful for the Financial Connections SDK.
     @_spi(STP) public var elementsSessionContext: ElementsSessionContext?
 
+    @_spi(STP) public var clientAttributionMetadata: STPClientAttributionMetadata?
```
tillh-stripe: Any reason we can't use the client attribution metadata in `ElementsSessionContext`? Is it not being set for FC Lite?

tillh-stripe (commented):
One clarifying question.

StripePaymentSheet/StripePaymentSheet/Source/Internal/FC Lite/FinancialConnectionsLite.swift
joyceqin-stripe: Good catch— that was dead code, but the dead code check didn't catch it because it's a public var. I've removed it. I was originally passing clientAttributionMetadata as a parameter everywhere before I moved it into ElementsSessionContext.

mats-stripe:
Let me know when this is ready for review @joyceqin-stripe

joyceqin-stripe:
> Let me know when this is ready for review @joyceqin-stripe

Should be ready now! Thanks

PR #5704, opened 2025-10-15
Fixed a bug with `FlowController` and `EmbeddedPaymentElement` in dark mode where `paymentOption.image` could be a blank icon for Satispay, Alma, Billie, Multibanco, Paynow, Promptpay, and Sunbit

yuki-stripe:
## Summary
**Context**
v1/e/s sends down image URLs for the payment method icons we’ll display. It sends a ‘light theme’ URL and an optional ‘dark theme’ URL.

When FC/Embedded has a dark background color, it downloads the “dark theme” images.

**The bug**
When the paymentOption image is created, it always looks for the “light theme” image. This will fail, since the light theme image was not downloaded, and fall back to the bundled image. If the bundled image doesn’t exist, it returns a 1x1 transparent image.

**Scope**
FC / Embedded + dark background colors (e.g. device is in dark mode + default appearance) + all PMs with “dark theme” image will either have

1. A potentially outdated icon, if we bundled it with the SDK
2. A blank icon, if we didn’t bundle it with the SDK

**The fix**
Make paymentOption respect the device’s dark mode setting. 

Note this doesn't fully solve the problem - it still vends the wrong image if the merchant uses a light background color in dark mode. Ticketing that up here: https://jira.corp.stripe.com/browse/MOBILESDK-4130

## Testing
Manually tested - there's no way to force dark mode in XCTest.

## Changelog
See CHANGELOG

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Categories/UIApplication+StripePaymentSheet.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Categories/UIApplication+StripePaymentSheet.swift
@@ -38,7 +38,7 @@ extension UIApplication {
         if activeScene == nil {
             let errorAnalytic = ErrorAnalytic(
                 event: .unexpectedPaymentSheetError,
-                error: Error.missingActiveScene,
+                error: Error.missingActiveScene
             )
             STPAnalyticsClient.sharedClient.log(analytic: errorAnalytic)
             stpAssertionFailure("Couldn't find active scene!")
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
-## 24.24.3 2025-10-13
+## X.Y.Z x-y-z
 ### PaymentSheet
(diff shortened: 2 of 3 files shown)


StripePaymentSheet/StripePaymentSheet/Source/Categories/UIApplication+StripePaymentSheet.swift
```diff
@@ -18,7 +19,33 @@ extension UIApplication {
             .sorted { firstWindow, _ in firstWindow.isKeyWindow }
         return windows.first
         #else
-        return windows.first { $0.isKeyWindow }
+        return activeScene?.windows.first { $0.isKeyWindow }
```
gbirch-stripe, line 22: What's the reason for this change?

yuki-stripe: [windows](https://developer.apple.com/documentation/uikit/uiapplication/windows) is deprecated!

gbirch-stripe (approved):

PR #5714, opened 2025-10-16
Pass new parameter to API when resending an SMS OTP code

mats-stripe:
## Summary

Adds a new parameter `is_resend_sms_code=true` to the `/start_verification` API when the 'Resend code' button is tapped for SMS verification.

## Motivation

https://jira.corp.stripe.com/browse/RUN_LINK_MOBILE-43

## Testing

Example request: https://admin.corp.stripe.com/request-log/req_m2v6Xne08uTjdf

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/STPAPIClient+Link.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/STPAPIClient+Link.swift
@@ -652,6 +652,7 @@ extension STPAPIClient {
         for consumerSessionClientSecret: String,
         type: ConsumerSession.VerificationSession.SessionType,
         locale: Locale,
+        isResendingSmsCode: Bool = false,
         requestSurface: LinkRequestSurface = .default,
         completion: @escaping (Result<ConsumerSession, Error>) -> Void
     ) {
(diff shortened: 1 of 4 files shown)


tillh-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/STPAPIClient+Link.swift
```diff
@@ -667,13 +668,18 @@ extension STPAPIClient {
...
-        let parameters: [String: Any] = [
+        var parameters: [String: Any] = [
             "credentials": ["consumer_session_client_secret": consumerSessionClientSecret],
             "type": typeString,
             "locale": locale.toLanguageTag(),
             "request_surface": requestSurface.rawValue,
         ]
 
+        // This parameter is specifically when resending SMS codes.
+        if isResendingSmsCode {
```
jkelle-stripe, line 679: [non-blocking] For my understanding, do we typically not send the analytic params if they're false?  I guess it saves space in the request if we assume no value present is falsy.

mats-stripe: I would say this is atypical in fact. But in this case the backend logic only changes when the value is `true`, so I'm omitting it other

PR #5715, opened 2025-10-16
Refactor AfterpayPriceBreakdownView, use system image instead of asset

gbirch-stripe:
## Summary
1. refactor AfterpayPriceBreakdownView to separate out the attributed string logic
2. remove info icon that was identical to system one, use system one instead

## Motivation
1. pave way for PMME to leverage the same attributed string logic
2. reduce SDK by removing unneeded asset

## Testing
existing snapshot tests

## Changelog
N/A


Tests/ReferenceImages_64/StripePaymentSheetTests.EmbeddedPaymentMethodsViewSnapshotTests/testEmbeddedPaymentMethodsView_flatWithDisclosure_customDisclosureView_info_icon@3x.png
gbirch-stripe, line 1: This was using the custom asset I removed. Its purpose is to test the rendering of SVG assets, so I replaced it with another

yuki-stripe (approved):

PR #5718, opened 2025-10-17
Link Inline Signup Mode Analytics

jkelle-stripe:
## Summary
Adds analytics to monitor the Link inline signup mode. Also, creates an analytics helper for Link surfaces to log to analytics v2.

## Motivation
Provide more observability for the mode used by Link inline signup. 

## Testing
Manual testing of each mode. Able to see analytics in our dashboard.

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
+++ b/Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
@@ -629,9 +629,9 @@ class PaymentSheetDeferredUITests: PaymentSheetUITestCase {
         XCTAssertEqual(
             // Ignore luxe_* analytics since there are a lot and I'm not sure if they're the same every time
             // filter out async passive captcha logs
-            analyticsLog.map({ $0[string: "event"] }).filter({ $0 != "luxe_image_selector_icon_from_bundle" && $0 != "luxe_image_selector_icon_downloaded" && !($0?.starts(with: "elements.captcha.passive") ?? false) && !($0?.starts(with: "link") ?? false) }),
+            analyticsLog.map({ $0[string: "event"] }).filter({ $0 != "luxe_image_selector_icon_from_bundle" && $0 != "luxe_image_selector_icon_downloaded" && !($0?.starts(with: "elements.captcha.passive") ?? false) }),
             // fraud detection telemetry should not be sent in tests, so it should report an API failure
-            ["mc_complete_init_applepay", "mc_load_started", "mc_load_succeeded", "fraud_detection_data_repository.api_failure", "mc_complete_sheet_newpm_show", "mc_lpms_render", "mc_form_shown"]
+            ["mc_complete_init_applepay", "mc_load_started", "mc_load_succeeded", "fraud_detection_data_repository.api_failure", "mc_complete_sheet_newpm_show", "mc_lpms_render", "mc_form_shown", "link.inline_signup.shown"]
         )
         XCTAssertEqual(analyticsLog.filter({ !($0[string: "event"]?.starts(with: "elements.captcha.passive") ?? false || $0[string: "event"]?.starts(with: "link") ?? false) }).last?[string: "selected_lpm"], "card")
 


StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Extensions/PaymentSheetAnalyticsHelper+Link.swift
```diff
@@ -0,0 +1,56 @@
...
+}
+
+enum LinkAnalyticsV2Event {
+
+    case inlineSignupShown(mode: LinkInlineSignupViewModel.Mode)
+
+    var eventName: String {
+        switch self {
+        case .inlineSignupShown:
+            "link_inline_signup_shown"
```
jkelle-stripe: Should this be `link.inline_signup_shown`?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory+Card.swift
```diff
@@ -148,7 +148,8 @@ extension PaymentSheetFormFactory {
                 accountService: accountService,
                 allowsDefaultOptIn: allowsLinkDefaultOptIn,
                 signupOptInFeatureEnabled: signupOptInFeatureEnabled,
-                signupOptInInitialValue: signupOptInInitialValue
+                signupOptInInitialValue: signupOptInInitialValue,
+                analyticsHelper: analyticsHelper
```
jkelle-stripe, line 152: The dependency injection for the analytics helper here is `LinkInlineSignupElement` -> `LinkInlineSignupViewModel` and the `LinkInlineSignupView` calls the analytics log function on the `LinkInlineSignupViewModel`.

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Extensions/PaymentSheetAnalyticsHelper+Link.swift
mats-stripe: How about `link.inline_signup.shown`? Matches the pattern of our other events like `link.account_lookup.complete`

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Extensions/PaymentSheetAnalyticsHelper+Link.swift
mats-stripe, line 1: I think it's good to keep all the Link events in their own file like this, but I think they should all be together. There's some more Link events here:

https://github.com/stripe/stripe-ios/blob/master/StripeCore/StripeCore/Source/Analytics/STPAnalyticEvent.swift#L142-L173

What do you think about either bringing all those events to this file, or moving this event to where the others are defined?

mats-stripe (commented):
Looks good! Few comments / suggestions

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Extensions/PaymentSheetAnalyticsHelper+Link.swift
jkelle-stripe: Totally agree that having them together is helpful. I had this in it's own file since this analytic is logging to the `R` instead of `Q` and I thought all of those analytics were being sent to `Q`, but maybe that distinction isn't valuable in keeping them separate?

mats-stripe: Ahh good point. Sorry for the churn on this, but let's stick with the existing implementation and send everything to `q` (using `STPAnalyticsClient` / `PaymentSheetAnalyticsHelper.analyticsClient`) so all our events show up in the same table

jkelle-stripe: All good! Will do

jkelle-stripe:
@mats Updated the backend from r to q! [3f5cdf3](https://github.com/stripe/stripe-ios/pull/5718/commits/3f5cdf31da9de8f41962a3dcd3d2bd4665cba3a7)

mats-stripe (approved):
Looks great!

Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
mats-stripe: Just curious, any reason to filter out the Link event(s) here instead of adding them to the list of expected events?

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
```diff
@@ -253,8 +253,8 @@ class PaymentSheetStandardUITests: PaymentSheetUITestCase {
         XCTAssertTrue(app.staticTexts["Card information"].waitForExistence(timeout: 2))
 
         // Should fire the `mc_form_shown` event w/ `selected_lpm` = card
-        XCTAssertEqual(analyticsLog.last?[string: "event"], "mc_form_shown")
-        XCTAssertEqual(analyticsLog.last?[string: "selected_lpm"], "card")
+        XCTAssertEqual(analyticsLog[12][string: "event"], "mc_form_shown")
```
mats-stripe: I know this is a unit test so it's not a huge deal, but could we use something like `analyticsLog.contains(where:)` instead of hardcoding the index here?

Example/PaymentSheet Example/PaymentSheetUITest/EmbeddedUITest.swift
jkelle-stripe: Fair point I didn't consider! I'll add it in

PR #5720, opened 2025-10-17
Add attempt id to paymenthandler

yuki-stripe:
Adds actionID, a uuid unique to a given `confirm` or `handleNextAction` call, so that we can group together all the analytics sent during a confirmation / next action handling.

Note that Session ID is not good enough, since a single session may have multiple calls to confirm.

## Testing
Add tests

## Changelog
Not user facing

### DIFF
--- a/StripePayments/StripePayments/Source/PaymentHandler/STPAnalyticsClient+STPPaymentHandler.swift
+++ b/StripePayments/StripePayments/Source/PaymentHandler/STPAnalyticsClient+STPPaymentHandler.swift
@@ -180,40 +180,4 @@ extension STPPaymentHandler {
         ])
         analyticsClient.log(analytic: analytic, apiClient: apiClient)
     }
-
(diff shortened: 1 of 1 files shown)


StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
```diff
@@ -643,7 +650,7 @@ public class STPPaymentHandler: NSObject {
                 return
             }
             if let setupIntent, error == nil {
-                self.handleNextAction(for: setupIntent, with: authenticationContext, returnURL: returnURL, completion: completion)
+                self.handleNextAction(for: setupIntent, with: authenticationContext, returnURL: returnURL, shouldSendAnalytic: false, completion: completion)
```
yuki-stripe, line 653: Whoops. This fixes a bug where we were double reporting analytics for handleNextAction(for setupintent)

PR #5721, opened 2025-10-19
Add check-scanning component and callback support

bredmond-stripe:
## Summary

Add check-scanning component and callback support

 * Add a new view controller for check-scanning which requires callback support
 * Add a fully generalized iOS implementation of the mobile supplemental function message protocol

Keeping these in the same PR because the new use case helps prove and test the callback implementation (would otherwise need a test-only supplemental function to do that)

## Motivation

Supporting the check-scanning component in iOS

## Testing

Unit test coverage and manual simulator for a test app 

## Changelog

skip dead code check

- Need to maintain a strong reference to `CheckScanController` even if it's not used

### DIFF
--- a/StripeConnect/StripeConnect/Source/Components/CheckScanningController.swift
+++ b/StripeConnect/StripeConnect/Source/Components/CheckScanningController.swift
@@ -11,7 +11,7 @@ public final class CheckScanningController {
 
     private(set) var webVC: ConnectComponentWebViewController!
 
-    private var retainedSelf: CheckScanningController?
+    var retainedSelf: CheckScanningController?
 
     public weak var delegate: CheckScanningControllerDelegate?
 


gimenete-stripe (approved):
LGTM! 👍

StripeConnect/StripeConnect/Source/Components/CheckScanningViewController.swift
```diff
@@ -0,0 +1,55 @@
+import UIKit
+
+@_spi(DashboardOnly)
+@available(iOS 15, *)
+public final class CheckScanningViewController: UIViewController {
```
torrance-stripe: Hmm so this needs to a "workflow" component like account onboarding that's full-screen only and can't be flexibly presented. The implementation should more closely resemble `AccountOnboardingController.swift`

bredmond-stripe: done

torrance-stripe (approved):
🚢

PR #5732, opened 2025-10-22
Send client_attribution_metadata on all confirmation calls + tests

joyceqin-stripe:
## Summary

Added client_attribution_metadata to confirm call for Apple Pay and confirmation tokens (already taken care of the rest) and added network stub testing with stub verification.

## Motivation

Previously, the expected behavior was that if payment method creation and intent confirmation happen on separate requests in the same button tap (like with deferred), include the metadata in payment method creation rather than intent confirmation.

But now the data team wants client_attribution_metadata sent on both in order to make it explicit that the intent-confirm and PM-create happened on the same session.
https://stripe.slack.com/archives/C07PXHWQMP0/p1760553051428099?thread_ts=1760449818.955979&cid=C07PXHWQMP0
## Testing

Updated stubs

Apple Pay confirm request: https://admin.corp.stripe.com/request-log/req_J5HL6cBFivDLLT
## Changelog

N/A

### DIFF
--- a/.github/workflows/check-translations.yml
+++ b/.github/workflows/check-translations.yml
@@ -46,12 +46,12 @@ jobs:
           else
             echo "Missing translations detected, continue workflow..."
           fi
-      - uses: peter-evans/find-comment@v3.0.0
+      - uses: davidme-stripe/find-comment@v3.0.0
         id: find_comment
         with:
           issue-number: ${{ github.event.pull_request.number }}
           body-includes: Missing Translations
-      - uses: peter-evans/create-or-update-comment@v4
+      - uses: davidme-stripe/create-or-update-comment@v4
         if: steps.missing.outputs.translations
         with:
           body: |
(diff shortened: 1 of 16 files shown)


StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/ClientAttributionMetadataCoverageTests.swift
```diff
@@ -0,0 +1,777 @@
...
+//  PaymentSheetClientAttributionMetadataTests.swift
+//  StripePaymentSheetTests
+//
+//  Created by Joyce Qin on 10/22/25.
+//
+
+@testable@_spi(STP) @_spi(CustomerSessionBetaAccess) @_spi(ConfirmationTokensPublicPreview) import StripePaymentSheet
+@testable@_spi(STP) import StripePaymentsTestUtils
+
+class ClientAttributionMetadataCoverageTests: STPNetworkStubbingTestCase {
```
porter-stripe: Do we really need all these? Why can't we add some assertions or modify the existing PaymentSheetLPMConfirmFlows? These look very similar/duplicative.

PR #5733, opened 2025-10-23
PMME networking

gbirch-stripe:
## Summary
PMME networking code

## Motivation
https://docs.google.com/document/d/13FuGJEOV0784KylxGzfSwaqYVhCv1lWcPMF5R0BlWrE/edit?pli=1&tab=t.0#heading=h.k2l70v5oo6d0

## Testing
parsing tests will be added later with PR that interprets API data

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+API.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+API.swift
@@ -83,9 +83,7 @@ extension PaymentMethodMessagingElement {
             let text: String
 
             struct IconInfo: Decodable {
-                let height: Int
                 let url: URL
-                let width: Int
             }
         }
 


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PaymentMethodMessagingElement.swift
```diff
@@ -0,0 +1,87 @@
...
+        /// Defaults to `nil`.
+        /// If `nil`, uses your preferences from the Stripe dashboard to show the relevant payment methods. See Dynamic payment methods for more information.
+        public var paymentMethodTypes: [STPPaymentMethodType]?
+
+        /// Describes the visual appearance of the PaymentMethodMessaingElement.
+        public var appearance: PaymentMethodMessagingElement.Appearance = PaymentMethodMessagingElement.Appearance()
+    }
+}
+
+public enum PaymentMethodMessagingElementError: Error, LocalizedError {
```
yuki-stripe: Ooh was this in the API Review?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+API.swift
```diff
@@ -0,0 +1,85 @@
...
+//
+
+import Foundation
+@_spi(STP) import StripeCore
+@_spi(STP) import StripePayments
+
+private let pmmeApiEndpoint = URL(string: "https://ppm.stripe.com/config")!
+
+extension PaymentMethodMessagingElement {
+    static func get(configuration: Configuration) async throws -> APIResponse {
```
yuki-stripe, line 15: Can we add a quick docstring for this method explaining what it returns, the method name`PaymentMethodMessagingElement.get` is not immediately obvious

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+API.swift
```diff
@@ -0,0 +1,85 @@
...
+        if let paymentMethods = configuration.paymentMethodTypes {
+            parameters["payment_methods"] = paymentMethods.map { $0.identifier }
+        }
+        return try await withCheckedThrowingContinuation { continuation in
+            configuration.apiClient.get(url: pmmeApiEndpoint, parameters: parameters) { (result: Result<APIResponse, Error>) in
+                switch result {
+                case .success(let apiResponse):
+                    continuation.resume(returning: apiResponse)
+                case .failure:
+                    continuation.resume(throwing: PaymentMethodMessagingElementError.unexpectedResponseFromStripeAPI)
```
yuki-stripe: sort of an api review discussion but i don't think we should obscure/overwrite these errors

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+API.swift
```diff
@@ -0,0 +1,85 @@
...
+                }
+            }
+        }
+    }
+}
+
+extension PaymentMethodMessagingElement {
+
+    struct APIResponse: Decodable {
```
yuki-stripe, line 66: Is there anything we can link to that future readers can reference to learn more, maybe a backend file or trailhead or something

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+API.swift
gbirch-stripe: This is an internal method that users won't interact with, so I didn't think that level of documentation is necessary? Happy to add it if you prefer, though

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+API.swift
gbirch-stripe: Got it, removed. My concern at the time was that users don't understand/want generic errors from iOS (like decoding failed), but they can see that it's coming from PMME and it seems we do that elsewhere, so it makes sense that we just pass them on as-is

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PaymentMethodMessagingElement.swift
gbirch-stripe: No, the API review didn't include any errors. Open to feedback or opening discussion about them. Tried to follow existing ps/embedded patterns as much as possible

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+API.swift
gbirch-stripe: I'll look into this!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PaymentMethodMessagingElement.swift
yuki-stripe: 👍 I think publicizing PaymentSheetError was a mistake, I'll write more in #mobile-api-review but for now to not block, can we just not make this public?

gbirch-stripe: done 👍

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+API.swift
yuki-stripe: Yeah not looking for full documentation here. I do think it could be helpful to spell it out for future maintainers. Just something like, this hits the endpoint and returns all the data PMME needs to display for the given configuration.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+Internal.swift
```diff
@@ -0,0 +1,22 @@
...
+//
+//  Created by George Birch on 10/23/25.
+//
+
+import Foundation
+
+enum PaymentMethodMessagingElementError: Error, LocalizedError {
+    case missingPublishableKey
+    case unexpectedResponseFromStripeAPI
+    case unknown
```
yuki-stripe: can delete these since they're unused?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+API.swift
```diff
@@ -0,0 +1,90 @@
...
+        }
+        if let stripeAccount = configuration.apiClient.stripeAccount {
+            parameters["stripe_account"] = stripeAccount
+        }
+        return try await withCheckedThrowingContinuation { continuation in
+            configuration.apiClient.get(url: pmmeApiEndpoint, parameters: parameters) { (result: Result<APIResponse, Error>) in
+                switch result {
+                case .success(let apiResponse):
+                    continuation.resume(returning: apiResponse)
+                case .failure(let error):
```
yuki-stripe, line 41: Ooh if we fail to parse we should probably send an [unexpected error](https://trailhead.corp.stripe.com/docs/mobile-sdk/analytics/error-analytics#unexpected-errors) so we get an alert

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+API.swift
```diff
@@ -39,6 +39,16 @@ extension PaymentMethodMessagingElement {
                 case .success(let apiResponse):
                     continuation.resume(returning: apiResponse)
                 case .failure(let error):
+                    if let decodingError = error as? DecodingError,
+                        case let .keyNotFound(codingKey, context) = decodingError {
```
yuki-stripe: Are not [all DecodingError cases](https://developer.apple.com/documentation/swift/decodingerror) unexpected and things we should alert on?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+API.swift
```diff
@@ -39,6 +39,16 @@ extension PaymentMethodMessagingElement {
...
+                        stpAssertionFailure(context.debugDescription)
+                        let problemKey = codingKey.stringValue
+                        let problemPath = context.codingPath
+                        let fullPath = problemPath.reduce("") { $0 + $1.stringValue + "." } + problemKey
+                        let error = PaymentMethodMessagingElementError.unexpectedResponseFromStripeAPI
```
yuki-stripe: Is this necessary? If you just send the decodingError as-is what happens? I think/hope it sends useful enough info by default.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+API.swift
gbirch-stripe: Agreed, thanks!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+API.swift
gbirch-stripe: When I wrote this I did a test of what actually gets logged to analytics and there unfortunately isn't any information about about how decoding failed. My thinking was that we probably usually avoid sending this sort of thing since the response structure of some API responses could be PII-y, but this API doesn't return any user data at all (other than which dynamic payment methods are enabled) and the structure is totally standardized, so it seems safe and useful here. Open to discussion if you think this info would be noise / not useful.

yuki-stripe: Just double checked, it sends something like this by default `(key: "error_type", value: "Swift.DecodingError"), (key: "error_code", value: "keyNotFound")`. So I think sending the problem key name is great (and safe)!

I think we could even add it to the `serializeForV1Analytics` method too so that everyone benefits.

PR #5737, opened 2025-10-23
Scale BNPL promo string images with font size

gbirch-stripe:
## Summary
- Scale the BNPL logo and info icon with the font size
- Reduce space between info icon and text
- Use text color instead of icon color for info icon

## Motivation
Better UI when using small or larger font

## Testing
- updated snapshot tests
- added unit test

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/Categories/NSAttributedString+Stripe.swift
```diff
@@ -21,18 +21,21 @@ extension NSMutableAttributedString {
         textColor: UIColor,
         infoIconColor: UIColor,
         template: String,
-        substitution: (String, UIImage)?
+        substitution: (String, UIImage, scale: CGFloat)?
```
yuki-stripe: I was going to ask to document this param, and asked an llm to do so:
```
 * - Parameters:
 *   - substitution: An optional tuple for partner logo replacement containing:
 *     - `String`: The placeholder text in the template to be replaced e.g., "{partner_logo}"
 *     - `UIImage`: The partner logo image to insert in place of the placeholder
 *     - `scale`: Optional scaling factor for the logo image relative to the font size (defaults to 1.0 if not provided)
 *
```

Not sure, is it possible to name these params to make it clearer?

ooc why can't we look at the font to determine the scale? Like the intention is to scale the image with the font size, which I might think is knowable from the `font`.

StripePaymentSheet/StripePaymentSheet/Source/Categories/NSAttributedString+Stripe.swift
```diff
@@ -77,19 +80,22 @@ extension NSMutableAttributedString {
 
     private static func attributedStringOfImageWithoutLink(
         uiImage: UIImage,
-        font: UIFont
+        font: UIFont,
+        scale: CGFloat
     ) -> NSAttributedString {
         let imageAttachment = NSTextAttachment()
-        imageAttachment.bounds = boundsOfImage(font: font, uiImage: uiImage)
+        imageAttachment.bounds = boundsOfImage(font: font, uiImage: uiImage, scale: scale)
```
yuki-stripe: Would love a comment here like // center the image and resize it according to `scale`.

StripePaymentSheet/StripePaymentSheet/Source/Categories/NSAttributedString+Stripe.swift
gbirch-stripe: I'll add the documentation and some more comments. The intention behind also having a scale param is that the desired height of the image relative to the text varies across context. For example in the Afterpay header, we want the afterpay logo to be a lot taller than the text, and the info icon we want to be only a bit larger

yuki-stripe: Oh I see, the image does already with the font size. The `scale` you pass in is an **additional** multiplier.

gbirch-stripe:
Temporarily converting to draft to await design feedback

PR #5741, opened 2025-10-24
Fix captcha timeout

joyceqin-stripe:
## Summary

Cancel all tasks after getting first result
## Motivation

Delayed confirmation
## Testing

Manual and added a test
## Changelog

N/A


StripePayments/StripePayments/Source/Captcha/PassiveCaptcha.swift
```diff
@@ -142,6 +142,7 @@ import Foundation
                 }
                 // Wait for first completion
                 let result = try await group.next()
+                group.cancelAll()
```
yuki-stripe: I'd like this to be alongside the other `tokenTask.cancel` so that the comment can apply to both (and maybe add a ⚠️  to the comment since it's so critical and it's so easy to mess up). 

Should we put this in the defer too? Or move the tokenTask.cancel over here? I've forgotten why we have the defer again >_>

PR #5750, opened 2025-10-27
PMME initialization from API data

gbirch-stripe:
## Summary
- implement logic tree from https://trailhead.corp.stripe.com/docs/payment-method-messaging/pmme-platform/elements-mobile
- implement concurrent fetching of images

## Motivation
https://docs.google.com/document/d/13FuGJEOV0784KylxGzfSwaqYVhCv1lWcPMF5R0BlWrE/edit?pli=1&tab=t.0#heading=h.k2l70v5oo6d0

## Testing
Going to add initialization tests once initialization logic tree is 100% finalized

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+Internal.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+Internal.swift
@@ -52,7 +52,7 @@ extension PaymentMethodMessagingElement {
 
             // invalid response scenario
             guard let infoUrl = firstPaymentPlan.content.learnMore?.url else {
-                Self.logAPIError(apiClient: configuration.apiClient)
+                // TODO(gbirch) add error analytics
                 throw PaymentMethodMessagingElementError.unexpectedResponseFromStripeAPI
             }
             guard let logo = try await Self.getIconSet(
@@ -62,7 +62,7 @@ extension PaymentMethodMessagingElement {
                 // This should never happen, but if it does we log an error and attempt to fall back to a multi-partner style
                 //      (so that we can use the promotion text, which doesn't require a logo, instead of inline) without logos
                 stpAssertionFailure("No images returned by API")
-                Self.logAPIError(apiClient: configuration.apiClient)
+                // TODO(gbirch) add error analytics
                 if let topLevelPromotion = apiResponse.content.promotion?.message {
                     self.init(
                         mode: .multiPartner(logos: []),
@@ -103,7 +103,7 @@ extension PaymentMethodMessagingElement {
 
             // invalid response scenario
             guard let infoUrl = apiResponse.content.learnMore?.url else {
-                Self.logAPIError(apiClient: configuration.apiClient)
+                // TODO(gbirch) add error analytics
                 throw PaymentMethodMessagingElementError.unexpectedResponseFromStripeAPI
             }
 
@@ -173,12 +173,6 @@ extension PaymentMethodMessagingElement {
             //    from [IconSet?] to [IconSet]
             return icons.compactMap { $0 }
         }
-    }
-
-    private static func logAPIError(apiClient: STPAPIClient) {
-        let error = PaymentMethodMessagingElementError.unexpectedResponseFromStripeAPI
-        let errorAnalytic = ErrorAnalytic(event: .unexpectedPMMEError, error: error)
-        STPAnalyticsClient.sharedClient.log(analytic: errorAnalytic, apiClient: apiClient)
     }
 }
 


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+Internal.swift
```diff
@@ -2,19 +2,147 @@
...
+
+            // invalid response scenario
+            guard let infoUrl = firstPaymentPlan.content.learnMore?.url else {
+                throw PaymentMethodMessagingElementError.unexpectedResponseFromStripeAPI
+            }
+            guard let logo = try await Self.getIconSet(
+                for: firstPaymentPlan.content.images,
+                style: appearance.style
+            ).first else {
+                throw PaymentMethodMessagingElementError.unknown
```
yuki-stripe: Should probably throw a more specific error. I would avoid `unknownErrors` in general, we can be more accurate.

yuki-stripe: Also should keep an eye on how often we fail to load all images, in case we want to be more robust (maybe just display an empty icon instead).

gbirch-stripe: I added handling that logs the error and reverts back to a multi-partner style display in this case if possible (since multi-partner doesn't actually require logos).

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+Internal.swift
```diff
@@ -51,21 +52,48 @@ extension PaymentMethodMessagingElement {
...
                 throw PaymentMethodMessagingElementError.unexpectedResponseFromStripeAPI
             }
             guard let logo = try await Self.getIconSet(
                 for: firstPaymentPlan.content.images,
-                style: appearance.style
+                style: configuration.appearance.style
             ).first else {
-                throw PaymentMethodMessagingElementError.unknown
+                // This should never happen, but if it does we log an error and attempt to fall back to a multi-partner style
+                //      (so that we can use the promotion text, which doesn't require a logo, instead of inline) without logos
```
yuki-stripe, line 63: Consider aligning with Android on this behavior!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+Internal.swift
```diff
@@ -75,6 +103,7 @@ extension PaymentMethodMessagingElement {
 
             // invalid response scenario
             guard let infoUrl = apiResponse.content.learnMore?.url else {
+                Self.logAPIError(apiClient: configuration.apiClient)
                 throw PaymentMethodMessagingElementError.unexpectedResponseFromStripeAPI
```
yuki-stripe, line 107: if we run into these will we have a way to debug? Can we look at the response that was sent or something to know that a client had a problem with the info URL? If not, maybe we should send a more specific `missingOrInvalidLearnMoreURL` or whatever error.

Same applies to the other no-images error.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+Internal.swift
```diff
@@ -2,19 +2,188 @@
...
+            } else {
+                // if we also don't have a top-level promo text, then this is a no content scenario
+                return nil
+            }
+        } else {
+            // multi-partner
+
+            // invalid response scenario
+            guard let infoUrl = apiResponse.content.learnMore?.url else {
+                Self.logAPIError(apiClient: configuration.apiClient)
```
yuki-stripe: Non-blocking: Are we planning to send a general "pmme started to load" and "pmme finished loading" analytics? Feels useful. If so, it might be a bit cleaner to not additionally log errors here, and instead alert on "pmme finished loading" analytic events with particular errors (possibly looking at error_code and/or error_type).

gbirch-stripe: Yes, there will be a proper analytics PR later. This makes sense, and so I'll remove these for now and add all of the analytics at once later so they can be better coordinated

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+Internal.swift
gbirch-stripe: What's thrown here goes to users, not our analytics, so we won't be using them to debug and I don't think there's much purpose to giving something more descriptive about what part of our API response we're having issues with. I agree about the above point about analytics and will be sure to add more descriptive and debuggable error analytics when they are added

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+Internal.swift
gbirch-stripe: pinged @tjclawson-stripe https://stripe.slack.com/archives/C09C4G4RLH3/p1761772631385129?thread_ts=1761596456.428429&cid=C09C4G4RLH3

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+Internal.swift
yuki-stripe: Sounds good! One way to go about the idea in the other comment is we throw a specific error here, which then ends up getting logged in a 'pmme failed to load' event.

PR #5751, opened 2025-10-27
PMME UI

gbirch-stripe:
## Summary
https://docs.google.com/document/d/13FuGJEOV0784KylxGzfSwaqYVhCv1lWcPMF5R0BlWrE/edit?pli=1&tab=t.0

## Motivation
https://docs.google.com/document/d/13FuGJEOV0784KylxGzfSwaqYVhCv1lWcPMF5R0BlWrE/edit?pli=1&tab=t.0

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMMEUIView.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMMEUIView.swift
@@ -11,8 +11,7 @@ import SafariServices
 @_spi(STP) import StripeUICore
 import UIKit
 
-// Abstract class for PMME UIViews. Use PMMESinglePartnerView or PMMEMultiPartnerView instead of displaying directly
-class PMMEUIView: UIStackView {
+class PMMEUIView: UIView {
 
     private let infoUrl: URL
     private let appearance: PaymentMethodMessagingElement.Appearance


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMMEUIView.swift
```diff
@@ -0,0 +1,184 @@
...
+//
+//  Created by George Birch on 10/9/25.
+//
+
+import SafariServices
+@_spi(STP) import StripeCore
+@_spi(STP) import StripeUICore
+import UIKit
+
+class PMMEUIView: UIStackView {
```
yuki-stripe: Hrm is there a reason we subclass instead of containing a stackview? We should prefer composition over inheritance in general.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMMEUIView.swift
```diff
@@ -0,0 +1,184 @@
...
+        self.appearance = viewData.appearance
+        self.didUpdateHeight = didUpdateHeight
+        super.init(frame: .zero)
+
+        setupView()
+
+        addGestureRecognizer(UITapGestureRecognizer(target: self, action: #selector(didTap)))
+    }
+
+    // Would be great to have a localized accessibility hint saying you can to open the info webview.
```
yuki-stripe: is this a TODO?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMMEUIView.swift
```diff
@@ -0,0 +1,184 @@
...
+        addArrangedSubview(promotionLabel)
+    }
+
+    required init(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    override func willMove(toSuperview newSuperview: UIView?) {
+        updateLogoStyles()
+    }
```
yuki-stripe: ooh i don't think we do this for embedded but probably should to respect its superview's traitCollection. Not sure what it respects today.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMMEUIView.swift
```diff
@@ -0,0 +1,184 @@
...
+        axis = .vertical
+        spacing = Self.verticalPadding
+
+        if case .alwaysDark = appearance.style {
+            overrideUserInterfaceStyle = .dark
+        } else if case .alwaysLight = appearance.style {
+            overrideUserInterfaceStyle = .light
+        }
+
+        // Add logos if in multi-partner style
```
yuki-stripe: Just my two cents, don't need to rewrite this, but I think I would have written two separate, even simpler views for the different UIs rather than combine them into one slightly complicated view. When I read the code I have to know there are two mutually exclusive UI layouts this thing can be in which makes it a bit more difficult to understand e.g. `logoViews` does not apply to single partner and is always empty.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMMEUIView.swift
```diff
@@ -0,0 +1,184 @@
...
+        // Construct themed info url
+        let themeParam = switch (appearance.style, traitCollection.isDarkMode) {
+        case (.alwaysLight, _), (.automatic, false): "stripe"
+        case (.alwaysDark, _), (.automatic, true): "night"
+        case (.flat, _): "flat"
+        }
+
+        let queryParam = URLQueryItem(name: "theme", value: themeParam)
+        guard var urlComponents = URLComponents(url: infoUrl, resolvingAgainstBaseURL: false) else {
+            stpAssertionFailure("Unable to generate URL components")
```
yuki-stripe, line 78: Would these stpassertions be alert-worthy?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMMEUIView.swift
yuki-stripe: Realized another reason to avoid subclassing here - it implies users can use it as a UIStackView, which isn't our intention e.g. you can call `addArrangedSubview` on it.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMMEMultiPartnerView.swift
```diff
@@ -0,0 +1,120 @@
...
+    required init(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    override func willMove(toSuperview newSuperview: UIView?) {
+        updateLogoStyles()
+    }
+
+    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        updateLogoStyles()
```
yuki-stripe, line 80: I think we need to call super here per the [docs](https://developer.apple.com/documentation/uikit/uitraitenvironment/traitcollectiondidchange(_:))

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMMESinglePartnerView.swift
```diff
@@ -0,0 +1,56 @@
...
+        promotionLabel.adjustsFontForContentSizeCategory = true
+        promotionLabel.numberOfLines = 0
+        addArrangedSubview(promotionLabel)
+    }
+
+    required init(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
```
yuki-stripe, line 47: same here call super

StripePaymentSheet/StripePaymentSheet.xcodeproj/project.pbxproj
```diff
@@ -331,11 +331,16 @@
 		9DEDA3E0FFF73F9275F5F8F0 /* AutoCompleteViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = EDDECF0A74497EE887DFCB46 /* AutoCompleteViewController.swift */; };
 		9E77F1E9F801AE970F1A5BE1 /* CustomerSheetConfiguration.swift in Sources */ = {isa = PBXBuildFile; fileRef = D7768E4377A7D7495A02655D /* CustomerSheetConfiguration.swift */; };
 		9F750611C4E8EAABE9F0B460 /* CustomerSheetTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = ED7ABCE56539213DCE501C54 /* CustomerSheetTests.swift */; };
+		A024A7A92EB2B6C500542C1E /* PMMEMultiPartnerView.swift in Sources */ = {isa = PBXBuildFile; fileRef = A024A7A82EB2B6C500542C1E /* PMMEMultiPartnerView.swift */; };
```
davidme-stripe: You can remove these changes, the xcodeproj will now pull in source files automatically

PR #5756, opened 2025-10-28
CryptoOnramp SDK: KYC Refresh Endpoints

mliberatore:
## Summary
- Adds new API requests to support KYC refreshing.
- Adds corresponding tests.

## Motivation
[See the PRD here](https://docs.google.com/document/d/1ZLfTbJljlyaJS_G16X98aZQqeghp6f3flxhNqE7uqHI/edit?tab=t.c0547slx70xm) for a description of the project, as well as the endpoint documentation.

## Testing
Ran the new tests.

## Changelog
N/A

### DIFF
--- a/StripeCryptoOnramp/StripeCryptoOnramp/Source/Internal/API Bindings/KYCRefreshRequest.swift
+++ b/StripeCryptoOnramp/StripeCryptoOnramp/Source/Internal/API Bindings/KYCRefreshRequest.swift
@@ -49,4 +49,3 @@ struct KYCRefreshRequest: Encodable {
         try container.encodeIfPresent(kycInfo.address.country, forKey: .country)
     }
 }
-
--- a/StripeCryptoOnramp/StripeCryptoOnrampTests/Unit/STPAPIClient+CryptoOnrampTests.swift
+++ b/StripeCryptoOnramp/StripeCryptoOnrampTests/Unit/STPAPIClient+CryptoOnrampTests.swift
@@ -260,7 +260,10 @@ final class STPAPIClientCryptoOnrampTests: APIStubbedTestCase {
         let apiClient = stubbedAPIClient()
 
         do {
-            let response = try await apiClient.refreshKycInfo(info: Constant.validKycRefreshInfo, linkAccountInfo: Constant.validLinkAccountInfo)
+            let response = try await apiClient.refreshKycInfo(
+                info: Constant.validKycRefreshInfo,
+                linkAccountInfo: Constant.validLinkAccountInfo
+            )
             XCTAssertTrue(response.allResponseFields.isEmpty)
         } catch {
             XCTFail("Expected a success response but got an error: \(error).")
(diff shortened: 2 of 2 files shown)


StripeCryptoOnramp/StripeCryptoOnramp/Source/Internal/API Bindings/EmptyRequestWithCredentials.swift
```diff
@@ -1,19 +1,19 @@
 //
-//  CustomerRequest.swift
+//  EmptyRequestWithCredentials.swift
```
mliberatore, line 2: Rather than duplicate the struct for a different request, I’ve just made the name more generic.

StripeCryptoOnramp/StripeCryptoOnramp/Source/Internal/API Bindings/KYCRefreshInfo.swift
```diff
@@ -0,0 +1,32 @@
...
+//
+//  Created by Michael Liberatore on 10/27/25.
+//
+
+import Foundation
+
+/// KYC information common to the KYC-refresh-related APIs:
+/// - /v1/crypto/internal/kyc_data_retrieve
+/// - /v1/crypto/internal/refresh_consumer_person
+struct KYCRefreshInfo {
```
mats-stripe, line 13: I might be missing something obvious, but is there any reason we can't reuse the existing `KycInfo` struct instead of this?

StripeCryptoOnramp/StripeCryptoOnramp/Source/Internal/API Bindings/KYCRefreshRequest.swift
```diff
@@ -0,0 +1,51 @@
...
+        case state
+        case zip
+        case country
+        case dob
+    }
+
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+        try container.encode(credentials, forKey: .credentials)
+        try container.encode(kycInfo.firstName, forKey: .firstName)
```
mats-stripe, line 39: Nice! Encoding from a nested structure to a flat structure, right?

mats-stripe (approved):
LGTM!

StripeCryptoOnramp/StripeCryptoOnramp/Source/Internal/API Bindings/KYCRefreshInfo.swift
mliberatore: Yes, good question. The meaningful difference is that these APIs expect/return the last 4 digits of the id number, whereas the request that expects `KycInfo` expects the full SSN. It's a minor difference, and we could essentially overload the property to mean both things, but I thought the minor duplication to name things more clearly was the better option. LMK if you disagree!

StripeCryptoOnramp/StripeCryptoOnramp/Source/Internal/API Bindings/KYCRefreshRequest.swift
mliberatore: Yup. The _retrieve_ API returns the response in a different format from the _refresh_ request body (i.e. what's nested, what's not, zip vs. postal_code ([see this thread for more info](https://lickability.slack.com/archives/C094P3BRBL1/p1761668312557619?thread_ts=1761597746.511609&cid=C094P3BRBL1))), requiring the explicit encoding and decoding methods, to not further duplicate similar models.

PR #5759, opened 2025-10-29
Generic function to run async operations with timeout

joyceqin-stripe:
## Summary

Implemented generic async with timeout functions.
https://docs.google.com/document/d/1DlK64SmYPhc4OW9bw2_dwSugi_sOURJEZFlq3K_7lVg/edit?usp=sharing
## Motivation

Support passive captcha and attestation with timeout

## Testing

Added tests, updated passive captcha
## Changelog

N/A


StripeCore/StripeCore.xcodeproj/project.pbxproj
```diff
@@ -99,6 +99,8 @@
 		9DCBC08C182ED76A962961E7 /* NSURLComponents+Stripe.swift in Sources */ = {isa = PBXBuildFile; fileRef = D475CD59778FAC1028670AB5 /* NSURLComponents+Stripe.swift */; };
 		9FBA50345D53E82AA974F672 /* STPAPIClient+FileUpload.swift in Sources */ = {isa = PBXBuildFile; fileRef = F4FC2FC535DAF9B340F9EA75 /* STPAPIClient+FileUpload.swift */; };
 		A03F7C3B2E79CB2B006F0FDC /* AsyncTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = A03F7C3A2E79CB28006F0FDC /* AsyncTests.swift */; };
+		A3DA747D2EB16F6200FFFCC9 /* Timeout.swift in Sources */ = {isa = PBXBuildFile; fileRef = A3DA747C2EB16F6200FFFCC9 /* Timeout.swift */; };
```
davidme-stripe: You can remove these changes, the xcodeproj will now pull in source files automatically

gbirch-stripe:
🔥 this is awesome @joyceqin-stripe !

gbirch-stripe:
Is there a need to increase the priority of the timeout task? i.e. is it possible that as a result of it having default priority there are situations in which it isn't actually initiated immediately and thus the timeout time isn't accurate?

StripeCore/StripeCore/Source/Helpers/Timeout.swift
```diff
@@ -0,0 +1,68 @@
...
+// MARK: - Timeout with TaskGroup and CheckedContinuation escape hatch
+/// Runs multiple operations in parallel with individual timeouts
+/// - Parameters:
+///   - timeout: The maximum time interval to wait for each operation to complete
+///   - operations: Variadic operations to run with timeout
+/// - Returns: Tuple of Results, where each Result contains either the task's value or its error
+ @_spi(STP) public func withTimeout<each T>(
+    _ timeout: TimeInterval,
+    _ operations: repeat @escaping () async throws -> each T
+ ) async -> (repeat Result<(each T)?, Error>) {
```
yuki-stripe: why's `(each T)?` optional? I'd think it'd be non-optional and T itself can be an optional if necessary.

StripeCore/StripeCore/Source/Helpers/Timeout.swift
```diff
@@ -0,0 +1,68 @@
...
+
+/// Runs a singular operation with a timeout
+/// - Parameters:
+///   - timeout: The maximum time interval to wait for the operation to complete
+///   - operation: The operation to run with a timeout
+/// - Returns: A Task whose result contains either the operation's value or its error
+private func withTimeout<T>(
+    _ timeout: TimeInterval,
+    _ operation: @escaping () async throws -> T
+) async throws -> T? {
```
yuki-stripe: Same here, why is this `T?` and not `T`?

StripeCore/StripeCore/Source/Helpers/Timeout.swift
```diff
@@ -0,0 +1,68 @@
...
+    let timeoutNs = UInt64(timeout) * 1_000_000_000
+
+    return try await withCheckedThrowingContinuation { (continuation: CheckedContinuation<T?, Error>) in
```
yuki-stripe: `withChecked..` is meant for bridging async <> completion based code which is not at all what we're using it for so it'd be great to explain why we're using it - would add comments here/below where we use `continuation` to help future readers.

StripeCore/StripeCore/Source/Helpers/Timeout.swift
```diff
@@ -0,0 +1,68 @@
...
+import Foundation
+
+@_spi(STP) public struct TimeoutError: Error {}
+
+// MARK: - Timeout with TaskGroup and CheckedContinuation escape hatch
+/// Runs multiple operations in parallel with individual timeouts
+/// - Parameters:
+///   - timeout: The maximum time interval to wait for each operation to complete
+///   - operations: Variadic operations to run with timeout
+/// - Returns: Tuple of Results, where each Result contains either the task's value or its error
```
yuki-stripe: really important to call out that this returns a tuple of results _in the same order as the operations_.

StripeCore/StripeCoreTests/Helpers/TimeoutTests.swift
```diff
@@ -0,0 +1,202 @@
+//
+//  TimeoutTests.swift
```
yuki-stripe, line 2: great tests!

StripeCore/StripeCore/Source/Helpers/Timeout.swift
joyceqin-stripe: It's because `group.next()` returns an optional— if the group is empty, it returns nil. In our case here, we expect it to never be nil. I can add a `guard`, make `TimeoutError` an enum instead with a `timeout` case and an `unexpected` case, unwrap the result, and throw the `unexpected` case if result couldn't be unwrapped.

yuki-stripe: Ah hrm. Throwing in that case rather than returning nil makes sense! 

I don't think this is semantically a case of `TimeoutError` though, should be its own error.

joyceqin-stripe:
> Is there a need to increase the priority of the timeout task? i.e. is it possible that as a result of it having default priority there are situations in which it isn't actually initiated immediately and thus the timeout time isn't accurate?

A good thought, but when I tested it, the priority on the tasks was `.high`, which I believe is already the highest priority and seems to be the default

EDIT: actually, it was `.high` because I was testing it on something that was calling from `@MainActor`— added Task priority

yuki-stripe (approved):
🔥

PR #5762, opened 2025-10-29
Attestation on confirmation

joyceqin-stripe:
## Summary

Add an AttestationChallenge wrapper around StripeAttest that prewarms attestation on init and fetches an assertion with a timeout.

Wrap PassiveCaptchaChallenge and AttestationChallenge in ConfirmationChallenge. Anywhere we previously passed PassiveCaptchaChallenge, we now pass in ConfirmationChallenge. ConfirmationChallenge creates the passive captcha and attestation challenges asynchronously in parallel and fetches their tokens with timeout in parallel and creates the radar options for the requests.

Added playground switch so that we can test with/without attestation, similar to the passive captcha switch. Will be removed on release.

Moved PassiveCaptchaChallenge into StripePaymentSheet, @_spi(STP) HCaptcha
## Motivation

Attestation support for HCaptcha project
## Testing

Added tests
Example request: https://admin.corp.stripe.com/request-log/req_YXMjaq42sec9gb
## Changelog

N/A


StripePaymentSheet/StripePaymentSheet.xcodeproj/project.pbxproj
```diff
@@ -345,6 +345,10 @@
 		A3B2F2B82CF1343F00C0E88C /* SavedPaymentMethodFormFactory+USBankAccount.swift in Sources */ = {isa = PBXBuildFile; fileRef = A3B2F2B72CF1343F00C0E88C /* SavedPaymentMethodFormFactory+USBankAccount.swift */; };
 		A3B2F2BA2CF134B500C0E88C /* SavedPaymentMethodFormFactory+SEPADebit.swift in Sources */ = {isa = PBXBuildFile; fileRef = A3B2F2B92CF134B500C0E88C /* SavedPaymentMethodFormFactory+SEPADebit.swift */; };
 		A3BE1C242D4161DC0061632C /* RemoveButton.swift in Sources */ = {isa = PBXBuildFile; fileRef = A3BE1C232D4161DC0061632C /* RemoveButton.swift */; };
+		A3DA75572EB20ECE00FFFCC9 /* AttestationChallenge.swift in Sources */ = {isa = PBXBuildFile; fileRef = A3DA75562EB20ECE00FFFCC9 /* AttestationChallenge.swift */; };
```
davidme-stripe: You can remove these changes, the xcodeproj will now pull in source files automatically

StripeCore/StripeCore/Source/Attestation/StripeAttest.swift
```diff
@@ -371,8 +380,18 @@ import UIKit
         if let keyId = storedKeyID {
             return keyId
         }
+        // Check if another task is already generating a key
```
joyceqin-stripe, line 383: Found a race condition— if you initialize AttestationChallenge and immediately try to fetch the assertion, both the `prepareAttestation()` and `assert()` functions will read a nil stored key and try and generate a key, overwriting each other and leading to a state mismatch later (client thinks we've attested because we did for one key, but the other key fetches the challenge from the server and gets that it needs to be attested)

StripeCore/StripeCoreTestUtils/Mocks/MockAppAttestService.swift
```diff
@@ -42,6 +44,14 @@ import UIKit
         attestationUsingDevelopmentEnvironment = value
     }
 
+    func setAttestationDelay(_ delay: TimeInterval) async {
```
joyceqin-stripe, line 47: Test-only to inject delays to force timeout

StripeCore/StripeCore/Source/Attestation/StripeAttest.swift
davidme-stripe: Yes, good catch!

PR #5764, opened 2025-10-29
Migrate xcodeproj files from groups to folders

davidme-stripe:
## Summary
Migrate the .xcodeproj file to use filesystem folders instead of groups (the new default). This makes git merges easier and allows Claude Code to create/delete/move files without manual intervention.

* Add tests that were inadvertently left out of the xcodeproj and not building, delete non-functional ones that weren't running.
* Update periphery to the latest version, as the 2.x series doesn't support the new `fileSystemSynchronizedGroups` type.
* Fix a failing Apple Pay test that was depending on the order in which the expectations were resolved

## Motivation
Modernizing our .xcodeprojs

## Testing
CI

## Changelog
N/A, not user facing

### DIFF
--- a/.periphery.yml
+++ b/.periphery.yml
@@ -6,19 +6,7 @@
 project: Stripe.xcworkspace
 
 schemes:
-- Stripe3DS2
-- StripeApplePay
-- StripeCameraCore
-- StripeCardScan
-- StripeConnect
-- StripeCore
-- StripeFinancialConnections
-- StripeIdentity
-- StripePaymentSheet
-- StripePayments
-- StripePaymentsUI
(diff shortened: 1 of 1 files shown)


davidme-stripe:
It seems like the "modifies public API" job is ordering-dependent and has changed with the new folder structure, so I'll mark this as "modifies public API" for now.

davidme-stripe:
This is ready to go: Size Analysis is broken from an earlier Bitrise incident, and "Verify public interface" is broken because we reordered all the source files so the change list is too large (but should fix itself once this is merged into master)

yuki-stripe (approved):

PR #5765, opened 2025-10-29
Fix ordering in API checker

davidme-stripe:
## Summary
The API checker has false positives if the ordering of the API functions in the `.swiftinterface` file changes, and Xcode seems to be compiling these files in a non-deterministic order now that we're using folder references. To work around this, try sorting them all before diffing. Parsing this file is messy and difficult, so I've outsourced it to Claude Code for now.

The code for itself is wild (it's using a series of regexes and special-cases), but it seems good enough for our intended use case of sorting two swiftinterface files for diffing.

## Motivation
Fix API checker errors

## Testing
CI

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Analytics/STPAnalyticsClient+LUXE.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Analytics/STPAnalyticsClient+LUXE.swift
@@ -2,7 +2,7 @@
 //  STPAnalyticsClient+LUXE.swift
 //  StripePaymentSheet
 //
-//  Copyright © 2022 Stripe, Inc. All rights reserved.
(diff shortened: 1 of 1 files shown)


StripePaymentSheet/StripePaymentSheet/Source/Analytics/STPAnalyticsClient+LUXE.swift
```diff
@@ -2,7 +2,7 @@
 //  STPAnalyticsClient+LUXE.swift
 //  StripePaymentSheet
 //
-//  Copyright © 2022 Stripe, Inc. All rights reserved.
+//  Copyright © 2025 Stripe, Inc. All rights reserved.
```
davidme-stripe, line 5: Updated this file to trigger a run of the interface checker

PR #5766, opened 2025-10-30
Sort swiftinterface files when diffing

davidme-stripe:
## Summary
The API checker has false positives if the ordering of the API functions in the `.swiftinterface` file changes, and Xcode seems to be compiling these files in a non-deterministic order now that we're using folder references. To work around this, try sorting them all before diffing. This makes the files semantically meaningless, but it *does* catch changes!

## Motivation
Fixing the API checker

## Testing
CI

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
@@ -3,7 +3,7 @@
 //  StripePaymentSheet
 //
 //  Created by Yuki Tokuhiro on 9/3/20.
-//  Copyright © 2020 Stripe, Inc. All rights reserved.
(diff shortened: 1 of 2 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
```diff
@@ -3,7 +3,7 @@
 //  StripePaymentSheet
 //
 //  Created by Yuki Tokuhiro on 9/3/20.
-//  Copyright © 2020 Stripe, Inc. All rights reserved.
+//  Copyright © 2025 Stripe, Inc. All rights reserved.
```
davidme-stripe, line 6: To trigger an API check build

yuki-stripe (approved):

PR #5768, opened 2025-10-30
[Identity] Fix layout issue in Identity flow

Pearapps:
Related to, but does not fix (because that issue specifically is fixed) https://github.com/stripe/stripe-identity-react-native/issues/221

---

This makes sure the views right about the flow buttons do not extend outside of the width of the view.

---

BEFORE:

Screenshot 2025-10-30 at 10 30 30 AM

After:

Screenshot 2025-10-30 at 10 58 13 AM

### DIFF
--- a/StripeIdentity/StripeIdentity/Source/NativeComponents/Views/IdentityFlowView.swift
+++ b/StripeIdentity/StripeIdentity/Source/NativeComponents/Views/IdentityFlowView.swift
@@ -311,7 +311,7 @@ extension IdentityFlowView {
 
         // Remove old buttons and create new ones and add them to the stack view
         buttons.forEach { $0.removeFromSuperview() }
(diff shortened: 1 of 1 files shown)


StripeIdentity/StripeIdentity/Source/NativeComponents/Views/IdentityFlowView.swift
```diff
@@ -307,7 +311,7 @@ extension IdentityFlowView {
 
         // Remove old buttons and create new ones and add them to the stack view
         buttons.forEach { $0.removeFromSuperview() }
-        buttons = buttonViewModels.enumerated().map { index, _ in
+        buttons = buttonViewModels.indices.map { index in
```
Pearapps, line 314: I was forced to change this due to the linter

luisv-stripe (approved):
LGTM

PR #5769, opened 2025-10-30
PMME - SwiftUI and public API

gbirch-stripe:
## Summary
- SwiftUI wrapper layer
- public API

## Motivation
https://docs.google.com/document/d/13FuGJEOV0784KylxGzfSwaqYVhCv1lWcPMF5R0BlWrE/edit?pli=1&tab=t.0

## Testing
still determining best way to test the swiftui wrapper. will add tests later

## Changelog
N/A (spi protected for now)

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PaymentMethodMessagingElement.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PaymentMethodMessagingElement.swift
@@ -12,7 +12,7 @@ import SwiftUI
 import UIKit
 
 /// An element that provides a view with information about how a purchase could be paid for using Buy Now, Pay Later payment methods.
-@_spi(PMME)
+@_spi(STP)
 public class PaymentMethodMessagingElement {
 
     /// A UIKit view of the element.
--- a/StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PMMESnapshotTests.swift
+++ b/StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PMMESnapshotTests.swift
@@ -7,7 +7,7 @@
 
 @_spi(STP)@testable import StripeCore
 import StripeCoreTestUtils
-@_spi(STP) @_spi(PMME) @testable import StripePaymentSheet
+@_spi(STP)@testable import StripePaymentSheet
 @_spi(STP)@testable import StripeUICore
 import UIKit
 


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PaymentMethodMessagingElement.swift
```diff
@@ -5,13 +5,41 @@
...
+        /// - Parameter Error: An `Error` object representing the reason the element failed to load
+        case failed(Error)
+    }
+
     /// Describes the visual appearance of the PaymentMethodMessagingElement.
-    public struct Appearance {
+    public struct Appearance: Equatable {
+
+        /// The default appearance for Payment Method Messaging Element.
+        public static let `default` = PaymentMethodMessagingElement.Appearance()
```
yuki-stripe: This wasn't in the API Review! We should get it reviewed, but why are we adding it?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PaymentMethodMessagingElement+SwiftUI.swift
```diff
@@ -2,15 +2,84 @@
...
+        /// Initializes a PaymentMethodMessagingElement SwiftUI View.
+        /// - Parameter config: Configuration for the PaymentMethodMessagingElement, such as the amount and currency of the purchase.
+        /// - Parameter content: Content to be displayed depending on the element's current phase.
+        public init(configuration: Configuration, @ViewBuilder content: @escaping (Phase) -> Content) {
+            self.config = configuration
+            self.content = content
+        }
+
+        /// Initializes a PaymentMethodMessagingElement SwiftUI View.
+        /// Use this initializer when you want manual control over the elment creation or to initialize it outside of the UI layer.
```
yuki-stripe:
```suggestion
        /// Use this initializer when you want manual control over the element creation or to initialize it outside of the UI layer.
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+SwiftUI+Internal.swift
```diff
@@ -0,0 +1,104 @@
...
+            ZStack {
+                Color.clear.frame(width: 0, height: 0)
+                content(phase)
+            }
+            // Anytime the config changes we reload.
+            // .task automatically manages the lifetime of the task to match the view.
+            // In the config (no content) integration style, the content will be an clear size 0 view to start, this
+            //      is necessary since an EmptyView will not enter the view hierarchy and thus .task has nothing to run on.
+                .task(id: config) {
+                    await load()
```
yuki-stripe: some qs
1. does this mean no retain cycle is possible here?
2. How often/when is this task called?  why not load in the init?

Also nit: funky spacing

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+SwiftUI+Internal.swift
```diff
@@ -0,0 +1,104 @@
...
+extension PaymentMethodMessagingElement.View {
+
+    @ViewBuilder func bodyImpl() -> some View {
+        if let config {
+            // If we were initialized with a config, load the config
+            ZStack {
+                Color.clear.frame(width: 0, height: 0)
+                content(phase)
+            }
+            // Anytime the config changes we reload.
```
yuki-stripe: What's this mean, the config is a `let`?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+SwiftUI+Internal.swift
```diff
@@ -0,0 +1,104 @@
...
+}
+
+@available(iOS 15.0, *)
+extension PaymentMethodMessagingElement.View {
+
+    @ViewBuilder func bodyImpl() -> some View {
+        if let config {
+            // If we were initialized with a config, load the config
+            ZStack {
+                Color.clear.frame(width: 0, height: 0)
```
yuki-stripe: ooc does it not work to just like
```
@ViewBuilder func bodyImpl() -> some View {
  ZStack {
    Color.clear.frame(width: 0, height: 0)
    content(phase)
  }.task(id: config) { .. }
}

gbirch-stripe: `task(id:)` needs a non-optional, so we have to bind `config`

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+SwiftUI+Internal.swift
gbirch-stripe: SwiftUI manages the lifecycle of the View, but `.task(id:)` lets us tie the task to a particular config. So SwiftUI may arbitrarily re-initialize but with the same config, and this way we'll still be using the same task instead of restarting it!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+SwiftUI+Internal.swift
gbirch-stripe: This task is called if and only if config changes (according to its `Equatable` conformance). SwiftUI may or may not call this View's initializer whenever it wants, but this guarantees that for each `PMME.View` in the user's view hierarchy, anytime the config data passed into it changes the task is launched again, but anytime the config data doesn't change it doesn't

gbirch-stripe: Yes we don't need to worry about retain cycles here! If the View is gone SwiftUI will clean up the task too!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PaymentMethodMessagingElement.swift
gbirch-stripe: Consistency with PS/Embedded. Users may be used to calling `x.Appearance.default` and I don't see a reason to make them do it the other way for PMME. Can remove for now if you'd prefer to have this go through the review process

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+SwiftUI+Internal.swift
gbirch-stripe: even though config is a `let` and technically doesn't change within a particular instance of the View, `.task {` is _not_ called again by SwiftUI when the config changes (and thus somewhere under the hood a new instance is created). That's why we need `.task(id:)` to tell SwiftUI that it actually _does_ need to run again

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PaymentMethodMessagingElement.swift
yuki-stripe: Yeah I think `default` was [accidentally](https://github.com/stripe-ios/stripe-ios-private/pull/886/files#diff-889b39a430619515606d9a00faf484fb2cb711be9dc0024810f9f15d2b562ee9R16) made public. I remember regretting when we were thinking about Liquid Glass APIs (can't remember the specifics right now) so I'd prefer not to add.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+SwiftUI+Internal.swift
yuki-stripe: I tried it out and it appears to work, it just needs an `Equatable`.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+SwiftUI+Internal.swift
yuki-stripe: Ah got it, neat.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+SwiftUI+Internal.swift
yuki-stripe: Mm ok. I can see how this is more idiomatic, but since `config` is immutable here we _could_ call this init for the ~same effect?

yuki-stripe: > Retaining a task object doesn’t indefinitely retain the closure, because any references that a task holds are released after the task completes. Consequently, tasks rarely need to capture weak references to values. [[src](https://developer.apple.com/documentation/swift/task)]

found this as well

gbirch-stripe: > Mm ok. I can see how this is more idiomatic, but since config is immutable here we could call this init for the ~same effect?

No we can't. Or rather it would be extremely difficult to do so. For the "same view" (meaning an instance of the view in the view hierarchy), SwiftUI may arbitrarily change the backing view struct instance. For example, if another param changes (such as content in our particular case), the initializer MUST be called again and a new view struct instance will be used for the same view. We would therefore have to find some way to NOT cancel the task and instead associate it and its result with the new struct instance. I don't even know how one would begin to manually implement this behavior.

Another major problem with calling the load in the init is that SwiftUI can call the initializer before the view is actually displayed. For example if a view is defined in a navigation link, its initializer is run right away even though it hasn't been displayed / entered the view hierarchy and may never do so.

In short, SwiftUI may or may not call the actual view initializer anytime it wants, and this causes myriad problems with putting behavior like triggering async work there, which is why it's an incorrect pattern. The correct pattern for triggering synchronous work for when a view appears is `.onAppear`, for triggering asynchronous work for when the view appears is `.task`, and for triggering asynchronous work that should be retriggered in response to data changes is `.task(id:)`

`.task(id:)` for free provides
1. starting when the view is displayed
2. cancelling when the view disappears
3. persistence across the life cycle of the view
4. not being restarted if non-dependent data changes
5. being restarted if dependent data changes

These things are very difficult to achieve by manually implementing them, and apple recommends you use `.task(id:)`

yuki-stripe (approved):

yuki-stripe (approved):

PR #5209, opened 2025-07-24
Rename STPPaymentIntentParams to STPPaymentIntentConfirmParams

wooj-stripe:

PR #5787, opened 2025-11-03
CryptoOnramp SDK: Adds KYC Refresh API + UI

mliberatore:
## Summary
Adds the new KYC Refresh API to `CryptoOnrampCoordinator`, with supporting UI in `StripePaymentSheet` to reduce the amount of `public` exposures needed in `StripePaymentSheet`.

There are several remaining items not included here, which will be added in follow-up PRs:
- String localizations
- Dismissing `BottomSheetViewController` logs analytics, and likely shouldn’t in our use case: `LOG ANALYTICS: mc_dismiss - [(key: "pay_var", value: "paymentsheet"), (key: "ocr_type", value: "none"), (key: "apple_pay_enabled", value: 1)]`
- Adding analytics to the crypto onramp implementation of the new API.
- Handling the new API’s result properly in the demo app, specifically presenting a new UI to re-collect the address to reinvoke the API to confirm the changes.

| Light | Dark |
| --- | --- |
|  |  |

## Motivation

- [Product requirements](https://docs.google.com/document/d/1ZLfTbJljlyaJS_G16X98aZQqeghp6f3flxhNqE7uqHI/edit?tab=t.c0547slx70xm#heading=h.xg0apl9hsn12)
- [Mobile API Review draft](https://docs.google.com/document/d/15pjTJiAn5pcajm0Z820A4yLt517q0l5wOVxu7YnLkaM/edit?tab=t.0#heading=h.570u4leviueu)
    - Contains link to the Figma

## Testing
1. Logged in with a user in the demo app.
2. After authenticating, used the new Verify KYC Info menu item.
3. Separately chose the 3 options to dismiss the UI (x button, pencil icon, and confirm button) ensuring that the proper status was reached in the demo UI code / print statements, and that the API requests succeeded.

https://github.com/user-attachments/assets/e7e330d3-8459-4db5-aad3-096361220736

## Changelog
N/A


Stripe.xcworkspace/contents.xcworkspacedata
```diff
@@ -25,6 +25,9 @@
       <FileRef
          location = "group:StripeConnectExample/StripeConnect Example.xcodeproj">
       </FileRef>
+      <FileRef
+         location = "group:CryptoOnramp Example/CryptoOnramp Example.xcodeproj">
+      </FileRef>
```
mliberatore, line 30: I’m not sure why but this got removed in #5209

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetSnapshotTests.swift
```diff
@@ -13,7 +13,7 @@ import UIKit
 import XCTest
 
 @_spi(STP)@testable import StripeCore
-@_spi(AppearanceAPIAdditionsPreview) @testable import StripePaymentSheet
+@_spi(AppearanceAPIAdditionsPreview) @_spi(STP) @testable import StripePaymentSheet
```
mliberatore, line 16: Unexpectedly, presumably due to the `BottomSheetViewController` `public` exposure, the following was generating a compiler error, which this fixes.

```
…/PaymentSheetSnapshotTests.swift:1156:57 'contentStack' is inaccessible due to 'internal' protection level
```

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Verify KYC/VerifyKYCContentViewController.swift
```diff
@@ -0,0 +1,213 @@
...
+        stackView.translatesAutoresizingMaskIntoConstraints = false
+        return stackView
+    }()
+
+    private lazy var headingLabel: UILabel = {
+        let label = UILabel()
+        label.translatesAutoresizingMaskIntoConstraints = false
+        label.font = LinkUI.font(forTextStyle: .title)
+        label.textColor = .linkTextPrimary
+        label.text = "Confirm your information"
```
mliberatore, line 82: Note: The nonlocalized strings in this PR will be taken care of in a separate PR.

Example/CryptoOnramp Example/CryptoOnramp Example/UI Helpers/AuthenticatedUserToolbarItemModifier.swift
```diff
@@ -79,4 +88,25 @@ private struct AuthenticatedUserToolbarItemModifier: ViewModifier {
...
+        Task {
+            do {
+                let result = try await coordinator.verifyKYCInfo(from: presentingVC)
+                switch result {
+                case .confirmed:
+                    print("KYC verification confirmed")
+                case .updateAddress:
+                    print("KYC verification: user chose to update address")
+                case .canceled:
+                    print("KYC verification canceled")
```
mliberatore, line 103: I’ll make actual use of the result in the example app in a separate PR.

Stripe.xcworkspace/contents.xcworkspacedata
mats-stripe: Thanks for asking there!

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
```diff
@@ -88,6 +88,18 @@ protocol CryptoOnrampCoordinatorProtocol {
     /// Throws if an authenticated Link user is not available, or an API error occurs.
     func attachKYCInfo(info: KycInfo) async throws
 
+    /// Initiates the KYC verification flow, which displays the user’s currently collected KYC information with the ability to confirm or update the displayed address.
+    ///
+    /// - Parameters:
+    ///   - updatedAddress: An optional updated address. Specify this parameter if the user has elected to change the address after a prior call to this API returned `VerifyKYCResult.updateAddress`. Otherwise, specify `nil` to show the user’s existing KYC information on the presented flow.
```
mats-stripe, line 94: When this parameter is provided will the UI be shown again? Or do / should we just perform an API request and return a `.confirmed` result

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
```diff
@@ -379,6 +391,61 @@ public final class CryptoOnrampCoordinator: NSObject, CryptoOnrampCoordinatorPro
         }
     }
 
+    public func verifyKYCInfo(updatedAddress: Address? = nil, from viewController: UIViewController) async throws -> VerifyKYCResult {
+        do {
+            let linkAccountInfo = try await self.linkAccountInfo
+            let apiClient = self.apiClient
```
mats-stripe: Just to confirm my understanding; this is captured here so it can be used in the `Task` closure?

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
```diff
@@ -379,6 +391,61 @@ public final class CryptoOnrampCoordinator: NSObject, CryptoOnrampCoordinatorPro
...
+
+                        switch result {
+                        case .canceled, .updateAddress:
+                            dismissAndResumeWithResult(.success(result))
+                        case .confirmed:
+                            Task {
+                                do {
+                                    // When confirming, we make the API call for confirmation before returning success.
+                                    // If the API call fails, the error will be caught and returned to the caller.
+                                    try await apiClient.refreshKycInfo(info: displayInfo, linkAccountInfo: linkAccountInfo)
```
mats-stripe, line 430: If this line `throw`s, will the sheet still get dismissed?

StripePaymentSheet/StripePaymentSheet/Resources/StripePaymentSheet.xcassets/KYC Refresh/icon_edit_outline.imageset/kyc_verify_edit.svg
mats-stripe, line 1: Minor: could we name this asset something more generic like `edit` or `pencil`? This will make it easier to reuse the asset in other situations if we ever need to!

StripePaymentSheet/StripePaymentSheet/Resources/StripePaymentSheet.xcassets/KYC Refresh/Contents.json
mats-stripe, line 1: Is this a folder?

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Verify KYC/VerifyKYCContentViewController.swift
```diff
@@ -0,0 +1,213 @@
+//
+//  VerifyKYCContentViewController.swift
```
mats-stripe, line 2: Wondering if you have thoughts on putting this under the `Link` [namespace](https://github.com/stripe/stripe-ios/tree/master/StripePaymentSheet/StripePaymentSheet/Source/Internal/Link) or keeping it separate?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Verify KYC/VerifyKYCContentViewController.swift
```diff
@@ -0,0 +1,213 @@
...
+
+@_spi(STP) import StripeUICore
+import UIKit
+
+/// The content view of `VerifyKYCViewController`, which displays a list of KYC fields with the optional ability to initiate editing of the address.
+final class VerifyKYCContentViewController: UIViewController, BottomSheetContentViewController {
+    private enum Constants {
+        static let layoutMargins = NSDirectionalEdgeInsets.insets(leading: 24, trailing: 24)
+        static let ctaLayoutMargins = NSDirectionalEdgeInsets.insets(amount: 24)
+        static let infoContainerCornerRadius: CGFloat = 12
```
mats-stripe: Wonder if we should use values from `LinkUI` here? No strong opinion either way

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Verify KYC/VerifyKYCContentViewController.swift
```diff
@@ -0,0 +1,213 @@
...
+        } else {
+            // fall back to strict manual formatting in the event of an unexpected error with `DateComponents`.
+            let mm = String(format: "%02d", info.dateOfBirthMonth)
+            let dd = String(format: "%02d", info.dateOfBirthDay)
+            let yyyy = String(format: "%04d", info.dateOfBirthYear)
+            return "\(mm)/\(dd)/\(yyyy)"
+        }
+    }
+
+    private var formattedAddress: String {
```
mats-stripe, line 121: Apparently there's a built in Swift address formatter: https://www.swiftbysundell.com/articles/exploring-some-of-the-lesser-known-formatter-types/#addresses though I'm not totally sure if the output matches to what we want

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Verify KYC/VerifyKYCContentViewController.swift
```diff
@@ -0,0 +1,213 @@
...
+        container.backgroundColor = .linkSurfaceSecondary
+        container.layer.cornerRadius = Constants.infoContainerCornerRadius
+        container.layer.masksToBounds = true
+
+        let stackView = UIStackView(arrangedSubviews: [
+            VerifyKYCInfoRowView(title: "Name", value: formattedName),
+            makeDivider(),
+            VerifyKYCInfoRowView(title: "Date of Birth", value: formattedDateOfBirth),
+            makeDivider(),
+            VerifyKYCInfoRowView(title: "Last 4 digits of SSN", value: info.idNumberLast4 ?? ""),
```
mats-stripe: Minor: Should we hide this row altogether if we don't have a `info.idNumberLast4`?

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Verify KYC/VerifyKYCContentViewController.swift
```diff
@@ -0,0 +1,213 @@
...
+    }
+
+    private func confirmButtonTapped() {
+       confirmButton.update(state: .spinnerWithInteractionDisabled)
+       onResult?(.confirmed)
+    }
+
+    // MARK: - BottomSheetContentViewController
+
+    func didTapOrSwipeToDismiss() {
```
mats-stripe, line 206: I wonder if we should disallow swipe to dismiss on this screen

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Verify KYC/VerifyKYCInfoRowView.swift
```diff
@@ -0,0 +1,103 @@
...
+//
+
+import UIKit
+
+@_spi(STP) import StripeUICore
+
+/// Displays an individual KYC detail (e.g. name) with an optional edit button.
+final class VerifyKYCInfoRowView: UIView {
+    private enum Constants {
+        static let editButtonSize = CGSize(width: 44, height: 44)
```
mats-stripe: Not ideal that the property is called `height`, but we could do:

```suggestion
        static let editButtonSize = CGSize(width: LinkUI.minimumButtonHeight, height: LinkUI.minimumButtonHeight)
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Verify KYC/VerifyKYCViewController.swift
```diff
@@ -0,0 +1,77 @@
...
+//
+//  Created by Michael Liberatore on 10/28/25.
+//
+
+import UIKit
+
+/// Container view controller that displays a list of KYC fields with the optional ability to initiate editing of the address.
+@_spi(STP) public final class VerifyKYCViewController: BottomSheetViewController {
+    private enum Constants {
+        static let sheetCornerRadius: CGFloat = 24
```
mats-stripe: Another instance where I think we should use values from `LinkUI`! 
```suggestion
        static let sheetCornerRadius: CGFloat = LinkUI.largeCornerRadius
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/BottomSheetViewController.swift
```diff
@@ -24,12 +24,12 @@ protocol BottomSheetContentViewController: UIViewController {
 /// A VC containing a content view controller and manages the layout of its SheetNavigationBar.
 /// For internal SDK use only
 @objc(STP_Internal_BottomSheetViewController)
-class BottomSheetViewController: UIViewController, BottomSheetPresentable {
+@_spi(STP) public class BottomSheetViewController: UIViewController, BottomSheetPresentable {
```
mats-stripe, line 27: Sorry, this might be obvious, but I don't fully grok why this needs to be `@_spi(STP) public` if `VerifyKYCViewController` is in the same module?

mliberatore: Because `VerifyKYCViewController` is declared `@_spi(STP) public` for use in crypto onramp, keeping this internal generates the following error:

```
…/VerifyKYCViewController.swift:11:31 Class cannot be declared public because its superclass is internal
```

The only way I can think of to prevent needing to make `BottomSheetViewController` `public` as well is if `VerifyKYCViewController` were used only in e.g. `LinkController`, and the crypto onramp coordinator called a `LinkController` API that masks which view controller is presented. It'd require some rework to the logic, though, especially since crypto onramp is responsible for making the crypto backend API requests / that might lead to a closure-heavy API.

Let me know your thoughts!

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
mliberatore: If the parameter is provided in a second call to this API yes, the UI will be shown again. They specifically want the user to re-confirm using Stripe UI after inputting their updated address in their own UI. The following screenshot highlights this in the PRD. Note also in this flow, there's a second call to post the updated address, we’re only posting it once, and will be communicating that to receive any further clarification: 
CleanShot 2025-11-04 at 09 44 29

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
mliberatore: Yes. The reason it's done this way is to make the caller responsible for presenting any errors to the user. Otherwise, they'd have to present errors above Stripe UI, which could get messy with regard to view controller inspection. We [asked about this in Slack](https://lickability.slack.com/docs/T025389C5/F09MSUUUXSB?focus_section_id=temp:C:QfJ3cf529d5ec664d1ba4ea1e93b), clarified [here](https://lickability.slack.com/archives/C09MSUUUXSB/p1761586013203989?thread_ts=1761264326.605939&cid=C09MSUUUXSB) and received an answer [here](https://lickability.slack.com/archives/C094P3BRBL1/p1761332506581499?thread_ts=1761264918.887209&cid=C094P3BRBL1) and [here](https://lickability.slack.com/archives/C09MSUUUXSB/p1761588682470759?thread_ts=1761264326.605939&cid=C09MSUUUXSB).


It will look like the following, which is easily tested by changing to a NY address. The client would be responsible for not losing the updated address.

https://github.com/user-attachments/assets/016b0466-c5a7-4074-9c92-7bf1f458dae2

StripePaymentSheet/StripePaymentSheet/Resources/StripePaymentSheet.xcassets/KYC Refresh/icon_edit_outline.imageset/kyc_verify_edit.svg
mliberatore: Sure. It turns out there's another pencil icon in the asset catalog called `icon_edit`, so I’m updating this new one to be `icon_edit_outline`.

StripePaymentSheet/StripePaymentSheet/Resources/StripePaymentSheet.xcassets/KYC Refresh/Contents.json
mliberatore: Yes, within the asset catalog:
CleanShot 2025-11-04 at 10 15 34

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
mats-stripe: Oh gotcha - I had just misread the code and thought the `dismissAndResumeWithResult` block wouldn't be called here, but it does in the `catch` below!

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Verify KYC/VerifyKYCContentViewController.swift
mliberatore: Possibly because of the embedded scroll view, but currently, I cannot swipe to dismiss this view controller. Only tapping outside near the status bar dismisses.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Verify KYC/VerifyKYCContentViewController.swift
mliberatore: Sure! Update incoming, looks like the following when hidden:
image

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Verify KYC/VerifyKYCContentViewController.swift
mliberatore: Trying this out, it does work, but the address format is quite different from what's in the Figma. My hunch is to try to match the design as closely as possible to prevent the need to make adjustments post-SDK release once this is being tested, but I’d love your thoughts on that. Going to stash the change for now, but please let me know!

image

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Verify KYC/VerifyKYCViewController.swift
mliberatore: Good call. I missed that specific constant in LinkUI!

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Verify KYC/VerifyKYCContentViewController.swift
mliberatore: Done!

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Verify KYC/VerifyKYCContentViewController.swift
mats-stripe: Let's leave it as-is! Thanks for trying it out!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/BottomSheetViewController.swift
mats-stripe: I wouldn't be totally exposed to exposing this via LinkController, but I'm peaceful with merging this for now!

mliberatore: Cool, I’ll try to swing back around to this once I‘ve gotten everything else merged in the down time while we wait on KYC Refresh API review. I’ll keep notes about what's been exposed as `public` here and revert those if we're able to make a sensible API.

PR #5789, opened 2025-11-03
Payment Sheet: Updates Localizations For KYC Refresh

mliberatore:
## Summary
For the UI introduced in #5787, there are several user-facing strings that need the localization treatment. This uses the existing localized strings, moves one from StripeIdentity to StripeUICore, and adds new ones.

I’ve also run:

```
./ci_scripts/upload_localized_strings_to_lokalise.sh
```

> [!CAUTION]
> While the validation occurs successfully, the upload fails, so I’ll need help completing this work from someone at Stripe.
> `./ci_scripts/upload_localized_strings_to_lokalise.sh: line 17: fetch-password: command not found` 

and 

```
./ci_scripts/l10n/lint.rb
```

Note that the removal of `giropay` is unrelated, but a result of that last script.

## Motivation
See #5787 

## Testing
Made sure the strings in the UI still show up appropriately.

Simulator Screenshot - iPhone 17 - 2025-11-03 at 12 33 44

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Verify KYC/VerifyKYCContentViewController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Verify KYC/VerifyKYCContentViewController.swift
@@ -5,6 +5,7 @@
 //  Created by Michael Liberatore on 10/30/25.
 //
 
+@_spi(STP) import StripeCore
 @_spi(STP) import StripeUICore
 import UIKit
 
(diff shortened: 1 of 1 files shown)


StripeIdentity/StripeIdentity/Resources/Localizations/bg-BG.lproj/Localizable.strings
```diff
@@ -28,8 +28,6 @@
 
 "Could not capture image" = "Заснемането беше неуспешно";
 
-"Date of Birth" = "Дата на раждане";
-
```
mliberatore, line 32: Note that this string moved to StripeUICore, but the translations did not move as part of the script. Is that to be expected, or perhaps an issue running the scripts without proper permissions?

mats-stripe: This should eventually be migrated on its own 👍

PR #5793, opened 2025-11-03
Payment Sheet: Prevents Payment Sheet Analytics When Closing VerifyKYCViewController

mliberatore:
## Summary
There were two areas where payment sheet analytics were being fired automatically when using the new `VerifyKYCViewController`, which seemed undesirable.
1. `BottomSheetViewController`’s `didTapOrSwipeToDismiss()`. This is already prevented since `VerifyKYCViewController` is a subclass and overrides `didTapOrSwipeToDismiss()`.
2. `SheetNavigationBar`‘s `didTapCloseButton()`. This is addressed in this PR.

I’ve elected to pass a `Bool` defaulted to `true` to allow for suppression of this event, and pass `false` from the KYC view controller. A number of other approaches exist, such as moving this to the delegate implementation, but I wanted to keep it simple, since that would require updating 15 delegate implementations to fire the same analytics. Happy to take other suggestions, however.

## Motivation
To keep analytics as clean as possible, and not log unintended events.

## Testing
1. Ran the crypto onramp example on the parent branch, brought up the KYC interface, and dismissed using the navigation bar close button.
2. Saw the following log to the console.
```
LOG ANALYTICS: mc_dismiss - [(key: "pay_var", value: "paymentsheet"), (key: "ocr_type", value: "none"), (key: "apple_pay_enabled", value: 1)]
```
3. Repeated these steps on this PR’s branch.
4. Saw no console log about payment sheet analytics.

## Changelog
N/A

### DIFF
--- a/StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
+++ b/StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
@@ -392,9 +392,11 @@ public final class CryptoOnrampCoordinator: NSObject, CryptoOnrampCoordinatorPro
     }
 
     public func verifyKYCInfo(updatedAddress: Address? = nil, from viewController: UIViewController) async throws -> VerifyKYCResult {
+        analyticsClient.log(.kycInfoVerificationStarted)
         do {
             let linkAccountInfo = try await self.linkAccountInfo
             let apiClient = self.apiClient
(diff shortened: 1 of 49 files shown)


mats-stripe (approved):
Dang nice catch! Not blocking, but while you're here, would you mind adding 

```
shouldLogPaymentSheetAnalyticsOnDismissal: false
``` 

to: `LinkFullConsentViewController` and `PayWithLinkViewController-BaseViewController`? 🙏

mats-stripe (approved):

PR #5796, opened 2025-11-04
PMME SwiftUI tests

gbirch-stripe:
## Summary
title

## Motivation
Better test coverage

## Testing
:)

## Changelog
N/A


gbirch-stripe:
> Consider testing appearance API too? Maybe just another test that sets everything to custom different values.

There's a test in here that tests large font, which I think is the most important part (ensure that the line wrapping works + the height is set correctly in SwiftUI + the logos are sized correctly). The other appearance customizations are already covered by the normal (non-SwiftUI) snapshot tests, so I think they would be redundant

PR #5797, opened 2025-11-04
PMME - Phase and CreationResult enums @frozen

gbirch-stripe:
## Motivation
- remove user need to use unknown default cases
- codify enum changes being a breaking change
- https://jira.corp.stripe.com/browse/MOBILESDK-4192

## Testing
N/A

## Changelog
N/A


gbirch-stripe:
The only remaining non-frozen public enum in PMME is the UserInterfaceStyle. Users never need to switch over this and I don't think it needs to be frozen

yuki-stripe (dismissed):

yuki-stripe (approved):

PR #5800, opened 2025-11-05
Fix handleDeferredIntentConfirmation hanging possibility

porter-stripe:
## Summary
- Injects the correct confirm handler for deferred confirmation (payment methods & CTs)
- Refactors the deferred confirmation to be async/await

## Motivation
- https://jira.corp.stripe.com/browse/MOBILESDK-4186

## Testing
- Updated existing tests

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+DeferredAPI.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+DeferredAPI.swift
@@ -71,7 +71,7 @@ extension PaymentSheet {
         isFlowController: Bool,
         allowsSetAsDefaultPM: Bool = false,
         mandateData: STPMandateDataParams? = nil,
-        confirmHandler: @escaping IntentConfiguration.ConfirmHandler,
+        confirmHandler: @escaping IntentConfiguration.ConfirmHandler
     ) async -> (result: PaymentSheetResult, deferredIntentConfirmationType: STPAnalyticsClient.DeferredIntentConfirmationType?) {
         do {
             var confirmType = confirmType


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+DeferredAPI.swift
```diff
@@ -67,36 +72,37 @@ extension PaymentSheet {
...
-                        let errorAnalytic = ErrorAnalytic(event: .unexpectedPaymentSheetConfirmationError,
-                                                          error: PaymentSheetError.unexpectedNewPaymentMethod,
-                                                          additionalNonPIIParams: ["payment_method_type": newPaymentMethod.type])
-                        STPAnalyticsClient.sharedClient.log(analytic: errorAnalytic)
-                    }
-                    stpAssert(newPaymentMethod == nil)
-                    paymentMethod = try await configuration.apiClient.createPaymentMethod(with: params, additionalPaymentUserAgentValues: makeDeferredPaymentUserAgentValue(intentConfiguration: intentConfig))
-                    confirmType = .new(params: params, paymentOptions: paymentOptions, paymentMethod: paymentMethod, shouldSave: shouldSave, shouldSetAsDefaultPM: shouldSetAsDefaultPM)
+        confirmHandler: @escaping IntentConfiguration.ConfirmHandler,
+        preparePaymentMethodHandler: IntentConfiguration.PreparePaymentMethodHandler?
```
joyceqin-stripe: Is there a need to pass `preparePaymentMethodHandler` in if we already have access to `intentConfig`? I understand passing the `confirmHandler` in since we have it as a non-Optional, but `preparePaymentMethodHandler` is Optional either way.

porter-stripe: Good callout! No I don't think we need to.

PR #5814, opened 2025-11-10
Fallback to system image of icon_x_standalone

porter-stripe:
## Summary
- In rare cases where the bundle asset for `icon_x_standalone` fails to load (e.g., bundle configuration issues or asset catalog corruption), the SDK would return a zero-sized image, potentially causing layout issues or crashes
  downstream. We now fallback to a system close icon instead rather than crash.

## Motivation
- https://jira.corp.stripe.com/browse/RUN_MOBILESDK-4812

## Testing
- Manual

## Changelog
N/A


porter-stripe:
Unsure if this is the right fix, it likely occurs with other assets due to some downloading issue. Gonna close.

PR #5819, opened 2025-11-11
Active captcha

joyceqin-stripe:
## Summary

Run Stripe js on a webview for active captcha.

For reference, this is the webview bridge for the challenge: https://git.corp.stripe.com/stripe-internal/pay-server/blob/master/frontend/ocs-mobile/mobile-confirmation-challenge/src/components/webview.tsx
## Motivation

HCaptcha project
## Testing

Test using the "stripe_shop_test" merchant and the card `4000 0000 0000 1208`.

https://github.com/user-attachments/assets/e65d2041-cf63-4229-81c9-ce262ab6546b

## Changelog

N/A


StripePayments/StripePayments/Source/Internal/IntentConfirmationChallengeViewController.swift
```diff
@@ -0,0 +1,322 @@
...
+        let configuration = WKWebViewConfiguration()
+        configuration.defaultWebpagePreferences.allowsContentJavaScript = true
+
+        // Setup message handlers
+        let contentController = WKUserContentController()
+        contentController.add(self, name: "getInitParams")
+        contentController.add(self, name: "onReady")
+        contentController.add(self, name: "onSuccess")
+        contentController.add(self, name: "onError")
+        contentController.add(self, name: "consoleLog")
```
toluo-stripe: We can take `consoleLog` out. We should follow the functions defined [here](https://stripe.sourcegraphcloud.com/stripe-internal/pay-server/-/blob/frontend/ocs-mobile/mobile-confirmation-challenge/src/components/webview.tsx?L36-41)

StripePayments/StripePayments/Source/Internal/IntentConfirmationChallengeViewController.swift
```diff
@@ -0,0 +1,322 @@
...
+                                    catch(e) { return String(arg); }
+                                }
+                                return String(arg);
+                            }).join(' ')
+                        });
+                    } catch(e) {}
+                };
+            });
+        })();
+        """
```
toluo-stripe: Just FYI, the frontend app will not call console.log. Might be useful if you want to add logs to the frontend server locally

StripePayments/StripePayments/Source/Internal/IntentConfirmationChallengeViewController.swift
```diff
@@ -0,0 +1,322 @@
...
+                background: transparent !important;
+            }
+
+            /* Hide react-root */
+            #react-root {
+                display: none !important;
+            }
+        `;
+        document.head.appendChild(style);
+        """
```
toluo-stripe: I think we can take this out now. [This PR](https://git.corp.stripe.com/stripe-internal/pay-server/pull/1253658) removed the frontend UI

joyceqin-stripe: Are there any plans to take out the spinner? I personally don't like that there are multiple spinners spinning. If you look in this video clip, you can see a blurple spinner right before the hCaptcha spinner shows up, and on top of that, the Processing button also already has a spinner.

https://github.com/user-attachments/assets/e387594f-1647-4835-a220-5cc8518961d7

toluo-stripe: Are you referring to the spinner on the modal? That's rendered by the hCaptcha js sdk, so we won't be able to get rid of that. 
There is one rendered by our frontend that I can remove. I think it's not showing here because of the .7s alpha animation.

joyceqin-stripe: I'd prefer to remove the frontend blurple spinner. It's showing here, but it's very brief

toluo-stripe: Ok, I'll take it out

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
```diff
@@ -1367,6 +1367,12 @@ public class STPPaymentHandler: NSObject {
                         returnURL = nil
                     }
                     _handleRedirect(to: redirectURL, withReturn: returnURL, useWebAuthSession: false)
+                case .intentConfirmationChallenge:
+                    if #available(iOS 14.0, *) {
+                        _handleIntentConfirmationChallenge()
+                    } else {
+                        // Fallback on earlier versions
```
wooj-stripe: What happens if _handleIntentConfirmationChallenge() doesn't execute?

StripePayments/StripePayments/Source/Internal/IntentConfirmationChallengeViewController.swift
```diff
@@ -0,0 +1,254 @@
...
+//  IntentConfirmationChallengeViewController.swift
+//  StripePayments
+//
+//  Created by Joyce Qin on 11/4/25.
+//
+
+import Foundation
+@_spi(STP) import StripeCore
+import UIKit
+@preconcurrency import WebKit
```
wooj-stripe, line 10: I've never seen @preconcurrency used for an import -- why do we need to do this?

StripePayments/StripePayments/Source/Internal/IntentConfirmationChallengeViewController.swift
```diff
@@ -0,0 +1,254 @@
+//
+//  IntentConfirmationChallengeViewController.swift
+//  StripePayments
+//
+//  Created by Joyce Qin on 11/4/25.
+//
+
+import Foundation
+@_spi(STP) import StripeCore
+import UIKit
```
wooj-stripe, line 9: Build works fine if we don't import Foundation, so seems like we can remove it

StripePayments/StripePayments/Source/Internal/IntentConfirmationChallengeViewController.swift
```diff
@@ -0,0 +1,254 @@
...
+    private func injectBridgeScripts(into contentController: WKUserContentController) {
+        // CRITICAL: Set initParams BEFORE page loads so it's available immediately
+        // The web page expects window.initParams to be set via window.setInitParams()
+        let initParamsScript = """
+        window.setInitParams = function(params) {
+            window.initParams = params;
+        };
+        window.setInitParams({
+            clientSecret: "\(clientSecret)",
+            publishableKey: "\(apiClient.publishableKey ?? "")"
```
wooj-stripe: Seems like we should assert loudly if publishable key is empty at this point

StripePayments/StripePayments/Source/Internal/IntentConfirmationChallengeViewController.swift
```diff
@@ -0,0 +1,254 @@
...
+
+    private func loadChallenge() {
+        let request = URLRequest(url: challengeURL)
+        webView.load(request)
+    }
+
+    // MARK: - Handlers
+    private func handleReady() {
+        DispatchQueue.main.async {
+            UIView.animate(withDuration: 0.7) {
```
wooj-stripe: feels a little long -- i think typical animations run at 0.3 seconds.

StripePayments/StripePayments/Source/Internal/IntentConfirmationChallengeViewController.swift
```diff
@@ -0,0 +1,254 @@
...
+        #if DEBUG
+        print("[IntentConfirmationChallenge] Error: \(error)")
+        #endif
+
+        completion(.failure(error))
+    }
+
+    /// Validates that the message comes from the expected Stripe origin
+    private func isValidMessageOrigin(_ message: WKScriptMessage) -> Bool {
+        let validHosts = ["pay.stripe.com", "js.stripe.com", "mobile-active-challenge-764603794666.us-central1.run.app"]
```
wooj-stripe: To avoid having to type this name out multiple times, maybe when we define the challenge url we can use:
```
    private static let challengeURLHost = "mobile-active-challenge-764603794666.us-central1.run.app"
    private let challengeURL = URL(string: "https://\(challengeURLHost)/")!
```
then, this line can be:
```
        let validHosts = ["pay.stripe.com", "js.stripe.com", Self.challengeURLHost]
```

toluo-stripe: I think we can take these out actually - `"pay.stripe.com", "js.stripe.com",`. We won't be receiving any messages from stripe js

StripePayments/StripePayments/Source/Internal/IntentConfirmationChallengeViewController.swift
toluo-stripe: https://git.corp.stripe.com/stripe-internal/pay-server/pull/1262398

StripePayments/StripePayments/Source/Internal/IntentConfirmationChallengeViewController.swift
joyceqin-stripe: Without it, CI fails with `add '@preconcurrency' to suppress 'Sendable'-related warnings from module 'WebKit'`. I referenced `ECEViewController.swift` to see if it handled it any differently, but the `@preconcurrency` import exists in there as well.

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
joyceqin-stripe: Good question. We can add OS version to the gate so that active captcha won't show up for anyone with iOS version under 14.0.

StripePayments/StripePayments/Source/Internal/IntentConfirmationChallengeViewController.swift
```diff
@@ -0,0 +1,302 @@
...
+    }
+
+    private func handleError(_ error: Error) {
+        if case ChallengeError.webError(_, let type, let code) = error {
+            STPAnalyticsClient.sharedClient.logIntentConfirmationChallengeError(errorType: type, errorCode: code, duration: Date().timeIntervalSince(startTime), fromBridge: true)
+        } else {
+            STPAnalyticsClient.sharedClient.logIntentConfirmationChallengeError(error: error, duration: Date().timeIntervalSince(startTime), fromBridge: false)
+        }
+        cleanup()
+        completion(.failure(error))
```
wooj-stripe, line 151: It looks like we've coupled receiving an error with shutting down/cleaning up the webview. I suspect this is expected behavior, but it feels like we're loosing some flexibility here in that the webview may want to throw an error, but not necessarily want to have the webview completely close/dismiss. Is there any reason to decouple this, or should we just keep this as-is? cc: @toluo-stripe

StripePayments/StripePayments/Source/Internal/IntentConfirmationChallengeViewController.swift
```diff
@@ -0,0 +1,302 @@
...
+            #if DEBUG
+            print("[IntentConfirmationChallenge] ⚠️ Invalid message origin: \(message.frameInfo.securityOrigin.host)")
+            #endif
+            return
+        }
+
+        switch message.name {
+        case "getInitParams":
+            #if DEBUG
+            print("[IntentConfirmationChallenge] getInitParams called (params already injected)")
```
wooj-stripe: Params already injected? should this never be called?

StripePayments/StripePayments/Source/Internal/IntentConfirmationChallengeViewController.swift
```diff
@@ -0,0 +1,302 @@
...
+@available(iOS 14.0, *)
+extension IntentConfirmationChallengeViewController: WKScriptMessageHandler {
+    func userContentController(
+        _ userContentController: WKUserContentController,
+        didReceive message: WKScriptMessage
+    ) {
+        // Validate message origin for security
+        guard isValidMessageOrigin(message) else {
+            #if DEBUG
+            print("[IntentConfirmationChallenge] ⚠️ Invalid message origin: \(message.frameInfo.securityOrigin.host)")
```
wooj-stripe: Should we stpAssert here?

StripePayments/StripePayments/Source/Internal/IntentConfirmationChallengeViewController.swift
```diff
@@ -0,0 +1,302 @@
...
+            if let errorType {
+                params["error_type"] = errorType
+            }
+            if let errorCode {
+                params["error_code"] = errorCode
+            }
+            log(
+                analytic: GenericAnalytic(event: .intentConfirmationChallengeError, params: params)
+            )
+        }
```
wooj-stripe: It is possible to not have an error, but there is an errorType and errorCode?

StripePayments/StripePayments/Source/Internal/IntentConfirmationChallengeViewController.swift
toluo-stripe: The only times we would receive an error through the bridge are when we fail to set up Stripe.js or when `stripeJs.handleNextAction` returns an error. Shutting down the webview is the correct behaviour in these cases. We don't anticipate introducing a scenario in which we throw and expect the webview to remain open

StripePayments/StripePayments/Source/Internal/IntentConfirmationChallengeViewController.swift
toluo-stripe: This should not be possible, from the bridge at least

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
```diff
@@ -1914,6 +1916,98 @@ public class STPPaymentHandler: NSObject {
...
+                presentingVC.present(challengeVC, animated: true, completion: nil)
+            }
+
+            if context.responds(to: #selector(STPAuthenticationContext.prepare(forPresentation:))) {
+                context.prepare?(forPresentation: doChallenge)
+            } else {
+                doChallenge()
+            }
+        } else { // Intent confirmation challenge should be gated to iOS versions 14.0+
+            let unsupportedVersionErrorMessage = "Unable to perform intent confirmation challenge. Requires iOS version 14.0 or later."
```
davidme-stripe, line 1999: Are we gating these challenges server-side based on the device OS?

joyceqin-stripe: I've talked to @toluo-stripe and he said he could add it to the gate. Conformance to `WKScriptMessageHandlerWithReply` is iOS 14+ only, so anyone running < iOS 14 would not be able to complete an active challenge at all. As we don't want to make it impossible for someone to pay, I feel that we shouldn't run captcha for < iOS 14.

joyceqin-stripe: I also recall seeing a thread that [iOS 13 and 14 are almost dead](https://stripe.slack.com/archives/C08FKV9B0PK/p1761852153549989) and we were considering dropping support for the major version, so doesn't seem like too many people would be affected, and if we end up dropping iOS 13 on the next major version, we can remove the OS-specific part of the gate

toluo-stripe: I'll add an os version gate to the backend

davidme-stripe: That all sounds good to me — we can gate on OS version for now, and I'm in support of dropping iOS 13 entirely soon.

StripePayments/StripePayments/Source/Internal/IntentConfirmationChallengeViewController.swift
```diff
@@ -0,0 +1,309 @@
...
+    // MARK: - Properties
+    private let publishableKey: String
+    private let clientSecret: String
+    private let completion: (Result<Void, Error>) -> Void
+
+    private var webView: WKWebView!
+    private var dimmedBackgroundView: UIView!
+
+    // Hard-coded challenge URL
+    private static let challengeHost = "mobile-active-challenge-764603794666.us-central1.run.app"
```
davidme-stripe, line 27: Sorry, one more thing here — can this either be hosted from a domain we control *or* client-side gated somehow so that we're not accessing this domain unless we're in the gate? I don't want this to end up in a production SDK until we've finalized the domain.

joyceqin-stripe: I don't think it's currently possible to trigger the active challenge and access the domain except from stripe_shop_test, and I believe @toluo-stripe is working on getting it stripe-hosted.

When you say a client-side gate, do you mean like with the configuration like with playground switches?

toluo-stripe: We're going to have an sdk version gate for active captcha, so this codepath will not be used until it's ready. We already have sdk version gates for passive captcha and device attestation
https://trailhead.corp.stripe.com/docs/mobile-sdk/runbooks/passive-captcha-runbook

davidme-stripe (approved):

PR #5830, opened 2025-11-14
PMME - anlaytics

gbirch-stripe:
## Testing
Analytics events added to existing tests.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+Internal.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+Internal.swift
@@ -29,6 +29,8 @@ extension PaymentMethodMessagingElement {
     /// - Parameter analyticsClient: Optional analytics client for testing. Defaults to shared client.
     /// - Returns: A `CreationResult` object representing the result of the attempt to load the element.
     static func create(configuration: Configuration, downloadManager: DownloadManager, analyticsClient: STPAnalyticsClientProtocol) async -> CreationResult {
+        // This being a singleton can theoretically cause problems when using multiple sessions-generating products at once
+        // TODO(ocs-mobile): Make this not a singleton
         AnalyticsHelper.shared.generateSessionID()
         analyticsClient.addClass(toProductUsageIfNecessary: PaymentMethodMessagingElement.self)
         let analyticsHelper = PMMEAnalyticsHelper(configuration: configuration, analyticsClient: analyticsClient)
@@ -89,6 +91,7 @@ extension PaymentMethodMessagingElement {
                 style: configuration.appearance.style,
                 downloadManager: downloadManager
             ).first else {
+                // There were no images in `paymentPlan.content.images`
                 // This should never happen, but if it does we log an error and attempt to fall back to a multi-partner style
                 //      (so that we can use the promotion text, which doesn't require a logo, instead of inline) without logos
                 _ = Self.assertAndLogMissingField("logo", apiClient: configuration.apiClient)
@@ -129,6 +132,7 @@ extension PaymentMethodMessagingElement {
                 throw Self.assertAndLogMissingField("info_url", apiClient: configuration.apiClient)
             }
 
+            // Use the list of images returned as the source of truth for what images to display and thus don't validate
             let apiImages = apiResponse.paymentPlanGroups.flatMap { $0.content.images }
             let logos = try await Self.getIconSet(
                 for: apiImages,
@@ -155,6 +159,7 @@ extension PaymentMethodMessagingElement {
         return error
     }
 
+    // Throws when the underlying downloadManager throws
     private static func getIconSet(for iconUrls: [APIResponse.Image], style: Appearance.UserInterfaceStyle, downloadManager: DownloadManager) async throws -> [LogoSet] {
         // Fetch all images concurrently
         // We want to preserve the order of the icons as provided by the API,
@@ -185,7 +190,7 @@ extension PaymentMethodMessagingElement {
                     // For all non-automatic styles, we fetch one image and use it for
                     //     both light and dark
                     taskGroup.addTask {
-                        let lightImage = try await DownloadManager.sharedManager.downloadImage(url: image.lightThemePng.url)
+                        let lightImage = try await downloadManager.downloadImage(url: image.lightThemePng.url)
                         return (
                             index: i,
                             iconSet: LogoSet(
@@ -198,7 +203,7 @@ extension PaymentMethodMessagingElement {
                     }
                 case .alwaysDark:
                     taskGroup.addTask {
-                        let darkImage = try await DownloadManager.sharedManager.downloadImage(url: image.darkThemePng.url)
+                        let darkImage = try await downloadManager.downloadImage(url: image.darkThemePng.url)
                         return (
                             index: i,
                             iconSet: LogoSet(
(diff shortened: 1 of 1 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+Internal.swift
```diff
@@ -28,16 +26,30 @@ extension PaymentMethodMessagingElement {
     /// Internal version of create() that allows injection of a DownloadManager for testing.
     /// - Parameter configuration: Configuration for the PaymentMethodMessagingElement.
     /// - Parameter downloadManager: The DownloadManager instance to use for downloading images.
+    /// - Parameter analyticsClient: Optional analytics client for testing. Defaults to shared client.
     /// - Returns: A `CreationResult` object representing the result of the attempt to load the element.
-    static func create(configuration: Configuration, downloadManager: DownloadManager) async -> CreationResult {
+    static func create(configuration: Configuration, downloadManager: DownloadManager, analyticsClient: STPAnalyticsClientProtocol) async -> CreationResult {
+        AnalyticsHelper.shared.generateSessionID()
```
yuki-stripe, line 34: This is already a problem today with CustomerSheet so we don't need to block on it but this can theoretically cause other product sessions to get messed up if you use multiple session-generating products at once. We really need to not make this a global singleton.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+Internal.swift
```diff
@@ -112,8 +126,7 @@ extension PaymentMethodMessagingElement {
 
             // unexpected / error case
             guard let infoUrl = apiResponse.content.learnMore?.url else {
-                // TODO(gbirch) add error analytics
-                throw PaymentMethodMessagingElementError.unexpectedResponseFromStripeAPI
+                throw Self.assertAndLogMissingField("info_url", apiClient: configuration.apiClient)
             }
 
             let apiImages = apiResponse.paymentPlanGroups.flatMap { $0.content.images }
```
yuki-stripe, line 136: should we validate images as well?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+Internal.swift
```diff
@@ -79,27 +91,29 @@ extension PaymentMethodMessagingElement {
             ).first else {
                 // This should never happen, but if it does we log an error and attempt to fall back to a multi-partner style
                 //      (so that we can use the promotion text, which doesn't require a logo, instead of inline) without logos
-                stpAssertionFailure("No images returned by API")
-                // TODO(gbirch) add error analytics
+                _ = Self.assertAndLogMissingField("logo", apiClient: configuration.apiClient)
```
yuki-stripe, line 97: image could fail to load for expected reasons too. Not sure what to do, maybe we should only consider 'request completed w/ non-200 http status code' as unexpected or something.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+Internal.swift
```diff
@@ -69,8 +82,7 @@ extension PaymentMethodMessagingElement {
 
             // unexpected / error cases
             guard let infoUrl = paymentPlan.content.learnMore?.url else {
-                // TODO(gbirch) add error analytics
-                throw PaymentMethodMessagingElementError.unexpectedResponseFromStripeAPI
+                throw Self.assertAndLogMissingField("info_url", apiClient: configuration.apiClient)
```
yuki-stripe, line 87: I think Android has different behavior based on that eng review doc, where I think this turns into no content instead of an error. Not blocking but it'd be nice to have the same story.

gbirch-stripe: Yes, I am changing this but in a subsequent PR. Want to keep this PR scoped to analytics

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+Internal.swift
gbirch-stripe: Hmm, I wasn't aware of this. I think many users will be using PMME alongside other products. Do you know the level of risk here? I'd hate to not have a session ID at all but understand it may be necessary

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+Internal.swift
gbirch-stripe: Mind elaborating on what validation we would do?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+Internal.swift
gbirch-stripe: This doesn't assert and log if `getIconSet` throws, only when there are no images returned from the API. The latter is definitely an unexpected scenario we should assert and log on. The former happens with the networking layer throws an error, which we consider a load failure and pass on to the user both here and for the main api response

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+Internal.swift
yuki-stripe: mmm if # of images doesn't match what we expect? e.g. 0 images?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+Internal.swift
yuki-stripe: Low risk. Each time we repeat this (have a new product call `generate`) it adds some amount more risk. e.g. of you initialized PMME and MPE and then interacted with PMME, its analytics will send MPE's session ID.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+Internal.swift
yuki-stripe: Yeah never mind, I was thinking if the image like 404s I'd like to know but we can catch that from our general load analytics.

yuki-stripe (dismissed):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMME+Internal.swift
gbirch-stripe: Unfortunately the list of objects in the `images` field of the API response is the source of truth. There's also a `payment_plans` field in the API response that lists payment methods, but those are sometimes different than what we should actually be showing, so we can't rely on them.
Additionally, we can't rely on the payment methods we request from the backend, because there are sometimes ones that we request that aren't available, and additionally we usually pass none and let it be determined by dynamic payment methods. 
In short, we unfortunately have to use this list as the source of truth of what PMs we should be showing.

PR #5833, opened 2025-11-17
Add billing details support to Affirm payment method initializer

jaynewstrom-stripe:
Added billingDetails parameter to STPPaymentMethodParams Affirm initializer with default value for backward compatibility. Updated both Swift and Objective-C APIs:
- Swift: Added billingDetails parameter with default nil value
- Objective-C: Kept original method and added new overload with billingDetails

Added test coverage for Affirm payment method with billing details.

Committed-By-Agent: claude


jaynewstrom-stripe:
I think it's better not to do this, and modify the RN SDK to add it on after.

PR #5834, opened 2025-11-18
CryptoOnramp SDK: (Example App) Swaps Unsupported SF Symbols

mliberatore:
## Summary
In the CryptoOnramp SDK’s example app, we're using SF Symbol `wallet.bifold`. However, on earlier OS versions, this silently fails since it's only available starting in iOS 18. Here's how it looked on modern OS versions:

| iOS 26 - Sign In | iOS 26 - Wallet Selection | iOS 26 - Review |
| --- | --- | --- |
|  |  |  |

But on earlier OS versions:

| iOS 16 - Log In | iOS 16 - Wallet Selection | iOS 16 - Review |
| --- | --- | --- |
|  |  |  |

To fix this, I’ve selected new SF Symbols for each of these views that appropriately match the step in the flow, and are available on older OS versions.

## Motivation
Ensure that icons in the example app always appear.

## Testing
I visited each screen in light and dark mode on iOS 16. See the table of screenshots below.

| iOS 16 - Log In | iOS 16 - Wallet Selection | iOS 16 - Review |
| --- | --- | --- |
|  |  |  |
|  |  |  |

## Changelog

N/A

### DIFF
--- a/Example/CryptoOnramp Example/CryptoOnramp Example/PaymentSummaryView.swift
+++ b/Example/CryptoOnramp Example/CryptoOnramp Example/PaymentSummaryView.swift
@@ -51,9 +51,10 @@ struct PaymentSummaryView: View {
                     .fill(.tint)
                     .frame(width: 44, height: 44)
 
-                Image(systemName: "wallet.bifold.fill")
+                Image(systemName: "cart.fill")
                     .font(.system(size: 20, weight: .semibold))
-                    .offset(x: 1, y: -1)
+                    .offset(x: -0.5, y: 1)
+                    .foregroundStyle(.white)
             }
 
             VStack(alignment: .leading, spacing: 2) {


Example/CryptoOnramp Example/CryptoOnramp Example/PaymentSummaryView.swift
```diff
@@ -51,9 +51,10 @@ struct PaymentSummaryView: View {
                     .fill(.tint)
                     .frame(width: 44, height: 44)
 
-                Image(systemName: "wallet.bifold.fill")
+                Image(systemName: "cart.fill")
                     .font(.system(size: 20, weight: .semibold))
-                    .offset(x: 1, y: -1)
+                    .offset(x: -0.5, y: 1)
+                    .foregroundStyle(.white)
```
mats-stripe, line 57: Thanks for fixing this tint 🙈

PR #5835, opened 2025-11-18
Promote CheckScanningController to PrivatePreviewConnect

bredmond-stripe:
## Summary

Promote CheckScanningController to PrivatePreviewConnect

## Motivation

https://docs.google.com/document/d/1OptB57egJxy2ibRh9V8iV8OD7xPA3JR7BbSY4Lmjnt0/edit?tab=t.0

## Testing

Manual testing in simulator, automated unit tests

### DIFF
--- a/StripeConnect/StripeConnectTests/Components/CheckScanningControllerTests.swift
+++ b/StripeConnect/StripeConnectTests/Components/CheckScanningControllerTests.swift
@@ -77,7 +77,7 @@ class CheckScanningControllerTests: XCTestCase {
     }
 
     @MainActor
-    func testCallbackInvoked() async throws {
+    func testCallbackSuccess() async throws {
         let controller = componentManager.createCheckScanningController()
 
         let expectationCallback = XCTestExpectation(description: "handleCheckScanSubmitted called")
@@ -91,14 +91,50 @@ class CheckScanningControllerTests: XCTestCase {
             {"setHandleCheckScanSubmitted":true}
             """)
 
-        try await controller.webVC.webView.evaluateMessage(
-            name: "callSupplementalFunction",
-            json: """
-                {"functionName": "handleCheckScanSubmitted","invocationId":"0","args":[{"checkScanToken": "uspchk_123"}]}
-            """
-        )
+        try await controller.webVC.webView.evaluateCallSupplementalFunction(functionName: .handleCheckScanSubmitted,
+                                                                            invocationId: "0",
+                                                                            args: "[{\"checkScanToken\": \"uspchk_123\"}]")
 
         await fulfillment(of: [expectationCallback], timeout: TestHelpers.defaultTimeout)
         XCTAssertEqual(delegate.received, .init(checkScanToken: "uspchk_123"))
+    }
+
+    class CheckScanningControllerDelegateWithCallbackError: CheckScanningControllerDelegate {
+        enum CallbackFailure: Error {
+            case fail
+        }
+
+        func checkScanning(_ checkScanning: CheckScanningController, didSubmitCheckScan: CheckScanningController.CheckScanDetails) async throws {
+            throw CallbackFailure.fail
+        }
+    }
+
+    @MainActor
+    func testCallbackError() async throws {
+        let controller = componentManager.createCheckScanningController()
+
+        let delegate = CheckScanningControllerDelegateWithCallbackError()
+
+        controller.delegate = delegate
+
+        try await controller.webVC.webView.evaluateMessageWithReply(name: "fetchInitComponentProps",
+                                                                    json: "{}",
+                                                                    expectedResponse: """
+            {"setHandleCheckScanSubmitted":true}
+            """)
+
+        let expectation = try controller.webVC.webView.expectationForMessageReceived(
+            sender: SupplementalFunctionCompletedSender(payload: .init(
+                functionName: .handleCheckScanSubmitted,
+                invocationId: "0",
+                result: .error("Error calling supplemental function")
+            ))
+        )
+
+        try await controller.webVC.webView.evaluateCallSupplementalFunction(functionName: .handleCheckScanSubmitted,
+                                                                            invocationId: "0",
+                                                                            args: "[{\"checkScanToken\": \"uspchk_123\"}]")
+
+        await fulfillment(of: [expectation], timeout: TestHelpers.defaultTimeout)
     }
 }


StripeConnect/StripeConnect/Source/Components/CheckScanningController.swift
```diff
@@ -24,11 +39,20 @@ public final class CheckScanningController {
...
-         analyticsClientFactory: ComponentAnalyticsClientFactory,
-         handleCheckScanSubmitted: @escaping HandleCheckScanSubmittedFn) {
-        let supplementalFunctions = SupplementalFunctions(handleCheckScanSubmitted: handleCheckScanSubmitted)
+         analyticsClientFactory: ComponentAnalyticsClientFactory) {
+        let supplementalFunctions = SupplementalFunctions(handleCheckScanSubmitted: { [weak self] details in
+              guard let self = self else {
+                 assertionFailure()
+                 throw CallbackError.controllerDisappeared
+              }
+              try await self.delegate?.checkScanning(self, didSubmitCheckScan: details)
```
bredmond-stripe: Should we have a runtime error if delegate is nil? Could also enforce that at `present`.

StripeConnect/StripeConnect/Source/Components/CheckScanningController.swift
```diff
@@ -1,8 +1,22 @@
 import UIKit
 
-@_spi(DashboardOnly)
+/// A view controller representing a check-scanning component
+/// - Seealso: [Check scanning component documentation](https://docs.stripe.com/connect/supported-embedded-components/check-scanning?platform=ios)
+@_spi(PreviewConnect)
```
torrance-stripe: Think we actually need a new annotation, `PrivatePreviewConnect`. `PreviewConnect` is currently only used for Payments + Payouts which are public preview (no gated access needed, but not yet GA-ready). 

And in the definition of `PrivatePreviewConnect`, you should point people towards docs which allows them to request access

bredmond-stripe: Is there an actual central definition of `PreviewConnect`? I only see it used not defined.

> you should point people towards docs which allows them to request access

For this feature it's gated but also we already have a list of users and don't need a bunch of inbound for some time

torrance-stripe: Oh huh, I think I was confusing the need to "define" `PreviewConnect` with how this works in Android. But regardless, we should still be using `PrivatePreviewConnect` if possible!. 

> For this feature it's gated but also we already have a list of users and don't need a bunch of inbound for some time

I guess you've already added docs link, so that should be good for now. Adding docs here also helps us redirect users who see this component in the API and then reach out to ask how to get access.

bredmond-stripe: done

StripeConnect/StripeConnect/Source/Components/CheckScanningController.swift
```diff
@@ -1,8 +1,22 @@
...
+/// - Seealso: [Check scanning component documentation](https://docs.stripe.com/connect/supported-embedded-components/check-scanning?platform=ios)
+@_spi(PrivatePreviewConnect)
 @available(iOS 15, *)
 public final class CheckScanningController {
+    typealias HandleCheckScanSubmittedFn = (CheckScanDetails) async throws -> Void
+
+    /// Contains details about a check scan
+    public struct CheckScanDetails: Decodable, Equatable {
+        /// Token identifying the check scan, see UsPaperCheck
+        public var checkScanToken: String
```
yuki-stripe: Should be `let`

StripeConnect/StripeConnect/Source/Components/CheckScanningController.swift
```diff
@@ -1,8 +1,22 @@
...
+/// A view controller representing a check-scanning component
+/// - Seealso: [Check scanning component documentation](https://docs.stripe.com/connect/supported-embedded-components/check-scanning?platform=ios)
+@_spi(PrivatePreviewConnect)
 @available(iOS 15, *)
 public final class CheckScanningController {
+    typealias HandleCheckScanSubmittedFn = (CheckScanDetails) async throws -> Void
+
+    /// Contains details about a check scan
+    public struct CheckScanDetails: Decodable, Equatable {
+        /// Token identifying the check scan, see UsPaperCheck
```
yuki-stripe: can we link to an API ref or something for UsPaperCheck?

StripeConnect/StripeConnect/Source/Components/CheckScanningController.swift
yuki-stripe: That's a good idea.

StripeConnect/StripeConnect/Source/Components/CheckScanningController.swift
```diff
@@ -85,11 +112,20 @@ public final class CheckScanningController {
...
-@_spi(DashboardOnly)
+@_spi(PrivatePreviewConnect)
 @available(iOS 15, *)
 public protocol CheckScanningControllerDelegate: AnyObject {
+    /// Called when the component fails to load (e.g., network issue during initial fetch). To try again, initialize a new CheckScanningController.
     func checkScanning(_ checkScanning: CheckScanningController,
                        didFailLoadWithError error: Error)
+
+    /// Called when the check scanning process is complete with a check in the `verified` or `manual_review` states.
+    /// If this function throws an Error, then the user will be given an opportunity to submit the check again, effectively resulting in a retry of this function.
```
yuki-stripe, line 132: do we test this behavior?

StripeConnect/StripeConnect/Source/Components/CheckScanningController.swift
bredmond-stripe: done

StripeConnect/StripeConnect/Source/Components/CheckScanningController.swift
bredmond-stripe: done. It points to gated docs though.

StripeConnect/StripeConnect/Source/Components/CheckScanningController.swift
bredmond-stripe: What specifically do you want to assert? That the submit button in the webview is clickable? That might be hard to express in a test here. We could possibly intercept the message back to the webview that indicates the function call completed (and assert that its result was an error)?

StripeConnect/StripeConnect/Source/Components/CheckScanningController.swift
bredmond-stripe: Added a runtime check in present() (which now `throws`, is that acceptable?) and force unwrapped delegate in the closures in the initializer

StripeConnect/StripeConnect/Source/Components/CheckScanningController.swift
```diff
@@ -51,7 +51,7 @@ public final class CheckScanningController {
                  assertionFailure()
                  throw CallbackError.controllerDisappeared
               }
-              try await self.delegate?.checkScanning(self, didSubmitCheckScan: details)
```
yuki-stripe: We should avoid force casting, since that crashes the app. I'd recommend undoing this change and adding an `assert`.

StripeConnect/StripeConnect/Source/Components/CheckScanningController.swift
```diff
@@ -62,13 +62,21 @@ public final class CheckScanningController {
             fetchInitProps: { Props(supplementalFunctions: supplementalFunctions) },
             didFailLoadWithError: { [weak self] error in
                 guard let self else { return }
-                delegate?.checkScanning(self, didFailLoadWithError: error)
+                delegate!.checkScanning(self, didFailLoadWithError: error)
```
yuki-stripe: We should undo this. Probably don't want to assert since it's valid to write code that does something like:

1. create the CheckScanningController
2. later, user dismisses this VC, causing MyVC to go out of scope
3. CheckScanningController fails to load, delegate is nil (used to be MyVC)

StripeConnect/StripeConnect/Source/Components/CheckScanningController.swift
```diff
@@ -62,13 +62,21 @@ public final class CheckScanningController {
...
         )
     }
 
+    enum DelegateError: Error {
+        case delegateNotSet
+    }
+
     /// Presents the check scanning experience
-    public func present(from viewController: UIViewController, animated: Bool = true) {
+    public func present(from viewController: UIViewController, animated: Bool = true) throws {
```
yuki-stripe: This is a change to the gaveled public API so things like this would need to be reviewed. And we shouldn't do this, asserting and maybe early returning here is fine.

StripeConnect/StripeConnect/Source/Components/CheckScanningController.swift
bredmond-stripe: done

StripeConnect/StripeConnect/Source/Components/CheckScanningController.swift
bredmond-stripe: done

yuki-stripe (dismissed):

StripeConnect/StripeConnect/Source/Components/CheckScanningController.swift
bredmond-stripe: We actually have an even weirder situation with `didFailLoadWithError` now that I think about it. Since the load happens async after initialization, we can't guarantee that the delegate is set here. Currently we try to enforce that the delegate is set before `present` because that guarantees that delegate is set before `didSubmitCheckScan` is called (from the webview, which requires user interaction).

@torrance-stripe I'm not sure we can guarantee that the app is able to set the delegate before `didFailLoadWithError` and it may sometimes be set and sometimes not. I assume we have this issue already in account-onboarding. What do you think we should do about this?

bredmond-stripe: From offline discussion: looks like this pattern spans all existing components and if we see this cause an issue in practice later we can make a consistent fix across all components. Although it would likely involve an interface change so maybe on the next major version.

StripeConnect/StripeConnect/Source/Components/CheckScanningController.swift
yuki-stripe: Specifically this behavior - "If this function throws an Error, then the user will be given an opportunity to submit the check again". Intercepting sounds good or maybe a E2E UI test. Not blocking.

yuki-stripe (dismissed):

StripeConnect/StripeConnect/Source/Components/CheckScanningController.swift
bredmond-stripe: Added a test which verifies the message sent back to the webview. Consider it a contract test. We have tests on the JS side that when such an error response is received, the form can be submitted again.

yuki-stripe (approved):

PR #5841, opened 2025-11-19
Dead code cleanup

gbirch-stripe:
## Summary
Removing new dead code found by turning on spi(STP)-aware periphery checking


StripeCore/StripeCore/Source/API Bindings/STPAPIClient+FileUpload.swift
```diff
@@ -165,48 +165,6 @@ extension STPAPIClient {
         )
     }
 
-    /// Uses the Stripe file upload API to upload a JPEG encoded image.
-    ///
```
davidme-stripe, line 169: Can we keep this comment? Can probably move it down to the function signature below (which is the one that is actually used, with a `FileAndUploadMetrics`)

PR #5844, opened 2025-11-24
Fix payment method decoding for IBP on FC Lite

mats-stripe:
## Summary

Some details on the issue: When the Instant Debits completes in a webview, the redirect URL contains a base64 encoded version of the payment method object. That looks like this:

```
stripe-auth://link-accounts/success?payment_method_id=pm_1SX3G9CeQ8xWuVRgK52G5g1G&payment_method=eyJpZCI6InBtXzFTWDNHOUNlUTh4V3VWUmdLNTJHNW...&bank_name=Payment+Success&last4=6789
```

On the client, we were attempting to decode the data into a `LinkBankPaymentMethod` object using `STPAPIClient.decodeResponse`, like this:

```swift
let result: Result<LinkBankPaymentMethod, Error> = STPAPIClient.decodeResponse(
    data: data,
    error: nil,
    response: nil
)
```

Notably, we weren't passing a response since we were just wanting to decode a `Data` object. A [recent change](https://github.com/stripe/stripe-ios/commit/6276b8d#diff-31ec0917418969f2f83d3d6ff1839bbcd0c0cd504d1a0202b87eae1df21ff57bR563-R567) added a guard condition in `STPAPIClient.decodeResponse` to check that the response was non-nil and had a 2xx http status. Since we were passing nil as the response, it would always result in `throw NSError.stp_genericFailedToParseResponseError()` 💥 

This fix switches our implementation to use `StripeJSONDecoder.decode` directly. I've also added a new analytic event to log when we're unable to complete the instant debits flow: 

```
stripeios.financialconnections.instant_debits_completion.failed
```

Separately, I think we should refactor `STPAPIClient.decodeResponse` to either no longer require the response be non-nil in the function implementation, or to mark it as non-optional in the function signature.

## Motivation

#ir-auburn-proof

## Testing

Before:

After:

https://github.com/user-attachments/assets/e3bd10d4-2d86-43b3-826b-bd38b2028680

## Changelog

```
- [Fixed] Fixes Instant Bank Payments when used in a webview flow.
```

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/Web/FinancialConnectionsWebFlowViewController.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/Web/FinancialConnectionsWebFlowViewController.swift
@@ -192,18 +192,13 @@ extension FinancialConnectionsWebFlowViewController {
                                 self.notifyDelegateOfSuccess(result: .linkedAccount(id: linkedAccountId))
                             } else {
                                 let error = FinancialConnectionsSheetError.unknown(
-                                    debugDescription: "Invalid payment_method returned"
+                                    debugDescription: "Missing payment_method or linked_account in return URL"
                                 )
-                                self.logInstantDebitsCompletionFailure(error: error)
-                                self.notifyDelegateOfFailure(error: error)
+                                throw error
                             }
                         } catch {
                             self.logInstantDebitsCompletionFailure(error: error)
-                            self.notifyDelegateOfFailure(
-                                error: FinancialConnectionsSheetError.unknown(
-                                    debugDescription: "Invalid payment_method returned"
-                                )
-                            )
+                            self.notifyDelegateOfFailure(error: error)
                         }
                     } else {
                         self.fetchSession()
@@ -333,13 +328,12 @@ extension FinancialConnectionsWebFlowViewController {
             event: .instantDebitsCompletionFailed,
             error: error,
             additionalNonPIIParams: [
-                "flow": "fc_sdk",
+                "flow": "fc_sdk_web",
             ]
         )
         STPAnalyticsClient.sharedClient.log(analytic: errorAnalytic)
     }
 }
-
 
 // MARK: - STPURLCallbackListener
 
--- a/StripePaymentSheet/StripePaymentSheet/Source/Internal/FC Lite/Controllers/FCLiteContainerViewController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Internal/FC Lite/Controllers/FCLiteContainerViewController.swift
@@ -93,8 +93,7 @@ class FCLiteContainerViewController: UIViewController {
                         completion(.completed(.linkedAccount(id: linkedAccountId)))
                     } else {
                         let error = FCLiteError.linkedBankUnavailable
-                        logInstantDebitsCompletionFailure(error: error)
-                        completion(.failed(error: error))
+                        throw error
                     }
                 } catch {
                     logInstantDebitsCompletionFailure(error: error)
@@ -303,4 +302,3 @@ extension FCLiteContainerViewController: UIAdaptivePresentationControllerDelegat
         viewController.present(alertController, animated: true)
     }
 }
-
--- a/StripeCore/StripeCore/Source/Connections Bindings/URL+LinkBankPaymentMethod.swift
+++ b/StripeCore/StripeCore/Source/Connections Bindings/URL+LinkBankPaymentMethod.swift
@@ -46,4 +46,3 @@ import Foundation
             .removingPercentEncoding
     }
 }
-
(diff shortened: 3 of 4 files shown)


StripeFinancialConnections/StripeFinancialConnections/Source/Web/FinancialConnectionsWebFlowViewController.swift
```diff
@@ -180,16 +180,25 @@ extension FinancialConnectionsWebFlowViewController {
...
+                                )
+                                self.notifyDelegateOfSuccess(result: .instantDebits(instantDebitsLinkedBank))
+                            } else if let linkedAccountId = returnUrl.extractValue(forKey: "linked_account") {
+                                self.notifyDelegateOfSuccess(result: .linkedAccount(id: linkedAccountId))
+                            } else {
+                                let error = FinancialConnectionsSheetError.unknown(
+                                    debugDescription: "Invalid payment_method returned"
+                                )
+                                self.logInstantDebitsCompletionFailure(error: error)
+                                self.notifyDelegateOfFailure(error: error)
```
tillh-stripe: Nit: Could just throw here to remove code duplication with the catch block.

StripeFinancialConnections/StripeFinancialConnections/Source/Web/FinancialConnectionsWebFlowViewController.swift
```diff
@@ -318,28 +327,33 @@ extension FinancialConnectionsWebFlowViewController {
 
         notifyDelegate(result: .failed(error: error))
     }
+
+    private func logInstantDebitsCompletionFailure(error: Error) {
+        let errorAnalytic = ErrorAnalytic(
+            event: .instantDebitsCompletionFailed,
+            error: error,
+            additionalNonPIIParams: [
+                "flow": "fc_sdk",
```
tillh-stripe: Should we add the FC SDK type?

StripePaymentSheet/StripePaymentSheet/Source/Internal/FC Lite/Controllers/FCLiteContainerViewController.swift
```diff
@@ -86,11 +86,18 @@ class FCLiteContainerViewController: UIViewController {
...
-                } else {
+                do {
+                    if let linkedBank = try createInstantDebitsLinkedBank(from: returnUrl) {
+                        completion(.completed(.instantDebits(linkedBank)))
+                    } else if let linkedAccountId = returnUrl.extractValue(forKey: "linked_account") {
+                        completion(.completed(.linkedAccount(id: linkedAccountId)))
+                    } else {
+                        let error = FCLiteError.linkedBankUnavailable
+                        logInstantDebitsCompletionFailure(error: error)
+                        completion(.failed(error: error))
```
tillh-stripe: Same here.

StripePaymentSheet/StripePaymentSheet/Source/Internal/FC Lite/Controllers/FCLiteContainerViewController.swift
```diff
@@ -247,6 +254,17 @@ class FCLiteContainerViewController: UIViewController {
             await self.completeFlow(result: .cancelled(.cancelledOutsideWebView))
         }
     }
+
+    private func logInstantDebitsCompletionFailure(error: Error) {
+        let errorAnalytic = ErrorAnalytic(
+            event: .instantDebitsCompletionFailed,
+            error: error,
+            additionalNonPIIParams: [
+                "flow": "fc_lite",
```
tillh-stripe, line 262: And same here.

StripeFinancialConnections/StripeFinancialConnections/Source/Web/FinancialConnectionsWebFlowViewController.swift
```diff
@@ -180,16 +180,25 @@ extension FinancialConnectionsWebFlowViewController {
...
+                                    from: returnUrl,
+                                    with: paymentMethod,
+                                    linkAccountSessionId: manifest.id
+                                )
+                                self.notifyDelegateOfSuccess(result: .instantDebits(instantDebitsLinkedBank))
+                            } else if let linkedAccountId = returnUrl.extractValue(forKey: "linked_account") {
+                                self.notifyDelegateOfSuccess(result: .linkedAccount(id: linkedAccountId))
+                            } else {
+                                let error = FinancialConnectionsSheetError.unknown(
+                                    debugDescription: "Invalid payment_method returned"
```
yuki-stripe: Is this description accurate? I think the error is that we expect a payment method or link account ID in the URL and found neither. This also seems like something you should consider alerting on if we ever see this, but I'm not sure you'll be able to distinguish this particular error case right now.

PR #5845, opened 2025-11-24
Captcha token retry

joyceqin-stripe:
## Summary

The captcha token can expire. On the back end, its max_age is set to 1800 seconds, or 30 minutes. We treat the token as expired at 29 minutes to prevent the case where a user confirms at, say, 1799 seconds, and the back end sees a token that has just expired. After expiration, when the user confirms, we fetch a new token.
## Motivation

HCaptcha project
## Testing

Added test
## Changelog

N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ConfirmationChallenge/PassiveCaptchaChallenge.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ConfirmationChallenge/PassiveCaptchaChallenge.swift
@@ -59,8 +59,7 @@ actor PassiveCaptchaChallenge {
         self.init(passiveCaptchaData: passiveCaptchaData, hcaptchaFactory: PassiveHCaptchaFactory())
     }
 
-    init(passiveCaptchaData: PassiveCaptchaData, hcaptchaFactory: HCaptchaFactory,
-         sessionExpiration: TimeInterval = 29 * 60) {
+    init(passiveCaptchaData: PassiveCaptchaData, hcaptchaFactory: HCaptchaFactory, sessionExpiration: TimeInterval = 29 * 60) {
         self.passiveCaptchaData = passiveCaptchaData
         self.hcaptchaFactory = hcaptchaFactory
         // The max_age of the token set on the backend is 1800 seconds, or 30 minutes. As a preventative measure, we expire the token a minute early so a user won't send an expired token


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ConfirmationChallenge/PassiveCaptchaChallenge.swift
```diff
@@ -116,18 +126,34 @@ actor PassiveCaptchaChallenge {
...
 
+    private func isSessionExpired() -> Bool {
+        // The session starts when we get our token
+        guard let sessionStartTime else { return false } // If sessionStartTime is nil, then we haven't gotten our token back yet
+        return Date().timeIntervalSince(sessionStartTime) >= hcaptchaFactory.sessionExpiration
+    }
+
+    private func setSessionStartTime(_ sessionStartTime: Date) {
+        self.sessionStartTime = sessionStartTime
+    }
```
wooj-stripe: Nit: It feels weird that you can pass in a Date() -- feels like it would be better to just have:
```
private func resetSessionStartTime() {
    self.sessionStartTime = Date()
}
```
I don't feel super strongly here, but it might be easier to calculate the expiration time based on the time you receive the token, rather than keeping the "sessionStartTime".

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/ConfirmationChallenge/PassiveCaptchaChallengeTests.swift
```diff
@@ -86,4 +97,34 @@ class PassiveCaptchaChallengeTests: XCTestCase {
...
+        try await Task.sleep(nanoseconds: 5_000_000_000)
+
+        // Check that isTokenReady triggers reset
+        let isReadyAfter = await passiveCaptchaChallenge.isTokenReady
+        XCTAssertFalse(isReadyAfter, "Token should not be ready after session expiration")
+
+        // Fetch a new token after expiration - should succeed with a new token
+        let secondToken = try await passiveCaptchaChallenge.fetchToken()
+        XCTAssertNotNil(secondToken)
+
```
wooj-stripe: Should we compare the firstToken against the secondToken and ensure they are not the same?

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/ConfirmationChallenge/PassiveCaptchaChallengeTests.swift
```diff
@@ -86,4 +97,34 @@ class PassiveCaptchaChallengeTests: XCTestCase {
...
+        // Wait for session to expire
+        try await Task.sleep(nanoseconds: 5_000_000_000)
+
+        // Check that isTokenReady triggers reset
+        let isReadyAfter = await passiveCaptchaChallenge.isTokenReady
+        XCTAssertFalse(isReadyAfter, "Token should not be ready after session expiration")
+
+        // Fetch a new token after expiration - should succeed with a new token
+        let secondToken = try await passiveCaptchaChallenge.fetchToken()
+        XCTAssertNotNil(secondToken)
```
wooj-stripe: Hmm i'm kind of questioning the call model here, rather than the test. If we're having users call isTokenReady() to detect if the token is usable, why not just 'auto fetch' when calling fetchToken(), rather than have users call isTokenReady first (they need to call `fetchToken()` anyways to get the token, right?)

In other words, it doesn't feel like we need to expose `isTokenReady` to users-- seems like we can use it as internal state and update the implementation of fetchToken() to ensure that it always returns a valid token (or throws)

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/ConfirmationChallenge/PassiveCaptchaChallengeTests.swift
joyceqin-stripe: In tests, the token values are always the same

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/ConfirmationChallenge/PassiveCaptchaChallengeTests.swift
joyceqin-stripe: In order to minimize the number of tokens fetched, if the session expires, we don't want to automatically fetch another one. We fetch again when the user attempts to confirm.

You're right that we shouldn't rely on `isTokenReady` to reset anything— I've updated it so that when the session expires, we reset. I still want `isTokenReady` to be exposed just for the `is_ready` field on the attach analytic.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ConfirmationChallenge/PassiveCaptchaChallenge.swift
```diff
@@ -116,18 +118,44 @@ actor PassiveCaptchaChallenge {
...
+    private func scheduleSessionExpirationReset() {
+        let sessionExpiration = hcaptchaFactory.sessionExpiration
+
+        // Cancel any existing expiration reset task
+        expirationResetTask?.cancel()
+
+        // Schedule a task to automatically reset the token when the session expires
+        expirationResetTask = Task { [sessionExpiration] in
+            try? await Task.sleep(nanoseconds: UInt64(sessionExpiration * 1_000_000_000))
+            guard !Task.isCancelled else { return }
```
wooj-stripe: Does using `Task.isCancelled` implicitly refer to `expirationResetTask`?  My brain has a tough time believing that `Task.isCancelled` works this way since it's a static variable -- my expectation would be that we'd need to query `expirationResetTask?.isCanceled` -- In any case, taking a step back..

Given that we now only have a single `public func` of `fetchToken()` it seems that we could just detect whether or not the token is expired when calling fetch token rather than scheduling/managing a task to expire it proactively.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/ConfirmationChallenge/PassiveCaptchaChallengeTests.swift
```diff
@@ -32,36 +32,48 @@ class PassiveCaptchaChallengeTests: XCTestCase {
         super.tearDown()
     }
 
-    struct TestDelayHCaptchaFactory: HCaptchaFactory {
+    struct TestHCaptchaFactory: HCaptchaFactory {
+        let shouldInjectDelay: Bool
```
wooj-stripe: Naming wise, we should probably keep this as "shouldDelayToken" to avoid introducing a new term "inject", then your create code can just be:

```
    hcaptcha.manager.shouldDelayToken = shouldDelayToken
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ConfirmationChallenge/PassiveCaptchaChallenge.swift
```diff
@@ -124,10 +147,16 @@ actor PassiveCaptchaChallenge {
 
 // Protocol for creating HCaptcha instances
 protocol HCaptchaFactory {
+    var sessionExpiration: TimeInterval { get }
```
wooj-stripe: To avoid having to define (29 * 60) in multiple places, I would have made sessionExpiration an injectable parameter on PassiveCaptchaChallenge rather than add it to this factory, but I don't feel super strongly about it.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ConfirmationChallenge/ConfirmationChallenge.swift
```diff
@@ -47,7 +47,7 @@ actor ConfirmationChallenge {
 
     func fetchTokensWithTimeout() async -> ChallengeTokens {
         let startTime = Date()
-        let isReady = await passiveCaptchaChallenge?.hasFetchedToken ?? false
+        let isReady = await passiveCaptchaChallenge?.isTokenReady ?? false
```
wooj-stripe, line 50: IsTokenReady seems to be used for analytics, and we are changing the underlying definition of it -- I'm guessing we're prepared to update any dashboards/queries to filter out older clients with the older definition?

joyceqin-stripe: HCaptcha on MPE is not public yet, so this isn't affecting anything.

wooj-stripe (approved):

PR #5855, opened 2025-11-25
Fix simulator setup script UUID extraction bug

davidme-stripe:
## Problem
The simulator setup script had a critical regex bug that extracted ALL parenthesized content from simulator listings instead of just the UUID. For a simulator like:
```
iPhone 12 mini (Stripe) (33E71AD9-C516-491B-B5E9-F8CADD5EE056) (Shutdown)
```
The regex `([^)]*)` would match "Stripe", the UUID, and "Shutdown", creating a malformed cache file that caused shell errors like:
```
/dev/fd/11:2: command not found: 33E71AD9-C516-491B-B5E9-F8CADD5EE056
Shutdown: NOT super-user
```

## Solution
This PR implements a comprehensive fix with five key improvements:

1. **Fixed UUID extraction** - Changed regex to specifically match UUID format (8-4-4-4-12 hex digits pattern)
2. **Added UUID validation** - Validates extracted UUIDs before caching to catch malformed data
3. **Improved cache format** - Cache file now uses proper shell script format with `export` statement
4. **Fixed sourcing behavior** - Added `exit_or_return()` helper to use `return` when sourced vs `exit` when executed directly
5. **Simplified usage** - Changed from `source <(./script.sh)` to simpler `source ci_scripts/setup_simulator.sh`

## Changes
- `ci_scripts/setup_simulator.sh`: Fixed regex, added validation, improved cache format, fixed sourcing behavior
- `CLAUDE.md`: Updated documentation with simplified usage pattern

## Testing
- ✅ Verified UUID extraction works correctly
- ✅ Verified cache creation and reuse
- ✅ Verified both bash and zsh compatibility
- ✅ Verified simulator ID is valid and usable with xcodebuild
- ✅ Verified clear cache functionality works
- ✅ Passed format and lint checks


ci_scripts/setup_simulator.sh
```diff
@@ -6,6 +6,23 @@
...
+# This affects whether we use return or exit
+if [[ "${BASH_SOURCE[0]}" != "${0}" ]]; then
+    SOURCED=1
+else
+    SOURCED=0
+fi
+
+# Exit or return based on how script was called
+exit_or_return() {
+    if [ "$SOURCED" = "1" ]; then
```
wooj-stripe: i would expect this to be:
```
if ["$SOURCED" == "1"]; then
```

davidme-stripe: I think it's right to use `=` and the extra spacing here when using `[`. (`[` works in any shell, as it's just a executable sitting in `/bin/[`. Here's [the man page](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/test.html) for it.)

Because the script is using bash, the [built-in `[[ expression ]]` syntax]( https://www.gnu.org/software/bash/manual/html_node/Conditional-Constructs.html#index-_005b_005b) would be fine too, e.g. `[["$SOURCED" == "1"]]`. The single-bracket and double-equals `[ "$SOURCED" == "1" ]` is technically incorrect, but bash uses a built-in implementation of `[` and is forgiving enough to allow it.

wooj-stripe: My original comment should have been:
```
if [ "$SOURCED" == "1" ]; then
```

That being said, I thought `==` was the ONLY operator for string comparison, and `=` was reserved for assignment, but seems that `=` can indeed be used for string comparison!!! TIL --oops.

wooj-stripe (approved):

gbirch-stripe:
Partially related: I noticed that when Claude tries to run these shell commands it tends to use `&amp;&amp;` instead of `&&` for the chaining on first try, probably because of some weird pre-processing that happens to the `CLAUDE.md` file. This causes a slight slowdown when it runs them since it fails the first time and then has to self correct. I tinkered with adding an instruction that tells it to avoid this error but that didn't work. One solution could be to keep all full shell commands in `.sh` files so Claude never has to interpret them out of the `.md`

PR #5864, opened 2025-11-25
Update render analytics

joyceqin-stripe:
## Summary

Renamed "mc_lpms_render" to "mc_initial_displayed_payment_methods" to match Android (and fits better given that now we'll include `saved`)
Include `saved` in the list when saved pms shown in vertical mode
Update calculations in horizontal and vertical screens to only count something as visible if it is _fully visible_ on the screen
Add `payment_method_layout` parameter
## Motivation

https://jira.corp.stripe.com/browse/RUN_MOBILESDK-4879
## Testing

Updated test
## Changelog

N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/PaymentSheetVerticalViewController.swift
```diff
@@ -577,22 +577,39 @@ class PaymentSheetVerticalViewController: UIViewController, FlowControllerViewCo
...
+        var visiblePaymentMethods: [String] = []
+        var hiddenPaymentMethods: [String] = []
+        if PaymentSheet.isApplePayEnabled(elementsSession: elementsSession, configuration: configuration), !shouldShowApplePayInList {
+            visiblePaymentMethods.append(RowButtonType.applePay.analyticsIdentifier)
         }
-        if PaymentSheet.isLinkEnabled(elementsSession: elementsSession, configuration: configuration) {
-            visibleLPMs.append(RowButtonType.link.analyticsIdentifier)
+        if PaymentSheet.isLinkEnabled(elementsSession: elementsSession, configuration: configuration), !shouldShowLinkInList {
+            visiblePaymentMethods.append(RowButtonType.link.analyticsIdentifier)
         }
```
wooj-stripe, line 595: This logic seems to be for the express buttons (e.g. PS.Complete & Vertical)..;

Maybe the code might be more clear like this:
```
let isShowingApplePayAsExpressButton = PaymentSheet.isApplePayEnabled(elementsSession: elementsSession, configuration: configuration), !shouldShowApplePayInList
if isShowingApplePayAsExpressButton {
    visiblePaymentMethods.append(RowButtonType.applePay.analyticsIdentifier)
}
```
or we add a comment explaining that we're logging apple_pay if it appears as an express button, or maybe added later as a button (e.g. flow controller)

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/PaymentSheetVerticalViewController.swift
```diff
@@ -577,22 +577,39 @@ class PaymentSheetVerticalViewController: UIViewController, FlowControllerViewCo
...
+                let rowButtonFrame = rowButton.convert(rowButton.bounds, to: scrollView)
+                let isFullyVisible = rowButtonFrame.minY >= visibleMinY && rowButtonFrame.maxY <= visibleMaxY
+                let identifier = rowButton.type.analyticsIdentifier
+                if isFullyVisible {
+                    visiblePaymentMethods.append(identifier)
+                } else {
+                    hiddenPaymentMethods.append(identifier)
+                }
+            }
+        }
```
wooj-stripe, line 603: Instead of using the scrollview's offset. I think this may be easier?
```
            for rowButton in rowButtons {
                let identifier = rowButton.type.analyticsIdentifier
                if rowButton.isEffectivelyVisibleOnScreen {
                    print("visible: \(identifier)")
                } else {
                    print("not visible: \(identifier)")
                }
            }

...

extension UIView {
  var isEffectivelyVisibleOnScreen: Bool {
    guard let window = window, !isHidden, alpha > 0.01 else { return false }
    let viewFrameInWindow = convert(bounds, to: window)
    return viewFrameInWindow.intersects(window.bounds)
  }
}

```

ensures the view actually has a window, it isn't hidden, and the alpha value is value -- then converts to the actual window to ensure it's on screen.  I suspect this approach might make the code for horizontal easier to reason about as well.

PR #5869, opened 2025-11-27
Add secure webview implementation of FC Lite for Instant Debits

mats-stripe:
Related backend PRs: [#1275019](https://git.corp.stripe.com/stripe-internal/pay-server/pull/1275019) and [#1280280](https://git.corp.stripe.com/stripe-internal/pay-server/pull/1280280)

## Summary

This updates FC Lite to use a secure webview instead of a `WKWebView` for unverified merchants using Instant Debit. The rational behind this is discussed in the Compass project / doc below. Note that 

## Motivation

[Compass](https://home.corp.stripe.com/compass/projects/instant-debits-mobile-expansion)

[Doc](https://docs.google.com/document/d/1ouIrcJWZWKGf105JSoRP80UgcgQGJ5vRGH64K3H5wBo/edit?usp=sharing)

## Testing

FCLiteUITests cover these changes. Here is the new instant debits flow, including a OAuth redirect:

https://github.com/user-attachments/assets/53ebc587-a38d-495a-8381-e8f4766971cb

Here is the FC Lite instant debits flow verified merchants would see (using the `WKWebView`):

https://github.com/user-attachments/assets/f6aa3042-ddec-4546-aef0-1c802c974384

And the ACH flow still works with the WKWebView implementation:

https://github.com/user-attachments/assets/f3ea6026-f603-4476-91c2-c27416dfee30

## Changelog

N/a

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheetUITest/FCLiteUITests.swift
+++ b/Example/PaymentSheet Example/PaymentSheetUITest/FCLiteUITests.swift
@@ -124,16 +124,13 @@ class FCLiteUITests: XCTestCase {
         // Continue with Link
         let continueWithLinkButtonPredicate = NSPredicate(format: "label CONTAINS[cd] 'Continue with Link'") // Link signup pane
         let continueWithLinkButton = app.webViews.firstMatch.buttons.containing(continueWithLinkButtonPredicate).firstMatch
-        XCTAssertTrue(continueWithLinkButton.waitForExistence(timeout: 5.0))
-        tapPrimaryButton()
+        XCTAssertTrue(continueWithLinkButton.waitForExistence(timeout: 10.0))
+        app.typeText(XCUIKeyboardKey.return.rawValue) // Enter key will continue to the next screen
 
-        // Payment Success bank
-        let paymentSuccessBankButtonPredicate = NSPredicate(format: "label CONTAINS[cd] 'Payment Success'") // Institution Picker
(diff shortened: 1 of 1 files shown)


StripePaymentSheet/StripePaymentSheet/Source/Internal/FC Lite/Controllers/FCLiteSecureAuthFlowViewController.swift
```diff
@@ -0,0 +1,128 @@
...
+            url: hostedAuthUrl,
+            callbackURLScheme: callbackScheme,
+            completionHandler: { [weak self] callbackUrl, error in
+                guard let self else {
+                    return
+                }
+
+                DispatchQueue.main.async {
+                    if let error = error {
+                        if let authenticationSessionError = error as? ASWebAuthenticationSessionError {
```
tillh-stripe, line 63: Nit:

```suggestion
                    if let authenticationSessionError = error as? ASWebAuthenticationSessionError {
```

mats-stripe: It can't be flattened because we want to also call the completion handler for non-ASWebAuthenticationSessionError errors. The nested `if let authenticationSessionError...` is to add special handling for the `cancelledWithinWebview` case

PR #5875, opened 2025-12-02
Add funding filtering API

porter-stripe:
## Summary
- Adds funding filtering API
- Adds playground setting

## Motivation
- https://docs.google.com/document/d/19cVJ7dkA7dOxu5KIwOzLNfxXmCWmXFtUBwBf9_cpDDg/edit?tab=t.0#heading=h.570u4leviueu

## Testing
- Manual

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
@@ -985,7 +985,7 @@ extension PaymentSheet {
 }
 
 extension PaymentSheet {
-    /// Card funding categories that can be filtered
+    /// Card funding types that can be filtered
     @_spi(CardFundingFilteringPrivatePreview) public enum CardFundingType: Equatable {
         /// Debit cards
         case debit
@@ -993,7 +993,8 @@ extension PaymentSheet {
         case credit
         /// Prepaid cards
         case prepaid
-        /// Unknown funding type
+        /// Unknown or undetermined funding type.
+        /// Include this if you want to accept cards where the funding type cannot be determined from card metadata.
         case unknown
     }
 


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElementConfiguration.swift
```diff
@@ -137,6 +137,10 @@ extension EmbeddedPaymentElement {
         /// Note: Card brand filtering is not currently supported by Link.
         public var cardBrandAcceptance: PaymentSheet.CardBrandAcceptance = .all
 
+        /// By default, the embedded payment element will accept cards of all funding types (credit, debit, prepaid, unknown).
+        /// You can specify which card funding types to allow.
+        @_spi(CardFundingFilteringPrivatePreview) public var allowedCardFundingTypes: PaymentSheet.CardFundingAcceptance = .all
```
wooj-stripe, line 142: We can chat about naming in the API review, but `allowed` feels too strongly given that the actual prevention of using these card types happens somewhere else, as this is purely a client side check.

Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
```diff
@@ -368,6 +377,15 @@ class PlaygroundController: ObservableObject {
             configuration.cardBrandAcceptance = .allowed(brands: [.visa])
         }
 
+        switch settings.cardFundingAcceptance {
+        case .all:
+            configuration.allowedCardFundingTypes = .all
+        case .creditOnly:
+            configuration.allowedCardFundingTypes = .allowed(fundingTypes: [.credit])
+        case .debitOnly:
+            configuration.allowedCardFundingTypes = .allowed(fundingTypes: [.debit])
```
wooj-stripe, line 386: We should probably take this offline, but I'm curious what happens if the metadata service fails to return a value, when a merchant specifies:
```
 configuration.allowedCardFundingTypes = .allowed(fundingTypes: [.credit])
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
```diff
@@ -979,3 +983,25 @@ extension PaymentSheet {
...
+    /// Card funding categories that can be filtered
+    @_spi(CardFundingFilteringPrivatePreview) public enum CardFundingType: Equatable {
+        /// Debit cards
+        case debit
+        /// Credit cards
+        case credit
+        /// Prepaid cards
+        case prepaid
+        /// Unknown funding type
+        case unknown
```
wooj-stripe: Kind of related to the comment above, but I think a better comment here is deserved -- Why would a user want to use this value?  Because the metadata service actually return `.unknown`, or are we also using this when the service times out/client/service error?

Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
porter-stripe: Yeah, we can talk about the failure more. The current aligned-on behavior is that we will not block checkout based on the metadata service funding type and will rather issue a warning that the user can still check out. Then the merchant should do a final check at confirm time of the funding type.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElementConfiguration.swift
porter-stripe: Yeah, this is what we aligned on in the API review: https://docs.google.com/document/d/19cVJ7dkA7dOxu5KIwOzLNfxXmCWmXFtUBwBf9_cpDDg/edit?disco=AAABwtHwJ5o

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
porter-stripe: Good point, let me improve this!

wooj-stripe: I think the comment is still vague, but we can handle this in API review. shipping for now to unblock development.

PR #5877, opened 2025-12-02
Filter funding types in Apple Pay and Saved payment methods

porter-stripe:
## Summary
- Forward funding filtering options to Apple Pay
- Filter saved payment methods based on the funding types allowed
- Not filtering in CustomerSheet is out of scope for private preview

Apple Pay filtering UX:
IMG_3498
IMG_3499

## Motivation
- https://docs.google.com/document/d/1eihnXys1Cx3WpSwb5xnj24ysi8_oak2aDR1Zr8eOgrI/edit?tab=t.0#heading=h.cz1xkpga7giy

## Testing
- Manual
- New unit tests

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/STPApplePayContext+PaymentSheetTest.swift
+++ b/StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/STPApplePayContext+PaymentSheetTest.swift
@@ -179,6 +179,7 @@ final class STPApplePayContext_PaymentSheetTest: XCTestCase {
         configuration.allowedCardFundingTypes = .all
         let intent = Intent._testValue()
         let sut = STPApplePayContext.createPaymentRequest(intent: intent, configuration: configuration, applePay: applePayConfiguration)
+        // When all funding types are accepted, the default merchant capabilities from StripeAPI.paymentRequest are preserved (.capability3DS)
         XCTAssertEqual(sut.merchantCapabilities, .capability3DS)
     }
 


porter-stripe:
cc @toluo-stripe

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CardFundingFilter.swift
```diff
@@ -0,0 +1,73 @@
...
+            return allowedFundingTypes.contains(fundingCategory)
+        }
+    }
+
+    /// Returns the `PKMerchantCapability` to use for Apple Pay based on the card funding acceptance configuration.
+    /// - Returns: A `PKMerchantCapability` option set that includes 3DS and any funding type restrictions
+    func applePayMerchantCapabilities() -> PKMerchantCapability {
+        switch cardFundingAcceptance {
+        case .all:
+            return .capability3DS
```
wooj-stripe: Given your test "testApplePayMerchantCapabilities_all" -- Seems like i'm missing something because, i would expect that `.all` should return capabilities including `.capabilityDebit`, `.capabilityCredit`, and `.capability3DS` ?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift
```diff
@@ -438,10 +438,21 @@ final class PaymentSheetLoader {
...
         return savedPaymentMethods.filter {
-            guard let cardBrand = $0.card?.preferredDisplayBrand else { return true }
-            return configuration.cardBrandFilter.isAccepted(cardBrand: cardBrand)
+            guard let card = $0.card else { return true }
+            // Filter by card brand
+            if !configuration.cardBrandFilter.isAccepted(cardBrand: card.preferredDisplayBrand) {
+                return false
+            }
+            // Filter by card funding type
+            if let fundingString = card.funding {
```
wooj-stripe: If `card.funding` is nil -- does that fall into the`.unknown` case? Is it really okay to return `true` if `.funding` is nil?

porter-stripe: This is a good point! I think it depends on if the merchant has specified `unknown` in the funding allowed list. I'll update this to reflect that.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CardFundingFilter.swift
porter-stripe: Yeah this is correct, although maybe `all` in this case seems misleading as it's essentially "default".  By default we only specify 3DS on all our Apple Pay requests, so that's where we got it: https://github.com/stripe/stripe-ios/blob/c1816624243907061efb367c450adb2a130870f5/StripeCore/StripeCore/Source/API%20Bindings/StripeAPI.swift#L170 

This is a bit confusing though, so I refactored this so instead it returns nil for "all" and we just use whatever is already on the PKPaymentRequest.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CardFundingFilter.swift
```diff
@@ -0,0 +1,76 @@
...
+
+    /// Returns the `PKMerchantCapability` to use for Apple Pay based on the card funding acceptance configuration.
+    /// - Returns: A `PKMerchantCapability` option set that includes 3DS and any funding type restrictions,
+    ///            or `nil` if no override is needed (use the default capabilities provided on the payment request).
+    func applePayMerchantCapabilities() -> PKMerchantCapability? {
+        switch cardFundingAcceptance {
+        case .all:
+            // When all funding types are accepted, don't override the merchant capabilities.
+            // The default capabilities on the payment request will be used.
+            return nil
```
wooj-stripe, line 41: nice!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift
```diff
@@ -438,10 +438,20 @@ final class PaymentSheetLoader {
...
+            if !configuration.cardBrandFilter.isAccepted(cardBrand: card.preferredDisplayBrand) {
+                return false
+            }
+            // Filter by card funding type
+            // If funding is nil, treat it as .other (unknown) and check if that's accepted
+            let fundingType: STPCardFundingType = card.funding.map { STPCard.funding(from: $0) } ?? .other
+            if !configuration.cardFundingFilter.isAccepted(cardFundingType: fundingType) {
+                return false
+            }
+            return true
```
wooj-stripe, line 454: nit: could be:
```
return configuration.cardFundingFilter.isAccepted(cardFundingType: fundingType) 
```

porter-stripe: will add in the next PR!

PR #5878, opened 2025-12-03
Speed up checkout action in dead code check

cttsai-stripe:
## Summary

Update the action version and remove `fetch-depth: 0` to speed up the action.
Reducing the checkout time from averaging ~2m to under 20s.
https://github.com/stripe/stripe-ios/actions/runs/19905920015/job/57061959041

Starting from [checkout-v4](https://github.com/actions/checkout?tab=readme-ov-file#checkout-v4):
> Only a single commit is fetched by default, for the ref/SHA that triggered the workflow. Set fetch-depth: 0 to fetch all history for all branches and tags.

## Motivation

In the worst scenario, it takes 9 minutes to checkout.
https://stripe.slack.com/archives/C02HQBW38JX/p1764783011920219

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -175,4 +175,3 @@ let package = Package(
         )
     ]
 )
-


.github/workflows/find-dead-code.yml
```diff
@@ -13,9 +13,7 @@ jobs:
 
     steps:
       - name: Checkout code
-        uses: actions/checkout@v2
-        with:
-          fetch-depth: 0
+        uses: actions/checkout@v6
```
porter-stripe, line 16: I think we do this on purpose so that we can compare with merge base. I think if you modified some swift code in this PR to get the check to run it might fail.

cttsai-stripe: https://github.com/stripe/stripe-ios/actions/runs/19905920015/job/57061959041 is the check that passed on a commit with swift code change. I uncommitted that change when ready for review. But if `fetch-depth` is necessary, I can close this PR and keep the issue open for the next iOS runner to investigate it.

PR #5880, opened 2025-12-03
Filtering funding type in card field

porter-stripe:
## Summary
- When a user types in at least 6 PAN digits we now fetch the funding type from the card metadata service. 
- We added a warningLabel to the TextFieldElement to allow showing a red warning message that is technically not a blocking error. Users can still checkout by tapping the CTA.

### Demo
https://github.com/user-attachments/assets/62c5f154-c905-4d97-9d44-9c8c333cfe12

## Motivation
- Card funding filtering

## Testing
- Manual
- New unit tests

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/CardSection/CardSectionElement.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/CardSection/CardSectionElement.swift
@@ -188,8 +188,6 @@ final class CardSectionElement: ContainerElement {
     var lastPanElementValidationState: ElementValidationState
     var lastDisallowedCardBrandLogged: STPCardBrand?
     var hasLoggedExpectedExtraDigitsButUserEntered16: Bool = false
-    /// Prevents infinite loop since the completion calls `setText` which triggers `didUpdate`.
-    private var lastFetchedFundingPrefix: String?
     func didUpdate(element: Element) {
         // Update the CVC field if the card brand changes
         let cardBrand = selectedBrand ?? STPCardValidator.brand(forNumber: panElement.text)
@@ -247,8 +245,7 @@ final class CardSectionElement: ContainerElement {
             return
         }
         let binPrefix = String(panElement.text.prefix(6))
-        guard panElement.text.count >= 6,
-              binPrefix != lastFetchedFundingPrefix else {
+        guard panElement.text.count >= 6 else {
             return
         }
 
@@ -258,9 +255,8 @@ final class CardSectionElement: ContainerElement {
             onlyFetchForVariableLengthBINs: false
         ) { [weak self] _ in
             guard let self = self else { return }
-            self.lastFetchedFundingPrefix = binPrefix
             // Trigger re-validation so warningLabel can read the now-cached funding data
-            self.panElement.setText(self.panElement.text)
+            delegate?.didUpdate(element: self)
         }
     }
 


porter-stripe:
Added the "ship without translations" as this feature will not deliver until early next year and we will have translations by then.

porter-stripe:
cc @toluo-stripe

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/CardSection/CardSectionElement.swift
```diff
@@ -229,6 +239,31 @@ final class CardSectionElement: ContainerElement {
...
+    func fetchAndCacheCardFunding() {
+        guard cardFundingFilter != .default else {
+            return
+        }
+        let binPrefix = String(panElement.text.prefix(6))
+        guard panElement.text.count >= 6,
+              binPrefix != lastFetchedFundingPrefix else {
+            return
+        }
+        lastFetchedFundingPrefix = binPrefix
```
wooj-stripe: seems like we should set `lastFetchedFundingPrefix` in the completion of calling `retrieveBINRanges` ?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/TextField/TextFieldElement+Card.swift
```diff
@@ -24,12 +24,20 @@ extension TextFieldElement {
...
         let cardFilter: CardBrandFilter
-
-        init(defaultValue: String? = nil, cardBrand: STPCardBrand? = nil, cardBrandDropDown: DropdownFieldElement? = nil, cardFilter: CardBrandFilter = .default) {
+        let cardFundingFilter: CardFundingFilter
+
+        init(
+            defaultValue: String? = nil,
+            cardBrand: STPCardBrand? = nil,
+            cardBrandDropDown: DropdownFieldElement? = nil,
+            cardFilter: CardBrandFilter = .default,
```
wooj-stripe: maybe we change `cardFilter` to `cardBrandFilter` ?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CardFundingFilter.swift
```diff
@@ -50,6 +53,31 @@ struct CardFundingFilter: Equatable {
...
+                // E.g. "debit"
+                return displayNames.first
+            } else if displayNames.count == 2 {
+                // E.g. "debit and pre-paid"
+                return String.Localized.x_and_y(displayNames[0], displayNames[1])
+            } else {
+                // For 3+ items: "credit, debit, and pre-paid"
+                let allButLast = displayNames.dropLast().joined(separator: ", ")
+                return String.Localized.x_and_y(allButLast, displayNames.last ?? "")
+            }
```
wooj-stripe: Super nit (feel free to ignore if you don't find this easier to read), but we could avoid the last nil coalesce, and indexing directly into displayNames, if we do:
```
            guard !displayNames.isEmpty,
                  let displayNamesFirst = displayNames.first,
                  let displayNamesLast = displayNames.last else { return nil }

            // Join with localized "and" for the last element
            if displayNames.count == 1 {
                // E.g. "debit"
                return displayNamesFirst
            } else if displayNames.count == 2 {
                // E.g. "debit and pre-paid"
                return String.Localized.x_and_y(displayNamesFirst, displayNamesLast)
            } else {
                // For 3+ items: "credit, debit, and pre-paid"
                let allButLast = displayNames.dropLast().joined(separator: ", ")
                return String.Localized.x_and_y(allButLast, displayNamesLast)
            }
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/TextField/TextFieldElement+Card.swift
porter-stripe: Yeah that is much clearer!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/CardSection/CardSectionElement.swift
porter-stripe: Makes sense!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CardFundingFilter.swift
porter-stripe: I like it!

StripeUICore/StripeUICore/Source/Elements/Element.swift
```diff
@@ -32,6 +32,9 @@ import UIKit
     /// Text to display to the user under the item, if any.
     var subLabelText: String? { get }
 
+    /// Warning text to display to the user under the item, if any. Displayed in danger color.
+    var warningLabelText: String? { get }
+
```
yuki-stripe, line 37: How does this interact with `subLabelText` if both are supplied? 
If the element is both invalid (validationState has an error) and has a warning label, what happens? e.g. if I have an invalid PAN and its BIN matches a filtered card type?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/CardSection/CardSectionElement.swift
```diff
@@ -181,6 +188,8 @@ final class CardSectionElement: ContainerElement {
     var lastPanElementValidationState: ElementValidationState
     var lastDisallowedCardBrandLogged: STPCardBrand?
     var hasLoggedExpectedExtraDigitsButUserEntered16: Bool = false
+    /// Tracks the last BIN prefix we fetched funding info for to avoid redundant calls
+    private var lastFetchedFundingPrefix: String?
```
yuki-stripe: STPBinController should already cache, does that not work for some reason?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/CardSection/CardSectionElement.swift
```diff
@@ -229,6 +239,31 @@ final class CardSectionElement: ContainerElement {
...
     }
 
+    // MARK: - Card funding check
+
+    /// Fetches BIN metadata from the card metadata service and caches it in `STPBINController`.
+    func fetchAndCacheCardFunding() {
+        guard cardFundingFilter != .default else {
+            return
+        }
+        let binPrefix = String(panElement.text.prefix(6))
```
yuki-stripe, line 247: where'd this logic come from? does the BIN endpoint need exactly 6 and no more?

porter-stripe: Yeah, that's right, we actually enforce it client-side too:

https://github.com/stripe/stripe-ios/blob/013dcf6934384f760f72941e4c0a9fc8bdb2dcfc/StripePayments/StripePayments/Source/Helpers/STPBINController.swift#L143

And on the backend:
https://stripe.sourcegraphcloud.com/stripe-internal/gocode/-/blob/gopiori/ingress/pkg/card_metadata/handler.go?L102-L113

StripeUICore/StripeUICore/Source/Elements/Element.swift
porter-stripe: Good question! The if statement that sets the `errorOrSubLabel` kinda has some priority baked in, probably not perfect but I think it fits our use case as we never show stacked sub labels under a field AFAIK. Lmk what you think.

error > warning > sublabel.

This shows the UX:


https://github.com/user-attachments/assets/8f1e4de7-c1f6-4590-98be-6e410b046043

yuki-stripe: TBD how it extends to the future but we can refactor at that point if needed. I would just call out the behavior / priority so that future-us knows that the ordering is intentional. Otherwise one day when we go try to build some other slightly different UX, we might not know the current behavior is intentional.

porter-stripe: Added a comment!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/CardSection/CardSectionElement.swift
porter-stripe: STPBINController caches but still invokes the completion handler when returning cached data. Since our completion calls setText -> didUpdate -> fetchAndCacheCardFunding, we'd end up in an infinite loop without this check. lastFetchedFundingPrefix skips the call entirely for the same prefix, avoiding the loop.

porter-stripe: I improved the comment here

yuki-stripe: Why do you need to `setText`? Does calling `didUpdate` not work?

porter-stripe: Ah good idea, looks like I can call `delegate?.didUpdate(element: self)` and we avoid the loop while getting the updated sub label!

PR #5882, opened 2025-12-04
Update README SDK requirements to Xcode 16

yuki-stripe:
.


gbirch-stripe:
cc @davidme-stripe iirc we sort of did a soft support drop for Xcode 15, but I think it makes sense to communicate that in the readme

PR #5883, opened 2025-12-04
Add check-scanning to StripeConnectExample

bredmond-stripe:
## Summary

Add check-scanning to StripeConnectExample

## Motivation

Updating StripeConnectExample now that check-scanning is in preview

## Testing

Simulator

### DIFF
--- a/Example/StripeConnectExample/StripeConnectExample/MainViewController.swift
+++ b/Example/StripeConnectExample/StripeConnectExample/MainViewController.swift
@@ -5,7 +5,7 @@
 //  Created by Mel Ludowise on 4/30/24.
 //
 
-@_spi(PreviewConnect) import StripeConnect
+@_spi(PrivatePreviewConnect) @_spi(PreviewConnect) import StripeConnect
 import SwiftUI
 import UIKit
 
@@ -29,8 +29,7 @@ class MainViewController: UITableViewController {
         case onboarding = "Account onboarding"
         case payouts = "Payouts"
         case payments = "Payments"
-        // TODO: add this once in preview
-        // case checkScanning = "Check scanning"
+        case checkScanning = "Check scanning"
 
         var label: String { rawValue }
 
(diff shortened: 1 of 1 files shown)


Example/StripeConnectExample/StripeConnectExample/MainViewController.swift
```diff
@@ -308,6 +318,30 @@ extension MainViewController: PaymentsViewControllerDelegate {
...
 
+// MARK: - CheckScanningControllerDelegate
+
+extension MainViewController: CheckScanningControllerDelegate {
+    func checkScanning(_ checkScanning: CheckScanningController, didFailLoadWithError error: Error) {
+        presentAlert(title: "Error loading check scanning", message: (error as NSError).debugDescription)
+    }
+
+    func checkScanning(_ checkScanning: CheckScanningController, didSubmitCheckScan: CheckScanningController.CheckScanDetails) async throws {
+        await MainActor.run {
```
yuki-stripe, line 329: ooc did the compiler complain without this?

Example/StripeConnectExample/StripeConnectExample/MainViewController.swift
```diff
@@ -308,6 +318,30 @@ extension MainViewController: PaymentsViewControllerDelegate {
...
+            checkScanning.dismiss(animated: false)
```
yuki-stripe, line 330: you should probably wait until the dismissal is finished before presenting/doing other UI stuff - add a completion block here and put the rest of the code inside it.

Example/StripeConnectExample/StripeConnectExample/MainViewController.swift
bredmond-stripe: I got a runtime error. I can't actually repro right now but I swear it was something along the lines of "UI updates must be on main thread".

Example/StripeConnectExample/StripeConnectExample/MainViewController.swift
bredmond-stripe: How is that expressed? There's no parameter for that in `dismiss`, but the `animated: false` should make it disappear quick. Do we need to change the interface here?

Example/StripeConnectExample/StripeConnectExample/MainViewController.swift
bredmond-stripe: from offline discussion: we'll follow up separately to address this, possibly calling this delegate method from the main thread

Example/StripeConnectExample/StripeConnectExample/MainViewController.swift
bredmond-stripe: from offline discussion: we'll follow up separately to address this, likely adding a completion handler parameter which will just forward to the inner `dismiss`

PR #5886, opened 2025-12-05
Add completion handler to CheckScanningController.dismiss

bredmond-stripe:
## Summary

Add completion handler to CheckScanningController.dismiss

Followup to https://github.com/stripe/stripe-ios/pull/5883

## Motivation

Final batch of design changes to the interface here:
 * Add completion handler to `dismiss`
 * Ensure delegate methods are run on the main thread

## Testing

simulator

### DIFF
--- a/StripeConnect/StripeConnect/Source/Components/CheckScanningController.swift
+++ b/StripeConnect/StripeConnect/Source/Components/CheckScanningController.swift
@@ -125,7 +125,7 @@ public final class CheckScanningController {
 
 @_spi(PrivatePreviewConnect)
 @available(iOS 15, *)
-@MainActor
+@preconcurrency @MainActor
 public protocol CheckScanningControllerDelegate: AnyObject {
     /// Called when the component fails to load (e.g., network issue during initial fetch). To try again, initialize a new CheckScanningController.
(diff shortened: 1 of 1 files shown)


Example/StripeConnectExample/StripeConnectExample/MainViewController.swift
```diff
@@ -326,9 +326,7 @@ extension MainViewController: CheckScanningControllerDelegate {
     }
 
     func checkScanning(_ checkScanning: CheckScanningController, didSubmitCheckScan: CheckScanningController.CheckScanDetails) async throws {
-        await MainActor.run {
-            checkScanning.dismiss(animated: false)
-
+        checkScanning.dismiss(animated: true) {
```
davidme-stripe, line 329: Cool, so this will always run on the main actor now?

davidme-stripe (approved):

Example/StripeConnectExample/StripeConnectExample/MainViewController.swift
bredmond-stripe: yup, was chatting with @yuki-stripe about this and we're doing a little addendum here

PR #5887, opened 2025-12-05
Update Card Funding API to match API review

porter-stripe:
## Summary
- Updates the card funding filtering API to match the result of the API review

## Motivation
- https://docs.google.com/document/d/19cVJ7dkA7dOxu5KIwOzLNfxXmCWmXFtUBwBf9_cpDDg/edit?tab=t.0#heading=h.570u4leviueu

## Testing
- CI

## Changelog
N/A

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
@@ -255,7 +255,7 @@ class PlaygroundController: ObservableObject {
 
         switch settings.cardFundingAcceptance {
         case .all:
-            configuration.allowedCardFundingTypes = .all
+            break // Default accepts all funding types
         case .creditOnly:
             configuration.allowedCardFundingTypes = .credit
         case .debitOnly:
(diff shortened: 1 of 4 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CardFundingFilter.swift
```diff
@@ -12,78 +12,85 @@ import PassKit
...
-    static let `default`: CardFundingFilter = .init(cardFundingAcceptance: .all)
+    static let `default`: CardFundingFilter = .init(allowedFundingTypes: .all)
 
-    private let cardFundingAcceptance: PaymentSheet.CardFundingAcceptance
+    private let allowedFundingTypes: PaymentSheet.CardFundingType
 
-    init(cardFundingAcceptance: PaymentSheet.CardFundingAcceptance) {
-        self.cardFundingAcceptance = cardFundingAcceptance
+    init(allowedFundingTypes: PaymentSheet.CardFundingType) {
+        self.allowedFundingTypes = allowedFundingTypes
```
wooj-stripe, line 20: We can now instantiate this class with:
```
 .init(allowedFundingTypes: [.all, .credit])
```

Where before, `.all` and `.credit` were separated in the the way we pass it in to init.

Is there an option where we just get rid of `.all` (so then no-existance of `allowedFundingTypes` implies all, or maybe we just don't make `.all` public?)

porter-stripe: Good catch, I like the read-ability of "all" in the API spec but we should make it internal!

PR #5891, opened 2025-12-08
Deflake IBP native test

mats-stripe:
## Summary

Deflakes the `testNativeInstantDebitsOnlyLinkPaymentController` which was unexpectedly not using the native flow https://app.bitrise.io/build/3b0c8c0e-105b-4154-85d3-1ca1b6b51063?tab=tests&tests_filter_status=failed

These changes should force the native flow in that test.

## Motivation

Keeping master green!

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
+++ b/Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
@@ -3692,6 +3692,7 @@ extension PaymentSheetUITestCase {
         app.pickerWheels.firstMatch.adjust(toPickerWheelValue: "🇺🇸 United States (+1)")
         app.toolbars.buttons["Done"].tap()
 
+        sleep(1) // Wait for keyboard to dismiss
         phoneTextField.tap()
         phoneTextField.typeText("4015006000")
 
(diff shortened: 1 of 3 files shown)


Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetUITest.swift
```diff
@@ -3692,6 +3692,7 @@ extension PaymentSheetUITestCase {
         app.pickerWheels.firstMatch.adjust(toPickerWheelValue: "🇺🇸 United States (+1)")
         app.toolbars.buttons["Done"].tap()
 
+        sleep(1) // Wait for keyboard to dismiss
```
gbirch-stripe, line 3695: Is it possible to do a waitForExistenceAndTap or something here? Waiting a full second for this minor detail seems pretty costly

mats-stripe: The button is already in frame at this point, but since the keyboard is still on the screen at the time this is tapped (if there's no delay), the tap isn't detected. Here's a full video. This points is at the 0:16 mark: 

https://github.com/user-attachments/assets/fa5262bf-12ce-469b-b5c7-7e431eeeaf36

PR #5898, opened 2025-12-10
Set API key in funding test

porter-stripe:
## Summary
- Fixes nightly build

### DIFF
--- a/StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/TextFieldElement+CardTest.swift
+++ b/StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/TextFieldElement+CardTest.swift
@@ -398,7 +398,9 @@ class TextFieldElementCardTest: STPNetworkStubbingTestCase {
     }
 
     func testCardSectionElement_cardFundingFiltering_showsWarningButRemainsValid() {
+        let originalPublishableKey = STPAPIClient.shared.publishableKey
         STPAPIClient.shared.publishableKey = STPTestingDefaultPublishableKey
+        defer { STPAPIClient.shared.publishableKey = originalPublishableKey }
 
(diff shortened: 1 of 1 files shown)


StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/TextFieldElement+CardTest.swift
```diff
@@ -398,6 +398,10 @@ class TextFieldElementCardTest: STPNetworkStubbingTestCase {
     }
 
     func testCardSectionElement_cardFundingFiltering_showsWarningButRemainsValid() {
+        let originalPublishableKey = STPAPIClient.shared.publishableKey
+        STPAPIClient.shared.publishableKey = STPTestingDefaultPublishableKey
+        defer { STPAPIClient.shared.publishableKey = originalPublishableKey }
```
wooj-stripe: I notice other tests in this file just set:
```
        STPAPIClient.shared.publishableKey = STPTestingDefaultPublishableKey
```
without a resetting it back to whatever the original is (e.g. testAlertsOnExpected19DigitCard) .  I'm not sure if the `defer` block is really needed since it has been working that way in other tests without the defer block, but if it is needed, then seems like we should be consistent throughout the file.

porter-stripe: I added it due to weird behavior I was seeing here: https://stripe.slack.com/archives/C02HQBW38JX/p1764872640875809

But let me try without

porter-stripe: Yep failed without gonna add it back.

PR #5908, opened 2025-12-13
Host app minimum iOS version in analytics

gbirch-stripe:
## Summary
Record the minimum iOS version of the host app in analytics

## Motivation
Gather data about what minimum iOS versions users are using


gbirch-stripe:
@davidme-stripe do you remember if there was a particular reason we didn't use this method back when we originally talked about it? I thought I remembered just having trouble making it work.... though it seems so simple I'm doubting my memory.
This seems to work fine for the example app

davidme-stripe:
Huh, no, this seems like a good idea! Let's try it!

davidme-stripe (approved):

PR #5916, opened 2025-12-16
Add pinning/auto-open feature to example app

gbirch-stripe:
## Summary
Add "pin" button to Payment Sheet Example app pages. Pinning a view causes it to be automatically opened when the app loads. Propagated via UserDefaults

## Motivation
Remove developer friction
https://jira.corp.stripe.com/browse/MOBILESDK-4287

## Testing
https://github.com/user-attachments/assets/a3246dff-94dd-42da-a04f-10a4859d86f8

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetExampleAppRootView.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetExampleAppRootView.swift
@@ -29,8 +29,6 @@ struct PaymentSheetExampleAppRootView: View {
                     }
                 }
             }
-            .navigationTitle("Examples")
-            .navigationBarTitleDisplayMode(.inline)
         }
         .onAppear {
             // Load pinned destination from UserDefaults and auto-open it


Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetExampleAppRootView.swift
```diff
@@ -12,33 +12,70 @@ struct PaymentSheetExampleAppRootView: View {
...
-                            }
-                            .accessibility(identifier: destination.displayTitle)
+                            navigationLink(for: destination)
                         }
                     }
                 }
             }
         }
-        .navigationTitle("Examples")
-        .navigationBarTitleDisplayMode(.inline)
```
gbirch-stripe, line 33: These were actually unused since they're on the wrong layer (the `NavigationView` itself instead of its contents, the `Form`). I tried putting them where they're supposed to go but I personally didn't like having the header bar and having the back buttons in the pages say "Examples" instead of "Back," so I removed them.

PR #5927, opened 2025-12-17
Fix deferred playground reused intent bug

joyceqin-stripe:
## Summary

Create a new intent before `intentCreationCallback` is invoked in `confirmHandler` and `confirmationTokenConfirmHandler` instead of reusing the intent that was created on playground load.
## Motivation

Screenshot 2025-12-17 at 3 13 18 PM

## Testing

Try to pay with Amazon Pay, cancel, try to pay with card

https://github.com/user-attachments/assets/8f4cf3aa-f00c-4e46-96b2-920f256a3a68

## Changelog

N/A

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
@@ -843,47 +843,6 @@ extension PlaygroundController {
         }
     }
 
-    /// Builds the common request body parameters used for both loading backend and creating intents
-    private func buildRequestBody(createCustomerKey: Bool = true) -> [String: Any] {
-        var body = [
-            "customer": customerIdOrType,
-            "currency": settings.currency.rawValue,
-            "amount": settings.amount.rawValue,
-            "merchant_country_code": settings.merchantCountryCode.rawValue,
-            "mode": settings.mode.rawValue,
-            "automatic_payment_methods": settings.apmsEnabled == .on,
-            "use_link": settings.linkPassthroughMode == .pm,
-            "link_mode": settings.linkEnabledMode.rawValue,
-            "use_manual_confirmation": settings.integrationType == .deferred_mc,
-            "require_cvc_recollection": settings.requireCVCRecollection == .on,
-            "is_confirmation_token": settings.confirmationMode == .confirmationToken && !settings.integrationType.isIntentFirst,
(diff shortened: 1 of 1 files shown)


Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
```diff
@@ -1009,6 +1018,57 @@ extension PlaygroundController {
...
+                "amount": settings.amount.rawValue,
+                "merchant_country_code": settings.merchantCountryCode.rawValue,
+                "mode": settings.mode.rawValue,
+                "automatic_payment_methods": settings.apmsEnabled == .on,
+                "use_link": settings.linkPassthroughMode == .pm,
+                "link_mode": settings.linkEnabledMode.rawValue,
+                "use_manual_confirmation": settings.integrationType == .deferred_mc,
+                "require_cvc_recollection": settings.requireCVCRecollection == .on,
+                "is_confirmation_token": settings.confirmationMode == .confirmationToken && !settings.integrationType.isIntentFirst,
+                "payment_method_options_setup_future_usage": settings.paymentMethodOptionsSetupFutureUsage.toDictionary(),
```
porter-stripe: What's the reasoning for omitting the customer session in this payload compared to the loadBackend?

Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
```diff
@@ -1009,6 +1018,57 @@ extension PlaygroundController {
             }
         }
     }
+
+    func createIntent() async throws -> String {
```
porter-stripe: I don't think this needs to return a string, it's never used.

porter-stripe: There's a lot of code duplication between loadBackend() and createIntent(), let's refactor to DRY it up. I think it will be easy for people to forget to add a new intent setting to two nearly identical payloads to the same endpoint.

Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
joyceqin-stripe: Strictly speaking, there's no need to add the customer session fields in the payload— we don't need those fields to create the intent

Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
joyceqin-stripe: Added a function to build the request payload

PR #5935, opened 2025-12-19
Rename app minimum OS version analytics

gbirch-stripe:
## Motivation
"OS" is platform agnostic in case android is ever able to use this same parameter in the future

### DIFF
--- a/StripeCore/StripeCore/Source/Analytics/AnalyticsClientV2.swift
+++ b/StripeCore/StripeCore/Source/Analytics/AnalyticsClientV2.swift
@@ -136,7 +136,7 @@ extension AnalyticsClientV2Protocol {
         }
         payload["app_name"] = Bundle.stp_applicationName() ?? ""
         payload["app_version"] = Bundle.stp_applicationVersion() ?? ""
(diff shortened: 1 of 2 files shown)


gbirch-stripe:
zoolander PR: https://git.corp.stripe.com/stripe-internal/zoolander/pull/637579

StripeCore/StripeCore/Source/Analytics/AnalyticsClientV2.swift
```diff
@@ -136,7 +136,7 @@ extension AnalyticsClientV2Protocol {
         }
         payload["app_name"] = Bundle.stp_applicationName() ?? ""
         payload["app_version"] = Bundle.stp_applicationVersion() ?? ""
-        payload["app_min_ios_version"] = Bundle.stp_minimumOSVersion()
+        payload["app_min_os_version"] = Bundle.stp_minimumOSVersion()
```
wooj-stripe: I think we'll want to default w/ `?? ""`, because i think, setting this to nil won't send the key as part of the payload.

wooj-stripe (approved):

PR #5938, opened 2025-12-22
Vertical mode floating button

joyceqin-stripe:
## Summary

In vertical mode, if the primary button is completely off-screen, float the button at the bottom of the screen.
## Motivation

https://jira.corp.stripe.com/browse/MOBILESDK-2909
## Testing

Long list UI test asserting hittable, CI

https://github.com/user-attachments/assets/f2eb5e44-3cae-4574-be77-4d81bfdc384b
## Changelog

N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/PaymentSheetVerticalViewController.swift
```diff
@@ -149,6 +149,14 @@ class PaymentSheetVerticalViewController: UIViewController, FlowControllerViewCo
     private lazy var errorLabel: UILabel = {
         ElementsUI.makeErrorLabel(theme: configuration.appearance.asElementsTheme)
     }()
+    private lazy var bottomSpacer: UIView = {
+        let spacer = UIView()
+        spacer.translatesAutoresizingMaskIntoConstraints = false
+        let buttonHeight = primaryButton.bounds.height
+        let totalBottomPadding = buttonHeight + buttonSpacing - 20 // account for stackView spacing
```
porter-stripe, line 156: Where does this `-20` come from? If it's the stack view spacing, let's reference that value directly instead of hardcoding.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/PaymentSheetVerticalViewController.swift
```diff
@@ -353,6 +360,80 @@ class PaymentSheetVerticalViewController: UIViewController, FlowControllerViewCo
...
+
+        NSLayoutConstraint.deactivate([
+            primaryButtonTopAnchorConstraint,
+            primaryButtonBottomConstraint,
+        ])
+
+        // Create floating constraint
+        primaryButtonFloatingBottomConstraint = primaryButton.bottomAnchor.constraint(equalTo: bottomSheet.view.safeAreaLayoutGuide.bottomAnchor)
+        stackViewBottomConstraint = stackView.bottomAnchor.constraint(equalTo: view.bottomAnchor)
+        stackViewBottomConstraint.priority = .defaultLow
```
porter-stripe, line 392: Why is this constraint needed and why `.defaultLow` priority?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/PaymentSheetVerticalViewController.swift
```diff
@@ -353,6 +360,80 @@ class PaymentSheetVerticalViewController: UIViewController, FlowControllerViewCo
...
+        NSLayoutConstraint.activate([
+            primaryButtonFloatingBottomConstraint,
+            stackViewBottomConstraint,
+        ])
+
+        // Add bottom spacer to allow scrolling past content
+        stackView.addArrangedSubview(bottomSpacer)
+
+        UIView.animate(withDuration: 0.3) {
+            bottomSheet.view.layoutIfNeeded()
```
porter-stripe, line 403: Why in activate and deactivate floating button are we animating different views?
`activateFloatingButton()` animates the bottomSheet but in deactivateFloatingButton we animate `self.view`.

porter-stripe (commented):
Some general feedback:

- The animate in of the floating button seems a bit unusual in our sheet. Can we check with design if this is desired? Seems like it should just always be there and not slide in.
- Why not have the button always float vs sometimes sticky to the bottom? I feel it would be simpler and desired to make the button always float (at least in the PM list)
- Seems like we need some kind of shadow on the button now to give it some “height” over the content it sits on top of

PR #5943, opened 2026-01-05
Update testRawExpiration to use the current year

joyceqin-stripe:
## Summary

Update testRawExpiration to use the current year so we don't have to update it every year
## Motivation

Test failing
## Testing

Updated test
## Changelog

N/A

### DIFF
--- a/Stripe/StripeiOSTests/STPPaymentCardTextFieldViewModelTest.swift
+++ b/Stripe/StripeiOSTests/STPPaymentCardTextFieldViewModelTest.swift
@@ -38,18 +38,21 @@ class STPPaymentCardTextFieldViewModelTest: XCTestCase {
     }
 
     func testRawExpiration() {
+        // Use current year to ensure test remains valid over time
+        let currentYear = String(format: "%02d", STPCardValidator.currentYear())
+
         // swiftlint:disable:next large_tuple
         let tests: [(String, String, String, String, STPCardValidationState)] = [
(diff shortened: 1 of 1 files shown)


porter-stripe (approved):

Stripe/StripeiOSTests/STPPaymentCardTextFieldViewModelTest.swift
```diff
@@ -38,18 +38,21 @@ class STPPaymentCardTextFieldViewModelTest: XCTestCase {
...
+        // Use current year to ensure test remains valid over time
+        let currentYear = String(format: "%02d", STPCardValidator.currentYear())
+
         // swiftlint:disable:next large_tuple
         let tests: [(String, String, String, String, STPCardValidationState)] = [
             ("", "", "", "", .incomplete),
-            ("12/25", "12/25", "12", "25", .valid),
-            ("1225", "12/25", "12", "25", .valid),
+            ("12/\(currentYear)", "12/\(currentYear)", "12", currentYear, .valid),
+            ("12\(currentYear)", "12/\(currentYear)", "12", currentYear, .valid),
```
wooj-stripe, line 48: I would prefer to keep this as a hardcoded string, as that does not change the expected value of the test.  If we switch to use currentYear(), my guess is that we are comparing STPCardValidator.currentYear() against STPCardValidator.currentYear(), which feels like it would defeat the purpose of this test.

PR #5944, opened 2026-01-05
Hardcode testRawExpiration  year

joyceqin-stripe:
## Summary

Hardcode testRawExpiration year
## Motivation

Don't compare STPCardValidator.currentYear() against itself; hardcode expected value
## Testing

Updated test
## Changelog

N/A


gbirch-stripe:
Won't these tests start failing in December of this year? Can we use something that won't need to be updated or will at least take many years to need to be updated?

PR #5954, opened 2026-01-07
throw download manager error for non-2xx http statuses, log http status in error analytics

gbirch-stripe:
## Motivation
failedToMakeImageFromData error analytics are very noisy, potentially because they are frequently firing for non-2xx http statuses
https://jira.corp.stripe.com/browse/RUN_MOBILESDK-4975


gbirch-stripe:
@wooj sanity checking that there's no situation in which we would want to proceed to attempt decoding of the image even though there is a non-2xx response?

wooj-stripe:
> @wooj sanity checking that there's no situation in which we would want to proceed to attempt decoding of the image even though there is a non-2xx response?

Maybe in a redirect situation (3xx) you could be sent another URL, in which case it could result to a 2xx. Perhaps some very creative enterprise firewall rules may lead to this situation?

gbirch-stripe:
> > @wooj sanity checking that there's no situation in which we would want to proceed to attempt decoding of the image even though there is a non-2xx response?
> 
> Maybe in a redirect situation (3xx) you could be sent another URL, in which case it could result to a 2xx. Perhaps some very creative enterprise firewall rules may lead to this situation?

Hmm I'm not well familiarized with 3xx behavior on iOS. Do you know if the OS networking layer automatically handle the redirection, return the correct data, but still return the original 3xx? If you don't know I can also explore this deeper or just include 3xx in case

wooj-stripe (approved):

PR #5956, opened 2026-01-07
Fix AddressViewController opening picker after autocompleting address

gbirch-stripe:
## Summary
Previously we forced a first responder resignation on the picker to stop it from being opened after the customer selects an autocomplete address. However, this is unreliable due to timing issues, causing the picker to quickly open and then close or not close at all. The better way to accomplish this is to not open it in the first place when the customer selects an autocomplete address by turning off `shouldAutoAdvance`.

## Motivation
https://jira.corp.stripe.com/browse/RUN_MOBILESDK-4862

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/AddressViewController/AddressViewController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/AddressViewController/AddressViewController.swift
@@ -547,13 +547,12 @@ extension AddressViewController: AutoCompleteViewControllerDelegate {
         }
 
         if let autocompleteCountryIndex = autocompleteCountryIndex {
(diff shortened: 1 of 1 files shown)


davidme-stripe (approved):

tillh-stripe:
Thanks for fixing this — can't wait to upgrade the Link app to use the next release and have this fixed 🫶

gbirch-stripe:
> Thanks for fixing this — can't wait to upgrade the Link app to use the next release and have this fixed 🫶

Np! I also saw an ask from your team about the sheet closing cancellation behavior. I'm not able to tackle that during run since it's a larger piece of work that our team has discussed and are tracking with a backlog ticket. We agree with you that that behavior should be changed, it's just not a run sized piece of work, so reach out if it needs to be escalated

PR #5967, opened 2026-01-12
Deflake testPassiveCaptchaConfirmationChallenge

joyceqin-stripe:
## Summary

Due to a race condition with the test log history when running tests simultaneously, sometimes an additional analytic gets logged. Instead of checking for the exact order of events, we really just care that we're seeing the success analytic and we want to check the param on the attach analytic.
## Motivation

Flaky test
## Testing

CI
## Changelog

N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/ConfirmButton.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/ConfirmButton.swift
@@ -18,55 +18,9 @@ private let spinnerMoveToCenterAnimationDuration = 0.35
 
 /// Buy or Continue button
 class ConfirmButton: UIView {
-    // MARK: Internal Properties
-    enum Status {
-        case enabled
-        case disabled
-        case processing
-        case spinnerWithInteractionDisabled
-        case succeeded
-    }
-    enum CallToActionType {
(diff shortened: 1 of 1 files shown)


StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/ConfirmationChallenge/ConfirmationChallengeTests.swift
```diff
@@ -105,8 +105,6 @@ class ConfirmationChallengeTests: XCTestCase {
         // didn't take the full timeout time, exited early
         XCTAssertLessThan(Date().timeIntervalSince(startTime), 10)
         XCTAssertNotNil(hcaptchaToken)
-        let passiveCaptchaEvents = STPAnalyticsClient.sharedClient._testLogHistory.map({ $0["event"] as? String }).filter({ $0?.starts(with: "elements.captcha.passive") ?? false })
-        XCTAssertEqual(passiveCaptchaEvents, ["elements.captcha.passive.init", "elements.captcha.passive.execute", "elements.captcha.passive.success", "elements.captcha.passive.attach"])
```
gbirch-stripe, line 109: Instead of entirely getting rid of this check, could we compare `Set(passiveCaptchaEvents)` and `Set(["elements.captcha....`? That way order doesn't matter but their presence is still required

PR #5976, opened 2026-01-14
Add render_error type to EmbeddedComponentError

hliana-stripe:
## Summary

Adds `render_error` type to `EmbeddedComponentError`

## Motivation

This will allow platforms to have visibility into render errors (when components don't render due to things like network errors or browser extensions that block ui frames from loading)


gimenete-stripe:
Answering your question on Slack: yes, I think we should include this change in the changelog.

hliana-stripe:
https://stripe.slack.com/archives/C01CY476ESJ/p1768412867394999?thread_ts=1768412659.019009&cid=C01CY476ESJ

should be small enough to not go into changelog based on SDK reviewer

PR #5977, opened 2026-01-14
ConfirmButton Refactor 6/6 - Move tap behavior and initialization

gbirch-stripe:
## Summary
Part 6/6 of a factor of the ConfirmButton. Previously the ConfirmButton class supported both the Apple Pay button and confirm buttons, but now it only supports confirm buttons and can be simplified, improve code clarity and SDK binary size.
Draft PR showing the whole change: #5678.

This PR moves the initialization and tap behavior into the BuyButton, renames it to the ConfirmButton, and removes the wrapper. See the first commit for the actual changes, which are small. The second commit is only a formatting pass to re-indent but generates a lot of git diff noise

### DIFF
--- a/Stripe/StripeiOSTests/ConfirmButtonTests.swift
+++ b/Stripe/StripeiOSTests/ConfirmButtonTests.swift
@@ -34,7 +34,7 @@ class ConfirmButtonTests: XCTestCase {
         ]
 
         for (backgroundColor, expectedForeground) in testCases {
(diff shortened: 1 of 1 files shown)


gbirch-stripe:
Review by commit please! The first commit is the actual changes and relatively small. The second commit includes a formatting pass that re-indents and generates a ton of git diff noise.

PR #5978, opened 2026-01-14
Add API model for CheckoutSession

porter-stripe:
## Summary
- Adds the API model of a CheckoutSession based on the API: https://docs.stripe.com/api/checkout/sessions
- Doesn't add every property from the API spec, just ones I think may be useful. We may add/remove as we see fit in the future.
- In future PRs the SDK will make requests to fetch a merchants CheckoutSession and we will need to decode it into this model object
- Adds tests for decoding

## Motivation
- CheckoutSession

## Testing
- Unit tests

## Changelog
N/A

### DIFF
--- a/StripePayments/StripePayments/Source/API Bindings/Models/CheckoutSessions/STPCheckoutSession.swift
+++ b/StripePayments/StripePayments/Source/API Bindings/Models/CheckoutSessions/STPCheckoutSession.swift
@@ -13,10 +13,10 @@ import Foundation
 @_spi(STP) public class STPCheckoutSession: NSObject {
 
     /// The Stripe ID of the CheckoutSession.
-    @objc public let stripeId: String
+    public let stripeId: String
 
     /// The client secret of the CheckoutSession. Used for embedded or custom UI modes.
-    @objc public let clientSecret: String?
+    public let clientSecret: String?
 
     /// Total of all items after discounts and taxes are applied.
     public let amountTotal: Int?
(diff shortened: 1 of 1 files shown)


StripePayments/StripePayments/Source/API Bindings/Models/CheckoutSessions/STPCheckoutSession.swift
```diff
@@ -0,0 +1,197 @@
...
+    @objc public let paymentIntentId: String?
+
+    /// The ID of the SetupIntent for Checkout Sessions in setup mode.
+    @objc public let setupIntentId: String?
+
+    /// The list of payment method types that this CheckoutSession is allowed to use.
+    public let paymentMethodTypes: [STPPaymentMethodType]
+
+    /// Payment-method-specific configuration for this CheckoutSession.
+    public let paymentMethodOptions: STPPaymentMethodOptions?
```
gbirch-stripe, line 46: Why are these not `@objc`?

StripePayments/StripePayments/Source/API Bindings/Models/CheckoutSessions/STPCheckoutSession.swift
```diff
@@ -0,0 +1,197 @@
...
+              let rawStatus = dict["status"] as? String
+        else {
+            return nil
+        }
+
+        let clientSecret = dict["client_secret"] as? String
+        let amountTotal = dict["amount_total"] as? Int ?? 0
+        let currency = dict["currency"] as? String ?? ""
+        let livemode = dict["livemode"] as? Bool ?? false
+        let createdUnixTime = dict["created"] as? TimeInterval ?? Date().timeIntervalSince1970
```
gbirch-stripe: Are these default values specified in documentation somewhere?

porter-stripe: Nope but good flag, this isn't what we want. I've improved the default handling/required fields we expect during decoding.

StripePayments/StripePayments/Source/API Bindings/Models/CheckoutSessions/STPCheckoutSession.swift
porter-stripe: Some of these can't be represented in Obj-C, but some of these were an oversight and should be marked @objc. On second thought we don't need the @objc annotation anyways since we won't use this class in Obj-C so I'm gonna remove them. 

`@objc public let paymentMethodTypes: [STPPaymentMethodType]` doesn't compile b/c Obj-C only let's you put objects in arrays not just plain `Int`'s!

PR #5979, opened 2026-01-14
Fix not sending mandate_data for PMO SFU + deferred intents + PMs

yuki-stripe:
## Summary
1. Don't set PMO SFU on the Intent if we're deferred in the test playground, to match [how we tell merchants to integrate](https://docs.corp.stripe.com/payments/mobile/accept-payment?platform=ios&type=paymentsfu#payment-method-options-setup-future-usage).
2. Fix the bug described in the PR title

## Testing
TODO

## Changelog
See changelog


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+DeferredAPI.swift
```diff
@@ -279,6 +283,17 @@ extension PaymentSheet {
...
     }
 
+    /// Sets mandate data if the PI params is setting up the PM type
+    static func setMandateDataIfNecessary(for paymentMethodType: STPPaymentMethodType, on paymentIntentParams: STPPaymentIntentConfirmParams) {
+        let isTopLevelSFUSet = [.onSession, .offSession].contains(paymentIntentParams.setupFutureUsage)
+        let isPMOSFUSet = [STPPaymentIntentSetupFutureUsage.offSession, STPPaymentIntentSetupFutureUsage.onSession].map({ $0.stringValue }).contains(
+            paymentIntentParams.nonnil_paymentMethodOptions.setupFutureUsage(for: paymentMethodType)
+        )
+        if STPPaymentMethodType.requiresMandateDataForPaymentIntent.contains(paymentMethodType) && (isTopLevelSFUSet || isPMOSFUSet) {
+            paymentIntentParams.mandateData = .makeWithInferredValues()
```
porter-stripe, line 293: should we ever not set `paymentIntentParams.mandateData` if it's already populated? I'm guessing it would already have been populated w/ `.makeWithInferredValues()` but just curious.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetLPMConfirmFlowTests.swift
```diff
@@ -1078,6 +1093,95 @@ extension PaymentSheet_LPM_ConfirmFlowTests {
                 ("Deferred PaymentIntent w/ setup_future_usage - client side confirmation with confirmation token", makeDeferredIntent(deferredCSCWithConfirmationToken)),
                 ("Deferred PaymentIntent w/ setup_future_usage - server side confirmation with confirmation token", makeDeferredIntent(deferredSSCWithConfirmationToken)),
             ]
+        case .paymentIntentWithPMOSetupFutureUsage:
```
porter-stripe, line 1096: 👍

porter-stripe (dismissed):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+DeferredAPI.swift
```diff
@@ -150,6 +150,10 @@ extension PaymentSheet {
                     // Set top-level SFU and PMO SFU to match the intent config
                     setSetupFutureUsage(for: paymentMethod.type, intentConfiguration: intentConfig, on: paymentIntentParams)
 
+                    // Now that we've potentially set SFU/PMO SFU, set mandate data accordingly again.
+                    // This is a bit hacky, since the above `makePaymentIntentParams` call already set mandate data
```
joyceqin-stripe, line 154: Could we intercept the payment method options and set the SFU values to match the intent config before we call `makePaymentIntentParams` so that we don't have to do this hack?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+DeferredAPI.swift
yuki-stripe: I can't think of anything that would break and the top-level SFU tests should exercise that case where we set `mandateData` twice. 

It's def. a code smell though and there's a better way to structure this code but I couldn't figure it out quickly.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+DeferredAPI.swift
yuki-stripe: @joyceqin-stripe Ooh say more? Like, pass the IntentConfig to `makePaymentIntentParams` so it knows if SFU/PMO SFU _will be_ set?  I can't think of any refactor yet that feels much better than what I've got here.

joyceqin-stripe: Yeah I don't have a fully formed idea, something like that could work or maybe like before we call `makePaymentIntentParams` we could update the payment method options on the `confirmPaymentMethodType` with the relevant SFU values based on the intent config and then update the check in `makePaymentIntentParams` that sets mandate data to also read from the params' sfu?
I don't know if that's better than your current approach but it's a little weird to attempt to make the mandate data twice in quick succession

yuki-stripe: Oh interesting, maybe! I'll explore that in a followup PR.

porter-stripe (approved):

PR #5983, opened 2026-01-15
Add API function to init a CheckoutSession

porter-stripe:
## Summary
- Calls the internal endpoint to init a payment page that returns back a CheckoutSession and a ElementsSession
- This API will later be used to be able to load PaymentSheet with a CheckoutSession ID

## Motivation
- CheckoutSession

## Testing
- New tests

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/STPAPIClient+CheckoutSession.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/STPAPIClient+CheckoutSession.swift
@@ -19,13 +19,9 @@ struct CheckoutSessionInitResponse {
 extension STPAPIClient {
 
     /// Initializes a CheckoutSession, fetching payment configuration data.
-    /// - Parameters:
-    ///   - checkoutSessionId: The ID of the checkout session (e.g., "cs_test_xxx")
-    ///   - completion: Completion handler with the result containing the session and elements session
-    func initCheckoutSession(
-        checkoutSessionId: String,
-        completion: @escaping (Result<CheckoutSessionInitResponse, Error>) -> Void
-    ) {
+    /// - Parameter checkoutSessionId: The ID of the checkout session (e.g., "cs_test_xxx")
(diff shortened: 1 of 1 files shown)


StripePayments/StripePayments/Source/API Bindings/Models/CheckoutSessions/STPCheckoutSession.swift
```diff
@@ -48,12 +48,6 @@ import Foundation
     /// Whether or not this CheckoutSession was created in livemode.
     public let livemode: Bool
 
-    /// Time at which the CheckoutSession was created.
-    public let created: Date
-
-    /// The timestamp at which the Checkout Session will expire.
-    public let expiresAt: Date
-
```
porter-stripe, line 56: The internal API for payment pages doesn't seem to return these back, I think that's fine we probably won't need them.

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/STPAPIClient+CheckoutSession.swift
```diff
@@ -0,0 +1,72 @@
...
+extension STPAPIClient {
+
+    /// Initializes a CheckoutSession, fetching payment configuration data.
+    /// - Parameters:
+    ///   - checkoutSessionId: The ID of the checkout session (e.g., "cs_test_xxx")
+    ///   - completion: Completion handler with the result containing the session and elements session
+    func initCheckoutSession(
+        checkoutSessionId: String,
+        completion: @escaping (Result<CheckoutSessionInitResponse, Error>) -> Void
+    ) {
```
gbirch-stripe: What's the reason to still include a completion-based version of this? Could we stick entirely to swift concurrency for Checkout Sessions?

porter-stripe: Good question! I didn't realize APIRequest.post had an async overload, so I was wrapping the completion-based version. Updated to use the async version directly.

PR #5984, opened 2026-01-15
Fix embedded playground confirm failing for interstitial integration

yuki-stripe:
## Summary
Embedded test playground always fails to confirm when using the 'interstitial' variation (see screenshot) because the presentingViewController was not in the view hierarchy

simulator_screenshot_1B0D7A2F-64D4-4BF5-9316-5A773345B9D2

## Testing
Manual

## Changelog
Not user facing


gbirch-stripe (approved):
ty for this fix!

PR #5989, opened 2026-01-16
update local form_specs with latest from server

gbirch-stripe:


gbirch-stripe:
replaced with https://github.com/stripe/stripe-ios/pull/6005

PR #5994, opened 2026-01-18
add eftpos apple pay support

franklingu-stripe:
## Summary

Add eftpos support for apple pay

## Motivation

Eftpos Apple Pay

## Testing

Added test


franklingu-stripe:
r?

StripeCore/StripeCore/Source/API Bindings/StripeAPI.swift
```diff
@@ -72,8 +72,8 @@ import PassKit
 
     /// The SDK accepts Amex, Mastercard, Visa, and Discover for Apple Pay.
     ///
-    /// Set this property to enable other card networks in addition to these, such as .JCB or .cartesBancaires.
-    /// For example, `additionalEnabledApplePayNetworks = [.JCB]` enables JCB (note this requires onboarding from JCB and Stripe).
+    /// Set this property to enable other card networks in addition to these, such as .JCB, .cartesBancaires, or .eftpos (for EFTPOS Australia).
+    /// For example, `additionalEnabledApplePayNetworks = [.eftpos]` enables EFTPOS Australia (note this requires the merchant to support EFTPOS Australia payment).
```
bernardo-stripe: The note about merchant support would also apply to other networks. We should probably instead move this to a comment to warn that they should make sure they support the networks before enabling them here.

bernardo-stripe (commented):
It's a nit, but Eftpos branding is `eftpos Australia` (all lowercase eftpos).
We should probably always format it like that on public docs.

Other than that looks good to me, similar to the changes for CB on #3402.

PR #5996, opened 2026-01-20
Add custom merchant with secret and publishable key input fields to playground

joyceqin-stripe:
## Summary

In the merchant selector, you can choose "custom" and input the secret key and publishable key for a merchant. This makes it easier to test with other merchants, including live testing, without having to add the keys on the playground backend.

This was also recently added to the Android playground.
## Motivation

Easier custom merchant testing, AB test bug bash
## Testing

Manual
## Changelog

N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Analytics/PaymentSheetAnalyticsHelper.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Analytics/PaymentSheetAnalyticsHelper.swift
@@ -555,6 +555,7 @@ extension PaymentElementConfiguration {
         payload["billing_details_collection_configuration"] = billingDetailsCollectionConfiguration.analyticPayload
         payload["preferred_networks"] = preferredNetworks?.map({ STPCardBrandUtilities.apiValue(from: $0) }).joined(separator: ", ")
         payload["card_brand_acceptance"] = cardBrandAcceptance != .all
+        payload["card_funding_acceptance"] = allowedCardFundingTypes != .all
         if let cpms = customPaymentMethodConfiguration?.customPaymentMethods {
             payload["custom_payment_methods"] = cpms.map { $0.id }
         }
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/CardSection/CardSectionElement.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/CardSection/CardSectionElement.swift
@@ -45,6 +45,9 @@ final class CardSectionElement: ContainerElement {
     let analyticsHelper: PaymentSheetAnalyticsHelper?
     let cardBrandFilter: CardBrandFilter
     let cardFundingFilter: CardFundingFilter
+    /// Separate BIN controller for funding filtering to avoid polluting
+    /// See: https://jira.corp.stripe.com/browse/RUN_MOBILESDK-5052
+    private let fundingBinController: STPBINController = STPBINController()
     private let opensCardScannerAutomatically: Bool
 
     private let linkAppearance: LinkAppearance?
@@ -120,7 +123,8 @@ final class CardSectionElement: ContainerElement {
             defaultValue: defaultValues.pan,
             cardBrandDropDown: cardBrandDropDown?.element,
             cardBrandFilter: cardBrandFilter,
-            cardFundingFilter: cardFundingFilter
+            cardFundingFilter: cardFundingFilter,
+            fundingBinController: fundingBinController
         ), theme: theme) { field, params in
             cardParams(for: params).number = field.text
             return params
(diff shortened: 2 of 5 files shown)


Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlaygroundSettings.swift
```diff
@@ -697,6 +698,8 @@ struct PaymentSheetTestPlaygroundSettings: Codable, Equatable {
     var currency: Currency
     var amount: Amount
     var merchantCountryCode: MerchantCountry
+    var customSecretKey: String?
+    var customPublishableKey: String?
```
porter-stripe, line 703: In case someone looks at our code can we add some comments that having the keys defined on the client is not best practice and it's just for testing?

Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlayground.swift
```diff
@@ -139,6 +139,14 @@ struct PaymentSheetTestPlayground: View {
                             SettingPickerView(setting: $playgroundController.settings.currency)
                         }
                         SettingPickerView(setting: merchantCountryBinding)
+                        if playgroundController.settings.merchantCountryCode == .custom {
+                            TextField("Custom secret key", text: customSecretKeyBinding)
+                                .autocapitalization(.none)
+                                .autocorrectionDisabled()
+                            TextField("Custom publishable key", text: customPublishableKeyBinding)
+                                .autocapitalization(.none)
+                                .autocorrectionDisabled()
```
porter-stripe: Small nice to have would be adding hints like "sk_test_..." or "pk_test_..." as placeholder text to guide users on the expected format.

Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
```diff
@@ -907,7 +907,9 @@ extension PlaygroundController {
                 self.customerId = json["customerId"]
                 self.customerSessionClientSecret = json["customerSessionClientSecret"]
                 self.paymentMethodTypes = json["paymentMethodTypes"]?.components(separatedBy: ",")
-                STPAPIClient.shared.publishableKey = json["publishableKey"]
+                if let publishableKey = json["publishableKey"] {
+                    STPAPIClient.shared.publishableKey = publishableKey
+                }
```
porter-stripe, line 912: If the backend does not return a publishableKey, the STPAPIClient.shared.publishableKey might retain a stale value from a previous load.

Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
```diff
@@ -1030,6 +1032,15 @@ extension PlaygroundController {
...
             //            "set_shipping_address": true // Uncomment to make server vend PI with shipping address populated
         ] as [String: Any]
+
+        // Send custom keys to backend if provided
+        if let customSecretKey = settings.customSecretKey, !customSecretKey.isEmpty {
+            body["custom_secret_key"] = customSecretKey
+        }
+        if let customPublishableKey = settings.customPublishableKey, !customPublishableKey.isEmpty {
+            body["custom_publishable_key"] = customPublishableKey
+        }
```
porter-stripe, line 1043: Maybe we should just set the publishable key on the API client if we have it.

```swift
if let customPublishableKey = settings.customPublishableKey, !customPublishableKey.isEmpty {
   body["custom_publishable_key"] = customPublishableKey
   STPAPIClient.shared.publishableKey = customPublishableKey
}

File: Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlaygroundSettings.swift (hunk 1 above)
Comment (porter-stripe) [reply Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlaygroundSettings.swift:703]: Let's also clear these when the user switches from "custom"

File: Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlaygroundSettings.swift [hunk 5]
```diff
@@ -164,6 +164,7 @@ struct PaymentSheetTestPlaygroundSettings: Codable, Equatable {
         case DE
         case IT
         case stripeShop = "stripe_shop_test"
+        case custom
```
porter-stripe, line 167: Nit: I kinda think "custom API keys" would be a bit more clear

Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
joyceqin-stripe: True, but if the backend does not return a publishableKey and we set STPAPIClient.shared.publishableKey to nil, we get an assertionFailure "You must use a valid publishable key. For more info, see https://stripe.com/docs/keys".

Example/PaymentSheet Example/PaymentSheet Example/PaymentSheetTestPlaygroundSettings.swift
joyceqin-stripe: The playground backend responds to "custom", so I'm going to keep it like this for now.

PR #5997, opened 2026-01-20
Implementing loading of a CheckoutSession

porter-stripe:
## Summary
- Adds some protected APIs that allow loading PaymentSheet, FlowController, and Embedded with a CheckoutSession id.
- Implements loading of the CheckoutSession to populate the fields required to load MPE.
- A follow up pull request will update our playground to fetch a CheckoutSession ID from the backend to use for loading, it was currently tested with hard coded values.

#### Example loading with a CheckoutSession ID:
Simulator Screen Recording - iPhone 17 Pro - 2026-01-21 at 11 22 30

## Motivation
- CheckoutSession

## Testing
- Manual
- Unit tests

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Intent.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Intent.swift
@@ -116,7 +116,7 @@ enum Intent {
                 return nil
             }
         case .checkoutSession(let session):
-            return session.amountTotal
+            return session.totalSummary?.total
         }
     }
 
--- a/StripePayments/StripePayments/Source/API Bindings/Models/CheckoutSessions/STPCheckoutSession.swift
+++ b/StripePayments/StripePayments/Source/API Bindings/Models/CheckoutSessions/STPCheckoutSession.swift
@@ -18,8 +18,8 @@ import Foundation
     /// The client secret of the CheckoutSession. Used for embedded or custom UI modes.
     public let clientSecret: String?
 
-    /// Total of all items after discounts and taxes are applied.
-    public let amountTotal: Int?
+    /// Summary of amounts for the CheckoutSession.
+    public let totalSummary: STPCheckoutSessionTotalSummary?
 
     /// Three-letter ISO currency code, in lowercase.
     public let currency: String?
@@ -71,7 +71,7 @@ import Foundation
         let props: [String] = [
             String(format: "%@: %p", NSStringFromClass(STPCheckoutSession.self), self),
             "stripeId = \(stripeId)",
-            "amountTotal = \(String(describing: amountTotal))",
+            "totalSummary = \(String(describing: totalSummary))",
             "clientSecret = <redacted>",
             "currency = \(String(describing: currency))",
             "mode = \(String(describing: allResponseFields["mode"]))",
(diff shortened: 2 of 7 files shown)


StripePayments/StripePayments/Source/API Bindings/Models/CheckoutSessions/STPCheckoutSession.swift
```diff
@@ -18,8 +18,8 @@ import Foundation
     /// The client secret of the CheckoutSession. Used for embedded or custom UI modes.
     public let clientSecret: String?
 
-    /// Total of all items after discounts and taxes are applied.
-    public let amountTotal: Int?
+    /// Summary of amounts for the CheckoutSession.
+    public let totalSummary: STPCheckoutSessionTotalSummary?
```
porter-stripe, line 22: Another difference in response shape from the public API for CheckoutSession compared to the payment_pages/init response :/

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
```diff
@@ -88,6 +88,37 @@ public final class EmbeddedPaymentElement {
         return embeddedPaymentElement
     }
 
+    /// An asynchronous failable initializer for CheckoutSession mode
+    /// Loads the CheckoutSession's payment methods and configuration.
+    /// - Parameter checkoutSessionId: The ID of a Stripe CheckoutSession object (e.g., "cs_test_xxx")
+    /// - Parameter configuration: Configuration for the PaymentSheet. e.g. your business name, customer details, etc.
+    /// - Returns: A valid EmbeddedPaymentElement instance
+    /// - Throws: An error if loading failed.
+    @_spi(CheckoutSessionPreview) public static func create(
```
gbirch-stripe, line 97: iirc the other CheckoutSessions APIs are marked `@_spi(STP)`, not `@_spi(CheckoutSessionPreview)`? Should we keep it as `@_spi(STP)` while it's in development and then change it to preview when it's in preview?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Intent.swift
```diff
@@ -54,14 +58,16 @@ enum Intent {
...
+            return false
         }
     }
 
     var intentConfig: PaymentSheet.IntentConfiguration? {
         switch self {
         case .deferredIntent(let intentConfig):
             return intentConfig
-        default:
+        case .paymentIntent, .setupIntent, .checkoutSession:
```
gbirch-stripe, line 70: nice, I dislike `default` cases :)

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
porter-stripe: Good eye! This is intentional, the CheckoutSession classes marked as `STP` such as `STPCheckoutSession` aren't intended to ever be made public to merchants (I don't think there's a use case AFAIK). It's more of an internal data class that the merchants shouldn't need to know about. But these `create`/init functions for MPE will be exposed to merchants.

PR #6001, opened 2026-01-21
Delay flow controller experiment exposure logging

joyceqin-stripe:
## Summary

When the FlowController makes its viewController, we resolve the layout based on the experiment. However, we want to delay the experiment exposure logging until the sheet is actually presented.
## Motivation

In our experiment to see when horizontal vs vertical layout leads to better conversion, our conversion metric is determined by dividing the number of success analytics divided by the unique sessions with logged experiment exposures. However, if we log the exposure for FlowController as soon as we make the view controller, then we would get sessions in our experiment where the layout has no real bearing on the user's actions. For example, if Apple Pay were already selected and the user confirmed with Apple Pay without presenting the sheet, this should not be included in the experiment results.
https://stripe.slack.com/archives/C0KCARHSL/p1769022897670379
## Testing

Manual
## Changelog

N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Resources/JSON/form_specs.json
+++ b/StripePaymentSheet/StripePaymentSheet/Resources/JSON/form_specs.json
@@ -138,56 +138,56 @@
                 "translation_id": "upe.labels.ideal.bank",
                 "items": [
                     {
-                      "display_text": "ABN Amro",
-                      "api_value": "abn_amro"
+                        "display_text": "ABN Amro",
+                        "api_value": "abn_amro"
                     },
                     {
-                      "display_text": "ASN Bank",
-                      "api_value": "asn_bank"
+                        "display_text": "ASN Bank",
+                        "api_value": "asn_bank"
                     },
                     {
-                      "display_text": "bunq B.V.",
-                      "api_value": "bunq"
+                        "display_text": "bunq B.V.",
+                        "api_value": "bunq"
                     },
                     {
-                      "display_text": "ING Bank",
-                      "api_value": "ing"
+                        "display_text": "ING Bank",
+                        "api_value": "ing"
                     },
                     {
-                      "display_text": "Knab",
-                      "api_value": "knab"
+                        "display_text": "Knab",
+                        "api_value": "knab"
                     },
                     {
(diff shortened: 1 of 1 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
```diff
@@ -691,6 +711,21 @@ extension PaymentSheet {
...
+            guard needsToLogLayoutExperimentExposure else { return }
+            needsToLogLayoutExperimentExposure = false
+
+            // Re-resolve the layout to log the exposure this time
+            var mutableConfig = configuration
+            _ = mutableConfig.resolveLayout(
+                loadResult: viewController.loadResult,
+                configuration: configuration,
+                analyticsHelper: analyticsHelper,
+                shouldLogExperimentExposure: true
```
gbirch-stripe: why do we resolve the layout a second time just to log?

joyceqin-stripe: Resolving it there doesn't really do anything besides log, but fair point that it's not necessary to do it that way. Refactored to separate the experiment creation process into a separate function so that we can create the experiments and then log them

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
```diff
@@ -268,52 +269,22 @@ extension PaymentSheet {
...
+                    loadResult: loadResult,
+                    configuration: configuration,
+                    analyticsHelper: analyticsHelper
+                )
 
                 experiments.forEach { experiment in
-                    let isExperimentActive = elementsSession.experimentsData?.experimentAssignments[experiment.name] != nil
-                    if isExperimentActive {
-                        // Log experiment exposure
+                    // Log experiment exposure (if not deferred)
```
gbirch-stripe: Shouldn't this read "if not flow controller" or "if sheet not yet presented" or something like that? "Deferred" makes it sound like we're talking about deferred intents here

PR #6012, opened 2026-01-23
Change integration type dimension from complete/custom to paymentsheet/flowcontroller

joyceqin-stripe:
## Summary

Use paymentsheet/flowcontroller instead of complete/custom based on bug bash feedback.
## Motivation

Bug bash feedback
## Testing

Updated tests
## Changelog

N/A

### DIFF
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -599,6 +599,12 @@ workflows:
     - xcode-start-simulator@0:
         inputs:
         - destination: $DEFAULT_TEST_DEVICE
+    - script@1:
+        inputs:
+        - content: |
+            xcrun simctl spawn booted defaults write com.apple.keyboard.preferences DidShowContinuousPathIntroduction -bool true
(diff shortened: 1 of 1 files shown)


StripePaymentSheet/StripePaymentSheet/Source/Analytics/Experiments/PaymentSheetLayoutExperiment.swift
```diff
@@ -42,18 +42,7 @@ struct PaymentSheetLayoutExperiment {
...
-                return "complete"
-            case .flowController:
-                return "custom"
-            case .embedded:
-                return "embedded"
-            default:
-                return integrationShape.analyticsValue
-            }
-        }()
+        self.integrationShape = integrationShape.analyticsValue
```
porter-stripe, line 45: much nicer!

PR #6023, opened 2026-01-28
Payment Method Form Hierarchy Tests

gbirch-stripe:
## Summary
New tests that codify the element structure of payment method forms.

## Motivation
Better track payment method form changes for the form specs removal project


StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/FormSpecHierarchyTests.swift
```diff
@@ -0,0 +1,775 @@
...
+            ]),
+        ])
+
+        XCTAssertEqual(actual, expected)
+    }
+}
+
+// MARK: - Mandate Tests (for payment methods that show mandate when setting up)
+
+class FormSpecMandateTests: XCTestCase {
```
yuki-stripe: This seems redundant with LPMConfirmTests, which also check mandate appears for PI+SFU and SI when necessary.

PR #6024, opened 2026-01-28
Confirm payment with CheckoutSession

porter-stripe:
## Summary
- Implement confirming new payment methods in CheckoutSessions
- Following PRs will add support for confirming saved PMs, Link, Apple Pay.

## Motivation
- CheckoutSessions

## Testing
- Add LPMConfirmFlowTests which test all confirm flows for new PMs

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+CheckoutSessionAPI.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+CheckoutSessionAPI.swift
@@ -39,7 +39,7 @@ extension PaymentSheet {
                                                       additionalNonPIIParams: ["payment_method_type": newPaymentMethod.type])
                     STPAnalyticsClient.sharedClient.log(analytic: errorAnalytic)
                 }
-                stpAssert(newPaymentMethod == nil)
+                stpAssert(newPaymentMethod == nil, "newPaymentMethod should be nil when confirming with a new payment method; the payment method is created from params.")
                 paymentMethodType = params.type
                 paymentMethodOptions = paymentOptions
                 params.clientAttributionMetadata = clientAttributionMetadata


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+CheckoutSessionAPI.swift
```diff
@@ -0,0 +1,154 @@
...
+//
+
+import Foundation
+@_spi(STP) import StripeCore
+@_spi(STP) import StripePayments
+
+extension PaymentSheet {
+
+    /// Confirms a checkout session with a new payment method
+    @MainActor
```
gbirch-stripe, line 15: What's the reason for `@MainActor` isolating this method? It doesn't seem to interact with shared data or anything that would indicate it needs to be isolated.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+CheckoutSessionAPI.swift
```diff
@@ -0,0 +1,154 @@
...
+            let paymentMethodOptions: STPConfirmPaymentMethodOptions?
+            switch confirmType {
+            case let .new(params, paymentOptions, newPaymentMethod, _, _):
+                if let newPaymentMethod {
+                    let errorAnalytic = ErrorAnalytic(event: .unexpectedPaymentSheetConfirmationError,
+                                                      error: PaymentSheetError.unexpectedNewPaymentMethod,
+                                                      additionalNonPIIParams: ["payment_method_type": newPaymentMethod.type])
+                    STPAnalyticsClient.sharedClient.log(analytic: errorAnalytic)
+                }
+                stpAssert(newPaymentMethod == nil)
```
gbirch-stripe: Could we add more details in the assert about why this isn't supposed to happen? I don't think I would know what to do if I didn't have context and hit this

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+CheckoutSessionAPI.swift
porter-stripe: When we do confirmation of a payment we generally always run on the MainActor to keep things simple as confirmation can require UI work/presentations and ensure the `PaymentSheetResult` is delivered back on the main actor IIRC. In the future we should probably update the STPPaymentHandler to require it's functions to be called on the MainActor because currently you can call them from any thread. This @MainActor annotation matches similar function definitions we have in PaymentSheet+DeferredAPI and PaymentSheet+ConfirmationTokens. We _may_ be able to remove the MainActor annotations but it might be slightly safer to keep it to ensure confirmation runs on the MainActor.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+CheckoutSessionAPI.swift
porter-stripe: Improved the language on the assert. It's a bit unfortunate that the `new` case has a `paymentMethod` field, it doesn't really make sense and is bit of a code smell of a bad pattern we should fix. This assert should _never_ fire, I can't recall where we use the newPaymentMethod param on this switch case...

PR #6026, opened 2026-01-29
Form Spec Removal 1 — Klarna, Affirm

gbirch-stripe:


Tests/ReferenceImages_64/StripePaymentSheetTests.PaymentSheetFormFactorySnapshotTest/testLpm_Klarna_AllFields_NoDefaults@3x.png
wooj-stripe, line 1: I'm a bit confused about this snapshot test? Both images seem empty?

gbirch-stripe:
> Do we need to re-record the testAffirmConfirmFlows and testKlarnaConfrimFlow tests?

The reason re-recording was needed was that the `additionalAPIParameters` (used by form_specs) encodes the billing details field names differently than when they are directly attached (used by explicit definitions), creating superficial changes in the API calls. Now that you point this out, it's pretty confusing and hard to review to have all of these .tail files changes that _could_ have an issue buried in them, so I created a PR to make the form_specs PMs align with the explicitly defined ones in how they attach these details: https://github.com/stripe/stripe-ios/pull/6082
After that merges, no re-records of the confirm tests will be needed for this and other migration PRs.

Tests/ReferenceImages_64/StripePaymentSheetTests.PaymentSheetFormFactorySnapshotTest/testLpm_Klarna_AllFields_NoDefaults@3x.png
gbirch-stripe: This is a chrome+github UI bug 🙃 . If you look at the PR in safari you can see the images.

wooj-stripe: fwiw, i can see it fine in chrome now. thanks!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
```diff
@@ -828,17 +818,8 @@ extension PaymentSheetFormFactory {
...
-                currency: currency,
-                appearance: configuration.appearance
-            ),
-            isHorizontalMode: configuration.isHorizontalMode
-        )
-    }
-
-    func makeKlarnaCountry(apiPath: String? = nil) -> PaymentMethodElement? {
+    /// Creates a country dropdown for Klarna form specs with API path support
+    func makeKlarnaCountry(apiPath: String?) -> PaymentMethodElement? {
```
wooj-stripe, line 822: nit: Maybe we update the comment here saying that this is to be deleted once we remove it out of PaymentSheetFormFactory+FormSpec?

wooj-stripe (approved):

PR #6028, opened 2026-01-29
Confirm CheckoutSessions in Apple Pay

porter-stripe:
## Summary
- Adds support for confirming payments using Apple Pay
- Future PRs will add the support for PI+SFU/SI in Apple Pay.

## Motivation
- CheckoutSessions

## Testing
- New unit test
- New E2E test for Apple Pay confirmation

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/CheckoutSessionConfirmResponse.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/CheckoutSessionConfirmResponse.swift
@@ -83,16 +83,19 @@ extension CheckoutSessionConfirmResponse {
     /// - Returns: The client secret string from the underlying intent
     /// - Throws: PaymentSheetError if the expected intent is missing
     func clientSecret(for mode: STPCheckoutSessionMode) throws -> String {
-        if mode == .setup {
+        switch mode {
+        case .setup:
             guard let setupIntent = setupIntent else {
                 throw PaymentSheetError.unknown(debugDescription: "Missing setup intent in confirm response")
             }
             return setupIntent.clientSecret
-        } else {
+        case .payment, .subscription:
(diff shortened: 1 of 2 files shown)


StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/CheckoutSessionConfirmResponse.swift
```diff
@@ -76,3 +76,23 @@ extension CheckoutSessionConfirmResponse: STPAPIResponseDecodable {
...
     }
 }
+
+extension CheckoutSessionConfirmResponse {
+    /// Extracts the client secret from the confirm response based on checkout session mode.
+    /// - Parameter mode: The checkout session mode (payment, setup, or subscription)
+    /// - Returns: The client secret string from the underlying intent
+    /// - Throws: PaymentSheetError if the expected intent is missing
+    func clientSecret(for mode: STPCheckoutSessionMode) throws -> String {
+        if mode == .setup {
```
gbirch-stripe: Should this be a switch on `mode` instead of an `if`/`else`? To protect against unexpected third cases / be adaptive to future `STPCheckoutSessionMode` changes

porter-stripe: Yeah good idea! I had to remove @objc for @frozen to get the good things about the switch statement, otherwise it wanted a default case.

gbirch-stripe: yuki and I have had some interesting discussions about `@frozen`. I think it makes sense here and 99% of places where it's `@_spi(STP) public` since these aren't real public declarations and enum changes should be breaking changes

PR #6030, opened 2026-01-29
MPE Load fixes/improvements

yuki-stripe:
## Summary
Various PaymentSheetLoader.load improvements and one bugfix. See commits for each.

## Motivation
Trying to make `load` analytics more robustly capture everything in the method, and ended up fixing a few things along the way.

## Testing
Relying on existing tests.

## Changelog
See changelog


yuki-stripe:
> I'd like to make sure that other things (like the Link experiments) won't be affected by the change in integrationShape from .complete -> .paymentSheet or at least that they're aware of the change, but otherwise looks fine to me

I believe that's `PaymentSheetAnalyticsHelper.IntegrationShape`, different from the `PaymentSheetLoader.IntegrationShape` that I changed here.

joyceqin-stripe (dismissed):

PR #6040, opened 2026-02-03
Captcha parse timeout

joyceqin-stripe:
## Summary

Parse elements/sessions-provided token timeout from response instead of using hard-coded value. The backend-provided value will already include buffer time (expiring early) so a user won't send an expired token
https://git.corp.stripe.com/stripe-internal/pay-server/pull/1331160
## Motivation

HCaptcha project
## Testing

Updated test
## Changelog

N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 ### PaymentSheet
 * [Changed] Afterpay/Clearpay no longer requires billing address by default. Set `billingDetailsCollectionConfiguration.address = .full` if you need to collect billing address for Afterpay.
 
+### CustomerSheet
+* [Fixed] Fixed a view controller lifecycle issue that could cause keyboard avoidance and layout problems on the Add Card screen.
+
 ### Identity
 
 * [Added] Added a best frame detector to document capture.
--- a/Stripe.xcworkspace/xcshareddata/xcschemes/AllStripeFrameworks.xcscheme
+++ b/Stripe.xcworkspace/xcshareddata/xcschemes/AllStripeFrameworks.xcscheme
@@ -333,6 +333,14 @@
                BlueprintName = "StripePaymentSheetTests"
                ReferencedContainer = "container:StripePaymentSheet/StripePaymentSheet.xcodeproj">
             </BuildableReference>
+            <SkippedTests>
+               <Test
+                  Identifier = "MPELatencyTest">
+               </Test>
+               <Test
+                  Identifier = "MPELatencyTest/testLoadLatency()">
+               </Test>
+            </SkippedTests>
          </TestableReference>
          <TestableReference
             skipped = "NO">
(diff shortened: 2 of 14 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ConfirmationChallenge/PassiveCaptchaChallenge.swift
```diff
@@ -32,9 +33,11 @@ struct PassiveCaptchaData: Equatable, Hashable {
 
         // Optional
         let rqdata = response["rqdata"] as? String
+        let tokenTimeoutMs = response["token_timeout_ms"] as? Double
```
toluo-stripe: https://stripe.sourcegraphcloud.com/stripe-internal/pay-server/-/blob/lib/elements/api/resources/elements_passive_captcha_resource.rb?L16

`optional :token_timeout_seconds, IntegerField.new`

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ConfirmationChallenge/PassiveCaptchaChallenge.swift
```diff
@@ -139,7 +139,10 @@ actor PassiveCaptchaChallenge {
     }
 
     private func setSessionExpirationDate() {
-        self.sessionExpirationDate = Date().addingTimeInterval(sessionExpiration)
+        guard let tokenTimeoutSeconds = passiveCaptchaData.tokenTimeoutSeconds else {
+            return
+        }
+        self.sessionExpirationDate = Date().addingTimeInterval(tokenTimeoutSeconds)
```
wooj-stripe: If i understand correctly this means that if the server does not set a value for `token_timeout_seconds` (or we fail to parse this value), then there is no sessionExpirationDate, which means hasSessionExpired will always return false, meaning, we will never call resetSession().

If this happens, it seems like on line 75, we'll check `if let tokenTask {` and we'll always return the that token without ever being able to refresh it.

Does it make sense to default the time out to 29*60, or is not setting `token_timeout_seconds` a feature to ensuring that you can never fetch another token?

joyceqin-stripe: On Android, it defaults to Int.max, which is realistically ~never. I guess there's no reason not to default to 29*60 instead though

wooj-stripe: If we are going to default to 29*30, I'd prefer to make `PassiveCaptchaData.tokenTimeoutSeconds` non optional. To do this, we can:
```
struct PassiveCaptchaData: Equatable, Hashable {
....
    let tokenTimeoutSeconds: TimeInterval
    static let defaultTokenTimeoutSeconds: TimeInterval = 29 * 30
    init(siteKey: String, rqdata: String?, tokenTimeoutSeconds: TimeInterval = defaultTokenTimeoutSeconds) {
        self.siteKey = siteKey
        self.rqdata = rqdata
        self.tokenTimeoutSeconds = tokenTimeoutSeconds
    }

```
when decoding:
```
        let tokenTimeoutSeconds = response["token_timeout_seconds"] as? TimeInterval ?? defaultTokenTimeoutSeconds
```

Issue #6013, opened 2026-01-24
[BUG] CustomerSheet - MainActor isolated crash runtime warning (Swift 6)


cvillanueva10:
## Summary

After upgrading to Swift 6, we are seeing a crash runtime warning when presenting CustomerSheet that results in completion(CustomerSheetResult.error). 

The warning is 
`\StripePaymentSheetPresenter.isPaymentMethodsPresented is isolated to the main actor. Accessing it via Binding from a different actor will cause undefined behaviors, and potential data races; This warning will become a runtime crash in a future version of SwiftUI.`

CustomerSheet+SwiftUI.swift

<img width="2874" height="674" alt="Image" src="https://github.com/user-attachments/assets/658cbdc0-5524-4a71-b312-b36d2471e5c0" />

## Code to reproduce

Present Customer Sheet but provide invalid ephemeral key or setup intent secret, causing completion to be called with Error

## iOS version

iOS 26

## Installation method

SPM 

## SDK version

25.5.0

porter-stripe:
@cvillanueva10 Thanks for reporting this! I have submitted a fix which should be included in an upcoming SDK release.

porter-stripe:
Now fixed in https://github.com/stripe/stripe-ios/releases/tag/25.6.2!

PR #6049, opened 2026-02-09
Fix MainActor warnings

porter-stripe:
## Summary
- Resolves the MainActor warning detailed in the below github issue

## Motivation
https://github.com/stripe/stripe-ios/issues/6013

## Testing
- Manual
- CI

## Changelog
See diff

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@
 * [Fixed] Fixed build for visionOS 26.2 SDK.
 * [Fixed] Fixed a Swift 6 actor isolation issue in the SwiftUI integration that could cause a crash when presenting PaymentSheet or CustomerSheet.
 
+### Payments
+* [Fixed] Fixed a Swift 6 actor isolation issue in the SwiftUI integration for `paymentConfirmationSheet` and `setupIntentConfirmationSheet`.
+
 ## 25.6.1 2026-02-03
 
 ### Identity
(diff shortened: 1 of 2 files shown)


porter-stripe:
1. > Does Coordinator.parent.presented always need to be set in the MainActor context, or are there instances where it doesn't need to?

It depends where the assignment occurs

Already on MainActor:
- Assignments in the presented property's didSet block (e.g., lines 281, 347 in PaymentSheet+SwiftUI.swift and lines 77, 83 in CustomerSheet+SwiftUI.swift) are called from updateUIView, which SwiftUI invokes on the main thread.

Needs MainActor wrapper:
- Assignments in completion handlers from async methods like present(from:completion:), confirm(from:completion:), and presentPaymentOptions(from:completion:). These callbacks may be invoked from a non-MainActor context, which is what this fix addresses.

2. presentConfirmationSheet in STPPaymentConfirmation+SwiftUI.swift
Yeah those have the same issue, this PR was originally just scoped to CustomerSheet and PaymentSheet. That file supports confirmation via STPPaymentHandler, but I can add them in here since you flagged.

3. @binding paymentMethodParams and @binding isComplete

These don't have the same issue because:

- paymentCardTextFieldDidChange
- cardFormView(_:didChangeToStateComplete:)

These are UIKit delegate methods, not async completion handlers. UIKit delegate callbacks are invoked synchronously on the main thread when user interaction occurs.

PR #6082, opened 2026-02-20
Directly attach billing details for form_specs payment method

gbirch-stripe:
## Summary
Previously, billing details were attached to the payment method params for form_specs payment methods via the `additionalAPIParameters`. However, this results in different URL % encoding compared with directly attaching. The means that superficial changes (the % encoding doesn't actually affect API behavior) appear when migrating payment methods from form_specs to explicit definitions, making it difficult to review changes.


gbirch-stripe:
Decided against this approach after conversation with wooj

PR #6218, opened 2026-03-17
When available, support adding bank accounts as first link payment method

josephroque-stripe:
## Summary

When available as a supported payment method, allow new link users to add a bank account as a payment method, instead of requiring a card first.

## Motivation

When a user creates a new link account, currently we require they add a card as a first payment method, but this requirement doesn't exist on web. On web, if the checkout supports bank accounts, then the user can add a bank account as a first payment method instead.

https://jira.corp.stripe.com/browse/RUN_LINK_MOBILE-96

## Testing

Manually testing the flow with a new account

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController-WalletViewController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController-WalletViewController.swift
@@ -740,7 +740,7 @@ extension PayWithLinkViewController.WalletViewController: LinkPaymentMethodPicke
         let newPaymentVC = PayWithLinkViewController.NewPaymentViewController(
             linkAccount: linkAccount,
             context: context,
-            isAddingFirstPaymentMethod: false
+            isAddingFirstPaymentMethod: viewModel.paymentMethods.isEmpty
         )
 
         bottomSheetController?.pushContentViewController(newPaymentVC)
--- a/StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController-WalletViewController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController-WalletViewController.swift
@@ -724,7 +724,7 @@ extension PayWithLinkViewController.WalletViewController: LinkPaymentMethodPicke
         paymentPicker.setAddButtonIsLoading(true)
         coordinator?.startFinancialConnections { [weak self] result in
             let completion = {
-                self?.confirmButton.update(status: .enabled)
+                self?.confirmButton.update(status: self?.viewModel.confirmButtonStatus ?? .disabled)
                 self?.paymentPicker.setAddButtonIsLoading(false)
             }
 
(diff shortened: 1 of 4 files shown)


StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController.swift
```diff
@@ -499,31 +499,22 @@ private extension PayWithLinkViewController {
...
+                    self.loadAndPresentWallet()
+                case .canceled:
+                    self.cancel(shouldReturnToPaymentSheet: false)
+                case .failed(let error):
+                    self.finish(withResult: .failed(error: error), deferredIntentConfirmationType: nil)
                 }
-                // Show a loading view while Financial Connections is being prepared
-                viewController = LoaderViewController(context: context)
-            } else {
-                let addPaymentMethodVC = NewPaymentViewController(
```
jkelle-stripe, line 501: Checking my understanding: We're removing this because when more than just bank accounts are supported and the user has no payment methods, we'll just show the wallet (line 519) so they can add any of the supported types. Is that right?

mats-stripe:
Looks good, but I found an issue when tapping the Pay button when there are no payment methods. Most likely wasn't introduced in this PR, but because this PR starts showing the empty wallet screen much more often, it feels like something we should fix here:

https://github.com/user-attachments/assets/b4db7c98-62f0-450d-89d5-6f6ca7e3dfae

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController.swift
josephroque-stripe: @jkelle-stripe exactly! the wallet can handle adding all supported types in this case. The `NewPaymentViewController` acts a bit janky in this case, the wallet view is more capable

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController-WalletViewController.swift
```diff
@@ -723,7 +724,7 @@ extension PayWithLinkViewController.WalletViewController: LinkPaymentMethodPicke
         paymentPicker.setAddButtonIsLoading(true)
         coordinator?.startFinancialConnections { [weak self] result in
             let completion = {
-                self?.confirmButton.update(status: .enabled)
+                self?.confirmButton.update(status: self?.viewModel.confirmButtonStatus ?? .disabled)
```
josephroque-stripe, line 727: Resolves another bug that came up: dismissing the financial connections view would set the confirm button to be enabled unconditionally, even though we had no payment methods, and tapping it would hit an `assertionFailure`

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController-WalletViewController.swift
```diff
@@ -57,7 +57,8 @@ extension PayWithLinkViewController {
             compact: viewModel.shouldUseCompactConfirmButton,
             linkAppearance: viewModel.linkAppearance,
             didTapWhenDisabled: { [weak self] in
-                self?.cardDetailsRecollectionSection.showAllValidationErrors()
+                guard let self, self.viewModel.shouldShowRecollectionSection else { return }
```
josephroque-stripe, line 60: @mats-stripe fixes the issue you mentioned, where tapping the disabled continue button would show the card expiration error. Technically, only either this or the change made in `PaymentElementMethodWrapper.swift` (to only show validation errors for visible fields) is required. The latter change is definitely required, for the case when cvc recollection is **required**, and expiry date update is **not**, the expiry date validation error **will show** due to `shouldShowRecollectionSection` being true, both validation errors will be calculated. 

I think it makes sense to keep both changes, as this prevents unnecessary work, but the other change actually fixes the underlying issue in all cases

mats-stripe (approved):

PR #6237, opened 2026-03-21
increase testSavedSepa network response allowance 10->15 seconds

gbirch-stripe:
## Summary
testSavedSepa has been failing in the nightly no-mocks CI run with a timeout failure. this happens inconsistently, indicating that it is a true timeout. increasing the allowance to see if this corrects it


joyceqin-stripe (commented):
I think a [change](https://github.com/stripe/stripe-ios/pull/6213) I made recently may have impacted this 😅 Previously, this test used a hard-coded customer, but every so often, the no-mocks tests would start failing due to the customer reaching the max limit for payment methods attached. I made a change when we ran into that issue again to create a new customer and payment method to attach to that customer instead of reusing a hardcoded one.

I think increasing the timeout should fix it most of the time, but if you look at run 2 of [this failure](https://app.bitrise.io/build/1e01e603-c44b-46ca-bc3b-ff43e4e71518?tab=tests&tests_filter_status=failed), you see "PaymentSheet.confirm failed - Error creating PaymentIntent: This object cannot be accessed right now because another API request or Stripe process is currently accessing it." I think maybe we need a mechanism before the PaymentSheet confirm intent loop to make sure previous API calls are finished before making another, maybe just a sleep or something?

PR #6238, opened 2026-03-23
Add Pay by Bank support

joyceqin-stripe:
## Summary

Add Pay by Bank support
## Motivation

Add Pay by Bank support
## Testing

Unit tests

https://github.com/user-attachments/assets/859bda6b-5a9c-4261-a866-f82e39455642

https://github.com/user-attachments/assets/927ea9b2-cedd-4d0d-8800-b033ef609949

## Changelog

- [Added] Added support for Pay by Bank payments (GA in GB, private preview in EU).


StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetLoaderTest.swift
```diff
@@ -978,8 +978,8 @@ final class PaymentSheetLoaderTest: STPNetworkStubbingTestCase {
 
         // Test successful load with valid payment method options
         let all_payment_methods_pmo_sfu_values: [STPPaymentMethodType: PaymentSheet.IntentConfiguration.SetupFutureUsage] = STPPaymentMethodType.allCases.reduce([:]) { partialResult, type in
-            // Skip unknown and wero (private preview, not yet recognized by /v1/elements/sessions)
-            guard type != .unknown, type != .wero else { return partialResult }
+            // Skip unknown, wero, and pay_by_bank (private preview, not yet recognized by /v1/elements/sessions)
+            guard type != .unknown, type != .wero, type != .payByBank else { return partialResult }
```
porter-stripe: nit: Let's make a list of PMs that should be excluded here now that we have more than 1.

StripePaymentSheet/StripePaymentSheet/Resources/JSON/form_specs.json
```diff
@@ -1132,5 +1132,24 @@
                 }
             }
         }
+    },
+    {
+        "type": "pay_by_bank",
```
porter-stripe, line 1137: do we actually need this entry in the form specs? I don't think we should use LUXE specs for new LPMs.

StripePayments/StripePayments/Source/API Bindings/Models/PaymentMethods/Types/STPPaymentMethodPayByBank.swift
```diff
@@ -0,0 +1,42 @@
...
+    @objc public override var description: String {
+        let props = [
+            // Object
+            String(format: "%@: %p", NSStringFromClass(STPPaymentMethodPayByBank.self), self)
+        ]
+
+        return "<\(props.joined(separator: "; "))>"
+    }
+
+    // MARK: - STPAPIResponseDecodeable
```
porter-stripe: Nit: STPAPIResponseDecodeable -> STPAPIResponseDecodable (extra 'e'). This typo exists in several other LPM files too, so not blocking.

Stripe/StripeiOSTests/STPPaymentMethodPayByBankTests.swift
```diff
@@ -0,0 +1,39 @@
...
+    static let payByBankPaymentIntentClientSecret = "pi_3TBdzVGoesj9fw9Q02JPHHQc_secret_NDNsHku6A1x1TNaAfVA4rhU6d"
+
+    func _retrievePayByBankJSON(_ completion: @escaping ([AnyHashable: Any]?) -> Void) {
+        let client = STPAPIClient(publishableKey: STPTestingGBPublishableKey)
+        client.retrievePaymentIntent(
+            withClientSecret: Self.payByBankPaymentIntentClientSecret,
+            expand: ["payment_method"]
+        ) { paymentIntent, _ in
+            XCTAssertNotNil(paymentIntent?.paymentMethod?.allResponseFields["pay_by_bank"])
+            let payByBankJson = try? XCTUnwrap(paymentIntent?.paymentMethod?.payByBank?.allResponseFields)
```
porter-stripe: try? XCTUnwrap(...) silently swallows failures — XCTUnwrap should be used with try (not try?) so the test fails with a useful message if payByBank is nil. (Same issue exists in the Wero test, but let's not copy it
 forward.)

porter-stripe (commented):
few comments

StripePaymentSheet/StripePaymentSheet/Resources/JSON/form_specs.json
joyceqin-stripe: This LPM has no input fields, so either we have this entry in the form specs or we add a case in PaymentSheetFormFactory.make that just returns an empty FormElement; otherwise, we hit the stpAssertionFailure that it's missing form specs. I don't see any other LPMs returning the empty FormElement, so I thought it made sense to keep an entry here, copied from https://stripe.sourcegraphcloud.com/r/stripe-internal/mint/-/blob/pay-server/lib/lpm_ui_platform/private/specs/pay_by_bank.json

porter-stripe (approved):

PR #6243, opened 2026-03-23
[Identity] Adding in new error logs and improving many existing ones

Pearapps:
This adds in a variety of new error logs so that we can gather more information on the error states in the identity flow.

Additionally, this adds in a bunch more details to existing and new logs to help us figure out more when we go to read the logs.

Went through a handful of scenarios to make sure I saw the analytics log logged to console (which we have enabled in debug)


StripeIdentity/StripeIdentity/Source/NativeComponents/ML/Helpers/MLModelLoader.swift
```diff
@@ -141,3 +153,22 @@ final class MLModelLoader {
         }
     }
 }
+
+private extension MLModelLoader {
+    static func logModelLoadingError(
```
cjmisenas-stripe, line 158: Does this also log which models had a loading error?

cjmisenas-stripe (changes requested):

StripeIdentity/StripeIdentity/Source/NativeComponents/ML/Helpers/MLModelLoader.swift
Pearapps: Not in this helper. The higher-level IdentityMLModelLoader logs include ml_model_type (document / face / shared), though.

cjmisenas-stripe (approved):

PR #6266, opened 2026-03-27
Remove version comparison in changelog_utils

amk-stripe:
## Summary

Remove version comparison in changelog_utils

## Motivation

- The deploy-dry-run job is failing because of the '>' comparison between arrays ([run ticket](https://jira.corp.stripe.com/browse/RUN_MOBILESDK-5267)).
- We don't need to compare versions in this step, because they are already compared in the preceding step.

This won't actually fix the deploy-dry-run but it does fix the release script, so I think we should merge it to unblock Monday's release

### DIFF
--- a/ci_scripts/changelog_utils.rb
+++ b/ci_scripts/changelog_utils.rb
@@ -71,8 +71,6 @@ module ChangelogUtils
     previous_parts = parse_version(previous_version)
     next_parts = parse_version(next_version)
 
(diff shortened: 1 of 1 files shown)


amk-stripe:
Doesn't fix the issue bc now the changelog verification fails...

gbirch-stripe:
We should do this to actually fix the issue: https://github.com/stripe/stripe-ios/pull/6271

amk-stripe:
fixed in another PR

PR #6277, opened 2026-03-31
[Identity] Update example app api to use new endpoint

Pearapps:
Maintains current functionality in the identity example app but with the new example API endpoint.

I tested with the happy paths on the native and web ui - and a bunch of playground options!


Example/IdentityVerification Example/IdentityVerification Example/ExampleVerificationViewController.swift
```diff
@@ -41,10 +41,17 @@ class ExampleVerificationViewController: UIViewController {
         // Make request to our verification endpoint
         let session = URLSession.shared
         let url = URL(string: baseURL + verifyEndpoint)!
+        let requestJson = try? JSONSerialization.data(
+            withJSONObject: [
+                "type": "document",
```
cjmisenas-stripe, line 46: Does this always set the type to `document` and that is intentional/only document verification is what this `ExampleVerificationViewController` does or this just falls back to `document` type?

Pearapps: Intentional,`ExampleVerificationViewController` is our fixed document-only sample, so moving from `/create-verification-session` to `/verification-sessions` means we now pass `type=document` explicitly, while the Playground is still the example where the type varies.

cjmisenas-stripe (approved):

PR #6300, opened 2026-04-06
[Identity] Updating the horizontal padding of the title in consent header 

Pearapps:
Bumps up the padding horizontally padding of the title on consent page.

Before: 

Screenshot 2026-03-24 at 2 35 05 PM

After:

Screenshot 2026-04-06 at 12 03 55 PM


cjmisenas-stripe (dismissed):

Tests/ReferenceImages_64/StripeIdentityTests.IndividualWelcomeViewControllerSnapshotTest/testViewIsConfiguredFromAPIAndReturnError_@3x.png
cjmisenas-stripe, line 1: I'm assuming that the padding of the title should match the padding of the content right? It looks like this does not match at 3X view?

cjmisenas-stripe (changes requested):

Tests/ReferenceImages_64/StripeIdentityTests.IndividualWelcomeViewControllerSnapshotTest/testViewIsConfiguredFromAPIAndReturnError_@3x.png
Pearapps: > I'm assuming that the padding of the title should match the padding of the content right? It looks like this does not match at 3X view?

This is actually matching, it just does not always perfectly align because the title is center aligned and the body is leading/left aligned - and line breaks as well. But yeah this should be good to go

mnn-stripe (approved):

cjmisenas-stripe (approved):

PR #6322, opened 2026-04-09
StripeCryptoOnramp SDK: Better Cancellation Signaling

Twigz:
## Summary
* Adds new `didSelectPaymentOption` `Bool` to signal from Link to Onramp that a payment option was selected or not. This helps CryptoOnramp know if a payment option was selected as part of collection, or was just preserved as part of a previous selection.
* Adds new state property `paymentSourceBeforeApplePay` to track an existing payment source during the Apple Pay flow, so that we can preserve the accuracy of `selectedPaymentSource` if the multi-step apple pay flow is not completed.

## Motivation
Making iOS payment collection more in line with how Android handles cancellation.

## Testing
1. Login.
2. Go to collect a card.
3. Cancel, make sure it receives a cancellation result.
4. Collect a card, and successfully collect one.
5. Go to collect a card, cancel, make sure you receive a cancellation result, but preserve previously selected card.
6. Make sure checkout still works.
7. Run these steps again, but with Apple Pay
8. Run these steps again, but switch back and forth between Link and Apple Pay.


jeanregisser (commented):
Thanks, I'll finish reviewing tomorrow,
Note that there's also this proposition: https://github.com/stripe/stripe-ios/pull/6354
We can discuss during our sync meeting.

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
```diff
@@ -174,6 +174,14 @@ public final class CryptoOnrampCoordinator: NSObject, CryptoOnrampCoordinatorPro
...
     private let analyticsClient: CryptoOnrampAnalyticsClient
     private var applePayCompletionContinuation: CheckedContinuation<ApplePayPaymentStatus, Swift.Error>?
+
+    /// Temporary snapshot of the selection that existed before starting an Apple Pay collection flow.
+    ///
+    /// We set this immediately before presenting Apple Pay so we can restore `selectedPaymentSource`
+    /// if the Apple Pay flow is canceled or fails after `didCreatePaymentMethod` has already updated
+    /// `selectedPaymentSource`. This value should only represent pre-existing selection state for the
+    /// active Apple Pay attempt, and must be cleared once that attempt completes, fails, or the user logs out.
+    private var paymentSourceBeforeApplePay: SelectedPaymentSource?
```
jeanregisser: Have we considered flipping the direction here? Instead of saving what we had before and restoring on cancel, we could use a `pendingApplePayPaymentSource` that only gets promoted to `selectedPaymentSource` in `didCompleteWith(.success)` - leaving it untouched on cancel/error, with nothing to restore. It'd also make cleanup symmetric: just `pendingApplePayPaymentSource = nil` at the end of `didCompleteWith` in all cases.

jeanregisser (commented):
Nice, this is a clean fix. One thought on the Apple Pay state. ✨

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
mliberatore: Good idea, hadn't thought about it that way. It reduces the overall management of that state. I‘ve implemented the suggestion in [73a636f](https://github.com/stripe/stripe-ios/pull/6322/commits/73a636fd45623ec38ca0e6814be44db553e6d01d).

jeanregisser (approved):
🚀

PR #6323, opened 2026-04-09
Add Checkout.Configuration

porter-stripe:
## Summary
- Adds `Checkout.Configuration` with an adaptive pricing API to allow merchants to opt in to Adaptive Pricing.

## Motivation
- https://docs.google.com/document/d/1CWglcOHR6I8cOWKjztPfnhnVG8axucpv-j5cWIoey7E/edit?tab=t.ixyh82ib6clh

## Testing
- Manual

## Changelog
N/A


joyceqin-stripe (commented):
Do we want to add some tests where adaptivePricingAllowed is false?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Configuration.swift
```diff
@@ -0,0 +1,60 @@
...
+    ///     configuration: config
+    /// )
+    /// ```
+    public struct Configuration {
+        /// Controls whether adaptive pricing is requested for this session.
+        ///
+        /// When allowed, Stripe may present prices in the customer's local
+        /// currency alongside the merchant's settlement currency.
+        ///
+        /// Default: ``AdaptivePricing/init()`` (`allowed: true`).
```
joyceqin-stripe, line 31: nit: .init instead of /

porter-stripe:
@joyceqin-stripe good point, I've changed those back to false, and added new tests w/ new network recordings to ensure we are sending it and getting the correct behavior.

joyceqin-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Configuration.swift
porter-stripe: This is interesting, I think `/` is correct in the markdown for Xcode doc comments but I've seen inconsistent conclusions.

PR #6326, opened 2026-04-10
Add `networkBusinessProfile` to seller payment methods

floriansuess-stripe:
# Summary and motivation
[APIREVIEW-4954](https://jira.corp.stripe.com/browse/APIREVIEW-4954).

iOS SDK counterpart to the Stripe.js change in [mint#2041413](https://git.corp.stripe.com/stripe-internal/mint/pull/2041413).

Exposes the API surface for agents to pass a seller's Business Network profile to Payment Element on iOS, enabling PE to display the seller's configured payment methods. Beta header `payment_element_seller_payment_methods_beta_1` is automatically included when `networkBusinessProfile` is provided.

The `/v1/elements/sessions` request already supports:

`GET /v1/elements/sessions`
```javascript
<<< REQUEST <<<
{
   seller_details?: {
      network_id?: string,                // was required
      external_id?: string,               // was required
      business_name?: string,             // was required (local-only, now also sent to API)
      network_business_profile?: string,  // new
   },
}
```

In the iOS SDK:

```swift
@_spi(SharedPaymentToken) public struct SellerDetails {
    public let networkId: String?              // was required
    public let externalId: String?             // was required
    public let businessName: String?           // was required
    public let networkBusinessProfile: String? // new
}
```

For backwards compatibility, the existing 3-arg initializer `SellerDetails(networkId:externalId:businessName:)` with non-optional String types is preserved. A new initializer with all-optional parameters enables flexible usage including the new `networkBusinessProfile` field.

The new `network_business_profile` field allows agents to specify a BizNet profile ID, from which Stripe resolves the seller's default PMC (provided `payment_element_seller_payment_methods_beta_1` is present).

All fields made optional. `networkBusinessProfile` added. `paymentMethodConfigurationId` and `sellerDetails` are now validated as mutually exclusive, passing both returns a `PaymentSheetError` client-side.

All this together looks like:

```swift
// 1. Agent creates intent configuration with seller's BizNet profile
@_spi(SharedPaymentToken) import StripePaymentSheet

let intentConfig = PaymentSheet.IntentConfiguration(
    sharedPaymentTokenSessionWithMode: .payment(amount: 1000, currency: "usd"),
    sellerDetails: .init(
        networkBusinessProfile: "bnp_abc123"  // the seller's profile
    ),
    preparePaymentMethodHandler: { paymentMethod, shippingAddress in
        // handle payment
    }
)

// 2. Payment Element renders the SELLER's payment methods
// The SDK automatically includes the beta header in the elements/sessions request
```

Note: what we do not do in this PR is act on it, i.e., take that network_business_profile value, resolve it to a seller's PMC, and return the right payment methods in the session response. That will come later.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetIntentConfiguration.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetIntentConfiguration.swift
@@ -58,14 +58,30 @@ public extension PaymentSheet {
 
         /// Seller details for facilitated payment sessions
         @_spi(SharedPaymentToken) public struct SellerDetails {
-            public let networkId: String
-            public let externalId: String
-            public let businessName: String
+            public let networkId: String?
+            public let externalId: String?
+            public let businessName: String?
+            public let networkBusinessProfile: String?
 
+            /// Backwards-compatible initializer with required fields
             public init(networkId: String, externalId: String, businessName: String) {
                 self.networkId = networkId
                 self.externalId = externalId
                 self.businessName = businessName
+                self.networkBusinessProfile = nil
+            }
+
+            /// Initializer with all optional fields including networkBusinessProfile
+            public init(
+                networkId: String? = nil,
+                externalId: String? = nil,
+                businessName: String? = nil,
+                networkBusinessProfile: String? = nil
+            ) {
+                self.networkId = networkId
+                self.externalId = externalId
+                self.businessName = businessName
+                self.networkBusinessProfile = networkBusinessProfile
             }
         }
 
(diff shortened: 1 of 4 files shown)


paulcarey-stripe (dismissed):
Looks good from LPMs.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetIntentConfiguration.swift
```diff
@@ -58,14 +58,30 @@ public extension PaymentSheet {
...
+            public let businessName: String?
+            public let networkBusinessProfile: String?
 
+            /// Backwards-compatible initializer with required fields
             public init(networkId: String, externalId: String, businessName: String) {
                 self.networkId = networkId
                 self.externalId = externalId
                 self.businessName = businessName
+                self.networkBusinessProfile = nil
+            }
```
toluo-stripe: I think we can delete this because the new constructor has a default value for `networkBusinessProfile`

floriansuess-stripe: Thanks @tulio-stripe, have addressed.

tulio-stripe: Thanks for updating it 🤣

toluo-stripe (commented):
LGTM, but I'm not an iOS codeowner. There's also an api change that may be a blocker. @wooj-stripe @yuki-stripe can you take a look?

PR #6327, opened 2026-04-10
Add new CheckoutSession settings pane

porter-stripe:
## Summary
- Adds all the supported CheckoutSessions settings on our backend to the playground so we can configure a checkout session

## Motivation
- CheckoutSessions

## Testing
- Manual

## Changelog
N/A

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
@@ -762,6 +762,16 @@ import UIKit
         }))
         rootViewController.present(vc, animated: true, completion: nil)
     }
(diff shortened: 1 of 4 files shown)


Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
```diff
@@ -762,6 +762,16 @@ import UIKit
         }))
         rootViewController.present(vc, animated: true, completion: nil)
     }
+    func checkoutSessionSettingsTapped() {
+        if #available(iOS 15.0, *) {
```
wooj-stripe, line 766: I guess it's okay if we silently fail here? Maybe we throw up an error if users are on ios 13 or 14?

wooj-stripe (approved):

PR #6337, opened 2026-04-14
Send is_saved_payment_method in MPE payment success/failure analytics

yuki-stripe:
## Summary
.

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-4425

## Testing
Manually tested PS/FC/Embedded + unit test

## Changelog
Not user facing

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Analytics/PaymentSheetAnalyticsHelper.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Analytics/PaymentSheetAnalyticsHelper.swift
@@ -410,6 +410,11 @@ final class PaymentSheetAnalyticsHelper {
             }
         }()
         var params: [String: Any] = [:]
+        if case .saved = paymentOption {
+            params["is_saved_payment_method"] = true
+        } else {
+            params["is_saved_payment_method"] = false
+        }
         if case let .new(confirmParams) = paymentOption {
             if let setAsDefault = confirmParams.setAsDefaultPM {
                 params["set_as_default"] = setAsDefault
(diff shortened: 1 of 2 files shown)


StripePaymentSheet/StripePaymentSheet/Source/Analytics/PaymentSheetAnalyticsHelper.swift
```diff
@@ -410,6 +410,11 @@ final class PaymentSheetAnalyticsHelper {
             }
         }()
         var params: [String: Any] = [:]
+        if case .saved = paymentOption {
+            params["is_saved_payment_method"] = true
+        } else {
+            params["is_saved_payment_method"] = false
+        }
```
wooj-stripe, line 417: Why not send something like `payment_option_type`, where the type could be new, external, saved, applepay, link?

yuki-stripe: What's the benefit?

wooj-stripe: Feels like it would be more future proof in the event we introduce a new paymentOption type?

yuki-stripe: Hmm possibly, it feels hard to judge without knowing what that future would be. I'd prefer to keep this targeted improvement if you don't feel strongly?

wooj-stripe (approved):

PR #6338, opened 2026-04-14
Fix vending card art by fixing pre-caching image

wooj-stripe:
## Summary
Fixes vending card art through our APIs by prefetching on load or when accessed on CustomerSheet retrievePaymentOptionSelection() api.  When vending, APIs return back what we currently have in cache. Approach is "best-effort" to reduce impact on loading payment sheet/customersheet.

## Motivation
Currently working only if the image happens to make itself into the cache

## Testing
Updated tests & manual testing

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSheet.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSheet.swift
@@ -373,6 +373,7 @@ extension CustomerSheet {
                 guard let matchingPaymentMethod = paymentMethods.first(where: { $0.stripeId == paymentMethodId }) else {
                     return nil
                 }
+                matchingPaymentMethod.preloadCardArtImage(cardArtEnabled: configuration.appearance.cardArtEnabled)
                 return CustomerSheet.PaymentOptionSelection.paymentMethod(matchingPaymentMethod)
             default:
                 return nil
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentOption+Images.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentOption+Images.swift
@@ -150,7 +150,7 @@ extension STPPaymentMethod {
     }
 
     func cachedCardArtImage(cardArtEnabled: Bool, downloadManager: DownloadManager = DownloadManager.sharedManager) -> UIImage? {
-        guard let cardArtURL = vendedCardArtURL(cardArtEnabled: cardArtEnabled) else {
+        guard let cardArtURL = cardArtCDNURL(cardArtEnabled: cardArtEnabled) else {
             return nil
         }
         let placeholder = downloadManager.imagePlaceHolder()
@@ -162,23 +162,20 @@ extension STPPaymentMethod {
         return image == placeholder ? nil : image.roundedWithBorder(radius: 3)
     }
 
+    // Populates the in-memory card art cache. Promotes from disk cache if available,
+    // then fires a best-effort network request to fetch the latest image.
     func preloadCardArtImage(cardArtEnabled: Bool, downloadManager: DownloadManager = DownloadManager.sharedManager) {
-        guard let cardArtURL = vendedCardArtURL(cardArtEnabled: cardArtEnabled) else {
+        guard let cardArtURL = cardArtCDNURL(cardArtEnabled: cardArtEnabled) else {
             return
         }
+        // Passing a non-nil updateHandler triggers a best-effort network
+        // download that refreshes the in-memory cache.
         let updateHandler: ((UIImage) -> Void)? = { _ in }
         _ = downloadManager.downloadImage(
             url: cardArtURL,
             placeholder: nil,
             updateHandler: updateHandler
         )
-    }
-
-    func vendedCardArtURL(cardArtEnabled: Bool) -> URL? {
-        guard cardArtEnabled, let cardArtURL = cardArtCDNURL(height: 26) else {
-            return nil
-        }
-        return cardArtURL
     }
 }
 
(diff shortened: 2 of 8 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSheet.swift
```diff
@@ -373,6 +373,7 @@ extension CustomerSheet {
                 guard let matchingPaymentMethod = paymentMethods.first(where: { $0.stripeId == paymentMethodId }) else {
                     return nil
                 }
+                matchingPaymentMethod.preloadCardArtImage(cardArtEnabled: configuration.appearance.cardArtEnabled)
```
yuki-stripe, line 376: I thought the idea was to block this on downloading the card art, since it's already async?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentOption+Images.swift
```diff
@@ -161,6 +161,25 @@ extension STPPaymentMethod {
         )
         return image == placeholder ? nil : image.roundedWithBorder(radius: 3)
     }
+
+    func preloadCardArtImage(cardArtEnabled: Bool, downloadManager: DownloadManager = DownloadManager.sharedManager) {
+        guard let cardArtURL = vendedCardArtURL(cardArtEnabled: cardArtEnabled) else {
+            return
+        }
+        let updateHandler: ((UIImage) -> Void)? = { _ in }
```
yuki-stripe, line 173: This is strange, can we add a comment explaining why we're passing an empty closure?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentOption+Images.swift
```diff
@@ -161,6 +161,25 @@ extension STPPaymentMethod {
...
+        _ = downloadManager.downloadImage(
+            url: cardArtURL,
+            placeholder: nil,
+            updateHandler: updateHandler
+        )
+    }
+
+    func vendedCardArtURL(cardArtEnabled: Bool) -> URL? {
+        guard cardArtEnabled, let cardArtURL = cardArtCDNURL(height: 26) else {
```
yuki-stripe: Each height caches a separate image, right? That seems unfortunate (and seems like it might actually be the source of the bug we found where you could open up Embedded/FC, _see_ the card art, and then not have it vended?). Would it be better to just request the bigger image and then resize instead? 

I also see we've changed the height from 28 to 26, just noting in case that's not intentional.

yuki-stripe: The difference between this `vendedCardArtURL` and `cardArtCDNURL` feels a bit unclear (when would I use one or the other?). Maybe just a docstring on what the purpose of this method is? It's responsible for the URL of an image to be vended to merchants in CustomerSheet/FC/Embedded.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift
```diff
@@ -219,6 +219,12 @@ final class PaymentSheetLoader {
                 elementsSession: elementsSession,
                 defaultPaymentMethod: elementsSession.customer?.getDefaultPaymentMethod()
             )
+
+            // Pre-fetch card art without waiting for this to finish
```
yuki-stripe: I think you said pre-loading payment option images in general was broken, is that still true?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSheet.swift
wooj-stripe: Doing so would slow down this function as well as be inconsistent with the behavior of payment sheet, so I ultimately decided against doing this.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentOption+Images.swift
wooj-stripe: Yes, each height caches a separate image.

Good idea. I was trying to get around downloading the most optimal image, but it think it does make things easier on us and the CDN to just standardize on the larger image size for all cases -- and yes, we've updated the 'larger' size to have a height of 26, which is intentional

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift
wooj-stripe: Correct in FlowController.swift,  `_ = flowController.paymentOption` does not seem to have any effect. That being said the images do seem to be preloaded by the creation of rows. I think this is something we should clean up in the future.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentOption+Images.swift
wooj-stripe: Removed this function and consolidated.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSheet.swift
yuki-stripe: Hrm I'm not following. The method's purpose is to asynchronously retrieve the PM's image, right?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift
yuki-stripe: So the images _are_ preloaded today, just not card art, and that's why we're adding this code? The thought behind this is - wondering if we can just generically preload images rather than special-case card art.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/CustomerSheet/CustomerSheet.swift
wooj-stripe: We're doing two things with this call, which is effectively what we're doing for payment sheet.

1. We're warming up the in-memory cache. This could mean it gets loaded synchronously from disk into memory.  This doesn't take much time, so if it's on disk, seems like we should promote it into memory, so that by the time we access it the next time, we would vend it.

2. Kicking off a download asynchronously to get the most up to date image. E.g. it's possible the image has changed on the backend, so we need to fetch it, and store it, but I don't think we should slow down this method by waiting for the most up-to-date version.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift
wooj-stripe: Updated comment per offline discussion

yuki-stripe (approved):

PR #6340, opened 2026-04-15
Update LinkController to use the apiClient from the caller instead of the shared one

rui-sofi:
## Summary
Configuration should use the apiClient passed in from the call instead of the shared one.

## Motivation
There are cases where an app might connect to multiple accounts and thus use multiple instances of apiClient.  When LinkController is created, it needs to use the apiClient passed in from the caller instead of relying on the shared apiClient.

## Testing
Manually tested.  This change won't have an impact if the app uses one apiClient.


porter-stripe (dismissed):

CHANGELOG.md
```diff
@@ -3,6 +3,9 @@ PATCH
 
 ## X.Y.Z - changes pending release
 
+### PaymentSheet
```
porter-stripe, line 6: Nit: LinkController is not part of PaymentSheet.

porter-stripe: Although it's in the module...

porter-stripe (approved):

PR #6342, opened 2026-04-15
Render Unknown Payment Methods in Link

jkelle-stripe:
## Summary
Display unknown payment methods in native Link by parsing the `display` metadata included in each `redacted_payment_details` entry. Expected structure includes looks like:

```
display: {
  label: 'Card nickname',
  icon: {
    default: 'cdn.stripe.com/.../amex.png'
  },
  sublabel: '**** 1234'
}
```

## Demos

<details><summary>Connecting a New PM</summary>
<p>

**Note**: The payment method previously existed (from creation on web), but is being used on a different device, so reconnection is needed.

https://github.com/user-attachments/assets/242cd4ab-26de-46a7-92f6-d466dd46f181

</p>
</details> 

<details><summary>Using an Existing PM</summary>
<p>

https://github.com/user-attachments/assets/3351ea1b-62f3-4c66-98dc-69bace66cec6

</p>
</details>


StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/PaymentDetails.swift
```diff
@@ -112,7 +117,8 @@ extension ConsumerPaymentDetails {
             // These are US bank accounts, so only check for US country code
             return allowedCountries.contains("US")
         case .unparsable:
-            return false
+            // Unknown types don't have country info; allow them if display metadata is present.
+            return display != nil
```
jkelle-stripe: My thought process here is that if they're not allowed, they would have been filtered by the backend (eg. for `brl` currencies, Pix is presented and not Crypto).

tillh-stripe (commented):
I haven't yet had the time to take a look myself, but a few things that Codex has been flagging:
1. Billing details recollection (when a merchant requests full billing details) might be triggered for an unknown type, but `createUpdateDetails` would fail for unknown types.
2. The `setLinkPaymentDetails` method in `ElementsCustomer.swift` wouldn't be able to render non-card and non-bank payment methods.
3. Do we need to update the logic in `getSupportedPaymentDetailsTypes` now?
4. The delete dialogs in `removePaymentMethod` uses an empty title for unknown types, which we don't want.

A quick scan of the codebase suggests to me that these are true — can you confirm that?

jkelle-stripe:
@tillh-stripe Thanks for flagging these!
1. Addressed in a0ccac09af11d312f4f1eb01e4737257d893624b
2. Working on this now. 
3. This function was updated before your comment. Do we think it needs more work?
4. Addressed in 57b8b5a4d1d38052b7fc1029590249509e2c8199


| Card | Bank | Crypto |
|--------|--------|--------|
|  |  |  |

StripePaymentSheet/StripePaymentSheet/Resources/Localizations/en.lproj/Localizable.strings
```diff
@@ -342,12 +339,22 @@ e.g, 'Pay faster at Example, Inc. and thousands of businesses.' */
...
+
 /* Title for confirmation alert to remove a card */
 "Remove card?" = "Remove card?";
 
 /* Title for a button that when tapped removes a linked bank account. */
 "Remove linked account" = "Remove linked account";
 
+/* Title for a button that when tapped removes a payment method.
+   Title for confirmation prompt when removing a saved payment method without a display label. */
+"Remove payment method" = "Remove payment method";
```
tillh-stripe, line 367: I wonder if we should just show `Remove` in the buttons, as the available space is limited. Not immediately required, but let's follow up with that?

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/PaymentDetails.swift
```diff
@@ -24,19 +24,22 @@ final class ConsumerPaymentDetails: Decodable {
     let billingAddress: BillingAddress?
     let billingEmailAddress: String?
     let nickname: String?
+    let display: DisplayMetadata?
```
tillh-stripe, line 27: If I remember correctly, we talked about this becoming non-nullable? Is that still the plan for later on?

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/PaymentDetails.swift
tillh-stripe: I'm not sure what the right logic is here, but I don't think this one is it. Maybe it needs to be similar to the card case and just look at billing address?

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/PaymentDetails.swift
```diff
@@ -375,7 +394,9 @@ extension ConsumerPaymentDetails {
         case .bankAccount(let bankAccount):
             return bankAccount.displayName(with: nickname)
         case .unparsable:
```
tillh-stripe, line 393: We should rename `unparsable`, but that's for a later pull request.

CHANGELOG.md
```diff
@@ -3,6 +3,9 @@ MINOR
 
 ## X.Y.Z - changes pending release
 
+### PaymentSheet
+* [Added] Added forward-compatiblity for backend-provided payment methods in the native Link payment sheet.
```
tillh-stripe: We should change this by either writing something that feels more relevant to the merchant (“Added support for new and upcoming payment methods in Link, such as UPI and Pix“) or by removing it entirely. I find that this is an implementation detail, so in favor of the latter.

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/PaymentDetails.swift
jkelle-stripe: @tillh-stripe Yep! The backend FF is still there so want to make sure that is removed before we remove nullability here.

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/PaymentDetails.swift
jkelle-stripe: @tillh-stripe Resolved in cce6779d68085ebd0d71d5fdafe407d3bc8c163b. Really good catch! 



 
| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/8a4ee47b-d10c-4ebb-851c-ad4c8cbf56e8" width="300" /> | <video src="https://github.com/user-attachments/assets/4632ccc7-6268-4145-b9d9-8e85139ea9d5" width="300" /> |

jkelle-stripe:
@tillh-stripe found a bug with billing address collection not include name for non-cards. Resolved in 7de1bcedc2d448999fdd208c2b7b966434709b83 See in the before that the billing address for a Pix PD created on web has a name and when we pay on MPE with billing address collection, we wipe the name and replace with the provided address. See in the after video, this is resolved.


| Before | After |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/aeacdaa6-c2d9-40bf-9898-89c52fe8316d" width="300" /> | <video src="https://github.com/user-attachments/assets/04cde95e-e098-40d7-9292-31b5a2ec7667" width="300" /> |

jkelle-stripe:
@tillh-stripe This is ready for review! Added back the changelog entry for this upcoming release to flag that a release is needed.

tillh-stripe (dismissed):

tillh-stripe (approved):

PR #6346, opened 2026-04-15
Create LinkAccountSession using the apiClient passed in from the caller instead of the shared one

rui-sofi:
## Summary
Create LinkAccountSession using the apiClient passed in from the caller instead of the shared one

## Motivation
This is to properly handle multiple apiClients/accounts using the same instance of the SDK

## Testing
Manually tested.  This doesn't impact the existing use-case with a single apiClient/account


jeanregisser (approved):
Thanks! 🚀

PR #6348, opened 2026-04-16
Add more metrics for card art

wooj-stripe:
## Summary
Adds additional card art metrics around confirmation and selecting saved pms

## Motivation
Analytics

## Testing
Added tests & manually


StripePaymentSheet/StripePaymentSheet/Source/Analytics/PaymentSheetAnalyticsHelper.swift
```diff
@@ -265,7 +265,11 @@ final class PaymentSheetAnalyticsHelper {
         guard let event else {
             return
         }
-        log(event: event, selectedLPM: selectedLPM)
+        var params: [String: Any] = [:]
+        if case .saved(let paymentMethod) = option {
+            params["has_card_art"] = hasCardArt(cardArtEnabled: configuration.appearance.cardArtEnabled, paymentMethod: paymentMethod)
```
porter-stripe, line 270: Can we just update the Appearance analytic payload to send this and we'd be able to know if on a given session card art is enabled: https://github.com/stripe/stripe-ios/blob/master/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/STPAnalyticsClient%2BPaymentSheet.swift#L65

porter-stripe (commented):
question

Wmc05061970:
What the heck is this

StripePaymentSheet/StripePaymentSheet/Source/Analytics/PaymentSheetAnalyticsHelper.swift
wooj-stripe: `appearance.cardArtEnabled` is only a client side flag to prevent the feature from being released prior to us QA-ing the feature. When we feel the feature is launch-ready, we'll remove the flag from PaymentSheet.Appearance, and then use launch flag/experimentation to dial up/down the experience.

porter-stripe (approved):

PR #6353, opened 2026-04-17
CheckoutSessions — PMO SFU

gbirch-stripe:
## Summary
Add PMO SFU support to CheckoutSessions
Note for testing: many payment methods have different PMO SFU options available as compared to Payment Intents

## Motivation
https://docs.google.com/document/d/1CWglcOHR6I8cOWKjztPfnhnVG8axucpv-j5cWIoey7E/edit?pli=1&tab=t.g4karprubz63

## Testing
- unit tests
- manual testing in playground


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Intent.swift
```diff
@@ -166,9 +166,8 @@ enum Intent {
                 return !setupFutureUsageValues.isEmpty
             }
             return nil
-        case .checkoutSession:
```
porter-stripe, line 169: Referring to line 133 here, can't select it.

For payment mode, it returns the top-level setupFutureUsage only. The setupFutureUsageString property is used in analytics. Is it intentional that this reports only the top-level value? I think this is consistent with others but just want to double check.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheet+APIMockTest.swift
```diff
@@ -537,6 +537,42 @@ final class PaymentSheetAPIMockTest: APIStubbedTestCase {
 
         waitForExpectations(timeout: 10)
     }
+
+    func testCheckoutSessionConfirmWithNonCardPaymentMethodIncludesSavePaymentMethod() {
```
porter-stripe, line 541: The tests in this PR are quite light considering what it's adding. Let's add `PaymentSheetLPMConfirmFlowTests` coverage: we already test PMO+SFU `paymentIntentWithPMOSetupFutureUsagefor` PI/deferred PI, but that suite doesn't test CheckoutSession+PMO SFU

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/STPCheckoutSession.swift
```diff
@@ -465,8 +473,16 @@ extension STPCheckoutSession: STPAPIResponseDecodable {
 // MARK: - Parsing Helpers
 
 extension STPCheckoutSession {
+    var isPaymentMethodOptionsSetupFutureUsageSet: Bool {
+        return !setupFutureUsageForPaymentMethodType.isEmpty
```
porter-stripe, line 477: Nit/question: The existing `STPPaymentMethodOptions.isSetupFutureUsageSet` (line 45-52 of STPPaymentMethodOptions.swift) checks whether any PMO entry contains the key setup_future_usage. Might be a bit safer to align with `STPPaymentMethodOptions`.

porter-stripe (commented):
few comments

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/STPCheckoutSession.swift
gbirch-stripe: Are you suggesting that we use the `payment_method_options` property on the checkout session as the data source for PMO SFU rather than `setup_future_usage_for_payment_method_type` or something else? If the former, I have a question about which to use down in the API team questions list, but for now my intuition is to use the `setup_future_usage_for_payment_method_type` property since it is specific to checkout sessions and what is used by web: https://stripe.sourcegraphcloud.com/deepsearch/c4c1b6ed-ecad-4140-a808-600a15fa1da9#answer-169850

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Intent.swift
gbirch-stripe: Yes, I intentionally made it only report the top-level to align with intent flows

gbirch-stripe:
skipping dead code detection — fetchCheckoutSessionPaymentMode and checkoutSessionAdditionalParameters are used in testing code

porter-stripe (approved):

PR #6355, opened 2026-04-17
Add a shimmering card view when loading card art

wooj-stripe:
## Summary
Adds shimmering view that can be used to indicate an image loading

## Motivation
Card art

## Testing
Manual testing

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentMethodCollectionView.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentMethodCollectionView.swift
@@ -361,11 +361,14 @@ extension SavedPaymentMethodCollectionView {
                         if let cardArtURL = paymentMethod.cardArtCDNURL(cardArtEnabled: cardArtEnabled) {
                             if paymentMethodLogo.tag != cardArtURL.hashValue {
                                 paymentMethodLogo.tag = cardArtURL.hashValue
-                                paymentMethodLogo.image = nil
+                                paymentMethodLogo.image = STPImageLibrary.cardBrandChoiceImage()
+                                paymentMethodLogoHeightConstraint.constant = CGFloat(STPPaymentMethod.cardArtHeight)
+                                paymentMethodLogo.addShimmer()
                             }
                             Task {
                                 let image = try? await DownloadManager.sharedManager.downloadImage(url: cardArtURL)
                                 guard paymentMethodLogo.tag == cardArtURL.hashValue else { return }
+                                paymentMethodLogo.removeShimmer()
                                 if let image {
                                     paymentMethodLogo.image = image.roundedWithBorder(radius: 3)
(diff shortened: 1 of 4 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/ShimmerView.swift
```diff
@@ -0,0 +1,53 @@
+//
+//  ShimmerView.swift
+//  StripePaymentSheet
+//
+
+import UIKit
+
+/// A view that displays a sliding shimmer animation overlay.
+/// Used as a loading indicator while card art images are downloaded.
+class ShimmerView: UIView {
```
porter-stripe, line 10: Should we add some snapshot tests for this view?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentMethodCollectionView.swift
```diff
@@ -361,11 +361,14 @@ extension SavedPaymentMethodCollectionView {
...
                                 paymentMethodLogo.tag = cardArtURL.hashValue
-                                paymentMethodLogo.image = nil
+                                paymentMethodLogo.image = STPImageLibrary.cardBrandChoiceImage()
+                                paymentMethodLogoHeightConstraint.constant = CGFloat(STPPaymentMethod.cardArtHeight)
+                                paymentMethodLogo.addShimmer()
                             }
                             Task {
                                 let image = try? await DownloadManager.sharedManager.downloadImage(url: cardArtURL)
                                 guard paymentMethodLogo.tag == cardArtURL.hashValue else { return }
+                                paymentMethodLogo.removeShimmer()
```
porter-stripe, line 371: Do we need to remove shimmer on line 376?

porter-stripe (commented):
questions

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/ShimmerView.swift
wooj-stripe: I didn't consider it due to the animation.

We could write some unit tests that test to make sure the gradient has various alpha components, but that doesn't seem too valuable?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentMethodCollectionView.swift
wooj-stripe: line 376 is showing me:
```
                                    paymentMethodLogo.image = paymentMethodCellImage
```
Removing the shimmer as proposed in the PR (on line 371) seems appropriate since we've added it prior to loading, and we should be removing it irregardless of the card art image loads or we fallback to the network logo.

porter-stripe (approved):

PR #6364, opened 2026-04-22
CryptoOnramp SDK: Exposes StripeCryptoOnramp via SPM

### LINKED ISSUE #6363: [BUG] [P1] StripeCryptoOnramp is referenced in iOS onramp docs but not exposed by stripe-ios-spm 25.11.0
**Summary**
`StripeCryptoOnramp` is referenced in the native iOS Crypto Onramp docs, but it is not exposed as a selectable product from `stripe-ios-spm` version 25.11.0.

**Code to reproduce**
In Xcode, select File > Add Package Dependencies...
Enter https://github.com/stripe/stripe-ios-spm
Choose version 25.11.0
Inspect the available products for the package

**Observed available products:**
* Stripe
* StripeApplePay
* StripeCardScan
* StripeConnect
* StripeFinancialConnections
* StripeIdentity
* StripeIssuing
* StripePayments
* StripePaymentSheet
* StripePaymentsUI

**Expected:**
StripeCryptoOnramp should also be available, based on the native iOS Crypto Onramp docs (https://docs.stripe.com/crypto/onramp/embedded-components-integration-guide?platform=ios#authentication)

**iOS version**
iOS 18.x

**Installation method**
Swift Package Manager

**SDK version*
25.11.0 (also tested on 25.10, 25.9, 25.8)

**Other information**
We are integrating native iOS Crypto Onramp using the current docs, which instruct us to:

* Add https://github.com/stripe/stripe-ios-spm
* Select the latest version
* Add the StripeCryptoOnramp product to the app target
However, StripeCryptoOnramp does not appear in Xcode’s package product list at all.

**Notes**
We cleared package caches and re-added the package
We verified this on version 25.11.0
Our app currently uses StripeCryptoOnramp successfully through CocoaPods
The onramp source appears present in the SDK, and the module already seems SwiftPM-aware. For example, `StripeCryptoOnrampBundleLocator.swift` includes `#if SWIFT_PACKAGE ` and Bundle.module

**Question**:
Is StripeCryptoOnramp supposed to be publicly exposed from stripe-ios-spm?
If yes, is the current SPM manifest/release missing that product?
If no, what is the supported non-CocoaPods integration path for native iOS Crypto Onramp?****

mliberatore:
## Summary
Early in development, `StripeCryptoOnramp` got removed from `Package.swift` before it was ready for public use. (#5185). This adds it back with its current dependencies.

## Motivation
Fixes #6363

## Testing
1. Created a test project.
2. File → Add Package Dependencies…
3. Add Local…
4. Selected my clone of `stripe-ios` pointed at this PR's branch.
5. Selected `StripeCryptoOnramp` to include.
6. Added `@_spi(CryptoOnrampAlpha) import StripeCryptoOnramp` to `ContentView.swift` and a call to `CryptoOnrampCoordinator.create()`.
7. Ran the app to ensure there were no linking errors and the dependency properly logged to the console.

## Changelog

Added.

### DIFF
--- a/StripeCryptoOnramp/StripeCryptoOnramp/Source/Internal/API Bindings/STPAPIClient+CryptoOnramp.swift
+++ b/StripeCryptoOnramp/StripeCryptoOnramp/Source/Internal/API Bindings/STPAPIClient+CryptoOnramp.swift
@@ -5,6 +5,7 @@
 //  Created by Michael Liberatore on 7/17/25.
 //
 
+import Foundation
 @_spi(STP) import StripeCore
 @_spi(STP) import StripePaymentSheet
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ MINOR
 ### Connect
 * [Changed] [#66351](https://github.com/stripe/stripe-ios/pull/6351) Theming tokens in connect embedded components is now GA.
 
+### CryptoOnramp (Alpha)
+* [Fixed] Fixed an issue (#6363) in which StripeCryptoOnramp was not properly exposed via Swift Package Manager.
+
 ## 25.11.0 2026-04-13
 ### PaymentSheet
 * [Added] Added support for Pay by Bank payments (GA in GB, private preview in EU).
--- a/Package.swift
+++ b/Package.swift
@@ -47,6 +47,10 @@ let package = Package(
         .library(
             name: "StripeConnect",
             targets: ["StripeConnect"]
+        ),
+        .library(
+            name: "StripeCryptoOnramp",
+            targets: ["StripeCryptoOnramp"]
         )
     ],
     targets: [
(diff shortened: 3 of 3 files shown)


StripeCryptoOnramp/StripeCryptoOnramp/Source/Internal/API Bindings/STPAPIClient+CryptoOnramp.swift
```diff
@@ -5,6 +5,7 @@
 //  Created by Michael Liberatore on 7/17/25.
 //
 
+import Foundation
```
mliberatore, line 8: This addressed a warning noticed in testing due to usage of `LocalizedError` without an explicit `import` when depended on via SPM.

jeanregisser (approved):
Awesome 🚀

PR #6369, opened 2026-04-23
Remove AP selector from PaymentSheet complete

porter-stripe:
## Summary
- Removes the AP selector inside of PaymentSheet
- Removes the ability to update PaymentSheet while open (no longer needed)
- Removes any related code

## Motivation
- Removing in favor of the standalone element

## Testing
- CI

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/Analytics/PaymentSheetAnalyticsHelper.swift
```diff
@@ -472,20 +472,6 @@ final class PaymentSheetAnalyticsHelper {
...
-    }
-
-    func logAdaptivePricingCurrencyToggled() {
-        log(event: .adaptivePricingCurrencyToggled)
-    }
-
-    func logAdaptivePricingCurrencyToggledFailed(error: Error) {
-        log(event: .adaptivePricingCurrencyToggledFailed, error: error)
-    }
-
```
gbirch-stripe, line 488: Will the new standalone AP currency selector also live in the PS module? If so could we leave this for that to re-use?

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Checkout/CheckoutCurrencySelectorViewSnapshotTests.swift
```diff
@@ -90,10 +90,35 @@ final class CheckoutCurrencySelectorViewSnapshotTests: STPSnapshotTestCase {
...
+        window.overrideUserInterfaceStyle = darkMode ? .dark : .light
+        window.isHidden = false
+
+        view.translatesAutoresizingMaskIntoConstraints = true
+        view.frame = CGRect(origin: .zero, size: size)
+        window.addSubview(view)
+        window.layoutIfNeeded()
+        view.layoutIfNeeded()
+
+        STPSnapshotVerifyView(view, file: file, line: line)
```
gbirch-stripe: iirc very similar code exists elsewhere — is there anything we can reuse?

StripePaymentSheet/StripePaymentSheet/Source/Analytics/PaymentSheetAnalyticsHelper.swift
porter-stripe: The standalone element (it already exists!) doesn't have an instance of PaymentSheetAnalyticsHelper, but does live in the PaymentSheet module, so we can't use them. See Checkout+CurrencySelectorView.swift.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Checkout/CheckoutCurrencySelectorViewSnapshotTests.swift
porter-stripe: Was able to simplify this quite a bit to where reuse probably isn't worth it now. lmk what you think

gbirch-stripe (approved):

PR #6370, opened 2026-04-23
Fix PaymentSheet displaying raw API error messages to end users

cttsai-stripe:
## Summary
Fixes [RUN_MOBILESDK-5326](https://jira.corp.stripe.com/browse/RUN_MOBILESDK-5326)

PaymentSheet's `nonGenericDescription` was falling back to raw API error messages from `STPError.errorMessageKey` for **all** error types, including developer-facing ones like `invalid_request_error`. This caused end consumers to see messages referencing internal API fields (e.g. `requirements.disabled_reason`) and developer concepts ("Connect account").

Per [Stripe API docs](https://stripe.com/docs/api/errors#errors-message): *"For card errors, these messages can be shown to your users."* Non-card error messages are not meant for end-user display.

### Changes
- **`Error+PaymentSheet.swift`**: Added a `card_error` type guard (via `STPError.stripeErrorTypeKey`) in both branches of `nonGenericDescription` before returning raw `errorMessageKey` values. Non-card errors now fall back to the existing generic `stp_unexpectedErrorMessage()` which is already localized.
- **`Error+PaymentSheetTests.swift`**: Replaced the single generic test with 5 targeted tests covering card errors, non-card errors, missing error types, and the `STPPaymentHandlerErrorDomain` path.
- **`CHANGELOG.md`**: Added changelog entry.

## Motivation
A merchant reported that disabled Connect accounts caused the iOS PaymentSheet to display: *"This Connect account cannot currently make live charges. The `requirements.disabled_reason` property on the account will provide information about why this account is currently disabled..."* — text that is meaningless and confusing to end consumers.

## Testing
- Updated unit tests cover all branches: card_error (shows message), invalid_request_error (shows generic), no error type (shows generic), and both the PaymentHandler underlying error path and the direct error path.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ MINOR
 * [Removed] Removed UPI support across the SDK.
 
 ### PaymentSheet
+* [Fixed] Fixed an issue where PaymentSheet displayed raw API error messages (e.g. `invalid_request_error` details) to end consumers instead of a user-friendly generic message. Only `card_error` messages are now shown to users.
 * [Fixed] Fixed an issue where `LinkController` used the shared `STPAPIClient` instead of the `apiClient` specified in the caller, affecting apps using multiple API client instances.
 
 ### Connect
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,6 @@ MINOR
 * [Removed] Removed UPI support across the SDK.
 
 ### PaymentSheet
-* [Fixed] Fixed an issue where PaymentSheet displayed raw API error messages (e.g. `invalid_request_error` details) to end consumers instead of a user-friendly generic message. Only `card_error` messages are now shown to users.
 * [Fixed] Fixed an issue where `LinkController` used the shared `STPAPIClient` instead of the `apiClient` specified in the caller, affecting apps using multiple API client instances.
 
 ### Connect
--- a/Stripe/StripeiOSTests/Error+PaymentSheetTests.swift
+++ b/Stripe/StripeiOSTests/Error+PaymentSheetTests.swift
@@ -37,12 +37,69 @@ class Error_PaymentSheetTests: XCTestCase {
         XCTAssertEqual(NSError.stp_unexpectedErrorMessage(), TestableError.generic.nonGenericDescription)
     }
 
-    func testError_HasGenericLocalizedDescription_WithServerError() {
-        let serverErrorMessage = "Test failed server response error messasge"
-        let info = [NSLocalizedDescriptionKey: NSError.stp_unexpectedErrorMessage(), STPError.errorMessageKey: serverErrorMessage]
+    func testError_HasGenericLocalizedDescription_WithServerError_CardError() {
+        let serverErrorMessage = "Your card was declined."
+        let info: [String: Any] = [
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -7,14 +7,12 @@ MINOR
 * [Removed] Removed UPI support across the SDK.
 
 ### PaymentSheet
+* [Fixed] Fixed an issue where PaymentSheet displayed raw API error messages (e.g. `invalid_request_error` details) to end consumers instead of a user-friendly generic message. Only `card_error` messages are now shown to users.
```
gbirch-stripe: IMO this falls below the threshold to include in the changelog. Usually we only write more notable things that would change developer behavior

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Error+PaymentSheet.swift
gbirch-stripe, line 1: Can we use a shared constant and/or single "isCardError" check to not create duplicate logic to check the error type? Unfortunately all of this logic is already so nasty as-is. Do we know under what circumstances "card_error" is the type at the top-level vs being wrapped? I'm curious why we even need to check for it two different places

Stripe/StripeiOSTests/Error+PaymentSheetTests.swift
gbirch-stripe, line 1: Can we simulate these scenarios with actual real Swift errors that occur in the SDK rather than manually building artificial NSErrors?

CHANGELOG.md
davidme-stripe: ++

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Error+PaymentSheet.swift
davidme-stripe: Possibly 3DS2? I am curious as well

cttsai-stripe: Extracted a shared `userSafeErrorMessage` computed property on NSError that both paths now call. The two paths remain because they inspect different NSError objects:

- Path 1: STPPaymentHandler wraps API errors as `NSUnderlyingError` during confirmation/3DS2 — the outer error has a useless default `localizedDescription`, so we dig into the wrapped error.
- Path 2: Direct STPAPIClient errors carry `stripeErrorTypeKey` in their own userInfo.

The check logic is now single-sourced; the paths just differ in *which* NSError to inspect.

Stripe/StripeiOSTests/Error+PaymentSheetTests.swift
cttsai-stripe: Updated to use `NSError.stp_error(errorType:stripeErrorCode:stripeErrorMessage:...)` for both direct and underlying errors. The one exception is the "no error type" test which uses manual NSError construction — this represents SDK-internal errors (e.g. connection failures) that genuinely bypass `stp_error()` and have no `stripeErrorTypeKey` in userInfo.

davidme-stripe (approved):

PR #6371, opened 2026-04-23
feat(FinancialConnections): Pass client_secret on calls to pollAccountNumbersForSelectedAccounts

dknudsen-stripe:
## Summary
Passes the `client_secret` parameter on calls to `/v1/link_account_sessions/poll_account_numbers`. This parameter is currently optional but it really shouldn't be.  All other APIs in the FC client here use the `client_secret` param. 

## Motivation
Having `client_secret` available on the backend helps us a lot with flagging, metrics, and logging. It's also our means of access control and it's weird we don't require it just on this API. I'll eventually want to make it required and kick old SDK versions over to webview.

## Testing
Added a new unit test.

## Changelog
Entirely internal and doesn't change any user facing behavior, so I don't think this needs a changelog entry.


mats-stripe (approved):
> This parameter is currently optional but it really shouldn't be.

Agreed, but it will need to continue to be optional while we have mobile clients in the wild not passing it up

dknudsen-stripe:
@mats-stripe see https://git.corp.stripe.com/stripe-internal/mint/pull/2113144/files

PR #6373, opened 2026-04-24
[link] Set up `LinkBrand` config

sophiah2-stripe:
## Summary

Initial set up for brand config for `PaymentSheet` with grey PaywithLinkButton Link logo and for FC with grey logo in the top bar as POC. More detailed UI updates will follow.

## Motivation

## Testing

Added testing for new `LinkBrand`.
This can be tested on the Example playground by toggling `Link brand`.

## Changelog

PaymentSheet:

| `Link` | `Notlink` |
| ------ | ------ |
| Simulator Screenshot - iPhone 16 Pro - 2026-04-24 at 15 57 34 | Simulator Screenshot - iPhone 16 Pro - 2026-04-24 at 15 57 29 |

FC:
Standalone
| `Link` | `Notlink` |
| ------ | ------ |
| Simulator Screenshot - iPhone 16 Pro - 2026-04-24 at 16 08 03 | Simulator Screenshot - iPhone 16 Pro - 2026-04-24 at 16 07 50 |

Payment Element
| `Link` | `Notlink` |
| ------ | ------ |
| Simulator Screenshot - iPhone 16 Pro - 2026-04-28 at 12 12 09 | Simulator Screenshot - iPhone 16 Pro - 2026-04-28 at 12 12 00 |


Example/FinancialConnections Example/FinancialConnections Example/Playground/PlaygroundView.swift
```diff
@@ -127,6 +127,22 @@ struct PlaygroundView: View {
                             }
                             .pickerStyle(.segmented)
                         }
+
+                        if viewModel.integrationType.wrappedValue == .standalone
```
jkelle-stripe: Can you explain why we need these conditions here?

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetDeferredUITests.swift
```diff
@@ -32,7 +32,8 @@ class PaymentSheetDeferredUITests: PaymentSheetUITestCase {
         // two wallet pms and 2 in the carousel
         XCTAssertEqual(
             (initialDisplayedPaymentMethodsEvent.map { $0["visible_payment_methods"] } as? [String])?.count,
-            4
+            5,
```
jkelle-stripe: We may not need these changes now that #6374 is merged

StripeFinancialConnections/StripeFinancialConnections.xcodeproj/project.pbxproj
```diff
@@ -277,6 +277,8 @@
 				"zh-Hans",
 				"zh-Hant",
 				Base,
+				"nn-NO",
```
jkelle-stripe: Do we need these here?

StripeFinancialConnections/StripeFinancialConnections/Source/Common/FinancialConnectionsNavigationController.swift
```diff
@@ -174,8 +174,11 @@ extension FinancialConnectionsNavigationController {
         let logoView: UIImageView? = {
             guard !shouldHideLogo else { return nil }
 
-            let logoImage = UIImageView(image: appearance.logo.makeImage(template: true))
-            logoImage.tintColor = appearance.colors.logo
+            let usesTemplate = appearance.logo != .notlink_logo
```
jkelle-stripe: Just checking if we'll still need this when we use the future logo?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentElementConfiguration.swift
```diff
@@ -49,6 +50,9 @@ protocol PaymentElementConfiguration: PaymentMethodRequirementProvider {
 }
 
 extension PaymentElementConfiguration {
+    func resolvedLinkBrand(elementsSession: STPElementsSession? = nil) -> LinkBrand {
+        link.brand ?? elementsSession?.linkBrand ?? .link
```
jkelle-stripe: IMO we should have a comment here describing the reasoning behind this priority / order. If there's no reasoning right now, that's ok, I'd just say that it's arbitrary. In case someone wants to change it in the future, they would know the implications

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
```diff
@@ -556,6 +560,14 @@ extension PaymentSheet {
             self.display = display
         }
 
+        @_spi(STP) public init(
```
jkelle-stripe, line 563: Is it possible for this init to be combined with one above on 557? I think we may not need both.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentElementConfiguration.swift
```diff
@@ -49,6 +50,9 @@ protocol PaymentElementConfiguration: PaymentMethodRequirementProvider {
 }
 
 extension PaymentElementConfiguration {
+    func resolvedLinkBrand(elementsSession: STPElementsSession? = nil) -> LinkBrand {
```
jkelle-stripe: I think `elementsSession` can be non-optional here looking at the call sites

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
```diff
@@ -528,6 +529,9 @@ extension PaymentSheet {
         /// The Link display mode.
         public var display: Display = .automatic
 
+        /// The brand to use for Link. Expected values are `.link` or `.notlink`.
+        @_spi(STP) public var brand: LinkBrand?
```
jkelle-stripe, line 533: Is it possible for this to be non-optional and use `.link` as default in the init below? Then we may not need [this comment](https://github.com/stripe/stripe-ios/pull/6373/changes#r3150161891)

jkelle-stripe (changes requested):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
sophiah2-stripe: The logic here is to have `brand` be the client-side override, which is only used in the playground app. So if there is no client override, `brand == nil`, so we would fallback to the backend response given in `elementsSession.linkBrand`, which would not work if we make `brand` non-optional. I'll add a comment to explain the fallback where you [commented](https://github.com/stripe/stripe-ios/pull/6373#discussion_r3150161891) about that.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
sophiah2-stripe: I'm keeping the `brand` out of the public API, but I can dedupe this by calling `.init` with `brand = nil` in line 560

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheetDeferredUITests.swift
sophiah2-stripe: Removing!

StripeFinancialConnections/StripeFinancialConnections.xcodeproj/project.pbxproj
sophiah2-stripe: also removing this

Example/FinancialConnections Example/FinancialConnections Example/Playground/PlaygroundView.swift
sophiah2-stripe: oh good catch, this was to limit the testing toggle to just one case, but since the same logic/asset is shared across multiple surfaces, the changes are already applied to more than the standalone IBP case. I'm updating this to only hide the toggle in the standalone FC case and adding the before/after screenshots for one payment element case.

StripeFinancialConnections/StripeFinancialConnections/Source/Common/FinancialConnectionsNavigationController.swift
sophiah2-stripe: I'm resolving some issues with light/dark mode for the assets, but will align the approach for tinting the logo asset for now and defer updating to the official asset after it's ready

sophiah2-stripe:
Comments addressed in [b6ceae3](https://github.com/stripe/stripe-ios/pull/6373/commits/b6ceae3540f7cea07d4d11edfe9bbb4bb23f62f8) and [6a35c17](https://github.com/stripe/stripe-ios/pull/6373/commits/6a35c17e76ef62ddbcff268b4aa6c35e3bbc13b4)

jkelle-stripe (approved):

PR #6379, opened 2026-04-27
[Identity] Adding in the ability to manually capture identity documents instead of just auto capture  

Pearapps:
•  adds a manual capture mode for identity documents directly in the existing capture screen with a Live / Manual switch
•  keeps the same camera/viewfinder UI in manual mode, disables live scanning feedback + auto-capture, and lets the user explicitly tap Take Photo
•  hides the switch on the timeout/error screen and updates manual capture copy from Scanned to Image taken
•  only allows manual capture when live verification is not required

You can see the switch and new UI here:

IMG_0072


StripeIdentity/StripeIdentity/Source/Helpers/String+Localized.swift
```diff
@@ -230,6 +230,20 @@ extension String.Localized {
         )
     }
 
+    static var liveCaptureMode: String {
+        STPLocalizedString(
+            "Live",
```
mnn-stripe: I'd like to go with `Auto` word to keep this consistent with web experience. 

Image

StripeIdentity/StripeIdentity/Source/NativeComponents/ViewControllers/DocumentCaptureViewController+Strings.swift
```diff
@@ -106,6 +106,13 @@ extension DocumentCaptureViewController {
         )
     }
 
+    static var imageTakenInstructionalText: String {
+        STPLocalizedString(
+            "Image taken",
```
mnn-stripe, line 111: This seems to be absent in en.lproj/Localizable.strings. Do we want to add it there?

StripeIdentity/StripeIdentity/Source/NativeComponents/ViewControllers/DocumentCaptureViewController.swift
```diff
@@ -576,6 +764,20 @@ extension DocumentCaptureViewController: ImageScanningSessionDelegate {
...
     ) {
+        let hadCapturedCameraFrame = latestCapturedCameraFrame != nil
+        latestCapturedCameraFrame = .init(
+            image: image,
+            scannerOutput: scannerOutputOptional,
+            exifMetadata: exifMetadata
+        )
+        guard !isManualCaptureMode else {
+            imageScanningSession.stopTimeoutTimer()
+            imageScanningSession.updateScanningState(nil)
```
mnn-stripe, line 775: When the user taps "Take Photo", the app immediately marks the image as captured and tells the camera to stop. But the camera doesn't stop instantly; it may have already finished processing one more video frame just before the stop message arrived. When that last frame's result gets delivered, the code tries to update the scanner state, but the scanner has already moved past the "scanning" phase. There's a built-in safety check that crashes the app (in debug builds) when this happens. ImageScanningSession/ImageScanningSession.swift:141

Before updating the scanner state for an incoming frame, we need to check that we're still actually in the scanning phase. If we're not, skip the update.

mnn-stripe (changes requested):

mnn-stripe:
Thanks for making the changes! PTAL at the comments and resolve bot's dead-code comment

StripeIdentity/StripeIdentity/Source/NativeComponents/ViewControllers/DocumentCaptureViewController.swift
Pearapps: I think this is already covered by `ImageScanningSession` before we get here.

`setStateScanned` moves us to `.scanned` before `stopScanning()`, and both new frames + in-flight scan completions have `.scanning` guards before calling `imageScanningSessionDidScanImage`. So if a late frame/result comes in after “Take Photo”, it should bail before reaching this manual branch / `updateScanningState(nil)`.

Happy to add the extra guard defensively if you still prefer, but I don’t think that path is actually reachable through the normal session flow.

mnn-stripe: yeah that makes sense! I missed this check ImageScanningSession.swift:337

mnn-stripe (dismissed):
LGTM!

cjmisenas-stripe (approved):

PR #6386, opened 2026-05-01
Upgrade CI to Xcode 26.4.1 and remove Xcode 16 support

davidme-stripe:
## Summary
- Update all Bitrise workflow stacks from osx-xcode-16.0.x to osx-xcode-26.4.x
- Keep around one Xcode 26.2 job to test older Xcode 26 versions
- Remove test-builds-xcode-16 and test-builds-xcode-16-release workflows
- Update GitHub Actions to macos-26 runner and Xcode_26.4.1.app
- Add skip_detect_devices to work around an issue with simctl
- Run `bundle update` to get the latest gems
- Update .periphery.yml to skip Xcode warnings, as they're not relevant to the API detector

## Test plan
- [x] Verify Bitrise builds trigger on the new stacks
- [x] Verify GitHub Actions workflows pass with macos-26 runner


fastlane/Fastfile
davidme-stripe, line 174: Device detection seems to be broken on Xcode 26.4, but we also don't need it here (we're specifying the device we want).

PR #6389, opened 2026-05-01
[link] Preliminary update for branded strings to use placeholder

sophiah2-stripe:
## Summary

Replace existing branded strings to use a placeholder that accommodates `LinkBrand` value. Remaining strings that require more refactoring/threading will be replaced in a future PR

## Motivation

https://jira.corp.stripe.com/browse/LINK_MOBILE-853

## Testing

Modified and added tests to check if strings are branded correctly.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Resources/Localizations/en.lproj/Localizable.strings
+++ b/StripePaymentSheet/StripePaymentSheet/Resources/Localizations/en.lproj/Localizable.strings
@@ -305,6 +305,9 @@ Pay faster at {Merchant Name} and thousands of businesses
 e.g, 'Pay faster at Example, Inc. and thousands of businesses.' */
 "Pay faster at %@ and thousands of businesses." = "Pay faster at %@ and thousands of businesses.";
 
+/* Subtitle for the Link signup screen. The placeholder is a Stripe brand name and should not be translated. */
+"Pay faster everywhere %@ is accepted." = "Pay faster everywhere %@ is accepted.";
+
 /* Subtitle for the Link signup screen */
 "Pay faster everywhere Link is accepted." = "Pay faster everywhere Link is accepted.";
 
@@ -469,3 +472,4 @@ the heading the screen itself. */
 
 /* An error message. */
 "Your IBAN should start with a two-letter country code." = "Your IBAN should start with a two-letter country code.";
+
--- a/StripePaymentSheet/StripePaymentSheet/Resources/Localizations/en.lproj/Localizable.strings
+++ b/StripePaymentSheet/StripePaymentSheet/Resources/Localizations/en.lproj/Localizable.strings
@@ -83,8 +83,8 @@
 /* Text providing link to terms for ACH payments */
 "By continuing, you agree to authorize payments pursuant to <terms>these terms</terms>." = "By continuing, you agree to authorize payments pursuant to <terms>these terms</terms>.";
 
-/* Text displayed below a credit card entry form when the card will be saved with the merchant and saved to Link. */
-"By continuing, you agree to save your information for future purchases with %@ and <link>Link</link> according to Link <terms>terms</terms> and <privacy>privacy</privacy>." = "By continuing, you agree to save your information for future purchases with %@ and <link>Link</link> according to Link <terms>terms</terms> and <privacy>privacy</privacy>.";
+/* Text displayed below a credit card entry form when the card will be saved with the merchant and saved to the Link brand. The first placeholder is a merchant name. The second placeholder is a Stripe brand name and should not be translated. */
+"By continuing, you agree to save your information for future purchases with %1$@ and <link>%2$@</link> according to %2$@ <terms>terms</terms> and <privacy>privacy</privacy>." = "By continuing, you agree to save your information for future purchases with %1$@ and <link>%2$@</link> according to %2$@ <terms>terms</terms> and <privacy>privacy</privacy>.";
 
 /* Text displayed below a credit card entry form when the card will be saved with the merchant. */
 "By continuing, you agree to save your information for future purchases with %@." = "By continuing, you agree to save your information for future purchases with %@.";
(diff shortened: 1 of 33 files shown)


StripePaymentSheet/StripePaymentSheet/Source/Categories/String+Localized.swift
```diff
@@ -72,8 +79,19 @@ extension String.Localized {
...
-    static var pay_with_link: String {
-        STPLocalizedString("Pay with Link", "Text for the 'Pay with Link' button. 'Link' is a Stripe brand, please do not translate the word 'Link'.")
+    static func pay_with_link(brand: LinkBrand) -> String {
+        switch brand {
+        case .link, .unparsable:
+            return existingLinkLocalizedString("Pay with Link")
+        case .onelink:
+            return String(
+                format: STPLocalizedString(
+                    "Pay with %@",
```
jkelle-stripe, line 89: Are these distinct so that we could make sure the parameterized string creates a new localization? Is the thought we'd use the parameterized version going forward?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/MandateVariant.swift
```diff
@@ -15,19 +15,27 @@ enum MandateVariant {
...
+                    merchantDisplayName: merchant,
+                    brand: brand
+                )
             } else {
                 String.Localized.by_continuing_you_agree_to_save_your_information_to_merchant
             }
         }
 
-        let terms = String(format: formatText, merchant).removeTrailingDots()
+        let terms = {
```
jkelle-stripe, line 33: Do you have a screenshot for this? Just not sure where we show this

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
```diff
@@ -142,6 +143,14 @@ class PaymentSheetFormFactory {
                   analyticsHelper: analyticsHelper,
                   paymentMethodIncentive: elementsSession.incentive,
                   linkAppearance: linkAppearance,
+                  linkBrand: {
```
jkelle-stripe: [nit] since we have more of these above the init (line 177 for example) can you move this up there to keep the standard going?

StripePaymentSheet/StripePaymentSheet/Source/Categories/String+Localized.swift
sophiah2-stripe: Yes, I'm preserving the current strings for non-rebranded strings to avoid any regressions while the new placeholder strings are waiting to be translated. Once those are in, the goal is to migrate all paths to use the new strings.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/MandateVariant.swift
sophiah2-stripe: I generated a one-off snapshot for this locally to test, but since it requires a specific backend response, I wasn't able to force this variant to show up in the example app:
Image

StripePaymentSheet/StripePaymentSheet/Resources/Localizations/en.lproj/Localizable.strings
```diff
@@ -83,8 +83,8 @@
...
 
-/* Text displayed below a credit card entry form when the card will be saved with the merchant and saved to Link. */
-"By continuing, you agree to save your information for future purchases with %@ and <link>Link</link> according to Link <terms>terms</terms> and <privacy>privacy</privacy>." = "By continuing, you agree to save your information for future purchases with %@ and <link>Link</link> according to Link <terms>terms</terms> and <privacy>privacy</privacy>.";
+/* Text displayed below a credit card entry form when the card will be saved with the merchant and saved to the Link brand. The first placeholder is a merchant name. The second placeholder is a Stripe brand name and should not be translated. */
+"By continuing, you agree to save your information for future purchases with %1$@ and <link>%2$@</link> according to %2$@ <terms>terms</terms> and <privacy>privacy</privacy>." = "By continuing, you agree to save your information for future purchases with %1$@ and <link>%2$@</link> according to %2$@ <terms>terms</terms> and <privacy>privacy</privacy>.";
```
jkelle-stripe, line 87: I didn't know about `%1$@`! My understanding is that this allows us to repeat the same argument in multiple positions in the output string. Cool!

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Elements/InlineSignup/LinkInlineSignupView-CheckboxElement.swift
```diff
@@ -99,8 +82,7 @@ extension LinkInlineSignupView {
                             textStyle: .caption
                         )
 
-                        // Replace "Link" with the icon
-                        let attachmentString = NSAttributedString(attachment: leadingIcon)
```
jkelle-stripe, line 103: I guess this was just unused previously

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/MandateVariant.swift
```diff
@@ -15,19 +15,27 @@ enum MandateVariant {
...
+            if case .updated(true) = self {
```
jkelle-stripe, line 34: [nit] This would be clearer if we did a switch case for both original and updated. That way if we add another case it would be a compiler error to not handle it here.

jkelle-stripe (approved):
One non-blocking nit so all good to merge and address in a future PR

StripePaymentSheet/StripePaymentSheet/Resources/Localizations/en.lproj/Localizable.strings
sophiah2-stripe: Specifically `%1$@` and `%2$@` are placeholders for the first and second arguments passed in, so in this case `%1$@ = [merchant]` and `%2$@ = [brand]`

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/MandateVariant.swift
sophiah2-stripe: Will address this in the next PR!

PR #6396, opened 2026-05-04
CryptoOnramp SDK: EU Compliance Backend APIs, Models, and Tests

mliberatore:
## Summary

For EU support in Crypto Onramp, this covers the new backend API communication and the associated responses, along with tests to ensure serialization/deserialization is working as expected.

This does NOT yet modify the `CryptoOnrampCoordinator` API to make use of these requests, nor does it update the example app to exercise that functionality. Those changes will be coming in separate PRs.

## Motivation
- [Mobile SDK EU KYC Additions](https://docs.google.com/document/d/17YUoOXK4bkGmU5ENihYyBzOMm9qav42cykc74Hl_wnE/edit?tab=t.0#heading=h.zbt1epa8d668)
- [Backend API Review](https://docs.google.com/document/d/1TGHh_Yyb8k--TO8iE2L9mFO_L3n65FBwShtrBrSDdTo/edit?tab=t.0#heading=h.qarkdnrrshe0)

Note that items in these docs may be out-of-date. The included JSON is taken from real network responses from the endpoints coded against in this PR.

## Testing
Ran the newly added tests. The response JSON was generated using real requests against the backend on branch `mliberatore/crypto-onramp-eu-support`.

## Changelog
✅


StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/Compliance/ComplianceIdentifierType.swift
```diff
@@ -0,0 +1,131 @@
...
+//  StripeCryptoOnramp
+//
+//  Created by Michael Liberatore on 5/4/26.
+//
+
+import Foundation
+
+/// The type of compliance identifier required or submitted for regulatory compliance.
+@_spi(CryptoOnrampAlpha)
+public struct ComplianceIdentifierType: Codable, Equatable, ExpressibleByStringLiteral, Hashable, RawRepresentable {
```
mliberatore, line 12: I’m on the fence about this API and whether we should just treat this as a `String` instead of its own type. On `mliberatore/crypto-onramp-eu-support` I have a mapping of the known identifiers from the extension below (the `static let`s) to displayable names located in the example app (I don't yet know whether the SDK would be providing localized names for each of these IDs).

I didn't opt for an `enum` due to the likelihood of the possible values expanding over time, and wanting to not fail parsing if the server returns a new identifier without an SDK update, so I chose a raw-representable. This could just be replaced with a `String`, rather than a named type. So I feel like the options are:

- `enum` with forward compatibility issues
- `RawRepresentable` wrapper type with named constants
- `String` likely without named constants

Happy to take opinions here.

jeanregisser: Your approach looks good!

jeanregisser (approved):
Awesome! 🚀

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/Compliance/ComplianceRegulation.swift
```diff
@@ -0,0 +1,19 @@
...
+/// The regulation requiring a compliance identifier.
+@_spi(CryptoOnrampAlpha)
+public enum ComplianceRegulation: String, Codable, Equatable, Hashable {
+
+    /// Common Reporting Standard (CRS) and Crypto-Asset Reporting Framework (CARF).
+    case euCARF = "eu_carf"
+
+    /// Markets in Crypto-Assets Regulation (MiCA).
+    case euMiCA = "eu_mica"
+}
```
jeanregisser, line 19: nit: `ComplianceIdentifierType` is a struct so unknown server values decode fine. This is an enum though if the backend adds a new regulation, the whole response will fail to decode. Worth using the same struct pattern here, or are we confident the regulation list won't grow?
I'll ask Rohan.

PR #6401, opened 2026-05-05
CryptoOnramp SDK: Changes CryptoOnramp-specific APIs in StripePaymentSheet to @_spi(CryptoOnrampAlpha)

mliberatore:
## Summary
There are a few APIs in `StripePaymentSheet` that were exposed only for `StripeCryptoOnramp`. We want to remove `@_spi(STP)` usage from React Native, and that requires migrating these to `@_spi(CryptoOnrampAlpha)`.

These changes are in:
- `LinkAppearance.swift`
- `VerifyKYCResult.swift`
- `STPImageLibrary.swift`

All other changes here are correcting the relevant imports to ensure a clean build.

## Motivation
This request came from Jean in Slack.

## Testing
- Allowed CI to build and run all tests.
- Tested locally with https://github.com/stripe/stripe-react-native/pull/2441 in the React Native project. 

## Changelog
✅


Example/CryptoOnramp Example/CryptoOnramp Example/UI Helpers/AuthenticatedUserToolbarItemModifier.swift
```diff
@@ -8,8 +8,9 @@
 import SwiftUI
 import UIKit
 
+@_spi(STP) import StripeCore
```
mliberatore: (Example app) We use `StripeError` here, requiring the new import since it was indirectly imported via `@_spi(STP) import StripePaymentSheet`. This does not affect the SDK.

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
```diff
@@ -12,7 +12,7 @@ import PassKit
 @_spi(STP) import StripeCore
 @_spi(STP) import StripeIdentity
 @_spi(STP) import StripePayments
-@_spi(STP) import StripePaymentSheet
+@_spi(STP) @_spi(CryptoOnrampAlpha) import StripePaymentSheet
```
mliberatore, line 15: `StripeCryptoOnramp` uses `StripePaymentSheet`'s `LinkController`, so `@_spi(STP)` stays here. This does not have an implication on usage in RN, as `LinkController` is not used directly by the RN SDK, nor is it expected to be.

StripePaymentSheet/StripePaymentSheet/Source/Helpers/STPImageLibrary.swift
```diff
@@ -65,7 +65,7 @@ public class PaymentSheetImageLibrary {
         return "default"
     }
 
-    @_spi(STP) @_spi(AppearanceAPIAdditionsPreview)
+    @_spi(CryptoOnrampAlpha) @_spi(AppearanceAPIAdditionsPreview)
     public class func bankIcon(for bank: String?, iconStyle: PaymentSheet.Appearance.IconStyle) -> UIImage {
```
yuki-stripe: Do you really mean for these to be part of the public API? From a quick look these are only internally used by stripe-react-native to power a crypto API over there,  not sure we intend for users of this SDK to access these methods.

If this is an internal API that has to be used by stripe-react-native, use `@_spi(ReactNativeSDK)` in the short term. Long term you should make sure this usage really is needed.

mliberatore: @yuki-stripe I will update it to `@_spi(ReactNativeSDK)` shortly, per your suggestion. Please see [this thread](https://lickability.slack.com/archives/C094P3BRBL1/p1758043262318759?thread_ts=1758043210.280779&cid=C094P3BRBL1) in Lickability Slack for the instructions we followed to expose this as an API for RN. (Paste the link to Slackbot in Stripe slack, and then it should resolve properly.)

Long term, happy to follow your desired solution here, just note that it will likely affect Android too. After reading the linked thread, please let me know how you'd like us to proceed.

yuki-stripe: Hmm I don't have the time right now to fully grok the thread. From a really quick read it seems like this method/functionality is not a react native specific API and would apply to the native SDKs, in which case we should add native APIs and then simply proxy to them in react native. 

I think this is something that will get resolved when this API goes through API Review, if we don't solve it now.

Example/CryptoOnramp Example/CryptoOnramp Example/UI Helpers/AuthenticatedUserToolbarItemModifier.swift
yuki-stripe: Why do we use StripeError in this example? Can we avoid doing this? This appears to be a user-facing example, in which case we shouldn't be using internal APIs.

StripePaymentSheet/StripePaymentSheet/Source/Helpers/STPImageLibrary.swift
mliberatore: @yuki-stripe would you mind approving this in its current state if migrating the API is to be tackled during API review?

Example/CryptoOnramp Example/CryptoOnramp Example/UI Helpers/AuthenticatedUserToolbarItemModifier.swift
mliberatore: Removed. This was used when building the feature, as we were working alongside the backend implementation, and it was helpful when debugging with error messages before the feature was complete.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ MINOR
 
 ### PaymentSheet
 * [Added] Added support for new and upcoming payment methods in native Link, such as UPI, Pix, and Crypto
+* [Changed] Changed `@_spi(STP)` exposure of `LinkAppearance`, `VerifyKYCResult`, `PaymentSheetImageLibrary.bankIconCode(for:)`, and `PaymentSheetImageLibrary.bankIcon(for:iconStyle:)` to `@_spi(CryptoOnrampAlpha)`.
```
yuki-stripe: I would mark this as [Fixed] and also don't forget to remove `PaymentSheetImageLibrary.bankIconCode(for:)`, and `PaymentSheetImageLibrary.bankIcon(for:iconStyle:)`.

yuki-stripe (dismissed):

CHANGELOG.md
mliberatore: Updated per your request. Re-requesting review.

yuki-stripe (approved):

PR #6402, opened 2026-05-05
Fix FlowController.update() losing paymentOption when mandate is introduced

jaynewstrom-stripe:
## Summary
- When `FlowController.update()` is called with a changed `setupFutureUsage` (e.g. `nil` → `.offSession`), a mandate requirement is introduced for cards. Previously this caused `paymentOption` to become `nil` because the mandate element hadn't been displayed yet.
- Fix: Before passing the previous payment option to the new view controller during `update()`, mark `didDisplayMandate = true` on its confirm params. This preserves the payment selection — the mandate will be shown when the user next presents the payment options sheet.
- Fixes the crash/broken flow reported by HousecallPro where calling `confirm()` after `update()` would hit an assertion failure.

## Motivation
https://jira.corp.stripe.com/browse/RUN_MOBILESDK-5344

## Testing
- Added unit tests for `mandateMarkedAsDisplayed()` on all PaymentOption cases
- Added integration test verifying `makeViewController` preserves `selectedPaymentOption` when mandate is introduced with `didDisplayMandate = true`
- Added form-level test verifying mandate element validity with/without `didDisplayMandate`
- All existing mandate tests (`testAppliesPreviousCustomerInput_for_mandate`, 102 form factory tests) continue to pass

## Test plan
- [ ] Verify new tests pass in CI
- [ ] Manual test: create FlowController with deferred intent, fill in card, call `update()` with `setupFutureUsage: .offSession`, verify `paymentOption` is non-nil
- [ ] Verify the mandate is still displayed when the user opens the payment sheet after `update()`

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
@@ -712,12 +712,13 @@ extension PaymentSheet {
                 case .success(let (loadResult, confirmationChallenge)):
                     // 2. Re-initialize PaymentSheetFlowControllerViewController to update the UI to match the newly loaded data e.g. payment method types may have changed.
 
+                    let previousPaymentOption = self.internalPaymentOption?.mandateMarkedAsDisplayed()
                     self.viewController = Self.makeViewController(
                         configuration: self.configuration,
                         loadResult: loadResult,
                         analyticsHelper: analyticsHelper,
                         walletButtonsViewState: walletButtonsViewState,
-                        previousPaymentOption: self.internalPaymentOption,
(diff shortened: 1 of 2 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
```diff
@@ -712,12 +712,13 @@ extension PaymentSheet {
                 case .success(let (loadResult, confirmationChallenge)):
                     // 2. Re-initialize PaymentSheetFlowControllerViewController to update the UI to match the newly loaded data e.g. payment method types may have changed.
 
+                    let previousPaymentOption = self.internalPaymentOption?.mandateMarkedAsDisplayed()
```
jaynewstrom-stripe, line 715: This doesn't seem right 🤔

PR #6403, opened 2026-05-05
Add `LinkController.present` API & demo

mats-stripe:
## Summary

Adds a new internally-public `present` API for the `LinkController` that handles:
1. Consumer lookup
2. Authentication
3. Payment method selection
4. Payment method creation

Full implementation plan is here: https://git.corp.stripe.com/gist/mats/dfd8299b4382fca63e18a98baf835099

This currently has no usage, besides the LinkController demo app.

## Motivation

https://docs.google.com/document/d/1b5gHtHsxE2Rl_YXvcJswcfPz2P0AtRSciAvpaqa_WKk/edit?tab=t.0#bookmark=id.v6evombxv9i3

## Testing

https://github.com/user-attachments/assets/a88cd7f2-c9e7-463e-b969-d341a5887a9d

## Changelog

N/a

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkController.swift
@@ -49,6 +49,13 @@ import UIKit
         case canceled
     }
 
+    @frozen @_spi(STP) public enum PaymentMethodResult {
+        /// The user selected a payment method. The associated value is the resulting `STPPaymentMethod`.
+        case completed(STPPaymentMethod)
+        /// The user dismissed the flow without selecting a payment method.
+        case canceled
+    }
+
     @frozen @_spi(STP) public enum AuthorizationResult {
         /// Authorization was consented by the user.
         case consented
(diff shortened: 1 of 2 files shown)


mliberatore (approved):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkController.swift
```diff
@@ -404,6 +411,81 @@ import UIKit
...
+    /// - Parameter email: The email address to look up and associate with the Link account.
+    /// - Parameter phoneNumber: Optional phone number in E.164 format to prefill during signup.
+    /// - Parameter supportedPaymentMethodTypes: The payment method types to support. If `nil`, all available types are shown.
+    /// - Parameter presentingViewController: The view controller from which to present the Link sheet.
+    /// - Parameter completion: A closure called with `.success(.completed(paymentMethod))` on selection,
+    ///   `.success(.canceled)` if the user dismisses the flow, or `.failure(error)` on API or network errors.
+    @_spi(STP) public func present(
+        email: String,
+        phoneNumber: String? = nil,
+        supportedPaymentMethodTypes: [LinkPaymentMethodType]? = nil,
```
tillh-stripe, line 432: One unrelated idea and thus not blocking: If we're only passing card or bank account here, should we update the label for adding a new payment method to be more specific, e.g. `Add bank account` for the bank-only flow?

tillh-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkController.swift
mats-stripe: We could, yeah. Maybe as part of the redesign!

PR #6404, opened 2026-05-05
Horizontal experiment ship

joyceqin-stripe:
## Summary

Update `automatic` layout logic to show `.vertical` when >= 3 pms available or elements/session flag is forcing vertical, `.horizontal` otherwise.

Log `payment_method_orientation` on ~all analytics after `load`, drop `payment_method_layout` from `mc_initial_displayed_payment_methods` in favor of `payment_method_orientation`.
## Motivation

Horizontal A/B test ship recommendation
## Testing

Added tests
## Changelog

[Changed] When paymentMethodLayout is set to automatic, the layout is now horizontal when there are 2 or fewer payment methods available.


StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/STPElementsSession.swift
```diff
@@ -395,6 +395,10 @@ extension STPElementsSession {
     var isCardFundingFilteringEnabled: Bool {
         flags["elements_mobile_card_funding_filtering"] == true
     }
+
+    var forceVerticalPaymentMethodLayout: Bool {
+        flags["elements_mobile_force_vertical_payment_method_layout"] == true
```
porter-stripe, line 400: nit:
can this just be `return flags["elements_mobile_force_vertical_payment_method_layout"]`

porter-stripe (approved):

PR #6411, opened 2026-05-06
PMM Experiment — Form header

gbirch-stripe:
## Summary
- refactored PMME promotion UI code to be reusable
- added new header view
- doesn't yet connect to data sources

## Motivation

## Testing
- snapshot tests
- unit tests


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMMEInfoModal.swift
```diff
@@ -127,3 +127,15 @@ extension PMMEInfoModal: WKNavigationDelegate {
...
+
+extension PMMEInfoModal {
+    static func present(
+        infoUrl: URL,
+        style: PaymentMethodMessagingElement.Appearance.UserInterfaceStyle,
+        from sourceView: UIView?
+    ) {
+        let infoController = PMMEInfoModal(infoUrl: infoUrl, style: style)
+        infoController.modalPresentationStyle = .formSheet
+        sourceView?.window?.findTopMostPresentedViewController()?.present(infoController, animated: true)
```
yuki-stripe: So if `sourceView` is nil, this does nothing? Is that intended?

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/BNPLFormHeaderViewTests.swift
```diff
@@ -0,0 +1,129 @@
...
+@_spi(STP) @testable import StripePaymentSheet
+
+@MainActor
+final class BNPLFormHeaderViewTests: STPSnapshotTestCase {
+    func testDefaultAppearance() {
+        let (headerView, _, window) = makeHeaderView()
+
+        verify(headerView)
+
+        window.isHidden = true
```
yuki-stripe: ooc why do we need this line in the tests?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
```diff
@@ -842,6 +842,50 @@ extension PaymentSheetFormFactory {
         return country
     }
 
+    private var bnplHeaderStyle: PaymentSheet.UserInterfaceStyle {
+        guard case .paymentElement(let configuration, _) = configuration else {
+            stpAssertionFailure("BNPL headers are only supported for PaymentSheet and EmbeddedPaymentElement.")
```
yuki-stripe: This sounds like we're saying FlowController is not supported, is that right? I think the enum case includes FC. Should this say we don't support CustomerSheet?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMMEInfoModal.swift
gbirch-stripe: I think there was a use for this at one point in prototyping but can be made non-optional. Thanks!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
gbirch-stripe: I meant FlowController implicitly as a sub-element of PS, but you're right that this is confusing and I updated it to clarify

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/BNPLFormHeaderViewTests.swift
gbirch-stripe: This is a new cleanup pattern introduced in ConfirmationChallengeTests.swift and PassiveCaptchaChallengeTests.swift. TL;DR it may be necessary since we are doing tap behavior to test the info modal. I can look into omitting, but everything works properly with it :). LLM explanation:
```
  The key difference from many of the other tests is that this one creates a real presentation window with
  makeKeyAndVisible() in stripe-ios/StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/
  BNPLFormHeaderViewTests.swift:69. That is stronger than window.isHidden = false: it makes the window active/key, not just
  renderable. The test needs that because the tap path in stripe-ios/StripePaymentSheet/StripePaymentSheet/Source/
  PaymentSheet/Views/BNPLFormHeaderView.swift:70 calls PMMEInfoModal.present(from: self), and stripe-ios/
  StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMMEInfoModal.swift:132 presents
  via sourceView.window. Without a real active window, that modal presentation is not exercising the real code path.

  Once we create that key window, we should undo it. A leftover visible/key window can leak process-global UI state into
  later tests: top-most presenter lookup, trait environment, safe-area/layout behavior, and modal presentation state can
  all be affected. Most tests get away without explicit cleanup because they only need layout, not a presentation window,
  or because they rely on implicit cleanup. Here we are mutating app-wide window state on purpose, so explicit cleanup is
  the safer version.
```

yuki-stripe (approved):

PR #6412, opened 2026-05-06
PMM Experiment — Expanding Row Buttons UI

gbirch-stripe:
## Summary
Adds new variant of RowButton that contains PMME data. This variant can expand and contract to display this data and can also be initialized without the data present and then later have it added (since the PMME data will be lazily loaded)

## Motivation


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButton.swift
```diff
@@ -52,11 +73,6 @@ class RowButton: UIView, EventHandler {
         return (rowStyle == .flatWithCheckmark || rowStyle == .flatWithDisclosure) && isEmbedded
     }
 
-    var hasSubtext: Bool {
-        guard let subtext = sublabel.text else { return false }
-        return !subtext.isEmpty
-    }
-
```
gbirch-stripe, line 59: Previously this was only used for setting the row heights so I wrapped it into that func for better legibility

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButton.swift
```diff
@@ -289,99 +508,194 @@ extension RowButton: UIGestureRecognizerDelegate {
...
-                  badgeText: badgeText,
-                  promoBadge: promoBadge,
-                  accessoryView: accessoryView,
-                  shouldAnimateOnPress: shouldAnimateOnPress,
-                  isEmbedded: isEmbedded,
-                  didTap: didTap
-              )
-          }
-      }
+    static func create(
```
porter-stripe, line 312: This double-switch on paymentMethodMessaging inside each style case creates a lot of repeated code. Can wer refactor each subclass to have a single initializer that accepts both subtext and paymentMethodMessagingContent (one being nil), or you could extract a helper that picks the right initializer based on the configuration.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButton.swift
```diff
@@ -479,7 +826,21 @@ extension RowButton {
             }
             return paymentMethodType.displayName
         }()
+
+        let resolvedPaymentMethodMessaging: PaymentMethodMessagingConfiguration = {
```
porter-stripe: This closure returns the same value it was given in every branch (.disabled -> .disabled, .enabled -> the same .enabled). The comment says "Until PMME is wired to real data, keep BNPL rows on the legacy subtitle path" but the code doesn't actually do that, it just passes through whatever was given. Seems like we should just remove for now even if the intention is to change line 833/34 later.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButton.swift
```diff
@@ -92,6 +115,106 @@ class RowButton: UIView, EventHandler {
...
+                isPopulated: paymentMethodMessagingContent != nil
+            ),
+            badgeText: badgeText,
+            promoBadge: promoBadge,
+            accessoryView: accessoryView,
+            shouldAnimateOnPress: shouldAnimateOnPress,
+            isEmbedded: isEmbedded,
+            didTap: didTap
+        )
+        sublabel.delegate = self
```
porter-stripe: The delegate assignment happens after self.init(...), but it looks like `setupUI()` is called in the `init`, could this cause issues?

porter-stripe (commented):
RowButton should be a dumb container: show an image, a label, maybe a sublabel, handle taps. The PMME behavior is adding a lot of weight: expand/collapse animations, UITextViewDelegate, presentation, and a SublabelVariant enum that forces pattern matching everywhere it touches the sublabel.

I think the sublabel should just be a UIView that RowButton lays out without knowing what's inside it. The PMME view can handle its own expand/collapse, text, and link taps internally.

For the height constraint piece, the sublabel view could notify RowButton via a closure when it needs more space and the RowButton deactivates its height constraint and triggers a layout pass without knowing why. Then updateSelectedState() just forwards selection state to the sublabel and the sublabel decides what to do with it.

gbirch-stripe:
Closing in favor of https://github.com/stripe/stripe-ios/pull/6513

PR #6413, opened 2026-05-06
Await tasks before presenting Checkout integrations

porter-stripe:
## Summary
- Made the playground hook into the UIKit delegate and update if the CheckoutSession had changed
- Serialize Checkout mutations with a FIFO task queue
- The previous implementation used a counter (sessionUpdateCount) to track in-flight operations and show a .loading state, but it didn't actually enforce ordering so concurrent calls would race at suspension points. This PR replaces that with a proper FIFO task queue (pendingOperations) via enqueueSessionUpdate, so each operation waits for the previous one to complete before running.

### Races fixed

**Race 1: Concurrent mutations: undefined final state**
```swift
// Fired from two Tasks (e.g. rapid UI taps, React Native bridge)
Task { try await checkout.applyPromotionCode("SAVE10") }
Task { try await checkout.removePromotionCode() }
```

Old behavior with withSessionUpdateGuard: both tasks incremented the counter and immediately fired their requests in parallel. Two refreshSession calls then raced to complete, whichever finished last became the final state. The outcome (promo applied or removed) was nondeterministic.

New behavior: removePromotionCode waits in the queue until applyPromotionCode finishes. Final state is always "promo removed," as written.

**Race 2: Mutate then present results in crash**
```swift
checkout.applyPromotionCode("SAVE10")  // in-flight network call
paymentSheet.present(from: vc) { ... } // called before apply finishes
```

Old behavior: PaymentSheet checked checkout.state.isLoading, hit the assertionFailure, and returned an integration error, forcing the caller to somehow know to retry after the op finished.

New behavior: present calls awaitPendingOperations(), waits for the apply to complete, then re-snapshots the session (now with the promo applied) and presents with the correct state. The developer doesn't need to do anything special.

## Motivation
- Improve integration

## Testing
- Manual
- Existing tests

## Changelog
N/A


porter-stripe:
Dead code check seems to be flagging a function that is actually used.

Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
```diff
@@ -1384,6 +1385,28 @@ extension AddressViewController.AddressDetails {
...
+    func checkout(_ checkout: Checkout, didChangeState state: Checkout.State) {
+        switch settings.uiStyle {
+        case .embedded:
+            Task { @MainActor in
+                _ = await embeddedPlaygroundViewController?.embeddedPaymentElement?.update(checkout: checkout)
+            }
+        case .flowController:
+            paymentSheetFlowController?.update(checkout: checkout) { error in
+                if let error {
+                    print("PaymentSheet.FlowController.update(checkout:) failed: \(error)")
```
gbirch-stripe, line 1401: Can we present the error in the UI at the bottom of the screen like we do for other loading errors?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Internal.swift
```diff
@@ -0,0 +1,156 @@
...
+        }
+    }
+
+    // MARK: - Session Updates
+
+    /// Replaces the current session, preserves client-side overrides, and notifies the delegate.
+    ///
+    /// - Parameter applyOverrides: Called with the new session after existing overrides are
+    ///   preserved but before state is published. Use this to set client-side properties
+    ///   (e.g. address overrides) that should be visible to the delegate and observers.
```
gbirch-stripe, line 32: Nit: does breaking up the parameter comment across lines break the Swift docs parsing? Should we `/*` this documentation?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Internal.swift
```diff
@@ -0,0 +1,156 @@
...
+    func updateSession(_ newSession: STPCheckoutSession, applyOverrides: ((STPCheckoutSession) -> Void)? = nil) {
+        // Preserve client-side address overrides on the new session.
+        newSession.billingAddressOverride = stpSession?.billingAddressOverride
+        newSession.shippingAddressOverride = stpSession?.shippingAddressOverride
+        applyOverrides?(newSession)
+        newSession.onConfirmed = { [weak self] response in
+            self?.updateSession(response)
+        }
+        let changed = stpSession?.allResponseFields as NSDictionary? != newSession.allResponseFields as NSDictionary
```
gbirch-stripe, line 41: What's the purpose for specially handling the case where it's the same session?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Internal.swift
```diff
@@ -0,0 +1,156 @@
...
+            delegate?.checkout(self, didChangeState: state)
+        }
+    }
+
+    /// Runs `body` as a tracked session update, serialized behind any in-flight ops.
+    ///
+    /// Operations execute in strict FIFO order: each task waits for the previous
+    /// task before running its body. While the queue is non-empty, ``state`` is
+    /// `.loading`; once the queue drains it returns to `.loaded`.
+    internal func enqueueSessionUpdate(
```
gbirch-stripe, line 53: I wonder if at some point we should have a generalized Swift Concurrency serial queue implementation. This _looks_ good to me but I have some FUD about throwing in some one-off highly technical Swift concurrency code without thorough automated testing added (like we have for other concurrency utilities). This also makes we wonder if we can leverage tools built into Swift such as old-fashioned GCD serial dispatch queues, Actors (which implicitly serialize operations) or AsyncSequences

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
```diff
@@ -116,6 +117,37 @@ public final class Checkout: ObservableObject {
...
+    // MARK: - Pending Operations
+
+    /// Waits for all in-flight session updates (refresh, mutations, etc.) to complete.
+    ///
+    /// - Returns immediately if no operations are pending.
+    /// - If any pending operation throws, the first such error is rethrown.
+    /// - If the wait exceeds `timeout`, throws ``CheckoutError.timedOut``.
+    ///
+    /// - Parameter timeout: Maximum time to wait, in seconds.
+    func awaitPendingOperations(
```
gbirch-stripe, line 134: Making sure I understand this function right — if we call this but before it's done another thread adds an operation to the queue, this function will _not_ wait execute that operation? If this is right, I feel like it should be documented

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
```diff
@@ -129,8 +161,8 @@ public final class Checkout: ObservableObject {
         guard integrationDelegate?.isSheetPresented != true else {
             throw CheckoutError.sheetCurrentlyPresented
         }
-        try await withSessionUpdateGuard {
-            try await refreshSession()
+        try await enqueueSessionUpdate {
+            try await self.refreshSession()
```
gbirch-stripe, line 170: Are these intentional strong references? If for some reason this task gets hung then I think it will keep the reference.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Internal.swift
```diff
@@ -0,0 +1,156 @@
...
+//  StripePaymentSheet
+//
+//  Created by Nick Porter on 5/5/26.
+//
+
+import Foundation
+@_spi(STP) import StripeCore
+@_spi(STP) import StripePayments
+
+extension Checkout {
```
porter-stripe, line 12: This file is all existing code, just moved from Checkout.swift.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
```diff
@@ -248,7 +256,7 @@ public class PaymentSheet {
     // MARK: - Internal Properties
 
     /// The initialization mode this instance was initialized with
-    let mode: InitializationMode
+    var mode: InitializationMode
```
gbirch-stripe: Intuitively this feels dangerous. It looks like here we're just using the mutability to update the session id? But I can imagine there would be problems if someone changed this value in a different circumstance since the mutability generally is unexpected and unhandled throughout the SDK

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
```diff
@@ -669,24 +671,53 @@ extension PaymentSheet {
...
             assert(Thread.isMainThread, "PaymentSheet.FlowController.update must be called from the main thread.")
             assert(!isPresented, "PaymentSheet.FlowController.update must be when PaymentSheet is not presented.")
-            guard let stpSession = checkout.state.session as? STPCheckoutSession else {
-                stpAssertionFailure("Expected STPCheckoutSession, got \(type(of: checkout.state.session))")
-                completion(PaymentSheetError.unknown(debugDescription: "Invalid checkout session type"))
-                return
+            let updateID = beginUpdate()
+            Task { @MainActor in
+                do {
+                    try await checkout.awaitPendingOperations()
```
gbirch-stripe, line 686: Could there be unexpected behavior out of `awaitPendingOperations` having the behavior where there may still be unfinished operations that were started after it was called when it finishes?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Internal.swift
porter-stripe: Looks like this is handled properly by Xcode!
Screenshot 2026-05-06 at 3 34 57 PM

Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
porter-stripe: Good idea, updated!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Internal.swift
porter-stripe: The goal is to avoid notifying the delegate for no-op refreshes. We still call `setSession(newSession)` so the current session object and callbacks stay up to date, but we only call `didChangeState` when the raw session payload changed.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
porter-stripe: Agreed, this was broader than necessary. I changed `mode` back to immutable and now use a local `loadMode` in `present` after waiting for pending Checkout updates.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
porter-stripe: Good question. I think the snapshot behavior is okay here: this waits for the Checkout updates that were already pending when `update(checkout:)` was called.

If another mutation starts after that, it’s a new state change and should trigger another delegate update / `update(checkout:)` call rather than extending this one indefinitely.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
porter-stripe: Yep, intentional. These operations mutate Checkout state and need the Checkout instance to stay alive until the queued update finishes and clears its loading state.

Using weak self here feels riskier to me: the operation could disappear halfway through a mutation and leave the caller with ambiguous behavior. If an operation truly hangs, the awaiting call is also hung, so Checkout being retained seems expected.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
porter-stripe: Yep, that’s right. `awaitPendingOperations()` waits on a snapshot of the queue from when it’s called. I added that to the doc comment so the behavior is explicit.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Internal.swift
porter-stripe: Yep, I’m adding focused tests for this now. I agree this is the part that deserves test coverage.

I don’t think an actor alone solves this shape, since `Checkout` is already `@MainActor` and actor reentrancy means async methods can still interleave at suspension points. We need an explicit chain/queue to guarantee each update waits for the prior one before running.

gbirch-stripe (dismissed):

gbirch-stripe (approved):

PR #6415, opened 2026-05-06
Payment Sheet Playground - Allow Custom Customer ID for Customer Sessions

jkelle-stripe:
## Summary
This PR adds a `custom` customer ID text field to the playground sample app. This allows for testing with known customers IDs in the playground. 

## Motivation
Testing Link LPM-funded SPMs was a little tricky without creating a customer in CD2 and using it here. This textfield makes testing easier.

## Screenshot

CleanShot 2026-05-06 at 13 19 52

Demo video found [here](https://git.corp.stripe.com/stripe-internal/sandbox-apps/pull/1164).

PR #6416, opened 2026-05-06
Saved Payment Methods Funded by Link LPMs

jkelle-stripe:
## Summary
Renders Link LPMs as SPMs for customers. 

## Motivation

Link is adding LPMs for native Link. These payment details can become saved payment methods for customers so this PR enables rendering of these in MPE.

## Testing
I recommend cherry picking the commits (if not already merged) from #6415 and #6410 to set the proper customer session params and a sample customer ID. Then, log in with Link LPMs custom merchant and the account `jkelle+260502@stripe.com`.

## Demo

https://github.com/user-attachments/assets/9eaf3c10-a29c-4b76-8d33-092fba09d811


StripePayments/StripePayments/Source/API Bindings/Models/PaymentMethods/LinkPaymentDetails.swift
```diff
@@ -10,6 +10,7 @@ import Foundation
 @_spi(STP) @frozen public enum LinkPaymentDetails {
     case card(Card)
     case bankAccount(BankDetails)
+    case generic(Generic)
```
jkelle-stripe, line 13: @tillh-stripe What do we think of `Generic`? We could use this to replace `unparsable` for the SafeEnumDecodable work

tillh-stripe: Works for me!

StripePayments/StripePayments/Source/API Bindings/Models/PaymentMethods/LinkPaymentDetails.swift
```diff
@@ -85,6 +115,8 @@ import Foundation
             return "•••• \(cardDetails.last4)"
         case .bankAccount(let bankAccountDetails):
             return "••••\(bankAccountDetails.last4)"
+        case .generic(let genericDetails):
+            return genericDetails.sublabel ?? genericDetails.label
```
tillh-stripe: When is `formattedLast4` used? Do we need to provide a value for generic types here?

tillh-stripe: Chatted over Slack. They're used in the horizontal payment method layout in MPE. We should show `{label} {sublabel}` there instead of just the sublabel, as the sublabel doesn't include the payment method type.

tillh-stripe (approved):

PR #6417, opened 2026-05-07
[link] Use new branding for display labels

sophiah2-stripe:
## Summary

Consider branding in display label logic (i.e. display data) for the Link payment method option. Also redirect URLs (e.g. terms and privacy, checkout).

## Motivation

## Testing

Click around all Link surfaces within `PaymentSheet Example` app playground with `Link brand` configured, and ensure all branded strings are surfacing the correct brand. Also check for regressions in the default branding case.

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheetUITest/FCLiteUITests.swift
+++ b/Example/PaymentSheet Example/PaymentSheetUITest/FCLiteUITests.swift
@@ -133,11 +133,12 @@ class FCLiteUITests: XCTestCase {
         // Continue with Link
         let continueWithLinkButtonPredicate = NSPredicate(format: "label CONTAINS[cd] 'Continue with Link'") // Link signup pane
         let continueWithLinkButton = app.webViews.firstMatch.buttons.containing(continueWithLinkButtonPredicate).firstMatch
-        XCTAssertTrue(continueWithLinkButton.waitForExistenceAndTap(timeout: 10.0))
-
         // Success bank
         let paymentSuccessBankButtonPredicate = NSPredicate(format: "label CONTAINS[cd] 'Disputed'") // Institution Picker
         let paymentSuccessBankButton = app.webViews.firstMatch.buttons.containing(paymentSuccessBankButtonPredicate).firstMatch
+        XCTAssertTrue(continueWithLinkButton.waitForExistence(timeout: 10.0))
+        continueWithLinkButton.forceTapElement()
+        XCTAssertTrue(paymentSuccessBankButton.waitForExistence(timeout: 10.0))
         XCTAssertTrue(paymentSuccessBankButton.waitForExistenceAndTap(timeout: 10.0))
 
         // Connect account - wait for button to appear, then tap via coordinate
--- a/Example/PaymentSheet Example/PaymentSheetUITest/FCLiteUITests.swift
+++ b/Example/PaymentSheet Example/PaymentSheetUITest/FCLiteUITests.swift
@@ -137,8 +137,10 @@ class FCLiteUITests: XCTestCase {
         let paymentSuccessBankButtonPredicate = NSPredicate(format: "label CONTAINS[cd] 'Disputed'") // Institution Picker
         let paymentSuccessBankButton = app.webViews.firstMatch.buttons.containing(paymentSuccessBankButtonPredicate).firstMatch
         XCTAssertTrue(continueWithLinkButton.waitForExistence(timeout: 10.0))
-        continueWithLinkButton.forceTapElement()
-        XCTAssertTrue(paymentSuccessBankButton.waitForExistence(timeout: 10.0))
+        advancePrimaryCTA(
+            continueButton: continueWithLinkButton,
+            nextPaneElement: paymentSuccessBankButton
+        )
         XCTAssertTrue(paymentSuccessBankButton.waitForExistenceAndTap(timeout: 10.0))
 
         // Connect account - wait for button to appear, then tap via coordinate
@@ -175,5 +177,25 @@ class FCLiteUITests: XCTestCase {
 
         // Primary button is at the bottom center of the webview (roughly 95% down, centered)
         app.webViews.firstMatch.coordinate(withNormalizedOffset: CGVector(dx: 0.5, dy: 0.95)).tap()
+    }
+
+    private func advancePrimaryCTA(
+        continueButton: XCUIElement,
+        nextPaneElement: XCUIElement
+    ) {
+        if app.keyboards.firstMatch.exists || app.toolbars.buttons["Done"].exists {
+            app.typeText(XCUIKeyboardKey.return.rawValue)
+            if nextPaneElement.waitForExistence(timeout: 2.0) {
+                return
+            }
+        }
+
+        if continueButton.waitForExistenceAndTap(timeout: 2.0),
+           nextPaneElement.waitForExistence(timeout: 2.0) {
+            return
+        }
+
+        tapPrimaryButton()
+        XCTAssertTrue(nextPaneElement.waitForExistence(timeout: 10.0))
     }
 }
--- a/Example/PaymentSheet Example/PaymentSheetUITest/FCLiteUITests.swift
+++ b/Example/PaymentSheet Example/PaymentSheetUITest/FCLiteUITests.swift
@@ -133,8 +133,7 @@ class FCLiteUITests: XCTestCase {
         // Continue with Link
         let continueWithLinkButtonPredicate = NSPredicate(format: "label CONTAINS[cd] 'Continue with Link'") // Link signup pane
         let continueWithLinkButton = app.webViews.firstMatch.buttons.containing(continueWithLinkButtonPredicate).firstMatch
-        XCTAssertTrue(continueWithLinkButton.waitForExistence(timeout: 10.0))
-        app.typeText(XCUIKeyboardKey.return.rawValue) // Enter key will continue to the next screen
+        XCTAssertTrue(continueWithLinkButton.waitForExistenceAndTap(timeout: 10.0))
 
         // Success bank
         let paymentSuccessBankButtonPredicate = NSPredicate(format: "label CONTAINS[cd] 'Disputed'") // Institution Picker
--- a/StripeCore/StripeCore/Source/Connections Bindings/LinkBrand.swift
+++ b/StripeCore/StripeCore/Source/Connections Bindings/LinkBrand.swift
@@ -32,6 +32,15 @@ import Foundation
         }
     }
 
+    @_spi(STP) public var checkoutURL: URL {
+        switch self {
+        case .link, .unparsable:
+            return URL(string: "https://checkout.link.com/")!
+        case .onelink:
+            return URL(string: "https://checkout.onelink.com/")!
+        }
+    }
+
     @_spi(STP) public var termsURL: URL {
         switch self {
         case .link, .unparsable:
--- a/StripeCore/StripeCore/Source/Connections Bindings/LinkBrand.swift
+++ b/StripeCore/StripeCore/Source/Connections Bindings/LinkBrand.swift
@@ -121,14 +121,6 @@ import Foundation
             components.scheme = "https"
             components.host = "support.onelink.com"
             return components.url ?? url
-        case "stripe.com", "www.stripe.com":
-            if components.path == "/legal/end-users", components.fragment == "linked-financial-account-terms" {
-                return URL(string: "https://onelink.com/terms#financial-connections-terms")!
-            }
-            if components.path == "/privacy" {
-                return privacyURL
-            }
-            return url
         default:
             return url
         }
--- a/StripeCore/StripeCore/Source/Connections Bindings/LinkBrand.swift
+++ b/StripeCore/StripeCore/Source/Connections Bindings/LinkBrand.swift
@@ -22,4 +22,106 @@ import Foundation
             return "Onelink"
         }
     }
+
+    @_spi(STP) public var websiteURL: URL {
+        switch self {
+        case .link, .unparsable:
+            return URL(string: "https://link.com")!
+        case .onelink:
+            return URL(string: "https://onelink.com")!
+        }
+    }
+
+    @_spi(STP) public var termsURL: URL {
+        switch self {
+        case .link, .unparsable:
+            return URL(string: "https://link.co/terms")!
+        case .onelink:
+            return URL(string: "https://onelink.com/terms")!
+        }
+    }
+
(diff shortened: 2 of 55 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/MandateVariant.swift
sophiah2-stripe, line 1: Addresses this comment from a previous PR: https://github.com/stripe/stripe-ios/pull/6389#discussion_r3195858490

StripeCore/StripeCore/Source/Connections Bindings/LinkBrand.swift
```diff
@@ -22,4 +22,115 @@ import Foundation
...
+            return URL(string: "https://checkout.link.com/")!
+        case .onelink:
+            return URL(string: "https://checkout.onelink.com/")!
+        }
+    }
+
+    @_spi(STP) public var termsURL: URL {
+        switch self {
+        case .link, .unparsable:
+            return URL(string: "https://link.co/terms")!
```
jkelle-stripe, line 47: @sophiah2-stripe Is the discrepancy between .co and .com here intentional? It's in a few URLs here

StripeCore/StripeCore/Source/Connections Bindings/LinkBrand.swift
```diff
@@ -22,4 +22,115 @@ import Foundation
...
+                return promotionTermsURL
+            default:
+                return url
+            }
+        case "support.link.co", "www.support.link.co":
+            components.scheme = "https"
+            components.host = "support.onelink.com"
+            return components.url ?? url
+        case "stripe.com", "www.stripe.com":
+            if components.path == "/legal/end-users", components.fragment == "linked-financial-account-terms" {
```
jkelle-stripe: Is there context on why this one becomes onelink.com instead of stripe.com?

StripePaymentSheet/StripePaymentSheet/Source/Categories/STPPaymentMethod+PaymentSheet.swift
```diff
@@ -167,3 +129,47 @@ private func makeCardAccessibilityLabel(cardBrand: STPCardBrand, last4: String)
...
+        } else {
+            return STPPaymentMethodType.card.displayName
+        }
+    case .USBankAccount:
+        if paymentMethod.isLinkPassthroughMode {
+            return brand.displayName
+        } else {
+            return paymentMethod.usBankAccount?.bankName ?? paymentMethod.type.displayName
+        }
+    case .link:
```
jkelle-stripe: Was there a reason to add `.link` here when it wasn't before? Out of context it makes sense, but wondering what your thought was here.

StripePaymentSheet/StripePaymentSheet/Source/Categories/STPPaymentMethod+PaymentSheet.swift
```diff
@@ -167,3 +129,47 @@ private func makeCardAccessibilityLabel(cardBrand: STPCardBrand, last4: String)
     let localized = String.Localized.card_brand_ending_in_last_4
     return String(format: localized, brand, last4Spaced)
 }
+
+func paymentOptionLabel(for paymentMethod: STPPaymentMethod, confirmParams: IntentConfirmParams?, brand: LinkBrand) -> String {
```
jkelle-stripe: What was the thought here about bringing these out as free functions outside of the the `STPPaymentMethod` type? Generally in Swift, it's preferred to avoid free functions because of discovery. It's harder to know about free functions because you can't have Xcode give you code suggestions for what functions a type has (concretely: if you type `paymentMethod.` into Xcode, you'll get all of the functions supported by that type in a popup, but now these functions will be gone from that list).

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Components/NavigationBar/LinkSheetNavigationBar.swift
```diff
@@ -56,7 +58,8 @@ class LinkSheetNavigationBar: SheetNavigationBar {
         return super.leftmostElement
     }
 
-    override init(isTestMode: Bool, appearance: PaymentSheet.Appearance, shouldLogPaymentSheetAnalyticsOnDismissal: Bool = true) {
+    init(isTestMode: Bool, appearance: PaymentSheet.Appearance, brand: LinkBrand = .link, shouldLogPaymentSheetAnalyticsOnDismissal: Bool = true) {
```
jkelle-stripe: [nit] I think it's better if we don't have the default value for brand here. We instantiate this type so infrequently that the convenience added isn't worth it for the possible bugs introduced

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Elements/LinkEmailElement.swift
```diff
@@ -89,11 +90,11 @@ class LinkEmailElement: Element {
         }
     }
 
-    public init(defaultValue: String? = nil, isOptional: Bool = false, showLogo: Bool, theme: ElementsAppearance = .default) {
+    public init(defaultValue: String? = nil, isOptional: Bool = false, showLogo: Bool, theme: ElementsAppearance = .default, brand: LinkBrand = .link) {
```
jkelle-stripe: Same here regarding default value

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Verification/LinkVerificationController.swift
```diff
@@ -22,6 +23,7 @@ final class LinkVerificationController {
     init(
         mode: LinkVerificationView.Mode = .modal,
         linkAccount: PaymentSheetLinkAccount,
+        brand: LinkBrand = .link,
```
jkelle-stripe: Same here with default value. I basically feel this way about the default values for this, because it's so easy to forgot that Link has two brands

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkPaymentController.swift
```diff
@@ -376,14 +387,58 @@ import UIKit
         }
     }
 
-    private func continueWithPaymentMethod(_ paymentMethodId: String) {
+    private func continueWithPaymentMethod(_ paymentMethodId: String) async {
+        await fetchLinkBrandIfNeeded()
         self.paymentMethodId = paymentMethodId
-        payWithLinkContinuation?.resume(returning: ())
+        let continuation = payWithLinkContinuation
```
jkelle-stripe, line 393: What was the issue this resolved?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkPaymentController.swift
```diff
@@ -376,14 +387,58 @@ import UIKit
...
+                    withIntentConfig: intentConfiguration,
+                    clientDefaultPaymentMethod: nil,
+                    configuration: configuration
+                )
+                linkBrand = elementsSession.linkBrand
+            case .checkoutSession:
+                linkBrand = nil
+            }
+
+            fetchedLinkBrand = linkBrand
```
jkelle-stripe, line 438: I'm surprised we need to fetch the elements session from here. I would have thought we had access to it already

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/PaymentSheet-LinkConfirmOption.swift
```diff
@@ -44,6 +45,19 @@ extension PaymentSheet {
 // MARK: - Helpers
 
 extension PaymentSheet.LinkConfirmOption {
+    func paymentSheetSubLabel(brand: LinkBrand) -> String? {
+        let sublabel = paymentSheetSubLabel
+        switch sublabel {
+        case nil:
```
jkelle-stripe: [nit] this is a bit irregular. I'd do a `guard let sublabel = paymentSheetSubLabel else { return nil }` above the switch so you don't need to do it in here

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/PaymentSheet-LinkConfirmOption.swift
```diff
@@ -136,3 +139,14 @@ extension PaymentSheet.LinkConfirmOption {
         }
     }
 }
+
+func paymentSheetLabel(for option: PaymentSheet.LinkConfirmOption, brand: LinkBrand) -> String {
```
jkelle-stripe: Same here re: free functions

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/IntentConfirmParams.swift
```diff
@@ -38,36 +38,19 @@ final class IntentConfirmParams {
     var instantDebitsLinkedBank: InstantDebitsLinkedBank?
 
     var paymentSheetLabel: String {
+        return paymentSheetLabel(brand: .link)
```
jkelle-stripe, line 41: Should this always be `.link`?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/IntentConfirmParams.swift
```diff
@@ -38,36 +38,19 @@ final class IntentConfirmParams {
...
+    }
+
+    func paymentSheetLabel(brand: LinkBrand) -> String {
         if let last4 = (financialConnectionsLinkedBank?.last4 ?? instantDebitsLinkedBank?.last4) {
             return "••••\(last4)"
+        } else if paymentMethodType == .instantDebits || paymentMethodType == .linkCardBrand {
```
jkelle-stripe, line 47: Just me not know about instant debits, but should this always be based the current Link brand? We don't market this outside of Link?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/IntentConfirmParams.swift
```diff
@@ -212,6 +195,27 @@ final class IntentConfirmParams {
     }
 }
 
+func expandedPaymentSheetLabel(for confirmParams: IntentConfirmParams, brand: LinkBrand) -> String {
```
jkelle-stripe: Same here re: free functions

jkelle-stripe (changes requested):

StripeCore/StripeCore/Source/Connections Bindings/LinkBrand.swift
sophiah2-stripe: I'm not entirely sure why this is the case, but I based these off of the previous links used, which was `https://support.link.co/...`.

StripeCore/StripeCore/Source/Connections Bindings/LinkBrand.swift
sophiah2-stripe: good catch 😅 i had codex migrate the URLs over and it added this case by accident. will remove

StripePaymentSheet/StripePaymentSheet/Source/Categories/STPPaymentMethod+PaymentSheet.swift
sophiah2-stripe: I see, this makes sense! will move these back into `STPPaymentMethod`

StripePaymentSheet/StripePaymentSheet/Source/Categories/STPPaymentMethod+PaymentSheet.swift
sophiah2-stripe: This is to replace the current fallback approach, which uses `type.displayName`, with `brand.displayName`. Internally, the payment method type is still labeled as "link", but on the display, we want the label to be brand aware.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/IntentConfirmParams.swift
sophiah2-stripe: I believe in this case, `.instantDebits` refers to specifically the Link IBP payment method, so it would be resolved as the Link brand if we allow it to fall through

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/IntentConfirmParams.swift
sophiah2-stripe: This variable is for backwards compatibility with call sites that aren't relevant to the rebranding, so we can safely default to `.link` in these cases. The relevant call sites should always call `func paymentSheetLabel(brand: LinkBrand) -> String` instead!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkPaymentController.swift
sophiah2-stripe: resolved on slack (this is ok to keep)

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/LinkPaymentController.swift
sophiah2-stripe: This was an issue marked by codex code review:

> Before this branch, the code effectively did:
`payWithLinkContinuation?.resume(returning: ())`
and in one path the continuation was nulled out elsewhere afterward. Once `continueWithPaymentMethod` started awaiting `fetchLinkBrandIfNeeded()`, that became more fragile for two reasons:
> - there is now a suspension point before resuming, so success/failure callbacks can race more easily
> - leaving `payWithLinkContinuation` set while calling resume makes it easier to accidentally resume the same continuation twice if another path re-enters before state is cleared

jkelle-stripe (dismissed):

Example/PaymentSheet Example/PaymentSheetUITest/FCLiteUITests.swift
sophiah2-stripe, line 1: Test is flaky: the text field doesn't always auto-focus, resulting in a missing keyboard in some cases. This new approach considers both focused and unfocused cases

jkelle-stripe (dismissed):

jkelle-stripe (approved):

PR #6418, opened 2026-05-07
Fix flaky testSavedSEPA by isolating Customer per intent kind

jaynewstrom-stripe:
## Summary
- Fixes intermittent lock contention failure in `PaymentSheetLPMConfirmFlowTests.testSavedSEPA`
- Root cause: all intent variants shared a single Customer object; each server-side confirmation triggers async processing (mandate creation, PM attachment) that locks the Customer, causing later iterations to fail with "This object cannot be accessed right now"
- Fix: create a fresh Customer and saved SEPA PaymentMethod per intent kind so lock pressure doesn't accumulate across iterations

## Test plan
- [ ] Run `testSavedSEPA` multiple times to verify it no longer flakes
- [ ] CI passes for StripePaymentSheet scheme


StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetLPMConfirmFlowTests.swift
```diff
@@ -402,49 +402,51 @@ final class PaymentSheetLPMConfirmFlowTests: STPNetworkStubbingTestCase {
...
-            config.returnURL = "https://foo.com"
-            config.customer = PaymentSheet.CustomerConfiguration(
-                id: customerAndCustomerSession.customer,
-                customerSessionClientSecret: customerAndCustomerSession.customerSessionClientSecret
-            )
-            return config
-        }()
-
         // Confirm saved SEPA with every confirm variation
+        // Use a fresh customer per intent kind to avoid lock contention on the Customer object
```
jaynewstrom-stripe, line 406: This was flakey because of lock contention on the customer object see http://go/o/req_C1MabOSKibtquq

joyceqin-stripe (approved):

PR #6427, opened 2026-05-07
Update data models for better usage

porter-stripe:
## Summary
Reshapes the data model we expose on `Checkout.Session` to be slightly more ergonomic and consistent across platforms.

## Motivation
- CheckoutSessions

## Testing
- Updated tests
- Tested in playground

## Changelog
N/A


StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Checkout/STPCheckoutSessionDiscountTests.swift
```diff
@@ -207,22 +208,20 @@ final class STPCheckoutSessionDiscountTests: XCTestCase {
...
+    func testCouponWithoutNameFallsBackToId() {
         let dict: [AnyHashable: Any] = [
+            "currency": "usd",
             "line_item_group": [
                 "discount_amounts": [
                     [
                         "amount": 250,
                         "coupon": [
-                            "name": "No ID Coupon",
+                            "id": "coupon_no_name",
```
wooj-stripe: Nit: maybe `coupon_no_id` ?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/STPApplePayContext+CheckoutSessionLineItems.swift
```diff
@@ -29,44 +29,52 @@ extension STPApplePayContext {
...
             let amount = NSDecimalNumber.stp_decimalNumber(
-                withAmount: lineItem.unitAmount * lineItem.quantity,
+                withAmount: unitMinorUnits * lineItem.quantity,
                 currency: currency
             )
             summaryItems.append(PKPaymentSummaryItem(label: label, amount: amount, type: .final))
         }
 
+        let shipping = total.shippingRate.minorUnitsAmount
+        let tax = total.taxExclusive.minorUnitsAmount + total.taxInclusive.minorUnitsAmount
```
wooj-stripe: For `tax`, seems like it should only look at taxInclusive?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/STPCheckoutSession.swift
```diff
@@ -315,56 +331,96 @@ extension STPCheckoutSession: STPAPIResponseDecodable {
...
-                total: total,
-                due: due,
-                discount: discounts.reduce(0) { $0 + $1.amount },
-                shipping: Self.parseSelectedShippingAmount(from: dict),
-                tax: taxAmounts.reduce(0) { $0 + $1.amount }
+            // Aggregate inclusive vs exclusive tax from internal taxAmounts.
+            let taxInclusiveValue = internalTaxAmounts.filter { $0.inclusive }.reduce(0) { $0 + $1.amount }
+            let taxExclusiveValue = internalTaxAmounts.filter { !$0.inclusive }.reduce(0) { $0 + $1.amount }
+            let shippingValue = Self.parseSelectedShippingAmount(from: dict)
+            let discountValue = discountAmounts.reduce(0) { $0 + $1.amount.minorUnitsAmount }
```
wooj-stripe, line 381: Just want to confirm that the calculations for taxInclusive/exclusive and discount are just to power the UI, correct? Assuming the answer is yes, i'm still surprised that we're having to calculate this on the client.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/STPCheckoutSession.swift
```diff
@@ -315,56 +331,96 @@ extension STPCheckoutSession: STPAPIResponseDecodable {
...
+                total: makeAmount(totalValue, currency: currency),
+                appliedBalance: makeAmount(appliedBalanceValue, currency: currency),
+                balanceAppliedToNextInvoice: balanceAppliedToNextInvoice
             )
         }()
 
-        // Parse saved payment methods offer save configuration
+        // Tax
+        let tax = Checkout.Tax(
+            status: .unknown,
```
wooj-stripe: why .unknown here?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/STPCheckoutSession.swift
```diff
@@ -510,12 +601,79 @@ extension STPCheckoutSession {
             return nil
         }
         let price = dict["price"] as? [AnyHashable: Any]
-        let amount = (price?["unit_amount"] as? Int) ?? 0
-        let currency = (price?["currency"] as? String) ?? defaultCurrency ?? "usd"
-        return Checkout.LineItem(id: id, name: name, quantity: quantity, unitAmount: amount, currency: currency)
+        let currency = (price?["currency"] as? String) ?? defaultCurrency
```
wooj-stripe, line 603: previously we defaulted to `usd`, but now currency can still be null.. is this expected?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/STPCheckoutSession.swift
```diff
@@ -533,48 +691,186 @@ extension STPCheckoutSession {
...
+                deliveryEstimate: deliveryEstimate
+            )
         }
         if let id = dict["shipping_rate"] as? String {
-            return Checkout.ShippingOption(id: id, displayName: "Shipping Option", amount: 0, currency: defaultCurrency ?? "usd")
+            let currency = defaultCurrency ?? "usd"
+            return Checkout.ShippingOption(
+                id: id,
+                displayName: nil,
+                amount: makeAmount(0, currency: currency),
```
wooj-stripe: why is shipping rate hardcoded to 0?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/STPCheckoutSession.swift
```diff
@@ -533,48 +691,186 @@ extension STPCheckoutSession {
...
         return nil
     }
 
+    private static func parseDeliveryEstimate(_ dict: [AnyHashable: Any]?) -> Checkout.DeliveryEstimate? {
+        guard let dict else { return nil }
+        let minimum = parseDeliveryBound(dict["minimum"] as? [AnyHashable: Any])
+        let maximum = parseDeliveryBound(dict["maximum"] as? [AnyHashable: Any])
+        if minimum == nil && maximum == nil {
+            return nil
+        }
```
wooj-stripe: Prefer guard over if

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/STPCheckoutSession.swift
```diff
@@ -533,48 +691,186 @@ extension STPCheckoutSession {
...
+
+    static func makeCurrencyOptions(
+        from metas: [STPCheckoutSessionLocalizedPriceMeta],
+        exchangeRateMeta: STPCheckoutSessionExchangeRateMeta?
+    ) -> [Checkout.CurrencyOption]? {
+        guard !metas.isEmpty else { return nil }
+        return metas.map { meta -> Checkout.CurrencyOption in
+            let conversion: Checkout.CurrencyConversion?
+            if let exchangeRateMeta,
+               meta.currency.lowercased() == exchangeRateMeta.localizedCurrency.lowercased() {
```
wooj-stripe: This logic reads awkwardly to me. I would have expected that if `meta.currency.lowercased() != exchangeRateMeta.localizedCurrency.lowercased()`, then there would be a conversion.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/STPCheckoutSession.swift
```diff
@@ -492,6 +565,24 @@ extension STPCheckoutSession {
...
+    /// Builds a ``Checkout/Amount`` from a minor-unit value plus the session currency.
+    static func makeAmount(_ minorUnits: Int, currency: String?) -> Checkout.Amount {
+        let formatted: String
+        if let currency, !currency.isEmpty {
+            formatted = String.localizedAmountDisplayString(for: minorUnits, currency: currency)
+        } else {
+            formatted = "\(minorUnits)"
+        }
+        return Checkout.Amount(amount: formatted, minorUnitsAmount: minorUnits)
+    }
```
wooj-stripe: I'm a confused by this function, is minorUnits the same as minorUnitsAmount, or is it something different?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/STPApplePayContext+CheckoutSessionLineItems.swift
porter-stripe: Good catch that this was wrong. I looked at what web does, they only use taxExclusive here since taxInclusive is already baked into the line item prices. Updated to use only taxExclusive to stay consistent.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/STPCheckoutSession.swift
porter-stripe: Yes, these are just to power the UI (Apple Pay line items + order summary breakdown). Web does the same client-side computation. I checked the server's TotalSummary struct and it only has subtotal, total, and due. It doesn't serve tax_exclusive, tax_inclusive, shipping_rate, or discount. So the server-read fallbacks were dead code. Simplified to always compute client-side, consistent with web.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/STPCheckoutSession.swift
porter-stripe: oof, I left this but forgot a TODO. I need some parsing code, good flag! Updated to compute the correct status from tax_meta, matching web's getTaxStatus logic: if computation_type is "automatic" and status is "requires_location_inputs", we check for shipping_address_collection to decide between .requiresShippingAddress and .requiresBillingAddress. Otherwise .ready.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/STPCheckoutSession.swift
porter-stripe: Yes, this is intentional. defaultCurrency is the session-level currency which is always present except in setup mode (where there are no line items anyway).

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/STPCheckoutSession.swift
porter-stripe: Good catch. Removed this fallback entirely. The server always sends shipping_rate as an expanded object. Returning nil is better than surfacing a shipping option with a misleading 0 amount.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/STPCheckoutSession.swift
porter-stripe: Refactored to use flatMap with a comment clarifying that conversion details are only attached to the converted (localized) currency option.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/STPCheckoutSession.swift
porter-stripe: yeah its the same, renamed!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/STPCheckoutSession.swift
```diff
@@ -202,53 +210,56 @@ class STPCheckoutSession: NSObject {
...
 
     private init(
-        stripeId: String,
+        id: String,
         clientSecret: String?,
-        totals: Checkout.Totals?,
+        businessName: String?,
         currency: String?,
-        mode: Checkout.Mode,
+        currencyOptions: [Checkout.CurrencyOption]?,
```
wooj-stripe: Feels odd to me that `currencyOptions` is the only optional List parameter.  All other list params are non optional (because i'm assuming that if those field don't have a value, they would just be empty lists)

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/STPCheckoutSession.swift
```diff
@@ -510,12 +600,79 @@ extension STPCheckoutSession {
...
+        case let v as String: decimal = Decimal(string: v)
+        default: decimal = nil
+        }
+        guard let decimal else { return nil }
+        let intValue = NSDecimalNumber(decimal: decimal).intValue
+        return Checkout.DecimalAmount(
+            amount: makeAmount(intValue, currency: currency).amount,
+            minorUnitsAmount: decimal
+        )
+    }
```
wooj-stripe: This feels bad that we're having to check the type (e.g. we don't even know if it's a number of a string?)-- we don't know the type that will be sent to us?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/STPCheckoutSession.swift
```diff
@@ -510,12 +600,79 @@ extension STPCheckoutSession {
...
+
+    private static func makeOptionalAmount(_ value: Any?, currency: String?) -> Checkout.Amount? {
+        guard let int = value as? Int else { return nil }
+        return makeAmount(int, currency: currency)
```
wooj-stripe: `guard let amount = value`?  using int as a variable feels weird.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/STPCheckoutSession.swift
```diff
@@ -533,48 +690,169 @@ extension STPCheckoutSession {
...
+            )
         }
         return nil
     }
 
+    private static func parseDeliveryEstimate(_ dict: [AnyHashable: Any]?) -> Checkout.DeliveryEstimate? {
+        guard let dict else { return nil }
+        let minimum = parseDeliveryBound(dict["minimum"] as? [AnyHashable: Any])
+        let maximum = parseDeliveryBound(dict["maximum"] as? [AnyHashable: Any])
+        guard minimum != nil || maximum != nil else { return nil }
```
wooj-stripe, line 731: I can't tell what the expected behavior is here.  Just to be clear, we want to ensure that there is a value for minimum or maximum?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/STPCheckoutSession.swift
porter-stripe: Yeah, the over-broad switch was unnecessary defensiveness on my part. We can simplify.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/STPCheckoutSession.swift
porter-stripe: Made currencyOptions non-optional [Checkout.CurrencyOption] to match the other list params.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/STPCheckoutSession.swift
porter-stripe: Renamed to amount.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/STPCheckoutSession.swift
porter-stripe: If neither bound parses, there's nothing useful to surface, so we return nil instead of an empty DeliveryEstimate(nil, nil). Added a comment.

wooj-stripe (approved):

PR #6428, opened 2026-05-08
[link]  Refactor Link brand propagation across PaymentSheet and Financial Connections

sophiah2-stripe:
## Summary

This PR cleans up how Link brand is resolved and threaded through SDK-owned surfaces.

## Motivation

## Testing

Check for any regressions

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentElementConfiguration.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentElementConfiguration.swift
@@ -51,8 +51,7 @@ protocol PaymentElementConfiguration: PaymentMethodRequirementProvider {
 
 extension PaymentElementConfiguration {
     func resolvedLinkBrand(elementsSession: STPElementsSession) -> LinkBrand {
-        // `link.brand` is an explicit client override; `nil` means defer to the elements session response.
-        link.brand ?? elementsSession.linkBrand ?? .link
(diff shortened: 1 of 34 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentElementConfiguration.swift
```diff
@@ -51,8 +51,7 @@ protocol PaymentElementConfiguration: PaymentMethodRequirementProvider {
 
 extension PaymentElementConfiguration {
     func resolvedLinkBrand(elementsSession: STPElementsSession) -> LinkBrand {
-        // `link.brand` is an explicit client override; `nil` means defer to the elements session response.
-        link.brand ?? elementsSession.linkBrand ?? .link
+        link.effectiveBrand(elementsSession: elementsSession)
```
jkelle-stripe, line 54: [nit] `effective` and `resolved` as words feel like they're solving the same problem but we're using both of them. Happy to keep terms both if you feel strongly, but if not, would love to just use one. Non blocking for this

jkelle-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentElementConfiguration.swift
sophiah2-stripe: Makes sense, I'll align these in the next PR

PR #6433, opened 2026-05-08
Add experimental logic to control card art

wooj-stripe:
## Summary
Puts card art behind a server-side experiment flag to support A/A and A/B testing

## Motivation
Exercising experimentation framework

## Testing
Will add tests

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift
@@ -222,20 +222,9 @@ final class PaymentSheetLoader {
                 paymentMethodTypes: paymentMethodTypes
             )
 
-            let loadResult = LoadResult(
-                intent: intent,
+            let paymentMethodOrientation = configuration.resolveLayout(
                 elementsSession: elementsSession,
-                savedPaymentMethods: filteredSavedPaymentMethods,
-                paymentMethodTypes: paymentMethodTypes,
-                paymentMethodMessagingPromotionsHelper: paymentMethodMessagingPromotionsHelper,
-                paymentMethodOrientation: configuration.resolveLayout(
-                    elementsSession: elementsSession,
-                    paymentMethodTypes: paymentMethodTypes
-                )
-            )
-            let confirmationChallenge = ConfirmationChallenge(
-                elementsSession: elementsSession,
-                stripeAttest: configuration.apiClient.stripeAttest
+                paymentMethodTypes: paymentMethodTypes
             )
 
             // This is hacky; the logic to determine the default selected payment method belongs to the SavedPaymentOptionsViewController. We invoke it here just to report it to analytics before that VC loads.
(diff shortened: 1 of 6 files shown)


porter-stripe (dismissed):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift
```diff
@@ -249,6 +238,19 @@ final class PaymentSheetLoader {
                 defaultPaymentMethod: elementsSession.customer?.getDefaultPaymentMethod()
             )
 
+            // Log card art experiment exposure
```
joyceqin-stripe, line 241: Why do we need to create and log the experiment before loading? Also, we generally only want to log the experiment exposures when we get to logic for whether or not to change behavior.

wooj-stripe: I added the creation and the logging of the experiment in loading because it's a central area shared by PS, PS.FC, and Embedded.  If we don't do this, then we'll need to create the experiment in different locations.

We are logging the experiment exposure exactly when we receive the response from elements/sessions -- are you saying that we should only be logging exposure when card art vended?  We thought about this, but thought that this would be more difficult to maintain, and difficult to deal with edge cases.  For example, there could be a case where we are in treatment, but there is no card art to show (because it's not available).

Are there reasons why not to create and log the experiment during loading?  Similarly, are there reasons only to log experiment exposures exactly when you see the behavior change? I believe I've effectively implemented what we have on Android, so, if this a problem, then we'll need to change it on both platforms. cc: @toluo-stripe

joyceqin-stripe: My thought here is that you don't want to log the exposure at a point where you're unsure if the user's experience would be different. At loading time, the user might not have any saved pms, so wouldn't it be possible for a user to just enter card details into the form and pay without seeing the card art experience? If so, then we can't really include that in the experiment results because the thing you're testing doesn't affect their experience or decisions in any way. The [trailhead](https://trailhead.corp.stripe.com/docs/mobile-sdk/running-an-experiment) says: "Log an exposure event every time you check an experiment's assignment. This is critical to avoid biased experiment results." cc @amk-stripe 

And even if we were to log the experiment here, I'm still wondering why we need to call `resolveLayout` and create the experiment prior to getting the `LoadResult`, which contains the `paymentMethodOrientation`

toluo-stripe: I think we discussed this with @amk-stripe when setting this up. We have dimensions that tell us if the user has a saved pm and if any of the pms have card art. We plan to use those filters to understand how card art performs

wooj-stripe: > And even if we were to log the experiment here, I'm still wondering why we need to call resolveLayout and create the experiment prior to getting the LoadResult, which contains the paymentMethodOrientation

I think i'm missing something, but here's my understanding:
Line 129: elementsSessions finishes loading
Line 225: paymentMethodOrientation is resolved
Line 248: paymentMethodOrientation is passed into the newly proposed CardArtExperiment so that its value can be a logged as a dimension
Line 263: elementsSession & paymentMethodOrientation are passed into an object called LoadResult.init() so that it can be returned to the user (we aren't "getting" a LoadResult, we are just constructing it with previously computed values)

joyceqin-stripe: > > And even if we were to log the experiment here, I'm still wondering why we need to call resolveLayout and create the experiment prior to getting the LoadResult, which contains the paymentMethodOrientation
> 
> I think i'm missing something, but here's my understanding: Line 129: elementsSessions finishes loading Line 225: paymentMethodOrientation is resolved Line 248: paymentMethodOrientation is passed into the newly proposed CardArtExperiment so that its value can be a logged as a dimension Line 263: elementsSession & paymentMethodOrientation are passed into an object called LoadResult.init() so that it can be returned to the user (we aren't "getting" a LoadResult, we are just constructing it with previously computed values)

Oh wait, sorry that's on me I got confused 😅 strike that

joyceqin-stripe: > I think we discussed this with @amk-stripe when setting this up. We have dimensions that tell us if the user has a saved pm and if any of the pms have card art. We plan to use those filters to understand how card art performs

If that's the case, I'll defer to @amk-stripe. My main concern is just that there will be logged exposures where card art had no bearing on the user's decisions, but if it'll be filtered out, I understand the convenience of doing this in load.

porter-stripe (approved):

PR #6437, opened 2026-05-11
Add Checkout.runServerUpdate

porter-stripe:
## Summary
- Adds `runServerUpdate`, similar to web: https://docs.stripe.com/js/custom_checkout/run_server_update

## Motivation
- CheckoutSessions

## Testing
- New unit tests

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
@@ -60,6 +60,9 @@ public final class Checkout: ObservableObject {
     /// Default timeout used by ``awaitPendingOperations(timeout:)``.
     nonisolated static let defaultPendingOperationsTimeout: TimeInterval = 30
 
+    /// Timeout enforced on the merchant's closure in ``runServerUpdate(_:)``.
+    nonisolated static let serverUpdateTimeout: TimeInterval = 20
+
     /// The single state writer. Publishes `.loading` if any op is queued, else `.loaded`.
     func setSession(_ session: Checkout.Session) {
         state = pendingOperations.isEmpty ? .loaded(session) : .loading(session)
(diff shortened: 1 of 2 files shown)


StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Checkout/CheckoutUnitTests.swift
```diff
@@ -146,7 +146,64 @@ final class CheckoutUnitTests: XCTestCase {
...
+            XCTFail("Expected CheckoutError.sheetCurrentlyPresented")
+        } catch let error as CheckoutError {
+            guard case .sheetCurrentlyPresented = error else {
+                XCTFail("Expected .sheetCurrentlyPresented, got \(error)")
+                return
+            }
+        } catch {
+            XCTFail("Unexpected error type: \(error)")
+        }
+    }
```
wooj-stripe: How important are these tests? They seem to just test the call to requireOpenSession (which i'd consider more or less a precondition of `runServerUpdate`, rather than testing the guts of `runServerUpdate`.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Checkout/CheckoutUnitTests.swift
```diff
@@ -146,7 +146,64 @@ final class CheckoutUnitTests: XCTestCase {
...
+
+    func testRunServerUpdateWrapsClosureError() async {
```
wooj-stripe: maybe we should test two more cases:
1. the function throws a timeout error
2. the function throws some generic error.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
```diff
@@ -295,6 +298,36 @@ public final class Checkout: ObservableObject {
...
+            let result = await withTimeout(Self.serverUpdateTimeout) {
+                try await updateFunction()
+            }
+            if case .failure(let error) = result {
+                if error is TimeoutError {
+                    throw CheckoutError.timedOut
+                }
+                throw CheckoutError.apiError(message: error.localizedDescription)
+            }
+            try await self.refreshSession()
```
wooj-stripe, line 328: In the time that we've called the user's server, do we need to check that we still have an open session?

porter-stripe: Good question, I think no, generally only occurs after confirmation happens, but it's possible the session expires. I'll add a check.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Checkout/CheckoutUnitTests.swift
porter-stripe: Added!

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Checkout/CheckoutUnitTests.swift
porter-stripe: Yeah, probably don't need them.

wooj-stripe (approved):

PR #6443, opened 2026-05-11
Switch CI from Xcode 16 to Xcode 26.4

davidme-stripe:
## Summary
- Remove all Xcode 16 build/test workflows (`test-builds-xcode-16`, `test-builds-xcode-16-release`)
- Switch default CI stack to `osx-xcode-26.4.x` for all jobs
- Add `xcodes runtimes install "iOS 16.4"` to most jobs (it's not bundled with Xcode 26)
- Add `install-runtime-15` workflow for `legacy-tests-15` (iOS 15.5)
- Remove redundant `test-builds-xcode-26` and `test-builds-xcode-26-release`, which should now be covered by other builds
- Keep one `test-builds-xcode-26-1` job on `osx-xcode-26.1.x` for backwards compatibility
- Update bundler

## Test plan
- [x] Verify main pipeline triggers and builds pass on Xcode 26.4
- [x] Run nightly pipeline to validate all nightly jobs: https://app.bitrise.io/app/b220f4ba85d134a7/build/1c0a9358-047f-426d-9d4f-5218ba54b4ed


Testers/IntegrationTester/IntegrationTesterUITests/IntegrationTesterUIPMTests.swift
```diff
@@ -41,17 +41,37 @@ class IntegrationTesterUIPMTests: IntegrationTesterUITests {
...
+            XCTAssertTrue(amexButton.waitForExistence(timeout: 10.0))
+            amexButton.forceTapElement()
+
+            let mastercardPredicate = NSPredicate(format: "label CONTAINS 'Simulated Card - MasterCard, ‪•••• 1234‬'")
+            let mastercardButton = applePay.buttons.containing(mastercardPredicate).firstMatch
+            XCTAssertTrue(mastercardButton.waitForExistence(timeout: 10.0))
+            mastercardButton.forceTapElement()
+
+            let payButton = applePay.buttons["Pay with Passcode"]
+            if payButton.waitForExistence(timeout: 5.0) {
```
wooj-stripe: This doesn't seem optional, feels like it should be :
```
XCTAssertTrue(payButton.waitForExistence(timeout: 5.0))
payButton.forceTapElement()
```

Testers/IntegrationTester/IntegrationTesterUITests/IntegrationTesterUITests.swift
```diff
@@ -135,16 +135,13 @@ class IntegrationTesterUITests: XCTestCase {
...
-        XCTAssertTrue(doneButton.waitForExistence(timeout: 10.0))
-        doneButton.forceTapElement()
-
-        let submitCodeButton = app.buttons["Submit"]
-        XCTAssertTrue(submitCodeButton.waitForExistence(timeout: 10.0))
-        submitCodeButton.forceTapElement()
+        // On iOS 26, the WKWebView form accessory "Done" button isn't reachable
+        // via standard XCTest queries. Type Return to dismiss the keyboard and
+        // submit the form.
+        enterCodeText.typeText("\n")
```
wooj-stripe, line 141: Do we need to put something like: `if ProcessInfo.processInfo.operatingSystemVersion.majorVersion >= 26 {` ?

bitrise.yml
```diff
@@ -49,10 +49,7 @@ pipelines:
         abort_on_fail: true
       framework-tests-no-mocks: {}
       framework-tests-ios-26: {}
-      test-builds-xcode-16: {}
-      test-builds-xcode-16-release: {}
-      test-builds-xcode-26: {}
-      test-builds-xcode-26-release: {}
+      test-builds-xcode-26-1: {}
```
wooj-stripe, line 52: What is the reason for removing -release builds?

bitrise.yml
```diff
@@ -536,10 +500,24 @@ workflows:
...
+        title: Install iOS 15.5 simulator runtime
+        inputs:
+        - content: |-
+            #!/usr/bin/env bash
+            set -euo pipefail
+            if xcrun simctl list runtimes | grep -q "iOS 15.5"; then
+              echo "iOS 15.5 runtime already installed"
+            else
+              echo "Installing iOS 15.5 runtime..."
+              sudo xcodes runtimes install "iOS 15.5"
```
wooj-stripe, line 519: Wow, we can just run as sudo here?!  I'm hoping this is just limited to the xcodes binary?  Maybe we should use the absolute path of xcodes?

bitrise.yml
davidme-stripe: I think it's covered by `deploy-dry-run`? We run `export_builds.rb` there.

bitrise.yml
davidme-stripe: This is running in a CI VM, so yes this is intentional! We do need root permission to write to /Library/Developer where the simulator images are stored. I'd prefer to not hardcode the path, the location of xcodes is an implementation detail of how Bitrise bundles the VM image.

Testers/IntegrationTester/IntegrationTesterUITests/IntegrationTesterUITests.swift
davidme-stripe: No, I think typing "\n" should work on older OSes as well.

Testers/IntegrationTester/IntegrationTesterUITests/IntegrationTesterUIPMTests.swift
davidme-stripe: Good point, fixed!

bitrise.yml
wooj-stripe: I looked into this a little bit, and `deploy-dry-run` does seem to run `export_builds.rb` which does cover this case, but `deploy-dry-run` only runs during the nightly pipeline, and not the `releases-trigger-pipeline`

bitrise.yml
wooj-stripe: I don't feel too strongly about this, sounds good.

bitrise.yml
davidme-stripe: That seems ~correct! It should be rare for release builds to break separately from debug ones (so nightly is a reasonable cadence), and `export_builds.rb` runs in the Xcode Cloud job on release.

wooj-stripe: Ok.. I thinking this through since it's 'rare', seems like nightly is fine.

wooj-stripe (approved):

PR #6450, opened 2026-05-14
Add Stripe flags to CurrencySelector

porter-stripe:
## Summary
- Adds downloading of Stripe flag icons from our CDN after we uploaded them to be shown in the currency selector
- More details here https://docs.google.com/document/d/1CWglcOHR6I8cOWKjztPfnhnVG8axucpv-j5cWIoey7E/edit?tab=t.ixyh82ib6clh#bookmark=id.6t3eoosszz64

## Motivation
- Design feedback

## Testing
- Snapshots

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/AdaptivePricingFlagImageManager.swift
```diff
@@ -0,0 +1,159 @@
...
+    private static func imageAttachment(for image: UIImage) -> NSAttributedString {
+        let attachment = NSTextAttachment()
+        attachment.image = image.withRenderingMode(.alwaysOriginal)
+        let imageHeight: CGFloat = 14
+        let imageWidth = imageHeight * (image.size.width / image.size.height)
+        let font = UIFont.preferredFont(forTextStyle: .footnote)
+        let yOffset = (font.capHeight - imageHeight) / 2
+        attachment.bounds = CGRect(x: 0, y: yOffset, width: imageWidth, height: imageHeight)
+        return NSAttributedString(attachment: attachment)
+    }
```
gbirch-stripe: Is there a reason we are hardcoding the image size (to height 14) regardless of the font size? I wrote a font-adaptive solution for PMME so that the image appears as the same size as the surrounding text: https://github.com/stripe/stripe-ios/blob/581e57b2b7d234ce29e8ff9c5dd9ec361c8b2dd1/StripePaymentSheet/StripePaymentSheet/Source/Categories/NSAttributedString%2BStripe.swift#L23-L35

One thing to note here is that if you use it it's preferable as a minor polish detail to re-render the image on `traitCollectionDidChange` so that it adapts to font size updates: https://github.com/stripe/stripe-ios/blob/581e57b2b7d234ce29e8ff9c5dd9ec361c8b2dd1/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMMESinglePartnerView.swift#L60-L63

porter-stripe: Great suggestion, incorporated this!

gbirch-stripe:
Could we add a snapshot test to verify the flag scaling functionality?

porter-stripe:
@gbirch-stripe Do you mind if I add that text in the immediately following PR? We're adding an appearance property sizeScaleFactor to scale fonts and will have a snapshot test for it.

gbirch-stripe (approved):

PR #6460, opened 2026-05-14
Update CurrencySelector Appearance to match figma

porter-stripe:
## Summary
- Updates design of selector to match the figma
- Matches the appearance API to the final list defined here: https://www.figma.com/design/M2S5vFdvpXYuEtplf8eAeZ/Checkout-Sessions?node-id=0-1&p=f&t=EzdcMpA0or397qRC-0

## Motivation
- Design

## Testing
- Snapshots

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/AdaptivePricingFlagImageManager.swift
```diff
@@ -73,10 +73,9 @@ final class AdaptivePricingFlagImageManager {
     ///
     /// If a downloaded image is available, the string contains a text attachment;
     /// otherwise it contains the corresponding emoji flag.
-    func flagIcon(for currency: String) -> NSAttributedString {
+    func flagIcon(for currency: String, font: UIFont) -> NSAttributedString {
```
gbirch-stripe, line 79: Could we add some documentation here about what the font parameter does? Not intuitive w/o context

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+CurrencySelectorViewAppearance.swift
```diff
@@ -12,62 +12,85 @@ import UIKit
...
+        /// The base font used throughout the selector. Weights are derived automatically.
+        /// Default is `.systemFont(ofSize: 14, weight: .medium)`.
+        public var font: UIFont = .systemFont(ofSize: 14, weight: .medium)
+
+        /// Multiplier applied to all font sizes. For example, `1.2` makes text 20% larger.
+        /// Must be greater than 0. Default is `1.0`.
+        public var sizeScaleFactor: CGFloat = 1.0 {
+            didSet {
+                if sizeScaleFactor <= 0.0 {
+                    assertionFailure("sizeScaleFactor must be a value greater than zero")
```
gbirch-stripe, line 51: Interesting pattern. I think usually we allow users to set anything and then throw errors on load. Any reason we're not doing that here? This approach seems fine as well but I'm curious.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+CurrencySelectorViewAppearance.swift
```diff
@@ -12,62 +12,85 @@ import UIKit
...
+        public var selectedText: UIColor = .label
+
+        /// Text color for caption text. Default is `.secondaryLabel`.
+        /// Alpha values below 0.5 are clamped to 0.5 to keep regulatory text legible.
+        public var textSecondary: UIColor = .secondaryLabel {
+            didSet {
+                // Re-assignment in didSet is safe for value types (no recursion).
+                if textSecondary.cgColor.alpha < 0.5 {
+                    textSecondary = textSecondary.withAlphaComponent(0.5)
+                }
```
gbirch-stripe, line 72: Should we also do an `assertionFailure` here like for size factor so that users notice that they're setting an invalid value and not getting the behavior they asked for?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/CurrencySelectorUtilities.swift
```diff
@@ -86,20 +95,43 @@ enum CurrencySelectorUtilities {
...
             formatter.minimumFractionDigits = 2
             formatter.maximumFractionDigits = 4
-            formattedRate = formatter.string(from: NSNumber(value: rateDouble)) ?? meta.exchangeRate
+            formattedRate = formatter.string(from: NSNumber(value: inverse)) ?? meta.exchangeRate
         } else {
             formattedRate = meta.exchangeRate
         }
 
-        return "1 \(sellCurrency) = \(formattedRate) \(buyCurrency)"
+        if meta.conversionMarkupBps > 0 {
```
gbirch-stripe, line 112: Surprised that this logic is client-side instead of the server. Is there a reason?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/TwoOptionSelectorView.swift
```diff
@@ -75,24 +101,37 @@ final class TwoOptionSelectorView: UIView {
...
+    private func trackAndPillCornerRadii() -> (track: CGFloat, pill: CGFloat) {
+        let h = appearance.height
+        let bw = appearance.borderWidth
+        let maxTrack = max((h - bw) / 2, 0)
+        let track = min(appearance.cornerRadius, maxTrack)
+        let innerH = h - 2 * trackPadding
+        let maxPill = innerH / 2
+        let pill = max(0, min(track - trackPadding, maxPill))
+        return (track, pill)
+    }
```
gbirch-stripe, line 113: Why can we no longer use the `applyCornerRadiusOrConfiguration` radius?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+CurrencySelectorViewAppearance.swift
porter-stripe: Good question! This follows the same pattern used in PaymentSheetAppearance.swift. Properties like sizeScaleFactor, height, and verticalModeRowPadding all validate in didSet with assertionFailure. Kept it consistent with that existing convention.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+CurrencySelectorViewAppearance.swift
porter-stripe: This is intentionally different from the sizeScaleFactor case. With sizeScaleFactor, the value is simply invalid (zero/negative would break layout). Here, we're applying a minimum floor for accessibility/readability reasons, not rejecting an invalid value. A user might reasonably set alpha to 0.3 for their primary text and we'd still want the secondary text to be legible. It's more of a "we'll respect your choice but ensure a minimum contrast" behavior rather than "you set something wrong." Open to other opinions though!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/CurrencySelectorUtilities.swift
porter-stripe: Yeah this is client-side for now, we'll move this to the server possibly eventually but web currently has this on the client as well.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/TwoOptionSelectorView.swift
porter-stripe: applyCornerRadiusOrConfiguration sets one radius on one layer, but here we need two coordinated radii. The pill's radius is derived from the track's radius minus the padding between them, and both need clamping so they don't exceed their respective heights.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/AdaptivePricingFlagImageManager.swift
porter-stripe: Added!

gbirch-stripe (approved):

PR #6462, opened 2026-05-15
[link] Update `LinkBrand` from consumer login and logout

sophiah2-stripe:
## Summary

This updates `PaymentSheet` to prefer the active consumer’s `link_brand` over the initial `ElementsSession `brand once a Link consumer is identified.

Expected behavior:
  - Logged out: UI uses explicit client override if present, otherwise `ElementsSession` brand
  - Logged in: UI prefers the active consumer’s `link_brand`
  - Logged out again: UI falls back to the pre-login brand source

`PaymentSheetLinkAccount` will determine if a verified consumer is logged in, and `PaymentElementConfiguration.resolvedLinkBrand(...)` uses the fallback logic of config override (this is only used within the playground app), elements session, then default `.link`.

## Motivation

## Testing

Added tests.

To further verify, build the example app and use the new playground toggle for forcing consumer branding to check for regressions.


Example/PaymentSheet Example/PaymentSheet Example/PlaygroundController.swift
sophiah2-stripe, line 1: Branding changes should only occur after user is verified for this new flag.

StripeCore/StripeCore/Source/Connections Bindings/LinkBrand.swift
sophiah2-stripe, line 1: Branding logic is not needed for the checkout URL

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Link/LinkURLGeneratorTests.swift
sophiah2-stripe, line 1: Removed logic to switch branded URLs

StripePaymentSheet/StripePaymentSheet/Source/Helpers/PaymentSheetLinkAccount.swift
```diff
@@ -534,6 +543,7 @@ struct LinkPMDisplayDetails {
     }
 
     func logout() {
+        LinkAccountContext.shared.account = nil
```
josephroque-stripe, line 546: Was this just incorrectly never cleared on logout previously?

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/ConsumerSession+LookupResponse.swift
```diff
@@ -36,20 +36,27 @@ extension ConsumerSession {
         }
 
         let responseType: ResponseType
+        let rawLinkBrand: String?
```
josephroque-stripe: Do we need to keep the `rawLinkBrand`? Doesn't seem like we use it anywhere, outside this class we're only concerned about `LinkBrand` and if it's `unparsable`

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/STPAPIClient+Link.swift
```diff
@@ -726,11 +732,16 @@ extension STPAPIClient {
...
             consumerPublishableKey: consumerPublishableKey,
-            completion: completion
+            completion: { (result: Result<EmptyResponse, Error>) in
+                completion(result)
+            }
         )
     }
 }
 
+private extension STPAPIClient {
```
josephroque-stripe: unused?

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/STPAPIClient+Link.swift
```diff
@@ -395,7 +399,9 @@ extension STPAPIClient {
             endpoint: endpoint,
             additionalHeaders: authorizationHeader(),
             parameters: parameters,
-            completion: completion
+            completion: { result in
+                completion(result)
+            }
```
josephroque-stripe: unused?

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/STPAPIClient+Link.swift
```diff
@@ -726,11 +732,16 @@ extension STPAPIClient {
             resource: endpoint,
             parameters: parameters,
             consumerPublishableKey: consumerPublishableKey,
-            completion: completion
+            completion: { (result: Result<EmptyResponse, Error>) in
+                completion(result)
+            }
```
josephroque-stripe, line 735: unused?

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Link/STPAPIClient+Link.swift
```diff
@@ -794,6 +805,12 @@ private extension STPAPIClient {
 
     struct SessionResponse: Decodable {
         let consumerSession: ConsumerSession
+        let linkBrand: LinkBrand?
+
+        private enum CodingKeys: String, CodingKey {
+            case consumerSession = "consumer_session"
```
josephroque-stripe: Does it make sense here that the `CodingKeys` changed 🤔 wouldn't it still be expecting `consumerSession`, not `consumer_session`

josephroque-stripe: Ah, we use a custom decoder that ignores camelcase/snakecase. I'm guessing this works fine if we don't define these overrides at all, can probably remove `CodingKeys` in this case entirely.

Tested, and seems to be the case that we don't need to override this, but can you double check? We should be able to remove the explicit `linkBrand = "link_brand"` override in `ConsumerSession+LookupResponse` as well, just leave `case linkBrand`

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
```diff
@@ -220,7 +220,7 @@ extension PaymentSheet {
                             sublabel: confirmParams.paymentSheetSublabel
                         )
                     } else {
-                        labels = Labels(label: option.brand.displayName, sublabel: option.displayPaymentSheetSubLabel())
+                        labels = Labels(label: linkBrand.displayName, sublabel: option.displayPaymentSheetSubLabel(brand: linkBrand))
```
josephroque-stripe, line 223: Clarifying my understanding: `PaymentOption` has a brand associated with it when it's created, but the `PaymentSheetFlowController` will as well, we want to always use the brand from the `PaymentSheetFlowController`? In what cases would the underlying `PaymentOption` need to define its own `linkBrand`?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
```diff
@@ -128,7 +128,7 @@ class PaymentSheetFormFactory {
         let linkBrand: LinkBrand = {
             switch configuration {
             case .paymentElement(let configuration, _):
-                return configuration.resolvedLinkBrand(elementsSession: elementsSession)
+                return configuration.resolvedLinkBrand(elementsSession: elementsSession, linkAccount: LinkAccountContext.shared.account)
```
josephroque-stripe: We already have a `linkAccount` passed in here, should we have used that one?

StripePaymentSheet/StripePaymentSheet/Source/Helpers/PaymentSheetLinkAccount.swift
sophiah2-stripe: Yea, I'm assuming a UI refresh wasn't needed until we needed the branding to propagate immediately after logout in this case.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
sophiah2-stripe: `PaymentSheetFlowController` has a `.paymentOptionDisplayData`, and `PaymentOption` passes `linkBrand` to `PaymentOption+Images` for brand-aware UI logic. Since the two touch on different surfaces, they each need to contain the brand logic.

josephroque-stripe (dismissed):
Tested and looks good to me, once questions are all addressed!

josephroque-stripe (approved):

PR #6464, opened 2026-05-15
Remove refresh() public API from Checkout

porter-stripe:
## Summary
- Removes the `Checkout.refresh` API in favor of the `Checkout.runServerUpdate` API.

## Motivation
- CheckoutSessions

## Testing
N/A

## Changelog
N/A


CHANGELOG.md
```diff
@@ -1,5 +1,5 @@
 The next release's version bump will so far be:
-PATCH
+MINOR
```
porter-stripe, line 2: This is to prevent React Native which matches patch versions from breaking

wooj-stripe (approved):

PR #6474, opened 2026-05-16
[iOS 26] Set card form postal code test country directly

davidme-stripe:
We can't swizzle NSLocale anymore in iOS 26, just set the country code to test this behavior instead.


Stripe/StripeiOSTests/STPCardFormViewTests.swift
gbirch-stripe, line 1: Why is this change being made on only some of the tests but not others?

davidme-stripe: Sorry Codex has gone rogue, these should all be drafts!

PR #6489, opened 2026-05-18
[link] Refactor FC link brand resolution

sophiah2-stripe:
## Summary

  - Use `synchronize.manifest.link_brand` as the base FC-native brand source, and override with a logged in or signed up consumer `link_brand` when available (passed by `confirm_verification` and `sign_up` responses)
  - Remove stale ElementsSession-based brand fallback from FC entry points
  - Update naming for Financial Connections playground app toggle
  - Rename ambiguous `brand` to `linkBrand`

## Motivation

## Testing

Modified tests


StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/Models/ConsumerSession/ConsumerSessionModels.swift
```diff
@@ -53,13 +53,49 @@ struct LinkSignUpResponse: Decodable {
...
+struct ConsumerSessionStartVerificationResponse: Decodable {
     let consumerSession: ConsumerSessionData
 }
+
+struct ConsumerSessionConfirmVerificationResponse: Decodable {
+    let consumerSession: ConsumerSessionData
+    let linkBrand: LinkBrand
+}
+
+extension LinkSignUpResponse {
```
jkelle-stripe, line 73: Why did we need to add the custom decoding? Don't doubt there was some reason we needed, just isn't clear to me.

StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPI.swift
```diff
@@ -8,7 +8,7 @@
 import Foundation
 @_spi(STP) import StripeCore
 
-protocol FinancialConnectionsAPI {
+protocol FinancialConnectionsAPI: AnyObject {
```
jkelle-stripe, line 11: [nit] I don't see a reason to have this restriction (requiring all conformers to be a class). If we don't have one, then I think we can get rid of this and make all of the call sites `var` instead of `let`.

StripePayments/StripePayments/Source/Helpers/STPBankAccountCollector.swift
```diff
@@ -305,7 +305,7 @@ public class STPBankAccountCollector: NSObject {
                 existingConsumer: nil,
                 style: self.style.asFinancialConnectionsConfigurationStyle,
                 elementsSessionContext: elementsSessionContext,
-                brand: elementsSessionContext?.linkSettings?.brand,
+                linkBrand: nil,
```
jkelle-stripe, line 308: Why the change here?

jkelle-stripe (approved):
Non-blocking questions/comments. Can address in a follow up

StripePayments/StripePayments/Source/Helpers/STPBankAccountCollector.swift
sophiah2-stripe: We want to set this one to `nil` to avoid any override from `ElementsSession` since this is part of an FC flow within the `PaymentSheet`; it would make more sense to derive the brand value from backend calls within the FC flows.

StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/Models/ConsumerSession/ConsumerSessionModels.swift
sophiah2-stripe: This was needed before I decided to split up the response handling for `confirm_verification` (returns `link_brand`) and `start_verification` (shouldn't return `link_brand`). Will clean this up in a follow up PR.

StripeFinancialConnections/StripeFinancialConnections/Source/API Bindings/FinancialConnectionsAPI.swift
sophiah2-stripe: Will also address this one in a follow up after release

Issue #6506, opened 2026-05-20
[BUG] AddressViewController does not collect City information for Japan addresses


bjhomer:
## Summary
When the user selects "Japan" as the country in an AddressViewController, the following fields are displayed:

English Label | Japanese Label | Notes
--|--|--
Postal Code | 郵便番号 |
Prefecture  | 都道府県 | roughly equivalent to "State" in the USA
Address Line 1 | 住所 (1 行目) |
Address Line 2 (optional) | 住所 (2 行目) (オプション) | 

**There is no field for the user to enter their city / ward / village information**. This would normally be present with a label like 市区町村. Without this field, many shipping orders cannot be completed.

### Screenshots
English | Japanese
---|---
<img width="379" height="448" alt="Image" src="https://github.com/user-attachments/assets/a5c9fbab-3428-407a-bf67-3d27384f1ee2" /> | <img width="385" height="461" alt="Image" src="https://github.com/user-attachments/assets/54de9159-36dc-4035-a0e4-524edc34b9d2" />

## Code to reproduce
```swift
let addressConfiguration = AddressViewController.Configuration(
  additionalFields: .init(phone: .required),
  allowedCountries: shippingCountries, // includes Japan
  title: "Shipping Address"
)

guard let addressViewController = AddressViewController(configuration: addressConfiguration, delegate: self) else { return }

let navigationController = UINavigationController(rootViewController: addressViewController)
present(navigationController, animated: true)
```

## iOS version
iOS 26.5 (but also present earlier)

## Installation method
Swift Package Manager

## SDK version
25.15.0, though also observed in 24.4.1

porter-stripe:
@bjhomer thank you for submitting this report! I've submitted a fix and it will be included in our next release.

porter-stripe:
This was fixed in 25.16.0!

bjhomer:
Thank you!

PR #6507, opened 2026-05-20
Fix Japan address form missing city field

porter-stripe:
## Summary
Japan's entry in localized_address_data.json was missing the %C (city) field token in its format string. This caused AddressSpec to produce a fieldOrdering without .city, so the city field was never rendered for Japanese addresses.

Changes:
- fmt: Add %C between %S and %n%A so city renders after prefecture
- lfmt: Add %C in the Latin format string
- require: Change ASZ to ACSZ so city is validated as required
- Add locality_name_type: city for proper field labeling

Simulator Screenshot - Snapshot - 2026-05-20 at 09 31 04
Simulator Screenshot - Snapshot - 2026-05-20 at 09 30 28

## Motivation
- https://github.com/stripe/stripe-ios/issues/6506

## Testing
- New unit test
- Manual

## Changelog
See diff

### DIFF
--- a/StripeUICore/StripeUICoreTests/Unit/Elements/AddressSectionElementTest.swift
+++ b/StripeUICore/StripeUICoreTests/Unit/Elements/AddressSectionElementTest.swift
@@ -268,6 +268,59 @@ class AddressSectionElementTest: XCTestCase {
         XCTAssertEqual(sut.addressDetails.address.state, "CA")
     }
 
+    func testJapanAddressIncludesCityField() {
+        // Verify that Japan's address spec includes a city field and that it's required.
+        // This is a regression test for a bug where the JP entry in localized_address_data.json
+        // was missing the %C token, causing the city field to not render.
+        let specProvider = AddressSpecProvider()
+        specProvider.addressSpecs = [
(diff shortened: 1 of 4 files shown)


StripeUICore/StripeUICoreTests/Unit/Elements/AddressSectionElementTest.swift
```diff
@@ -268,6 +268,59 @@ class AddressSectionElementTest: XCTestCase {
...
+        )
+        let section = sut.addressSection
+
+        // Verify that the city field exists
+        XCTAssertNotNil(sut.city, "Japan address should include a city field")
+
+        // Verify the field ordering contains .city
+        let spec = specProvider.addressSpecs["JP"]!
+        XCTAssertTrue(spec.fieldOrdering.contains(.city), "JP field ordering should contain .city")
+        XCTAssertTrue(spec.requiredFields.contains(.city), "JP required fields should contain .city")
```
wooj-stripe, line 307: The test only checks that it exists, but maybe we should check that it comes after state?

wooj-stripe (approved):

PR #6510, opened 2026-05-20
CryptoOnramp SDK: Makes LinkAppearance.PrimaryButtonConfiguration Optionality Match Android

mliberatore:
## Summary
Previously, unspecified values on `init` made `PrimaryButtonConfiguration`'s properties default to `.zero`. This led to a mismatch in mapping behavior since the properties couldn’t be `nil`, and individually apply corner radius or height, falling back to Link  defaults for the other.

## Motivation
This comment: https://github.com/stripe/stripe-react-native/pull/2404#discussion_r3134337573

## Testing
Ensured that when applied to the react native repo (https://github.com/stripe/stripe-react-native/pull/2475), we were able to set each value independently while relying on a default for the other. I used intentionally obvious values to ensure they were being applied, and differed from sensible defaults.

| Corner radius set only | Height set only | Both set | Neither set (both defaults) |
| --- | --- | --- | --- |
|  |  |  |  |

## Changelog

✅

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ PATCH
 
 ## X.Y.Z - changes pending release
 ### PaymentSheet
+* [Fixed] Fixed `LinkAppearance.PrimaryButtonConfiguration` styling so unspecified custom height and corner radius values use Link defaults.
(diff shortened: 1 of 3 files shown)


jeanregisser (approved):
Thanks for following up on this! 🙌

PR #6512, opened 2026-05-21
Autocomplete using endpoint

joyceqin-stripe:
## Summary

Use autocomplete endpoints for address autocomplete. Debounce and don't start making API calls until at least 2 characters have been typed.

Added playground switches, make API requests, add Google attribution
## Motivation

Address autocomplete project
## Testing

Snapshots, unit tests, network tests
## Changelog

N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/AutoComplete/AutoCompleteViewController.swift
```diff
@@ -95,6 +104,37 @@ class AutoCompleteViewController: UIViewController {
...
+        let container = UIView()
+        container.addSubview(imageView)
+
+        var constraints = [
+            imageView.leadingAnchor.constraint(equalTo: container.leadingAnchor, constant: tableView.layoutMargins.left + indendationWidth),
+            imageView.heightAnchor.constraint(equalToConstant: UIFont.preferredFont(forTextStyle: .footnote).lineHeight),
+            imageView.widthAnchor.constraint(equalTo: imageView.heightAnchor, multiplier: image.size.width / image.size.height),
+            imageView.topAnchor.constraint(equalTo: container.topAnchor, constant: tableView.layoutMargins.top),
+        ]
+        if #available(iOS 26, *) {
```
davidme-stripe: Why do we only add this separator on iOS 26?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/AutoComplete/AutoCompleteViewController.swift
```diff
@@ -95,6 +104,37 @@ class AutoCompleteViewController: UIViewController {
...
+            container.addSubview(separator)
+            constraints += [
+                separator.topAnchor.constraint(equalTo: container.topAnchor),
+                separator.leadingAnchor.constraint(equalTo: container.leadingAnchor, constant: tableView.layoutMargins.left),
+                separator.trailingAnchor.constraint(equalTo: container.trailingAnchor, constant: -tableView.layoutMargins.right),
+                separator.heightAnchor.constraint(equalToConstant: 1.0),
+            ]
+        }
+        NSLayoutConstraint.activate(constraints)
+        container.frame.size.height = 64
```
davidme-stripe: Might be missing something, but I think this should be defined as an autolayout constraint with a fixed height instead of setting the frame directly.

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Address/AutocompleteResponse.swift
```diff
@@ -0,0 +1,154 @@
...
+//  StripePaymentSheet
+//
+//  Created by Joyce Qin on 4/20/26.
+//
+
+import Foundation
+
+private extension String {
+    /// Converts a half-open code-point range [start, end) to an NSRange of UTF-16 code units.
+    func utf16Range(fromCodePointOffsets start: Int, to end: Int) -> NSRange? {
```
davidme-stripe, line 14: I'm a little confused by this, could we add more explanation? I thought the Google Maps API would return UTF-8, and Swift uses UTF-8 internally these days afaik, so should we be operating on a UTF8View?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/AutoComplete/AutoCompleteViewController.swift
joyceqin-stripe: Okay I used to think that maybe iOS 26 didn't have an ending separator because the snapshot tests came out like this, but after further inspection it appears that I just had different constraints that were messing it up/covering it. Yay, I get to remove this block!
image

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/AutoComplete/AutoCompleteViewController.swift
joyceqin-stripe: Yep, you're right— it was once again because I was using different constraints before (I was using a centerY anchor instead) and it was looking weird. I can remove this now

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Address/AutocompleteResponse.swift
joyceqin-stripe: The Google API returns zero based Unicode character offsets, which _could_ theoretically include non-BMP characters (though I think it's unlikely), and highlightSearchString uses NSAttributedString, which is UTF-16, so the idea is to convert to UTF-16 to get the right ranges

davidme-stripe: Ah, right, it's NSAttributedString. Let's just make sure we have some tests around this that check that we can go from a string representation from the service to an NSAttributedString with the proper range? Maybe with some emoji and various language blocks

joyceqin-stripe: I updated the comment and I have a test testMatchRangeWithNonBMPCharacterBefore with an emoji to test the offset translates to the right range

davidme-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/Address/AutocompleteResponse.swift
davidme-stripe: Cool cool, thanks!

PR #6514, opened 2026-05-22
[link] Modify accessibility label for `LinkBrand`

sophiah2-stripe:
## Summary

Modify the accessibility label to correct the pronunciation of "Onelink".

## Motivation

https://jira.corp.stripe.com/browse/LINK_MOBILE-870

## Testing

Tested on physical device and Accessibility Inspector

### DIFF
--- a/StripeCore/StripeCore/Source/Connections Bindings/LinkBrand.swift
+++ b/StripeCore/StripeCore/Source/Connections Bindings/LinkBrand.swift
@@ -27,7 +27,7 @@ import Foundation
         case .link, .unparsable:
             return "Link"
         case .onelink:
-            return "one link"
+            return "One-link"
         }
     }
 
--- a/StripeCore/StripeCore/Source/Connections Bindings/LinkBrand.swift
+++ b/StripeCore/StripeCore/Source/Connections Bindings/LinkBrand.swift
@@ -37,7 +37,7 @@ import Foundation
         }
         return text
             .replacingOccurrences(of: displayName, with: accessibilityDisplayName)
-            .replacingOccurrences(of: rawValue, with: accessibilityDisplayName)
+            .replacingOccurrences(of: displayName.lowercased(), with: accessibilityDisplayName)
     }
 
     @_spi(STP) public var websiteURL: URL {
--- a/StripeCore/StripeCore/Source/Connections Bindings/LinkBrand.swift
+++ b/StripeCore/StripeCore/Source/Connections Bindings/LinkBrand.swift
@@ -22,6 +22,24 @@ import Foundation
         }
     }
 
+    @_spi(STP) public var accessibilityDisplayName: String {
+        switch self {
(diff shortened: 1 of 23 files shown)


StripeCore/StripeCore/Source/Connections Bindings/LinkBrand.swift
```diff
@@ -22,6 +22,24 @@ import Foundation
...
+        }
+    }
+
+    @_spi(STP) public func accessibilityText(from text: String) -> String {
+        guard self == .onelink else {
+            return text
+        }
+        return text
+            .replacingOccurrences(of: displayName, with: accessibilityDisplayName)
+            .replacingOccurrences(of: rawValue, with: accessibilityDisplayName)
```
jkelle-stripe: Is the goal here is upper and lower case detection? If so, I think we should make this explicit by calling displayName.lowercased() in place of rawValue. We may end up changing the rawValue and not know that we're breaking accessibility. Unlikely, but just a good practice to limit indirection IMO

StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/PaneLayoutView/PaneLayoutView+Extensions.swift
```diff
@@ -11,12 +11,21 @@ import SwiftUI
...
+            self.accessibilityText = accessibilityText ?? text
+        }
+    }
 
     @available(iOSApplicationExtension, unavailable)
     static func createContentView(
         iconView: UIView?,
-        title: String?,
-        subtitle: String?,
+        accessibleTitle: AccessibleText?,
```
jkelle-stripe: I like updating the type here, but on reading the `accessibleTitle` property label, I first thought "how do I add a `title` and `subtitle` along with the `accessibleTitle`"? I think we should keep the `title` and `subtitle` naming for the properties to convey that they will set the rendered title on screen and let the types describe that you include both the text and the accessibility text.

jkelle-stripe (changes requested):

StripeFinancialConnections/StripeFinancialConnections/Source/Native/Shared/PaneLayoutView/PaneLayoutView+Extensions.swift
sophiah2-stripe: The issue here is with call sites that pass in `nil` values, since we have two `createContentView` function definitions, one using `String?` and one with `AccessibleText?`. I pushed some changes for this!

jkelle-stripe: Resolved in Slack. Let's leave these as `accessibleTitle`

jkelle-stripe (approved):

PR #6516, opened 2026-05-22
Checkout Sessions — US Bank Account / Financial Connections support

gbirch-stripe:
## Summary
Add US bank account / FC support for checkout sessions by using the deferred intent link account session endpoint. 
See decision here: https://docs.google.com/document/d/15b-rFbJZhLVS7HVBk99couT93kUcBk64WG11nL9kVFM/edit?tab=t.0

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-4480

## Testing
Added new STPBankAccountCollectorTests test


porter-stripe:
@gbirch-stripe Can you add a PR description? The summary is blank.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentMethodFormViewController.swift
```diff
@@ -550,10 +550,33 @@ extension PaymentMethodFormViewController {
...
+                stpAssertionFailure("subscription and unknown modes not supported")
+                amount = nil
+            }
+            client.collectBankAccountForDeferredIntentOrCheckoutSession(
+                sessionId: elementsSession.sessionID,
+                returnURL: configuration.returnURL,
+                onEvent: nil,
+                amount: amount,
+                currency: checkoutSession.currency,
+                onBehalfOf: nil,
```
porter-stripe, line 566: Is this correct to always pass nil for onBehalfOf?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentMethodFormViewController.swift
```diff
@@ -653,10 +676,33 @@ extension PaymentMethodFormViewController {
                 additionalParameters: additionalParameters,
                 elementsSessionContext: elementsSessionContext,
                 from: viewController,
+                intentType: .deferred,
+                financialConnectionsCompletion: financialConnectionsCompletion
+            )
+        case .checkoutSession(let checkoutSession):
```
porter-stripe, line 675: This seems nearly identical to the code block above starting on 556, can we DRY it?

Stripe/StripeiOSTests/STPBankAccountCollectorTests.swift
```diff
@@ -202,7 +202,36 @@ final class STPBankAccountCollectorTests: APIStubbedTestCase {
...
+                XCTFail("Result was not completed")
+            }
+            XCTAssertEqual(linkAccountSession?.stripeID, "las_789")
+            exp.fulfill()
+        }
+        waitForExpectations(timeout: 2.0)
+    }
+
+    func testCollectBankAccountForCheckoutSessionSucceeds() {
+        stubCreateLinkAccountSessionForDeferredIntent()
```
porter-stripe, line 220: Why are we stubbing an account session for deferred intent in a checkout session test?

Stripe/StripeiOSTests/STPBankAccountCollectorTests.swift
```diff
@@ -193,7 +193,7 @@ final class STPBankAccountCollectorTests: APIStubbedTestCase {
         let collector = STPBankAccountCollector(apiClient: stubbedAPIClient())
         let exp = expectation(description: "completion")
 
-        collector.collectBankAccountForDeferredIntent(
+        collector.collectBankAccountForDeferredIntentOrCheckoutSession(
```
porter-stripe, line 196: nit: Should this really be `collectBankAccountForDeferredIntentOrCheckoutSession`? The "or" is a bit smelly, why not two functions or something?

porter-stripe (commented):
Can you also include a recording for these payment flows showing they work now? I also think we may be missing test coverage for the instant debits path.

Stripe/StripeiOSTests/STPBankAccountCollectorTests.swift
gbirch-stripe: I originally did that but decided against it because the functions are totally identical except for a different `intent_type` param in the analytics logs, and so it's a bunch of code duplication. Open to reverting it back, so I'm curious to hear your thoughts on the advantage

Stripe/StripeiOSTests/STPBankAccountCollectorTests.swift
gbirch-stripe: Checkout sessions uses the deferred_intent API endpoint (`link_account_sessions_for_deferred_payment`), and so that's the one need to stub

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentMethodFormViewController.swift
gbirch-stripe: Good idea, we actually already have an `expectedAmount()` function on the session that does this. It threw to guard against an amount-less payment mode session but I think we can change this to an assert since it's a scenario that has no reason to ever occur

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentMethodFormViewController.swift
gbirch-stripe: Digging into this more. Will update PR after

porter-stripe (approved):

PR #6519, opened 2026-05-26
CryptoOnramp SDK: API Error Refactor

mliberatore:
## Summary
Refactors API errors to adopt what's being proposed in [Proposal: Errors are a UI](https://docs.google.com/document/d/1q2ifBLNzr5AEITpQOkFIq58suY2Lo2vFP0eElQ5JlYE/edit?tab=t.0#heading=h.oxy9g9klck2y) for attestation errors and any other API errors that are not currently tracked. 

Existing concrete error types within CryptoOnramp are untouched. See the "note" below.

This PR:
* Added protocol `StripeCryptoOnrampError` for rich Crypto Onramp errors and sub-protocol `StripeCryptoOnrampAPIError` for API-backed Crypto Onramp errors, with separate user-facing and developer-facing messages, stable error codes, documentation URLs, and underlying errors.
    * Added `AppAttestationAPIError`, `UncategorizedAPIError` as the first concrete implementations of `StripeCryptoOnrampAPIError`, and `APIErrorContext` as a concrete struct to store common properties for all `StripeCryptoOnrampAPIError`s to reduce duplication.
    * `AppAttestationAPIError` is the main goal of this PR, where `UncategorizedAPIError` catches any API errors currently not routed to other existing error types.

> [!NOTE]
> We have additional error types that `CryptoOnrampCoordiantor` currently can `throw`. These are not folded in yet to the new protocols, and we'll either consider making the existing errors conform to the new protocols, or introduce new concrete tpyes that conform to the protocols and move away from the `enum` based error types:
> - `CryptoOnrampCoordinator.Error`
> - `CheckoutError`
> - `CryptoOnrampAIError` (primarily short circuits from checks made before making a request)
> - `ApplePayPaymentStatus.Error` (has an unlikely fallback error case)

## Motivation

[Proposal: Errors are a UI](https://docs.google.com/document/d/1q2ifBLNzr5AEITpQOkFIq58suY2Lo2vFP0eElQ5JlYE/edit?tab=t.0#heading=h.oxy9g9klck2y).

## Testing

Tested by ensuring I’d trigger an attestation error by using a dummy app id (reverted in [57b2336](https://github.com/stripe/stripe-ios/pull/6519/commits/57b2336a9f4a1e09af83926869e5143112e974aa)). Here, `localizedDescription` surfaces the following error message:
Simulator Screenshot - iPhone 17 - 2026-05-26 at 12 28 47

And in the debugger, a lot more developer-specific detail can be retrieved:
CleanShot 2026-05-26 at 12 27 32

## Changelog

Updated.


StripeCryptoOnramp/StripeCryptoOnramp/Resources/Localizations/en.lproj/Localizable.strings
```diff
@@ -1,3 +1,27 @@
+/* Error message shown when the user tries to create a Link account for an email that already has one */
+"A Link account already exists for this email. Log in instead." = "A Link account already exists for this email. Log in instead.";
```
mliberatore: To make an API contract guarantee, I‘ve added user-facing descriptions for the few existing `CryptoOnrampCoordinator.Error` cases as well. If these aren't desired at this time, just let me know and I‘ll fall back to the underlying error's `localizedDescription`. Today, that's a generic message.

StripeCryptoOnramp/StripeCryptoOnramp/Resources/Localizations/en.lproj/Localizable.strings
```diff
@@ -1,3 +1,27 @@
...
+"Invalid phone format. Please use E.164 format (e.g., +12125551234)." = "Invalid phone format. Please use E.164 format (e.g., +12125551234).";
+
+/* Error message shown when there is no usable payment method selected */
+"No payment method is ready to use. Please collect a payment method and try again." = "No payment method is ready to use. Please collect a payment method and try again.";
+
+/* Error message shown when the SDK is missing required information from the server */
+"Required information was missing. Please try again later." = "Required information was missing. Please try again later.";
+
+/* Error message shown when app attestation fails */
+"This app couldn't be verified due to an attestation error. Please try again later or contact the developer if the issue persists." = "This app couldn't be verified due to an attestation error. Please try again later or contact the developer if the issue persists.";
```
mliberatore, line 5: This is the attestation error user-facing message. Note that these localized strings are drafts that I created. Please let me know the process for getting user-facing copy approved or suggested by the appropriate party.

jeanregisser (commented):
Thanks for pushing this forward. I think this is a strong first step toward the "Errors are a UI" direction.

The parts I like most are that this moves beyond simply passing through backend/API errors and starts giving the SDK ownership over the developer and user-facing experience: preserving the raw Stripe error, adding SDK-local context like operation/mode/SDK version/request ID, separating user-facing copy from developer diagnostics, and making attestation errors feel guided instead of opaque.

The main thing I am wondering about is the shape of the abstraction before this pattern spreads further. I can see why this PR keeps the surface centered on `CryptoOnrampCoordinator.Error` for the first iteration: it is conservative, scoped, and avoids introducing a bigger abstraction before we have proved the shape. The tradeoff is that the common rich-error contract is implicit rather than explicit.

Do you think it is worth introducing a small shared protocol now, so future cases can plug into the same shape without making callers switch over coordinator-specific cases?

The proposal is intentionally inspired by Viem. I do not think we should copy Viem's class hierarchy into Swift, but I do think there is a useful architectural lesson there: rich errors have a shared contract. They have a stable code, a developer-oriented rendered message, user-safe copy, docs/cause metadata where available, and concrete error types can add source-specific fields. In Swift, that probably wants to be protocol-oriented.

For example, not necessarily exact API or final name:

```swift
@_spi(CryptoOnrampAlpha)
public protocol StripeCryptoOnrampError: Error, LocalizedError, CustomDebugStringConvertible {
    var code: String { get }
    var userMessage: String { get }
    var developerMessage: String { get }
    var docURL: URL? { get }
    var underlyingError: Error? { get }
}
```

With defaults along these lines:

```swift
public extension StripeCryptoOnrampError {
    var errorDescription: String? {
        userMessage
    }

    var debugDescription: String {
        developerMessage
    }
}
```

Then an API-specific refinement could carry API v2 fields without forcing every SDK/local error to pretend it came from an API response:

```swift
@_spi(CryptoOnrampAlpha)
public protocol StripeCryptoOnrampAPIError: StripeCryptoOnrampError {
    var type: String? { get }
    var requestID: String? { get }
    var apiMessage: String? { get }
    var apiUserMessage: String? { get }
}
```

That would keep us aligned with API v2 terminology (`message`, `user_message`, `code`, `type`) while still respecting Swift conventions. In particular, I think `localizedDescription` / `errorDescription` should remain user-safe, because iOS developers commonly put that into UI. The richer developer diagnostic should be explicit via `developerMessage` / `debugDescription`.

The current implementation already moves in this direction by making `errorDescription` return the user-facing copy and `debugDescription` return the developer diagnostic. That also seems consistent with patterns elsewhere in the SDK, like `PaymentSheetError`. If we do add the protocol in this PR, I think that convention should live there rather than being repeated per concrete error type.

`CryptoOnrampCoordinator.Error` can still exist as the product-specific thrown enum, but callers would have a case-agnostic way to handle rich Crypto Onramp errors:

```swift
catch let error as StripeCryptoOnrampError {
    show(error.userMessage) // or map/override it in the host app's own UI
    logger.error(error.developerMessage)
}
```

This might also help with the follow-up work called out in the PR description. If `CheckoutError`, `CryptoOnrampAIError`, and `ApplePayPaymentStatus.Error` eventually fold into the same umbrella, they could conform to the base rich-error protocol without needing to pretend they are API-backed errors.

One concrete implication: for uncategorized API errors, I would avoid falling back from `userMessage` to raw `localizedDescription` / backend prose. Unknown errors should still fail safely for app users, while preserving raw API details in the developer/API-specific fields. That seems more consistent with the proposal's two-audience model.

One DevX note: I like that `logAndThrow` already prefers `developerDescription` for mapped coordinator errors. If we introduce a common protocol, I would generalize that logging path to check the protocol instead of the concrete coordinator enum. Throwing an error will not automatically print the developer diagnostic everywhere, so the explicit logging/debug surface matters.

Overall, I think this is going in the right direction. My main question is whether this PR should make the common rich-error contract explicit now, while this is still scoped to Crypto Onramp, so we do not accidentally bake the first implementation into a coordinator-specific shape that is harder to generalize later.

mliberatore:
@jeanregisser updated according to your feedback and opened for re-review. Note that one change I‘ve made to make the approach more clear is that I’m not adapting `CryptoOnrampCoordinator+Error` to the new error protocols yet. These error cases will (for now) live alongside the new error protocols with the only concrete error types conforming to the new format being `AppAttestationAPIError` and `UncategorizedAPIError`. Every other non-API concrete error currently thrown by the coordinator remains unchanged in this PR. This way, we can more methodically chip away at those errors, maybe remove the `enum` structure of them all together, and create new, individual, concrete conformers to the new protocols to migrate those.

Let me know your thoughts on that, as I can always revert and make `CryptoOnrampCoordinator+Error` conform instead.

jeanregisser (commented):
Thanks for the update. I like this direction!

One smaller formatting/design thought: I think the final developer-message rendering should probably live in a common rich-error renderer rather than on `APIErrorContext`. `APIErrorContext` can still own API/request-specific data, but the final layout of the developer message feels like part of the shared error UX.

This is mostly about preserving the readability target from the original proposal: the developer error should read like a polished diagnostic, not just a dump of fields. I think the concrete error should own the meaningful content, but the shared layer should own the final shape so every rich error has the same visual grammar.

For example, not necessarily exact API:

```swift
public protocol StripeCryptoOnrampError: Error, LocalizedError, CustomDebugStringConvertible {
    var code: String { get }
    var userMessage: String { get }
    var developerMessage: String { get }
    var docURL: URL? { get }
    var sdkVersion: String { get }
    var underlyingError: Error? { get }
}
```

Concrete errors would build the main developer body. API-backed errors can include a `Request Context` block in that body, omitting fields that are not set. A shared renderer would append footer-style metadata consistently.

The output shape I would personally like us to aim for is something like:

```text
App attestation failed: this app is not registered as a trusted application.

Request Context:
  operation: has_link_account
  app_id: com.example.app
  mode: test
  reason: app_not_registered
  request_id: req_123
  type: invalid_request_error

Code: link_failed_to_attest_request
Next step: Register this app's bundle ID or package name as a trusted application with Stripe, then retry the Onramp flow.
Docs: https://stripe.com/docs/...
SDK: stripe-ios@25.16.0
```

So conceptually:

```swift
StripeCryptoOnrampErrorRenderer.render(
    developerBody: developerBody,
    code: code,
    nextStep: nextStep,
    docURL: docURL,
    sdkVersion: sdkVersion
)
```

That keeps the current protocol small, but avoids each concrete error inventing its own formatting as we add more rich errors. It also mirrors the part of Viem that I think is most relevant here: centralizing the final error rendering while letting concrete errors own the useful content.

Side note: I would also like the RN side to be able to include its version in the SDK line eventually:

```text
SDK: stripe-ios@25.16.0, stripe-react-native@x.y.z
```

I am not sure we need to solve that in this PR, but it is worth keeping the renderer design wrapper-friendly if that does not add much complexity.

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/Errors/StripeCryptoOnrampAPIError.swift
```diff
@@ -0,0 +1,40 @@
...
+    var operation: String { get }
+
+    /// The bundle identifier for the app using the SDK, if one is available.
+    var appIdentifier: String? { get }
+
+    /// The Stripe mode associated with this error, if it can be determined.
+    var mode: String? { get }
+
+    /// The Stripe iOS SDK version.
+    var sdkVersion: String { get }
```
jeanregisser: I think this should be moved to `StripeCryptoOnrampError`.

jeanregisser (dismissed):
Awesome, thanks for working with me on these changes!

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/Errors/StripeCryptoOnrampAPIError.swift
mliberatore: Updated in [9255e12](https://github.com/stripe/stripe-ios/pull/6519/commits/9255e129787df39fef343d259258c802f2ac42a7).

jeanregisser (dismissed):

mliberatore:
> Side note: I would also like the RN side to be able to include its version in the SDK line eventually:
> 
> ```
> SDK: stripe-ios@25.16.0, stripe-react-native@x.y.z
> ```
> I am not sure we need to solve that in this PR, but it is worth keeping the renderer design wrapper-friendly if that does not add much complexity.

@jeanregisser makes sense that we'd want to do this. I’d like to make sure we don't have to majorly refactor anything here to pull that off. One solution that comes to mind is the following, but I don't think it matches your intention with this being part of the renderer design. Let me know your thoughts or if you would like to propose alternative suggestions:

We can make `sdkVersion` use a new type instead of a `String`. Something like:

```swift
struct SDKVersion {
    let name: String
    let version: String

    var debugDescription: String {
        "\(name)@\(version)"
    }
}
```

and we store it on `StripeCryptoOnrampError` as `var sdkVersion: SDKVersion { get }`.

We don’t store an array, as that’s going to be less extensible (we’d need every concrete error to store the property, and make it `{ get set }` rather than being able to provide it as a default protocol conformance).

Then, in the RN codebase ([WIP seen here](https://github.com/stripe/stripe-react-native/blob/affec24af7a15dc004775b18299ccc7c69b02c31/ios/OnrampErrors.swift#L67)), in our mapping logic, where we map `sdkVersion`, at least in iOS bindings, it looks like:

```swift
"sdkVersion": error.sdkVersion
```

We can instead make it something like:
```swift
"sdkVersion": "\(error.sdkVersion.developerDescription), \(reactNativeSDKVersion.developerDescription)
```
 where `reactNativeSDKVersion` is a private computed property or static instance declared somewhere around our mapping logic:

```swift
private static let reactNativeSDKVersion = SDKVersion(
    name: "stripe-react-native"
    version: WhateverPropertyAccessesTheRNVersionHere
)
```

Then the actual rendered output line when inspecting the developer message in RN would be:

```
SDK: stripe-ios@25.16.0, stripe-react-native@x.y.z
```

jeanregisser:
Thanks @mliberatore , I guess since the rendered developer message includes the SDK line, I wonder if a pragmatic choice would be to have a global variable / function that the RN SDK could call to set it's own extra data for this.
Otherwise we'll have mismatches between the property and the message.

What do you think?

mliberatore:
> Thanks @mliberatore , I guess since the rendered developer message includes the SDK line, I wonder if a pragmatic choice would be to have a global variable / function that the RN SDK could call to set it's own extra data for this. Otherwise we'll have mismatches between the property and the message.
> 
> What do you think?

I’m curious about "mismatches between the property and the message", but want to make sure we're on the same page. What I’m proposing is that the mapped error in RN, which is converted from our native `StripeCryptoOnrampError`, would have its `sdkVersion` property become e.g. `stripe-ios@25.16.0, stripe-react-native@x.y.z`. So if the developer inspected `sdkVersion` in react native code, the value would be `"stripe-ios@25.16.0, stripe-react-native@x.y.z"`. That is, in the translation layer between the native Swift type and the RN typescript type, we amend `sdkVersion` so that the RN type's `sdkVersion` has both values as a single string. The RN type wouldn't need to necessarily match the native version in separating the values into a `struct SDKVersion` like in swift. It could remain a string that has both sdk versions accounted for, comma-separated.

jeanregisser:
> I’m curious about "mismatches between the property and the message", but want to make sure we're on the same page.

The `developerMessage` includes the SDK text `SDK: stripe-ios@x.yz` and it is rendered on the native side.
Since the RN error object has both the `developerMessage` and `sdkVersion` props, I'd like that part to match.
If we just change the `sdkVersion` prop in the RN dictionary, it won't match the SDK text in the developer message, right?

mliberatore:
> The `developerMessage` includes the SDK text `SDK: stripe-ios@x.yz` and it is rendered on the native side.
> Since the RN error object has both the `developerMessage` and `sdkVersion` props, I'd like that part to match.
> If we just change the `sdkVersion` prop in the RN dictionary, it won't match the SDK text in the developer message, right?

@jeanregisser yes sorry, I see what you mean, you're right. I’ll think on this a bit and propose an alternative. I'd like to avoid adding a global if possible to solve this. We could pass something in during initial configuration, for example. I’ll try to think of solutions closer to the actual error implementation to see if I can work out something intuitive!

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/Errors/APIErrorContext.swift
```diff
@@ -0,0 +1,172 @@
...
+    public var apiUserMessage: String?
+
+    /// A URL to documentation for this error, if one is available.
+    public var docURL: URL?
+
+    /// The original error that was mapped to this error.
+    public var underlyingError: Swift.Error
+
+    /// SDK versions included in developer diagnostics, including Stripe iOS and any additional wrapper SDK versions.
+    public var sdkVersions: [SDKVersion]
```
mliberatore, line 115: Note: `sdkVersions` still lives on the root protocol (`StripeCryptoOnrampError`) so will be present when we have concrete types that _aren't_ API errors. This property just provides de-duplicative storage for API-specific errors which we have concrete implementations of now.

mliberatore:
The latest changes support additional sdk versions passed at configuration time for the onramp coordinator. With this, testing out the same attestation error as in the description of this PR, we now see on native iOS:

Slack 2026-06-04 15 31 50

and React Native (when passing in the appropriate version in our binding):

Google Chrome 2026-06-04 15 26 22

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/Errors/StripeCryptoOnrampError.swift
```diff
@@ -0,0 +1,81 @@
...
+@_spi(CryptoOnrampAlpha)
+public struct SDKVersion: CustomDebugStringConvertible, Equatable {
+
+    /// The SDK name.
+    public let name: String
+
+    /// The SDK version.
+    public let version: String
+
+    static let stripeIOS = SDKVersion(name: "stripe-ios", version: STPAPIClient.STPSDKVersion)
```
mliberatore, line 23: This is `internal`, but I’m flexible on the naming here. I don't love the casing since it makes the i in iOS capitalized. Happy to take any suggestions.

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/Errors/StripeCryptoOnrampErrorRenderer.swift
```diff
@@ -0,0 +1,102 @@
...
+        developerBody: String,
+        code: String,
+        nextStep: String,
+        docURLString: String?,
+        sdkVersions: [SDKVersion]
+    ) -> String {
+        var lines = [
+            developerBody,
+            "",
+            "Code: \(code)",
```
jeanregisser, line 61: So the code stands out and "Next step:` is given more visual room.
```suggestion
            "Code: \(code)",
            "",
```

jeanregisser (dismissed):
Love it ❤️

jeanregisser (dismissed):

jeanregisser (approved):

Issue #6423, opened 2026-05-07
[Feature] [Identity] Project Orion Redesign Part 1


Pearapps:
## Project Orion Redesign

As part 1 of this project, we are updating necessary parts of the identity flow to accommodate the new selfie flow.

### Scope

- [ ] Update colors throughout the flow
- [ ] Update relevant icons, primarily in the selfie flow
- [ ] Update the biometric consent flow
- [ ] Add static tick marks
- [ ] Update the selfie viewfinder
- [ ] Update selfie flow functionality
- [ ] Update selfie errors and toasts
- [ ] Update the overall flow to account for the new selfie flow
- [ ] Add sound design updates
- [ ] Add various animation tweaks
- [ ] Update small miscellaneous UI details

PR #6523, opened 2026-05-27
[Identity] Project Orion

Pearapps:
Progress on https://github.com/stripe/stripe-ios/issues/6423

This started as a pass on the selfie capture UX but has grown into the full Orion / 3D selfie capture flow on iOS: new capture UI, MediaPipe head pose detection, front → left → right pose capture, server-driven experiment gating, uploads and submission of the additional 3D data, plus a handful of related Identity UX updates.

- updates the overall selfie capture flow
  - removes the extra `selfies captured` screen at the end
  - keeps the user on the camera after the final image instead of moving to a separate review state
  - freezes and blurs the last captured image while we check and save everything
  - shows the saving and error states directly inside the camera UI
  - adds the `Having trouble?` action below the viewfinder
  - keeps the existing legacy flow when 3D capture is not enabled

- adds the new oval / tick mark face guide (`CaptureTickMarksView`)
  - draws the capture guide using animated ticks around the viewfinder
  - shows the centered shadow treatment when the user is in the right position
  - updates the ticks as the user moves through the different capture states
  - adds progress, success, and transition animations instead of abruptly changing states
  - supports the different front, left, and right pose instructions
  - updates the status treatment depending on whether we are positioning, capturing, checking, saving, or showing an error

- adds the full front → left → right 3D capture flow
  - starts with the normal front-facing selfie capture
  - asks the user to slowly turn left and then right
  - shows directional guidance for each side
  - looks for the best frame during each pose capture window
  - gives haptic and audio feedback (`selfie_capture_confirmed.wav`) when a pose is accepted
  - adds a fallback path (pose capture fallback timer) so the flow does not get completely stuck if someone cannot reach the ideal side pose
  - transitions between poses without leaving the camera

- adds MediaPipe face pose detection
  - adds `LocalPackages/MediaPipeSPM`, a local Swift package wrapping the MediaPipe Tasks Vision / Common artifacts
    - graph archives repackaged as `MediaPipeCommonGraphLibraries.xcframework` so SPM can link the Face Landmarker calculators
    - includes refresh scripts and bundles the upstream Apache 2.0 license
  - bundles the face landmarker model (`face_landmarker.task`) and loads it with the other Identity ML models
  - new `MediaPipeFacePoseDetector` / `FaceGeometryDetector` layer uses face landmarks and blendshapes to calculate yaw, pitch, and roll in real time
  - passes that information through the existing face scanner (`FaceScannerOutput`, `FaceCaptureScanningState`)
  - uses the pose information (±15° yaw thresholds) to determine when the user has reached the requested left or right pose
  - keeps the MediaPipe work behind the 3D capture gate so it is not loaded for the legacy flow
  - reports a `media_pipe_face_detector_unavailable` analytics error if the detector cannot be created

- makes the 3D flow server driven
  - reads the selfie `pose_sequence` returned by the Verification Page API
  - also reads the new top-level `experiments` response (`idprod_3d_face_capture_mobile`)
  - uses those fields to decide whether to
    - load the MediaPipe face models
    - show the new 3D capture UI
    - capture the left and right poses
    - submit the additional 3D data
  - leaves the local override as `nil` so the normal behavior is controlled by the server
  - keeps the experiment exposure analytics connected to the selfie screen being presented

- uploads and submits the new 3D selfie data
  - still sends the first, best, and last front images in high and low resolution
  - also uploads the selected left and right images as `left_high_res_image` and `right_high_res_image`
  - includes per-frame data for each accepted capture (first, best, last, left, right)
    - yaw, pitch, and roll
    - face bounds and input size
    - face score and score variance
    - blur score
    - capture timestamp and capture order
    - camera information
    - compacted face landmark / blendshape results
  - compacts the face landmark result so it stays below the API's 5000 character limit
  - only includes the new side-pose fields when 3D capture is enabled for the session

- adds new face validation feedback
  - adds `tooFar` / `tooClose` validation issues based on face size in the frame
  - shows "move closer" style guidance when the user is too far away, with debouncing so the prompt does not flicker

- other Identity UX updates included here
  - selfie warmup screen: adds the optional training consent section ("Help us improve verifications") with Allow / Decline buttons, HTML body text with tappable links, and the new warmup illustration
  - adds `brandColor` to `IdentityVerificationSheet.Configuration` (new public API) and themes the native flow's primary buttons with it, picking a contrast-appropriate text color
  - success screen: banner header treatment and centered body text
  - camera preview container supports a separate viewfinder shadow style

## Build / CI changes to be aware of

- **arm64-only simulator slices**: the MediaPipe artifacts are arm64-only, so the `x86_64` simulator slice is removed
  - `ci_scripts/api_diff/diff_public_interface.rb` now inspects the `ios-arm64-simulator` slice instead of `ios-arm64_x86_64-simulator`
  - `ci_scripts/api_diff/build_xcframeworks.rb` and `.periphery.yml` build with `ARCHS=arm64`
  - this means **StripeIdentity no longer runs on Intel (x86_64) simulators** — anyone on an Intel Mac or running x86_64 sim builds will be affected
- **new local SPM package**: `Package.swift`, `Stripe.xcworkspace`, and `StripeIdentity.podspec` all reference `LocalPackages/MediaPipeSPM`
  - CocoaPods integration vendors the MediaPipe xcframeworks, adds the private headers, and ships the model + audio in the resource bundle
  - adds ~3.7 MB for the bundled face landmarker model
- **release artifacts**: `ci_scripts/export_builds.rb` now copies `LICENSE`, `NOTICE`, and `LocalPackages/MediaPipeSPM/LICENSE-MediaPipe` into the exported build archives
- **NOTICE** updated with the MediaPipe attribution (Apache 2.0)
- bitrise.yml now forces the AllStripeFrameworks simulator build to ARCHS=arm64 so CI can link the arm64-only MediaPipe artifacts.

## Example app

- adds a "3D face capture enabled" switch to the Identity example and playground apps
- passes `3d_face_capture_enabled` through to the test backend

## Tests

- coverage for the experiment and `pose_sequence` gating
- coverage for enabling and disabling the 3D flow
- coverage for submitting the left and right image fields
- coverage for the additional frame metadata
- coverage for compacting the face landmark payload
- coverage for capture ordering and timestamps
- coverage for the warmup training consent screen
- updates the affected selfie snapshots for the new oval, instructions, and saving states


fedefrappi (dismissed):

StripeIdentity.podspec
```diff
@@ -18,9 +18,10 @@ Pod::Spec.new do |s|
   s.swift_version		               = '5.0'
   s.weak_framework = 'SwiftUI'
   s.source_files                   = 'StripeIdentity/StripeIdentity/**/*.swift'
-  s.ios.resource_bundle            = { 'StripeIdentityBundle' => 'StripeIdentity/StripeIdentity/Resources/**/*.{lproj,json,png,xcassets}' }
+  s.ios.resource_bundle            = { 'StripeIdentityBundle' => 'StripeIdentity/StripeIdentity/Resources/**/*.{lproj,json,png,xcassets,task}' }
```
fedefrappi, line 21:
```suggestion
  s.ios.resource_bundle            = { 'StripeIdentityBundle' => 'StripeIdentity/StripeIdentity/Resources/**/*.{lproj,json,png,xcassets,task,wav}' }
```

StripeIdentity/StripeIdentity/Source/API Bindings/Models/VerificationPage/VerificationPage.swift
```diff
@@ -56,6 +56,31 @@ extension StripeAPI {
...
+        return experiments.contains { experiment in
+            experiment.experimentName == ThreeDFaceCaptureExperiment.experimentName
+        }
+    }
+
+    var enable3DFaceCapture: Bool {
+        if let localOverride = ThreeDFaceCaptureExperiment.localOverride {
+            return localOverride
+        }
+        return has3DFaceCaptureExperiment
```
fedefrappi, line 77: I think on Android this enabling logic also checks whether `pose_sequence` contains `left` or `right` and enables the feature independently of the experiment, I am not sure which one is correct

StripeIdentity/StripeIdentity/Source/NativeComponents/ViewControllers/SelfieWarmupViewController.swift
```diff
@@ -14,18 +14,70 @@ final class SelfieWarmupViewController: IdentityFlowViewController {
...
+        }()
+        let buttonTopContentViewModel: HTMLTextView.ViewModel? = {
+            guard let trainingConsentText, !trainingConsentText.isEmpty else {
+                return nil
+            }
+
+            return .init(
+                text: Self.trainingConsentHTMLText(trainingConsentText),
+                style: .html(makeStyle: Self.trainingConsentHTMLStyle),
+                didOpenURL: { url in
```
fedefrappi, line 65: we may need a `[weak self]` here

StripeIdentity/StripeIdentity/Source/API Bindings/SelfieUploader+API.swift
```diff
@@ -51,7 +61,7 @@ extension StripeAPI.VerificationPageDataFace {
                 let exposureDuration = properties.exposureDuration
 
                 if exposureDuration.isNumeric {
-                    return Int(properties.exposureDuration.seconds * 1000)
+                    return Int(properties.exposureDuration.seconds * 1_000_000)
```
fedefrappi, line 64: Is this change of unit expected for both old and new data?

Example/IdentityVerification Example/IdentityVerification Example/PlaygroundViewController.swift
```diff
@@ -319,6 +340,10 @@ class PlaygroundViewController: UIViewController {
         task.resume()
     }
 
+    private func setLocal3DFaceCaptureOverride() {
+        IdentityVerificationSheet.local3DFaceCaptureOverride = is3DFaceCaptureEnabled
```
fedefrappi, line 344: just making sure we are keeping the local override and are not waiting for some backend change that will let us remove this before merging

StripeIdentity/StripeIdentity/Source/NativeComponents/Views/Selfie/SelfieScanningView.swift
```diff
@@ -468,3 +1041,891 @@ extension SelfieScanningView: CheckboxButtonDelegate {
         return false
     }
 }
+
+private final class CaptureTickMarksView: UIView {
```
fedefrappi, line 1045: I'd move this to a separate file because the main one is already 1000+ lines and this class is almost the same size

StripeIdentity/StripeIdentity/Source/IdentityVerificationSheet.swift
```diff
@@ -12,6 +12,8 @@ import UIKit
 /// A drop-in class that presents a sheet for a user to verify their identity.
 /// This class is in beta; see https://stripe.com/docs/identity for access
 final public class IdentityVerificationSheet {
+    /// Overrides server-driven 3D face capture enablement when non-nil.
+    @_spi(STP) public static var local3DFaceCaptureOverride: Bool?
```
fedefrappi, line 16: same as before, is this something we need to keep or can it be removed before merging?

StripeIdentity/StripeIdentity/Source/API Bindings/SelfieUploader+API.swift
```diff
@@ -64,7 +74,215 @@ extension StripeAPI.VerificationPageDataFace {
...
+            {
+                compactCategory["category_name"] = displayName
+            }
+            return compactCategory
+        }
+
+        let compactPayload: [String: Any] = [
+            "categories": compactCategories,
+        ]
+        guard JSONSerialization.isValidJSONObject(compactPayload),
```
fedefrappi, line 252: Will this ever be false considering how it is built above? I think we can remove the guard but it won't hurt if we keep it

StripeIdentity/StripeIdentity/Source/API Bindings/SelfieUploader+API.swift
```diff
@@ -64,7 +74,215 @@ extension StripeAPI.VerificationPageDataFace {
...
+    private static func encodedCameraInfo(
+        from exifMetadata: CameraExifMetadata?
+    ) -> String? {
+        guard let cameraLabel = exifMetadata?.lensModel else {
+            return nil
+        }
+        let payload: [String: Any] = [
+            "cameraLabel": cameraLabel,
+        ]
+        guard JSONSerialization.isValidJSONObject(payload),
```
fedefrappi, line 281: Same as above, will this ever be false considering how it is built above? I think we can remove the guard but it won't hurt if we keep it

StripeIdentity/StripeIdentity/Source/NativeComponents/ViewControllers/SelfieCaptureFeedbackGenerator.swift
```diff
@@ -0,0 +1,61 @@
...
+        let bundle = StripeIdentityBundleLocator.resourcesBundle
+
+        return bundle.url(
+            forResource: Constants.confirmationSoundName,
+            withExtension: Constants.confirmationSoundExtension,
+            subdirectory: Constants.confirmationSoundSubdirectory
+        ) ?? bundle.url(
+            forResource: Constants.confirmationSoundName,
+            withExtension: Constants.confirmationSoundExtension
+        )
```
fedefrappi, line 59: do we need to keep both here?

StripeIdentity/StripeIdentity/Source/API Bindings/SelfieUploader+API.swift
Pearapps: Yes. This field is shared by both flows, and the API expects it in microseconds. The previous calculation was sending milliseconds

StripeIdentity/StripeIdentity/Source/NativeComponents/Views/Selfie/CaptureTickMarksView.swift
```diff
@@ -0,0 +1,899 @@
...
+        checkmarkPath.addLine(to: CGPoint(x: bounds.width * 0.70, y: bounds.height * 0.38))
+        checkmarkPath.lineCapStyle = .round
+        checkmarkPath.lineJoinStyle = .round
+        checkmarkPath.lineWidth = 2.6
+        context.setBlendMode(.clear)
+        UIColor.white.setStroke()
+        checkmarkPath.stroke()
+    }
+}
+
```
cjmisenas-stripe: `StripeIdentity/StripeIdentity/Source/NativeComponents/Views/Selfie/CaptureTickMarksView.swift:899:1: error: Trailing Newline Violation: Files should have a single trailing newline (trailing_newline)`

cjmisenas-stripe (changes requested):

fedefrappi (dismissed):

fedefrappi (dismissed):

fedefrappi (dismissed):

cjmisenas-stripe (dismissed):

mnn-stripe (dismissed):

mnn-stripe (approved):

PR #6533, opened 2026-06-01
PMM Experiment — Update PaymentMethodMessagingPromotionsHelper for exposure logging

gbirch-stripe:
## Summary
- Simplify code to be better suited for new exposure logging strategy
- Add exposure logging when getting experiment data. We wait until the experiment data is actually used to log an exposure. This is because we want to minimize the number of sessions that log an exposure but where the experiment data doesn't actually impact the customer experience. If we logged an exposure when we first fetch the experiment data (upon payment sheet loading), we could potentially gather extraneous exposures.
- Make the PaymentMethodMessagingPromotionsHelper not `@MainActor` isolated and instead enforce thread safety with a lock. `PaymentMethodMessagingPromotionsHelper` will be used from non-`@MainActor` contexts that cannot easily be made `@MainActor` and so it is easiest to just protect the read/write to the one piece of async fetched code this way.

## Motivation

## Testing
- added unit tests for PaymentMethodMessagingPromotionsHelper
- added integration tests that verify that PMM experiment logging is completed for both PS and embedded

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingPromotionsHelper.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingPromotionsHelper.swift
@@ -74,11 +74,9 @@ final class PaymentMethodMessagingPromotionsHelper {
         self.paymentMethodTypes = paymentMethodTypes
         self._experiment = PaymentMethodMessagingPromotionsExperiment(elementsSession: elementsSession)
         self.analyticsHelper = analyticsHelper
-
-        fetchData()
     }
 
-    private func fetchData() {
+    func fetchData() {
         // Check experiment group, and only proceed if in treatment group
         guard experiment.group == .treatment else {
             return
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingPromotionsHelper.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingPromotionsHelper.swift
@@ -9,130 +9,121 @@ import Foundation
 @_spi(STP) import StripeCore
 @_spi(STP) import StripePayments
 
-private let paymentSheetPMMESupportedPaymentMethodTypes: [STPPaymentMethodType] = [
-    .afterpayClearpay,
-    .affirm,
-    .klarna,
-]
-private let paymentSheetPMMESupportedPaymentMethodIdentifiers = Set(paymentSheetPMMESupportedPaymentMethodTypes.map(\.identifier))
+@_spi(STP) public enum PaymentSheetTestHooks {
+    private static let paymentMethodMessagingPromotionLoadDelayLock = NSLock()
+    private static var _paymentMethodMessagingPromotionLoadDelay: TimeInterval = 0
 
-@MainActor
+    // Humans and LLMs: this is test code for the PMM experiment test branch only. Double-check before propagating it.
+    @_spi(STP) public static var paymentMethodMessagingPromotionLoadDelay: TimeInterval {
+        get {
+            paymentMethodMessagingPromotionLoadDelayLock.lock()
+            defer { paymentMethodMessagingPromotionLoadDelayLock.unlock() }
+            return _paymentMethodMessagingPromotionLoadDelay
+        }
+        set {
+            paymentMethodMessagingPromotionLoadDelayLock.lock()
+            defer { paymentMethodMessagingPromotionLoadDelayLock.unlock() }
+            _paymentMethodMessagingPromotionLoadDelay = max(0, newValue)
+        }
+    }
+}
+
 final class PaymentMethodMessagingPromotionsHelper {
+
+    static let supportedPaymentMethods: [PaymentSheet.PaymentMethodType] = [
+        .stripe(.afterpayClearpay),
+        .stripe(.affirm),
+        .stripe(.klarna),
+    ]
+
     struct PromotionContent: Equatable {
         let promotion: String
         let learnMoreText: String
         let infoUrl: URL
     }
 
-    private enum FetchState: Equatable {
-        case idle
-        case loading
-        case completed([String: PromotionContent])
+    private let elementsSession: STPElementsSession
+    private let intent: Intent
+    private let configuration: PaymentElementConfiguration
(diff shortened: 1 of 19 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingPromotionsHelper.swift
```diff
@@ -9,167 +9,135 @@ import Foundation
...
 @_spi(STP) import StripePayments
 
-private let paymentSheetPMMESupportedPaymentMethodTypes: [STPPaymentMethodType] = [
-    .afterpayClearpay,
-    .affirm,
-    .klarna,
-]
-private let paymentSheetPMMESupportedPaymentMethodIdentifiers = Set(paymentSheetPMMESupportedPaymentMethodTypes.map(\.identifier))
-
-@MainActor
```
yuki-stripe, line 19: I checked out the branch and it looks like this compiles if I mark `_mockFlowController` as `@MainActor`, did you find other trouble?

gbirch-stripe: Once hooked up, the PaymentSheetFormFactory will need to fetch the promotion from the PaymentMethodMessagingPromotionsHelper. The form factory is non-isolated and from my tinkering seems to be very tricky / IMO not worth it to try to main actor isolate.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingPromotionsHelper.swift
```diff
@@ -9,167 +9,135 @@ import Foundation
...
-    private static func makeConfiguration(
-        intent: Intent,
-        elementsSession: STPElementsSession,
-        configuration: PaymentElementConfiguration,
-        paymentMethodTypes: [PaymentSheet.PaymentMethodType]
-    ) -> PaymentMethodMessagingElement.Configuration? {
+        // Only fetch data if we have an amount and currency (for example setup mode won't have this)
         guard let amount = intent.amount, let currency = intent.currency else {
-            return nil
+            return
```
yuki-stripe, line 87: just double checking but i assume it's fine that a client could be in treatment and have behavior identical to control?

yuki-stripe: This scenario happens in all these guard conditions and also when the promotions loads *after* `promotion()` is accessed, I think?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift
```diff
@@ -215,11 +199,13 @@ final class PaymentSheetLoader {
             let prefetchedSavedPaymentMethods = try await prefetchedSavedPaymentMethodsTask.value
             let filteredSavedPaymentMethods = filterSavedPaymentMethods(intent: intent, elementsSession: elementsSession, configuration: configuration, prefetchedSPMs: prefetchedSavedPaymentMethods, loadTimings: loadTimings)
 
-            let paymentMethodMessagingPromotionsHelper = PaymentMethodMessagingPromotionsHelper(elementsSession: elementsSession)
-            paymentMethodMessagingPromotionsHelper.prefetchIfNeeded(
+            // Initializing PaymentMethodMessagingPromotionsHelper kick off loading the PMM/BNPL info
```
yuki-stripe: ooc why remove the 'prefetchIfNeeded', i liked it explicit

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingPromotionsHelper.swift
yuki-stripe: Got it. Unfortunate! We should all of that MainActor at some point.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingPromotionsHelper.swift
gbirch-stripe: Setup mode should at some point be excluded from the experiment: https://stripe.slack.com/archives/C09C4G4RLH3/p1779401280964029?thread_ts=1779401056.255699&cid=C09C4G4RLH3. Will update this code to assert if/when that happens

yuki-stripe (dismissed):

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheet+APIMockTest.swift
```diff
@@ -156,7 +156,7 @@ final class PaymentSheetAPIMockTest: APIStubbedTestCase {
         }
 
         stub { urlRequest in
-            guard let pathComponents = urlRequest.url?.pathComponents else { return false }
+            guard let pathComponents = urlRequest.url?.pathComponents, pathComponents.count >= 3 else { return false }
```
gbirch-stripe, line 159: This is unrelated to the current change, but I was getting lots of test crashes on the line below because this isn't guarded against.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingPromotionsHelper.swift
yuki-stripe: Just pointing out that there are other conditions where client is in treatment but we early return, like PM types not containing bnpls + the promotion async edge case mentioned before!

yuki-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingPromotionsHelper.swift
gbirch-stripe: Good thought! The goal is to de-dilute the data by minimizing these scenarios. For example, sessions not using BNPLs are already excluded from the experiment.

PR #6537, opened 2026-06-02
Reduce network calls when updating a CheckoutSession

porter-stripe:
## Summary
- Reduces a extra network call to get the latest session as the update call gives it back to us

## Motivation
- Latency

## Testing
- Existing tests

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+CurrencySelectorView.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+CurrencySelectorView.swift
@@ -238,7 +238,8 @@ extension Checkout.CurrencySelectorView: TwoOptionSelectorViewDelegate {
         // Disable interaction during the API call
         selectorView?.setEnabled(false)
 
-        Task {
+        Task { [weak self] in
+            guard let self else { return }
             do {
                 try await checkout.selectCurrency(id)
                 STPAnalyticsClient.sharedClient.log(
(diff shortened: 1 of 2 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+CurrencySelectorView.swift
```diff
@@ -238,7 +238,8 @@ extension Checkout.CurrencySelectorView: TwoOptionSelectorViewDelegate {
         // Disable interaction during the API call
         selectorView?.setEnabled(false)
 
-        Task {
+        Task { [weak self] in
+            guard let self else { return }
```
gbirch-stripe, line 242: What's the case where this makes a difference? Wouldn't this code execute nearly instantaneously after the `Task` initialization?

porter-stripe: Since selectCurrency is a network call, without `[weak self]`, if the view is dismissed mid-request, the Task keeps it alive until the response arrives and then does pointless UI updates. With `[weak self]`, it deallocates immediately and the block no-ops.

gbirch-stripe: Makes sense, got it!

gbirch-stripe (approved):

PR #6540, opened 2026-06-02
PMM Experiment — Use correct UserInterfaceStyle when showing PaymentMethodMessaging modal in sublabel view

gbirch-stripe:
## Summary
- Use correct UserInterfaceStyle when showing PaymentMethodMessaging modal in sublabel view
- Simplify derivation of UserInterfaceStyle for BNPLFormHeaderView by just checking the style in the view (which is already set according to the config)

## Testing
- existing unit and snapshot tests

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/BNPLFormHeaderView.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/BNPLFormHeaderView.swift
@@ -8,7 +8,6 @@ import UIKit
 
 final class BNPLFormHeaderView: UIView {
     private let appearance: PaymentSheet.Appearance
-    let style: PaymentSheet.UserInterfaceStyle
     let promotion: String
     let learnMoreText: String
     let infoUrl: URL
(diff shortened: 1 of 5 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/BNPLFormHeaderView.swift
```diff
@@ -68,18 +56,7 @@ final class BNPLFormHeaderView: UIView {
...
-    private var pmmeStyle: PaymentMethodMessagingElement.Appearance.UserInterfaceStyle {
-        switch style {
-        case .automatic:
-            return .automatic
-        case .alwaysLight:
-            return .alwaysLight
-        case .alwaysDark:
-            return .alwaysDark
-        }
+        PMMEInfoModal.present(infoUrl: infoUrl, style: traitCollection.isDarkMode ? .alwaysDark : .alwaysLight, from: self)
```
yuki-stripe, line 59: just double checking that this still works when the system is light mode, and you've set `alwaysDark` on Configuration or have customized appearance w/ a dark background color on this row?

yuki-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/BNPLFormHeaderView.swift
gbirch-stripe: - yes I verified that this properly responds to `alwaysDark/alwaysLight` when the system is set to the other one
- IMO this shouldn't respond to the custom appearance of the row background (and it doesn't). This determines the theming of the information popup that appears when tapping to learn more. This popup is fullscreen and thus its contrast with surrounding colors is not relevant, unlike place where we use `.contrastingColor` to dynamically treat surfaces as light or dark mode elsewhere in the SDK

PR #6541, opened 2026-06-02
CryptoOnramp SDK: Updates EU Declaration Acceptance Screen to Use HTML

mliberatore:
## Summary
- The backend is providing basic HTML for formatting+links. We now render this HTML using `NSAttributedString`.
- We adjust the sheet size to fit the content instead of using a full-screen presentation, as the declaration text is expected to stay relatively small.
- Updated the CTA to "Accept" to match the declaration text.

## Motivation
To properly render the contents sent by the backend as opposed to raw html.

## Testing
1. Created a new EU user account in the example app.
2. Once I reached the declaration screen, tested in both light and dark modes to ensure everything was legible.
3. Made sure that tapping the links opened SafariViewController properly.

| Dark | Light | Opened Link |
| --- | --- | --- |
|  |  |  |

## Changelog

N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/CRSCARF Declaration/CRSCARFDeclarationContentViewController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/CRSCARF Declaration/CRSCARFDeclarationContentViewController.swift
@@ -9,7 +9,6 @@ import SafariServices
 @_spi(STP) import StripeCore
 @_spi(STP) import StripeUICore
 import UIKit
-import WebKit
 
 /// The content view of `CRSCARFDeclarationViewController`, which displays CRS/CARF declaration HTML with a confirmation button.
 final class CRSCARFDeclarationContentViewController: UIViewController, BottomSheetContentViewController {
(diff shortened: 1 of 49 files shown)


StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/CRSCARF Declaration/CRSCARFDeclarationContentViewController.swift
```diff
@@ -52,41 +74,6 @@ final class CRSCARFDeclarationContentViewController: UIViewController, BottomShe
         fatalError("init(coder:) has not been implemented")
     }
 
-    private lazy var scrollView: UIScrollView = {
```
mliberatore, line 55: Note: this bottom sheet component wraps its content in a scroll view already for when content is larger than the sheet, so we don't need to provide our own.

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/CRSCARF Declaration/CRSCARFDeclarationContentViewController.swift
```diff
@@ -97,25 +84,47 @@ final class CRSCARFDeclarationContentViewController: UIViewController, BottomShe
...
+    private var attributedDeclarationHTML: NSAttributedString {
+        guard let attributedString = try? NSMutableAttributedString(
+            data: Data(html.utf8),
+            options: [
+                .documentType: NSAttributedString.DocumentType.html,
+                .characterEncoding: String.Encoding.utf8.rawValue,
+            ],
+            documentAttributes: nil
+        ) else {
+            return NSAttributedString(string: html, attributes: declarationBaseAttributes)
```
mliberatore, line 112: Not a great fallback here, but in my testing I haven't been able to the declaration to fail. Open to ideas, but a secondary HTML-handling approach feels like too much for a fallback to an error that we can't reliably produce.

jeanregisser (approved):
Great! 🚀

PR #6542, opened 2026-06-03
PMM Experiment — Add analytics; wire up form header to data source

gbirch-stripe:
## Summary
- add analytics (not exposure logging — that was done in a previous PR)
- final wiring — enable the experiment form header by attaching it to the data source

## Motivation
https://docs.google.com/document/d/1pi9vehSvFpT80lgEYwBmE_NCcDE5nPTnRrhBrYqlzUQ/edit?tab=t.0#heading=h.4y6tyjyk6w3h

## Testing
- added analytics tests


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButton+PaymentMethodMessagingSublabelView.swift
```diff
@@ -66,7 +66,11 @@ extension RowButton {
 
         func updateSelectedState(_ isRowSelected: Bool, willDisplayForm: Bool) {
             if isRowSelected && !willDisplayForm {
-                if let promotionContent = promotionsHelper.promotion(for: paymentMethodType) {
+                // If selected, get PMME content, log an attempt to display, and expand if applicable
+                // PMM data is not always available on initial load/display of a RowButton, so we check for it right before attempting to display
+                let promotionContent = promotionsHelper.promotion(for: paymentMethodType)
+                promotionsHelper.logDisplayedAnalytic(displayedSuccessfully: promotionContent != nil)
```
yuki-stripe: does this fire for non-BNPL row buttons? 

Also is it ok to fire multiple times in a session, like when we reload the UI after an update?

yuki-stripe (dismissed):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButton+PaymentMethodMessagingSublabelView.swift
gbirch-stripe: > does this fire for non-BNPL row buttons?

No, this is only called on the PaymentMethodMessagingSublabelView

> Also is it ok to fire multiple times in a session, like when we reload the UI after an update?

It should be called every time we display the messaging, which may be multiple times if the customer taps back and forth between payment methods. Do you mind elaborating on the UI update scenario? Are you referring to the update that happens when `.setIncentive` is called? At a high level since these analytics aren't being used for the experiment, the exact quantity of events is less important — we're mostly interested in seeing the rates at which we are succeeding or failing to display the data because of the lazy loading

gbirch-stripe:
Added prevention for duplicate logging

porter-stripe (dismissed):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButton+PaymentMethodMessagingSublabelView.swift
yuki-stripe: When users call `update` methods like FlowController.update(..) (or in the future CS integration where we'll also trigger updates ourselves), our implementation just rebuilds the UI from scratch. That might trigger this code unexpectedly. It's something you could test by going to the embedded playground and toggling the Mode at the top. But if exact quantity is not important, sounds like it's a non-issue either way!

porter-stripe (approved):

PR #6545, opened 2026-06-03
CryptoOnramp SDK: EU Declaration: Fixes `<ul>` Rendering

mliberatore:
## Summary
Applying a `paragraphStyle` wholesale to the declarations HTML string overwrites any `paragraphStyle` attributes specified by `NSAttributedString.DocumentType.html`. This was causing the new `<ul>` to render incorrectly without bullets and indentation. I’ve fixed the issue by modifying the existing paragraph style for each range to include the desired `lineHeightMultiple` instead of applying it wholesale.

## Motivation

To fix the rendering of the HTML string from the backend.

## Testing

Tested by comparing the existing behavior to the latest in the example app when registering a new EU user.
| Before | After (Light Mode) | After (Dark Mode) |
| --- | --- | --- |
|  |  |  |

## Changelog

N/A


jeanregisser (approved):
🚀

PR #6554, opened 2026-06-04
Add ui mode headless support to StripeCryptoOnramp

rohan-stripe:
Always send ui_mode=headless since this is the headless-only SDK. This will enable integration against the correct flows for headless in EU, while remaining consistent behavior for headless US. 
This param needs to show up in getPlatformSettings and createPaymentToken

Committed-By-Agent: claude

### DIFF
--- a/StripeCryptoOnramp/StripeCryptoOnramp/Source/Internal/API Bindings/CreatePaymentTokenRequest.swift
+++ b/StripeCryptoOnramp/StripeCryptoOnramp/Source/Internal/API Bindings/CreatePaymentTokenRequest.swift
@@ -9,9 +9,7 @@ import Foundation
 
 /// Encodable model passed to the `/v1/crypto/internal/payment_token` endpoint.
 struct CreatePaymentTokenRequest: Encodable {
-    /// The crypto wallet address to register.
     let paymentMethod: String
-
-    /// The crypto customer ID.
     let cryptoCustomerId: String
+    let uiMode: String = "headless"
 }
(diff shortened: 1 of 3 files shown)


StripeCryptoOnramp/StripeCryptoOnramp/Source/Internal/API Bindings/CreatePaymentTokenRequest.swift
```diff
@@ -9,9 +9,13 @@ import Foundation
...
 /// Encodable model passed to the `/v1/crypto/internal/payment_token` endpoint.
 struct CreatePaymentTokenRequest: Encodable {
-    /// The crypto wallet address to register.
     let paymentMethod: String
-
-    /// The crypto customer ID.
     let cryptoCustomerId: String
+    let uiMode: String = "headless"
+
+    enum CodingKeys: String, CodingKey {
```
jeanregisser: The `CodingKeys` enum isn't needed here - `StripeJSONEncoder` already converts camelCase to snake_case automatically (same pattern as `RegisterWalletRequest`, etc.).

We can drop it and just keep `let uiMode: String = "headless"`.

jeanregisser (commented):
Thanks 🙌

`testCreatePaymentTokenSuccess` in `STPAPIClient+CryptoOnrampTests.swift` still asserts `parameters.count == 2` and doesn't check `ui_mode` - would be good to update it to match the coverage added for `getPlatformSettings`.

jeanregisser (approved):
🚀

PR #6557, opened 2026-06-04
Rename LinkPaymentController to InstantBankPaymentsController

mats-stripe:
## Summary

Renames the `LinkPaymentController` to `InstantBankPaymentsController` in anticipation of the `LinkController`. Also adds a simple analytic event when the class is initialized.

## Motivation

https://docs.google.com/document/d/1VoMxNhXmKix5kiFzWO3eJc2YzE3dpEO63rKp_Kb9F6E/edit?disco=AAAB8E45KpU

## Testing

Ensure tests still pass 

## Changelog

```
* [Changed] Renamed `LinkPaymentController` to `InstantBankPaymentsController`.
```

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/InstantBankPaymentsController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/InstantBankPaymentsController.swift
@@ -1,5 +1,5 @@
 //
-//  LinkPaymentController.swift
+//  InstantBankPaymentsController.swift
 //  StripePaymentSheet
 //
 //  Created by Bill Meltsner on 12/9/22.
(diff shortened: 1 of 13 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/InstantBankPaymentsController.swift
```diff
@@ -103,13 +102,17 @@ import UIKit
             configuration.defaultBillingDetails = billingDetails
         }
         self.configuration = configuration
+        super.init()
+        STPAnalyticsClient.sharedClient.log(
```
jkelle-stripe, line 106: Just so I have context what's our goal with this analytic? Did we have an analytic for this previously?

jkelle-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Link/InstantBankPaymentsController.swift
mats-stripe: Correct, we didn't have any analytics here previously. Goal is just to get a sense of how much usage it gets!

PR #6562, opened 2026-06-05
Refactor intent to hold a Checkout instance

porter-stripe:
## Summary
  - Refactors Intent.checkoutSession(STPCheckoutSession) to Intent.checkout(Checkout), passing the SDK's lifecycle manager instead of the raw API model
  - Removes the onConfirmed closure indirection on STPCheckoutSession; PaymentSheet now calls checkout.updateSession(_:) directly

## Motivation
- This will make it easier for future users to make updates to Checkout from MPE, like updating the billing address or other. Now we can just call `checkout.updateX(...)` within MPE easily.

## Testing
- Existing tests

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
@@ -45,9 +45,12 @@ public final class Checkout: ObservableObject {
 
     /// The underlying `STPCheckoutSession` backing the current public ``state``.
     ///
-    /// Internal callers that need `STPCheckoutSession`-specific data (e.g.
-    /// `allResponseFields`, address overrides, the expanded intent objects)
-    /// should read this rather than casting from `state.session`.
+    /// Marked `nonisolated(unsafe)` because PaymentSheet internals read this from non-MainActor
+    /// contexts. This is safe: reads only occur after the session is loaded and while the payment
+    /// UI is presented, a window during which no mutations occur. Writes are always on MainActor
+    /// because they go through `Checkout`'s MainActor-isolated mutation methods.
+    /// Requiring full MainActor isolation would propagate `@MainActor` through nearly all of
+    /// PaymentSheet's internal types, which is not warranted by the actual concurrency risk.
     nonisolated(unsafe) private(set) var stpSession: STPCheckoutSession!
 
     weak var integrationDelegate: CheckoutIntegrationDelegate?
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
@@ -110,6 +110,7 @@ public final class Checkout: ObservableObject {
         }
     }
 
+#if DEBUG
     /// Internal initializer for unit tests that injects a pre-loaded session.
     init(
         clientSecret: String,
@@ -125,7 +126,6 @@ public final class Checkout: ObservableObject {
         self.state = .loaded(session.makePublicSession())
     }
 
-    #if ENABLE_STPASSERTIONFAILURE
     /// Synchronous test-only initializer that wraps a pre-loaded session without async work.
     init(session: STPCheckoutSession) {
         self.clientSecret = ""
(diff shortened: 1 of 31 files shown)


porter-stripe:
`Checkout.swift:130 warning: Unused initializer 'init(session:)'` is used in tests.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
```diff
@@ -123,11 +123,19 @@ public final class Checkout: ObservableObject {
...
         self.stpSession = session
         self.state = .loaded(session.makePublicSession())
-        session.onConfirmed = { [weak self] response in
-            self?.updateSession(response)
-        }
     }
 
+    #if ENABLE_STPASSERTIONFAILURE
+    /// Synchronous test-only initializer that wraps a pre-loaded session without async work.
+    init(session: STPCheckoutSession) {
```
gbirch-stripe, line 130: Why can we not use the init on line 114 for this?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
```diff
@@ -123,11 +123,19 @@ public final class Checkout: ObservableObject {
         await flagImageManager.prefetchFlagImages(for: session)
         self.stpSession = session
         self.state = .loaded(session.makePublicSession())
-        session.onConfirmed = { [weak self] response in
-            self?.updateSession(response)
-        }
     }
 
+    #if ENABLE_STPASSERTIONFAILURE
```
gbirch-stripe: Is this true if and only if we are in a testing scheme? Seems like a non-idiomatic pattern that I don't see used elsewhere

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
```diff
@@ -45,10 +45,13 @@ public final class Checkout: ObservableObject {
...
-    /// `allResponseFields`, address overrides, the expanded intent objects)
-    /// should read this rather than casting from `state.session`.
-    private(set) var stpSession: STPCheckoutSession?
+    /// Marked `nonisolated(unsafe)` because PaymentSheet internals read this from non-MainActor
+    /// contexts. This is safe: reads only occur after the session is loaded and while the payment
+    /// UI is presented, a window during which no mutations occur. Writes are always on MainActor
+    /// because they go through `Checkout`'s MainActor-isolated mutation methods.
+    /// Requiring full MainActor isolation would propagate `@MainActor` through nearly all of
+    /// PaymentSheet's internal types, which is not warranted by the actual concurrency risk.
+    nonisolated(unsafe) private(set) var stpSession: STPCheckoutSession!
```
gbirch-stripe, line 54: I'm trying to brainstorm ways to better codify this and thus future proof it. A couple of (non-mutually-exclusive) ideas for thought:
1. Separate underlying data from access with a wrapper didSet/get computed property. You can separately specify `@MainActor` on the `didSet` and `nonisolated` on the `get`
2. Have some sort of lock/semaphore that is set to an open state after loading is completed and before then blocks reads.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
porter-stripe: The one on 114 is async, b/c it needs to load the flag images for snapshot tests. The sync init is just for convince in writing unit tests.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
porter-stripe: Changed to if DEBUG instead which seems more common.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
porter-stripe: Reached out in Slack, I think this is a good callout and something we should cover in the "CheckoutSessions Binding Mechanism Design Doc", I think it fits well into that read/write category and it's interesting that MPE is not MainActor but Checkout is.

porter-stripe: Chatted offline, confirmed this will be covered in the design doc.

gbirch-stripe (approved):

PR #6574, opened 2026-06-08
[Xcode 27] Fix UIButton edge inset deprecation warnings

davidme-stripe:
## Summary
Uses UIButton.Configuration instead of setting `contentEdgeInsets`, which are no longer supported as of iOS 15. We'll remove all the `else` clauses (including the calls to `contentEdgeInsets`) when we remove support for iOS 13 in a follow-up PR.

This will update all the snapshot tests, as the insets appear to be calculated slightly differently when using a `UIButtonConfiguration`. The changes look fine to me: The labels weren't perfectly centered before (e.g. the distance was 21px on one side and 20px on the other, and now it's 20px on one and 21px on the other), so I think this is an implementation detail of UIButton.

To see the snapshots yourself, check here: [View snapshot diffs](https://artifact.services.bitrise.io/files/apps/b220f4ba85d134a7/builds/ba00ef2d-9b86-48e3-9300-7c6cf15f5feb/d0846d8a4bdf74fd4fe2815bade343b6f9330e637a870a256c50a3226eba209e1c521359b8b099027638c582141ad790173bf929653fe2d07afb7d58f3932d48/index.html)


davidme-stripe:
^ The API checker seems wrong here, we're just not importing CloudKit locally. I'm not sure why we imported CloudKit here to begin with!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Views/CardScanButton.swift
```diff
@@ -19,18 +17,44 @@ extension UIButton {
...
+            }
+            config.baseForegroundColor = tintColor
+
+            scanButton = UIButton(configuration: config)
+        } else {
+            scanButton = UIButton(type: .system)
+            scanButton.titleLabel?.adjustsFontSizeToFitWidth = true
+            scanButton.setTitle(String.Localized.scan_card, for: .normal)
+            scanButton.setImage(image, for: .normal)
+            if scanButton.effectiveUserInterfaceLayoutDirection == .leftToRight {
```
davidme-stripe, line 48: This is just extracted from UIButton+Stripe.swift (which is now deleted), I'll delete this else in the next PR to remove iOS 13 support

wooj-stripe (approved):
Shipping because

PR #6577, opened 2026-06-09
Derive `isLinkPassthroughMode` from wallet type

tillh-stripe:
## Summary

This pull request updates how `isLinkPassthroughMode` is computed.

Previously, we relied on `isLinkOrigin` to be provided as part of the Elements session response. However, other calls such as `/payment_details/share` don't return this information. Therefore, we now update the logic as follows:

```
isLinkPassthroughMode = isLinkOrigin || card?.wallet.type == .link
```

With this change, we continue to take `isLinkOrigin` from the Elements sessions response while also correctly labeling newly created payment methods as originating from Link.

## Motivation

[RUN_LINK_MOBILE-87](https://jira.corp.stripe.com/browse/RUN_LINK_MOBILE-87)

## Recording

`isLinkPassthroughMode` is `true` for card and bank payments in passthrough mode. For payments in payment method mode, the method isn't called and the log line is therefore not executed.

https://github.com/user-attachments/assets/658e41c1-a774-490f-8fb3-d686e0297e82

## Changelog

N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/ElementsCustomer.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/ElementsCustomer.swift
@@ -77,14 +77,13 @@ struct ElementsCustomer: Equatable, Hashable {
                     if let linkDetails = paymentMethodWithLinkDetails.linkDetails {
                         paymentMethod.setLinkPaymentDetails(from: linkDetails)
                     } else {
-                        paymentMethod.isLinkPassthroughMode = paymentMethodWithLinkDetails.isLinkOrigin
+                        paymentMethod.isLinkOrigin = paymentMethodWithLinkDetails.isLinkOrigin
                     }
                     paymentMethod.card?.cardArt = cardArtByPaymentMethodId[paymentMethod.stripeId]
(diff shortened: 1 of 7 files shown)


StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/v1-elements-sessions/ElementsCustomer.swift
```diff
@@ -77,14 +77,13 @@ struct ElementsCustomer: Equatable, Hashable {
...
                     } else {
-                        paymentMethod.isLinkPassthroughMode = paymentMethodWithLinkDetails.isLinkOrigin
+                        paymentMethod.isLinkOrigin = paymentMethodWithLinkDetails.isLinkOrigin
                     }
                     paymentMethod.card?.cardArt = cardArtByPaymentMethodId[paymentMethod.stripeId]
                     paymentMethods.append(paymentMethod)
                 }
             } else {
                 if let paymentMethod = STPPaymentMethod.decodedObject(fromAPIResponse: json) {
-                    paymentMethod.isLinkPassthroughMode = paymentMethod.card?.wallet?.type == .link
```
tillh-stripe, line 87: Not needed anymore as this is now a computed property.

mindy-stripe (approved):

jkelle-stripe (approved):

PR #6578, opened 2026-06-09
Update the AP selector on session updates

porter-stripe:
## Summary
- Fixes a bug where the AP selector would not respond to changes in amount that are displayed in the selector. After adding a new style to the selector this became obvious.
- Fixed an issue where elements_session instances were not attached to the update responses.

## Motivation
- Bug fixes

## Testing
- New tests
- Manual

## Changelog
- N/A


StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/STPAPIClient+CheckoutSession.swift
```diff
@@ -59,10 +59,14 @@ extension STPAPIClient {
         checkoutSessionId: String,
         parameters: [String: Any]
     ) async throws -> STPCheckoutSession {
+        var params = parameters
+        params["elements_session_client"] = [
+            "is_aggregation_expected": true,
```
porter-stripe, line 64: This makes sure we have an elements/session attached to the update response.

davidme-stripe (approved):

PR #6580, opened 2026-06-09
PMM Experiment — Disable double tap selection behavior

gbirch-stripe:
## Summary
Previously, double tapping the PMM link in either the standalone PMME or the PMM in MPE would result in the double-tapped link text being selected, which it should not be. This prevents it by intercepting double tap gestures.

## Motivation
https://docs.google.com/document/d/1E-EQ4hEIvIkQFQF9doshXYKiKh1_3Y-tIecjgapSkdY/edit?tab=t.0
https://jira.corp.stripe.com/browse/MOBILESDK-4566

## Testing
Manual testing of standalone PMME and PMM in MPE

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMMEPromotionTextView.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMMEPromotionTextView.swift
@@ -27,4 +27,18 @@ class PMMEPromotionTextView: LinkOpeningTextView {
     required init?(coder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
+
+    // To avoid double-tap selection behavior we block double tap gestures. These double taps will instead open the link
+    public override func gestureRecognizerShouldBegin(_ gestureRecognizer: UIGestureRecognizer) -> Bool {
+        // Intercept tap gesture recognizers
+        if let tapGesture = gestureRecognizer as? UITapGestureRecognizer {
+            // Block the gesture if it requires a double-tap
+            if tapGesture.numberOfTapsRequired == 2 {
+                return false
+            }
+        }
+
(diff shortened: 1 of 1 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMMEPromotionTextView.swift
```diff
@@ -27,4 +27,18 @@ class PMMEPromotionTextView: LinkOpeningTextView {
...
+        // Intercept tap gesture recognizers
+        if let tapGesture = gestureRecognizer as? UITapGestureRecognizer {
+            // Block the gesture if it requires a double-tap
+            if tapGesture.numberOfTapsRequired == 2 {
+                return false
+            }
+        }
+
+        // Allow all other system gestures (scrolling, link clicks, etc.)
+        return super.gestureRecognizerShouldBegin(gestureRecognizer)
```
wooj-stripe, line 50: Did we ever figure out why this is only happening the first time the view loads?

gbirch-stripe: I think it's just some undefined behavior. The link text is not actually set to `isSelectable = false` but also tapping on it opens the link. We can't actually make the link non-selectable because otherwise it won't work, but we can disable the double tap selection gesture

wooj-stripe: I noticed that it extends from "LinkOpeningTextView" where `isSelectable` is overridden to always be `true`, and seems to allow the users to select part of the text, which i don't think we are using?  Maybe we avoid subclassing from LinkOpeningTextView if we aren't using this functionality?

wooj-stripe (approved):

PR #6582, opened 2026-06-10
Preserve Link metadata on saved payment method update

tillh-stripe:
## Summary

This pull request fixes an issue where updating a Link-branded saved payment method would remove the branding. We fix it by moving the preservation of local fields to `STPPaymentMethod` directly.

## Motivation

[RUN_LINK_MOBILE-168](https://jira.corp.stripe.com/browse/RUN_LINK_MOBILE-168)

## Testing

https://github.com/user-attachments/assets/81274969-a0c4-45df-ba86-43da61fb67c1

## Changelog

N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Categories/STPPaymentMethod+PaymentSheet.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Categories/STPPaymentMethod+PaymentSheet.swift
@@ -77,6 +77,12 @@ extension STPPaymentMethod {
         return components.joined(separator: " ")
     }
 
+    func updateLocalFields(from original: STPPaymentMethod) {
+        // The update endpoint returns a plain PaymentMethod, so preserve Link presentation state.
+        linkPaymentDetails = original.linkPaymentDetails
+        isLinkOrigin = original.isLinkOrigin
+    }
+
     func hasUpdatedCardParams(_ updatedParams: STPPaymentMethodCardParams?) -> Bool {
         guard let currCard = self.card,
               let updatedParams = updatedParams else {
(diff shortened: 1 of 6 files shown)


StripePaymentSheet/StripePaymentSheet/Source/Categories/STPPaymentMethod+PaymentSheet.swift
```diff
@@ -77,6 +77,12 @@ extension STPPaymentMethod {
         return components.joined(separator: " ")
     }
 
+    func updateLocalFields(from original: STPPaymentMethod) {
+        // The update endpoint returns a plain PaymentMethod, so preserve Link presentation state.
+        linkPaymentDetails = original.linkPaymentDetails
```
jkelle-stripe, line 82: Would we expect that the `linkPaymentDetails` would be updated by the server even if they're not returned to us? In your demo video, we update CVC of the SPM, then we call [apiClient.updatePaymentMethod](http://github.com/stripe/stripe-ios/pull/6582/changes#diff-8f47303afb8f04eed6d4094c9b208d0e7712580bcbe0cc6fe65e4a9609988eb8L47) with update params including the new CVC. Do we think that the `original` has the updated CVC in it's `linkPaymentDetails`? I'm worried there will be a mismatch of CVCs. I may not be understanding though!

tillh-stripe: No, `linkPaymentDetails` only come from the Elements session. We're primarily using the label and last4 fields from the `linkPaymentDetails`, so no concerns with outdated card information.

jkelle-stripe (approved):

PR #6586, opened 2026-06-10
Fix blank screen in specific Financial Connections scenario

tillh-stripe:
## Summary

This pull request fixes an issue in the Financial Connections flow where we'd temporarily show a blank screen when connecting a single account. As a fix, we keep the `Retrieving accounts…` placeholder on screen until the next screen is pushed on top.

## Motivation

[LINK_MOBILE-969](https://jira.corp.stripe.com/browse/LINK_MOBILE-969)

### Before

https://github.com/user-attachments/assets/0990726a-6c0e-4a61-98d8-e5a1ed419468

### After

https://github.com/user-attachments/assets/084f26ad-f9ff-4077-ba31-bb589ff08032

## Changelog

N/A

### DIFF
--- a/StripeFinancialConnections/StripeFinancialConnections/Source/Native/AccountPicker/AccountPickerViewController.swift
+++ b/StripeFinancialConnections/StripeFinancialConnections/Source/Native/AccountPicker/AccountPickerViewController.swift
@@ -150,6 +150,11 @@ final class AccountPickerViewController: UIViewController {
             .pollAuthSessionAccounts()
             .observe(on: .main) { [weak self] result in
                 guard let self = self else { return }
+
+                // There are scenarios where we want to keep the "Retrieving accounts…" view on the screen
+                // even though retrieval is done, as we otherwise would just render a blank screen.
+                var keepShowingLoadingView = false
+
                 switch result {
                 case .success(let accountsPayload):
                     self.dataSource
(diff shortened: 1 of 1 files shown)


StripeFinancialConnections/StripeFinancialConnections/Source/Native/AccountPicker/AccountPickerViewController.swift
```diff
@@ -150,6 +150,11 @@ final class AccountPickerViewController: UIViewController {
             .pollAuthSessionAccounts()
             .observe(on: .main) { [weak self] result in
                 guard let self = self else { return }
+
+                // There are scenarios where we want to keep the "Retrieving accounts…" view on the screen
+                // even though retrieval is done, as we otherwise would just render a blank screen.
```
mats-stripe, line 155: "There are scenarios" feels vague to me - what are the scenarios exactly? Could we write them here so it's a bit easier to understand when this happens 🙏

mats-stripe (approved):
One non-blocking comment. Haven't tophatted but trusting the "After" video

StripeFinancialConnections/StripeFinancialConnections/Source/Native/AccountPicker/AccountPickerViewController.swift
tillh-stripe: The scenarios should be apparent from the conditional below where `keepShowingLoadingView` is set to `true`. I'd rather not explicitly enumerate this in the comment to avoid it being out of date later on.

PR #6588, opened 2026-06-11
CryptoOnramp SDK: Adds Rich Error Detail to Additional Attestation Failure

mliberatore:
## Summary
This is the native implementation version of https://github.com/stripe/stripe-react-native/pull/2486

## Motivation
To continue improving common errors seen during integration to reduce developer support load.

## Testing
1. Unit test added.
2. Modified `func deviceCanUseNativeLink(useAttestationEndpoints:,apiClient:)` to return `false`, and launched the example app.
3. Saw the error print to the console and the user message shown on screen.
CleanShot 2026-06-23 at 11 11 45

## Changelog

N/A


StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/Errors/Concrete Error Types/AppAttestationAPIError.swift
```diff
@@ -63,6 +67,12 @@ public struct AppAttestationAPIError: StripeCryptoOnrampAPIError, APIErrorContex
             return "App attestation failed: the App Attest environment does not match this Stripe mode."
         case "ios_attestation_validation_failed":
             return "App attestation failed: the App Attest attestation could not be validated."
+        case "app_attestation_unavailable":
+            return """
+            App attestation unavailable: this app isn't configured to use Stripe Crypto Onramp.
+
+            This usually means app attestation isn't enabled for this Stripe account, or this app isn't registered as a trusted application. Use your iOS bundle ID and contact Stripe to enable app attestation or register the app for this account.
```
mliberatore: Note: The messages are otherwise the same ones from https://github.com/stripe/stripe-react-native/pull/2486, but I‘ve removed android-specific terminology from this particular line.

jeanregisser (commented):
Great!

Since you offered the refactor: a thin concrete type conforming to `StripeCryptoOnrampError` (not `StripeCryptoOnrampAPIError`) would be a cleaner fit here, since this error isn't API-backed and the current approach needs synthetic reason/apiErrorCode values plus carries inapplicable fields like `requestID` and `apiMessage`.

mliberatore:
Updated. Here's the latest developer message log screenshot using the testing steps above. 
Xcode 2026-06-15 09 21 03

StripeCryptoOnramp/StripeCryptoOnrampTests/Unit/CryptoOnrampCoordinatorErrorMappingTests.swift
```diff
@@ -70,6 +71,46 @@ final class CryptoOnrampCoordinatorErrorMappingTests: XCTestCase {
...
+
+        let richError = attestationError as StripeCryptoOnrampError
+        XCTAssertEqual(richError.code, "app_attestation_unavailable")
+        XCTAssertEqual(richError.userMessage, attestationError.userMessage)
+        XCTAssertEqual(richError.developerMessage, attestationError.developerMessage)
+
+        XCTAssertEqual(attestationError.developerMessage, """
+        App attestation unavailable: this app isn't configured to use Stripe Crypto Onramp.
+
+        This usually means app attestation isn't enabled for this Stripe account, or this app isn't registered as a trusted application. Use your iOS bundle ID and contact Stripe to enable app attestation or register the app for this account.
```
jeanregisser, line 103: Can we include the mode (live/test) and bundle id as context here?

jeanregisser (commented):
Nice!

StripeCryptoOnramp/StripeCryptoOnrampTests/Unit/CryptoOnrampCoordinatorErrorMappingTests.swift
mliberatore: Here's the latest update (also reflected in the main PR description):
Slack 2026-06-23 12 23 13


Notable API changes are also reflected in `CHANGELOG.md`. Storage for diagnostic details about the error have moved to `DiagnosticContext`, a not-publicly-exposed helper type to handle storage for these properties for both API and non-API errors. The `developerMessage` "context" properties discussed in Slack are now removed from the public APIs and only seen in the `developerMessage` (with the addition of `sdkVersions`), while other properties remain on the public error types.

```swift
struct DiagnosticContext {
    let sdkVersions: [SDKVersion]
    let operation: String
    let appPackageName: String?
    let mode: String?
}
```
As a result `context: APIErrorContext` becomes `apiErrorContext: APIErrorContext` for better disambiguation where available on the same type.

The reason it's not a small simple change to expose `mode` in the `developerMessage` is that it is not available statically (relies on the current `apiClient` used by the coordinator, not the singleton instance). Furthermore, getting `operation` into the `developerMessage` requires the plumbing for additional stored properties, so this should be more future-proof for additional properties that may be added to the `developerMessage`.

jeanregisser (approved):
Awesome 🚀 

One thought is if we introduce other rich errors and want additional context fields (only relevant to those new errors), the current way is not directly extensible from what I see.

PR #6589, opened 2026-06-11
Remove PaymentSheet.init(Checkout)

porter-stripe:
## Summary
- Removes PaymentSheet.init(Checkout), replaced by PS.FC and EPE
- Replaces/removes PaymentSheet as an option in the playground when CheckoutSessions is selected

## Motivation
- Product brief

## Testing
- Manual
- CI

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetLoaderStubbedTest.swift
+++ b/StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetLoaderStubbedTest.swift
@@ -529,36 +529,6 @@ class PaymentSheetLoaderStubbedTest: APIStubbedTestCase {
         wait(for: [loaded], timeout: 2)
     }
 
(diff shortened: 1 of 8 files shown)


porter-stripe:
The code flagged is actually used but in STP functions which is why it is flagged.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetLoaderStubbedTest.swift
```diff
@@ -529,36 +529,6 @@ class PaymentSheetLoaderStubbedTest: APIStubbedTestCase {
         wait(for: [loaded], timeout: 2)
     }
 
-    func testCheckoutSessionWithoutEmailThrowsError() {
```
porter-stripe, line 532: Not needed anymore, we will just enforce this at confirm time not loading time.

yuki-stripe (approved):

PR #6595, opened 2026-06-11
PMM Experiment — Link out to external browser if non-stripe link is hit

gbirch-stripe:
## Summary
- if a non-stripe-host navigation is attempted, link out to external browser
- when returning from the external browser, refresh the page to get rid of any transition/loading state

## Motivation
3rd party UI such as Affirm's prequalification flow has UX issues when embedded into the PMMEInfoModal. Thus, we link out to the external browser if a non-stripe navigation happens
https://docs.google.com/document/d/1E-EQ4hEIvIkQFQF9doshXYKiKh1_3Y-tIecjgapSkdY/edit?tab=t.0#heading=h.cz1xkpga7giy

## Testing
https://github.com/user-attachments/assets/af5f1aa7-f9e2-4537-9bc3-1a7a73a55970

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMMEInfoModal.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMMEInfoModal.swift
@@ -115,6 +115,17 @@ final class PMMEInfoModal: UIViewController {
 
 extension PMMEInfoModal: WKNavigationDelegate {
 
+    func webView(_ webView: WKWebView, decidePolicyFor navigationAction: WKNavigationAction, decisionHandler: @escaping (WKNavigationActionPolicy) -> Void) {
+        // Non-Stripe links (e.g. partner sites) should open in the external browser
+        // rather than navigating within the modal webview.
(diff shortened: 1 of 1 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMMEInfoModal.swift
```diff
@@ -67,11 +69,17 @@ final class PMMEInfoModal: UIViewController {
...
+        self.webView = webView
+        self.themedUrl = themedUrl
         webView.navigationDelegate = self
         webView.isOpaque = false
         webView.translatesAutoresizingMaskIntoConstraints = false
         view.addAndPinSubview(webView)
 
+        // Reload original page when returning from external browser
+        // This is needed to reset the state of the page in case there was a loading animation or anything from the link-out
+        NotificationCenter.default.addObserver(self, selector: #selector(appWillEnterForeground), name: UIApplication.willEnterForegroundNotification, object: nil)
```
davidme-stripe, line 81: This feels a little off — for one thing, the view will reload every time we switch apps. Would it work to reset the state of the page after the linkout instead?

gbirch-stripe:
Replacing with https://github.com/stripe/stripe-ios/pull/6596

PR #6601, opened 2026-06-15
Update MPE automatically in Checkout

porter-stripe:
## Summary
  - Adds checkoutDidUpdate(_:) to CheckoutIntegrationDelegate so integrations (FlowController, EmbeddedPaymentElement) can re-render before the merchant's delegate is notified
  - Makes updateSession async/throws and always calls the integration delegate first, then the regular delegate
  - Refactors away the applyOverrides closure. Address updates now set the value on the current session before calling updateSession, which naturally preserves it via the existing copy step
  - Fixes a bug where local-only address updates (no tax API call) bypassed updateSession entirely, meaning the integration delegate was never notified

## Motivation
- CheckoutSessions

## Testing
- Updated/new tests
- Manual

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
@@ -19,7 +19,7 @@ import Foundation
 /// ```
 ///
 /// The async initializer loads the session from Stripe before returning,
-/// so ``state`` is guaranteed to be ``State/loaded(_:)`` immediately after initialization.
+/// so ``state`` is guaranteed to be ``State.loaded(_:)`` immediately after initialization.
 ///
 /// Observe session changes with SwiftUI by using ``state`` (published via `ObservableObject`),
 /// or in UIKit by setting a ``delegate``.
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
@@ -146,7 +146,9 @@ public final class Checkout: ObservableObject {
     /// - If any pending operation throws, the first such error is rethrown.
     /// - If the wait exceeds `timeout`, throws ``CheckoutError.timedOut``.
     ///
-    /// - Parameter timeout: Maximum time to wait, in seconds.
+    /// - Parameters:
+    ///   - timeout: Maximum time to wait, in seconds.
+    ///   - excludingCurrent: If true, excludes the last enqueued operation from the wait.
     func awaitPendingOperations(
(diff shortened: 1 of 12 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
```diff
@@ -148,9 +148,13 @@ public final class Checkout: ObservableObject {
     ///
     /// - Parameter timeout: Maximum time to wait, in seconds.
     func awaitPendingOperations(
-        timeout: TimeInterval = Checkout.defaultPendingOperationsTimeout
+        timeout: TimeInterval = Checkout.defaultPendingOperationsTimeout,
+        excludingCurrent: Bool = false
```
gbirch-stripe, line 154: Can we add documentation for this new param?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
```diff
@@ -173,7 +173,9 @@ public final class EmbeddedPaymentElement {
         checkout: Checkout
     ) async -> UpdateResult {
         do {
-            try await checkout.awaitPendingOperations()
+            // Exclude the last pending operation (the current mutation that triggered
+            // this update) to avoid deadlocking on ourselves.
+            try await checkout.awaitPendingOperations(excludingCurrent: true)
```
gbirch-stripe: I find this pretty confusing — my intuition would be that the checkout session updates and then you call `Embedded.update(checkout)` but it sounds like this function expects you to call it immediately after the update begins but before it finishes?

Looking at `Checkout+Internal.updateSession`, it seems that this is called with a new session after the update has completed, but it ends up calling `Embedded.update`?

gbirch-stripe (commented):
Is there a design doc that explains these changes?

porter-stripe:
@gbirch-stripe yes design doc here: https://docs.google.com/document/d/1CWglcOHR6I8cOWKjztPfnhnVG8axucpv-j5cWIoey7E/edit?tab=t.uzy1hgt1h5jt#heading=h.a2lhvmal87zq

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
porter-stripe: From the design doc: "The auto update to MPE is awaited by the Checkout mutating API: `Checkout.applyPromoCode()` -> update CS via PP API -> update MPE -> return from `.applyPromoCode()`"

Because the mutation awaits the MPE update before returning, its task is still in `pendingOperations` when we get here. We exclude it so we don't deadlock waiting on the operation that's waiting on us.

gbirch-stripe (approved):

PR #6603, opened 2026-06-16
Fail confirm if mutations are in flight

porter-stripe:
## Summary
-  Fail EPE/FlowController confirm() if the backing Checkout has in-flight mutations.
- When a Checkout mutation (e.g. applyPromotionCode) is in progress, the session state is stale. Confirming against it could charge the wrong amount. This adds a guard so confirm() fails immediately if checkout.pendingOperations is non-empty.

## Motivation
- https://jira.corp.stripe.com/browse/MOBILESDK-4583

## Testing
- Manual
- Quite difficult to unit test, this is just a safe guard.

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
@@ -302,6 +302,7 @@ extension PaymentSheet {
             }
         }
 
+        private weak var checkout: Checkout?
         private var isPresented = false
         private(set) var didPresentAndContinue: Bool = false
         private var confirmationChallenge: ConfirmationChallenge?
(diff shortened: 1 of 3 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
wooj-stripe, line 575: Still coming up to speed on CheckoutSession -- what is the reason for making this weak?

porter-stripe: This memory management story is going to get fleshed out a bit more by a API design @jaynewstrom-stripe is doing. But for now: The weak reference prevents a retain cycle. Checkout already holds a weak var integrationDelegate pointing back to EPE/FC. If EPE/FC held a strong reference to Checkout, neither could be deallocated.

wooj-stripe (approved):

PR #6604, opened 2026-06-16
Await updates before presenting FC sheet

porter-stripe:
## Summary
- If there are updates pending when the user calls presentPaymentOptions on flow controller, we will show a sheet with a loading indicator and await those tasks to finish presenting the sheet.
- If one of the updates fails, the sheet presentation will silently fail and the loading sheet will dismiss. 

#### Demo (using runServerUpdate to run a simulated a 2 second update)

https://github.com/user-attachments/assets/2738bdb5-307d-4d25-8487-7d36e67356e7

## Motivation
- https://docs.google.com/document/d/1CWglcOHR6I8cOWKjztPfnhnVG8axucpv-j5cWIoey7E/edit?tab=t.uzy1hgt1h5jt#bookmark=id.7lu47ty3ex39

## Testing
- Manual, difficult to write unit tests for

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
@@ -511,7 +511,7 @@ extension PaymentSheet {
             }
             presentPaymentOptionsCompletionWithResult = wrappedCompletion
 
-            // If Checkout mutations are in-flight, present a loading sheet and await them.
+            // Mutations in-flight: show loading, await them, then present payment options internally.
             // TODO(porter): Remove assumeIsolated once presentPaymentOptions is @MainActor (blocked on new FC API designs)
             if let checkout, MainActor.assumeIsolated({ !checkout.pendingOperations.isEmpty }) {
                 presentPaymentOptionsAwaitingMutations(
@@ -519,6 +519,7 @@ extension PaymentSheet {
                     checkout: checkout,
                     completion: wrappedCompletion
                 )
+                // Early exit, presentPaymentOptionsAwaitingMutations will present after awaiting mutations
                 return
             }
 
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
@@ -304,6 +304,7 @@ extension PaymentSheet {
 
         private weak var checkout: Checkout?
         private var isPresented = false
+        private var pendingPresentTask: Task<Void, Never>?
         private(set) var didPresentAndContinue: Bool = false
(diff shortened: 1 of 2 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
```diff
@@ -510,6 +511,16 @@ extension PaymentSheet {
             }
             presentPaymentOptionsCompletionWithResult = wrappedCompletion
 
+            // If Checkout mutations are in-flight, present a loading sheet and await them.
+            if let checkout, MainActor.assumeIsolated({ !checkout.pendingOperations.isEmpty }) {
```
gbirch-stripe, line 516: Reading the Apple documentation, it looks like `assumeIsolated` crashes if called from outside of the MainActor. That makes this look like pretty dangerous code — is there some other way we can accomplish this?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
```diff
@@ -539,6 +550,57 @@ extension PaymentSheet {
             showPaymentOptions()
         }
 
+        /// Presents a loading sheet while awaiting in-flight Checkout mutations,
+        /// then swaps to the payment options view controller on success or dismisses on failure.
+        private func presentPaymentOptionsAwaitingMutations(
```
gbirch-stripe, line 557: I'm a bit confused here — it seems that `presentPaymentOptionsAwaitingMutation()` presents the payment options, but both `presentPaymentOptionsAwaitingMutation()` AND `showPaymentOptions()` are called above in `presentPaymentOptions()`. Which one actually ends up being responsible for showing the payment options? What happens to the one that happens second?

porter-stripe: They're mutually exclusive, there's a `return` on line 522 after `presentPaymentOptionsAwaitingMutations` so `showPaymentOptions()` is never reached when mutations are in-flight. The structure is:

1. If mutations in-flight: `presentPaymentOptionsAwaitingMutations` (shows loading, then swaps to payment options on success), return
2. Otherwise: `showPaymentOptions()`

Added a comment to try and clarify, lmk what you think.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
porter-stripe: I agree, assumeIsolated isn't great. But we should be able to fix this soon, @jaynewstrom-stripe is working on some new API designs, and I believe in these new designs we will have the liberty to make presentPaymentOptions and confirm @MainActor, in which case we can remove these. I'll leave a TODO reminder for myself here.

gbirch-stripe (approved):

PR #6607, opened 2026-06-18
Fix Link Inline Signup GB Test Case

jkelle-stripe:
## Summary
Use the accessibility brand label for Link in GB instead of the US brand name.

## Motivation
Fix `testLinkInlineSignup_gb` UI test.


porter-stripe (approved):

sophiah2-stripe (approved):
🙏

PR #6609, opened 2026-06-20
Simplify Checkout session update internals

porter-stripe:
## Summary

Simplifies the Checkout session update internals by consolidating duplicated patterns into a single performUpdate function. Also updates the confirm flow to use runServerUpdate instead of directly calling commitSession.

Before
```swift
  func applyPromotionCode(_ code: String) async throws {
      try requireOpenSession()
      try await enqueueSessionUpdate {
          try await self.performAPIUpdate(.setPromotionCode(code))
      }
  }

  // performAPIUpdate called updateSession, which called setSession
  // setSession stored the session and published state
  // updateSession also notified delegates
  // Two separate functions for one logical operation
```
  After

```swift
  func applyPromotionCode(_ code: String) async throws {
      try await performUpdate(.setPromotionCode(code))
  }

  // performUpdate handles everything:
  // 1. Enqueues on the serial queue
  // 2. Sets .loading state
  // 3. Calls the API
  // 4. Applies local mutations (e.g. address overrides)
  // 5. Calls commitSession (stores session, publishes state, notifies delegates)
  // 6. On error, restores .loaded state
```

Confirm flow change

The confirm flow previously called updateSession directly after getting a confirm response. This is now handled through runServerUpdate, which refreshes the session from our server. This means the confirm update goes through the same serialized queue as everything else.

  What changed

  - setSession + updateSession merged into commitSession: one function that stores, publishes, and notifies
  - performAPIUpdate removed: its logic lives inside performUpdate
  - Loading/error state management lives in one place instead of scattered in enqueueSessionUpdate
  - enqueueSessionUpdate is now a pure serialization primitive with no state side effects

## Motivation
- The old pattern required each caller to manually compose queue enrollment, API calls, and session updates. The new shape makes the happy path and error path explicit in one place.

## Testing
- Existing tests

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Internal.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Internal.swift
@@ -90,19 +90,6 @@ extension Checkout {
 
     // MARK: - Validation
 
-    /// Validates that the session is open and no sheet is presented.
-    @discardableResult
-    func requireOpenSession() throws -> STPCheckoutSession {
-        guard let currentSession = stpSession else {
-            stpAssertionFailure("Expected STPCheckoutSession, got \(type(of: state.session))")
-            throw CheckoutError.apiError(message: "Unexpected session type: expected STPCheckoutSession")
-        }
-        guard currentSession.status?.type == .open else {
-            throw CheckoutError.sessionNotOpen
-        }
-        return currentSession
-    }
-
     func requireSheetNotPresented() throws {
         guard integrationDelegate?.isSheetPresented != true else {
             throw CheckoutError.sheetCurrentlyPresented
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Internal.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Internal.swift
@@ -22,21 +22,6 @@ extension Checkout {
 
     // MARK: - Session Updates
 
-    /// Replaces the current session, preserves client-side overrides, and notifies delegates.
-    ///
-    /// Client-side address overrides are copied from the current session to `newSession`
-    /// automatically. To update an address, set it on `stpSession` before calling this method.
-    func commitSession(_ newSession: STPCheckoutSession) async throws {
-        // Preserve client-side address overrides on the new session.
-        newSession.billingAddress = stpSession?.billingAddress
-        newSession.shippingAddress = stpSession?.shippingAddress
-        stpSession = newSession
-        let publicSession = newSession.makePublicSession()
-        state = pendingOperations.isEmpty ? .loaded(publicSession) : .loading(publicSession)
-        try await integrationDelegate?.checkoutDidUpdate(self)
-        delegate?.checkout(self, didChangeState: state)
-    }
-
     /// Runs `body` as a tracked session update, serialized behind any in-flight ops.
     ///
     /// Operations execute in strict FIFO order: each task waits for the previous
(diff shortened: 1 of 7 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Internal.swift
```diff
@@ -109,20 +135,11 @@ extension Checkout {
...
     }
 
-    /// Validates that the session is open and no sheet is presented.
-    @discardableResult
-    func requireOpenSession() throws -> STPCheckoutSession {
-        guard let currentSession = stpSession else {
-            stpAssertionFailure("Expected STPCheckoutSession, got \(type(of: state.session))")
-            throw CheckoutError.apiError(message: "Unexpected session type: expected STPCheckoutSession")
-        }
-        guard currentSession.status?.type == .open else {
```
porter-stripe, line 104: This is handled by the server, no need to check client-side.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
```diff
@@ -322,7 +297,6 @@ public final class Checkout: ObservableObject {
                 }
                 throw CheckoutError.apiError(message: error.localizedDescription)
             }
-            try self.requireOpenSession()
             try await self.refreshSession()
```
gbirch-stripe: We should also flatten `runServerUpdate` and `refreshSession` like we discussed?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
```diff
@@ -51,7 +51,7 @@ public final class Checkout: ObservableObject {
     /// because they go through `Checkout`'s MainActor-isolated mutation methods.
     /// Requiring full MainActor isolation would propagate `@MainActor` through nearly all of
     /// PaymentSheet's internal types, which is not warranted by the actual concurrency risk.
-    nonisolated(unsafe) private(set) var stpSession: STPCheckoutSession!
+    nonisolated(unsafe) var stpSession: STPCheckoutSession!
```
gbirch-stripe: We are we changing these protection levels?

porter-stripe: There was a function in Checkout+Internal that accessed this (commitSession), but I'm going to move it back to Checkout.swift and revert the protection level change.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
porter-stripe: Yeah, I intentionally left out runServerUpdate. I tried to find a clean way to unify it with performUpdate, but it's unique enough that flattening them together adds more complexity than it removes. A few things that make it different:

  1. It runs a merchant-provided closure
  2. It applies a 20-second timeout (we could apply this to our other mutations too, but currently don't need to)
  3. It has to manually fetch the session afterward, which isn't true for our other updates since updateCheckoutSession returns the session directly

Came to the conclusion that just writing out runServerUpdate inline within the enqueueSessionUpdate is the simplest approach given these differences.

As some compromise I removed refreshSession as it's only used in runServerUpdate and in-lined it.

gbirch-stripe: That all makes sense to me! Inlining `refreshSession` into `runServerUpdate` was all I meant — it makes sense to keep everything separate otherwise

gbirch-stripe (approved):

PR #6610, opened 2026-06-23
Update LinkController.present to return ConfirmationTokens

mats-stripe:


mats-stripe:
No longer what we want to do

PR #6612, opened 2026-06-24
Fix behavior on spm's with no address w/ default payment method on update screen

wooj-stripe:
## Summary
Previously, when a user edited a saved payment method that had no billing details stored, we ignored the merchant-provided defaultBillingDetails from the configuration and instead defaulted the country to the device's locale. This change ensures we respect the merchant's configured defaults before falling back to the device locale.

To reproduce (within payment sheet example)
1. Attach a payment method without any billing address information.
2. Switch device region to United Kingdom (Settings app → General → Language & Region → Region → United Kingdom)
3. Open payment sheet w/ "Default Billing Address" set to off
4. Notice that when adding a new payment method, card form defaults to "United kingdom"
5. Notice that when updating a new payment method, card form defaults to "United kingdom".
6. Close payment sheet.
7.  Open payment sheet w/ "Default billing address" set to `on` (so that default billing address is now set 510 Townsend, SF, 94102, CA, US)
8. Notice that when adding a new payment method, card form defaults to "US"
9. Notice that when updating a new payment method, card form defaults to "UK"

Ultimately, there is an inconsistency between the add screen and the update screen in how we use the default billing address. Note this also impacts CustomerSheet, which was initially reported here:
https://github.com/stripe/stripe-react-native/issues/2493

## Motivation
Updated intended behavior for defaultBillingDetails on the updatePage to be consistent with the add payment method page.

## Testing
Manually tested & added unit tests

## Changelog
Updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@ The next release's version bump will so far be:
 MINOR
 
 ## X.Y.Z - changes pending release
+### PaymentSheet
+* [Fixed] Fixed defaultBillingDetails not being applied when editing a saved payment method that has no billing address.
+
+### CustomerSheet
+* [Fixed] Fixed defaultBillingDetails not being applied when editing a saved payment method that has no billing address.
 
 ### CryptoOnramp (Alpha)
 * [Changed] Renamed `CryptoOnrampCoordinator.presentCRSCARFDeclaration(from:)` to `presentUserAttestation(from:)`, and renamed `CRSCARFDeclarationResult` to `UserAttestationResult`.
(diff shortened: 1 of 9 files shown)


porter-stripe:
I don't think we should be applying `defaultBillingDetails` to the update screen for existing saved PMs. Here's my thinking:

- `defaultBillingDetails` is meant for pre-filling form fields when a user is adding a new payment method. It's a convenience so the merchant can save their customer some typing. It's not a statement about what should appear on every saved PM that happens to not have an address.
- The update screen should reflect what the server actually has for that PM. If the billing address is empty, it should show empty. Pre-filling it with client-side defaults is misleading. The user sees a filled-in address and reasonably thinks "oh, that's what's on my card." They leave without changing anything, and the PM still has no address.
- I kinda think of it this way: a merchant setting `defaultBillingDetails` is saying "I know my customer's info for new PMs." They're not saying "backfill this onto all their existing saved PMs that are missing it.

PR #6619, opened 2026-06-25
[link] Update login copy

lng-stripe:
## Summary
Replace "log in" with "sign in", etc. 

Same as https://github.com/stripe/stripe-android/pull/13292

## Motivation
[RUN_LINK_MOBILE-197](https://go/jira/RUN_LINK_MOBILE-197)

## Testing

CleanShot 2026-07-16 at 12 48 34@2x

CleanShot 2026-07-16 at 12 48 48@2x


StripePaymentSheet/StripePaymentSheet/Source/Categories/String+Localized.swift
```diff
@@ -583,13 +583,18 @@ extension String.Localized {
...
-                "Log out of %@",
-                "Title of the logout action. The placeholder is a Stripe brand name and should not be translated."
-            ),
-            brand.displayName
-        )
+        switch brand {
+        case .link, .unparsable:
+            return STPLocalizedString("Sign out of Link", "Title of the sign-out action.")
+        case .onelink:
+            return String(
```
tillh-stripe: Can't we always use this variant?

lng-stripe: Ah, yeah we can.

tillh-stripe (approved):

PR #6621, opened 2026-06-25
Split state into isLoading and session, update delegate calls

gbirch-stripe:
## Summary
- get rid of unified state
- use isLoading: Bool and session: Checkout.Session
- update isLoading exclusively from the updateQueue code itself
- change from unified didUpdateState delegate call to beginLoad, endLoad, didUpdateSession
- always call delegate functions when we update the isLoading and session

## Motivation
Design doc: https://docs.google.com/document/d/1CWglcOHR6I8cOWKjztPfnhnVG8axucpv-j5cWIoey7E/edit?tab=t.6mxhxefxz2br

## Testing
- tests that verify that the correct delegate and Combine Published emissions happen at the correct times
- test that throwing code in the updateQueue works properly
- test that ensures that there isn't de-duping
- update existing tests to always test delegate calls + Combine Published emissions + that each happen the correct amount


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
```diff
@@ -321,13 +332,10 @@ public final class Checkout: ObservableObject {
...
         newSession.shippingAddress = stpSession?.shippingAddress
         stpSession = newSession
-        let publicSession = newSession.makePublicSession()
-        state = pendingOperations.isEmpty ? .loaded(publicSession) : .loading(publicSession)
         // Skip delegate if another op is queued—it'll notify when it commits.
         if isLastPendingOperation {
             try await integrationDelegate?.checkoutDidUpdate(self)
         }
-        delegate?.checkout(self, didChangeState: state)
+        session = newSession.makePublicSession()
```
porter-stripe: Whats the reason for moving this after the `integrationDelegate?.checkoutDidUpdate(self)`? I see two possible issues:
1. If integrationDelegate?.checkoutDidUpdate throws, session is never set. stpSession is updated but @Published session is stale and out of date w/ the server.
2. Future concern: When MPE is migrated to read from session (the public Checkout.Session) instead of stpSession directly, the integrationDelegate?.checkoutDidUpdate(self) call will trigger MPE's update with the old session value since session hasn't been set yet at that point.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Internal.swift
```diff
@@ -30,17 +30,30 @@ extension Checkout {
     internal func enqueueSessionUpdate(
         _ body: @MainActor @escaping () async throws -> Void
     ) async throws {
+        // Update the loading state to `isLoading = true`.
+        // Avoid triggering the `checkoutDidBeginLoading` delegate if we were already loading.
+        if !isLoading {
```
porter-stripe: Can we move this check into the didSet for isLoading didSet so this logic is preserved if we need to set isLoading from somewhere else?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Internal.swift
```diff
@@ -30,17 +30,30 @@ extension Checkout {
...
             try await body()
         }
 
         pendingOperations.append(operation)
 
         defer {
             pendingOperations.removeAll { $0 == operation }
+
+            // If the queue is now empty, update the loading state to `isLoading = false`.
+            if pendingOperations.isEmpty {
```
porter-stripe: Any reason we can't update isLoading in the didSet for pendingOperations?
```swift
  var pendingOperations: [Task<Void, Error>] = [] {
      didSet {
          isLoading = !pendingOperations.isEmpty
      }
  }
```

porter-stripe (commented):
Can we audit Checkout.swift and Checkout+Internal.swift for stale docs/comments, I see about 5-6 that still reference .loading/.loaded/state.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
gbirch-stripe: I was naively going off of the old delegtae call placement, but I think you're right that we should be setting the session first

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
```diff
@@ -60,7 +71,23 @@ public final class Checkout: ObservableObject {
     let apiClient: STPAPIClient
 
     /// Serial queue of in-flight session updates. Each task waits for the previous task before running.
-    var pendingOperations: [Task<Void, Error>] = []
+    var pendingOperations: [Task<Void, Error>] = [] {
+        didSet {
+            // If the queue has gone from empty to non-empty, we set
+            //  isLoading to true. We avoid setting it if the queue
+            //  was already non-empty to prevent duplicate delegate calls
+            if !pendingOperations.isEmpty && !isLoading {
```
porter-stripe, line 78: The conditional delegate logic (&& !isLoading / && isLoading) should probably live in the didSet of isLoading itself rather than pendingOperations.didSet. pendingOperations shouldn't need to know that isLoading has delegate side-effects, it should just set isLoading =! pendingOperations.isEmpty and let isLoading's own didSet handle whether to fire delegates or not.

porter-stripe (commented):
LGTM, with one minor comment!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
porter-stripe: Maybe like this (double check me!)
```swift
var pendingOperations: [Task<Void, Error>] = [] {
    didSet {
        isLoading = !pendingOperations.isEmpty
    }
}

@Published public internal(set) var isLoading: Bool = false {
    didSet {
        guard oldValue != isLoading else { return }
        isLoading ? delegate?.checkoutDidBeginLoading(self) : delegate?.checkoutDidFinishLoading(self)
    }
}
```

gbirch-stripe: Makes sense!

porter-stripe (commented):
@gbirch-stripe I think my original comment here about stale doc comments still applies: https://github.com/stripe/stripe-ios/pull/6621#pullrequestreview-4582189269

Example: https://github.com/stripe/stripe-ios/blob/1d1b7a875404f2d9f01ad2344d1fa358a1dae68f/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift#L22

Can we do a quick audit? Then looks good to merge.

gbirch-stripe:
> @gbirch-stripe I think my original comment here about stale doc comments still applies: [#6621 (review)](https://github.com/stripe/stripe-ios/pull/6621#pullrequestreview-4582189269)
> 
> Example:
> 
> https://github.com/stripe/stripe-ios/blob/1d1b7a875404f2d9f01ad2344d1fa358a1dae68f/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift#L22
> 
> Can we do a quick audit? Then looks good to merge.

My bad, missed this! Updated all the ones I could find.

Also, I decided to change the de-duping of `isLoading` updates back to the `pendingOperations` `didSet`. With it in the `isLoading` `didSet` it de-dupes the delegate calls but not the Combine emissions. IMO Combine emitting excess changes _shouldn't_ matter if you're using Combine properly, but in case merchants have side effects in their listeners it could cause problems for them, so it seems a bit safer to just not have `isLoading` be set if it's not changing. Open to other thoughts, though.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Checkout/CheckoutUnitTests.swift
```diff
@@ -500,7 +500,7 @@ final class CheckoutUnitTests: XCTestCase {
 
         try await checkout.commitSession(updatedSession)
 
-        XCTAssertEqual(callOrder, ["integration", "regular"])
+        XCTAssertEqual(callOrder, ["regular", "integration"])
```
porter-stripe, line 533: This is an interesting change in behavior due to the didSet usage, I think it's ok, but we did gavel on integration being called first, so MPE is done loading by the time the merchant delegate is called. But I think this is ok, and matches SwiftUI.

porter-stripe (dismissed):

porter-stripe (approved):

PR #6623, opened 2026-06-26
Fix dyld crash on with LinkOpeningTextView

wooj-stripe:
## Summary
Changes LinkOpeningTextView back to `public` from `open`

While checking for regressions, we noticed that the workaround which hooks into the double-tap gesture recognizer was not working for iOS26.

The expected behavior was that:
* Text should not be selectable
* We should not show a system menu
However, we were observing both.

As such, we found a different workaround and verified a fix on both iOS18 and iOS26. 

While testing on iOS16.4, we noticed a minor visual artifact when double tapping the link. To work around this, we could add back the gesture recognizer, but given that traffic to iOS16.4 is quite low, we're opting to accept this minor visual artifact on iOS16.4

## Motivation
May crash when subclassing open classes across SPM modules

## Testing
Manually

## Changelog
Updated.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMMEPromotionTextView.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMMEPromotionTextView.swift
@@ -5,24 +5,28 @@
 //  Created by George Birch on 4/30/26.
 //
 
-@_spi(STP) import StripeUICore
 import UIKit
 
 /// Shared text view used by PMME surfaces that need tappable links without selectable text.
-class PMMEPromotionTextView: LinkOpeningTextView {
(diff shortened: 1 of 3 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentMethodMessagingElement/PMMEPromotionTextView.swift
```diff
@@ -5,24 +5,28 @@
...
+
+    override var isSelectable: Bool {
+        get {
+            return isTextSelectable
+        }
+        set {
+            super.isSelectable = true
+            isTextSelectable = newValue
+        }
+    }
```
gbirch-stripe: Can we simplify this down? We could just hardcode `point()` to have the isTextSelectable = false behavior since that's the only thing we use for PMMEPromotionTextView

gbirch-stripe (approved):

PR #6624, opened 2026-06-29
Fix flag display for special currencies (XAF, XOF, ANG, etc.)

porter-stripe:
## Summary
We were naively grabbing the first two characters of any currency code as the country code for flag display. That works for most currencies (USD -> US, GBP -> GB) but breaks for X-prefixed multi-country currencies like XAF and XOF where there's no single country to show a flag for, and for ANG which should map to NL not the no longer Netherlands Antilles code.

Now regionCode(for:) returns nil for those X-prefixed currencies (so we just don't show a flag) and has an override for ANG -> NL. The flag image prefetcher also handles the nil case gracefully instead of trying to download a flag for a
  nonsense country code.

## Motivation
- AP

## Testing
- New tests

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/CurrencySelectorUtilities.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/CurrencySelectorUtilities.swift
@@ -163,8 +163,31 @@ enum CurrencySelectorUtilities {
 
     // MARK: - Flag emoji
 
+    // Most currency codes already start with the country code (USD→US, GBP→GB) per ISO 4217.
+    // ANG is the one exception among Stripe-supported currencies — it maps to NL, not the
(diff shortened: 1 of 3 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/CurrencySelectorUtilities.swift
gbirch-stripe, line 1: I could definitely see these mappings continuing to change or becoming outdated — have we considered putting it on the server?

porter-stripe: I think it's a reasonable suggestion, although today we have no where to put it. I'll add it to my list a doc I have of good candidates to be driven by the server for the future when we can make changes to the translation layer. I do think this list is quite stable, changes to currencies and such is rare.

gbirch-stripe (approved):

PR #6630, opened 2026-06-29
Derive correct billingDetailsCollectionConfiguration

porter-stripe:
## Summary
-  When a CheckoutSession has billing_address_collection: required, we now upgrade the SDK's billing address collection from .automatic to .full so we always collect the full billing address the server expects. We also disable .never for Checkout (with an assertion + fallback to .automatic) since we don't support collecting billing outside MPE in private preview.

## Motivation
- https://docs.google.com/document/d/1rzyRGIMHO0tYYstS71cYg3ZKyr_LL0ks5EoNT-6Goq4/edit?pli=1&tab=t.5vtzz0k7xzui#heading=h.ka5ya26b6y7y

## Testing
- New unit tests

## Changelog
N/A


porter-stripe:
cc @cttsai-stripe

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/STPCheckoutSession+AddressMerging.swift
```diff
@@ -52,4 +54,21 @@ extension STPCheckoutSession {
...
+    /// `.never` isn't valid for Checkout — fall back to `.automatic`, then upgrade to `.full` if the session requires billing.
+    private func reconcileBillingAddressCollection(
+        on config: inout PaymentSheet.BillingDetailsCollectionConfiguration
+    ) {
+        if config.address == .never {
+            stpAssertionFailure(
+                "billingDetailsCollectionConfiguration.address = .never is not supported"
+                + " with CheckoutSession. Falling back to .automatic."
+            )
+            config.address = .automatic
```
gbirch-stripe: Should we instead throw an integration error at configuration time like other invalid configs?

porter-stripe: Yeah good point. I've moved the assert into PaymentSheetLoader next to the others, and removed this function and moved the remaining code into the function above. Added/removed some tests.

gbirch-stripe (approved):

PR #6634, opened 2026-06-30
Checkout Sessions — Make confirm call on update queue

gbirch-stripe:
## Summary
Put the Embedded and FlowController confirmation calls on the update queue in the checkout session case

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-4619

## Testing
This is unfortunately really hard to test. Ideally we would start a confirmation call, before that is complete add something else to the queue, verify that the second operation hasn't started, finish the confirmation, and then verify that the second operation has started. Unfortunately, though, we don't have a way to ensure that the second operation is queued before the confirmation finishes.


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement+Internal.swift
```diff
@@ -545,17 +539,38 @@ extension EmbeddedPaymentElement {
...
+        let result: PaymentSheetResult
+        let deferredIntentConfirmationType: STPAnalyticsClient.DeferredIntentConfirmationType?
+
+        if let checkout {
+            if !checkout.pendingOperations.isEmpty {
+                let errorMessage = "confirm was called while the Checkout session is still loading. Wait until the Checkout state is .loaded before calling confirm."
+                let error = PaymentSheetError.integrationError(nonPIIDebugDescription: errorMessage)
+                return (.failed(error: error), nil)
+            }
+            var confirmResult: (PaymentSheetResult, STPAnalyticsClient.DeferredIntentConfirmationType?)!
```
porter-stripe: Why the force unwrap "!"? Any way to avoid this?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
```diff
@@ -707,28 +697,49 @@ extension PaymentSheet {
...
+                        assertionFailure("`confirm` should not be called while the Checkout session is loading.")
+                        let error = PaymentSheetError.flowControllerConfirmFailed(
+                            message: "confirmPayment was called while the Checkout session is still loading. Wait until the Checkout state is .loaded before calling confirm."
+                        )
+                        completion(.failed(error: error))
+                        return
+                    }
+                    Task { @MainActor in
+                        await checkout.enqueueSessionUpdate {
+                            confirmBlock()
```
porter-stripe, line 740: Do we need to await this confirmBlock?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Internal.swift
```diff
@@ -46,6 +46,30 @@ extension Checkout {
         try await operation.value
     }
 
+    /// Non-throwing variant of ``enqueueSessionUpdate(_:)-throws``.
+    ///
+    /// Use this when the enqueued work cannot fail. The operation is still
+    /// serialized behind any in-flight ops in the same FIFO order.
+    internal func enqueueSessionUpdate(
```
porter-stripe: Any reason this can't just be 
```swift
internal func enqueueSessionUpdate(_ body: @MainActor @escaping () async -> Void) async {
       try? await enqueueSessionUpdate(body as @MainActor () async throws -> Void)
 }
```

To remove the duplication of code and queue logic?

porter-stripe:
Let's land this before: https://github.com/stripe/stripe-ios/pull/6641

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
gbirch-stripe: No because the confirmBlock works via a completion. Added a comment to explain this but also open to more discussion if you have other thoughts. This code is pretty confusing but I don't immediately see another way around since the rest of this context is completion-based.

porter-stripe (approved):

PR #6635, opened 2026-06-30
Checkout Sessions — Add testing to verify that confirmation calls made while there is work on the queue fails

gbirch-stripe:
## Summary
- change FlowController confirm assertions from normal assertionFailures to stpAssertionFailures
- verify that confirmation calls have the behavior in the title
- start CLAUDE.md file section for acronyms and terms. I'm having Claude automatically update this to add speed and accuracy to human <-> Claude communication

## Motivation

## Testing
the testing is the PR


CLAUDE.md
```diff
@@ -164,3 +164,7 @@ Run `bundle install && bundle exec fastlane stripeios_tests` initially to instal
 
 ### Special Testing Notes
 - Legacy iOS versions: Separate fastlane lanes for iOS 13-16 compatibility testing
+
+### Acronyms and terms
+- EPE = EmbeddedPaymentElement
+- FC = FlowController(PaymentSheet.FlowController)
```
porter-stripe: Nit, this might be confusing for financial connections, I might say PS.FC instead.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
```diff
@@ -302,7 +302,7 @@ extension PaymentSheet {
             }
         }
 
-        private weak var checkout: Checkout?
+        weak var checkout: Checkout?
```
porter-stripe, line 305: Can this retain private(set) at least?

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Checkout/CheckoutPendingOperationsTests.swift
```diff
@@ -163,6 +165,124 @@ final class CheckoutPendingOperationsTests: XCTestCase {
...
+        var configuration = PaymentSheet.Configuration()
+        configuration.customer = .init(id: "cus_test", ephemeralKeySecret: "ek_test")
+        let fc = PaymentSheet.FlowController(
+            configuration: configuration,
+            loadResult: loadResult,
+            analyticsHelper: ._testValue()
+        )
+        fc.checkout = checkout
+
+        STPAssertTestUtil.shouldSuppressNextSTPAlert = true
```
porter-stripe, line 259: Do we need to reset this in between tests?

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Checkout/CheckoutPendingOperationsTests.swift
```diff
@@ -163,6 +165,124 @@ final class CheckoutPendingOperationsTests: XCTestCase {
...
+            default:
+                XCTFail("Expected confirm to fail due to pending operations, got: \(result)")
+            }
+            expectation.fulfill()
+        }
+
+        await fulfillment(of: [expectation], timeout: 2.0)
+
+        XCTAssertTrue(
+            STPAssertTestUtil.lastAssertMessage.contains("Checkout session is loading"),
```
porter-stripe, line 278: Do we need to clear this between tests?

porter-stripe (commented):
I have #6638 up that adds shared checkout test infra. If we land that first, you can rebase and drop most of the boilerplate setup in your new tests. Alternatively we could land this and I can refactor the tests, up to you!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
gbirch-stripe: testEPEConfirmFailsWhenCheckoutPendingOperationsExist and testFCConfirmFailsWhenCheckoutPendingOperationsExist preferably set the checkout directly. There's not a route to initialize FC and EPE with checkout sessions without calling .create() and having it run the API call, which greatly increases the complexity and time cost of these tests. I opted for making FC.checkout internal, which is what EPE already has, though I'm open to further discussion if you think this is a concern

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Checkout/CheckoutPendingOperationsTests.swift
gbirch-stripe: Good thought. This is possible if the code being tested incorrectly doesn't assert. It looks like this is actually an issue throughout our tests that use this functionality, and I've added a solution I tested and works.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Checkout/CheckoutPendingOperationsTests.swift
gbirch-stripe: See comment above

porter-stripe (approved):

porter-stripe:
Merging so I can pull these changes into a related branch!

PR #6636, opened 2026-06-30
Fix currency selector revert bug and clean up utilities

porter-stripe:
## Summary
Fix a bug where the currency selector can't revert on error during the first tap, plus a couple small cleanups in the utilities file.

## Motivation
- AP

## Testing
- Existing

## Changelog
- N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/CurrencySelectorUtilities.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/CurrencySelectorUtilities.swift
@@ -98,6 +98,13 @@ enum CurrencySelectorUtilities {
         return formatExchangeRate(from: meta)
     }
 
+    private static let exchangeRateFormatter: NumberFormatter = {
+        let formatter = NumberFormatter()
+        formatter.minimumFractionDigits = 2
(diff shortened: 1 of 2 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/CurrencySelectorUtilities.swift
```diff
@@ -138,11 +142,7 @@ enum CurrencySelectorUtilities {
     }
 
     private static func formatConversionFeePercent(bps: Int) -> String {
-        let percent = Double(bps) / 100.0
-        if percent.truncatingRemainder(dividingBy: 1) == 0 {
-            return String(format: "%.0f", percent)
-        }
-        return String(format: "%g", percent)
+        String(format: "%g", Double(bps) / 100.0)
```
gbirch-stripe: Should this still be a separate function instead of inlined if it's only used once and is only one line?

porter-stripe: Yeah good point, now that it's been simplified no need for the func.

gbirch-stripe (approved):

PR #6644, opened 2026-06-30
Format GB postal codes for display

toluo-stripe:
Committed-By-Agent: codex

## Summary

Format GB postal codes for display

## Motivation

https://jira.corp.stripe.com/browse/RUN_MOBILESDK-5378

### DIFF
--- a/StripeUICore/StripeUICore/Source/Elements/Factories/TextFieldElement+AddressFactory.swift
+++ b/StripeUICore/StripeUICore/Source/Elements/Factories/TextFieldElement+AddressFactory.swift
@@ -190,6 +190,15 @@ import UIKit
                 return .valid
             }
 
+            func makeDisplayText(for text: String) -> NSAttributedString {
+                if countryCode == "GB", text.count > 3 {
+                    let splitIndex = text.index(text.endIndex, offsetBy: -3)
+                    return NSAttributedString(string: "\(text[..<splitIndex]) \(text[splitIndex...])")
+                }
+
+                return NSAttributedString(string: text)
+            }
+
             func keyboardProperties(for text: String) -> TextFieldElement.KeyboardProperties {
                 // CA and GB use alphanmeric. US uses numeric only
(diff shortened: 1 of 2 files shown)


StripeUICore/StripeUICore/Source/Elements/Factories/TextFieldElement+AddressFactory.swift
```diff
@@ -190,6 +190,15 @@ import UIKit
                 return .valid
             }
 
+            func makeDisplayText(for text: String) -> NSAttributedString {
+                guard countryCode == "GB", text.count > 3 else {
```
joyceqin-stripe: nit: while this works great when only GB has specialized formatting, if we were to add other formatting client-side, I think it'd work better as an if so we could add else ifs for other cases.

StripeUICore/StripeUICore/Source/Elements/Factories/TextFieldElement+AddressFactory.swift
mats-stripe, line 193: Is this something we could use [`CNPostalAddressFormatter`](https://www.swiftbysundell.com/articles/exploring-some-of-the-lesser-known-formatter-types/#addresses) for? Also - is this method being called outside of tests?

toluo-stripe: I don’t think `CNPostalAddressFormatter` works for this case. It formats a full postal address, but it doesn’t
  normalize the postal code itself. I checked `CNPostalAddressFormatter` with a GB address whose `postalCode` is `SW1X7XL`, and it outputs `SW1X7XL`, not `SW1X 7XL`.

And yes, this is used outside tests: `TextFieldElement.viewModel` calls `configuration.makeDisplayText(for:)` when building the attributed text shown in the field.

joyceqin-stripe (approved):

PR #6648, opened 2026-07-01
Make elementsSession non-optional on STPCheckoutSession

porter-stripe:
## Summary
Makes `elementsSession` non-optional on `STPCheckoutSession`. The backend now always returns this field, so we can enforce it at decode time and drop the guard/unwrap in PaymentSheetLoader. Also passes `elements_session_client` on the detach and confirm endpoints so the backend knows to include it.

## Motivation
- CheckoutSessions

## Testing
- Existing tests

## Changelog
N/A


StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Checkout/STPCheckoutSessionTest.swift
```diff
@@ -341,6 +359,7 @@ class STPCheckoutSessionTest: XCTestCase {
             "mode": "payment",
             "payment_status": "unpaid",
             "payment_method_types": ["card"],
+            "elements_session": minimalElementsSession,
```
porter-stripe, line 362: In a follow up PR I'm gonna DRY up all these stubs

gbirch-stripe (approved):

PR #6650, opened 2026-07-01
Checkout Sessions — Fix integration delegate errors reporting as API errors

gbirch-stripe:
## Summary
Previously, `commitSession()` was called in `performUpdate` in a `do` block that rewrites all errors at API errors. This was inaccurate since the errors coming from `commitSession()` are not necessarily API errors. 

Removed `commitSession()` from this block.

## Testing
Added test to enforce this behavior

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Internal.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Internal.swift
@@ -110,6 +110,8 @@ extension Checkout {
             }
 
             localMutation?()
+            /* Commit the session. Note that the above `catch` block still throws, so this
+             * is not reached in that case, which is why we manually notify the delegate. */
             try await self.commitSession(updatedSession)
         }
     }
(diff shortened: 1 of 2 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Internal.swift
```diff
@@ -87,29 +87,32 @@ extension Checkout {
...
-            } catch {
-                // If a prior op skipped the delegate and we're failing before we
-                // get to commitSession ourselves, still notify so the UI updates.
-                if self.isLastPendingOperation {
-                    try? await self.integrationDelegate?.checkoutDidUpdate(self)
+                } catch {
+                    // If a prior op skipped the delegate and we're failing before we
+                    // get to commitSession ourselves, still notify so the UI updates.
+                    if self.isLastPendingOperation {
+                        try? await self.integrationDelegate?.checkoutDidUpdate(self)
```
porter-stripe, line 103: We gaveled on bubbling up these errors, so this breaks the design. The only way the integration delegate throws is if the API returned a session state that the UI can't handle (like zero supported payment methods after an update). Wrapping it as .apiError isn't totally wrong, that's basically what happened. I'm not sure we should distinguish this for the merchant either. "MPE failed to update" doesn't make much sense to them. They should handle an auto update failure the same way they do a standard API failure: retry.

https://docs.google.com/document/d/1CWglcOHR6I8cOWKjztPfnhnVG8axucpv-j5cWIoey7E/edit?tab=t.uzy1hgt1h5jt

gbirch-stripe: Got it, thank you!

PR #6651, opened 2026-07-01
Checkout Sessions — Use public Checkout.Session throughout Checkout, elements, and MPE code

gbirch-stripe:
## Summary
- Rename STPCheckoutSession to STPCheckoutSessionAPIResponse
- Move internal properties, computed properties, and functions used throughout the lifecycle of the session over to Checkout.Session
- Do not store or pass around STPCheckoutSessionAPIResponse after API calls — instead take any immediate actions and then create the public Checkout.Session to store data
- Remove stpSession from Checkout
- Have all Checkout, elements, and MPE code use Checkout's Published `session` property (of type Checkout.Session) instead of stpSession (of type STPCheckoutSession)
- When mutating sessions, we now create a copy instead of editing the properties. This is because we want to keep Checkout.Session's data immutable. We need to make edits on the Checkout.Session rather than the STPCheckoutSessionAPIResponse because when we are making local-only mutations we no longer have access to the STPCheckoutSessionAPIResponse object.
- Move execution of localMutation to commitSession

Strategy for MPE:
During MPE loading, we are already MainActor isolated, and thus can safely access Checkout.session. We use this opportunity to do so and save the reference onto the Intent object, where it can be used throughout MPE. This Checkout.Session has only immutable data, and the reference itself is immutable on the Intent, so we can now safely access it and its data across isolation contexts throughout MPE. 
Ideally, we will eventually MainActor isolate MPE properly, at which point we can drop the Checkout.Session reference off of the Intent and directly access checkout.session.

## Motivation


gbirch-stripe:
> 
> ```diff
> STPCheckoutSessionAPIResponse+Customer.swift:27 warning: Assign-only property 'name' is assigned, but never used
> STPCheckoutSessionAPIResponse+Customer.swift:30 warning: Assign-only property 'phone' is assigned, but never used
> STPCheckoutSessionAPIResponse+SavedPaymentMethodsOfferSave.swift:20 warning: Assign-only property 'status' is assigned, but never used
> STPCheckoutSessionAPIResponse+TaxAmount.swift:21 warning: Assign-only property 'taxableAmount' is assigned, but never used
> STPCheckoutSessionAPIResponse.swift:24 warning: Assign-only property 'clientSecret' is assigned, but never used
> STPCheckoutSessionAPIResponse.swift:83 warning: Assign-only property 'paymentMethodTypes' is assigned, but never used
> STPCheckoutSessionAPIResponse.swift:103 warning: Assign-only property 'cancelUrl' is assigned, but never used
> STPCheckoutSessionAPIResponse.swift:109 warning: Unused property 'totalTaxAmount'
> STPCheckoutSessionAPIResponse.swift:123 warning: Assign-only property 'requiresBillingAddress' is assigned, but never used
> STPCheckoutSessionAPIResponse.swift:130 warning: Unused property 'requiresShippingAddress'
> Checkout.swift:128 warning: Unused initializer 'init(clientSecret:configuration:apiResponse:apiClient:)'
> Checkout.swift:143 warning: Unused initializer 'init(apiResponse:)'
> Checkout.swift:344 warning: Unused parameter 'skipIntegrationNotification'
> ```
> 
> 
> 
> 
> [find-dead-code]

@porter-stripe — these were already present and unused before my changes. Is there a particular reason we're keeping them around?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Intent.swift
```diff
@@ -21,14 +21,17 @@ enum Intent {
     case paymentIntent(STPPaymentIntent)
     case setupIntent(STPSetupIntent)
     case deferredIntent(intentConfig: PaymentSheet.IntentConfiguration)
-    case checkout(Checkout)
+    // TODO(gbirch): Remove Checkout.Session associated value once MPE is MainActor-isolated;
+    // we can then access checkout.session directly. This is a temporary stopgap to provide a
+    // threadsafe version of the checkout session data.
+    case checkout(Checkout, Checkout.Session)
```
porter-stripe: The STPCheckoutSession -> Checkout.Session migration and making the session data immutable (copy-on-write) is great. But I think storing a Checkout.Session snapshot on the Intent puts us in a spot where we now have two sources of truth for the session: `checkout.session` and the `Checkout.Session` on the intent. It's easy to grab the wrong one and hard to audit correctness, and undoes what we set up in #6562 where we specifically moved to passing `Checkout` through the stack so everyone reads from one place.

I know this is working around the fact that `checkout.session` is a `@Published var` on a MainActor-isolated type and we want a stable reference we can use throughout MPE. But the problem it's solving is more theoretical and would take some intentional internal changes to actually become an issue. The compiler can't see it, but we already have guards (not assumptions) that ensure the session can't be mutated while MPE is using it:

- `requireSheetNotPresented()` and `isSheetPresented` on the integration delegate prevent any mutation while the sheet is up
- `awaitPendingOperations` blocks presentation and confirm until in-flight work settles, and confirm fails immediately if `pendingOperations` is non-empty
- EPE is already `@MainActor` and PaymentSheet/FlowController already `assert(Thread.isMainThread)` on every entry point, so in practice the session is never accessed off the main thread

As an alternative, could we just keep `.checkout(Checkout)` and change the existing `nonisolated(unsafe)` property's type from `STPCheckoutSession` to `Checkout.Session`? It gives us the same status quo read without duplicating state on the Intent enum, and the guards already ~guarantee it won't change out from under us. Then once MPE is @MainActor we can delete the nonisolated and just use the `.session` directly.

Also, the PR summary says once MPE is properly MainActor-isolated we can drop the snapshot off the Intent and read `checkout.session` directly. But `checkout.session` is still a `@Published var` that gets reassigned on every write. MainActor isolation just gives us compiler-checked thread safety for something we're mostly enforcing at runtime with the guards. So once we resolve the TODO we're practically back to where we already are today, just with the compiler aware of it.

porter-stripe:
> @porter-stripe — these were already present and unused before my changes. Is there a particular reason we're keeping them around?

The last three are test helpers so can be ignored.

The ones before are probably safe to delete unless they are being currently used in the public API but not consumed by anyone yet. Some seem worth keeping (like below), but maybe some don't. I'll defer to what you think when factoring in the public API!

````
STPCheckoutSessionAPIResponse+Customer.swift:27 warning: Assign-only property 'name' is assigned, but never used
STPCheckoutSessionAPIResponse+Customer.swift:30 warning: Assign-only property 'phone' is assigned, but never used
````

**Comment by gbirch-stripe** [comment]
> > @porter-stripe — these were already present and unused before my changes. Is there a particular reason we're keeping them around?
> 
> The last three are test helpers so can be ignored.
> 
> The ones before are probably safe to delete unless they are being currently used in the public API but not consumed by anyone yet. Some seem worth keeping (like below), but maybe some don't. I'll defer to what you think when factoring in the public API!
> 
> ```
> STPCheckoutSessionAPIResponse+Customer.swift:27 warning: Assign-only property 'name' is assigned, but never used
> STPCheckoutSessionAPIResponse+Customer.swift:30 warning: Assign-only property 'phone' is assigned, but never used
> ```

I lean towards leaving everything for now and then doing a full audit and getting rid of stuff later on since there's still churn for now on what we are and aren't using

File: StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift [hunk 2]
```diff
@@ -289,6 +289,7 @@ extension PaymentSheet {
                     Task { @MainActor in
                         let result = await handleCheckoutSessionConfirmation(
                             checkout: checkout,
+                            session: checkout.nonisolatedSession,
```
porter-stripe: Why pass `checkout` and `session`?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
```diff
@@ -121,41 +122,43 @@ public final class Checkout: ObservableObject {
...
         self.clientSecret = clientSecret
         self.configuration = configuration
         self.apiClient = apiClient
-        await flagImageManager.prefetchFlagImages(for: session)
-        self.stpSession = session
-        self.session = session.makePublicSession()
+        let loadedSession = apiResponse.makePublicSession()
+        await flagImageManager.prefetchFlagImages(for: loadedSession)
+        self.session = loadedSession
+        self.nonisolatedSession = session
```
porter-stripe, line 152: seems like we can remove this given the didSet on self.session.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
gbirch-stripe: Moved this structure back to how it was before!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
porter-stripe: nvm on all these, I think they are all in initalizers.

gbirch-stripe: `didSet` doesn't trigger when initially setting properties in initializers, so in each init we have to manually set `nonisolatedSession`

porter-stripe (approved):

PR #6652, opened 2026-07-02
DRY up CheckoutSession test stubs with shared helpers

porter-stripe:
## Summary
Adds shared builder helpers (`baseSessionJSON`, `makeSession(_:)`, `makeSessionJSON(_:)`) to `CheckoutTestHelpers` and migrates all CheckoutSession test stubs across several files.

## Motivation
- CheckoutSessions

## Testing
- Existing tests

## Changelog
N/A


StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Checkout/CheckoutTestHelpers.swift
```diff
@@ -78,26 +78,52 @@ class CheckoutEmissionRecorder {
...
+    // MARK: - Checkout-flow helpers
+
     static func makeCheckoutWithOpenSession() async -> Checkout {
         let session = makeOpenSession()
         return await Checkout(clientSecret: "cs_test_123_secret_abc", session: session)
     }
 
     static func makeOpenSessionJSON() -> [AnyHashable: Any] {
-        [
+        makeSessionJSON([
```
wooj-stripe: Feels brittle to call makeSessionJSON from a function with no params.  i think `makeOpenSessionJSON`, `makeOpenSession`, `makeClosedSession` are fine to be left as just 'constant' without any need for transforming anything.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Checkout/STPCheckoutSessionTest.swift
```diff
@@ -14,24 +14,9 @@ import XCTest
...
-            "livemode": false,
-            "mode": "payment",
-            "payment_status": "unpaid",
-            "payment_method_types": ["card"],
-            "customer": ["id": "cus_123"],
-            "elements_session": minimalElementsSession,
-        ]
-        overrides.forEach { json[$0.key] = $0.value }
-        return STPCheckoutSession.decodedObject(fromAPIResponse: json)!
+        CheckoutTestHelpers.makeSession(["customer": ["id": "cus_123"]].merging(overrides) { _, new in new })
```
wooj-stripe: Naming here feels awkward -- we have a function called `makeCheckoutSession` calling `makeSession`, which makes it more difficult to reason about when to call makeCheckoutSession vs makeSession

Maybe we use decorator pattern by add an extension in this class like:
```
extension STPCheckoutSession {
    func withCustomerOverride(customerId: String = "cus_123") -> STPCheckoutSession {
...
    }
}
```
That way, users can call
```
makeSession(overrides: overrides)
 .withCustomerOverride()
```

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/SavedPaymentMethodManagerTests.swift
```diff
@@ -286,18 +286,7 @@ extension SavedPaymentMethodManagerTests {
...
-            "mode": "payment",
-            "payment_status": "unpaid",
-            "payment_method_types": ["card"],
-            "elements_session": [
-                "session_id": "es_test",
-                "payment_method_preference": ["ordered_payment_method_types": ["card"]],
-            ],
-        ]
-        return STPCheckoutSession.decodedObject(fromAPIResponse: json)!
+        CheckoutTestHelpers.makeSession(["session_id": id])
```
wooj-stripe: We can also use 'decorator pattern' here by adding an extension on STPCheckoutSession

porter-stripe: Good idea!

wooj-stripe (approved):

PR #6656, opened 2026-07-02
Add EmbeddedPaymentElement option to Checkout playground

porter-stripe:
## Summary
- Adds an integration type picker to the Checkout playground so you can switch between FlowController and the EmbeddedPaymentElement.

## Motivation
- We need this to unblock the writing billing address in the sheet

## Testing
- Manual

## Changelog
N/A

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/Checkout Playground/CheckoutCartEmbeddedPaymentView.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/Checkout Playground/CheckoutCartEmbeddedPaymentView.swift
@@ -1,0 +1,223 @@
+//
+//  CheckoutCartEmbeddedPaymentView.swift
+//  PaymentSheet Example
+//
+//  Created by Nick Porter on 7/2/26.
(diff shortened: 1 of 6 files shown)


Example/PaymentSheet Example/PaymentSheet Example/Checkout Playground/CheckoutCartEmbeddedPaymentView.swift
```diff
@@ -0,0 +1,223 @@
...
+
+    private func load() async {
+        isLoading = true
+        do {
+            var configuration = EmbeddedPaymentElement.Configuration()
+            configuration.returnURL = "payments-example://stripe-redirect"
+            try await viewModel.load(checkout: checkout, configuration: configuration)
+        } catch {
+            loadError = error
+        }
```
wooj-stripe, line 181: For testing, i guess this is fine -- once it errors out, you have to do a full reload, because there's no way to reset loadError.

wooj-stripe (approved):

PR #6678, opened 2026-07-07
Retry UI tests when the test runner hangs

davidme-stripe:
## Summary

We're intermittently seeing UI tests fail with this shape:
> `PaymentSheetUITest-Runner (10939) encountered an error (The test runner hung before establishing connection.)`

I'm not sure *why* this is happening, but it causes the entire test to fail. What's interesting is that we've seen this even in cases where the tests *succeed* by running on a parallel worker.

[Example build](https://app.bitrise.io/app/b220f4ba85d134a7/build/9c470399-d6a6-47e0-bb68-ec6adbfef168?tab=tests&tests_filter_status=failed): all 51 tests passed on `Clone 1 … Runner (10321)`; the build went red solely because a second runner (`10939`) hung before connecting.

## Approach

We've added a script to intercept the failure and determine if it's a real failure or one of these spurious test runner failures. We'll also mark `xcode-test` as skippable, and then add a second copy of the `xcode-test` job.

* If it succeeds, great, we fall through.
* If it fails for reasons related to the tests failing, we fail.
* If it's failing due to this spurious issue, we set `STRIPE_RETRY_XCODE_TEST=yes`, which triggers a new copy of the build job.

## Testing

CI

## Changelog

N/A — CI only.

### DIFF
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -230,11 +230,20 @@ workflows:
     - prep_all
     steps:
     - xcode-test@6:
-        inputs:
+        is_skippable: true
+        inputs: &financial_connections_stability_tests_inputs
         - destination: $DEFAULT_TEST_DEVICE
         - test_repetition_mode: retry_on_failure
         - maximum_test_repetitions: "2"
         - scheme: FinancialConnections Example
+    - script@1:
+        is_always_run: true
+        title: Retry gate for test-runner hang
+        inputs:
+        - content: ./ci_scripts/xcode_test_runner_hang_gate.sh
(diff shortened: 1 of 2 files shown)


mats-stripe (commented):
Could we also do this for `financial-connections-stability-tests`? I've been seeing the same error happen there as well: https://app.bitrise.io/app/b220f4ba85d134a7/build/a37a455b-ba75-4f63-9572-eac97e215a59?tab=tests&tests_filter_status=failed

mats-stripe (approved):
Thank you!

PR #6680, opened 2026-07-07
Update Checkout billing address prior to dismissing MPE

porter-stripe:
## Summary
When the user taps Continue in FlowController or EmbeddedPaymentElement, we now sync the billing address from the form onto the checkout session before dismissing.

- Shows a processing/disabled state while the update is in flight
- On success, dismisses normally
- On failure, re-enables the UI and shows the error inline
- Skips the update entirely for Apple Pay and Link (no billing to sync)
- Only applies to checkout sessions, no-ops for other intent types

Also:
- Made updateBillingAddress internal (no external callers per the product brief)
- Removed the standalone billing address section from the checkout playground (now handled in-sheet)
- Added Apple Pay config + full billing collection defaults to the playground for testing

### EPE Demo

https://github.com/user-attachments/assets/15d7e3f8-e4de-4f0f-bce3-c5b328d6582c

### PS.FC Demo

https://github.com/user-attachments/assets/1cae128b-f318-4d52-b984-1d5661c87344

## Motivation
- CheckoutSessions

## Testing
- Manual

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+BillingAddress.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+BillingAddress.swift
@@ -12,6 +12,8 @@ import Foundation
 extension Checkout {
     /// Syncs the billing address from the payment method's billing details onto the checkout session.
     func syncBillingAddress(from billingDetails: STPPaymentMethodBillingDetails?) async throws {
+        // We need at least a country to build an Address for tax region calculation. Billing details
+        // are optional on payment methods, so it's fine to just skip if we don't have enough info.
         guard let billingDetails,
               let country = billingDetails.address?.country?.nonEmpty else {
             return
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+BillingAddress.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+BillingAddress.swift
@@ -1,39 +1,34 @@
 //
-//  Checkout+PaymentSheet.swift
+//  Checkout+BillingAddress.swift
 //  StripePaymentSheet
 //
 //  Created by Nick Porter on 7/7/26.
 //
 
 import Foundation
+@_spi(STP) import StripeCore
 @_spi(STP) import StripePayments
 
 extension Checkout {
     /// Syncs the billing address from the payment method's billing details onto the checkout session.
     func syncBillingAddress(from billingDetails: STPPaymentMethodBillingDetails?) async throws {
         guard let billingDetails,
-              let country = billingDetails.address?.country, !country.isEmpty else {
+              let country = billingDetails.address?.country?.nonEmpty else {
             return
         }
         let source = billingDetails.address
         let address = Address(
             country: country,
-            line1: source?.line1?.nilIfEmpty,
-            line2: source?.line2?.nilIfEmpty,
-            city: source?.city?.nilIfEmpty,
-            state: source?.state?.nilIfEmpty,
-            postalCode: source?.postalCode?.nilIfEmpty
+            line1: source?.line1?.nonEmpty,
+            line2: source?.line2?.nonEmpty,
+            city: source?.city?.nonEmpty,
+            state: source?.state?.nonEmpty,
+            postalCode: source?.postalCode?.nonEmpty
         )
         try await updateBillingAddress(
             name: billingDetails.name,
             phone: billingDetails.phone,
             address: address
         )
-    }
-}
-
-private extension String {
-    var nilIfEmpty: String? {
-        isEmpty ? nil : self
     }
 }
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+BillingAddress.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+BillingAddress.swift
@@ -30,7 +30,8 @@ extension Checkout {
         try await updateBillingAddress(
             name: billingDetails.name,
             phone: billingDetails.phone,
-            address: address
+            address: address,
+            canUpdateWhileSheetPresented: true
         )
     }
 }
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
@@ -243,7 +243,7 @@ public final class Checkout: ObservableObject {
     ///     to a country-only region, pass a ``Checkout.Address`` with just the country.
     /// - Throws: ``CheckoutError`` if the session is not open, or if
     ///   the server request fails.
-    public func updateBillingAddress(
+    func updateBillingAddress(
         name: String? = nil,
         phone: String? = nil,
         address: Address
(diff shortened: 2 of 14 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+BillingAddress.swift
```diff
@@ -0,0 +1,34 @@
...
+import Foundation
+@_spi(STP) import StripeCore
+@_spi(STP) import StripePayments
+
+extension Checkout {
+    /// Syncs the billing address from the payment method's billing details onto the checkout session.
+    func syncBillingAddress(from billingDetails: STPPaymentMethodBillingDetails?) async throws {
+        guard let billingDetails,
+              let country = billingDetails.address?.country?.nonEmpty else {
+            return
```
gbirch-stripe, line 19: Can we document why this is a no-op rather than an error case? Not clear to me from reading the code

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
```diff
@@ -253,11 +253,11 @@ public final class Checkout: ObservableObject {
         if session.shouldSendTaxRegion(for: "billing") {
             try await performUpdate(.setTaxRegion(address), applying: { session in
                 session.makeCopyOverriding(billingAddress: contactAddress)
-            })
+            }, canUpdateWhileSheetPresented: true)
```
gbirch-stripe: Should `updateBillingAddress` universally be allowed to happen while the sheet is open? What if it is called at a different point while the sheet is displayed other than the timing we call it now? As far as I know we wouldn't correctly handle it being called while displaying the payment method list or a payment method form?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedFormViewController.swift
```diff
@@ -377,15 +377,48 @@ class EmbeddedFormViewController: UIViewController {
...
+            syncCheckoutBillingIfNeeded {
+                self.delegate?.embeddedFormViewControllerDidContinue(self)
+            }
             return
         }
 
         pay(with: selectedPaymentOption)
     }
 
+    private func syncCheckoutBillingIfNeeded(completion: @escaping () -> Void) {
```
gbirch-stripe: Why are we using a completion instead of Swift concurrency?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
```diff
@@ -191,8 +191,9 @@ public final class EmbeddedPaymentElement {
             return result
         }
 
-        // If we currently have a sheet presented fail the update
-        guard !(presentingViewController?.presentedViewController is StripePaymentSheet.BottomSheetViewController) else {
+        // If we currently have a sheet presented fail the update (unless it's a checkout session update, which may occur during billing sync)
+        if !mode.isCheckout,
```
gbirch-stripe, line 195: Is it correct to universally allow updates in the checkout sessions case? Wouldn't this still be an invalid scenario if it weren't 1) a billing address update that's 2) occurring during the "continue" step

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/PaymentSheetVerticalViewController.swift
```diff
@@ -773,6 +775,36 @@ class PaymentSheetVerticalViewController: UIViewController, FlowControllerViewCo
         }
     }
 
+    private func syncCheckoutBillingIfNeeded(completion: @escaping () -> Void) {
```
gbirch-stripe: Same comment as above

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetFlowControllerViewController.swift
```diff
@@ -472,13 +472,53 @@ class PaymentSheetFlowControllerViewController: UIViewController, FlowController
...
                 addPaymentMethodViewController.logBillingAddressCompletionIfNeeded()
-                self.flowControllerDelegate?.flowControllerViewControllerShouldClose(self, didCancel: false)
+                syncCheckoutBillingIfNeeded {
+                    self.flowControllerDelegate?.flowControllerViewControllerShouldClose(self, didCancel: false)
+                }
+            }
+        }
+    }
+
+    private func syncCheckoutBillingIfNeeded(completion: @escaping () -> Void) {
```
gbirch-stripe: Same comment as above

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
```diff
@@ -773,17 +790,8 @@ extension PaymentSheet {
                 return
             }
             assert(Thread.isMainThread, "PaymentSheet.FlowController.update must be called from the main thread.")
-            assert(!isPresented, "PaymentSheet.FlowController.update must be when PaymentSheet is not presented.")
```
gbirch-stripe, line 776: Same here, should we be universally allowing this?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+BillingAddress.swift
porter-stripe: Added!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
porter-stripe: Made this a param on the function. Today it is only ever called from within MPE and it won't be a public API.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedFormViewController.swift
porter-stripe: Good flag. Refactored to remove the completion block entirely. The function is now syncCheckoutBillingThenClose (or ThenContinue in Embedded) which bakes in the success action.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
porter-stripe: Good callout. I looked into scoping this more tightly (e.g. adding an `isDismissing` flag to the VC protocol) but it adds a fair bit of plumbing across both VC types and EPE for limited benefit.

The reason this is safe: the internal `update(checkout:)` is only invoked via the `checkoutDidUpdate` delegate callback, which only fires from `Checkout.commitSession`. And `commitSession` is only reachable while the sheet is presented if `canUpdateWhileSheetPresented: true` was passed, which only happens from `syncBillingAddress`. So the Checkout SDK already gates this to allow updating when we want.

gbirch-stripe: Makes sense!

gbirch-stripe (approved):

PR #6684, opened 2026-07-08
CryptoOnramp SDK: Override PKPaymentRequest MCC Code to 6051

randyliu-stripe:
## Summary

- override PKPaymentRequest.merchantCategoryCode to 6051 

## Motivation

- hide unsupported payment methods when rendering Apple Pay (e.g. Apple Cash)

## Testing

ran unit tests
manually tested on Onramp Example App

Image from iOS (3)

## Changelog

    - [Fixed] Automatically set PKPaymentRequest.merchantCategoryCode so unsupported payment methods are hidden from Apple Pay

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,11 @@ MINOR
 ### StripeConnect
 * [Changed] `PaymentsViewController`, `PaymentsViewControllerDelegate`, `PayoutsViewController`, `PayoutsViewControllerDelegate`, and `EmbeddedComponentManager.PaymentsListDefaultFiltersOptions` are now part of the public API and no longer require `@_spi(PreviewConnect)` to access.
 
-## 26.2.0 2026-07-06
 ### CryptoOnramp (Alpha)
 * [Fixed] Automatically set PKPaymentRequest.merchantCategoryCode so unsupported payment methods are hidden from Apple Pay
+
+## 26.2.0 2026-07-06
+### CryptoOnramp (Alpha)
 * [Added] Added `AppAttestationUnavailableError`, a rich `StripeCryptoOnrampError` surfaced when configuration fails because app attestation is missing or native Link is unavailable.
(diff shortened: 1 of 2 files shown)


StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
```diff
@@ -574,6 +574,9 @@ public final class CryptoOnrampCoordinator: NSObject, CryptoOnrampCoordinatorPro
         case .applePay(let paymentRequest):
             // This presents Apple Pay and promotes the pending payment source on success.
             pendingApplePayPaymentSource = nil
+            if #available(iOS 18.0, *) {
+                paymentRequest.merchantCategoryCode = PKPaymentRequest.MerchantCategoryCode(rawValue: 6051)
```
jeanregisser, line 578: Not sure if there's a real reason a merchant would set this on its own,
but what do you think of setting it only if the merchant did NOT set something for that property already?
An escape hatch if someone ever wants a different value in the future.

jeanregisser (dismissed):

jeanregisser:
We should fill out the CHANGELOG.md unreleased section.
It's used in the release process

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
randyliu-stripe: The only reason I can think of that a merchant might want to override this, would be to specifically test a real card decline. There are other ways for the merchant to do this (e.g. testmode, adding a block on a card) that I don't think it's super valuable. 
I think it is more likely for the merchant to accidentally sent this value and get worse auth rates due to declines.

As we think about adding new MoR (Swapped) and supporting Remittence use cases, I think we should just abstract MCC entirely from the merchant.

randyliu-stripe:
> We should fill out the CHANGELOG.md unreleased section. It's used in the release process

👍 added

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ MINOR
```
jeanregisser, line 8: The changelog entry should be moved under the pending release here.

randyliu-stripe: Fixed

jeanregisser (approved):

PR #6704, opened 2026-07-14
Collect minimum tax billing fields for Checkout Sessions

porter-stripe:
## Summary
- When automatic tax is sourced from the billing address, expand the address collection mode to satisfy country-specific tax requirements (e.g. full address for the US, state/province for Canada).
- Checkout Sessions with automatic tax need enough billing address detail to compute tax. Without this, users in certain countries would submit incomplete billing info and tax calculation would fail.

## Motivation
- CheckoutSessions

## Testing
- New and existing tests

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
```diff
@@ -230,6 +234,43 @@ class PaymentSheetFormFactory {
...
+        let formElement = (form as? PaymentMethodElementWrapper<FormElement>)?.element ?? form
+        guard let section = formElement.getAllUnwrappedSubElements()
+            .compactMap({ $0 as? AddressSectionElement }).first else {
+            return
+        }
+
+        let widenForSelectedCountry: (AddressSectionElement) -> Void = { section in
+            let requirement = CountryTaxRequirement(country: section.selectedCountryCode)
+            let target = section.collectionMode.widened(toSatisfy: requirement)
+            guard section.collectionMode != target else { return }
```
yuki-stripe: Why can't we set the collection mode when we build the AddressSectionElement in `makeBillingAddressSectionIfNecessary`?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
```diff
@@ -230,6 +234,43 @@ class PaymentSheetFormFactory {
...
+            section.delegate = delegate
+        }
+
+        widenForSelectedCountry(section)
+
+        let previousDidUpdate = section.didUpdate
+        section.didUpdate = { [weak section] addressDetails in
+            previousDidUpdate?(addressDetails)
+            guard let section else { return }
+            widenForSelectedCountry(section)
```
yuki-stripe: Why only widen? Seems like a behavior gap that going from e.g. US -> CA stays full address.

porter-stripe: Good flag! I had this implemented but broke it unintentionally in commit #ac775024f6a

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
porter-stripe: The collection mode depends on the *selected* country, and that changes at runtime as the user moves through the country dropdown. `makeBillingAddressSectionIfNecessary` only runs once when we build the form, so it can set the initial mode but can't react to a US -> CA switch. 

The other option is to push this into `AddressSectionElement`, so it recomputes the tax fields per country internally. That gets the collection mode mutation out of the form factory at the cost of baking tax-specific logic into the address element which is shared in StripeUICore.

yuki-stripe: Sure, why can't we set collection mode *and `didUpdate`` when we build the AddressSectionElement? `applyAutomaticTaxFieldCollection` is a bit hacky, grabbing the section and modifying it. It means the other code that creates an AddressSectionElement is incomplete. 

I assume we did it this way out of convenience and there was something tricky about just building the AddressSectionElement correctly, in one place (ok two places - one for LUXE and one for hand-written PMs).

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
```diff
@@ -249,9 +249,13 @@ class PaymentSheetFormFactory {
...
 
-        let widenForSelectedCountry: (AddressSectionElement) -> Void = { section in
+        // Recompute from the base mode every time so we narrow as well as widen (e.g. US -> CA should drop
+        // the full address back down to just the state).
+        let baseCollectionMode = section.collectionMode
+
+        let applyTaxRequirementForSelectedCountry: (AddressSectionElement) -> Void = { section in
             let requirement = CountryTaxRequirement(country: section.selectedCountryCode)
-            let target = section.collectionMode.widened(toSatisfy: requirement)
+            let target = baseCollectionMode.widened(toSatisfy: requirement)
```
yuki-stripe: how does this interact with other 'compute address fields based on country' logic (like in the card form)?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
```diff
@@ -993,12 +1005,15 @@ extension PaymentSheetFormFactory {
...
-            || (configuration.billingDetailsCollectionConfiguration.address == .automatic && requiredByPaymentMethod) {
-            let countries = configuration.billingDetailsCollectionConfiguration.allowedCountries.isEmpty
-                ? nil
-                : Array(configuration.billingDetailsCollectionConfiguration.allowedCountries)
-           return makeBillingAddressSection(countries: countries)
+        let address = configuration.billingDetailsCollectionConfiguration.address
+        let countries = configuration.billingDetailsCollectionConfiguration.allowedCountries.isEmpty
+            ? nil
+            : Array(configuration.billingDetailsCollectionConfiguration.allowedCountries)
+        if address == .full || (address == .automatic && requiredByPaymentMethod) {
```
yuki-stripe: rename requiresPaymentMethod -> fullAddressRequiredByPaymentMethod

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
porter-stripe: I moved this into AddressSectionElement itself. It now takes a `countryCollectionModeOverrides: [String: CollectionMode]` init param and handles swapping to the override internally as the selected country changes. That let me delete `applyAutomaticTaxFieldCollection`, which was grabbing the section after the fact to mutate it and wire up didUpdate, so no more post-hoc mutation.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
porter-stripe: I _think_ the refactor in https://github.com/stripe/stripe-ios/pull/6704/commits/e4aa7e144621deacd3a0afcdc6f442324bc4b169 should address this or make it more clear.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory+FormSpec.swift
```diff
@@ -162,9 +162,15 @@ extension PaymentSheetFormFactory {
...
-                ? makeBillingAddressSection(countries: configuration.billingDetailsCollectionConfiguration.allowedCountriesArray)
-                : nil
+            let address = configuration.billingDetailsCollectionConfiguration.address
+            if address == .full {
+                return makeBillingAddressSection(countries: configuration.billingDetailsCollectionConfiguration.allowedCountriesArray)
+            } else if collectsTaxFromBillingAddress {
+                // Start with country+postal, widen later based on country
+                return makeBillingAddressSection(collectionMode: .countryAndPostal(), countries: configuration.billingDetailsCollectionConfiguration.allowedCountriesArray)
+            } else {
+                return nil
```
yuki-stripe, line 179: I feel like switching on `address` would be clearer and safer. We probably want to still return nil if address == `.never`. 

Also the comment is out of date right, we don't just widen? AddressElement just handles it? In which case does the mode even matter?  Why start with countryAndPostal?

yuki-stripe: I don't understand the collectionMode here too because the correct mode now depends on the country.

yuki-stripe: Maybe the real mode here is 'minimal address fields required for tax, varies per country'.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/CountryTaxRequirement.swift
```diff
@@ -0,0 +1,70 @@
...
+@_spi(STP) import StripeUICore
+
+/// Per-country billing-address collection modes needed to compute tax for Checkout Sessions whose tax
+/// is sourced from the billing address. Countries whose requirement the base mode already satisfies
+/// are omitted so the form falls back to the base mode as the user changes country.
+enum CountryTaxRequirement {
+    private static let requiredModesByCountry: [String: AddressSectionElement.CollectionMode] = [
+        "US": .autoCompletable,
+        "CA": .countryPostalAndState,
+    ]
```
yuki-stripe: **LLM prompt**: roughly, "requiredModesByCountry - do we not need other countries? look at ewcs."
**Response:**

EWCS appears to use shared country lists rather than a two-country inline map:

  - `STRIPE_TAX_COUNTRIES_REQUIRING_FULL_ADDRESS = ['US', 'PR']`
    Evidence: https://stripe.sourcegraphcloud.com/stripe-internal/mint/-/blob/pay-server/stripe-js-v3/src/shared/constants/STRIPE_TAX_COUNTRIES_REQUIRING_FULL_ADDRESS.ts

  - `STRIPE_TAX_POSTAL_CODE_CHECK_COUNTRIES = [...POSTAL_CODE_CHECK_COUNTRIES, 'IN']`, where `POSTAL_CODE_CHECK_COUNTRIES = ['US', 'CA', 'GB', 'PR']`
    Evidence: https://stripe.sourcegraphcloud.com/stripe-internal/mint/-/blob/pay-server/stripe-js-v3/src/shared/constants/STRIPE_TAX_POSTAL_CODE_CHECK_COUNTRIES.ts
    Evidence: https://stripe.sourcegraphcloud.com/stripe-internal/mint/-/blob/pay-server/stripe-js-v3/src/shared/constants/POSTAL_CODE_CHECK_COUNTRIES.ts

  - The API enum docs for `automatic_tax.address_collection_precision=minimal` say it typically collects country plus postal code for `US/CA/IN`.
    Evidence: https://stripe.sourcegraphcloud.com/stripe-internal/mint/-/blob/pay-server/lib/payment_pages/api/api_enum/automatic_tax_address_collection_precision_api_enum.rb

 This looks like it may miss PR and IN, and possibly GB depending which EWCS behavior this should mirror. Also, EWCS minimal behavior seems to describe CA as postal-code collection, while this PR collects province/state via .countryPostalAndState.

  Is the iOS list intentionally narrower, or should this be generated/shared from the same tax country requirements as EWCS?

StripeUICore/StripeUICore/Source/Elements/Factories/Address/AddressSectionElement.swift
```diff
@@ -137,13 +139,19 @@ import UIKit
...
+    /// See the `countryCollectionModeOverrides` init parameter.
+    private let countryCollectionModeOverrides: [String: CollectionMode]
     public var selectedCountryCode: String {
         get {
             return countryCodes[country.selectedIndex]
         }
         set {
             guard let index = countryCodes.firstIndex(of: newValue) else { return }
             country.selectedIndex = index
+            applyCollectionModeOverride(for: newValue)
```
yuki-stripe: Something codex told me - should check this edge case out and make sure it's ok.

One review note: the new country override is applied on normal country changes, but the “billing same as shipping” checkbox path mutates country.selectedIndex directly and then calls updateAddressFields without applyCollectionModeOverride. If automatic-tax billing can combine with that
  checkbox, toggling same-as-shipping to a US/CA default address may leave the previous collection mode in place.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/CountryTaxRequirement.swift
porter-stripe: Thanks for the double check. I was looking at [tax_product.rb origin_address_required_fields](https://stripe.sourcegraphcloud.com/stripe-internal/mint/-/blob/pay-server/lib/globalization/private/data/tax_product.rb?L803-805) which is the merchant origin address, not the customer billing address. I think we should match EWCS here.

StripeUICore/StripeUICore/Source/Elements/Factories/Address/AddressSectionElement.swift
porter-stripe: Ah good edge case, I centralized how we set the country to avoid this. Added a test as well.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory+FormSpec.swift
yuki-stripe: Eh I take this last comment back.

porter-stripe:
@yuki-stripe this is ready for another look, the noCountry edge case will come in a following PR.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/CountryTaxRequirement.swift
```diff
@@ -0,0 +1,22 @@
...
+
+import Foundation
+@_spi(STP) import StripeUICore
+
+/// Billing address fields needed to compute tax from the billing address.
+enum CountryTaxRequirement {
+    /// Collection mode per country; unlisted countries need country only.
+    static let collectionModeByCountry: [String: AddressSectionElement.CollectionMode] = [
+        "US": .autoCompletable,
+        "PR": .autoCompletable,
```
yuki-stripe: I'd think the autocomplete style (.autoCompletable vs the other autocomplete modes) is orthogonal to 'please collect a full billing address for US', and we shouldn't _hardcode_ it to this particular style in here. 

This method should express 'please collect a full billing address for US', and let the autocomplete style be determined elsewhere (no clue what the logic is for the different styles).

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/CountryTaxRequirement.swift
```diff
@@ -0,0 +1,22 @@
...
+        "CA": .countryAndPostal(countriesRequiringPostalCollection: ["CA"]),
+        "GB": .countryAndPostal(countriesRequiringPostalCollection: ["GB"]),
+        "IN": .countryAndPostal(countriesRequiringPostalCollection: ["IN"]),
```
yuki-stripe: The `countriesRequiringPostalCollection: ["CA"]` part of this doesn't make sense to me. "For CA, collect country and postal (and the countries that require postal collection is CA)"? This mode is specific to CA, so it should just be '.country'?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory+Card.swift
```diff
@@ -94,11 +94,21 @@ extension PaymentSheetFormFactory {
                 : Array(configuration.billingDetailsCollectionConfiguration.allowedCountries)
             switch configuration.billingDetailsCollectionConfiguration.address {
             case .automatic:
```
yuki-stripe, line 96: `BDCC = .automatic` + server `billing_address_collection = required` should result in a full billing address. So this code, at face value, looks wrong.

However, we confusingly _overwrite_ BDCC in this case to be `.full`. It think it would be better to not do that, and have a 
```
func resolveBillingCollectionLevel(Checkout.Session, BDCC) -> BDCC? Or actually, maybe this should return a CollectionMode and we can include the logic below in here.
```

sort of function.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory+Card.swift
```diff
@@ -94,11 +94,21 @@ extension PaymentSheetFormFactory {
...
             case .never:
-                return nil
+                // Still collect tax fields if the merchant opted out
+                guard collectsTaxFromBillingAddress else { return nil }
+                return makeBillingAddressSection(
+                    collectionMode: .perCountry(CountryTaxRequirement.collectionModeByCountry),
+                    countries: countries,
+                    includeEmail: shouldIncludeEmail,
+                    includePhone: shouldIncludePhone
+                )
```
yuki-stripe, line 103: `.never` + `collectsTaxFromBillingAddress == true` should not happen, CheckoutElements doesn't support `.never`. We should assert.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
```diff
@@ -992,15 +996,22 @@ extension PaymentSheetFormFactory {
...
+        case .full:
+            return makeBillingAddressSection(countries: countries)
+        case .automatic where fullAddressRequiredByPaymentMethod:
+            return makeBillingAddressSection(countries: countries)
+        case .automatic, .never:
+            guard collectsTaxFromBillingAddress else { return nil }
+            return makeBillingAddressSection(
+                collectionMode: .perCountry(CountryTaxRequirement.collectionModeByCountry),
+                countries: countries
+            )
```
yuki-stripe: same here, assert if never + collectsTaxFromBillingAddress

StripeUICore/StripeUICore/Source/Elements/Factories/Address/AddressSectionElement.swift
```diff
@@ -137,16 +140,15 @@ import UIKit
             }
         }
     }
+    /// Collection mode passed at init. Restored on country change when using ``CollectionMode/perCountry(_:)``.
+    private let baseCollectionMode: CollectionMode
```
yuki-stripe: This tells me that 'per country' mode is _not_ the same as CollectionMode.

I think it should be something like:
```
// The fields to collect
enum FieldsToCollect {
  case country
  case countryAndPostal
  case all
}

// Autocomplete style
enum AutocompleteStyle {
  // all the different styles, only applies when we collect all fields
}

// === Initializer  ===
// If the selected country in the AddressSectionElement is not in countryToFieldsOverrides, uses `defaultFieldsToCollect`
AddresssSectionElement(defaultFieldsToCollect: FieldsToCollect, countryToFieldsOverrides: [String: FieldsToCollect] = [:]) 

// === Examples ===
// PI Cards: 
AddresssSectionElement(defaultFieldsToCollect: .country, countryToFieldsMap: [US: .countryAndPostal, ... ])

// CS Cards + Tax billing 
AddresssSectionElement(defaultFieldsToCollect: .country, countryToFieldsMap: [US: .all, ... ])
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory+Card.swift
yuki-stripe: I'm preserving this confusing overwrite behavior in my own PR too that adds the Checkout default billing address, so maybe we should punt on this for now to avoid merge conflicts.

StripeUICore/StripeUICore/Source/Elements/Factories/Address/AddressSectionElement.swift
yuki-stripe: This is probably better actually, since autocomplete only applies to all fields:

```
// The fields to collect
enum FieldsToCollect {
  case country
  case countryAndPostal
  case all(autocompleteStyle: AutocompleteStyle)
}

struct AutocompleteStyle {...whatever we need }

// === Initializer  ===
// If the selected country in the AddressSectionElement is not in countryToFieldsOverrides, uses `defaultFieldsToCollect`
AddresssSectionElement(defaultFieldsToCollect: FieldsToCollect, countryToFieldsOverrides: [String: FieldsToCollect] = [:]) 

// === Examples ===
// PI Cards: 
AddresssSectionElement(defaultFieldsToCollect: .country, countryToFieldsMap: [US: .countryAndPostal, ... ])

// CS Cards + Tax billing 
AddresssSectionElement(defaultFieldsToCollect: .country, countryToFieldsMap: [US: .all, ... ])
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory+Card.swift
porter-stripe: Isn't this handled here? https://github.com/stripe/stripe-ios/blob/master/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/STPCheckoutSessionAPIResponse%2BAddressMerging.swift#L25

yuki-stripe: That is where we do the confusing overwrite.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/CountryTaxRequirement.swift
porter-stripe: Adopted the FieldsToCollect which should resolve this.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/CountryTaxRequirement.swift
porter-stripe: Adopted the FieldsToCollect which should resolve this.

StripeUICore/StripeUICore/Source/Elements/Factories/Address/AddressSectionElement.swift
porter-stripe: Adopted the `FieldsToCollect` enum and the per-country overrides map and removed `indirect case perCountry`. And now, `.all` no longer carries an autocomplete style, the element resolves it and picks the style.

I didn't end up going with `defaultFieldsToCollect` or `AutocompleteStyle`, each feels like it needs more thought / its own refactor. Mapping `FieldsToCollect` to a `CollectionMode` seemed about equivalent here.

porter-stripe:
@yuki-stripe also merged https://github.com/stripe/stripe-ios/pull/6727 which collapsed the auto complete cases into one single case.

Also worth noting in this PR previously we were checking the BDCC API value many times, but not that is mostly centralized, e.g were checking this in quite a few spots. `configuration.billingDetailsCollectionConfiguration.address == .full`

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
```diff
@@ -968,14 +982,27 @@ extension PaymentSheetFormFactory {
...
-            || (configuration.billingDetailsCollectionConfiguration.address == .automatic && requiredByPaymentMethod) {
-            let countries = configuration.billingDetailsCollectionConfiguration.allowedCountries.isEmpty
-                ? nil
-                : Array(configuration.billingDetailsCollectionConfiguration.allowedCountries)
-           return makeBillingAddressSection(countries: countries)
-        } else {
+    /// Builds the billing address section, or nil if there shouldn't be one.
+    /// - Parameters:
+    ///   - fullAddressRequiredByPaymentMethod: Whether the payment method requires the full address for `.automatic`
+    ///     collection (independent of tax). Tax can still force country collection.
```
yuki-stripe, line 988:
```suggestion
    ///   - fullAddressRequiredByPaymentMethod: Whether the payment method requires a full address.
```

StripeUICore/StripeUICore/Source/Elements/Factories/Address/AddressSectionElement.swift
```diff
@@ -180,6 +192,8 @@ import UIKit
        - locale: Locale used to generate the display names for each country
        - addressSpecProvider: Determines the list of address fields to display for a selected country
        - defaults: Default address to prepopulate address fields with
+       - collectionMode: Which address fields to collect
```
yuki-stripe, line 195: I see we've kept `CollectionMode`. Why wouldn't collectionMode's type be FieldsToCollect, when it is literally which fields to collect?

PR #6705, opened 2026-07-14
Implement PaymentElement UIKit view

yuki-stripe:
## Summary
Implements PaymentElementUIView

## Motivation
CheckoutSessions

## Testing
Coming later - will probably have to make bugfixes here.

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
@@ -190,7 +190,7 @@ public final class EmbeddedPaymentElement {
             return result
         }
 
-        // If we currently have a sheet presented fail the update (unless it's a checkout session update, which may occur during billing sync)
+        // If we currently have a sheet presented, fail the update (unless it's a checkout session update, which may occur during billing sync)
         if !mode.isCheckout,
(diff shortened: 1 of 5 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
```diff
@@ -473,6 +474,30 @@ extension EmbeddedPaymentElement {
     var isPresentingPaymentUI: Bool {
         return presentingViewController?.presentedViewController is BottomSheetViewController
     }
+
+    /// Returns the explicitly configured presenting view controller or tries to find one if nil.
+    var resolvedPresentingViewController: UIViewController? {
```
porter-stripe, line 479: I feel like we already have some logic for this we could reuse, somewhere in the Embedded SwiftUI code?

yuki-stripe: The code that finds the VC is reused (`UIWindow.visibleViewController`).

yuki-stripe: not sure we can reasonably share the other parts of this method with Embedded.

porter-stripe (approved):

PR #6706, opened 2026-07-14
Implement PaymentElement SwiftUI view

yuki-stripe:
## Summary
Implements SwiftUI view. Borrows from Embedded. 

## Motivation
CheckoutSessions

## Testing
Added one test for height change updates, rest coming later. 

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Payment Element/PaymentElement.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Payment Element/PaymentElement.swift
@@ -80,8 +80,9 @@ public final class PaymentElement {
             checkout: checkout,
             configuration: embeddedConfiguration
         )
+        self.embeddedPaymentElement.notifiesDelegateOnInitialHeight = true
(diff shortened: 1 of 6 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Payment Element/PaymentElement.swift
```diff
@@ -81,8 +80,10 @@ public final class PaymentElement {
             checkout: checkout,
             configuration: embeddedConfiguration
         )
-        self.view = PaymentElementView()
-        self.uiView = PaymentElementUIView(contentView: embeddedPaymentElement.view)
+        self.embeddedPaymentElement.notifiesDelegateOnInitialHeight = true
```
yuki-stripe, line 83: this `notifiesDelegateOnInitialHeight` thing fixes an issue with swiftui embedded where it was having trouble getting a height. Turns out embedded's first layout calculates its height correctly but then *doesn't* inform the delegate, and so SwiftUI can't update its height.  This flag fixes that.

porter-stripe (approved):

PR #6708, opened 2026-07-14
Sync apple pay billing address for tax

porter-stripe:
## Summary
- Updates the billing address when Apple Pay is presented and when a new PM or the billing address changes
- Tax is reflected in the line items
- We do not reset the billing address to what it was prior to Apple Pay launching if a user closes it because Apple Pay remains selected, so the billing will be updated whenever the user selects a new payment method type and provides billing for that payment method.

#### Demo

https://github.com/user-attachments/assets/da4773bd-bf0b-43ea-8b3b-73fcbb1b758c

## Motivation
- CheckoutSessions

## Testing
- Unit test
- Manual

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/STPApplePayContext+PaymentSheet.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/STPApplePayContext+PaymentSheet.swift
@@ -193,23 +193,21 @@ private class ApplePayContextClosureDelegate: NSObject, ApplePayContextDelegate 
         paymentInformation: PKPayment,
         context: STPApplePayContext
     ) async throws -> String {
-        // Sync the full billing address (now available post-authorization) so the server
-        // computes final tax before we read expectedAmount.
-        if let stpPaymentMethod = STPPaymentMethod.decodedObject(fromAPIResponse: paymentMethod.allResponseFields) {
-            try await checkout.syncBillingAddress(from: stpPaymentMethod.billingDetails)
-        }
-
         let checkoutSession: Checkout.Session = checkout.nonisolatedSession
 
+        // 1. Build client attribution metadata
         let clientAttributionMetadata = STPClientAttributionMetadata.makeClientAttributionMetadata(
             intent: intent,
             elementsSession: elementsSession
         )
 
+        // 2. Get expected amount from checkout session
         let expectedAmount = checkoutSession.expectedAmount()
 
+        // 3. Extract shipping details from PKPayment (if provided)
         let shipping = makeShippingDetailsParams(from: paymentInformation)
 
+        // 4. Call confirm API with the Apple Pay payment method
         let response = try await context.apiClient.confirmCheckoutSession(
             sessionId: checkoutSession.id,
             paymentMethod: paymentMethod.id,
@@ -221,7 +219,10 @@ private class ApplePayContextClosureDelegate: NSObject, ApplePayContextDelegate 
             clientAttributionMetadata: clientAttributionMetadata
         )
 
+        // 5. Update the Checkout instance with the confirmed session response
         try await checkout.commitSession(response)
+
+        // 6. Return client secret based on checkout session mode
         return try response.intentClientSecret()
     }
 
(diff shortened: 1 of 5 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/STPApplePayContext+PaymentSheet.swift
```diff
@@ -305,6 +312,19 @@ private class ApplePayContextClosureDelegate: NSObject, ApplePayContextDelegate
                 handler(PKPaymentRequestShippingContactUpdate())
             }
         }
+
+    func applePayContext(
+        _ context: STPApplePayContext,
+        didSelectPaymentMethod paymentMethod: PKPaymentMethod,
```
yuki-stripe, line 318: > We do not reset the billing address to what it was prior to Apple Pay launching if a user closes it because Apple Pay remains selected, so the billing will be updated whenever the user selects a new payment method type and provides billing for that payment method.

why not?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/STPApplePayContext+PaymentSheet.swift
```diff
@@ -305,6 +312,19 @@ private class ApplePayContextClosureDelegate: NSObject, ApplePayContextDelegate
...
+        handler: @escaping (PKPaymentRequestPaymentMethodUpdate) -> Void) {
+            if let paymentMethodUpdateHandler {
+                paymentMethodUpdateHandler(paymentMethod) { result in
+                    handler(result)
+                }
+            } else {
+                handler(PKPaymentRequestPaymentMethodUpdate(paymentSummaryItems: []))
```
yuki-stripe, line 327: just want to double check, it's ok to pass `PKPaymentRequestPaymentMethodUpdate(paymentSummaryItems: [])` here? In particular the empty paymentSummaryItems? That won't e.g. override the existing summary items?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/STPApplePayContext+PaymentSheet.swift
porter-stripe: When looking at the UX I feel like it doesn't make sense to revert b/c Apple Pay remains selected even if you close the sheet until you choose something else. Seems fine to leave the billing as-is, lmk what you think.

https://github.com/user-attachments/assets/31fd2b3d-ba36-47c1-8447-f571ab128715

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/STPApplePayContext+PaymentSheet.swift
porter-stripe: This code is unreachable in practice, this is following a similar pattern to what we did in `didSelectShippingContact` and `didSelect`

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/STPApplePayContext+PaymentSheet.swift
yuki-stripe: I'm peaceful.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/STPApplePayContext+PaymentSheet.swift
yuki-stripe: Ah ok, can we add a comment similar to those too then? `        // Note: this method isn't called unless our delegate implements it (see this class's `responds(to:)` override)
`

porter-stripe: Added!

yuki-stripe: sorry can we also stpassert here with that same comment too?

porter-stripe: Added!

yuki-stripe (dismissed):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/STPApplePayContext+PaymentSheet.swift
porter-stripe: @yuki-stripe I had to add a `responds(to:)` since we have the assert. The others no-op so I guess they haven't needed it.

yuki-stripe (approved):

PR #6709, opened 2026-07-14
Add Checkout payment option state

yuki-stripe:
## Summary
  Adds `Checkout.Session.paymentOption` and `Checkout.clearPaymentOption()` for Checkout PaymentElement integrations.

PaymentElement now mirrors selected payment option display data into the Checkout session from EmbeddedPaymentElement and FlowController updates. 

`clearPaymentOption()` + FC is not supported yet. 
\

  ## Testing
`ci_scripts/run_tests.rb --test StripePaymentSheetTests/CheckoutUnitTests/testSessionPaymentOptionUpdatesAndClears`

  ## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Internal.swift
```diff
@@ -20,6 +20,15 @@ extension Checkout {
         try await performUpdate(.setCurrency(currency))
     }
 
+    // MARK: - Payment Option
+
+    func setPaymentOption(_ paymentOption: Session.PaymentOptionDisplayData?) {
+        guard session.paymentOption != paymentOption else {
```
porter-stripe: Does this guard ever fire? I think we need to compare the pngData() of the payment option. We might need to add a "==" override rather than just declare the Equatable conformance. Similar to EPE.PaymentOptionDisplayData.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Payment Element/PaymentElement.swift
```diff
@@ -85,6 +88,18 @@ public final class PaymentElement {
         self.view = PaymentElementView(viewModel: PaymentElementViewModel(uiView: uiView))
         self.uiView = uiView
         self.embeddedPaymentElement.delegate = self
+        self.checkout = checkout
+        self.paymentSheetFlowController.$paymentOption
+            // Without dropFirst, @Published immediately emits FlowController's current value before we setPaymentOption from EmbeddedPaymentElement below.
+            .dropFirst()
+            .sink { [weak self] paymentOption in
+                self?.checkout?.setPaymentOption(paymentOption.map(Checkout.Session.PaymentOptionDisplayData.init))
```
porter-stripe: I'm having a hard time reasoning about having both of these elements hooked up. But does this override EPE's selection when the session is updated? When the session is updated we call PS.FC.update(checkout), which calls updatePaymentOption() triggering this sink.

Thinking:
user selects card in EPE -> session.paymentOption = card
any session update  -> PS.FC.update(checkout) -> emits nil (PS.FC has no selection)
sink -> setPaymentOption(nil) -> session.paymentOption is cleared

EPE would still have card selected, so I think they'd be out of sync

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Internal.swift
yuki-stripe: Good point. Do we even want to dedupe 'didUpdateSession' / the publisher of session? Or is it okay if it fires sometimes w/o changes? 

If we do want to dedupe and enforce a "only notified when session is actually mutated" rule, I think we want to make Session Equatable to do it.

IMO It's not clear atm that we need to dedupe. The only reason I can think of is as a minor optimization to reduce amount of UI updates, so I'm leaning towards removing this guard and not worrying about it.

porter-stripe: I think it's fine not to de-dupe.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Internal.swift
```diff
@@ -20,6 +20,15 @@ extension Checkout {
...
+            return
+        }
+        session = session.makeCopyOverriding(paymentOption: .newValue(paymentOption))
```
yuki-stripe: A big problem I'm realizing - AFAIU we should put *all* session mutations on the queue. If that's true (and that's what our eng design says) then we should structure the code to make it harder to make this mistake.

porter-stripe: Yes, that's right,  all mutations should go through the queue. Previously the only way to mutate a session (outside init) was via the mutation APIs: every write funneled through a `enqueueSessionUpdate` operation. `setPaymentOption` is the first thing to write `session` directly, off-queue AFAICT.

idk how to enforce no one mutates `session` off-queue, but it'd be nice to make it hard to do by accident. Right now the queue is just a convention. Any ideas?

porter-stripe: Some random more thoughts here: this is somewhat a significant change b/c now you may need to batch numerous mutations into one single operation, where previously it was one single operation per queue task.

before:
updateX -> payment_pages/update -> update session -> update elements
now:
updateX -> payment_pages/update -> update session -> update MPE -> back propagate changes to session that MPE may have made -> update elements

That back-propagation step is really us breaking unidirectional data flow. Before, session was the single source of truth: it flowed one direction down into the elements, and the only way back up was through the queue. Now the element is both a consumer and a producer of session state, so we get a loop. 

It makes me wonder if paymentOption being on Checkout is correct, it's the only property (as of now) that has this issue. Makes me wonder if CheckoutController should just expose a .confirm() and not a paymentOption and instead leave that on payment element. Also I'm guessing this pattern won't scale well if other elements decide to start mutating the session upon updating too.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Payment Element/PaymentElement.swift
yuki-stripe: 💀💀💀  ooof yep, really great call. I fixed it in the latest commit.

porter-stripe (dismissed):

porter-stripe (approved):

PR #6710, opened 2026-07-14
Point CLAUDE.md to AGENTS.md

yuki-stripe:
## Summary
  - 0ae6f347ddd CLAUDE.md imports AGENTS.md so there's one source of truth. Claude supports importing AGENTS.md. 
  - 8e7666aa8d2 adds the Given/When/Then test guidance to AGENTS.md

.


porter-stripe (dismissed):

davidme-stripe (dismissed):
I do think it's useful to keep these separate in case different hints are helpful for different models, but that hasn't really been the case for the last few GPT/Claude models. So I guess we can combine these for now.

yuki-stripe:
I moved some stuff that was newer/better in CLAUDE.md over (according to codex).

> I do think it's useful to keep these separate in case different hints are helpful for different models, but that hasn't really been the case for the last few GPT/Claude models. So I guess we can combine these for now.

Yeah we can still add claude specific stuff to CLAUDE.md afaiu, not sure about codex-specific stuff(codex says we could use `.codex/config.toml`).

porter-stripe (approved):

PR #6713, opened 2026-07-14
Fix FC interaction bug

porter-stripe:
## Summary
- After filling out details and returning into the FlowController sheet, it showed up in a bad, non-interactive state.
- Interaction was never re-enabled on the VC before closing
- DRY'd up EPE's version of this too

## Motivation
- CheckoutSessions bug fix

## Testing
- Manual

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedFormViewController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedFormViewController.swift
@@ -406,15 +406,16 @@ class EmbeddedFormViewController: UIViewController {
             guard let self else { return }
             do {
                 try await checkout.syncBillingAddress(from: paymentOption.billingDetails)
-                self.isPaymentInFlight = false
-                self.isUserInteractionEnabled = true
-                self.delegate?.embeddedFormViewControllerDidContinue(self)
             } catch {
-                self.isPaymentInFlight = false
-                self.isUserInteractionEnabled = true
                 self.error = error
-                self.updateError()
-                self.updatePrimaryButton()
+            }
+            self.isPaymentInFlight = false
+            self.isUserInteractionEnabled = true
+            self.updateError()
+            self.updatePrimaryButton()
+
+            if self.error == nil {
+                self.delegate?.embeddedFormViewControllerDidContinue(self)
             }
         }
     }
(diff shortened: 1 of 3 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedFormViewController.swift
```diff
@@ -406,15 +406,16 @@ class EmbeddedFormViewController: UIViewController {
...
-                self.updatePrimaryButton()
+            }
+            self.isPaymentInFlight = false
+            self.isUserInteractionEnabled = true
+            self.updateError()
+            self.updatePrimaryButton()
+
+            if self.error == nil {
+                self.delegate?.embeddedFormViewControllerDidContinue(self)
             }
```
wooj-stripe, line 419: I guess this is somewhat safe since it's all in the main actor, but i don't like that that we're having to access self.error, and feels very 'side-effecty' without too much expressiveness... Maybe it would be better to:
```
do {
    try await checkout.syncBillingAddress(from: paymentOption.billingDetails)
    defer {
        self.delegate?.embeddedFormViewControllerDidContinue(self)
    }
}catch{
}
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetFlowControllerViewController.swift
```diff
@@ -494,32 +494,38 @@ class PaymentSheetFlowControllerViewController: UIViewController, FlowController
...
+
+            if let error = self.error {
                 self.errorLabel.text = error.nonGenericDescription
                 UIView.animate(withDuration: PaymentSheetUI.defaultAnimationDuration) {
                     self.errorLabel.setHiddenIfNecessary(false)
                 }
-                self.updateButton()
+            } else {
+                self.flowControllerDelegate?.flowControllerViewControllerShouldClose(self, didCancel: false)
             }
```
wooj-stripe, line 517: seems like we can use defer blocks here too?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedFormViewController.swift
porter-stripe: I tried the defer version, but Swift warns: `'defer' statement at end of scope always executes immediately; replace with 'do' statement to silence this warning`. It also means the continue callback would run before we reset the in-flight/UI state, so I’m going to keep as-is. Lmk what you think!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetFlowControllerViewController.swift
porter-stripe: I don't mind the branching on this one as much, I do think defer in the above case would be nice though.

wooj-stripe (approved):

PR #6716, opened 2026-07-15
Return to login screen after logging out of Link wallet in standalone flow

mats-stripe:
Resolves https://jira.corp.stripe.com/browse/RUN_LINK_MOBILE-212

## Summary

Updates Link Standalone logout behavior of returning to the login screen rather than dismissing the flow entirely. When the Link wallet is presented via MPE, logging out returns back to MPE, but in this integration there is no MPE to return to. This fixes that.

This integration is captured by `context.shouldShowSecondaryCta == false`, but that name wasn't really accurate. I renamed it to `canPayAnotherWay`.

## Motivation

https://jira.corp.stripe.com/browse/RUN_LINK_MOBILE-212

## Testing

Before:

https://github.com/user-attachments/assets/a18fe23e-b398-4ac3-8661-1b36900127b2

After:

https://github.com/user-attachments/assets/5a96c66f-a0ec-4bf7-bfc1-33411f2c5e03

MPE -> Link Wallet behavior of returning to MPE is unaffected:

https://github.com/user-attachments/assets/b4bb14d4-43f0-4b31-aa3d-0d41f289bc92

## Changelog

N/a

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController.swift
@@ -85,7 +85,7 @@ final class PayWithLinkViewController: BottomSheetViewController {
         var linkBrand: LinkBrand
         let shouldOfferApplePay: Bool
         let shouldFinishOnClose: Bool
-        let shouldShowSecondaryCta: Bool
+        let canPayAnotherWay: Bool
         let launchedFromFlowController: Bool
         let initiallySelectedPaymentDetailsID: String?
         let callToAction: ConfirmButton.CallToActionType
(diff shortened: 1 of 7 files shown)


StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController.swift
```diff
@@ -85,7 +85,7 @@ final class PayWithLinkViewController: BottomSheetViewController {
         var linkBrand: LinkBrand
         let shouldOfferApplePay: Bool
         let shouldFinishOnClose: Bool
-        let shouldShowSecondaryCta: Bool
+        let canPayAnotherWay: Bool
```
tillh-stripe: Nit: This is also true for non-payment cases, so we could name this `canContinueWithoutLink`?

tillh-stripe (dismissed):

StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController.swift
mats-stripe: I named it this way because it also determines whether the `Pay another way` button is shown - I'll leave it as-is  if you don't have a strong opinion!

tillh-stripe: This is what I mean: https://github.com/stripe/stripe-ios/blob/master/StripePaymentSheet/StripePaymentSheet/Source/Internal/Link/Controllers/PayWithLinkViewController.swift#L100

mats-stripe: Done in https://github.com/stripe/stripe-ios/pull/6716/commits/e619b8d36758c875041eaaf90e3412784123825e

tillh-stripe (approved):

PR #6717, opened 2026-07-15
Fall back to saved payment method billing details

porter-stripe:
## Summary
- `PaymentOption.billingDetails` now falls back to a saved payment method's own billing details. Previously the `.saved` case only read `confirmParams`, which is nil for saved payment methods, so it returned nil.
- When a customer selects a saved payment method, its billing address was dropped. That address feeds the checkout billing update, so the checkout session never received it and tax calculation was wrong.
- Only impacts CheckoutSessions

## Motivation
- CheckoutSessions

## Testing
- New unit test

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
@@ -138,8 +138,8 @@ extension PaymentSheet {
             switch self {
             case .new(let confirmParams):
                 return confirmParams.paymentMethodParams.billingDetails
-            case .saved(_, let confirmParams):
-                return confirmParams?.paymentMethodParams.billingDetails
+            case .saved(let paymentMethod, let confirmParams):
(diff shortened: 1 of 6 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
```diff
@@ -138,8 +138,8 @@ extension PaymentSheet {
             switch self {
             case .new(let confirmParams):
                 return confirmParams.paymentMethodParams.billingDetails
-            case .saved(_, let confirmParams):
-                return confirmParams?.paymentMethodParams.billingDetails
+            case .saved(let paymentMethod, let confirmParams):
+                return confirmParams?.paymentMethodParams.billingDetails ?? paymentMethod.billingDetails
```
wooj-stripe: Description says that this "Only impacts CheckoutSessions", but this seems to be in the FlowController class?

porter-stripe: Yeah that's confusing, I moved it to it's own file!

wooj-stripe (approved):

PR #6718, opened 2026-07-15
Add defaults.billingAddress to Checkout configuration

yuki-stripe:
## Summary
- Add `Checkout.Configuration.defaults.billingAddress` public API
- Route default billing details into the PaymentElement configuration and Apple Pay. Note Link consumes the MPE config, so it gets it for free.
- On Checkout init, call `updateBillingAddress`.
  - Note: The initial payment_pages /init request does not support being passed defaults.

## Follow-ups
- The other defaults: `shippingDetails`, `phone`, and `email`

## Testing
See tests.


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Embedded/EmbeddedPaymentElement.swift
```diff
@@ -112,23 +112,21 @@ public final class EmbeddedPaymentElement {
         configuration: Configuration
     ) async throws -> EmbeddedPaymentElement {
         try await checkout.awaitPendingOperations()
-        var config = configuration
-        checkout.session.applyAddressOverrides(to: &config)
```
porter-stripe, line 116: With these `applyAddressOverrides` calls removed, how do local Checkout address updates make it back into PaymentElement?

Today, `checkout.updateBillingAddress(...)` stores `session.billingAddress`, then `commitSession(...)` calls `paymentElement.update(checkout:)`. Before this change, FlowController/EPE merged `session.billingAddress`, `session.shippingAddress/billing/email` into their configs during update to prefill fields.

The new pattern only seems to copy `configuration.defaults.billingDetails` via `defaultBillingDetails.set(...)`, which only sets name/address from the config defaults. So `Checkout update shipping address -> PaymentElement update` no longer seems to prefill MPE from the updated session local fields.

porter-stripe: I see shippingDetails, phone, and email on Defaults, but we don't read them?

yuki-stripe: > I see shippingDetails, phone, and email on Defaults, but we don't read them?
Will deal with in a future PR. 

FYI this PR was a draft when you commented!

yuki-stripe: I added `applyAddressOverrides` back, agree it's needed although `session.billing/updateBillingAddress` will go away.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Payment Element/PaymentElement+Configuration.swift
```diff
@@ -45,8 +45,12 @@ extension PaymentElement {
         /// Describes how billing details should be collected.
         public var billingDetailsCollectionConfiguration: BillingDetailsCollectionConfiguration = .init() {
             didSet {
-                paymentSheetConfiguration.billingDetailsCollectionConfiguration = billingDetailsCollectionConfiguration
-                embeddedConfiguration.billingDetailsCollectionConfiguration = billingDetailsCollectionConfiguration
+                paymentSheetConfiguration.billingDetailsCollectionConfiguration = billingDetailsCollectionConfiguration.paymentSheetConfiguration(
+                    billingAddressCollection: .automatic
```
porter-stripe: Why hardcode .automatic here? Does this get overwritten somewhere?

yuki-stripe: Shoot, thanks. We shouldn't hardcode, but also this gets overwritten by `Checkout+Session+AddressMerging` and is unnecessary. Updated.

porter-stripe (approved):

PR #6723, opened 2026-07-16
Use Link PM type, and attach mandate data to confirm params

mats-stripe:
## Summary

1. Attaches mandate data to the setup intent before confirming it. This is required for bank payments, which would fail otherwise with `Missing required param: mandate_data`.
2. Sets `pmt=link` on the intent

## Motivation

1 is required for bank payments to work, 2 is needed for those to be treated as IBP, not ACH

## Testing

Tested manually

## Changelog

N/a

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/ExampleLinkControllerPreviewView.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/ExampleLinkControllerPreviewView.swift
@@ -67,6 +67,15 @@ struct ExampleLinkControllerPreviewView: View {
                         }
                     }
 
+                    DemoSection(title: "Intent Payment Method Types") {
+                        Picker("Intent Payment Method Types", selection: $configuration.paymentMethodTypesMode) {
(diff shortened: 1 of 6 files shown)


mats-stripe:
Updated it so that the `payment_method_types` used is passed in via the `LinkConfiguration`, instead of hardcoding it to `link`. By default, it will be `nil`, which will use automatic PM types. https://github.com/stripe/stripe-ios/pull/6723/commits/76e6482979ac86f6596cd53ea5f7fa252982c3ff + https://github.com/stripe/stripe-ios/pull/6723/commits/01afdbe47d9ed72ea41eec44179ea5225adbb5f2

CHANGELOG.md
```diff
@@ -1,5 +1,5 @@
 The next release's version bump will so far be:
-PATCH
+MINOR
```
tillh-stripe, line 2: Is this accidental?

mats-stripe: No, it was required for the public API check GH action to pass

porter-stripe (approved):

PR #6724, opened 2026-07-16
[🪿] Make STPAPIClient.betas public

porter-stripe:
## Summary
Previously `@_spi(STP)`, the ability to pass beta version headers is inherent to any Stripe API client and has valid merchant-facing use cases (e.g. beta-header-gated payment methods where the server requires a preview flag on the Stripe-Version header). Exposing it publicly is low risk since it simply widens access to an existing property.

Also updated the doc comment to use the `["alipay_beta=v1"]` example form and to note that beta headers should only be used when a feature can't be gated via `@_spi` and the server requires the header; most SDK betas are opted into via other APIs.

## Motivation
https://stripe.slack.com/archives/C0353FPANQ0/p1784214006249639

## Testing
Added tests covering single/multiple beta appending to the Stripe-Version header and the empty default.

## Changelog
See diff

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 The next release's version bump will so far be:
-PATCH
+MINOR
 
 ## X.Y.Z - changes pending release
 ### StripeCore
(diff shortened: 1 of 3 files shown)


gbirch-stripe (commented):
Should this make the version bump MINOR since it's an API change? Since it's just adding an API it doesn't break anything but seems more semantically correct to be minor?

gbirch-stripe (approved):

PR #6729, opened 2026-07-16
Intent.checkout -> Intent.checkoutSession

yuki-stripe:
## Summary
- Change Intent.checkout to store Checkout.Session instead of Checkout
- Add a PaymentElement retain-cycle regression test

## Testing
- ci_scripts/run_tests.rb --test StripePaymentSheetTests/PaymentElementTest/testCheckoutAndPaymentElementDoNotRetainEachOther
- ci_scripts/run_tests.rb --scheme StripePaymentSheet --build-only
- ci_scripts/format_modified_files.sh
- ci_scripts/lint_modified_files.sh

Also verified the retain-cycle test fails when temporarily reverting Intent.checkout back to Checkout.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/CheckoutSessionBillingAddressUpdater.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/CheckoutSessionBillingAddressUpdater.swift
@@ -1,0 +1,64 @@
+//
+//  CheckoutSessionBillingAddressUpdater.swift
+//  StripePaymentSheet
+//
+//  Created by Yuki Tokuhiro on 7/17/26.
+//
+
+@_spi(STP) import StripePayments
+
+/// Narrow Checkout interface for MPE billing address updates.
+///
+/// This keeps MPE plumbing from depending on the full `Checkout` object or reading
+/// `checkout.session` directly. Callers must use these sanctioned update methods to sync
+/// billing changes or commit session updates.
+@MainActor
+protocol CheckoutSessionBillingAddressUpdater: AnyObject {
+    // TODO: Delete this when CheckoutSession confirmation no longer uses `PaymentSheet.confirm`.
+    func commitSession(
+        _ apiResponse: PaymentPagesAPIResponse?,
+        applying localMutation: (@MainActor @Sendable (Checkout.Session) -> Checkout.Session)?
+    ) async throws
+
+    func updateBillingAddressForPaymentSheet(
+        name: String?,
+        phone: String?,
+        address: Checkout.Address,
+        canUpdateWhileSheetPresented: Bool
+    ) async throws -> Checkout.Session
+}
+
+extension CheckoutSessionBillingAddressUpdater {
+    func commitSession(_ apiResponse: PaymentPagesAPIResponse) async throws {
+        try await commitSession(apiResponse, applying: nil)
+    }
+
+    func updateBillingAddressForPaymentSheet(
+        address: Checkout.Address,
+        canUpdateWhileSheetPresented: Bool
+    ) async throws -> Checkout.Session {
+        try await updateBillingAddressForPaymentSheet(
+            name: nil,
+            phone: nil,
+            address: address,
+            canUpdateWhileSheetPresented: canUpdateWhileSheetPresented
+        )
+    }
+}
+
+extension Checkout: CheckoutSessionBillingAddressUpdater {
+    func updateBillingAddressForPaymentSheet(
+        name: String?,
+        phone: String?,
+        address: Address,
+        canUpdateWhileSheetPresented: Bool
+    ) async throws -> Session {
+        try await updateBillingAddress(
+            name: name,
+            phone: phone,
+            address: address,
+            canUpdateWhileSheetPresented: canUpdateWhileSheetPresented
+        )
+        return session
+    }
+}
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Intent.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Intent.swift
@@ -16,19 +16,19 @@ import UIKit
 
 // MARK: - Intent
 
-/// An internal type representing either a PaymentIntent, SetupIntent, a "deferred Intent", or a Checkout
+/// An internal type representing either a PaymentIntent, SetupIntent, a "deferred Intent", or a Checkout Session
 enum Intent {
     case paymentIntent(STPPaymentIntent)
     case setupIntent(STPSetupIntent)
     case deferredIntent(intentConfig: PaymentSheet.IntentConfiguration)
-    case checkout(Checkout)
+    case checkout(Checkout.Session)
 
     var stripeId: String? {
         switch self {
         case .paymentIntent(let intent): intent.stripeId
         case .setupIntent(let intent): intent.stripeID
         case .deferredIntent: nil
-        case .checkout(let checkout): checkout.nonisolatedSession.id
+        case .checkout(let session): session.id
         }
     }
 
@@ -45,8 +45,8 @@ enum Intent {
             case .setup:
                 return false
             }
-        case .checkout(let checkout):
-            return checkout.nonisolatedSession.mode == .payment || checkout.nonisolatedSession.mode == .subscription
+        case .checkout(let session):
+            return session.mode == .payment || session.mode == .subscription
         }
     }
 
@@ -97,8 +97,8 @@ enum Intent {
             case .setup(let currency, _):
                 return currency
             }
-        case .checkout(let checkout):
-            return checkout.nonisolatedSession.currency
+        case .checkout(let session):
+            return session.currency
         }
     }
 
@@ -115,8 +115,8 @@ enum Intent {
             case .setup:
                 return nil
             }
-        case .checkout(let checkout):
-            return checkout.nonisolatedSession.expectedAmount()
+        case .checkout(let session):
+            return session.expectedAmount()
         }
     }
 
@@ -129,10 +129,10 @@ enum Intent {
                 return setupFutureUsage?.rawValue
             }
             return nil
-        case .checkout(let checkout):
-            switch checkout.nonisolatedSession.mode {
+        case .checkout(let session):
+            switch session.mode {
             case .payment:
-                return checkout.nonisolatedSession.setupFutureUsage
+                return session.setupFutureUsage
             case .setup:
                 return nil
             case .subscription, .unknown:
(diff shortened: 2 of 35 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+API.swift
```diff
@@ -42,6 +42,7 @@ extension PaymentSheet {
         paymentHandler: STPPaymentHandler,
         integrationShape: IntegrationShape = .complete,
         paymentMethodID: String? = nil,
+        checkout: Checkout? = nil,
```
yuki-stripe: I would not look too closely at PaymentSheet+API / confirm because I'm probably going to rewrite confirm to some degree later.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Checkout/PaymentElementTest.swift
```diff
@@ -70,6 +70,32 @@ final class PaymentElementTest: XCTestCase {
         XCTAssertEqual(checkout.session.paymentOption?.paymentMethodType, "paynow")
     }
 
+    func testCheckoutAndPaymentElementDoNotRetainEachOther() async throws {
```
yuki-stripe, line 73: this test failed before and passes now.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Intent.swift
```diff
@@ -16,19 +16,19 @@ import UIKit
...
 // MARK: - Intent
 
-/// An internal type representing either a PaymentIntent, SetupIntent, a "deferred Intent", or a Checkout
+/// An internal type representing either a PaymentIntent, SetupIntent, a "deferred Intent", or a Checkout Session
 enum Intent {
     case paymentIntent(STPPaymentIntent)
     case setupIntent(STPSetupIntent)
     case deferredIntent(intentConfig: PaymentSheet.IntentConfiguration)
-    case checkout(Checkout)
+    case checkout(Checkout.Session)
```
porter-stripe, line 24: Intent.checkout holds a Checkout.Session captured at load time but the `Checkout` we pass around is live and can mutate, so now we have two sources of truth for the session. The underlying session on the `Checkout` can become out of sync with what lives on the `Intent` in cases like Apple Pay where we update the billingAddress and attempting to confirm the session I believe can mutate it leaving intent.checkout.session stale.

This also adds this code smell where we are injecting a `Checkout` everywhere we need it like Apple Pay, and adding in asserts when we hit a case where we expect it to be populated compared to previously we always knew we had an instance b/c it on

Is there some way we can leave Checkout on the Intent and solve the memory issue? Maybe with a wrapper?
```swift
  final class WeakCheckoutRef {  // exists only to hold the weak ref
      weak var checkout: Checkout?
      init(_ checkout: Checkout) { self.checkout = checkout }
  }
```

Then 
`case checkout(WeakCheckoutRef)`

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet+Link.swift
```diff
@@ -68,7 +68,13 @@ extension PaymentSheet {
         shouldFinishOnClose: Bool,
         onClose: (() -> Void)? = nil
     ) {
-        let payWithNativeLink = PayWithNativeLinkController(mode: .full, intent: intent, elementsSession: elementsSession, configuration: configuration, analyticsHelper: analyticsHelper, confirmationChallenge: confirmationChallenge)
+        let checkout: Checkout? = {
+            guard case .checkout(let checkout) = mode else {
+                return nil
+            }
+            return checkout
+        }()
```
porter-stripe: If we do keep this can we DRY it up, maybe a property on the Mode?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Intent.swift
yuki-stripe: > Intent.checkout holds a Checkout.Session captured at load time but the Checkout we pass around is live and can mutate, so now we have two sources of truth for the session.

That's true. I want to make Intent.checkout hold a Session because it cannot mutate, _just like the other Intents_ (PI, SI, DeferredIntentConfig). MPE is not designed or architected to handle the Intent changing. 

To address 'two sources of truth' I can replace `Checkout` a "Checkout...Helper" protocol that only contains the methods we need. 

Regarding the fact that MPE _does_ in fact mutate the session, whenever that happens the sheet immediately dismisses + MPE is reloaded. Hence "MPE can't handle Intent changing" - we handle it by throwing away MPE and creating a new one.

> This also adds this code smell where we are injecting a Checkout everywhere
We inject Checkout either standalone or inside Intent, so I think the critique here is that having it separate from the Intent is more boilerplate, which I feel like is minor? 

The confirm case (which includes Apple Pay's injection of checkout) will be reworked in a future PR, and I think it will remove the injection. When it's reworked, we won't want to use a weak checkout inside Intent.checkout because we'll be invoking confirm directly from the checkout instance and therefore will have access to a strong reference.

yuki-stripe: I'm also not even sure we will use STPApplePayContext at all. I think ECE might be better off writing their own thing, and we would use that.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/CheckoutSessionBillingAddressUpdater.swift
```diff
@@ -0,0 +1,64 @@
...
+/// This keeps MPE plumbing from depending on the full `Checkout` object or reading
+/// `checkout.session` directly. Callers must use these sanctioned update methods to sync
+/// billing changes or commit session updates.
+@MainActor
+protocol CheckoutSessionBillingAddressUpdater: AnyObject {
+    // TODO: Delete this when CheckoutSession confirmation no longer uses `PaymentSheet.confirm`.
+    func commitSession(
+        _ apiResponse: PaymentPagesAPIResponse?,
+        applying localMutation: (@MainActor @Sendable (Checkout.Session) -> Checkout.Session)?
+    ) async throws
```
yuki-stripe, line 21: please note this is probably going away when we actually build `confirm`.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Intent.swift
porter-stripe: > Regarding the fact that MPE does in fact mutate the session, whenever that happens the sheet immediately dismisses + MPE is reloaded. Hence "MPE can't handle Intent changing" - we handle it by throwing away MPE and creating a new one.

This is mostly true, but not for Apple Pay. We mutate the CheckoutSession when the Apple Pay sheet is opened and PMs are selected within it. It’s hard to picture what the confirm case will look like after the changes you are referencing, but today I think this breaks Apple Pay with auto tax. I can’t verify, though, because I’m hitting `Assertion failed: Payment Element assumes that the FlowController's payment option is the same as the Embedded's on first load!`

The failure mode is that the snapshot may have $10 as the expected amount, but opening the Apple Pay sheet mutates the session by applying a billing address, which changes the total due to tax. Now the snapshot session is out of sync with what’s on `Checkout.Session`, and I think confirm will fail with an `amount` / `expectedAmount` mismatch.

This makes me think we definitely need better testing around the automatic tax angle, but also that the restriction against allowing the Intent to mutate while the sheet is open should be dumped. We added it here (and partially removed it later): https://github.com/stripe/stripe-ios/pull/6260

porter-stripe: One thing on the memory issue: I don't think `Intent.checkout(Checkout)` is what introduced the retain cycle. Before `PaymentElement`, MPE owned the `Intent.checkout` and Checkout's reference back to MPE was via a weak delegate, so I don't think there was a strong reference loop there.

Stepping back across the recent PE changes, could the new ownership/API shape be what's making this harder? Checkout now owns PE, PE owns both PS.FC and EPE and we need logic to mirror payment option state then propagate it back to Checkout. And now here we're changing Intent from a live Checkout to a snapshot and passing a separate helper around. I think I preferred the earlier shape where each Element was initialized with a Checkout. The ownership and data flow were linear and easy to understand, and the merchant API was just as usable. I do like having a single Checkout configuration, can we get that configuration win without Checkout also owning all Elements and introducing the additional state/memory complexity?

yuki-stripe: > This is mostly true, but not for Apple Pay. ...
Ah, so MPE != Apple Pay, and I think when we say 'sheet' in this context it should only refer to our sheet, not Apple Pay's sheet. 

MPE can't mutate the CheckoutSession - if it does, you have to re-load it. Apple Pay can mutate. They're separate code; the latter doesn't need `Intent.checkout(CheckoutSession)` in order to mutate. The Apple Pay flow is started by calling `confirm` and `STPApplePayContext` can get the `checkout` object directly. 


--
> could the new ownership/API shape be what's making this harder? 

I disagree this is harder! What is hard? I think the model in this PR is simple:
```
+--------------------+                +----------+
| CheckoutController | ----strong---->| Elements |
|        (C)         | < - - weak - - |   (E)    |
+--------------------+                +----------+
```

This model is analogous to 
```
+--------------------+                +----------+
| ViewController.    | ----strong---->| View.   |
|        (VC)        | < - - weak - - |   (V)    |
+--------------------+                +----------+
```

or any parent -> child relationship where the weak connection back to the parent is implemented as a delegate.


> ..the restriction against allowing the Intent to mutate while the sheet is open should be dumped. We added it here (and partially removed it later): https://github.com/stripe/stripe-ios/pull/6260

The constraint it imposes - that MPE's Session is immutable - is _good_, because it really can't handle the Intent being mutated. I believe that PR shows this - it doesn't rely on the Intent being mutable, it re-creates MPE with a new Intent - and can still be implemented the same way.

porter-stripe: I agree Apple Pay's sheet is separate, but the Apple Pay flow starts inside MPE with it's state and I see how the protocol prevents MPE from reading the live Checkout now and improves the two sources of truth to an extent. But the billing address update returns the updated session and Apple Pay uses it for the summary items while (I think) confirmation still uses the snapshot. What's the rule after a mutation? Which session should the rest of the flow use?

I also agree the ARC diagram is simple, but the state flow isn't. PE needs to reconcile PS.FC and EPE, flow the payment option back into Checkout, and this PR adds a delegate path that can return an updated Checkout session while the Intent still holds the original snapshot. It feels like the previous MPE(checkout:) flow was more linear and easier to follow and implement.

 Maybe a future Checkout owned confirm resolves this, so maybe I'm missing something about the final state.

porter-stripe (approved):

PR #6730, opened 2026-07-16
Remove test-only Checkout initializers

yuki-stripe:
## Summary
- remove DEBUG-only Checkout initializers
- route Checkout tests through real Checkout(configuration:) with stubbed API client/configuration fixtures
- keep existing recorded loader tests on the recording path where already available

## Testing
Existing tests


StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Checkout/CheckoutUnitTests.swift
```diff
@@ -475,9 +475,9 @@ final class CheckoutUnitTests: XCTestCase {
 
         try await checkout.commitSession(updatedSession)
 
-        XCTAssertEqual(callOrder, ["regular"])
-        XCTAssertEqual(delegate.updateSessionCallCount, 1)
-        XCTAssertEqual(recorder.sessions.count, 1)
+        XCTAssertEqual(callOrder, ["regular", "regular"])
```
porter-stripe, line 485: Can we drop a comment explaining why it's 2 and not 1, like: "one emission for the committed session, one for PaymentElement re-syncing the payment option"? Otherwise the next person can't tell whether 2 is correct or a bug someone rubber-stamped by bumping the number.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Checkout/CheckoutUnitTests.swift
```diff
@@ -475,9 +475,9 @@ final class CheckoutUnitTests: XCTestCase {
...
+        XCTAssertEqual(delegate.updateSessionCallCount, 2)
```
porter-stripe, line 487: Is the double checkoutDidUpdateSession per commit actually the behavior we want to expose to merchants?

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Checkout/CheckoutTestHelpers.swift
```diff
@@ -96,9 +100,58 @@ enum CheckoutTestHelpers {
...
     // MARK: - Checkout-flow helpers
 
-    static func makeCheckoutWithOpenSession() async -> Checkout {
-        let session = makeOpenSession()
-        return await Checkout(clientSecret: "cs_test_123_secret_abc", apiResponse: session)
+    @MainActor
+    static func makeConfiguration(
+        apiResponse: PaymentPagesAPIResponse = makeOpenSession(),
+        configuration: Checkout.Configuration? = nil,
+        stubAllOutgoingRequests: Bool = true
```
porter-stripe, line 113: Can we explain what this does, at first glance it's not clear what/where the stub is.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Checkout/CheckoutUnitTests.swift
yuki-stripe: I'm not sure how much this matters, I thought of 'reducing # of updates' as a very minor optimization that probably has no effect if we're not making lots of updates. Like it only matters to the extent it makes the UI laggy due to rendering too much? Is that wrong?

porter-stripe: I'm peaceful w/ 2, I think could be a little confusing for merchants who make 1 call but see 2 updates.

yuki-stripe: I filed https://jira.corp.stripe.com/browse/MOBILESDK-4701 to track

StripeApplePay.podspec
```diff
@@ -3,7 +3,7 @@ Pod::Spec.new do |s|
 
   # Do not update s.version directly. 
   # Instead, update the VERSION file and run ./ci_scripts/update_version.sh 
-  s.version                        = '26.4.0' 
+  s.version                        = '26.4.0'
```
porter-stripe, line 6: ?

porter-stripe (approved):

StripeApplePay.podspec
yuki-stripe: I used codex for the merge (well, I use it for almost everything these days) and it says it uses `git diff --cached --check` which will call out trailing whitespace. That's why it noticed it, and it fixed it because LLM 🤷

PR #6737, opened 2026-07-17
Refactor AddressSectionElement autocomplete

porter-stripe:
## Summary
- Renames `CollectionMode` to `FieldsToCollect`
- Moves autocomplete onto `.all` as an associated value, and drops the separate compact/expanded presentation
- `AddressSectionElement` now manages the compact -> expanded autocomplete transition internally (default address values, unsupported country, or manual entry all trigger it), so call sites no longer set `collectionMode`/`fieldsToCollect` to force expansion
- Based on suggestion here: https://github.com/stripe/stripe-ios/pull/6704#discussion_r3599285293

Follow up PRs will add ability to override `FieldsToCollect` by country for auto tax.

## Motivation
- CheckoutSessions auto tax

## Testing
- Existing

## Changelog
N/A


Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheet+AddressTests.swift
```diff
@@ -494,12 +494,13 @@ RO
         // ...should update PaymentSheet.FlowController
         app.buttons["Payment method"].waitForExistenceAndTap()
         XCTAssertEqual(app.textFields["Country or region"].value as? String, "Uruguay")
+        XCTAssertFalse(app.textFields["Postal code"].exists)
+        XCTAssertFalse(app.textFields["ZIP"].exists)
 
         // If you change the billing address, however...
-        let updatedBillingAddressPostalCode = "12345"
```
porter-stripe, line 499: Worth calling out, this slightly changes existing behavior on purpose. Today adding shipping details will automatically expand billing into a full address field form, which seems accidental and can make us collect fields we don't need (e.g. postal code for Uruguay ). After this we don't expand, we still show the minimum fields required for billing if we are in .automatic in MPE even if we have prefills from a shipping address.

porter-stripe:
@yuki-stripe this is ready, we refactored out "compact" and "expanded" auto complete modes. Now AddressSectionElement just expands when it needs to. Removes quite a bit of duplicated logic.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
```diff
@@ -889,7 +872,7 @@ extension PaymentSheetFormFactory {
 
         let countries = configuration.billingDetailsCollectionConfiguration.allowedCountriesArray
         let addressElement = billingConfiguration.address == .full
-            ? makeBillingAddressSection(collectionMode: .autocomplete(), countries: countries)
+            ? makeBillingAddressSection(fieldsToCollect: .all(autocomplete: .init()), countries: countries)
```
porter-stripe: this is technically wrong, we should be passing in a list of countries like AddressElement. This is an existing issue/bug, will tackle in: https://jira.corp.stripe.com/browse/RUN_MOBILESDK-5529

StripeUICore/StripeUICore/Source/Elements/Factories/Address/AddressSectionElement.swift
```diff
@@ -65,28 +65,25 @@ import UIKit
...
-        ///   - autocompleteCountries: Countries that support autocomplete. If nil, all countries are supported.
-        ///     Unsupported countries use manual entry in expanded mode.
-        ///     Compact mode callers should expand when an unsupported country is selected.
-        ///   - presentation: Whether to show only the autocomplete entry point or the full address form.
-        case autocomplete(autocompleteCountries: [String]? = nil, presentation: AutocompletePresentation = .compact)
+    /// Describes which address fields to collect.
+    public enum FieldsToCollect: Equatable {
+        /// Shows all address fields, optionally with autocomplete.
+        /// - Parameter autocomplete: The autocomplete configuration to use. If nil, autocomplete is disabled.
+        case all(autocomplete: Autocomplete? = nil)
```
yuki-stripe: I still think this should be outside of `FieldsToCollect`. I think w/ the current PR we'll have to preserve the autocomplete config as we switch between FieldsToCollect for different countries. In that world - where FIeldsToCollect varies per country - the current PR also means you can specify different `autocompleteCountries` per country, which is wrong.


I'd propose:
```
class AddressSectionElement {
  /// When ASE is collecting a full billing address and the country is in this list, it will allow autocomplete.
  var autocompleteCountries: [String] = defaultAutocompleteCountries
}
```


I'm pretty sure we'll be glad we did it this way in future PRs, and I _think_ it's a small refactor, but it's fine if you want to ship this as-is and deal with it later.

Example/PaymentSheet Example/PaymentSheetUITest/PaymentSheet+AddressTests.swift
yuki-stripe: Niice

StripeUICore/StripeUICore/Source/Elements/Factories/Address/AddressSectionElement.swift
porter-stripe: Good idea! I've refactored autocomplete out of the .all case!

StripeIdentity/StripeIdentity/Source/Elements/IdentityElementsFactory.swift
```diff
@@ -94,6 +94,7 @@ struct IdentityElementsFactory {
             countries: countries,
             locale: locale,
             addressSpecProvider: addressSpecProvider,
+            disableAutocomplete: true,
```
porter-stripe, line 97: Just flagging, some consumes of ASE don't use auto complete so we need this flag which was previously expressed as just .all to disable auto complete. I think there are some improvements we can make here on that relation but thats for another time.

porter-stripe:
snapshots approved
1. some cases are now just adding a magnifying class when expanded, this is expected b/c now auto complete is always on unless opted out of
2. testLpm_Klarna_AllFields_NoDefaults: When using .all, it now is condensed for auto complete!

yuki-stripe (approved):

PR #6738, opened 2026-07-17
Add Tempo support to CryptoOnramp

jeanregisser:
## Summary

Adds Tempo as a supported CryptoOnramp network and pins CryptoOnramp internal API calls to the same preview API version used by the web SDK. This lets mobile register wallets for newer networks that are behind preview API features without changing the global iOS SDK API version.

A small SPI `apiVersionOverride` parameter was added in STPAPIClient so the CryptoOnramp SDK can own this versioning internally instead of defaulting to the one set by STPAPIClient (2020-08-27, which does not work for some crypto onramp internal calls).

Jira: https://jira.corp.stripe.com/browse/ONRAMP-611

## Testing

- Focused STPAPIClientCryptoOnrampTests.testCollectWalletAddressSuccess passed on the iPhone 17 simulator.
- Built and manually ran the CryptoOnramp Example app on the iPhone 17 simulator; Tempo wallet registration succeeded with a fresh address.


StripeCryptoOnramp/StripeCryptoOnramp/Source/Internal/API Bindings/STPAPIClient+CryptoOnramp.swift
```diff
@@ -367,9 +371,20 @@ private extension STPAPIClient {
...
             }
         }
     }
 }
+
+private enum CryptoOnrampAPI {
+    // Use a preview API version for networks and parameters behind preview API features.
+    // Bump this when new onramp features require a newer API version.
+    static let stripeAPIVersion = "2026-03-25.preview"
+    static let additionalHeaders = ["Stripe-Version": stripeAPIVersion]
```
porter-stripe: `additionalHeaders` here fully replaces `Stripe-Version` in `configuredRequest` rather than merging. If a merchant has enabled a beta on the shared `apiClient` (e.g. via FinancialConnections' `apiClient.betas.insert(...)`), that beta silently gets dropped on these crypto-internal calls since `CryptoOnrampCoordinator` defaults to the same shared `STPAPIClient`. 

I think we should probably merge them, especially b/c we are making betas public: https://github.com/stripe/stripe-ios/pull/6724

jeanregisser: Good callout, thanks. I agree we shouldn’t drop any betas already configured on the shared `apiClient`.

Instead of using raw `additionalHeaders` for this, what do you think about adding an SPI `apiVersionOverride` parameter? `STPAPIClient` could use that as the base version, then append `betas` the same way it already does today. CryptoOnramp would pass `2026-03-25.preview`, and existing client betas would be preserved automatically.

That feels a bit closer to what we mean here: override the base API version for this module, while keeping normal `STPAPIClient` header behavior.

jeanregisser:
Thanks @porter-stripe, I updated this to use a per-request SPI `apiVersionOverride` instead of overriding the header directly. Crypto Onramp now applies `2026-03-25.preview` through its request helpers, matching the updated approach.

porter-stripe (dismissed):

porter-stripe (approved):

PR #6740, opened 2026-07-18
Revert persisted PaymentSheet selection on cancellation

porter-stripe:
## Summary
- When a customer cancels PaymentSheet, any selection change they made is now reverted back to the selection that was persisted before the sheet was presented.
- This aligns with Android

### Demo
Before: pay with SPM -> select a different SPM card 4242 -> cancel -> re-open card 4242 selected
After: (pay with SPM -> select a different SPM  -> cancel -> re-open bank selected

https://github.com/user-attachments/assets/443f15f3-56d1-4c91-8e0c-68d32bdd079e

## Motivation
Previously, canceling out of PaymentSheet after switching payment methods would still persist the newly selected method, even though the customer never confirmed the change. This restores the original selection on cancellation.

## Testing
Added `PaymentSheetCancelPersistenceTests.swift` covering cancellation with and without a prior persisted selection.

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
@@ -130,6 +130,21 @@ public class PaymentSheet {
         Task { @MainActor in
             // Overwrite completion closure to retain self until called
             let completion: (PaymentSheetResult) -> Void = { status in
+                // Every terminal path, including native Link, reaches this completion.
+                switch status {
+                case .canceled:
+                    if let paymentSheetViewController = self.bottomSheetViewController.contentStack.first(
+                        where: { $0 is PaymentSheetViewControllerProtocol }
+                    ) as? PaymentSheetViewControllerProtocol {
+                        self.revertPersistedSelectionAfterCancellation(
+                            using: paymentSheetViewController.savedPaymentMethods
+                        )
+                    }
+                case .completed, .failed:
+                    // PaymentSheet is finishing without cancellation, so discard its cancellation snapshot.
+                    self.persistedSelectionSnapshotBeforePresentation = nil
+                }
+
                 // Dismiss if necessary
                 if let presentingViewController = self.bottomSheetViewController.presentingViewController {
                     // Calling `dismiss()` on the presenting view controller causes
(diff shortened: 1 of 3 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
```diff
@@ -128,28 +128,49 @@ public class PaymentSheet {
         completion: @escaping (PaymentSheetResult) -> Void
     ) {
         Task { @MainActor in
-            // Overwrite completion closure to retain self until called
-            let completion: (PaymentSheetResult) -> Void = { status in
-                // Dismiss if necessary
+            let merchantCompletion = completion
+            // Retain PaymentSheet until the presentation finishes.
+            let finishPresentation: (PaymentSheetResult) -> Void = { result in
+                // Clear this before dismissing so another terminal callback can't finish the presentation twice.
```
gbirch-stripe: What's a "terminal callback"?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
```diff
@@ -128,28 +128,49 @@ public class PaymentSheet {
...
+                self.completion = nil
+
+                let completeAfterDismissal = {
+                    switch result {
+                    case .canceled:
+                        // Native Link can finish without calling the PaymentSheet cancel delegate.
```
gbirch-stripe: What does this code block have to do with Link?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
gbirch-stripe, line 1: Could we maybe refactor / clean up this code to not have so many closures defined in-line? I find it very confusing to read and understand

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
porter-stripe: Good call. I replaced "terminal callback" with a explanation of why the completion is cleared.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
porter-stripe: PaymentSheet can present native Link, so dimissing PaymentSheet also dismisses Link. I clarified this in the comment.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheet.swift
porter-stripe: Yep. I extracted the inline closures into named private methods to make the completion flow easier to follow.

gbirch-stripe (approved):

PR #6741, opened 2026-07-18
Revert FlowController selection on cancellation

porter-stripe:
## Summary
Extends selection-revert-on-cancel behavior to `PaymentSheet.FlowController`. Canceling out of the payment options sheet now reverts to the selection that existed before it was presented, for both horizontal and vertical layouts.

- Snapshots the accepted selection (and, where relevant, the completed form behind it) before presenting payment options
- On cancel, reverts the persisted default and rebuilds the view controller with the snapshotted selection if needed
- Handles the saved-payment-method-deleted-while-open case by falling back to whatever's now selected instead of restoring a payment method that no longer exists
- Handles linked bank accounts selected through the Instant Debits/Link Card Brand form, which are represented as `.saved` but need to be restored through their form, not a saved-PM row
- Fixes a pre-existing init-ordering bug in `PaymentSheetVerticalViewController` where `regenerateUI()` ran before `linkConfirmOption` was restored

## Motivation
FlowController had the same gap as PaymentSheet: canceling the payment options sheet could leave a new selection applied even though the customer backed out. This snapshots the selection before presenting and reverts it on cancellation.

## Testing
- New tests
- Existing
- Manual

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/PaymentSheetVerticalViewController.swift
```diff
@@ -183,13 +198,12 @@ class PaymentSheetVerticalViewController: UIViewController, FlowControllerViewCo
         self.analyticsHelper = analyticsHelper
         super.init(nibName: nil, bundle: nil)
 
-        regenerateUI()
-
         if case let .link(linkConfirmOption) = previousPaymentOption {
             self.linkConfirmOption = linkConfirmOption
         }
+        regenerateUI()
```
gbirch-stripe, line 192: What's the reasoning for moving this line?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/FlowControllerStateSnapshot.swift
```diff
@@ -0,0 +1,76 @@
...
+internal struct FlowControllerStateSnapshot {
+    enum Source {
+        case viewController
+        case presented(FlowControllerStateSnapshot)
+        case restored(Restoration)
+    }
+
+    struct Selection {
+        var paymentOption: PaymentOption?
+    }
```
gbirch-stripe: Can we get rid of this struct and just directly use the underlying type?

porter-stripe: The follow up PR: #6742 expands this type a bit, so I'd like to keep this.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/PaymentSheetVerticalViewController.swift
porter-stripe: This fixes a pre-existing init ordering bug exposed by this change. When rebuilding a form-only controller with Link selected, regnerateUI() ran before the linkConfirmOption was restored.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFlowController.swift
```diff
@@ -843,25 +871,63 @@ extension PaymentSheet {
...
+                    prioritizing: selection.paymentOption
+                ),
                 analyticsHelper: analyticsHelper,
-                walletButtonsViewState: self.walletButtonsViewState,
-                previousPaymentOption: self.internalPaymentOption
+                walletButtonsViewState: walletButtonsViewState,
+                initialState: .restoringAfterCancellation(selection)
             )
             self.viewController.flowControllerDelegate = self
-            updatePaymentOption()
```
porter-stripe, line 864: Removal is intentional. The new wrapped completion always calls updatePaymentOption before invoking merchant code.

gbirch-stripe (approved):

PR #6744, opened 2026-07-18
Add country override to AddressSectionElement

porter-stripe:
## Summary
Adds per-country minimum address field requirements to `AddressSectionElement`. `makeBillingAddressSection` now takes a `defaultFieldsToCollect` plus a `minimumFieldsToCollectByCountry` map, so a country can require more fields (e.g. postal code) than the default without lowering it for other countries. Also renames `.countryOnly` to `.country` and simplifies `.countryAndPostal` to always collect postal code.

## Motivation
Billing address collection needed a way to require additional fields for specific countries while keeping a lighter default for the rest.

## Testing
- Updated tests

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory+Card.swift
```diff
@@ -94,9 +100,9 @@ extension PaymentSheetFormFactory {
                 : Array(configuration.billingDetailsCollectionConfiguration.allowedCountries)
             switch configuration.billingDetailsCollectionConfiguration.address {
             case .automatic:
-                return makeBillingAddressSection(fieldsToCollect: .countryAndPostal(), countries: countries, includeEmail: shouldIncludeEmail, includePhone: shouldIncludePhone)
+                return makeBillingAddressSection(defaultFieldsToCollect: .country, minimumFieldsToCollectByCountry: Self.cardMinimumFieldsToCollectByCountry, countries: countries, includeEmail: shouldIncludeEmail, includePhone: shouldIncludePhone)
             case .full:
-                return makeBillingAddressSection(fieldsToCollect: .all, countries: countries, includeEmail: shouldIncludeEmail, includePhone: shouldIncludePhone)
+                return makeBillingAddressSection(defaultFieldsToCollect: .all, minimumFieldsToCollectByCountry: Self.cardMinimumFieldsToCollectByCountry, countries: countries, includeEmail: shouldIncludeEmail, includePhone: shouldIncludePhone)
```
yuki-stripe: We don't need to pass `minimumFieldsToCollectByCountry` here right? It does nothing?

yuki-stripe (dismissed):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory+Card.swift
porter-stripe: Yep, you're right. `.all` always wins, so the minimum has no effect here. I'll remove it.

yuki-stripe (approved):

PR #6745, opened 2026-07-19
Collect automatic tax billing address minimums

porter-stripe:
## Summary
Payment method forms now collect the minimum billing address fields required for automatic tax calculation, based on the customer's country. This applies even to LPM forms that don't normally collect an address.

## Motivation
Automatic tax needs certain billing address fields (e.g. postal code, state) to calculate tax correctly. Some payment method forms don't collect enough address detail today, or don't collect an address at all, leading to inaccurate tax calculation.

## Testing
Added unit tests covering the new minimum field requirements per country and the case where a form has no existing address section.

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
```diff
@@ -230,6 +238,11 @@ class PaymentSheetFormFactory {
     }
 
     func make() -> PaymentMethodElement {
+        let form = makePaymentMethodForm()
+        return applyAutomaticTaxMinimumsIfNecessary(to: form)
```
porter-stripe, line 238: I think ideally this is makePaymentMethodForm -> applyTaxAndBDCC(to:form) eventually

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/STPApplePayContext+PaymentSheet.swift
```diff
@@ -427,6 +427,10 @@ extension STPApplePayContext {
             currency: intent.currency ?? "USD"
         )
         paymentRequest.requiredBillingContactFields = makeRequiredBillingDetails(from: configuration)
+        // Apple Pay can't vary billing fields by country, so automatic tax from billing needs the postal address.
```
yuki-stripe, line 430: Is this what EwCS does too?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
```diff
@@ -958,16 +961,28 @@ extension PaymentSheetFormFactory {
...
+        let defaultFieldsToCollect: AddressSectionElement.FieldsToCollect? = {
+            switch configuration.billingDetailsCollectionConfiguration.address {
+            case .automatic where requiredByPaymentMethod:
+                return .all
+            case .automatic where collectsTaxFromBillingAddress:
+                // Tax always needs the country; country overrides collect any additional fields required for that country.
+                return .country
+            case .full:
+                return .all
+            case .automatic, .never:
```
yuki-stripe: This is a nit but I'm also paranoid we could change line 971 and introduce a bug, can we do `case .automatic where !requiredByPaymentMethod, .never`

yuki-stripe (commented):
SO CLEAN!!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
porter-stripe: Not a bad idea, but when I add the case the complier complains that we still need to be exhaustive and handle the .automatic case and return nil which feels about equal. I think testing should catch this 🤷

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/STPApplePayContext+PaymentSheet.swift
porter-stripe: It's hard to tell exactly what web does in the code, I can test it out. But in our case, we need to collect the max that we could possibly need b/c the user could update the country from CA to US, and we don't have a way to show more fields if a user changes country in Apple Pay like we do ASE. Lmk and I can dig in more but I think in our situation we have to collect the absolute max we could need for any country b/c we can't react to country changes in Apple Pay.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetFormFactory/PaymentSheetFormFactory.swift
porter-stripe: Nvm, added!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/STPApplePayContext+PaymentSheet.swift
yuki-stripe: Yeah the other option is to collect less info and rely on geo IP or whatever fallback. I'm just worried about the conversion hit for collecting full billing, is that misplaced?

yuki-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/STPApplePayContext+PaymentSheet.swift
porter-stripe: I think the risk is low-ish, I assume most Apple Pay cards already have a full billing details. I'll look into it more tho.

porter-stripe: Checked the web code for EwCS. They always request full billing from Apple Pay, it's not even gated on automatic tax. billingAddressRequired: true is hardcoded in the click handler: https://stripe.sourcegraphcloud.com/stripe-internal/mint@6e7c51789742cec82a854183be7f5e63ebbe2004/-/blob/pay-server/stripe-js-v3/src/lib/hostedPaymentUi/modules/paymentRequest/components/ElementsExpressCheckoutElement.tsx?L493

Their tax min-field stuff (country + postal for non-US, full for US/PR) is only for the form-based card path. Apple Pay is all or nothing at the API level so you can't request less.

PR #6756, opened 2026-07-21
Fix Alipay EVO redirect return URL

porter-stripe:
## Summary
For merchants migrated to Alipay+ (EVO), the `alipay_handle_redirect` next action's `return_url` is an intermediate pm-redirects.stripe.com trampoline, not the app's actual return URL. We were passing that straight through as the redirect's return URL, so the SDK never recognized the final redirect back into the app and reported failure even though the payment succeeded.

- Use the confirm-time return URL (`currentAction.returnURLString`) as the expected redirect callback when it's set, falling back to `alipayHandleRedirect.returnURL` for merchants not on EVO
- Added a UI test (`testAlipayEVO`) against a China merchant enrolled in the EVO migration gate
- Added CNY/CN cases to the playground settings enums to support the new test
- Added unit tests covering both the confirm-time URL and legacy fallback paths

https://github.com/stripe/stripe-android/pull/13544

## Motivation
Alipay EVO migration

## Testing
- New unit tests
- New UI test

## Changelog
See diff

### DIFF
--- a/StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
+++ b/StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
@@ -1146,10 +1146,14 @@ public class STPPaymentHandler: NSObject {
 
         case .alipayHandleRedirect:
             if let alipayHandleRedirect = authenticationAction.alipayHandleRedirect {
+                // Post-EVO Alipay's next-action return URL is an intermediate tramopline.
+                // Register the confirm-time return URL as the expected callback instead.
+                let returnURL = currentAction.returnURLString.flatMap(URL.init(string:))
+                    ?? alipayHandleRedirect.returnURL
                 _handleRedirect(
                     to: alipayHandleRedirect.nativeURL,
                     fallbackURL: alipayHandleRedirect.url,
-                    return: alipayHandleRedirect.returnURL,
+                    return: returnURL,
                     useWebAuthSession: false
                 )
             } else {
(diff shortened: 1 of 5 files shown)


StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
```diff
@@ -1146,10 +1146,14 @@ public class STPPaymentHandler: NSObject {
 
         case .alipayHandleRedirect:
             if let alipayHandleRedirect = authenticationAction.alipayHandleRedirect {
+                // Post-EVO Alipay's next-action return URL is an intermediate tramopline.
+                // Register the confirm-time return URL as the expected callback instead.
+                let returnURL = currentAction.returnURLString.flatMap(URL.init(string:))
```
davidme-stripe, line 1151: Doesn't this skip the trampoline? Should we follow the trampoline instead to notify Stripe, or will it work to go directly to the URL?

porter-stripe: Good question, I don't think this should skip the trampoline. We still open the unchanged nativeURL. The returnURL passed here is only used to register a listener with `STPURLCallbackHandler`, it doesn't change the redirect destination. So the trampoline should still be hit, this only changes which URL we recognize when the redirect returns to the app.

davidme-stripe: Ah, I misread. Carry on then!

davidme-stripe (approved):

PR #6762, opened 2026-07-22
Add Checkout default email, updateEmail

yuki-stripe:
## Summary

- Add `updateEmail()`
- Add `configuration.defaults.email`. Defaults and `updateEmail()` store email locally; neither can override server email (`customer_email` or `Customer.email`).
- PaymentElement email fields are prefilled from local or server email. Only email collected by the payment method goes onto PM billing details.
- Checkout confirm sends local email via `collected_information[email]`, omitting it when server email exists. Without a PM, the legacy path sends `customer_data[email]` from `session.email`.

### Unit tests

- Defaults and `updateEmail()` set/clear local email and publish Session updates; queued updates preserve the latest value.
- Server-email decoding and Payment Element email prefill, including changes after `updateEmail()`.
- Confirmation sends local email separately from PM billing email. Form-collected PM email is preserved; Checkout email is not injected into new PMs, Link, or Apple Pay.

### Functional tests

- Load `Customer.email` as server email; confirm cards without PM billing email using either `Customer.email` or Checkout Session `customer_email`.
- Confirm SEPA Debit with `customer_email` and a different form-collected PM email.
- Confirm $0 sessions: server email with no PM, and local default email with a card.

### UI test

- Existing checkout synchronization test passes using `Customer.email`: AP localization, currency switching, shipping/tax updates, and card confirmation.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Configuration.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Configuration.swift
@@ -120,6 +120,8 @@ extension Checkout.Configuration {
         public var shippingDetails: ShippingDetails?
 
         /// The customer's email address.
+        /// This cannot be set if the Checkout Session was created with `customer_email`
+        /// or the Checkout Session's Customer already has an email.
         public var email: String?
 
         /// Creates default customer details.
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Configuration.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Configuration.swift
@@ -119,6 +119,9 @@ extension Checkout.Configuration {
         /// Default shipping details.
         public var shippingDetails: ShippingDetails?
 
+        /// The customer's email address.
+        public var email: String?
+
         /// Creates default customer details.
         public init() {}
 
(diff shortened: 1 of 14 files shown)


StripePaymentSheet/StripePaymentSheet/Source/Internal/API Bindings/STPAPIClient+CheckoutSession.swift
```diff
@@ -93,7 +110,8 @@ extension CheckoutSessionConfirmationRequestParameters {
             returnURL: configuration.returnURL,
             shipping: STPPaymentIntentShippingDetailsParams(paymentSheetConfiguration: configuration),
             paymentMethodOptions: paymentMethodOptions,
-            clientAttributionMetadata: clientAttributionMetadata
+            clientAttributionMetadata: clientAttributionMetadata,
+            collectedInformation: .init(email: checkoutSession.localState.email)
```
porter-stripe, line 114: What happens if we set a local email, then runServerUpdate adds an email to the Customer? Would we still send the stale local email rather than the server email?

yuki-stripe: Mmm interesting, that's not something a merchant should ever be doing. We'd still send the 'stale' local email (debatable if it's stale, it's not clear why the merchant would do this), and IDK what else would happen - maybe something eventually errors on the backend.

porter-stripe (approved):

PR #6765, opened 2026-07-22
Add Checkout default shipping details

yuki-stripe:
## Summary
- Add `Checkout.Configuration.defaults.shippingDetails` for default shipping name/address.
- Apply default shipping during Checkout initialization before PaymentElement loads, which can update tax_region.
- Align `Checkout.Address` model with Checkout Session/EwCS name+address name + shape.

## Motivation
Checkout Configuration already supports billing defaults. This adds the matching default shipping address path.
API Review: https://docs.google.com/document/d/1mi1xmUzONmuX0tUwjDQ1PUBZZNWgaEuWTje2V1maXuE/edit?tab=t.0 

## Testing
- `ci_scripts/run_tests.rb --test StripePaymentSheetTests/CheckoutAddressMergingTests --test StripePaymentSheetTests/CheckoutUnitTests --test StripePaymentSheetTests/CheckoutTests --test StripePaymentSheetTests/CheckoutDefaultsInitializationTests --test StripePaymentSheetTests/PaymentElementTest`

## Changelog
Not added; Checkout APIs are SPI/private preview scoped.


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/ContactAddress.swift
```diff
@@ -1,32 +1,83 @@
...
+        ///   - address: The customer's shipping address.
+        public init(name: String? = nil, address: Address) {
             self.name = name
-            self.phone = phone
             self.address = address
         }
     }
+
+    /// A postal address used by Checkout's billing and shipping address APIs.
+    public struct Address: Equatable {
```
gbirch-stripe: Why are we adding this in addition to the existing Checkout.Address struct in Address.swift?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Session+AddressMerging.swift
```diff
@@ -11,10 +11,13 @@ import Foundation
...
 
-    /// Populates empty fields in the configuration with checkout-collected addresses.
-    /// Configuration values always take precedence over checkout-collected values.
+    /// Applies Checkout Session customer details to a PaymentElement configuration.
+    /// Copies `session.shippingAddress` into `configuration.shippingDetails`,
+    /// sets the default billing email from `session.email` when the configuration
+    /// has none, and requires full billing address collection when the Checkout
+    /// Session requires it.
     func applyAddressOverrides<C: PaymentElementConfiguration>(to configuration: inout C) {
-        if let shipping = shippingAddress, configuration.shippingDetails() == nil {
```
gbirch-stripe, line 17: Makes sense that we should no longer prioritize the MPE config's shipping details! Should we also make setting shipping details on the config invalid / an error? We will always be ignoring it and using the Checkout-level data as the source of truth

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/ContactAddress.swift
yuki-stripe: Good catch, let's just use `Checkout.Address`. I'll delete this.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Session+AddressMerging.swift
yuki-stripe: Yeah we should stpassert here, it's our error if that happens.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Session+AddressMerging.swift
```diff
@@ -17,6 +17,7 @@ extension Checkout.Session {
     /// has none, and requires full billing address collection when the Checkout
     /// Session requires it.
     func applyAddressOverrides<C: PaymentElementConfiguration>(to configuration: inout C) {
+        stpAssert(configuration.shippingDetails() == nil, "Nothing should set configuration.shippingDetails, we overwrite it unconditionally to match checkout.session.shipping")
```
gbirch-stripe: Can we throw an error or assert at load time with merchant-friendly messaging? I don't think we actually prevent the merchant from setting shipping details on the configuration today, and it feels like it should be in the same category as [these restrictions](https://github.com/stripe/stripe-ios/blob/7bf464fd2bf327adfa37cf4cb919b94f0fbcde98/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetLoader.swift#L83-L90) (setting customer on config, etc.).

yuki-stripe: Ah, merchants don't have access to MPE's config. The proposed API is pretty different from our world a few weeks ago.

yuki-stripe: We should remove those asserts, or at least - they're not integration errors, they'd be our own fault since we set MPE config based on `CheckoutController.Configuration.PaymentElementConfiguration`.

gbirch-stripe: Got it, that makes sense then! My original concern was mostly about disallowing merchants to set a default address other than the Checkout one

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Configuration.swift
```diff
@@ -74,10 +73,19 @@ extension Checkout.Configuration {
...
+        /// Default shipping details.
+        public struct ShippingDetails {
+            /// The customer's full name.
+            public var name: String?
+
+            /// The customer's shipping address.
+            public var address: Checkout.Address?
+
+            /// Creates default shipping details.
+            public init() {}
```
gbirch-stripe, line 88: What's the purpose of having separate `ShippingDetails` and `Checkout.Address` structs? Are there situations where we need to specifically exclude `name` from a data structure?

yuki-stripe: No particular reason except that it matches EwCS, I think. It makes shipping/billing more future proof if they diverge somehow.

Is there a benefit to putting name in `Checkout.Address`?  I guess it makes it marginally easier to set both default billing and shipping to the same thing?

gbirch-stripe: Makes sense — wasn't thinking about how `Checkout.Address` can be reused by both billing and shipping

gbirch-stripe (approved):

PR #6766, opened 2026-07-22
ECE scaffolding

joyceqin-stripe:
## Summary

Adding basic ECE shape

Other PRs planned to render the buttons, add wallet visibility configurations, and appearance before trying to wire up Apple Pay/Link functionality.

ExpressCheckoutElement will be accessed by users via checkout.getExpressCheckoutElement(), similar to PaymentElement. I've added a basic ApplePayConfiguration and LinkConfiguration based on the API review doc.

Generally referencing CurrencySelectorElement to inform ECE patterns
## Motivation

ECE
## Testing

N/A
## Changelog

N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Express Checkout Element/ExpressCheckoutElement.swift
```diff
@@ -0,0 +1,46 @@
...
+
+import SwiftUI
+
+/// A SwiftUI view that displays wallet payment buttons (Apple Pay, Link) for express checkout.
+///
+/// Place this view on your cart or checkout page to let customers pay quickly using their
+/// saved payment methods.
+///
+/// ```swift
+/// ExpressCheckoutElement(checkout: checkout)
```
yuki-stripe: It's hard for me to give feedback on this since I assume it's not what we intend to ship necessarily but this isn't what I'd expect the API to look like, users won't initialize ECE directly, they'll get it through CheckoutController (the class called Checkout today).

yuki-stripe (commented):
Can you give more context on your plans? I assume this is some sort of prototype to start implementation but the final API shape is TBD?  If so you should explain in the PR what parts are to be reviewed and what parts I should ignore (the API?).

Can you also give more context on what the overall design is here, assuming that's in scope for review? e.g. is the SwiftUI story meant to copy an existing pattern?

joyceqin-stripe:
Yeah the docstring was misleading 😅 I've updated it.

This PR is just to add basic files/structs to build off of. I have other PRs planned to render the buttons, add wallet visibility configurations, and appearance before trying to wire up Apple Pay/Link functionality.

ExpressCheckoutElement will be accessed by users via `checkout.getExpressCheckoutElement()`, similar to PaymentElement. I've added a basic ApplePayConfiguration and LinkConfiguration based on the API review doc.

The SwiftUI story is meant to copy CurrencySelectorElement. If there's something wrong with that approach, feel free to let me know— my understanding was that CurrencySelectorElement, ECE, and Shipping Address Element would follow similar patterns.

joyceqin-stripe:
I see there's a [PR](https://github.com/stripe/stripe-ios/pull/6773) that changes the CurrencySelectorElement pattern to match PaymentElement more— I'm going to convert this back to draft and do the same

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
```diff
@@ -59,6 +59,9 @@ public final class Checkout: ObservableObject {
     /// The PaymentElement for this Checkout instance.
     private(set) var paymentElement: PaymentElement!
 
+    /// The ExpressCheckoutElement for this Checkout instance.
+    private(set) var expressCheckoutElement: ExpressCheckoutElement!
```
yuki-stripe: I'm planning to change this w/ PaymentElement, not great to have force unwraps and also I don't think we want to give it the entire checkout instance. It's probably better to give it some `ECEDelegate` for any Checkout methods it needs + probably the `$session`.

joyceqin-stripe: Can you share any more on planned changes for PaymentElement patterns?

yuki-stripe: Check out what @porter-stripe did with AP

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Express Checkout Element/ExpressCheckoutElement.swift
```diff
@@ -0,0 +1,43 @@
...
+
+    init(
+        sessionSource: CheckoutSessionSource,
+        configuration: Checkout.Configuration,
+        delegate: ExpressCheckoutElementDelegate
+    ) {
+        let uiView = ExpressCheckoutElementUIView(session: sessionSource.initialSession, configuration: configuration, delegate: delegate)
+        let viewModel = ExpressCheckoutElementViewModel(sessionSource: sessionSource, configuration: configuration, uiView: uiView)
+        self.uiView = uiView
+        self.view = ExpressCheckoutElementView(viewModel: viewModel)
```
yuki-stripe, line 41: Can you explain the design here at a high level? The UIView, View, ViewModel, and ECE. The structure confuses me a bit, looks like View owns ViewModel owns UIView? 

```
View
  - VM
    - UIView
```

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
```diff
@@ -133,6 +136,14 @@ public final class Checkout: ObservableObject {
 
             // Load elements
             self.paymentElement = try await PaymentElement(checkout: self)
+            self.expressCheckoutElement = ExpressCheckoutElement(
+                sessionSource: CheckoutSessionSource(
+                    initialSession: session,
+                    sessionPublisher: $session
+                ),
```
yuki-stripe: nit can we DRY `CheckoutSessionSource`? Do they need to be separate?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Express Checkout Element/ExpressCheckoutElement.swift
joyceqin-stripe: It's the same structure as CurrencySelectorElement (and I believe it matches PaymentElement as well). I think the idea is that UIKit's UIView and the SwiftUI View both need to observe and update changes from the session publisher, so we create one UIView and wire it so that both callers (UIKit and SwiftUI) can share it. The view model can help UIKit and SwiftUI observe session changes via uiView.update and observing isAvailable

yuki-stripe (approved):

PR #6767, opened 2026-07-23
Allow billing details collection configuration in LinkController

mats-stripe:
## Summary

This adds `billingDetailsCollectionConfiguration` as a field to the LinkConfiguration object to allow merchants specify which billing details they'd like to collect. It uses the existing `PaymentSheet.BillingDetailsCollectionConfiguration` object as the type.

## Motivation

https://stripe.slack.com/archives/C08G211PLER/p1784815838646309

## Testing

I set all the parameters to `.always` in the demo app and see this after selecting a payment method:

Simulator Screenshot - iPhone 17 Pro - 2026-07-23 at 15 41 35

## Changelog

```
### PaymentSheet
* [Added] Added `billingDetailsCollectionConfiguration` to `LinkConfiguration` (private preview).
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,10 @@
 The next release's version bump will so far be:
-PATCH
+MINOR
 
 ## X.Y.Z - changes pending release
+
+### PaymentSheet
+* [Added] Added `billingDetailsCollectionConfiguration` to `LinkConfiguration`, allowing `LinkControllerPreview` consumers to configure billing details collection in the Link sheet (private preview).
 
 ## 26.4.1 2026-07-23
 ### Payments
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
@@ -683,7 +683,21 @@ extension PaymentSheet {
             }
         }
 
-        public init() {}
+        public init(
+            name: CollectionMode = .automatic,
+            phone: CollectionMode = .automatic,
+            email: CollectionMode = .automatic,
+            address: AddressCollectionMode = .automatic,
+            attachDefaultsToPaymentMethod: Bool = false,
+            allowedCountries: Set<String> = []
+        ) {
+            self.name = name
+            self.phone = phone
+            self.email = email
+            self.address = address
+            self.attachDefaultsToPaymentMethod = attachDefaultsToPaymentMethod
+            self.allowedCountries = allowedCountries
+        }
     }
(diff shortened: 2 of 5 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
```diff
@@ -682,6 +682,8 @@ extension PaymentSheet {
                 allowedCountries = Set(allowedCountries.map { $0.uppercased() })
             }
         }
+
+        public init() {}
```
mats-stripe: I had to add a public initializer to this already public struct in order for instances of it to be passed in to LinkController. The other public structs here in this class already have a public initializers. Let me know if there's any FUD with this approach!

Image

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
```diff
@@ -682,6 +682,22 @@ extension PaymentSheet {
                 allowedCountries = Set(allowedCountries.map { $0.uppercased() })
             }
         }
+
+        public init(
```
tillh-stripe, line 686: Why is this now necessary?

CHANGELOG.md
```diff
@@ -1,8 +1,11 @@
 The next release's version bump will so far be:
-PATCH
+MINOR
 
 ## X.Y.Z - changes pending release
 
+### PaymentSheet
+* [Added] Added `billingDetailsCollectionConfiguration` to `LinkConfiguration`, allowing `LinkControllerPreview` consumers to configure billing details collection in the Link sheet (private preview).
```
tillh-stripe, line 7: Do we need to make this a minor bump for a private-preview feature?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/PaymentSheetConfiguration.swift
mats-stripe: https://github.com/stripe/stripe-ios/pull/6767#discussion_r3645819095

CHANGELOG.md
mats-stripe: Yes! See this comment: https://github.com/stripe/stripe-ios/pull/6767#issuecomment-5061262677

>SPI API changes (other than declarations using only @_spi(STP) and/or @_spi(ReactNativeSDK)) require a MINOR or MAJOR version bump.

The `Verify public interface` CI step will fail otherwise

tillh-stripe (approved):

PR #6770, opened 2026-07-23
Render ECE buttons

joyceqin-stripe:
## Summary

Render Apple Pay and Link buttons when included on the session and add ECE to the playground. Allow standalone ECE or with PE
## Motivation

ECE
## Testing

image
image

## Changelog

N/A


Example/PaymentSheet Example/PaymentSheet Example/Checkout Playground/CheckoutPlaygroundTypes.swift
```diff
@@ -124,13 +124,29 @@ enum CheckoutPlayground {
     enum IntegrationType: String, CaseIterable, Identifiable {
         case flowController
         case embedded
+        case disabled
```
porter-stripe: what does integration type disabled mean?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Express Checkout Element/ExpressCheckoutElementUtilities.swift
```diff
@@ -5,9 +5,21 @@
...
 //
 
+@_spi(STP) import StripeCore
+
 enum ExpressCheckoutElementUtilities {
     static func resolveButtons(for session: Checkout.Session, configuration: Checkout.Configuration) -> [ExpressButton] {
-        // TODO: Compute from session
-        return []
+        var buttons: [ExpressButton] = []
+        for button in session.availableExpressButtonTypes {
```
porter-stripe, line 13: Few questions on the availability checks:
1. Do we need to check elementsSession.isApplePayEnabled?
2. Do we need to check configuration.linkConfiguration.display?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Express Checkout Element/ExpressCheckoutElementUIView.swift
```diff
@@ -13,20 +15,96 @@ import UIKit
...
+        // TODO: Appearance
+        button.cornerRadius = 6
+        button.translatesAutoresizingMaskIntoConstraints = false
+        button.heightAnchor.constraint(equalToConstant: 44).isActive = true
+        button.addTarget(self, action: #selector(handleApplePayTapped), for: .touchUpInside)
+        return button
+    }
+
+    private func makeLinkButton() -> UIView {
+        let button = PayWithLinkButton()
```
porter-stripe: Do we to pass in the brand like wallet header view? Do we need to use one link brand?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Express Checkout Element/ExpressCheckoutElementUtilities.swift
joyceqin-stripe: Yep, I was going to add the wallet visibility check on a later pr but it makes sense to add it in this one. Added the elementsSession.isApplePayEnabled check in availableExpressButtonTypes— the separation I was thinking for availability logic is that availableExpressButtonTypes is the list of available buttons based on the session/server-side things and in resolveButtons we consider configs (including, later, ECE wallet visibility configs)

Example/PaymentSheet Example/PaymentSheet Example/Checkout Playground/CheckoutPlaygroundTypes.swift
joyceqin-stripe: That wasn't a great name— renamed to "none", but the purpose is so that we can hide PE and allow ECE-only in the playground too

porter-stripe: nit: Can we name it exactly what it is? I think none is also confusing. Just eceOnly or something?

porter-stripe (approved):

PR #6773, opened 2026-07-24
Move AP to Checkout Controller

porter-stripe:
## Summary
- Moves AP to Checkout Controller

## Motivation
- CheckoutSessions

## Testing
- CI

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Currency Selector Element/CurrencySelectorElementUIView.swift
```diff
@@ -0,0 +1,235 @@
...
+@_spi(STP) import StripeCore
+@_spi(STP) import StripePayments
+@_spi(STP) import StripeUICore
+import UIKit
+
+/// A UIKit view that displays an Adaptive Pricing currency selector.
+@_spi(STP)
+@_spi(ReactNativeSDK)
+@MainActor
+public final class CurrencySelectorElementUIView: UIView {
```
porter-stripe, line 17: Not interesting, copied from master

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Currency Selector Element/CurrencySelectorElementView.swift
```diff
@@ -0,0 +1,61 @@
...
+//  Created by Nick Porter on 7/22/26.
+//
+
+import Combine
+import SwiftUI
+
+/// A SwiftUI view that displays an Adaptive Pricing currency selector.
+@_spi(STP)
+@_spi(ReactNativeSDK)
+public struct CurrencySelectorElementView: View {
```
porter-stripe, line 14: Not interesting, copied from master

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Checkout/CurrencySelectorElementViewTests.swift
```diff
@@ -0,0 +1,75 @@
...
+//  Created by Nick Porter on 7/22/26.
+//
+
+@testable @_spi(STP) import StripePaymentSheet
+import SwiftUI
+import UIKit
+import XCTest
+
+@MainActor
+final class CurrencySelectorElementViewTests: XCTestCase {
```
porter-stripe, line 14: Not interesting, copied from master

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
```diff
@@ -127,7 +129,9 @@ public final class Checkout: ObservableObject {
 
             // Load elements
             self.paymentElement = try await PaymentElement(checkout: self)
-            await flagImageManager.prefetchFlagImages(for: session) // TODO: This should probably just load currency selector and not be a global singleton
+            if configuration.adaptivePricing.allowed {
```
porter-stripe, line 136: I'm not sure what our plan is for lazy loading elements, but I feel like this is reasonable if the user has explicitly enabled AP to always load the selector.

porter-stripe: I'm on the fence about the "getXElement" API shape for AP and future elements, since this one may be optional (data may not exist to construct it)

porter-stripe:
The snapshot diffs are not related to these changes. It is b/c our stubbing changed and we now 400 for the CDN to download the flag assets in these. I think thats fine for snapshots.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
yuki-stripe: Is there a better API shape? I don't see how we can avoid an optional, since whether or not CSE exists is only known at run-time.

porter-stripe: That's probably true! I think the previous API shape had this issue where the init could return nil if the required data wasn't on the CS.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Currency Selector Element/CurrencySelectorElementSessionSource.swift
```diff
@@ -0,0 +1,14 @@
...
+//
+//  Created by Nick Porter on 7/27/26.
+//
+
+import Combine
+
+/// Provides Checkout Session state without exposing Checkout itself.
+struct CurrencySelectorElementSessionSource {
+    let initialSession: Checkout.Session
+    let sessionPublisher: Published<Checkout.Session>.Publisher
```
yuki-stripe: Huh this is unfortunately clunky, thought we could just pass $session.

WDYT about renaming it to be more generalized since it has nothing to do with CurrencySelectorElement? `CheckoutSessionSource`?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
```diff
@@ -59,6 +59,9 @@ public final class Checkout: ObservableObject {
     /// The PaymentElement for this Checkout instance.
     private(set) var paymentElement: PaymentElement!
 
+    /// The CurrencySelectorElement for this Checkout instance, when Adaptive Pricing is available.
+    private var currencySelectorElement: CurrencySelectorElement?
```
yuki-stripe, line 63: Can we make this a `let` and assign the delegate after it's initialized

porter-stripe: This is tricky b/c the complier complains that we are using `self` before it is initialized and throws a bunch of errors.

yuki-stripe: Oh, because of the self.session usage in addition to the delegate. Hm ok. I wonder if we should actually just call `currencySelectorElement.update(with: session)` in `session.didSet` rather than make the `Source` thing and deal with this var. 

It's possible we will want Checkout to have the ability to coordinate the ordering of updates or insert logic (eg don't update element x if y).

yuki-stripe (dismissed):

yuki-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
porter-stripe: Ah that might work, like all elements inside of the session.didSet? Would maybe work for all besides the initialization case.

PR #6776, opened 2026-07-24
Sync initial saved payment option's billing address to Checkout tax region

porter-stripe:
## Summary
- When `PaymentElement` loads `Checkout`'s initial payment option from Embedded on load, also sync that payment option's billing address to the Checkout Session's automatic tax region (via `Checkout.syncBillingAddress(from:)`).
- This keeps a saved card's billing address in sync with automatic tax on first load, matching the behavior already applied on subsequent payment option changes.

## Test plan
- New tests

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Payment Element/PaymentElement.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Payment Element/PaymentElement.swift
@@ -112,6 +112,7 @@ public final class PaymentElement {
             embeddedPaymentElement.paymentOption.map(Checkout.Session.PaymentOptionDisplayData.init)
         )
         paymentOptionSourceOfTruthIsFlowController = false // We used embedded's payment option
+        try await checkout.syncBillingAddress(from: embeddedPaymentElement._paymentOption?.checkoutBillingDetails)
     }
 }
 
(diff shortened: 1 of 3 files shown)


gbirch-stripe (commented):
What happens if the merchant loads the Checkout Controller + PE and then updates the billing address themselves (via runServerUpdate) — we're back to being out of sync, right?

porter-stripe:
@gbirch-stripe the merchant is not able to update billing address on the server for two reasons
1. billing address is more of a client side concept and it actually maps to tax_region on the backend
2. Merchants can't update tax_region themselves: https://docs.stripe.com/api/checkout/sessions/update

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Payment Element/PaymentElement.swift
```diff
@@ -112,6 +112,7 @@ public final class PaymentElement {
             embeddedPaymentElement.paymentOption.map(Checkout.Session.PaymentOptionDisplayData.init)
         )
         paymentOptionSourceOfTruthIsFlowController = false // We used embedded's payment option
+        try await checkout.syncBillingAddress(from: embeddedPaymentElement._paymentOption?.checkoutBillingDetails)
```
yuki-stripe, line 115: Ideally checkout just has code like 
```
func setPaymentOption() {
  ...
  // update our billing addr
  self.syncBillingAddress(..)
}
```
that way we eliminate the possibility PaymentOption is updated _without_ syncing billing addr, but understand that due to MPE's existing implementation etc it's not possible/practical.

yuki-stripe (approved):

PR #6777, opened 2026-07-24
Require return URL for Checkout configuration

yuki-stripe:
## Summary
- Add a required `returnURL` to `Checkout.Configuration`.
- Add debug-time validation that the return URL can be routed through `StripeAPI.handleURLCallback(with:)`, and that custom URL schemes are registered in `CFBundleURLTypes`.
- Update Checkout playground and existing Checkout call sites/tests to pass a return URL.
- Leave MPE config and CheckoutSession confirm wiring unchanged.

## Motivation
Checkout needs to capture the app return URL at configuration initialization time so follow-up Checkout confirm handling can use it separately. The debug validation catches malformed or unregistered return URLs earlier during integration.

## Testing
- `ci_scripts/run_tests.rb --test StripePaymentSheetTests/PaymentElementTest/testConfigurationSetsCheckoutDefaultBillingDetails`
- `ci_scripts/format_modified_files.sh`
- `ci_scripts/lint_modified_files.sh`

## Changelog
Not added; Checkout is SPI and this only updates the configuration API.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Configuration.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Configuration.swift
@@ -56,6 +56,7 @@ extension Checkout {
             self.returnURL = returnURL
         }
 
+        // MARK: - Debug-only return URL validation
 #if DEBUG
         /// Debug-only listener used to verify Checkout return URLs can be routed through
         /// `StripeAPI.handleURLCallback(with:)`.
@@ -72,10 +73,8 @@ extension Checkout {
                 return true
             }
         }
-#endif
 
         func validateReturnURL() {
-            #if DEBUG
             guard let url = URL(string: returnURL),
                   let scheme = url.scheme,
                   !scheme.isEmpty else {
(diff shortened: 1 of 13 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Configuration.swift
```diff
@@ -44,9 +50,61 @@ extension Checkout {
...
         /// - Parameter clientSecret: The client secret for your Checkout Session.
-        public init(clientSecret: String) {
+        /// - Parameter returnURL: A custom URL scheme that redirects back to your app after authenticating a payment method, e.g. `my-app://stripe-redirect`. Register this URL scheme in your app and forward incoming URLs to `StripeAPI.handleURLCallback(with:)`.
+        public init(clientSecret: String, returnURL: String) {
             self.clientSecret = clientSecret
+            self.returnURL = returnURL
+        }
+
+        // MARK: - Debug-only return URL validation
+#if DEBUG
```
yuki-stripe, line 60: I thought this would be neat to do. It's way nicer than having to debug 'why isn't it redirecting back to my app?' manually.

joyceqin-stripe (approved):
Debug URL validation is cool!

PR #6780, opened 2026-07-24
Default FlowController to Link wallet button

yuki-stripe:
## Summary
- Default FlowController to Link when Link is the customer default but is rendered as a wallet/header button instead of a selectable row. 
- Covers both horizontal and vertical mode.
- **Note** that Link is only the customer default when it was previously selected, which in this scenario where Link is a wallet button means you successfully confirmed with Link (merely tapping the Link button won't do it).

https://github.com/user-attachments/assets/4ead1348-9c65-43ec-8a0f-675861e2967f

https://github.com/user-attachments/assets/3d5c1d98-21fe-412b-97e6-a7f20e6599cb

## Motivation
PaymentElement expects FlowController and Embedded to agree on the default payment option. If they don't, since we don't know which one the merchant is using, we can't know what the default payment option is.

In the scenario of the PR, FlowController reported payment option = nil while Embedded reported payment option = link. 

**Embedded where Link is the default**
CleanShot 2026-07-24 at 17 08 36

## Testing
- ci_scripts/format_modified_files.sh
- ci_scripts/lint_modified_files.sh
- ci_scripts/run_tests.rb --test StripePaymentSheetTests/PaymentSheetFlowControllerTests
- ci_scripts/run_tests.rb --test StripePaymentSheetTests/PaymentSheetVerticalViewControllerTest
- ci_scripts/run_tests.rb --test StripePaymentSheetTests/PaymentElementTest

## Changelog
see changelog


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/PaymentSheetVerticalViewController.swift
```diff
@@ -573,13 +601,17 @@ class PaymentSheetVerticalViewController: UIViewController, FlowControllerViewCo
...
+        isLinkWalletButtonSelected = false
+        linkConfirmOption = nil
+        updatePrimaryButton()
     }
 
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
         logInitialDisplayedPaymentMethods()
-        isLinkWalletButtonSelected = false
-        linkConfirmOption = nil
```
yuki-stripe, line 582: Moved this to viewWillAppear so that we can sync the primary button with this state before UI is shown instead of after.

davidme-stripe (approved):

PR #6786, opened 2026-07-27
Make Checkout mode-agnostic

porter-stripe:
## Summary
- Removes `Checkout.Mode` and switches on `payment_status` instead via a new `isSetupStyle` flag.

First of a few PRs to support unified mode.

`payment_status == no_payment_required` is what EWCS itself uses as the setup-intent signal (https://stripe.sourcegraphcloud.com/stripe-internal/mint/-/blob/pay-server/stripe-js-v3/src/checkout/outer/customCheckout/actions/confirm/handleNextAction.ts?L267), and we don't support subscriptions, so it's a safe stand-in for mode.

## Motivation
CheckoutSessions

## Testing
Existing tests updated

## Changelog
N/A


porter-stripe:
The snapshot diffs are not related to these changes. It is b/c our stubbing changed and we now 400 for the CDN to download the flag assets in these. I think thats fine for snapshots.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/PaymentPagesAPIResponse.swift
```diff
@@ -65,8 +65,8 @@ class PaymentPagesAPIResponse: NSObject {
 
     // MARK: - Internal SDK-only fields
 
-    /// The mode of the Checkout Session (payment, setup, or subscription).
-    let mode: Checkout.Mode
+    /// The top-level payment status, which is present even when `status` is omitted.
```
yuki-stripe: huh can status actually be omitted and nil? I don't think so.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Session+Internal.swift
```diff
@@ -26,6 +26,11 @@ extension Checkout.Session {
     var collectsTaxFromBillingAddress: Bool {
         return shouldSendTaxRegion(for: "billing")
     }
+
+    /// Whether confirmation follows setup-style semantics.
+    var isSetupStyle: Bool {
```
yuki-stripe: ooc why 'style'?

yuki-stripe: Maybe we should just call this `noPaymentRequired`.

yuki-stripe (dismissed):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/PaymentPagesAPIResponse.swift
porter-stripe: yeah, confirmed in pay-server w/ claude.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Session+Internal.swift
porter-stripe: yeah, that's a better name. updated!

yuki-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/PaymentPagesAPIResponse.swift
porter-stripe: actually, good flag. checked with the server and it IS required: https://stripe.sourcegraphcloud.com/r/stripe-internal/mint/-/blob/pay-server/lib/payment_pages/api/payment_page/resources/payment_page_resource.rb?L84-95

PR #6797, opened 2026-07-29
Fix embedded automatic tax constraints

porter-stripe:
## Summary
When automatic tax is enabled and a saved payment method exists, EPE was mutating an already-built address section's minimum fields in place after it had already been laid out, then rebuilding its address fields on top of that. That produced broken Auto Layout constraints on load.

Now the tax-driven minimum fields are folded in before the address section is built, so it's constructed once with the right fields from the start. Removed the now-unused in-place mutation method on `AddressSectionElement` along with its `widened(toMeet:)` helper, and re-added that helper where it's actually used in `PaymentSheetFormFactory+Tax`.

## Motivation
Embedded

## Testing
- Existing tests

## Changelog
N/A


Example/PaymentSheet Example/PaymentSheetUITest/CheckoutSessionUITests.swift
```diff
@@ -27,6 +27,22 @@ class CheckoutSessionUITests: PaymentSheetUITestCase {
...
+        settings.integrationType = .checkoutSession
+        settings.customerMode = .returning
+        settings.csAutomaticTax = .on
+        loadPlayground(app, settings)
+
+        // When the embedded element is presented
+        app.buttons["Present embedded payment element"].waitForExistenceAndTap()
+
+        // Then building its payment method forms doesn't break any layout constraints
+        XCTAssertFalse(app.alerts["Broken constraint!"].waitForExistence(timeout: 1))
```
joyceqin-stripe: Is 1s enough/is there another signal or indicator we could use to make sure the form is loaded? Is it possible for the form building to take longer for some reason

porter-stripe: I didn't even mean to commit this test! We should remove it. It's pretty hard to test for this efficiently.

joyceqin-stripe (approved):

PR #6801, opened 2026-07-29
Add unified Checkout Session payload support

porter-stripe:
## Summary
- Supports Checkout Sessions returned in the unified payload shape. Maps checkout items, discounts, and taxes so sessions load, update shipping addresses, and use adaptive pricing as expected.
- Disables non card LPM flow tests

## Motivation
CheckoutSessions

## Testing
- Existing tests
- Updated

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/PaymentPagesAPIResponse.swift
```diff
@@ -491,60 +491,54 @@ extension PaymentPagesAPIResponse {
...
-            taxAmounts: lineTaxAmounts,
-            adjustableQuantity: adjustableQuantity
+            subtotal: nil,
+            discount: nil,
+            taxExclusive: nil,
+            taxInclusive: nil,
+            total: nil,
+            discountAmounts: [],
+            taxAmounts: [],
+            adjustableQuantity: nil
```
porter-stripe, line 541: These are nil/empty until we get the new public API implemented for our Session data model.

porter-stripe:
skipping dead flow test, the function is used but only in tests.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/PaymentPagesAPIResponse.swift
```diff
@@ -652,11 +632,19 @@ extension PaymentPagesAPIResponse {
         return Checkout.DeliveryEstimate.Bound(unit: unit, value: value)
     }
 
+    /// Reads an aggregate array from `recurring_details`. Despite its name, `recurring_details`
+    /// also contains one-time order totals.
```
yuki-stripe, line 636: odd!

StripePayments/StripePaymentsTestUtils/STPTestingAPIClient+Swift.swift
```diff
@@ -195,165 +195,41 @@ extension STPTestingAPIClient {
         let publishableKey: String
     }
 
-    static func checkoutSessionAdditionalParameters(
+    /// Creates a unified (modeless) Checkout Session backed by `checkout_items`.
+    func fetchCheckoutSession(
```
yuki-stripe: Can we call this `createCheckoutSession`? "Fetch" made me think this test was retrieving the CheckoutSession, and then I was wondering why the tests don't use the SDK's 'retrieve PP' method to fetch a CheckoutSession.

yuki-stripe (approved):

PR #6809, opened 2026-07-30
Sync Checkout billing from embedded saved methods

porter-stripe:
## Summary
When a customer picks or updates a saved payment method in embedded/vertical mode, we now sync its billing address to the Checkout session so automatic tax stays in sync.
- https://docs.google.com/document/d/1CWglcOHR6I8cOWKjztPfnhnVG8axucpv-j5cWIoey7E/edit?tab=t.fqk76h942k4r#heading=h.a2lhvmal87zq

#### Happy path

https://github.com/user-attachments/assets/2dc67f1d-70e5-455f-807b-fc0b06a23dfc

#### Error path with recovery

https://github.com/user-attachments/assets/65f496aa-b258-4dd3-879e-fa4aeae1c281

## Motivation
Checkout

## Testing
New tests

## Changelog
N/A

PR #6810, opened 2026-07-30
Sync Checkout billing from horizontal saved methods

porter-stripe:
## Summary
Extends the billing sync from #6809 to the horizontal (FlowController) saved payment method screen. Selecting a saved payment method there now syncs its billing address to the Checkout session for automatic tax before closing the sheet, and restores the previous selection if the sync fails. The FlowController confirm button only kicks off a sync when the selected payment option actually needs one, instead of always making a network call.

#### No CTA happy path

https://github.com/user-attachments/assets/c94cd181-9ee8-44d5-b1f2-2fa84ca3d39b

#### No CTA error then recovery
https://github.com/user-attachments/assets/dce98661-0de4-49ae-93ce-8b4591702f25
#### CTA happy path

https://github.com/user-attachments/assets/e2232058-d833-4378-920a-6c397ac410a7

#### CTA error then recovery

https://github.com/user-attachments/assets/9dea3b0e-ac12-4217-980e-d8bbcdf66cba

## Motivation
Checkout

## Testing
New tests

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentMethodCollectionView.swift
```diff
@@ -280,6 +292,22 @@ extension SavedPaymentMethodCollectionView {
             update()
         }
 
+        /// Replaces the selected checkmark with a spinner while loading.
+        func setLoading(_ loading: Bool) {
+            guard loading != spinner.isAnimating else {
+                return
+            }
+
+            selectedIcon.isHidden = !loading && !isSelected
```
gbirch-stripe: Should we be using `setHiddenIfNecessary` here? Usually that's the safer option to avoid weird edge cases and UI behavior

gbirch-stripe (commented):
Where did the error copy come from? "Failed to select try again" seems wrong (should be "Failed to select. Try again." or something like that)

porter-stripe:
@gbirch-stripe the copy was just something I hardcoded for demo purposes, the real error would come from a buyer friendly error: https://docs.google.com/document/d/1ExcnE4Wb7Svk2lp4tRP4A7Kqn5eXOvzOS2xohQkzYkE/edit?tab=t.dm2vlx61qbml#heading=h.7woxa655m2yz

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Saved Payment Method Screen/SavedPaymentMethodCollectionView.swift
porter-stripe: true, good idea!

gbirch-stripe (commented):
In the automatic tax case the billing address update will update the checkout session which will then trigger a flow controller update — in this case will the in-progress payment method selection go through properly, or will it get overwritten (or other weird behavior) by the FC reload?

porter-stripe:
@gbirch-stripe the automatic tax case is the one shown in the videos! The tax update does reload FC, but it's fine b/c the tapped SPM is persisted first, so the rebuilt VC restores that selection and publishes the correct payment option when the sheet dismisses.

gbirch-stripe (approved):

PR #6812, opened 2026-07-30
Rename autocomplete analytics params

joyceqin-stripe:
## Summary

Rename some params used for autocomplete
## Motivation

Match with Android, I don't have strong opinions but Android does so it'd be better for analytics to match
https://docs.google.com/document/d/1wNbAj7MTAHO6zg4mKAqTBGiDzR_7z0qlLgOITLmOt4Y/edit?tab=t.wjw4suuiahq8#heading=h.ug025we8zb0

## Testing

Updated tests
## Changelog

N/A


StripePaymentSheet/StripePaymentSheet/Source/Analytics/STPAnalyticsClient+Address.swift
```diff
@@ -70,35 +70,35 @@ extension STPAnalyticsClient {
         log(analytic: AddressAnalytic(event: .addressAutocompleteStart, params: [:]), apiClient: apiClient)
     }
 
-    func logAddressAutocompleteSuggestions(characterCount: Int, sessionToken: String, source: String, duration: TimeInterval, latency: TimeInterval?, apiClient: STPAPIClient) {
+    func logAddressAutocompleteSuggestions(queryLength: Int, autocompleteSessionToken: String, source: String, sessionElapsed: TimeInterval, timeToFetch: TimeInterval?, apiClient: STPAPIClient) {
         assert(apiClient.publishableKey?.nonEmpty != nil) // A publishable key is required to be set at this point so we can send it in our analytics payload
         var params: [String: Any] = [
-            "character_count": characterCount,
-            "session_token": sessionToken,
+            "query_length": queryLength,
```
wooj-stripe: Based on this [doc](https://docs.google.com/document/d/1wNbAj7MTAHO6zg4mKAqTBGiDzR_7z0qlLgOITLmOt4Y/edit?tab=t.wjw4suuiahq8#heading=h.ug025we8zb0):

for `mc_address_autocomplete_suggestions`, i see that params should include:
* autocomplete_session_token ✅ 
* source  ✅ 
* time_to_fetch ❌ 
* result_count ❌ 
* session_elapsed ✅

StripePaymentSheet/StripePaymentSheet/Source/Analytics/STPAnalyticsClient+Address.swift
```diff
@@ -70,35 +70,35 @@ extension STPAnalyticsClient {
...
+            "autocomplete_session_token": autocompleteSessionToken,
             "source": source,
-            "duration": duration,
+            "time_to_complete": timeToComplete,
         ]
         if let latency { params["latency"] = latency }
         log(analytic: AddressAnalytic(event: .addressAutocompleteComplete, params: params), apiClient: apiClient)
```
wooj-stripe: does .addressAutocompleteComplete  need to map to "mc_address_completed"?  it seems to map to  "mc_address_autocomplete_complete"

StripePaymentSheet/StripePaymentSheet/Source/Analytics/STPAnalyticsClient+Address.swift
```diff
@@ -70,35 +70,35 @@ extension STPAnalyticsClient {
...
     }
 
-    func logAddressAutocompleteError(error: Error, sessionToken: String, duration: TimeInterval, apiClient: STPAPIClient) {
+    func logAddressAutocompleteError(errorType: Error, autocompleteSessionToken: String, duration: TimeInterval, apiClient: STPAPIClient) {
         assert(apiClient.publishableKey?.nonEmpty != nil) // A publishable key is required to be set at this point so we can send it in our analytics payload
         log(analytic: AddressAnalytic(event: .addressAutocompleteError, params: [
-            "error": error.localizedDescription,
-            "session_token": sessionToken,
+            "error_type": errorType.localizedDescription,
+            "autocomplete_session_token": autocompleteSessionToken,
```
wooj-stripe: based on the doc, looks like we're supposed to include error_code and request_id?

joyceqin-stripe: Included in error.serializeForV1Analytics

StripePaymentSheet/StripePaymentSheet/Source/Analytics/STPAnalyticsClient+Address.swift
joyceqin-stripe: Looks like Android also changed the event names, addressAutocompleteComplete is now addressAutocompleteSelected "mca_address_autocomplete_selected" instead

StripePaymentSheet/StripePaymentSheet/Source/Analytics/STPAnalyticsClient+Address.swift
```diff
@@ -36,13 +36,14 @@ extension STPAnalyticsClient {
...
-    func logAddressCompleted(addressCountyCode: String, autoCompleteResultedSelected: Bool, editDistance: Int?, apiClient: STPAPIClient) {
+    func logAddressCompleted(addressCountyCode: String, autoCompleteResultedSelected: Bool, editDistance: Int?, timeToComplete: TimeInterval?, apiClient: STPAPIClient) {
         assert(apiClient.publishableKey?.nonEmpty != nil) // A publishable key is required to be set at this point so we can send it in our analytics payload
         let analyticData = AddressAnalyticData(addressCountryCode: addressCountyCode,
                                                autoCompleteResultedSelected: autoCompleteResultedSelected,
                                                editDistance: editDistance)
-
-        self.logAddressControllerEvent(event: .addressCompleted, addressAnalyticData: analyticData, apiClient: apiClient)
+        var additionalParams: [String: Any] = ["address_data_blob": analyticData.analyticsPayload]
+        if let timeToComplete { additionalParams["time_to_complete"] = timeToComplete }
```
wooj-stripe: I understand this is the name specified in the doc, but instead of "time_to_complete" a better name would be 'ms_to_complete' (same with "time_to_fetch" -> "ms_to_fetch")

StripePaymentSheet/StripePaymentSheet/Source/Analytics/STPAnalyticsClient+Address.swift
```diff
@@ -36,13 +36,14 @@ extension STPAnalyticsClient {
         self.logAddressControllerEvent(event: .addressShow, addressAnalyticData: analyticData, apiClient: apiClient)
     }
 
-    func logAddressCompleted(addressCountyCode: String, autoCompleteResultedSelected: Bool, editDistance: Int?, apiClient: STPAPIClient) {
+    func logAddressCompleted(addressCountyCode: String, autoCompleteResultedSelected: Bool, editDistance: Int?, timeToComplete: TimeInterval?, apiClient: STPAPIClient) {
```
wooj-stripe: Seems like timeToComplete should not be optional. Are there legitimate cases where timeToComplete can be nil?

StripePaymentSheet/StripePaymentSheet/Source/Analytics/STPAnalyticsClient+Address.swift
```diff
@@ -65,42 +66,44 @@ extension STPAnalyticsClient {
...
+    func logAddressAutocompleteSuggestions(resultCount: Int, sessionToken: String, source: String, sessionElapsed: TimeInterval, timeToFetch: TimeInterval?, apiClient: STPAPIClient) {
         assert(apiClient.publishableKey?.nonEmpty != nil) // A publishable key is required to be set at this point so we can send it in our analytics payload
         var params: [String: Any] = [
-            "character_count": characterCount,
-            "session_token": sessionToken,
+            "result_count": resultCount,
+            "autocomplete_session_token": sessionToken,
             "source": source,
-            "duration": duration,
+            "session_elapsed": sessionElapsed,
```
wooj-stripe: 1. I think ms_session_elapsed would be a better name
2. The doc doesn't call out needing `source` but, i suppose having additional data is fine?

StripePaymentSheet/StripePaymentSheet/Source/Analytics/STPAnalyticsClient+Address.swift
```diff
@@ -65,42 +66,44 @@ extension STPAnalyticsClient {
...
+            "query_length": queryLength,
+            "autocomplete_session_token": sessionToken,
             "source": source,
-            "duration": duration,
+            "session_elapsed": sessionElapsed,
         ]
-        if let latency { params["latency"] = latency }
-        log(analytic: AddressAnalytic(event: .addressAutocompleteComplete, params: params), apiClient: apiClient)
+        if let placeId { params["place_id"] = placeId }
+        if let timeToFetch { params["time_to_fetch"] = timeToFetch }
```
wooj-stripe: Again, is `timeToFetch` really optional?

wooj-stripe: I noticed that AutoCompleteViewController has:
```
    private var elapsedTimeSinceAutocompleteStart: TimeInterval {
        guard let startTime = autocompleteStartTime else { return 0 }
        return Date().timeIntervalSince(startTime)
    }
```

 I think we need to find a better solution here to avoid optionals / awkward looking code. Maybe we:

```
class AutoCompleteViewcontroller {
     private var autocompleteStartTime: Date = Date()

.... snip...

    override func viewWillAppear(_ animated: Bool) {
        super.viewWillAppear(animated)
        autocompleteStartTime = Date()
    }   
}
```

I think the reason this works is because all the metrics from this class only get sent after viewDidAppear().


Maybe we can do something similar for `mapKitQueryStartTime` in this class but i haven't looked too deeply.

StripePaymentSheet/StripePaymentSheet/Source/Analytics/STPAnalyticsClient+Address.swift
joyceqin-stripe: Done, renamed to `ms_to_complete` and `ms_to_fetch`.

StripePaymentSheet/StripePaymentSheet/Source/Analytics/STPAnalyticsClient+Address.swift
joyceqin-stripe: Done, renamed to `ms_session_elapsed`. And `source` is meant to be there, talked to Nidhi— there was some confusion on the purpose of it

StripePaymentSheet/StripePaymentSheet/Source/Analytics/STPAnalyticsClient+Address.swift
joyceqin-stripe: `timeToFetch` can be optional. When a suggestion already includes a full address (no place-details fetch needed), or when MapKit resolves via `asAddress`, there's no request to time.

StripePaymentSheet/StripePaymentSheet/Source/Analytics/STPAnalyticsClient+Address.swift
joyceqin-stripe: Both are now included: `result_count` and `ms_to_fetch` (renamed from `time_to_fetch` to use the `ms_` prefix).

StripePaymentSheet/StripePaymentSheet/Source/Analytics/STPAnalyticsClient+Address.swift
joyceqin-stripe: Made it not optional

StripePaymentSheet/StripePaymentSheet/Source/Analytics/STPAnalyticsClient+Address.swift
```diff
@@ -65,42 +64,58 @@ extension STPAnalyticsClient {
...
+        logAddressControllerEvent(event: .addressAutocompleteSuggestions,
+                                  addressAnalyticData: analyticData,
+                                  additionalParams: params,
+                                  apiClient: apiClient)
     }
 
-    func logAddressAutocompleteComplete(characterCount: Int, sessionToken: String, source: String, duration: TimeInterval, latency: TimeInterval?, apiClient: STPAPIClient) {
+    func logAddressAutocompleteSelected(addressCountryCode: String, queryLength: Int, sessionToken: String, source: String, sessionElapsed: TimeInterval, placeId: String?, msToFetch: TimeInterval?, apiClient: STPAPIClient) {
         assert(apiClient.publishableKey?.nonEmpty != nil) // A publishable key is required to be set at this point so we can send it in our analytics payload
+        let analyticData = addressCountryCode.isEmpty ? nil : AddressAnalyticData(addressCountryCode: addressCountryCode, autoCompleteResultedSelected: nil, editDistance: nil)
```
wooj-stripe: nit, this should probably just be:
```
        let analyticData = AddressAnalyticData(addressCountryCode: addressCountryCode, autoCompleteResultedSelected: nil, editDistance: nil)
```

`AddressAnalyticData::analyticsPayload` can be `[String: Any?]?` which will centralize your logic w.r.t  there being an analytic payload or not (rather than repeating `addressCountryCode.isEmpty ? nil`) .

StripePaymentSheet/StripePaymentSheet/Source/Analytics/STPAnalyticsClient+Address.swift
```diff
@@ -14,15 +14,12 @@ extension STPAnalyticsClient {
     func logAddressControllerEvent(
         event: STPAnalyticEvent,
         addressAnalyticData: AddressAnalyticData?,
```
wooj-stripe: This can be:
```
addressAnalyticData: AddressAnalyticData,
```

wooj-stripe (dismissed):
Don't want to block on my nit

wooj-stripe (approved):

PR #6815, opened 2026-07-30
Allow manual capture when live capture is required.

Pearapps:
Just allows manual capture even when the live capture is required.


luisv-stripe (approved):
LGTM, thanks!

PR #6817, opened 2026-07-30
Refactor Checkout confirm ownership

yuki-stripe:
## Summary
- add the public Checkout.confirm wrapper and route Checkout Session confirmation through Checkout
- move Checkout Session-specific confirmation out of PaymentSheet+API.confirm
- restrict PaymentSheet.confirm to PI/SI/deferred confirm intents

## Testing
- ci_scripts/run_tests.rb --scheme StripePaymentSheet --build-only


porter-stripe (approved):

porter-stripe:
Mentioned this offline, but we should consider moving confirm off the queue to avoid deadlocks w/ apple pay

PR #6827, opened 2026-08-03
ECE + basic apple pay

joyceqin-stripe:
## Summary

Add a CheckoutApplePayContext to handle Checkout Apple Pay presentation and confirmation. Support basic Apple Pay (no automatic tax or shipping/billing address handling).
## Motivation

ECE
## Testing

https://github.com/user-attachments/assets/240e8ad7-44b2-4928-b4e0-297bcea2286e

## Changelog

N/A


Example/PaymentSheet Example/PaymentSheet Example/Checkout Playground/CheckoutCartView.swift
```diff
@@ -128,6 +153,9 @@ struct CheckoutCartView: View {
                 merchantId: "merchant.com.stripe.paymentsheet.example"
             )
             config.currencySelectorElement.appearance = currencySelectorAppearance
+            config.expressCheckoutElement.confirmHandler = { [self] result in
```
yuki-stripe: does this need to be [weak self] to avoid retain cycle?

yuki-stripe: Do we show ECE in the example?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Express Checkout Element/ExpressCheckoutElement+Configuration.swift
```diff
@@ -10,6 +10,12 @@
 extension ExpressCheckoutElement {
     /// Configuration options for ``ExpressCheckoutElement``.
     public struct Configuration {
+        /// A closure called after a wallet payment confirmation completes.
+        public typealias ConfirmHandler = (_ result: Checkout.ConfirmResult) -> Void
+
+        /// Called after a wallet payment confirmation completes.
+        public var confirmHandler: ConfirmHandler?
```
yuki-stripe: probably needs to be non-nil

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Express Checkout Element/ExpressCheckoutElement.swift
```diff
@@ -8,7 +8,8 @@
 /// Handles Checkout mutations requested by an ExpressCheckoutElement.
 @MainActor
 protocol ExpressCheckoutElementDelegate: AnyObject {
-    // TODO: Add delegate methods for Apple Pay and Link button taps
+    /// Called when the user taps the Apple Pay button.
+    func confirmApplePay() async -> Checkout.ConfirmResult
```
yuki-stripe: let's rename using standard delegate naming convention (eceNeedsConfirm or something)

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Confirm+Link.swift
```diff
@@ -10,7 +10,7 @@
 
 extension Checkout {
     static func confirmLink(
-        checkoutSession: Session,
+        checkout: CheckoutConfirmDataSource,
```
yuki-stripe: Let's rename `checkout` -> `checkoutConfirmDataSource` if we keep it. But see comment below first.

yuki-stripe: Why does `confirmLink` now need allll of the stuff in ConfirmDataSource?

yuki-stripe: Also note paymentHandler is duplicated in confirmdatasource and here

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Confirm.swift
```diff
@@ -1,7 +1,28 @@
...
+@MainActor
+protocol CheckoutConfirmDataSource: AnyObject {
+    var applePayConfiguration: Checkout.ApplePayConfiguration? { get }
+    var session: Checkout.Session { get }
+    var apiClient: STPAPIClient { get }
+    var paymentHandler: STPPaymentHandler { get }
+    var returnURL: String? { get }
+    var merchantDisplayName: String { get }
+    func commitSession(_ response: PaymentPagesAPIResponse) async throws
+}
```
yuki-stripe: Let's abstract out the things that Apple Pay specifically needs, like `protocol StuffThatApplePayNeeds`. This is too vague.

Do we have any alternatives to `commitSession`? Is there something more specific?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/CheckoutApplePayContext.swift
```diff
@@ -0,0 +1,373 @@
...
+//
+
+import Foundation
+import PassKit
+@_spi(STP) import StripeApplePay
+@_spi(STP) import StripeCore
+@_spi(STP) import StripePayments
+import UIKit
+
+/// Manages an active Apple Pay session.
```
yuki-stripe: i don't like this comment, dunno what 'active Apple Pay session' means (what is active? what is session?).

I'd say like it is responsible for handling the Apple Pay flow end to end, including displaying line items, collecting shipping, updating tax region, confirmation, etc.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/CheckoutApplePayContext.swift
```diff
@@ -0,0 +1,373 @@
...
+        case success
+    }
+
+    private weak var checkout: CheckoutConfirmDataSource?
+    private let session: Checkout.Session
+    private let merchantLabel: String
+    private let apiClient: STPAPIClient
+    private let paymentHandler: STPPaymentHandler
+    private let returnURL: String?
+    var authorizationController: PKPaymentAuthorizationController
```
yuki-stripe: why a var? why does this take this in the init?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/CheckoutApplePayContext.swift
```diff
@@ -0,0 +1,373 @@
...
+                    paymentMethod: paymentMethod.id,
+                    expectedAmount: currentSession.expectedAmount(),
+                    expectedPaymentMethodType: paymentMethod.type?.rawValue ?? STPPaymentMethodType.card.identifier,
+                    savePaymentMethod: savePaymentMethod,
+                    returnURL: self.returnURL,
+                    shipping: self.makeShippingDetailsParams(from: payment),
+                    clientAttributionMetadata: clientAttributionMetadata
+                )
+
+                // 3. Handle next actions
```
yuki-stripe: We can't handle ~any next actions here, we can't present anything. Not sure what to do about that exactly though in code. We really just want to confirm and then uhh do the post-next-action work that hasn't been implemented yet. CS confirm is a 3 step thing
1. confirm
2. handle next actions
3. post-next-action work tbd (it's gonna be polling)

Really we just want to do (1) and (3) here. Let's at least add a TODO comment about this.

yuki-stripe: That might let us remove paymentHandler, authContext, returnURL dependencies from here, which make ~no sense for Apple Pay.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Confirm.swift
joyceqin-stripe: I named it commitSession because Checkout already has that

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/CheckoutApplePayContext.swift
joyceqin-stripe: Still need returnURL for confirmCheckoutSession because otherwise we get an error during confirmation because it expects a returnURL, but removing the next action handling removed the other dependencies

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Internal.swift
```diff
@@ -8,8 +8,46 @@
...
+            STPAnalyticsClient.sharedClient.log(analytic: analytic)
+            return .failed(CheckoutError.unknown(debugDescription: errorMessage))
+        }
+        let confirmationContext = confirmationContext(for: paymentMethod)
+        let authenticationContext = AuthenticationContext(
+            presentingViewController: presentingViewController,
+            appearance: .default
+        )
+
+        let result = await Checkout.confirm(
```
yuki-stripe: Hrmm I think we don't want to call the static `confirm` because that lacks all the protections of the instance method. Not sure what to do about this footgun.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Express Checkout Element/ExpressCheckoutElement+Configuration.swift
```diff
@@ -10,6 +10,12 @@
 extension ExpressCheckoutElement {
     /// Configuration options for ``ExpressCheckoutElement``.
     public struct Configuration {
+        /// A closure called after a wallet payment confirmation completes.
+        public typealias ConfirmHandler = (_ result: Checkout.ConfirmResult) -> Void
+
+        /// Called after a wallet payment confirmation completes.
+        public var confirmHandler: ConfirmHandler = { _ in }
```
yuki-stripe, line 17: Seems like strictly a footgun if this has a default value. It's an API Review thing though so happy to discuss it there and leave as-is for now.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Internal.swift
yuki-stripe: I also don't like calling the main `confirm` b/c Apple Pay could invoke the pre-confirm stuff b/c I don't think that's even 'legal' to show our own UI when the Apple Pay button is tapped (I might be contradicting what I said earlier, if so sorry).

yuki-stripe: I recognize this is an existing flaw with PaymentSheet.confirm today.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Confirm.swift
```diff
@@ -1,7 +1,24 @@
...
+// MARK: - CheckoutApplePayDataSource
+
+/// Convenience bag of everything Apple Pay needs for confirmation
+@MainActor
+protocol CheckoutApplePayDataSource: AnyObject {
+    var applePayConfiguration: Checkout.ApplePayConfiguration? { get }
+    var apiClient: STPAPIClient { get }
+    var returnURL: String? { get }
+    var merchantDisplayName: String { get }
+    func commitSession(_ response: PaymentPagesAPIResponse) async throws
```
yuki-stripe: I think I'd like to separate this out into its own protocol. The other things probably shouldn't be a protocol but a struct.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Express Checkout Element/ExpressCheckoutElement+Configuration.swift
joyceqin-stripe: Yeah, I left some comments in the API Review, but basically making ConfirmHandler non-optional would make it so that the ECE config on Checkout.Configuration couldn't do a default init, and it might never get initialized. There's probably a better way to do things, but I didn't like the ece config being nil

joyceqin-stripe: I think on Android, you can use ECE without configuring it in any way. Maybe for iOS it'd make more sense for the ECE config to be optional and assert if it's nil in ECE saying that they have to implement the confirmHandler or something

joyceqin-stripe: Going to leave as-is for now though

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Confirm+ApplePay.swift
```diff
@@ -0,0 +1,30 @@
...
+//
+
+import Foundation
+@_spi(STP) import StripeCore
+@_spi(STP) import StripePayments
+
+extension Checkout {
+    static func confirmApplePay(
+        checkoutSession: Session,
+        confirmationContext: ApplePayConfirmationContext,
```
yuki-stripe: nit call this `applePayConfirmationContext`

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Confirm.swift
```diff
@@ -1,9 +1,28 @@
...
 
 // MARK: - Confirm
 
+extension Checkout {
+    /// Convenience bag of everything Apple Pay needs for confirmation, aside from committing the session
+    /// back to `Checkout` (see ``CheckoutSessionBillingAddressUpdater`` for that).
+    struct ApplePayConfirmationContext {
+        let applePayConfiguration: Checkout.ApplePayConfiguration?
+        let apiClient: STPAPIClient
+        let returnURL: String?
```
yuki-stripe: shouldn't need to be optional.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Analytic.swift
```diff
@@ -12,6 +12,8 @@ struct UnexpectedCheckoutElementsErrorAnalytic: Analytic {
     enum ErrorCode: String {
         case paymentElementPresentingViewControllerUnavailable =
             "payment_element_presenting_view_controller_unavailable"
+        case expressCheckoutElementPresentingViewControllerUnavailable =
```
yuki-stripe: can't quickly find the relevant code but make sure we're grabbing the right VC to present from, ECE's presenting VC isn't necessarily the same as Checkout.confirm's presenting VC.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/CheckoutApplePayContext.swift
```diff
@@ -0,0 +1,369 @@
...
+                }
+
+                // 3. Commit the confirmed session back to Checkout so its state stays current.
+                // The payment already succeeded at this point, so a failure here shouldn't be
+                // reported to the customer as a failed payment - just log it.
+                do {
+                    try await self.sessionUpdater?.commitSession(response)
+                } catch {
+                    let errorAnalytic = ErrorAnalytic(
+                        event: .unexpectedPaymentSheetConfirmationError,
```
yuki-stripe: Let's use unexpectedCheckoutElementsError

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/CheckoutError.swift
```diff
@@ -21,11 +21,25 @@ public enum CheckoutError: Error, LocalizedError, Sendable {
...
 
+    /// The Apple Pay configuration is missing from ``Checkout/Configuration``.
+    case applePayNotConfigured
+
+    /// Apple Pay could not be presented. The device may not support Apple Pay or no payment cards are set up in Wallet.
+    case applePayUnavailable
+
+    /// The Checkout Session's PaymentIntent or SetupIntent still requires further action (e.g. 3DS) after
+    /// confirming with Apple Pay. We can't present that action while the Apple Pay sheet is still up.
+    case applePayRequiresNextAction(status: String)
```
yuki-stripe: Seems like strictly a bug on our part no, these should never happen? We shouldn't show Apple Pay if it's not configured or unavailable. Should probably send unexpected error analytic and just use `unknown` here.

StripePaymentSheet/StripePaymentSheet/Source/Analytics/PaymentSheetAnalyticsHelper.swift
```diff
@@ -288,8 +298,11 @@ final class PaymentSheetAnalyticsHelper {
     }
 
     func logSavedPaymentMethodRemoved(paymentMethod: STPPaymentMethod) {
-        let event: STPAnalyticEvent = {
+        let event: STPAnalyticEvent? = {
             switch integrationShape {
+            case .expressCheckout:
+                // TODO: Implement analytics for ExpressCheckoutElement.
+                return nil
```
yuki-stripe: Doesn't seem right to add ECE to integration shape here, given that `logSavedPaymentMethodRemoved` does not apply to ECE? Same with `logSavedPMScreenOptionSelected` and probably others.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Internal.swift
```diff
@@ -8,8 +8,51 @@
...
+        guard let expressCheckoutConfirmationContext = confirmationContext(for: paymentMethod) else {
+            let error = CheckoutError.unknown(debugDescription: "CheckoutController.expressCheckoutElementShouldConfirm() could not build a confirmation context for \(paymentMethod).")
+            STPAnalyticsClient.sharedClient.log(analytic: ErrorAnalytic(event: .unexpectedCheckoutElementsError, error: error))
+            return .failed(error)
+        }
+        let authenticationContext = AuthenticationContext(
+            presentingViewController: presentingViewController,
+            appearance: expressCheckoutConfirmationContext.configuration.appearance
+        )
+        let result = await CheckoutController.confirm(
```
yuki-stripe, line 33: can we add a todo here as discussed so i don't forget

StripePaymentSheet/StripePaymentSheet/Source/Analytics/PaymentSheetAnalyticsHelper.swift
joyceqin-stripe: I think some of them still apply— Link in ECE will emit some of these events

yuki-stripe: like, native Link? 

I see we already have LinkController as an integration shape. So ECE is going to..have a switch on this line between Apple Pay and Link, and in the Link case, do something? 

The point of this class was that we have different MPE variants but they have basically the same lifecycle, so it makes sense to have a shared analytics helper to track that lifecycle.

It's not clear to me why Link or ECE wants to share this.

yuki-stripe: The desire of my comment was not to delete the TODOs, I'm questioning adding ECE to the integration shape enum / to this class. Why'd we add it?

joyceqin-stripe: Originally added because PaymentSheetAnalyticsHelper is required in ConfirmationContext, and I thought Link would eventually log with integration shape. But turns out it's suppressed in native Link (logPayment: false). I'm wondering whether we can remove analytics helper from ConfirmationContext— ConfirmationContext.analyticsHelper isn't extracted for any payment option confirm flow except Link .

For this PR, I'll just construct a throwaway PaymentSheetAnalyticsHelper.

.linkController is actually a separate thing— it's a standalone Link integration that isn't used anywhere else in our SDK

yuki-stripe (approved):

PR #6833, opened 2026-08-04
[Identity] Migrate code from Future/Promise to Swift concurrency

### LINKED ISSUE #5557: [Feature] Convert Promises to Swift Concurrency
The async `Promise` architecture is being deprecated, so we will need to convert any promises in Identity to use async/await instead.

fedefrappi:
## Summary

Migrates StripeIdentity’s internal networking, verification flows, image scanning and uploading, and ML model loading from Promise/Future and completion-handler APIs to Swift concurrency.

Also adds an async `IdentityVerificationSheet.present(from:)` overload while preserving the existing completion-handler API, and removes the now-unused internal Promise/Future API client overloads.

The existing public completion-handler API remains unchanged to avoid a breaking change for current integrations.

## Motivation

StripeCore’s Promise/Future implementation is deprecated and not fully thread-safe. Migrating StripeIdentity to async/await removes that dependency, simplifies asynchronous control flow, and makes execution and actor-isolation requirements explicit.

## Testing

Ran Unit Test and checked the sample app keeps working as expected


StripeIdentity/StripeIdentity/Source/NativeComponents/ViewControllers/DocumentFileUploadViewController.swift
```diff
@@ -575,14 +575,16 @@ final class DocumentFileUploadViewController: IdentityFlowViewController {
...
     func didTapContinueButton() {
-        isSubmitting = true
-        sheetController?.saveDocumentBackAndTransition(
-            from: analyticsScreenName,
-            documentUploader: self.documentUploader
-        ) { [weak self] in
-            self?.isSubmitting = false
-            self?.continueButtonEnabled = false
+        Task {
+            isSubmitting = true
```
Pearapps: The behavior here is subtly changed, where the state is being changed inside the task instead of synchronously in this method -- im not sure if that could cause issues. I am also generally wondering if we need to short circuit these if we are currently submitting regardless.

Some other areas to check out that might have similar issues: 

`BiometricConsentViewController.didTapButton`, `ErrorViewController.didTapContinueButton`, `DocumentCaptureViewController.saveFrontAndDecideBack`, in `IndividualViewController`

StripeIdentity/StripeIdentity/Source/NativeComponents/Coordinators/VerificationSheetController.swift
```diff
@@ -238,258 +222,285 @@ final class VerificationSheetController: VerificationSheetControllerProtocol {
...
     private func saveDocumentFront(
         from fromScreen: IdentityAnalyticsClient.ScreenName,
         forceConfirm: Bool,
         documentUploader: DocumentUploaderProtocol,
-        onCompletion: @escaping (_ isBackRequired: Bool) -> Void
-    ) {
+    ) async -> Bool {
+        guard let frontUploadResult = await documentUploader.frontUploadResult()
+        else {
+            // TODO: throw an error for missing data instead?
```
Pearapps: Hm, this return is a behavior change that I am not sure is fully safe.. could use some extra auditing / testing

Pearapps: same with `saveDocumentBack` and `saveSelfieFileDataAndTransition`

StripeIdentity/StripeIdentity/Source/NativeComponents/ViewControllers/DocumentFileUploadViewController.swift
```diff
@@ -722,16 +724,14 @@ extension DocumentFileUploadViewController: UIDocumentPickerDelegate {
...
-                }
-            }
-        )
+        guard let sheetController else { return }
+        Task { @MainActor in
+            _ = await sheetController.saveDocumentFrontAndDecideBack(
+                from: analyticsScreenName,
+                documentUploader: documentUploader
+            )
+            updateUI()
```
Pearapps: In the old code, `isBackRequired` is checked before calling this, idk if that is important but a behavior change here

StripeIdentity/StripeIdentity/Source/NativeComponents/Coordinators/ImageScanner/ImageScanningConcurrencyManager.swift
```diff
@@ -128,26 +162,20 @@ final class ImageScanningConcurrencyManager: ImageScanningConcurrencyManagerProt
...
-        }
-
-        future.observe(on: concurrentQueue) { result in
-            switch result {
-            case .success(let scannerOutput):
+        concurrentQueue.async {
+            defer { self.semaphore.signal() }
+
+            do {
+                let scannerOutput = try scanner.scanImage(
```
Pearapps: This is a major change in behavior, moving this call from the caller's thread/queue into this concurrent queue. Ultimately, I think we should refactor this whole system to fully get off of queues - but either way I _think_ this will probably cause issues or unforeseen consequences at some point

StripeIdentity/StripeIdentity/Source/NativeComponents/Coordinators/ImageUploaders/IdentityImageUploader.swift
```diff
@@ -36,9 +36,6 @@ final class IdentityImageUploader {
     let analyticsClient: IdentityAnalyticsClient
     let sheetController: VerificationSheetControllerProtocol
 
-    /// Worker queue to encode the image to jpeg
-    let imageEncodingQueue = DispatchQueue(label: "com.stripe.identity.image-encoding")
```
Pearapps, line 40: Ultimately, this type of change is probably for the best --- although this does mean that all the images will be uploaded concurrently.... worth a sanity check because that can spike memory a lot. perhaps we do limit it somehow --- especially worth thinking about older devices

StripeIdentity/StripeIdentity/Source/API Bindings/STPAPIClient+Concurrency.swift
```diff
@@ -0,0 +1,83 @@
+import Foundation
```
Pearapps, line 8: should this have a file header?

StripeIdentity/StripeIdentity/Source/API Bindings/STPAPIClient+Concurrency.swift
```diff
@@ -0,0 +1,83 @@
...
+@_spi(STP) import StripeCore
+import UIKit
+
+extension STPAPIClient {
```
Pearapps, line 11: Idk if it opens a can of worms, but this probably is more of something that should go in StripeCore

StripeIdentity/StripeIdentity/Source/NativeComponents/Coordinators/VerificationSheetController.swift
```diff
@@ -131,7 +123,8 @@ private enum VerificationSheetControllerError: String, AnalyticLoggableStringErr
     case missingVerificationPageResponseForClearDataCalculation
 }
 
-final class VerificationSheetController: VerificationSheetControllerProtocol {
+@MainActor
```
Pearapps, line 131: do we need both @mainactor annotations here?

StripeIdentity/StripeIdentity/Source/NativeComponents/Coordinators/VerificationSheetController.swift
```diff
@@ -238,258 +222,285 @@ final class VerificationSheetController: VerificationSheetControllerProtocol {
...
-                    guard let verificationPageResponse = self.verificationPageOrLogError(
-                        missingError: .missingVerificationPageResponseForFallbackUpdate,
-                        assertionMessage: "Fail to get VerificationPageResponse is nil"
-                    ) else {
-                        return
-                    }
-                    self.verificationPageResponse = .success(verificationPageResponse.copyWithNewMissings(newMissings: resultData.requirements.missing))
-                    // clear collected data
-                    self.collectedData = StripeAPI.VerificationPageCollectedData()
+            self.isVerificationPageSubmitted = resultData.submitted == true && resultData.closed == true
```
Pearapps, line 286: I am a bit confused on this, because its also set on line 268

Pearapps:
a few higher level thoughts: 

1. Theres a lot of nuanced behavioral changes, especially after function reentry -- a triple check on everything definitely might be wise here 
2. I know the description mentions keeping completion handlers - but is this a good opportunity to remove those APIs? that could allow us to simplify a decent amount of these systems 
3. There's a handful of places that `[weak self]` was in the implementation of some asynchronous code but now that is not longer there - its a behavior change in those cases and theoretically cause some odd behavior like presenting VCs inappropriately.

I also wonder if there are a handful of more tests we could add to verify behaviors of some of the changes, even if that might be a bit tricky --- thinking of one off the top of my head is the `ImageScanningConcurrencyManager`

StripeIdentity/StripeIdentity/Source/NativeComponents/Coordinators/ImageScanner/ImageScanningConcurrencyManager.swift
fedefrappi: Correct, only the result of the scanImage call was observed in the concurrentQueue previously. I'll move the function call back outside of the dispatch.

StripeIdentity/StripeIdentity/Source/NativeComponents/ViewControllers/DocumentFileUploadViewController.swift
fedefrappi: Agreed, better to keep the state change outside of the task as well as adding a short circuit just in case

StripeIdentity/StripeIdentity/Source/API Bindings/STPAPIClient+Concurrency.swift
fedefrappi: 100%, once we're confident with the final shape of the PR I'll try to raise this

StripeIdentity/StripeIdentity/Source/NativeComponents/Coordinators/VerificationSheetController.swift
fedefrappi: yes, weird syntax but the class annotation doesn't impact the protocol implementation and viceversa

StripeIdentity/StripeIdentity/Source/NativeComponents/Coordinators/VerificationSheetController.swift
fedefrappi: yes this was pre-existing but it's duplicate code that doesn't really seem to be useful, I'll remove!

StripeIdentity/StripeIdentity/Source/NativeComponents/Coordinators/VerificationSheetController.swift
fedefrappi: Thanks for catching the TODO here.

This was actually exposing a pre-existing potential issue where if the `documentUploader.frontUploadFuture` (which became the `frontUploadResult` we're guarding now) was nil, this function never called the completion block and callers hanged indefinitely / never continued.

This is now impossible at syntax level because the function must return something, I'll throw an error or return an optional Bool so that we can try to keep the upstream behavior as close as possible as what it is today

PR #6836, opened 2026-08-04
SAE — Extract address form builder

gbirch-stripe:
## Summary
Extract the logic that builds the address form for the AddressViewController into a separate component. This will allow us to normalize default addresses passed to us in Checkout Sessions to prevent mismatches between internal data and customer-facing UI data

## Motivation
https://stripe.slack.com/archives/C09MLFD2KU7/p1785447926041409

## Testing
Added unit tests

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/AddressViewController/AddressFormNormalizer.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/AddressViewController/AddressFormNormalizer.swift
@@ -2,8 +2,7 @@
 //  AddressFormNormalizer.swift
 //  StripePaymentSheet
 //
-//  Copyright © 2026 Stripe, Inc. All rights reserved.
-//
+//  Created by George Birch on 8/4/26.
 
 import Foundation
 @_spi(STP) import StripeCore
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/AddressViewController/AddressFormNormalizer.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/AddressViewController/AddressFormNormalizer.swift
@@ -1,5 +1,5 @@
 //
-//  AddressViewControllerAddressNormalizer.swift
+//  AddressFormNormalizer.swift
 //  StripePaymentSheet
 //
 //  Copyright © 2026 Stripe, Inc. All rights reserved.
@@ -9,7 +9,7 @@ import Foundation
 @_spi(STP) import StripeCore
 @_spi(STP) import StripeUICore
 
-enum AddressViewControllerAddressNormalizer {
+enum AddressFormNormalizer {
     enum AddressSource {
         case defaultAddress
         case fallbackAddress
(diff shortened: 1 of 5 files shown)


gbirch-stripe:
> ⚠️ New dead code detected in this PR but ignored:
> 
> ```diff
> AddressFormNormalizer.swift:17 warning: Unused function 'normalize(defaultAddress:fallbackAddress:allowedCountries:addressSpecProvider:)'
> ```
> 
> This dead code check has been bypassed with the `skip dead code check` label.
> 
> 
> [find-dead-code]

This will be used the Checkout Sessions Shipping Address Element integration

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/AddressViewController/AddressFormNormalizer.swift
```diff
@@ -0,0 +1,88 @@
...
+        case .defaultAddress:
+            address = defaultAddress
+        case .fallbackAddress:
+            guard let fallbackAddress else { return nil }
+            address = fallbackAddress
+        case nil:
+            return nil
+        }
+
+        let addressSection = AddressSectionElement(
```
porter-stripe, line 38: I think normalize should own loading the address specs, possibly by being async or explicitly enforce/document that they are required to be loaded. A fresh `AddressSpecProvider` has no countries, so ASE would get an empty country list and eventually probably crashes. Might also need a test on the behavior of this API if AddressSpecProvider has not yet loaded.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/AddressViewController/AddressFormNormalizer.swift
```diff
@@ -0,0 +1,88 @@
...
+@_spi(STP) import StripeCore
+@_spi(STP) import StripeUICore
+
+enum AddressFormNormalizer {
+    enum AddressSource {
+        case defaultAddress
+        case fallbackAddress
+    }
+
+    static func normalize(
```
porter-stripe, line 17: If we are given an empty default address with a name and a valid billing fallback, will normalize choose billing, while AddressViewController renders the name only default with an empty address?

Default shipping has a name but no address -> normalizer doesn't see the name -> considers the default address empty -> fallback to the billing address -> internal data becomes "123 fake st." -> AddressViewController sees the name and preserves the empty shipping -> user sees blank shipping-address fields -> internal data says billing address while the UI shows no address

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/AddressViewController/AddressFormNormalizer.swift
porter-stripe, line 38: Why do we construct the ASE here separately from the AddressViewController.makeDefaultAddresSection? Since the goal is for normalization to match the user facing form, could we extract one builder used by both paths? Otherwise changes to either init can make them fall out of sync.

gbirch-stripe:
Discussed and decided against this approach.

PR #6851, opened 2026-08-05
Return from Link early when no funding sources available

mats-stripe:
## Summary

Instead of falling back to the card form when no funding sources are available, we now throw an error. Funding sources are determined by taking an intersection between:

1. `elementsSession.linkFundingSources` what the merchant has configured as allowed on their Stripe account
2. `consumerSession.supportedPaymentDetailsTypes` what the server says this consumer can use
3. `linkConfiguration.supportedPaymentMethodTypes` what the merchant has configured as allowed on the Link mobile SDK 

And if none are available, we used to fallback to `[card]`, which would cause payments to fail for IBP-only merchants.

## Motivation

Resolves https://jira.corp.stripe.com/browse/RUN_LINK_MOBILE-231

## Testing

Before: 

https://github.com/user-attachments/assets/fadae1f7-dc73-4831-a9ec-b9ecc045db8f

After:

https://github.com/user-attachments/assets/e4072b3e-3b50-47b1-84bf-e28acf3e05f9

## Changelog

```
### PaymentSheet
* [Fixed] LinkController (private preview) now returns an error when no funding sources are available for a Link session, rather than silently falling back to card.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@ The next release's version bump will so far be:
 PATCH
 
 ## X.Y.Z - changes pending release
+### PaymentSheet
+* [Fixed] LinkController (private preview) now returns an error when no funding sources are available for a Link session, rather than silently falling back to card.
 
 ## 26.5.0 2026-08-03
 ### CryptoOnramp (Alpha)
--- a/StripePaymentSheet/StripePaymentSheet/Source/Helpers/PaymentSheetLinkAccount.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/Helpers/PaymentSheetLinkAccount.swift
@@ -712,11 +712,6 @@ extension PaymentSheetLinkAccount {
             }
         }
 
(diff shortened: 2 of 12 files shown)


jkelle-stripe (commented):
Code looks good! Just want to make sure we're not losing a default behavior some merchants may be relying on. I'm not sure of the original intent to show a card if the intersection was empty, but it could be that some merchants are relying on this bug. We could have this be a setting, although that's a slippery slope. Agree with the motivation though that this could be considered a bug

mats-stripe:
>Just want to make sure we're not losing a default behavior some merchants may be relying on.

I don't think that's the case. The default behavior is still "all payment method types are supported", this change just affects a different scenario where no payment method types are supported. Even when the server was telling us that cards weren't supported, we were still showing the card form. It was an intentional decision at the time, but I think at the time we didn't really have the concept of "IBP only", and the assumption was that cards should always be supported (which we now know isn't the case)

jkelle-stripe (approved):

PR #6853, opened 2026-08-05
Checkout Sessions — ignore default shipping address if not in allowed countries

gbirch-stripe:
## Summary
If the merchant passes a default shipping address that is not in the allowed shipping countries, we silently treat it as nil.

## Motivation
https://docs.google.com/document/d/1CWglcOHR6I8cOWKjztPfnhnVG8axucpv-j5cWIoey7E/edit?tab=t.6mlp9ite2vpp#heading=h.fbn38s9qfcm0

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
@@ -359,7 +359,7 @@ extension Checkout {
            let address = billingDetails.address {
             try await updateBillingTaxRegionIfNecessary(address: address)
         }
-        
+
         // Set the passed default shipping address IF it is non-nil and the country is in the allowed shipping countries
         if let shippingDetails = defaults.shippingDetails,
            let address = shippingDetails.address,
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
@@ -359,8 +359,11 @@ extension Checkout {
            let address = billingDetails.address {
             try await updateBillingTaxRegionIfNecessary(address: address)
         }
+        
+        // Set the passed default shipping address IF it is non-nil and the country is in the allowed shipping countries
         if let shippingDetails = defaults.shippingDetails,
-           let address = shippingDetails.address {
+           let address = shippingDetails.address,
+           session.allowedShippingCountries?.contains(address.country) != false {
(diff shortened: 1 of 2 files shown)


gbirch-stripe:
> 
> ### StripePaymentSheet
> #### Public API
> ```diff
> +  @_Concurrency.MainActor @preconcurrency @objc dynamic public func presentationControllerShouldDismiss(_ presentationController: UIKit.UIPresentationController) -> Swift.Bool
> ```
> 
> If you are adding a new public API consider the following:
> 
> * Do these APIs need to be `public` or can they be protected with `@_spi(STP)`?
> * If these APIs need to be `public`, assess whether they require an API review.
> 
> If you are modifying or removing a public API:
> 
> * Does this require a breaking version change?
> * Do these changes require API review?
> 
> If you confirm these APIs need to be added/updated and have undergone necessary review, add the label `modifies public API` to this PR to acknowledge the interface change. Additionally, if you modified or removed an existing API, ensure you update the changelog to reflect the necessary version bump for your changes. Regular public API changes require a MAJOR version bump, and SPI API changes (other than declarations using only `@_spi(STP)` and/or `@_spi(ReactNativeSDK)`) require a MINOR or MAJOR version bump.
> 
> ℹ️ If this comment appears to be left in error, make sure your branch is up-to-date with `master`.

This is incorrect. Including and then stuck because of a force-pushing issue

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
```diff
@@ -359,8 +359,11 @@ extension Checkout {
            let address = billingDetails.address {
             try await updateBillingTaxRegionIfNecessary(address: address)
         }
+
+        // Set the passed default shipping address IF it is non-nil and the country is in the allowed shipping countries
         if let shippingDetails = defaults.shippingDetails,
-           let address = shippingDetails.address {
+           let address = shippingDetails.address,
+           session.allowedShippingCountries?.contains(address.country) != false {
```
porter-stripe: `updateShippingAddress` already checks allowedShippingCountries and throws `CheckoutError.invalidShippingCountry` when the country isn't allowed. This PR adds a second, silent check at the call site in applyDefaults that swallows the same condition instead. So now we have two places encoding "is this country allowed". One that throws, one that silently skips. Why not let `updateShippingAddress` do the validation (it already knows how) and just catch/ignore `CheckoutError.invalidShippingCountry` here?

gbirch-stripe: Good thought, updated!

porter-stripe (approved):

PR #6856, opened 2026-08-06
RTL(1/7): Add shared RTL UI foundations

porter-stripe:
## Summary
Introduces the mirroring logic our text field, picker, and card brand views need for right-to-left interfaces. `TextFieldView` and `PickerFieldView` now align their text to the leading edge under RTL instead of always defaulting left, while keeping values like emails and card numbers left-to-right internally. `FloatingPlaceholderTextFieldView` scales its floating placeholder from the correct edge, and `CardBrandView` mirrors the CBC indicator relative to the card logo instead of leaving it pinned to a fixed side.

Also adds a `forceRightToLeftLayout()` test helper on `UIView` so we can flip a view hierarchy into RTL for a snapshot/unit test without touching the simulator's language or region.

## Motivation
RTL support

## Testing
- New unit and snapshot tests covering RTL alignment for text fields, dropdowns, address forms, and the card brand/CBC indicator
- Existing tests updated

## Changelog
N/A


Tests/ReferenceImages_64/StripePaymentsUITests.STPPaymentCardTextFieldSnapshotTests/testCardBrandViewRightToLeftCBCIndicatorFollowsLogoAndReturnError_@3x.png
porter-stripe, line 1: @joyceqin-stripe it seems like Link still uses the old CBC dropdown, did we miss this?

joyceqin-stripe: STPPaymentCardTextField was intentionally left out because it's a legacy surface, so Mastercard did not require that we update it— https://stripe.slack.com/archives/C0AENPRB354/p1773423475376539

PR #6857, opened 2026-08-06
Add missing FPX bank support

ryanjeffares-stripe:
## Summary

Added support for FPX bank options that are available in Stripe.js but have been missing from the iOS SDK: Agrobank, Bank of China, Deutsche Bank, Public Bank Enterprise, MBSB Bank, BNP Paribas, and Citibank. Refactored some of the code to use map lookups instead of long switch statements. Displaying banks in ascending alphabetical order as is required by the financial partner.

## Motivation

New FPX bank support was added to the API [here](https://docs.stripe.com/changelog/dahlia/2026-07-29/fpx-payments-new-banks) and is available in Stripe.js, bringing this SDK to parity.

### DIFF
--- a/StripePayments/StripePayments/Source/API Bindings/Models/STPFPXBankBrand.swift
+++ b/StripePayments/StripePayments/Source/API Bindings/Models/STPFPXBankBrand.swift
@@ -24,6 +24,8 @@ import Foundation
     case ambank
     /// Affin Bank
     case affinBank
+    /// Agrobank
+    case agrobank
     /// Alliance Bank
     case allianceBank
     /// Bank Islam
@@ -32,8 +34,12 @@ import Foundation
     case bankMuamalat
     /// Bank Rakyat
     case bankRakyat
+    /// Bank of China
+    case bankOfChina
     /// BSN
     case BSN
+    /// Deutsche Bank
+    case deutscheBank
     /// HSBC BANK
     case HSBC
     /// KFH
@@ -42,10 +48,18 @@ import Foundation
     case maybank2E
     /// OCBC Bank
     case ocbc
+    /// Public Bank Enterprise
+    case publicBankEnterprise
     /// Standard Chartered
     case standardChartered
     /// UOB Bank
     case UOB
+    /// MBSB Bank
+    case mbsb_bank
+    /// BNP Paribas
+    case bnp_paribas
+    /// Citibank
+    case citibank
     /// An unknown bank
     case unknown
 }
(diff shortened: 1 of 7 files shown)


StripePayments/StripePayments/Source/API Bindings/Models/STPFPXBankBrand.swift
gbirch-stripe, line 12: We should probably be adding banks to the end of this list rather than in the middle. `STPFPXBankBrand` has an `Int` raw value type that defaults to the order of the enum cases. Adding to the middle of the list means that e.g. `allianceBank` used to have raw value 8 but now has raw value 9. This has the potential to cause issues for merchants

StripePayments/StripePayments/Source/API Bindings/Models/STPFPXBankBrand.swift
```diff
@@ -218,118 +102,133 @@ public class STPFPXBank: NSObject {
...
+        .HSBC: (business: "HSBC0223", individual: "HSBC0223"),
+        .KFH: (business: "KFH0346", individual: "KFH0346"),
+        .maybank2E: (business: "MBB0228", individual: "MBB0228"),
+        .maybank2U: (business: nil, individual: "MB2U0227"),
+        .ocbc: (business: "OCBC0229", individual: "OCBC0229"),
+        .publicBank: (business: "PBB0233", individual: "PBB0233"),
+        .RHB: (business: "RHB0218", individual: "RHB0218"),
+        .standardChartered: (business: "SCB0215", individual: "SCB0216"),
+        .UOB: (business: "UOB0228", individual: "UOB0226"),
+        .agrobank: (business: "AGRO01", individual: "AGRO02"),
```
gbirch-stripe: This is the opposite of Stripe.js and pay-server, which have AGRO02 has business and AGRO01 has individual. Can you please verify if this is correct or not and if there is any other incorrect data in this PR?

StripePayments/StripePayments/Source/Categories/Enums+CustomStringConvertible.swift
```diff
@@ -132,18 +134,30 @@ extension STPFPXBankBrand: CustomStringConvertible {
             return "bankIslam"
         case .bankMuamalat:
             return "bankMuamalat"
+        case .bankOfChina:
+            return "bankOfChina"
         case .bankRakyat:
             return "bankRakyat"
+        case .bnp_paribas:
```
gbirch-stripe: This should be camel case not snake case

StripePayments/StripePayments/Source/Categories/Enums+CustomStringConvertible.swift
```diff
@@ -132,18 +134,30 @@ extension STPFPXBankBrand: CustomStringConvertible {
...
+            return "citibank"
+        case .deutscheBank:
+            return "deutscheBank"
         case .hongLeongBank:
             return "hongLeongBank"
         case .maybank2E:
             return "maybank2E"
         case .maybank2U:
             return "maybank2U"
+        case .mbsb_bank:
```
gbirch-stripe: same here

gbirch-stripe (changes requested):
I see that Deutsche Bank, Public Bank Enterprise, BNP Paribas, and Citibank all have only a business code and no individual. However as far as I know the mobile SDK only supports individual, and so these banks would always reject these selctions

StripePayments/StripePayments/Source/API Bindings/Models/STPFPXBankBrand.swift
ryanjeffares-stripe: Good point, fixed.

ryanjeffares-stripe:
> I see that Deutsche Bank, Public Bank Enterprise, BNP Paribas, and Citibank all have only a business code and no individual. However as far as I know the mobile SDK only supports individual, and so these banks would always reject these selctions

Ah okay, I thought it had similar functionality to Stripe.js. I've removed the business-only banks for now and we will discuss with product about supporting B2B on mobile. Also removed the bank codes (e.g. AGRO01) because they were not used, which takes care of your other comment about that.

EDIT: had to re-add the bank code logic because removing the function counted as a breaking API change requiring a major update :/ still not adding B2B-only banks.

gbirch-stripe (approved):

PR #6863, opened 2026-08-06
CryptoOnramp SDK: Exposes Settings to Seamless Sign-In Flow

mliberatore:
## Summary
In the example app, the menu that allows for toggling livemode, KYC L0 Mode, and adjusting oauth scopes is currently only available on the manual Sign In / Sign Up view. This change makes it available as part of the Seamless Sign-In flow as well, so that these options can be toggled on subsequent launches where a seamless token is available. Previously, if you had started with L0 KYC mode enabled, and relaunched, signing in seamlessly, you'd be stuck in regular KYC mode.

## Motivation

To make testing more flows possible in the example app.

## Testing

1. Created a new account with KYC L0 mode enabled and reached the KYC Info view.
2. Backgrounded and relaunched the app.
3. On the seamless sign in view, made sure the options menu was available and enabled L0 KYC Mode for that launch.
4. Verified that the KYC info view was in L0 mode, leaving SSN and DOB optional.
5. Signed out and switched L0 KYC Mode off.
6. Signed in seamlessly again.
7. Made sure the KYC Info view required SSN and DOB, as we were no longer in L0 KYC Mode.

| Seamless Sign In Options | KYC Info View |
| --- | --- |
|  |  |

## Changelog

N/A, example app changes only.


Example/CryptoOnramp Example/CryptoOnramp Example/CryptoOnrampExampleView.swift
```diff
@@ -56,21 +58,62 @@ struct CryptoOnrampExampleView: View {
...
+                            // Livemode is disabled on the simulator.
+                            .disabled(isRunningOnSimulator)
+
+                            Toggle(isOn: $isL0KYCModeEnabled) {
+                                Label("L0 KYC Mode", systemImage: "person.text.rectangle")
+                            }
+
+                            Divider()
+
+                            Button {
```
jeanregisser: Now that this menu is shared with the seamless screen, "OAuth Scopes…" shows up there too - but seamless sign-in goes through `authenticateUserWithToken` and never reads `selectedScopes`.

So changing scopes there is a silent no-op. Should we hide it when the seamless view is showing? Livemode and L0 KYC both still work, it's just the scopes item.

jeanregisser (commented):
Great 🙌

Example/CryptoOnramp Example/CryptoOnramp Example/CryptoOnrampExampleView.swift
mliberatore: Good point. Updated with your suggestion:
Simulator Screenshot - iPhone 17 - 2026-08-18 at 16 14 12 Simulator Screenshot - iPhone 17 - 2026-08-18 at 16 14 18

jeanregisser (approved):

PR #6865, opened 2026-08-06
SAE — Normalize Checkout shipping defaults

gbirch-stripe:
## Summary
- normalize configured shipping defaults through the Shipping Address Element before applying them to Checkout
- leave invalid or disallowed defaults in the form for correction instead of updating the Checkout Session
- initialize payment elements after the normalized shipping address is applied

## Testing
added unit tests


StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Checkout/CheckoutDefaultsInitializationTests.swift
```diff
@@ -68,7 +69,148 @@ final class CheckoutDefaultsInitializationTests: XCTestCase {
...
+        // When Checkout initializes
+        let checkout = try await Checkout(configuration: configuration)
+
+        // Then the required state dropdown remains blank and the invalid address is not applied
+        XCTAssertEqual(checkout.getShippingAddressElement().addressViewController.addressSection?.state?.rawData, "")
+        XCTAssertEqual(requestRecorder.requests.map(\.kind), [.initSession])
+        XCTAssertNil(checkout.session.shippingAddress)
+    }
+
+    func testInitDoesNotApplyDisallowedShippingDefault() async throws {
```
porter-stripe: Is this a duplicate pretty much of testInitIgnoresShippingDefaultsWithDisallowedCountry? How does it differ?

gbirch-stripe: Good point — now that we have handling + testing of this case via the Checkout initialization we don't need to test it on the SAE

porter-stripe (approved):

PR #6867, opened 2026-08-06
Simplify currency selector availability handling

porter-stripe:
## Summary

Adaptive Pricing availability is fixed when a Checkout Session is created, but the currency selector currently builds an empty view and watches for Adaptive Pricing data to appear or disappear across updates.

Make element creation failable based on the initial session data and keep the resulting UIKit and SwiftUI selector stable for the session lifetime.

## Motivation

Checkout Elements

## Testing

- New and updated unit tests
- 46 focused tests

## Changelog

N/A


gbirch-stripe (approved):
nit: from a code cleanliness and testability perspective I think the CurrencySelectorElementUIView should just take displayable data and the view model should be handling extracting that out of the Checkout.Session and such. But that step is pretty simple here so it probably doesn't matter much

PR #6868, opened 2026-08-06
CheckoutSessions — Initialize elements in parallel

gbirch-stripe:
## Summary
- Initialize elements in parallel instead of sequentially
- SAE is excluded since we need to initialize it to get the shipping address and update tax info
- PE needs the checkout itself to initialize. Other elements can and should use the sessionSource for thread safety across parallel initializations

## Motivation
Reducing loading latency

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
@@ -167,20 +167,20 @@ public final class Checkout: ObservableObject {
             )
 
             // PE
-            async let paymentElement = PaymentElement(checkout: self)
+            let peTask = Task { try await PaymentElement(checkout: self) }
 
             // CSE
-            async let currencySelectorElement = configuration.adaptivePricing.allowed ?
-                CurrencySelectorElement(
+            let cseTask = Task {
+                configuration.adaptivePricing.allowed ? await CurrencySelectorElement(
                     sessionSource: sessionSource,
                     configuration: configuration.currencySelectorElement,
                     delegate: self
-                )
-            : nil
+                ) : nil
+            }
 
             // Wait for all of the initializations to finish.
-            self.paymentElement = try await paymentElement
-            self.currencySelectorElement = await currencySelectorElement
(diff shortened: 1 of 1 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
```diff
@@ -154,22 +154,33 @@ public final class Checkout: ObservableObject {
...
             self.expressCheckoutElement = ExpressCheckoutElement(
                 sessionSource: sessionSource,
                 configuration: configuration,
                 delegate: self
             )
-            if configuration.adaptivePricing.allowed {
-                self.currencySelectorElement = await CurrencySelectorElement(
+
+            // PE
+            let peTask = Task { try await PaymentElement(checkout: self) }
```
porter-stripe, line 170: Why use unstructured tasks rather than like `async let`?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
```diff
@@ -154,22 +154,33 @@ public final class Checkout: ObservableObject {
...
+
+            // CSE
+            let cseTask = Task {
```
porter-stripe, line 173: This could possibly now introduce a bug, I _think_ or expose a weakness in CSE but was fine previously. Could CSE miss a session update now? PE could update the session with tax from the default saved card billing address while CSE is downloading flag images. CSE subscribes after that so it could miss that billing address update which may update the totals for tax.

I think if we move to this parallelization we should consider doing PE first, then the others in parallel or in CSE remove the .dropFirst on it's subscribe. dropFirst was added to skip duplicate emissions from PE as an optimization originally.

This change is somewhat more interesting than it seems b/c the session can literally change while all the elements are loading due to the loading of a different element. Are all of the element initializers safe if the session changes underneath them? This feels potentially fragile ordering assumption that might be worth documenting or testing.

gbirch-stripe:
Closing this in favor of keeping the sequential order but documenting it: https://github.com/stripe/stripe-ios/pull/6892

PR #6876, opened 2026-08-09
Reshape Checkout Session totals

yuki-stripe:
## Summary

- Stop decoding unused `return_url`, `shipping_rate`, and `total_summary` fields from the Payment Pages response.
- Replace optional `Checkout.Total` with required `Checkout.Session.Totals`.
- Calculate unified/modeless totals from supported `one_time_price` checkout items.
- Return a zero discount because discounts are not implemented in unified mode.

Each total is formatted as an `Amount` using the Payment Pages response currency. `total_summary` does not build `Session.Totals`.

## Payment Pages response to Session.Totals

```
┌──────────────────────────────────────────┐       ┌──────────────────────────┐
│ Payment Pages response                   │       │ Session.Totals           │
│                                          │       │                          │
│ checkout_items[]                         │       │                          │
│   type: one_time_price                   │       │                          │
│                                          │       │                          │
│   one_time_price.subtotal ───────────────┼──Σ───▶│ subtotal: Amount         │
│                                          │       │                          │
│   one_time_price.items[]                 │       │                          │
│     .tax_exclusive ──────────────────────┼──Σ───▶│ taxExclusive: Amount     │
│     .tax_inclusive ──────────────────────┼──Σ───▶│ taxInclusive: Amount     │
│                                          │       │                          │
│   one_time_price.total ──────────────────┼──Σ───▶│ total: Amount            │
│                                          │       │                          │
│ currency ────────────────────────────────┼──────▶│ formats every Amount     │
│                                          │       │                          │
│ N/A (not implemented) ──────────────────┼──────▶│ discount: Amount = 0     │
└──────────────────────────────────────────┘       └──────────────────────────┘
```

`Σ` means sum across every supported `one_time_price` checkout item. Unsupported checkout item types fail response decoding.

## Tests

Added one test that passes two one-time-price groups and verifies that their subtotals, inclusive tax, exclusive tax, and totals are summed, discount is zero, and the results use the response currency.

Updated existing tests for the required totals model, item-derived tax values, aggregate tax metadata remaining separate, session updates preserving totals, and Apple Pay using the required totals value.

CleanShot 2026-08-14 at 16 15 00@2x

## Why recorded API responses

Handcrafting a full Payment Pages response is a strict anti-pattern. These response shapes will keep evolving, and the tests intentionally modify fields to exercise decoding edge cases; maintaining a parallel synthetic representation makes it easy for unrelated fields to drift into combinations the API could never return.

Payment Pages should be the source of truth. Recorded Mocktail responses are easy to refresh, keep the complete response internally consistent, and still replay offline during normal test runs. Focused malformed-response tests can load a recording and mutate only the field under test instead of duplicating the whole payload.


StripePaymentSheet/StripePaymentSheet/Resources/Localizations/bg-BG.lproj/Localizable.strings
```diff
@@ -290,8 +290,6 @@
 
 "Set up" = "Създаване";
 
-"Shipping" = "Доставка";
```
porter-stripe, line 293: Why are these included?

yuki-stripe: this pr removes our usage of the "Shipping" localized string here https://github.com/stripe/stripe-ios/pull/6876/changes#diff-c1bbb3403d26c83db5fad81bb67ca5769737ab332550fd65bcae0903367c139cL70

and I think some localization check fails if we don't clean this up too

porter-stripe (approved):

PR #6882, opened 2026-08-10
CryptoOnramp SDK: Addresses Flaky Typing in UI Integration Tests

mliberatore:
## Summary
In recent sporadic CI failures for the CryptoOnramp UI tests, the artifacts show that occasionally characters are being missed while typing. It's not reliably the first character, so it's unlikely to be due to e.g. not waiting for the keyboard to appear.

Most of text field modification is already running through an `enterText` helper, so these changes use it more consistently, and now that helper checks the field after typing, and retries once if the output does not match the input. 

## Motivation

Sporadic test failures. For example, in this test, the wallet address should have the prefix `DBhBRyb9…` but the `h` is missing.
CleanShot 2026-08-10 at 11 38 35

And in this test, the street address should have "456 Fake St." not "456 ake St."

CleanShot 2026-08-10 at 11 46 41

## Testing

Ran the updated tests, and ensured they pass locally and on CI.

## Changelog

N/A

### DIFF
--- a/Example/CryptoOnramp Example/CryptoOnrampExampleUITests/CryptoOnrampExampleUITestHelpers.swift
+++ b/Example/CryptoOnramp Example/CryptoOnrampExampleUITests/CryptoOnrampExampleUITestHelpers.swift
@@ -23,7 +23,7 @@ extension CryptoOnrampExampleUITests {
     ) {
         let email = "crypto-onramp-ui-tests-\(UUID().uuidString.lowercased())@stripe.com"
         waitForLoadingToFinish(file: file, line: line)
-        enterText(email, in: app.textFields["Enter email address"].firstMatch, file: file, line: line)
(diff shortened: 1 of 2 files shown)


Example/CryptoOnramp Example/CryptoOnrampExampleUITests/CryptoOnrampExampleUITestHelpers.swift
```diff
@@ -189,15 +169,83 @@ extension CryptoOnrampExampleUITests {
         )
     }
 
+    func enterText(
```
mliberatore, line 172: This is simply a second helper to reduce the duplication of `enterText` where we're passing `app.textFields["…"].firstMatch`, as it's the most common occurrence across the tests.

jeanregisser (approved):
🚀

PR #6889, opened 2026-08-11
SAE — presentation API + implementations

gbirch-stripe:
## Summary

- add presentation APIs
- switch checkout element playground to use SAE instead of legacy address element

NOTE: save functionality is still not implemented, only presentation. The playground data will not update when the SAE dismisses

## Motivation
- eng design doc: https://docs.google.com/document/d/1CWglcOHR6I8cOWKjztPfnhnVG8axucpv-j5cWIoey7E/edit?tab=t.6mlp9ite2vpp
- API review: https://docs.google.com/document/d/1pIrRtTJQzu7hAhnCjumM64cwatgOjNuR9N22QzGJUAg/edit?tab=t.0#heading=h.5iezaxih5aiu

## Testing
- added unit tests

## Changelog
N/A


gbirch-stripe:
> 
> ```diff
> AddressViewController.swift:614 warning: Unused parameter 'addressDetails'
> ```
> 
> 
> 
> 
> [find-dead-code]

Unused param is intentionally included in delegate (will be used for the CS save functionality in a future PR)

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Shipping Address Element/ShippingAddressElement.swift
```diff
@@ -50,13 +52,57 @@ public final class ShippingAddressElement {
...
+            let analytic = UnexpectedCheckoutElementsErrorAnalytic(
+                errorCode: .shippingAddressElementPresentingViewControllerUnavailable,
+                errorMessage: errorMessage
+            )
+            STPAnalyticsClient.sharedClient.log(analytic: analytic)
+            completion?()
+            return
+        }
+
+        guard presentingViewController.presentedViewController == nil else {
```
porter-stripe, line 93: What happens if present is called twice before the first presentation finishes? e.g. the customer double taps "Add shipping address"? I think we would present the same addressViewController twice and overwrite the first presentationCompletion.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Shipping Address Element/ShippingAddressElement.swift
```diff
@@ -50,13 +52,57 @@ public final class ShippingAddressElement {
...
 }
 
 extension ShippingAddressElement: AddressViewControllerDelegate {
     public func addressViewControllerDidFinish(
         _ addressViewController: AddressViewController,
         with address: AddressViewController.AddressDetails?
     ) {
-        // TODO(gbirch): populate when implementing presentation
+        addressViewController.dismiss(animated: true) {
+            self.presentationCompletion?()
```
porter-stripe: Can we clear this before calling it? If the completion immediately presents this element again, that call stores the second completion before line 105 clears it. Probably an edge case.
```swift
shippingAddressElement.present {
   shippingAddressElement.present {
   }
}
```

I think the second completion would never run.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Shipping Address Element/ShippingAddressElement.swift
gbirch-stripe: Good thought. I actually tinkered with handling of this but decided against it since we don't do it for PS or FC. Re-adding it for extra safety, though

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Shipping Address Element/ShippingAddressElement.swift
gbirch-stripe: Changes for the simultaneous presentation case should also handle this

porter-stripe (approved):

PR #6892, opened 2026-08-12
Checkout Sessions — document element initialization order

gbirch-stripe:
## Summary
Add documentation explaining the order of Checkout Elements.

## Motivation
Initialization order is important and shouldn't be unintentionally changed. Adding documentation makes this safer

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
@@ -133,35 +133,14 @@ public final class Checkout: ObservableObject {
             self.session = loadedSession
             self.nonisolatedSession = loadedSession // temporary hack
 
-            let defaultShippingAddress: Session.ShippingAddress?
-            if let shippingDetails = configuration.defaults.shippingDetails,
-               let address = shippingDetails.address {
-                defaultShippingAddress = Session.ShippingAddress(
-                    name: shippingDetails.name,
-                    address: address
-                )
-            } else {
-                defaultShippingAddress = nil
-            }
-
             // Element initialization is intentionally sequential:
 
             // 1. Initialize SAE so that its form can normalize the raw default shipping address before it is applied to the session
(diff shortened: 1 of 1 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
```diff
@@ -144,7 +144,9 @@ public final class Checkout: ObservableObject {
                 defaultShippingAddress = nil
             }
 
-            // Initialize the SAE with the raw default so its form can normalize the address.
+            // Element initialization is intentionally sequential:
+
+            // 1. Initialize SAE so that its form can normalize the raw default shipping address before it is applied to the session
```
porter-stripe, line 138: Can we put the SAE init into it's own function? It's got some interesting logic and that it applies defaults, etc, and all the other elements are 1 liners

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
```diff
@@ -195,6 +174,39 @@ public final class Checkout: ObservableObject {
         }
     }
 
+    private static func makeShippingAddressElement(
```
porter-stripe, line 177: ooc why static?

porter-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
gbirch-stripe: We can't call any instance methods on Checkout from its initializer before all of its properties have been initialized. So to call `makeShippingAddressElement` before `self.shippingAddressElement` is set, either `makeShippingAddressElement` needs to be static or `self.shippingAddressElement` needs to be implicitly unwrapped (like `PaymentElement`). I'm really not a fan of having these implicit unwraps though since they're pretty dangerous

PR #6901, opened 2026-08-12
SAE — save shipping addresses

gbirch-stripe:
## Summary
- Save Shipping Address Element changes to the Checkout Session.
- Show loading UI when the address save is in progress

Stacked on #6889.

## Motivation
- eng design doc: https://docs.google.com/document/d/1CWglcOHR6I8cOWKjztPfnhnVG8axucpv-j5cWIoey7E/edit?tab=t.6mlp9ite2vpp
- API review: https://docs.google.com/document/d/1pIrRtTJQzu7hAhnCjumM64cwatgOjNuR9N22QzGJUAg/edit?tab=t.0#heading=h.5iezaxih5aiu

## Testing
- added unit tests
- added a snapshot test for the loading state

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/AddressViewController/AddressViewController.swift
```diff
@@ -366,18 +366,21 @@ extension AddressViewController {
...
+                return
+            }
             do {
                 try await self.integrationDelegate?.save(
                     addressDetails: addressDetails,
-                    // TODO(gbirch) fill in loading UI behavior
-                    setLoading: { _ in }
+                    setLoading: { [weak self] isLoading in
+                        self?.setLoading(isLoading)
+                    }
```
porter-stripe, line 379: nit: Why does the IntegrationDelegate control setLoading? AddressViewController.didContinue should already know about the lifetime of the save.

porter-stripe (approved):

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/AddressViewController/AddressViewController.swift
gbirch-stripe: We ideally don't want to set loading at all in the legacy address element case since its loading is actually synchronous and we might get a weird flash of loading state. I'll poke at this some more and make sure this is actually necessary to avoid the UI issue

PR #6902, opened 2026-08-12
Improve right-to-left support in Link

porter-stripe:
## Summary

Fixes right-to-left layout in Link. Navigation chevrons now point in the expected direction, payment method carousels begin at the RTL leading edge, and trailing controls mirror without overlapping Link branding. Adds RTL snapshot and accessibility coverage for these flows.

#### Link verification header

| Before | After |
| :---: | :---: |
| Before: Link verification header is not mirrored | After: Link verification header is mirrored |

#### Link wallet header

| Before | After |
| :---: | :---: |
| Before: Link wallet header is not mirrored | After: Link wallet header is mirrored |

#### Link add payment method navigation

| Before | After |
| :---: | :---: |
| Before: RTL back chevron points left | After: RTL back chevron points right |

## Motivation

Right-to-left language support

## Testing

- Manual Simulator before/after captures
- Focused RTL layout and VoiceOver traversal tests

## Changelog
N/A


mats-stripe (approved):
Tested it and LGTM!

PR #6914, opened 2026-08-13
Add Checkout Session order summary items

yuki-stripe:
## Summary

Add `Session.orderSummaryItems` for unified one-time-price Checkout Sessions. Decode `checkout_items[].one_time_price`, expose its nested product and amount details, and use those items for Apple Pay line items.

## Mapping

```text
PP /init response                              Checkout.Session
+-------------------------------------+        +----------------------------------+
| mode = "modeless"                   |        |                                  |
| currency ---------------------------+------->| Formats every Amount              |
|                                     |        |                                  |
| checkout_items[] -------------------+------->| orderSummaryItems[]               |
|   Must be non-empty                 |        | One output item per input item    |
|   Only "one_time_price" supported   |        |                                  |
+-------------------------------------+        +----------------------------------+

For each checkout_items[] entry:

PP one_time_price                            Session.OrderSummaryItem.oneTimePrice
+------------------------------------+       +------------------------------------+
| key -------------------------------+------>| key                                |
| N/A                                |       | description = nil                  |
|                                    |       |                                    |
| one_time_price.items[] ------------+------>| items[]                            |
| one_time_price.subtotal -----------+------>| amountDetails.subtotal             |
| one_time_price.total --------------+------>| amountDetails.total                |
| N/A                                |       | amountDetails.discount = $0        |
|                                    |       |                                    |
| sum(items[].tax_inclusive) --------+------>| amountDetails.taxInclusive         |
| sum(items[].tax_exclusive) --------+------>| amountDetails.taxExclusive         |
| flatten(items[].tax_amounts) ------+------>| amountDetails.taxAmounts           |
|                                    |       | nil when flattened list is empty   |
+------------------------------------+       +------------------------------------+

For each nested one_time_price item:

PP nested item                               Session OneTimePrice.Item
+------------------------------------+       +------------------------------------+
| price.id --------------------------+------>| key                                |
| price.product.name ---------------+------>| displayName                        |
| price.product.images -------------+------>| images                             |
|                                    |       |                                    |
| unit_amount                        |       |                                    |
|   or price.unit_amount ------------+------>| unitAmount                         |
|   or 0 for decimal-only Prices     |       |                                    |
| unit_amount_decimal --------------+------>| unitAmountDecimal                  |
| unit_label ------------------------+------>| unitLabel                          |
| quantity --------------------------+------>| quantity                           |
|                                    |       |                                    |
| adjustable_quantity.enabled=false +------>| adjustableQuantity = nil           |
| adjustable_quantity.enabled=true  +------>| adjustableQuantity                 |
|   maximum -------------------------+------>|   maximum (defaults to 99)         |
|   minimum -------------------------+------>|   minimum (defaults to 0)          |
+------------------------------------+       +------------------------------------+
```

```text
Strict decoding boundary
------------------------

Wrong mode
Missing or empty checkout_items
Unsupported checkout item type
Missing one_time_price data
Malformed Price, Product, amount, or tax data
Currency mismatch
                         |
                         v
              Reject the entire PP response

Malformed rows are never skipped, and malformed amounts never silently become $0.
```

## Tests

Added a decoder test that checks one-time-price product details, amounts, taxes, and adjustable quantity.

Added a decoder test that checks an empty one-time-price group.

Updated the Checkout functional test to check the live one-time-price order summary.

Updated Apple Pay tests to check line items built from order-summary items.


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/PaymentPagesAPIResponse+Session.swift
```diff
@@ -88,69 +92,135 @@ extension PaymentPagesAPIResponse {
...
+        formatter.numberStyle = .currency
+        formatter.usesGroupingSeparator = true
+        formatter.locale = locale
+        formatter.currencyCode = currency
+        let formatted = formatter.string(from: decimalizedAmount)
+            ?? "\(formatter.currencySymbol ?? "")\(decimalizedAmount)"
+        return Checkout.Session.Amount(
+            amount: formatted,
+            minorUnitsAmount: minorUnitsAmount
+        )
```
gbirch-stripe, line 125: Are there any plans to have this data come from the translation layer formatted like we want? This feels like pretty nasty logic to have on the clients

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/PaymentPagesAPIResponse+Session.swift
```diff
@@ -88,69 +92,135 @@ extension PaymentPagesAPIResponse {
...
+                    let currency = price.currency
+                    let unitAmount = item.unitAmount ?? price.unitAmount ?? 0
+                    let adjustableQuantity: Checkout.Session.AdjustableQuantity?
+                    if let rawAdjustableQuantity = item.adjustableQuantity,
+                       rawAdjustableQuantity.enabled {
+                        // TODO: The server should guarantee minimum and maximum when adjustable quantity is enabled.
+                        adjustableQuantity = Checkout.Session.AdjustableQuantity(
+                            enabled: true,
+                            maximum: rawAdjustableQuantity.maximum ?? 99,
+                            minimum: rawAdjustableQuantity.minimum ?? 0
```
gbirch-stripe, line 156: Where do these default fallbacks come from? Should we be asserting on having this data or do we expect to use these fallback values?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/PaymentPagesAPIResponse.swift
```diff
@@ -154,50 +154,164 @@ class PaymentPagesAPIResponse: NSObject {
...
+            guard items.count == itemDicts.count else {
+                return nil
+            }
+            self.items = items
+            self.subtotal = subtotal
+            self.total = total
         }
     }
 
     struct OneTimePriceItem {
```
gbirch-stripe, line 194: Same here. Are all of these validations with `nil` return escapes expected scenarios?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/PaymentPagesAPIResponse+Session.swift
```diff
@@ -88,69 +92,135 @@ extension PaymentPagesAPIResponse {
...
-    private static func makeMinorUnitsAmountDivisor(currency: String?) -> Int? {
-        guard let currency else { return nil }
+    private static func makeSessionAmount(
+        _ minorUnitsAmount: Double,
+        currency: String,
+        locale: Locale
+    ) -> Checkout.Session.Amount {
+        let minorUnitValueInMajorUnits = NSDecimalNumber.stp_decimalNumber(withAmount: 1, currency: currency) // e.g. USD: 0.01
+        let decimalizedAmount = NSDecimalNumber(value: minorUnitsAmount).multiplying(by: minorUnitValueInMajorUnits) // e.g. 49,900 × 0.01 = 499.00
+        let formatter = NumberFormatter()
```
gbirch-stripe, line 110: Do we need (or would there ever be a need) to support sub-cent precision? NumberFormatter doesn't support this so it will round/truncate

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/STPApplePayContext+CheckoutSession.swift
```diff
@@ -32,16 +32,20 @@ extension STPApplePayContext {
...
-            )
-            summaryItems.append(PKPaymentSummaryItem(label: itemLabel, amount: amount, type: .final))
+        for orderSummaryItem in session.orderSummaryItems {
+            switch orderSummaryItem {
+            case .oneTimePrice(let oneTimePrice):
+                for item in oneTimePrice.items {
+                    let itemLabel = item.quantity > 1
+                        ? String.Localized.lineItemLabel(name: item.displayName, quantity: item.quantity)
+                        : item.displayName
+                    let amount = NSDecimalNumber(value: item.unitAmount.minorUnitsAmount * Double(item.quantity))
```
gbirch-stripe: Should this be unitAmount or unitAmountDecimal? Decimal-only prices would be 0

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/PaymentPagesAPIResponse+Session.swift
yuki-stripe: Agreed, no plans yet but let's add a TODO.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/PaymentPagesAPIResponse+Session.swift
yuki-stripe: PP API models them as optional but sets these defaults. Added a TODO to make them non-optional.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/PaymentPagesAPIResponse+Session.swift
yuki-stripe: Not for us, probably, but yes it's possible to eg bill $.001 cents per token or whatever. It's okay that this amount is truncated but the `unitAmountDecimal` should not be, updating.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/PaymentPagesAPIResponse.swift
yuki-stripe: This was reworked (you reviewed this before I intended!). But there are plenty of cases where the current code returns nil when it needs to throw a 'missing required field' error (really a downside of `STPAPIResponseDecodable`). 

I won't address this here; I'll follow up in a PR to either manually throw all those errors or move this to use `StripeDecodable` instead.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/STPApplePayContext+CheckoutSession.swift
yuki-stripe: True, will prefer unitAmountDecimal

porter-stripe (approved):

PR #6938, opened 2026-08-17
Checkout Sessions — Add default shipping address to playground

gbirch-stripe:
## Summary
Add default shipping address setting to CS playground

## Motivation
Test default shipping address

## Testing

https://github.com/user-attachments/assets/258aecdc-e194-4206-91bf-5124500c5844

### DIFF
--- a/Example/PaymentSheet Example/PaymentSheet Example/Checkout Playground/CheckoutPlaygroundSections.swift
+++ b/Example/PaymentSheet Example/PaymentSheet Example/Checkout Playground/CheckoutPlaygroundSections.swift
@@ -170,6 +170,8 @@ struct CheckoutPlaygroundLineItemCard: View {
 struct CheckoutPlaygroundFeaturesSection: View {
     let customerType: CheckoutPlayground.CustomerType
     @Binding var shippingAddressCollection: Bool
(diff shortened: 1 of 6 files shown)


Example/PaymentSheet Example/PaymentSheet Example/Checkout Playground/CheckoutPlaygroundSections.swift
```diff
@@ -224,6 +240,103 @@ struct CheckoutPlaygroundFeaturesSection: View {
...
+    }
+}
+
+private struct CheckoutPlaygroundShippingAddressEditor: View {
+
+    @Binding var address: CheckoutPlayground.DefaultShippingAddress
+
+    var body: some View {
+        VStack(spacing: 12) {
+            CheckoutPlaygroundShippingAddressField(
```
porter-stripe, line 277: Can we just reuse address element to collect custom rather than creating this entire thing?

gbirch-stripe: I considered it but I think it limits our testing ability. The merchants will be inputting arbitrary strings, and part of the thing we want to test with this is what happens when they input data that isn't allowed by the Address Element

porter-stripe (approved):

PR #6939, opened 2026-08-17
Checkout Sessions — fix content hidden in playground cart view

gbirch-stripe:
## Summary
Adjust bottom padding so that content isn't covered by ECE

## Testing

simulator_screenshot_57586DEF-A5D0-4ED8-B5E2-68AB5247B6F4


Example/PaymentSheet Example/PaymentSheet Example/Checkout Playground/CheckoutCartContentView.swift
```diff
@@ -33,7 +34,7 @@ struct CheckoutCartContentView: View {
                 }
                 orderSummarySection
 
-                Spacer().frame(height: 160)
+                Spacer().frame(height: bottomContentPadding)
```
porter-stripe: Is this even needed now with the safeAreaInset?

gbirch-stripe: You're right we can fix this by getting rid of the spacer entirely!

porter-stripe (approved):

PR #6944, opened 2026-08-18
Align Checkout Session API with reviewed proposal

yuki-stripe:
## Summary

- rename the public `Checkout` class to `CheckoutController`
- rename `CheckoutController.isLoading` to `isUpdating`
- make `CheckoutController.configuration` internal
- make the initializers for `Session.PresentmentDetails`, `Session.DiscountAmount`, `Session.ShippingAddress`, `Session.Totals`, and `Session.Amount` internal
- remove the write-only `nonisolatedSession` property
- move `STPCheckoutSessionExchangeRateMeta` to `CheckoutController.Session.ExchangeRateMeta`
- remove the unused exchange-rate `id`, `buyCurrency`, and `sellCurrency` fields
- remove the unused localized-price `id` field
- remove unused `Equatable` conformances from the exchange-rate and localized-price metadata types

## Testing

No new tests.


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout.swift
```diff
@@ -16,62 +16,44 @@ import UIKit
...
+    @Published public internal(set) var isUpdating: Bool = false
 
     /// The Checkout Session, updated from Stripe after every mutation.
-    @Published public private(set) var session: Session {
-        didSet {
-            nonisolatedSession = session
-            // Just some notes: Setting session causes the publisher to fire even when it didn't change.
-            // AFAICT that's okay, deduping sees like a minor optimization to slightly reduce the amount of UI updates.
-        }
-    }
```
yuki-stripe, line 46: cc @gbirch-stripe this became unused!

> nonisolatedSession became unused in commit 553aa960e0c on July 20, 2026: “Intent.checkout -> Intent.checkoutSession (#6729).”
>
>  That commit changed Intent.checkout to store a Checkout.Session directly instead of retaining the controller. It replaced all 31 external reads of checkout.nonisolatedSession with the associated session value, fixing a CheckoutController/PaymentElement retain cycle.
>
>  The property was originally introduced on July 6 in 2f82fee08b0 as a non-main-actor snapshot for PaymentSheet internals. After 553aa960e0c, only its initialization and didSet assignments remained.

porter-stripe (approved):

PR #6945, opened 2026-08-18
Add protocol conformances to Checkout session models

porter-stripe:
## Summary

Adds `Hashable` and `Sendable` to Checkout's address models. Merchants can include these values in their own hashable state and move them across Swift concurrency boundaries without creating wrapper types.

Also adds `Identifiable` to the Session and order summary models that have stable server IDs. SwiftUI can preserve rows, local state, focus, tasks, and animations when Checkout publishes an updated Session snapshot.

Before, merchants may reach for value equality as identity:

```swift
ForEach(session.orderSummaryItems, id: \.self) { item in
    OrderSummaryRow(item: item)
}
```

Changing an amount or quantity changes the value, so SwiftUI can treat it as a new row. Now the models provide stable identity directly:

```swift
ForEach(session.orderSummaryItems) { item in
    OrderSummaryRow(item: item)
}
```

#### `CheckoutController.Address`

Adds `Hashable` so an address can be part of a merchant's own hashable app state, sets, dictionaries, or diffable data source identifiers. Adds `Sendable` so it can cross tasks and actors cleanly under strict concurrency.

#### `CheckoutController.Session.ShippingAddress`

Adds `Hashable` so the name and address can be treated as one hashable value in merchant state. Adds `Sendable` so the complete shipping value can move between tasks and actors without a wrapper.

#### `CheckoutController.Session`

Adds `Identifiable` using the Checkout Session ID. The ID stays the same across refreshed snapshots, so SwiftUI can keep views associated with the same Session.

#### `CheckoutController.Session.OrderSummaryItem`

Adds `Identifiable` using a case-qualified Checkout item key. Merchants can pass `session.orderSummaryItems` directly to `List` or `ForEach`, and SwiftUI keeps the same row when its amount changes.

#### `OrderSummaryItem.OneTimePrice`

Adds `Identifiable` using its Checkout item key, giving the extracted one-time Price group the same stable identity as its order summary row.

#### `OrderSummaryItem.OneTimePrice.Item`

Adds `Identifiable` using `inner_item_key`. SwiftUI can keep the same product row when its quantity changes, and multiple items that use the same Price still have distinct identities.

We only add `Identifiable` where Checkout provides a stable ID. We don't generate client-side IDs because they would change when the Session is rebuilt.

## Motivation

https://jira.corp.stripe.com/browse/MOBILE_APIREVIEW-227

## Testing

- CI

## Changelog

N/A


porter-stripe:
We didn't add Identifiable to the DiscountAmount or the TaxAmount b/c they have no ID from the backend. Maybe we should change that but seems like an optimization.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Session.swift
```diff
@@ -99,12 +99,28 @@ extension CheckoutController {
...
+            case oneTimePrice(String)
+        }
+
+        /// The stable identity of this order summary item within its Checkout Session.
+        public var id: ID {
+            switch self {
+            case let .oneTimePrice(oneTimePrice):
+                return .oneTimePrice(oneTimePrice.key)
+            }
+        }
```
gbirch-stripe: Why do we need to have an enum for the ID? Can we just use the oneTimePrice.key (and later equivalents for the other OrderSummerItem types?

porter-stripe: Yeah that's fair, it's a bit over architected for this use case.

gbirch-stripe (approved):

PR #6947, opened 2026-08-18
PS MainActor — create CheckoutController weak reference wrapper

gbirch-stripe:
## Summary
Currently we snapshot the CheckoutController.session on MPE load and attach that to the `Intent` for use throughout PS code. Now that we want to utilize PS MainActor isolation to access the current CheckoutController.session directly, we need to attach a weak reference to the CheckoutController to attach to the `Intent` enum case


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/CheckoutController.swift
```diff
@@ -13,6 +13,35 @@ import Foundation
 @_spi(STP) import StripePayments
 import UIKit
 
+// Wraps the Checkout Controller in a weak reference so that it can be passed through MPE code without a retain cycle
+@MainActor
+final class CheckoutIntentContext {
+
+    weak var checkout: CheckoutController?
```
porter-stripe, line 20: Why can callers replace `checkout`? This context represents the controller that created it, but the internal setter allows mutating it or clearing it. Can this be a let or private(set)?

yuki-stripe: > Now that we want to utilize PS MainActor isolation to access the current CheckoutController.session directly, we need to attach a weak reference to the CheckoutController to attach to the Intent enum case

Hm I intentionally didn't want to put the CheckoutController in the Intent at all.

Reasoning - It feels better to have an immutable Checkout.Session and it matches the other immutable Intents in the Intent enum. I'd rather expose the other functionality (updateBillingAddress, basically) via a limited protocol vs. giving access to the whole CheckoutController for safety/explicitness/maybe easier testing.

gbirch-stripe:
Discussed with Yuki. Skipping this work.

PR #6952, opened 2026-08-18
[PaymentSheet] Move FlowController UI work to MainActor

gbirch-stripe:


gbirch-stripe:
This gets into risky complicated territory and is not necessary for now. closing

PR #6960, opened 2026-08-19
CI — do LPMConfirmFlowTests in separate job

gbirch-stripe:
## Summary
Split LPMConfirmFlowTests into a separate job

## Motivation
framework-tests had become the limiting bottleneck for CI as it has grown. LPMConfirmFlowTests alone account for 12 minutes of this job and can be cleanly split into a separate one


porter-stripe (approved):
LGTM, can we verify we are ok with the build count increase before merging?

PR #6962, opened 2026-08-19
Fix decline messages after 3DS authentication

### LINKED ISSUE #3037: [BUG] Fix error mapping when returning from 3DS
## Summary
The payment error message in PaymentSheet for "Insufficient funds decline" case is too generic and can lead to confusion for users. Without knowing the specific reason (i.e., insufficient funds), users may repeatedly try the same card, thinking it might be a temporary issue.
https://user-images.githubusercontent.com/5547520/277351452-80cb89cf-1542-4a62-bf6c-e0e87749c479.png

Can you fix your mapping as Android side to replace "Card declined" by a more specific message like "insufficient funds"?
The reason: https://github.com/stripe/stripe-android/issues/7478#issuecomment-1779221544

## Help to reproduce
You can test it with your testing usecase "Insufficient funds decline" (card number: 4000 0000 0000 9995)
Use your PaymentSheet when 3DS is mandatory for a transaction

## iOS version
Any

## Installation method
N/A

## SDK version
N/A

## Other information
Already solved on Android side
https://github.com/stripe/stripe-android/pull/7499

porter-stripe:
## Summary
When a card is declined after 3DS authentication, we were replacing the specific server-provided error with a generic message based on the top-level error code. PaymentSheet now keeps the specific card message, so insufficient-funds declines show “Your card has insufficient funds.” instead of “Card declined.”

We still fall back to the localized SDK message or generic error when the server does not provide one.

## Motivation
Fixes https://github.com/stripe/stripe-ios/issues/3037

## Testing
- Added coverage for specific PaymentIntent and SetupIntent card errors after authentication.
- Added coverage for the localized fallback when the server message is missing.
- Ran the focused payment handler tests.

## Changelog
See diff

### DIFF
--- a/StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
+++ b/StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
@@ -2389,8 +2389,8 @@ public class STPPaymentHandler: NSObject {
                 ?? "There was an error confirming the Intent. Inspect the `paymentIntent.lastPaymentError` or `setupIntent.lastSetupError` property."
 
             userInfo[NSLocalizedDescriptionKey] =
-                apiErrorCode.flatMap({ NSError.Utils.localizedMessage(fromAPIErrorCode: $0) })
-                ?? userInfo[NSLocalizedDescriptionKey]
+                userInfo[NSLocalizedDescriptionKey]
+                ?? apiErrorCode.flatMap({ NSError.Utils.localizedMessage(fromAPIErrorCode: $0) })
                 ?? NSError.stp_unexpectedErrorMessage()
 
         // Client secret format error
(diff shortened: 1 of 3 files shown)


StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
```diff
@@ -2389,8 +2389,8 @@ public class STPPaymentHandler: NSObject {
                 ?? "There was an error confirming the Intent. Inspect the `paymentIntent.lastPaymentError` or `setupIntent.lastSetupError` property."
 
             userInfo[NSLocalizedDescriptionKey] =
-                apiErrorCode.flatMap({ NSError.Utils.localizedMessage(fromAPIErrorCode: $0) })
-                ?? userInfo[NSLocalizedDescriptionKey]
+                userInfo[NSLocalizedDescriptionKey]
+                ?? apiErrorCode.flatMap({ NSError.Utils.localizedMessage(fromAPIErrorCode: $0) })
```
porter-stripe, line 2393: Our order of preference was just out of order. We were originally falling back to the mapping of API errors to our hard coded generic error messages rather than whats from the server.

wooj-stripe (approved):

PR #6963, opened 2026-08-19
Validate required Payment Pages response fields

porter-stripe:
## Summary
Payment Pages guarantees required values within optional tax, shipping, saved payment method, recurring totals, and discount objects. We were decoding those values as optional and silently turning malformed responses into valid-looking Checkout session state. Now malformed nested payloads fail decoding, and discount fixtures use Payment Pages' required coupon code.

Corresponds to item #4 in the audit.

## Motivation
Checkout Sessions

## Testing
- Added coverage that missing or unknown required tax, shipping, saved payment method offer, recurring total, discount, and coupon values fail decoding.
- Updated existing Checkout session response fixtures and ran 115 unit tests.
- Formatting and lint.

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/PaymentPagesAPIResponse.swift
gbirch-stripe, line 487: Are we confident that these are the only possible values? It looks like this might fail decoding if a new value is added

porter-stripe: Good catch. These match the current contract, but additive enum values shouldn’t fail the entire session. I updated these to use `ParsedEnum` and handle unknown values.

gbirch-stripe (approved):

PR #6978, opened 2026-08-20
ECE Apple Pay Billing

joyceqin-stripe:
## Summary

ECE Configuration BillingDetailsCollectionConfiguration + Apple Pay can require billing fields
## Motivation

ECE
API still in review: https://docs.google.com/document/d/18Mn6y4NG7V-pkiJLRW7wi1w_a0GGAnJWnBkTGYK9Ato/edit?usp=sharing
## Testing

https://admin.corp.stripe.com/request-log/req_XGQowxa1463cG4

image

## Changelog

N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Express Checkout Element/ExpressCheckoutElement+Configuration.swift
```diff
@@ -27,6 +30,56 @@ extension ExpressCheckoutElement {
...
 
+    /// Configuration for how billing details are collected during checkout.
+    public struct BillingDetailsCollectionConfiguration: Equatable {
+        /// Billing details fields collection options.
+        public enum CollectionMode: String, CaseIterable {
+            /// The field will be collected depending on the Payment Method's requirements.
+            case automatic
+            /// The field will never be collected.
+            /// If this field is required by the Payment Method, you must provide it as part of `defaultBillingDetails`.
+            case never
```
porter-stripe: do we need a callout or some way to inform this cannot apply to name?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+BillingAddress.swift
```diff
@@ -33,3 +34,13 @@ extension CheckoutSessionBillingAddressUpdater {
...
     }
 }
+
+/// A billing details collection configuration that can compute which contact fields Apple Pay must require,
+/// implemented by both ``PaymentSheet/BillingDetailsCollectionConfiguration`` and
+/// ``ExpressCheckoutElement/BillingDetailsCollectionConfiguration`` so confirmation code
+/// can accept either one without converting between them.
+protocol CheckoutBillingDetailsCollectionConfiguration {
+    var requiredBillingContactFields: Set<PKContactField> { get }
+    var requiredShippingContactFields: Set<PKContactField> { get }
```
porter-stripe: Why do we need this protocol? PaymentSheet.BillingDetailsCollectionConfiguration and ECE.BillingDetailsCollectionConfiguration both implement the same Apple Pay contact field mapping, so we still have two sources of truth. Could we convert the ECE config to the PaymentSheet config or something?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Express Checkout Element/ExpressCheckoutElement+Configuration.swift
joyceqin-stripe: Things are still subject to change as the API is not finalized yet, but we decided to remove phone and [email](https://stripe.slack.com/archives/C09MLFD2KU7/p1787594252562309) from the BDCC (and it's possible name will be removed as well). For now, I'm just removing the `never` case for name.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+BillingAddress.swift
joyceqin-stripe: My thought was that it'd be nice to not have to transform into explicitly `PaymentSheet`-namespace, as it makes me a little uneasy sometimes because I like the separation, but it's not worth the duplicated logic here. Removed protocol, ECE.BDCC converts.

porter-stripe (approved):

PR #6979, opened 2026-08-20
Fix Apple Pay presentation window for Catalyst

wooj-stripe:
## Summary
Pass the presenting window through Checkout’s Apple Pay flow and implement the Catalyst-required delegate method.

## Motivation
Breaks CI

## Testing
Didn't test exact breakage, but should work.


porter-stripe (approved):

joyceqin-stripe (changes requested):
I think this will have some conflicts with Yuki's refactor on yuki/checkout-session-confirm-refactor, so we won't be able to source a presentationWindow from authenticationContext

PR #6987, opened 2026-08-21
CheckoutSessions — add CS Playground setting model

gbirch-stripe:
## Summary
Part 2/4 to add CS playground setting persistence via UserDefaults

## Motivation
Improved debugging experience


Example/PaymentSheet Example/PaymentSheet Example/Checkout Playground/CheckoutPlaygroundTypes.swift
```diff
@@ -219,7 +219,7 @@ enum CheckoutPlayground {
         }
     }
 
-    struct LineItemConfig: Identifiable {
+    struct LineItemConfig: Identifiable, Codable {
         let id = UUID()
```
porter-stripe: Do we encode this? Since this is a let with an initialize value Decodable can't overwrite it. I feel like I've seen warnings about this in the past, might be wrong.

gbirch-stripe: I don't think it would have mattered with the current code state (id just used for `ForEach`), but I fixed just in case there's a weird edge case in the future!

porter-stripe (approved):

PR #6992, opened 2026-08-21
ECE Apple Pay Billing Tax

joyceqin-stripe:
## Summary

ECE Apple Pay automatic tax by billing address
## Motivation

ECE
## Testing

https://github.com/user-attachments/assets/32ef6465-2fde-4ba0-9fda-6af73dcff0e1

## Changelog

N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/CheckoutApplePayContext.swift
```diff
@@ -186,8 +190,22 @@ final class CheckoutApplePayContext: NSObject, PKPaymentAuthorizationControllerD
...
-        // TODO: Update billing tax region when the user switches cards.
-        handler(PKPaymentRequestPaymentMethodUpdate(paymentSummaryItems: summaryItems()))
+        guard session.collectsTaxFromBillingAddress,
+              let postalAddress = paymentMethod.billingAddress?.postalAddresses.first?.value,
+              let address = STPApplePayContext.makeCheckoutAddress(from: postalAddress),
+              let checkoutWalletUpdater else {
+            handler(PKPaymentRequestPaymentMethodUpdate(paymentSummaryItems: summaryItems()))
+            return
+        }
+        Task { @MainActor in
```
porter-stripe: What happens if Apple Pay calls didFinish before this task complete? I think we could have some weird race condition occur. Can we keep this task tied to the lifetime of the outer operation, maybe by tracking it and waiting for it before didFinish resumes?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/CheckoutApplePayContext.swift
```diff
@@ -158,6 +159,7 @@ final class CheckoutApplePayContext: NSObject, PKPaymentAuthorizationControllerD
         switch paymentState {
         case .notStarted:
             Task {
+                await paymentMethodUpdateTask?.value
```
porter-stripe, line 163: What does this do?

joyceqin-stripe: We keep track of the Task to update the billing tax region, and we wait for it before didFinish resumes. .notStarted is the only state where didFinish can end the outer confirmation while a payment-method tax update may still be running, as .pending doesn't resume, and Apple Pay waits for the didSelectPaymentMethod completion handler before allowing authorization to proceed for the success and error states

joyceqin-stripe: This reminded me of the old discussion about resetting the tax region updates... I'll have to implement that

joyceqin-stripe: I think I'm going to reset tax region updates in a follow-up PR

porter-stripe (approved):

PR #6993, opened 2026-08-21
[Identity] Fix simulator export for arm64-only MediaPipe

mnn-stripe:
## Summary

Keep mixed-architecture release exports buildable after Project Orion made the bundled MediaPipe simulator artifacts arm64-only.

- Limit the simulator restriction to `StripeIdentity` and its production dependent, `StripeCryptoOnramp`, across Xcode and CocoaPods builds.
- Keep arm64 and x86_64 simulator slices for every unaffected SDK framework.
- Make the release exporter enforce that architecture matrix in nightly and release builds.
- Document the restriction in both SDK requirements pages and the pending changelog.

## Motivation

[Project Orion](https://github.com/stripe/stripe-ios/pull/6523) deliberately strips x86_64 from the bundled MediaPipe simulator artifacts. The release export still tried to link `StripeIdentity` for x86_64, so Xcode discarded MediaPipe and failed with undefined `MPP*` and `mediapipe::*` symbols.

The restriction must also cover `StripeCryptoOnramp`: it imports and links `StripeIdentity` in production and is exported by the same archive. Applying `ARCHS=arm64` to the shared archive would hide the linker failure by dropping Intel simulator support from every Stripe iOS framework.

CocoaPods generates its own targets rather than importing settings from the checked-in Xcode projects. Both podspecs therefore declare the same simulator exclusion so source-based CocoaPods builds enforce the documented architecture boundary too.

The aggregate build-for-testing remains arm64-only because its test bundles consume the thin MediaPipe modules; it cannot exercise the release contract. The nightly dry run and Xcode Cloud release build use the unconstrained exporter, which fails if an affected or unaffected framework contains the wrong simulator slices.

closes [RUN_IDXP-11667](https://jira.corp.stripe.com/browse/RUN_IDXP-11667)

## Testing

- Added a test that accepts a matching arm64-only simulator binary.
- Added a test that accepts matching arm64 and x86_64 slices in either order.
- Added a test that rejects a simulator architecture mismatch and reports the expected and actual slices.
- Added a test that rejects a failed `lipo` inspection and reports the underlying error.
- All four validator tests pass locally with 10 assertions.
- Added release-export integration validation for `StripeIdentity`, `StripeCryptoOnramp`, every unaffected framework, and `Stripe3DS2`.
- Parsed both modified Xcode projects and verified their Debug and Release production target settings.
- Ruby syntax checks passed for both modified podspecs. Static inspection confirmed that Identity preserves its existing linker flags and both pods exclude x86_64 on the simulator. No CocoaPods install or integration build ran locally.
- Parsed the module and Bitrise configuration, including the nightly release-export workflow and feature-PR validator unit gate.
- An earlier feature-branch run of full Xcode archive validation passed in [Bitrise](https://app.bitrise.io/build/911b58b7-2ff8-4702-9dcd-0d842778c844): the two affected frameworks were arm64-only and the other 14 retained arm64 and x86_64. Current feature PRs run the four validator tests; nightly Bitrise and Xcode Cloud release builds run the full exporter.

## Changelog

Updated the pending Identity entry to state that `StripeIdentity` and `StripeCryptoOnramp` are arm64-only on the simulator. Other SDK frameworks retain Intel simulator support.

### Notify
cc @stripe-internal/codeowners-identity-product-services
r?

### DIFF
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -459,6 +459,7 @@ workflows:
         inputs:
         - scheme: AllStripeFrameworks
         - destination: generic/platform=iOS Simulator
+        - xcodebuild_options: ARCHS=arm64
     - deploy-to-bitrise-io@2: {}
     - save-spm-cache@1: {}
     before_run:
@@ -651,6 +652,10 @@ workflows:
         inputs:
         - content: ./ci_scripts/check_image_format.sh
         title: Check png image format is 8-bit
+    - script@1:
+        inputs:
+        - content: ruby ci_scripts/tests/framework_architecture_validator_test.rb
+        title: Test release build scripts
     - fastlane@3:
         inputs:
         - lane: build_extras
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -34,7 +34,6 @@ pipelines:
       framework-tests: {}
       lpm-confirm-flow-tests: {}
       framework-tests-and-builds-xcode-26-1: {}
-      deploy-dry-run: {}
       install-and-integration-tests: {}
       size-report: {}
       ui-tests-paymentsheet:
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -34,6 +34,7 @@ pipelines:
       framework-tests: {}
       lpm-confirm-flow-tests: {}
       framework-tests-and-builds-xcode-26-1: {}
+      deploy-dry-run: {}
       install-and-integration-tests: {}
       size-report: {}
       ui-tests-paymentsheet:
@@ -458,7 +459,6 @@ workflows:
         inputs:
         - scheme: AllStripeFrameworks
         - destination: generic/platform=iOS Simulator
-        - xcodebuild_options: ARCHS=arm64
     - deploy-to-bitrise-io@2: {}
     - save-spm-cache@1: {}
     before_run:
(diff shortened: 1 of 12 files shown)


Pearapps (dismissed):

bitrise.yml
```diff
@@ -34,6 +34,7 @@ pipelines:
       framework-tests: {}
       lpm-confirm-flow-tests: {}
       framework-tests-and-builds-xcode-26-1: {}
+      deploy-dry-run: {}
```
wooj-stripe: Why are we adding deploy-dry-run here?

modules.yaml
```diff
@@ -225,6 +231,8 @@ modules:
 - podspec: StripeCryptoOnramp.podspec
   scheme: StripeCryptoOnramp
   framework_name: StripeCryptoOnramp
+  simulator_architectures:
+    - arm64
```
wooj-stripe, line 235: having claude review this, they mentioned that we should also update the identity/crypto .podspec files to:
```
'EXCLUDED_ARCHS[sdk=iphonesimulator*]' => 'x86_64',
```
wdyt

wooj-stripe (dismissed):

bitrise.yml
mnn-stripe: This one will help running ci_scripts/framework_architecture_validator.rb while on the dry run build. Will catch such issues faster.

modules.yaml
mnn-stripe: yeah I missed this one. Let me update!

wooj-stripe (approved):

PR #7000, opened 2026-08-22
Add GET request timeout and retry options

yuki-stripe:
## Summary
- Allow modern STPAPIClient GET requests to set a per-attempt timeout
- Allow callers to control HTTP 429 retry attempts while preserving existing defaults

## Motivation
Will be useful later when CheckoutSessions confirm code polls and wants to control this

## Testing
- Verifies a custom GET timeout is applied to the outgoing URL request.
- Verifies a GET request configured with zero retries makes only one request after an HTTP 429 response.

### DIFF
--- a/StripeCore/StripeCore/Source/API Bindings/STPAPIClient.swift
+++ b/StripeCore/StripeCore/Source/API Bindings/STPAPIClient.swift
@@ -272,7 +272,7 @@ extension STPAPIClient {
     /// Make a GET request using the passed parameters.
     /// - Parameters:
     ///   - timeout: Optional timeout for each request attempt. The total request duration can be longer when retries are enabled.
-    ///   - retryCount: Number of times to retry an HTTP 429 response. Defaults to `StripeAPI.maxRetries`.
+    ///   - retriesEnabled: Whether to retry HTTP 429 responses using the standard retry policy.
     @_spi(STP) public func get<T: Decodable>(
         resource: String,
         parameters: [String: Any],
@@ -280,7 +280,7 @@ extension STPAPIClient {
         consumerPublishableKey: String? = nil,
         apiVersionOverride: String? = nil,
         timeout: TimeInterval? = nil,
-        retryCount: Int = StripeAPI.maxRetries,
+        retriesEnabled: Bool = true,
         completion: @escaping (
             Result<T, Error>
         ) -> Void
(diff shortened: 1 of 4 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/PaymentPagePollResponse.swift
gbirch-stripe, line 17: It looks like there's also a `failed_async_payment` status that isn't included here (messaged in slack)

StripeCore/StripeCore/Source/API Bindings/STPAPIClient.swift
```diff
@@ -569,7 +590,8 @@ extension STPAPIClient {
                         STPAPIClient.decodeResponse(data: data, error: error, response: response, request: request)
                     )
                 }
-            }
+            },
+            retryCount: retryCount
```
gbirch-stripe, line 594: It looks like passing `retryCount` here feeds it into `stp_performDataTask` code that makes assumptions about the amount of retries and does math accordingly. It's supposed to go 1->4->9 second delay between retires, but if you just pass in 1 retry it goes straight to a 9 second delay. Similarly if you pass in 4 it will start with a 0 second delay. So I think this code isn't suited for arbitrary retry counts being passed?

StripeCore/StripeCoreTests/API Bindings/STPAPIClientTest.swift
```diff
@@ -0,0 +1,79 @@
...
+
+import Foundation
+import OHHTTPStubs
+import OHHTTPStubsSwift
+@testable @_spi(STP) import StripeCore
+import StripeCoreTestUtils
+import XCTest
+
+final class STPAPIClientTest: APIStubbedTestCase {
+    func testGetAppliesTimeoutAndRetryCount() {
```
gbirch-stripe: Can we add more tests that test different `retryCount` behavior?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/PaymentPagePollResponse.swift
yuki-stripe: TY!

StripeCore/StripeCore/Source/API Bindings/STPAPIClient.swift
yuki-stripe: Oof confusing, changed to a retryEnabled flag

gbirch-stripe (approved):

PR #7001, opened 2026-08-22
Add Checkout Session poller

yuki-stripe:
## Summary

Add a Checkout Session poller for the confirm flow.

- Poll for up to 30 seconds with minimum start-to-start request spacing and exponential backoff after rate limits.
- Return an explicit outcome describing why polling stopped: completed, timed out, requires a payment method, failed async payment, or invalid/expired.

## Motivation

https://docs.google.com/document/d/1CWglcOHR6I8cOWKjztPfnhnVG8axucpv-j5cWIoey7E/edit?tab=t.t7xn7s6bnklb#heading=h.wtruc9k2ym04

## Testing

- Verified that each continuing poll state causes another request and each successful terminal state returns the completed outcome.
- Verified that invalid and expired states return their distinct outcome.
- Verified that a payment object requiring a new payment method returns its distinct outcome.
- Verified that a failed asynchronous payment returns its distinct outcome.
- Verified that poll request failures are retried.
- Verified start-to-start request spacing for both fast and slow responses and the remaining timeout passed to each request.
- Verified exponential rate-limit backoff and the timed-out outcome.
- Verified both modern and legacy rate-limit error representations.
- Verified that only a successful poll resets rate-limit backoff.

## Changelog

N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/CheckoutSessionPoller.swift
```diff
@@ -0,0 +1,182 @@
...
+    init(
+        apiClient: any CheckoutSessionPollingAPIClient,
+        clock: CheckoutSessionPollingClock = CheckoutSessionPollingClock()
+    ) {
+        self.apiClient = apiClient
+        self.clock = clock
+    }
+
+    /// Returns the reason polling stopped. Request failures are retried until timeout.
+    func poll(checkoutSessionId: String) async -> Outcome {
```
porter-stripe, line 72: if this is canceled, I think we retry for the full 30 seconds. Should this be async throws so we can propagate the cancellation? I'm not sure how this will be used but if it's like
```swift
let task = Task { await poll.checkoutSessionid: id) }
task.cancel()
```
It looks like the polling continues

yuki-stripe: Yeah I intentionally don't want it to be cancelable. It should only do full polling in the success case; error cases should drop off before then.

porter-stripe (dismissed):

porter-stripe (approved):

PR #7007, opened 2026-08-24
Preserve Checkout card selection after session updates

porter-stripe:
## Summary
Selecting Card in PE horizontal mode could revert to Link after collecting a shipping address. PE now preserves the completed Card selection across Checkout Session updates.

## Motivation
- Bug fix

## Testing
- New test
- Existing

## Changelog
N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetFlowControllerViewController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetFlowControllerViewController.swift
@@ -274,7 +274,9 @@ class PaymentSheetFlowControllerViewController: UIViewController, FlowController
         }
         self.savedPaymentOptionsViewController.delegate = self
         self.addPaymentMethodViewController.delegate = self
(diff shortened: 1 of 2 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/ViewControllers/PaymentSheetFlowControllerViewController.swift
```diff
@@ -274,7 +274,9 @@ class PaymentSheetFlowControllerViewController: UIViewController, FlowController
         }
         self.savedPaymentOptionsViewController.delegate = self
         self.addPaymentMethodViewController.delegate = self
-        if shouldUseLinkOnlyWalletHeader && Self.customerDefaultIsLink(configuration: configuration, elementsSession: elementsSession) {
+        if initialState.paymentOption == nil,
```
yuki-stripe, line 277: The name `initialState` is confusing me, is this representing the _previous state_?

porter-stripe: It’s the state the new VC should initialize with. During a normal rebuild it carries the current selection forward and after cancellation it restores the selection from before presentation. That’s why “previous state” isn’t quite accurate and we renamed it.

yuki-stripe (approved):

PR #7009, opened 2026-08-24
Pass up permissions to & use merchant keys during IBP+FC flow

mats-stripe:
## Summary

In an effort to support IBP+FC flows, this PR does a few things:

- Passes up the merchant token and the requested permissions when creating the Link Account session
  - If no permissions are requested, both fields are omitted
- When permissions are requested, we use the merchant's API keys during the IBP flow instead of the consumer's.
  - The equivalent web logic is here https://stripe.sourcegraphcloud.com/r/stripe-internal/mint@b091717d9ea01e2ab3af5a255d9da64a84f5129a/-/blob/pay-server/stripe-js-v3/src/linkedAccounts/shared/helpers.ts?L19-24
- When permissions are requested, we no longer call the `v1/consumers/payment_details` API at the end of the flow, and instead grab the payment details ID from the `generated_payment_detail_ids` field in the `v1/link_account_sessions/attach_payment_account` response. 

## Motivation

Correctly associate the requested permissions to the Link Account Session.

## Testing

https://github.com/user-attachments/assets/9dcff413-1e82-4415-8160-df5dbc310b45

## Changelog

N/a


StripeCore/StripeCore/Source/Connections Bindings/HostedAuthUrlBuilder.swift
```diff
@@ -103,6 +126,7 @@ import Foundation
             ? ""
             : "&"
         let updatedUrlString = urlString + joiningCharacter + joinedParameters
+        print("**** \(updatedUrlString)")
```
tillh-stripe: 👀

tillh-stripe (dismissed):

tillh-stripe (approved):

PR #7015, opened 2026-08-25
Move Checkout Apple Pay configuration to PaymentElement

yuki-stripe:
## Summary

- Move Checkout Apple Pay configuration under Payment Element configuration.
- Fix a pre-existing bug where Checkout did not pass its Apple Pay configuration to either underlying Payment Element presentation, preventing Apple Pay from appearing.
- Derive the merchant country from the Checkout Session response and require it during decoding.

This PR is stacked on #7016, which makes FlowController the single resolver for Checkout PaymentElement’s initial selection and seeds EmbeddedPaymentElement with that result.

## Testing

- Added coverage that Checkout rejects responses without a merchant country.
- Added coverage that Apple Pay settings and a server-provided merchant country reach both Payment Element presentations and that both initially select Apple Pay. Without #7016, this test fails because FlowController selects Apple Pay while EmbeddedPaymentElement has no initial selection.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Payment Element/PaymentElement+Configuration.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Payment Element/PaymentElement+Configuration.swift
@@ -17,7 +17,7 @@ extension PaymentElement {
         public init() {}
 
         /// Configuration for Apple Pay.
-        public var applePay: ApplePayConfiguration?
+        public var applePayConfiguration: ApplePayConfiguration?
 
         /// PaymentSheet offers users an option to save some payment methods for later use.
         /// Default value is `.automatic`.
@@ -144,7 +144,7 @@ extension PaymentElement {
             configuration.apiClient = apiClient
             configuration.returnURL = returnURL
             configuration.apply(linkConfiguration: linkConfiguration)
-            configuration.applePay = applePay?.paymentSheetConfiguration(
+            configuration.applePay = applePayConfiguration?.paymentSheetConfiguration(
                 merchantCountryCode: merchantCountryCode
             )
             configuration.merchantDisplayName = merchantDisplayName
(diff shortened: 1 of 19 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Payment Element/PaymentElement+Configuration.swift
```diff
@@ -16,6 +16,9 @@ extension PaymentElement {
         /// Initializes a Configuration with default values.
         public init() {}
 
+        /// Configuration for Apple Pay.
+        public var applePay: ApplePayConfiguration?
```
joyceqin-stripe: Do we want to call it `applePay` or `applePayConfiguration`? It's fine by me either way, just want to make sure we end up keeping it the same name across ECE and PE

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Payment Element/PaymentElement.swift
joyceqin-stripe, line 103: Why do we need to keep them in sync? What is the reason FlowController and Embedded have different defaults?

joyceqin-stripe: Payment Element only shows either FlowController or Embedded, right? Can it switch from one to the other? I'm having trouble understanding why we need to keep them synced here and with flowControllerDefaultsToApplePay

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Payment Element/PaymentElement+Configuration.swift
yuki-stripe: Oh I guess we call it `applePayConfiguration`, will change.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Payment Element/PaymentElement.swift
yuki-stripe: Hmmm oh no, something went wrong, this was part of #7016 not sure why it ended up here.

yuki-stripe: The problem is that PE's default payment option - and therefore CheckoutController's default, initial payment option - is calculated *before* we know whether you're using FC or Embedded.

FC and Embedded can have different defaults for ~uninteresting reasons / different choices made at different times by different people 🤷  i think limited to apple pay and link.

joyceqin-stripe (approved):

PR #7023, opened 2026-08-25
Remove Checkout Session from PaymentSheet playground

yuki-stripe:
## Summary

- Remove Checkout Session as an integration type from the PaymentSheet test playground
- Remove the disabled Checkout Session UI tests and stale test-plan exclusions
- Make the Checkout-backed EmbeddedPaymentElement and FlowController factories internal to Checkout Elements
- Keep the dedicated Checkout Elements playground unchanged

## Testing

No new tests.


Example/PaymentSheet Example/PaymentSheet Example-Shard2.xctestplan
```diff
@@ -31,7 +31,6 @@
     {
       "parallelizable" : true,
       "skippedTests" : [
-        "CheckoutSessionUITests",
```
porter-stripe, line 34: should we consider re-writing these two smoke tests for our long term playground?

porter-stripe (approved):

PR #7029, opened 2026-08-26
Move Checkout item amounts to sub-items

yuki-stripe:
## Summary
Move one-time Price amount details from the Checkout item group to each sub-item, per changes in https://docs.google.com/document/d/1KUPx71jTOsRU4rUpyF21jYG8Pg5oe-FlUInsQ1TWuCU/edit?tab=t.mbyu6dbsos8j#heading=h.8pcoo71xz60e
```
oneTimePrice.amountDetails
  → oneTimePrice.items[].amountDetails
```
Will add discounts in another PR.

## Some context 
 A unified-mode one-time-price checkout item is a group containing one or more configured Prices:
```
  one_time_price
  ├── item: T-shirt × 2
  └── item: Hat × 1

  Payment Pages already computes amounts for each individual Price:

  {
    "one_time_price": {
      "items": [
        {
          "inner_item_key": "shirt",
          "quantity": 2,
          "subtotal": 2000,
          "total": 2148,
          "tax_inclusive": 0,
          "tax_exclusive": 148,
          "tax_amounts": []
        }
      ],
      "subtotal": 2000,
      "total": 2148
    }
  }
```
  So the backend gives us both:
```
  Group totals:     one_time_price.subtotal / total
  Per-price totals: one_time_price.items[].subtotal / total / tax
```
## Testing
- Added coverage showing that two one-time Price sub-items preserve their distinct subtotals, totals, and inclusive or exclusive taxes.
- Updated coverage to verify that responses missing required per-sub-item subtotal or total values are rejected.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/PaymentPagesAPIResponse.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/PaymentPagesAPIResponse.swift
@@ -290,8 +290,6 @@ extension PaymentPagesAPIResponse {
 
     struct OneTimePrice: Decodable {
         let items: [OneTimePriceItem]
-        let subtotal: Int
-        let total: Int
     }
 
     struct OneTimePriceItem: Decodable {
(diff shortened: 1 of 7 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/API Models/PaymentPagesAPIResponse.swift
wooj-stripe: seems like this needs to be removed if we're moving it to OneTimePriceItem?

yuki-stripe: We don't need to but we could...I think that's probably better, since the API doesn't need to send these anymore, I'll make the change!

porter-stripe (approved):

PR #7031, opened 2026-08-26
Make Payment Element configuration opt-in

porter-stripe:
## Summary

Payment Element is now created only when its configuration is provided. `getPaymentElement()` remains non-optional and asserts if it is called without opting in to Payment Element.

## Motivation

Mobile Checkout Elements API review

## Testing

- New unit tests
- Existing Checkout and Payment Element tests updated

## Changelog

N/A


Example/PaymentSheet Example/PaymentSheet Example/Checkout Playground/CheckoutCartViewController.swift
```diff
@@ -850,12 +853,13 @@ final class CheckoutCartViewController: UIViewController {
...
             Task {
-                await checkout.getPaymentElement().present(from: self)
+                await checkout.getPaymentElement()?.present(from: self)
                 renderCheckout()
             }
         case .embedded:
+            guard let paymentElement = checkout.getPaymentElement() else { return }
             let viewController = CheckoutEmbeddedPaymentViewController(
-                paymentElement: checkout.getPaymentElement()
+                paymentElement: paymentElement
```
yuki-stripe: ah oh no this kind of sucks that we force everyone to deal with this being optional

porter-stripe: This was an error on my end! We should make this non-optional but make it assert like you suggested.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Configuration.swift
```diff
@@ -45,8 +45,11 @@ extension CheckoutController {
         /// Default customer details used to pre-populate Checkout integrations.
         public var defaults: Defaults = Defaults()
 
-        /// Configuration for PaymentElement.
-        public var paymentElement: PaymentElement.Configuration = .init()
+        /// Configuration for Payment Element.
+        ///
+        /// Set this property to use Payment Element with this CheckoutController.
+        /// The default value is `nil`.
+        public var paymentElement: PaymentElement.Configuration?
```
joyceqin-stripe, line 52: OOC, why did we decide to go with the pattern of requiring Payment Element configuration to be explicitly set in order to opt in? Why shouldn't a merchant be able to just use it with default configuration values without instantiating it on their own? To me, calling `getPaymentElement()` already seems like an explicit opt-in to use Payment Element

porter-stripe: You can find more details here! https://docs.google.com/document/d/1CWglcOHR6I8cOWKjztPfnhnVG8axucpv-j5cWIoey7E/edit?tab=t.uxtqimnqdt9j#heading=h.82la9li1w67k

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/CheckoutController.swift
```diff
@@ -314,7 +319,9 @@ public final class CheckoutController: ObservableObject {
 
     /// Returns the PaymentElement for this CheckoutController instance.
     public func getPaymentElement() -> PaymentElement {
-        return paymentElement
+        assert(configuration.paymentElement != nil, "Set Configuration.paymentElement before calling getPaymentElement().")
+        stpAssert(paymentElement != nil, "PaymentElement should be initialized when Configuration.paymentElement is set.")
```
yuki-stripe, line 332: :applause: correct usage of assert vs stpAssert

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Checkout/CheckoutConfirmationTests.swift
```diff
@@ -575,7 +577,7 @@ final class CheckoutConfirmationTests: APIStubbedTestCase {
     private func makePaymentMethodFlow(
         for checkout: CheckoutController,
         option: CheckoutController.PaymentMethodConfirmationParameters.Option
-    ) -> CheckoutController.CheckoutConfirmationFlow {
+    ) throws -> CheckoutController.CheckoutConfirmationFlow {
```
yuki-stripe: sorry why is this throws now?

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Checkout/CheckoutTestHelpers.swift
```diff
@@ -159,16 +159,21 @@ enum CheckoutTestHelpers {
     /// - Parameters:
     ///   - apiResponse: The Checkout Session response returned by the stubbed `/init` request.
     ///   - configuration: An optional base configuration for test-specific settings.
+    ///   - includePaymentElement: Whether to configure Payment Element when the base configuration omits it.
     ///   - stubAllOutgoingRequests: Whether to stub every outgoing API request made by the client, or only the initialization request.
     @MainActor
     static func makeConfiguration(
         apiResponse: PaymentPagesAPIResponse = makeOpenSession(),
         configuration: CheckoutController.Configuration? = nil,
+        includePaymentElement: Bool = true,
```
yuki-stripe: should probably be `paymentElement: PaymentElement.Configuration? = nil or .init()` for more flexibility.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Checkout/CheckoutConfirmationTests.swift
porter-stripe: erroneous change, reverted. thanks!

yuki-stripe (approved):

PR #7035, opened 2026-08-27
Assert when Currency Selector Element is not configured

porter-stripe:
## Summary

Calling `getCurrencySelectorElement()` without configuring Currency Selector Element now triggers an SDK assertion. The getter still returns `nil` without asserting when it was configured but Adaptive Pricing is unavailable. The Checkout playground tracks the opt-in state before calling the getter.

## Motivation

Mobile Checkout Elements API review

## Testing

- Added coverage for accessing Currency Selector Element without configuration
- Existing Currency Selector tests
- PaymentSheet Example build

## Changelog

N/A


gbirch-stripe (dismissed):
lgtm

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/Checkout/CheckoutUnitTests.swift
```diff
@@ -68,24 +68,6 @@ final class CheckoutUnitTests: XCTestCase {
         XCTAssertNil(configuration.currencySelectorElement)
     }
 
-    func testGetCurrencySelectorElementAssertsWhenNotConfigured() async throws {
```
gbirch-stripe, line 71: Why was this test removed?

porter-stripe: Asserts (unlike stpAsserts) aren't testable

gbirch-stripe: Understand now — only stp asserts are testable, not normal asserts

gbirch-stripe (approved):

PR #7039, opened 2026-08-27
Allow clearing Checkout shipping addresses

porter-stripe:
## Summary

`updateShippingAddress` can now clear the Checkout Session's local shipping address. When shipping drives automatic tax, clearing the address removes its detailed tax region fields while retaining the previous country required by the server.

## Motivation

Checkout Elements API review

## Testing

- Added coverage for clearing a local shipping address.
- Added coverage that automatic tax clearing retains only the previous country.
- Added coverage that a failed tax update preserves the previous shipping address.
- All Checkout unit tests and Shipping Address Element presentation tests pass.

## Changelog

N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/CheckoutController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/CheckoutController.swift
@@ -267,6 +267,7 @@ public final class CheckoutController: ObservableObject {
         if session.shouldSendTaxRegion(for: "shipping") {
             // The Checkout Session update endpoint requires tax_region[country] and does not
             // support clearing tax_region, so keep the previous country.
+            // TODO(porter) When migrating to the CheckoutClient API, stop sending country only and send nil
             let countryOnlyAddress = Address(country: shippingAddress.address.country)
             try await performUpdate(
                 .setTaxRegion(countryOnlyAddress),
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/CheckoutController.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/CheckoutController.swift
@@ -237,32 +237,37 @@ public final class CheckoutController: ObservableObject {
         try await performUpdate(.setTaxRegion(address), canUpdateWhileSheetPresented: canUpdateWhileSheetPresented)
     }
 
-    /// Sets the shipping address for this checkout.
+    /// Sets or clears the shipping address for this checkout.
     ///
     /// The address is stored locally and merged into PaymentSheet configuration
     /// when presenting payment UI. If automatic tax is enabled and the tax
     /// address source is "shipping", the address is also sent to the server to
-    /// compute updated tax amounts.
+    /// compute updated tax amounts. Clearing the address removes its detailed tax
+    /// region fields, but keeps its country because the server requires one.
     ///
     /// - Parameters:
     ///   - name: The customer's full name.
-    ///   - address: The shipping address to set. To reset tax computation
+    ///   - address: The shipping address to set, or `nil` to clear it. To reset tax computation
     ///     to a country-only region, pass a ``CheckoutController.Address`` with just the country.
     /// - Throws: ``CheckoutError`` if the session is not open, or if
     ///   the server request fails.
     public func updateShippingAddress(
         name: String? = nil,
-        address: Address
+        address: Address?
     ) async throws {
-        if let allowedCountries = session.allowedShippingCountries,
+        if let address,
+           let allowedCountries = session.allowedShippingCountries,
            !allowedCountries.contains(address.country) {
             throw CheckoutError.invalidShippingCountry(countryCode: address.country)
         }
-        let shippingAddress = Session.ShippingAddress(name: name, address: address)
+        let shippingAddress = address.map { Session.ShippingAddress(name: name, address: $0) }
         guard session.shippingAddress != shippingAddress else { return }
-        if session.shouldSendTaxRegion(for: "shipping") {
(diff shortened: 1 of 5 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/CheckoutController.swift
```diff
@@ -237,23 +237,28 @@ public final class CheckoutController: ObservableObject {
...
 
-    /// Sets the shipping address for this checkout.
+    /// Sets or clears the shipping address for this checkout.
     ///
     /// The address is stored locally and merged into PaymentSheet configuration
     /// when presenting payment UI. If automatic tax is enabled and the tax
     /// address source is "shipping", the address is also sent to the server to
-    /// compute updated tax amounts.
+    /// compute updated tax amounts. Clearing the address removes its detailed tax
+    /// region fields, but keeps its country because the server requires one.
```
yuki-stripe: Oh no, this docstring is no good. Hard to tell what, if any, info in here is relevant to users vs. an implementation detail. We can probably just go with the API Review here: "    /// Use this method to update the Customer's shipping address."

This is part of the stuff we need to audit.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/CheckoutController.swift
```diff
@@ -237,23 +237,28 @@ public final class CheckoutController: ObservableObject {
...
     ///
     /// - Parameters:
     ///   - name: The customer's full name.
-    ///   - address: The shipping address to set. To reset tax computation
+    ///   - address: The shipping address to set, or `nil` to clear it. To reset tax computation
     ///     to a country-only region, pass a ``CheckoutController.Address`` with just the country.
```
yuki-stripe: > To reset tax computation to a country-only region, pass a ``CheckoutController.Address`` with just the country.

Is this useful information?  I can't tell what it means. 

I think the contract is just, 'set the shipping address', you don't need to worry about 'resetting tax computation' and we can delete this.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/CheckoutController.swift
```diff
@@ -270,6 +275,21 @@ public final class CheckoutController: ObservableObject {
         }
     }
 
+    private func clearShippingAddress() async throws {
+        guard let shippingAddress = session.shippingAddress else { return }
+        if session.shouldSendTaxRegion(for: "shipping") {
+            // The server requires a country, so keep the previous one when clearing the tax region.
+            let countryOnlyAddress = Address(country: shippingAddress.address.country)
```
yuki-stripe: Is this to say the update API can't clear an address?  This is a bit surprising, I don't see why it couldn't.

porter-stripe: Yeah, also find this surprising but country is required and the request fails if not provided:
https://stripe.sourcegraphcloud.com/r/stripe-internal/mint@acfa6a48bac14d036599e9b58e6139ae002f543f/-/blob/pay-server/lib/payment_pages/api/payment_page/update_method.rb?L119-123

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/CheckoutController.swift
porter-stripe: Changed this doc comment to match the API review

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/CheckoutController.swift
porter-stripe: Good point, I'll take a pass and audit the doc comments.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/CheckoutController.swift
yuki-stripe: Does ewcs also just send the previous country?  I'm just worried if something goes wrong with this behavior; thinking through it - worst case you clear the address and...the tax is incorrectly calculated instead of marked as pending/requires shipping address? 

cc @gbirch-stripe did you come across this restriction?

porter-stripe: Yeah web does this but kinda differently, instead they reset it to a derived default country, with precedence.
```
 1. PaymentPage.geocoding.country_code
  2. Currency mapping for USD, GBP, AUD, CAD, EUR, MXN, or SGD
  3. The device locale’s region
  4. The first allowed shipping country
  5. US
 ```
I can implement this in a follow up, but honestly I feel like using the previously set country is better than that logic IMO. https://stripe.sourcegraphcloud.com/r/stripe-internal/mint/-/blob/pay-server/stripe-js-v3/src/checkout/outer/customCheckout/actions/addressHelpers/AutomaticTaxHelpers.ts?L377

yuki-stripe: Ack thanks, we shouldn't put this logic on the client. I'm going to bring this up w/ the backend, if this is really what we want, we can put it there, and just send country = nil to the API.

 Can you add a TODO to stop setting the country when we move to the CheckoutClient for now?

porter-stripe: Added!

yuki-stripe (approved):

PR #7040, opened 2026-08-27
Add placeholder projects for erroneous Xcode Cloud checks

davidme-stripe:
This didn't work


davidme-stripe:
Hosted result: Xcode Cloud found the placeholder project and scheme, archived it, and then attempted App Store distribution export. Making the placeholder exportable could cause the external tenant to sign or submit an unintended blank app. We are abandoning this workaround as outside acceptable safety bounds.

PR #7045, opened 2026-08-27
Add Korean cards support to PaymentSheet

porter-stripe:
## Summary
Adds Korean cards as a supported payment method in PaymentSheet. Merchants can collect and confirm Korean card payments and save them for future use with the required mandate.

### Korean cards

https://github.com/user-attachments/assets/38e49907-fc3c-4729-9a90-97ccd8d03dfb

https://github.com/user-attachments/assets/80df2ef6-a5cd-4865-a83a-c7d234c82065

https://github.com/user-attachments/assets/84a28da6-d694-4b3f-a2c9-c258098ed448

## Motivation
Korean payment methods

## Testing
- Verifies intent parameters include mandate data.
- Verifies the mandate for payments saved for future use and SetupIntents.
- Verifies payment and setup confirmation flows.
- Verifies the return URL requirement for payment and setup.

## Changelog
See diff


wooj-stripe (approved):
Code seems fine, but do we need to penny test before releasing?

porter-stripe:
@wooj-stripe typically for these LPMs in a web view we don't, also would need someone in Korea to do so. I can reach out to the LPM team and ask.

PR #7049, opened 2026-08-28
Add ACSS Debit API bindings

porter-stripe:
## Summary
Adds ACSS Debit API bindings, including payment method params and response decoding. Also handles asynchronous processing and microdeposit verification states in PaymentHandler.

PaymentSheet continues to treat ACSS Debit as unsupported in this PR. Compatibility cases keep its exhaustive switches compiling until #7048 enables and renders the payment method.

## Motivation
https://jira.corp.stripe.com/browse/MOBILESDK-2064

## Testing
- New tests cover parameter encoding, response decoding with nullable bank details, type conversion, and inferred mandate data.
- New tests cover SetupIntents awaiting microdeposits and asynchronous PaymentIntents that are still processing.
- A clean all-framework build passes.

## Changelog
N/A


StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
```diff
@@ -772,6 +772,8 @@ public class STPPaymentHandler: NSObject {
                 if let setupIntent = setupIntent,
                    error == nil,
                    setupIntent.status == .succeeded
+                    || (setupIntent.status == .requiresAction
+                        && strongSelf.isNextActionSuccessState(nextAction: setupIntent.nextAction))
```
porter-stripe, line 776: We need this because `verifyWithMicrodeposits` completes successfully while the SetupIntent remains in `requiresAction`. This matches the existing [`confirmSetupIntent` handling](https://github.com/stripe/stripe-ios/blob/554c4c489404bbed4ab91b33d6a7bf22249f6e38/StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift#L577-L584), which treats that terminal next-action state as success.

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
gbirch-stripe: Is this still true with the behavior change below?

gbirch-stripe: Same here

gbirch-stripe (commented):
STPIntentAction:52 also needs to be updated. Can you get someone who is experienced with STPPaymentHandler to look at this? The changes in that file make me pretty nervous and I don't think I have the expertise to properly verify

StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift
porter-stripe: yeah should be updated!

porter-stripe:
@gbirch-stripe The STPPaymentHandler changes already exist in `confirmSetupIntent`: https://github.com/stripe/stripe-ios/blob/5d9e573d31a/StripePayments/StripePayments/Source/PaymentHandler/STPPaymentHandler.swift#L577-L585. 

That API confirms the SetupIntent in the SDK and handles its next action. `handleNextAction` is the separate path for merchants who confirm on their server and then pass a `requires_action` SetupIntent to the SDK. Its completion wrapper was missing the same check. This makes the two paths consistent, so I guess it was a previous gap/bug.

cc @david-stripe would you mind double checking the STPPaymentHandler changes?

PR #7061, opened 2026-08-28
Rename runServerUpdate closure parameter 🪿✨

porter-stripe:
[Minion run](https://orbit.corp.stripe.com/sessions/ad457673-cd94-47f3-88d7-be3649309028)

## Summary
Renames the `runServerUpdate` closure parameter from `updateFunction` to `update` in its declaration, implementation, and documentation.

## Motivation
Uses a concise local parameter name for the server-update closure.

## Testing
Not run; this is a local identifier-only rename. `git diff --check` passes.

## Changelog
No changelog entry; this does not change runtime behavior.


gbirch-stripe (approved):
Do you mind linking the API decision for the change next time?

PR #7070, opened 2026-08-28
Add SeQura API bindings

porter-stripe:
## Summary

Adds API bindings for SeQura.

## Motivation

Adds the bindings needed to support SeQura in the mobile SDKs. See [EMEAPP-205](https://jira.corp.stripe.com/browse/EMEAPP-205).

## Testing

Added coverage for decoding and encoding SeQura payment methods and parameters.

Verified deferred intents exclude SeQura from unsupported off-session payment method options.

## Changelog

Added a Payments changelog entry.

### DIFF
--- a/StripePayments/StripePayments/Source/API Bindings/Models/PaymentMethods/STPPaymentMethodEnums.swift
+++ b/StripePayments/StripePayments/Source/API Bindings/Models/PaymentMethods/STPPaymentMethodEnums.swift
@@ -108,6 +108,8 @@ import Foundation
     case naverPay
     /// A PAYCO payment method
     case payco
+    /// A SeQura payment method
+    case sequra
     /// An unknown type.
     case unknown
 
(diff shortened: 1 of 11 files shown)


StripePayments/StripePayments/Source/API Bindings/Models/PaymentMethods/STPPaymentMethodEnums.swift
```diff
@@ -214,6 +216,8 @@ import Foundation
             return "Naver Pay"
         case .payco:
             return "PAYCO"
+        case .sequra:
+            return "SeQura"
```
davidme-stripe, line 220: Weirdly they're not consistently using either "seQura" or "SeQura" on their own site, so I think either is fine here. https://www.sequra.com/es/terminos-legales

davidme-stripe (approved):

PR #7075, opened 2026-08-28
Make Checkout payment option clearing async

porter-stripe:
## Summary

Makes CheckoutController.clearPaymentOption async and throwing so merchants can wait for related Checkout Session updates and handle failures. When billing-based automatic tax uses the selected payment method's address, clearing the option recalculates tax with only its country. A failed update leaves the payment option selected.

## Motivation

CheckoutSessions

## Testing

- Clearing a payment option recalculates automatic tax with only the previous billing country.
- A failed tax update returns an error and keeps the payment option selected.
- Existing Checkout and Payment Element tests.

## Changelog

N/A

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Internal.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Internal.swift
@@ -77,6 +77,18 @@ extension CheckoutController {
         )
     }
 
+    /// Resets billing-based automatic tax to the selected payment option's country.
+    ///
+    /// - Warning: Only call this from an operation already serialized by
+    ///   `enqueueSessionUpdate(_:)`.
+    func resetBillingTaxRegionIfNecessary() async throws {
+        guard session.shouldSendTaxRegion(for: "billing"),
+              let country = session.paymentOption?.billingDetails?.address.country?.nonEmpty else {
+            return
+        }
+        try await applySessionUpdate(.setTaxRegion(Address(country: country)))
+    }
+
     // MARK: - Session Updates
 
     /// Waits for all in-flight session updates (mutations, etc.) to complete.
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Internal.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Internal.swift
@@ -71,56 +71,10 @@ extension CheckoutController {
 
     // MARK: - Payment Option
 
-    /// Updates only the published payment option state.
-    /// Use `updatePaymentOption(to:updateElement:)` when the change and its Checkout Session
-    /// side effects must succeed atomically.
     func setPaymentOption(_ paymentOption: Session.PaymentOptionDisplayData?) {
         dangerouslySetSessionDirectly(
             session.makeCopyOverriding(paymentOption: .newValue(paymentOption))
         )
-    }
-
-    /// Updates the Checkout Session and Payment Element with a new payment option.
-    ///
-    /// For billing-based automatic tax, this updates the tax region from the new option's
-    /// billing address. Clearing an option retains only its previous billing country because
-    /// the Checkout Session update endpoint does not support removing the tax region.
-    /// `updateElement` runs only after any required Checkout Session update succeeds.
-    func updatePaymentOption(
-        to paymentOption: Session.PaymentOptionDisplayData?,
-        updateElement: @MainActor @escaping () -> Void
-    ) async throws {
-        try await enqueueSessionUpdate {
-            if let billingTaxRegion = self.billingTaxRegion(for: paymentOption) {
-                try await self.applySessionUpdate(.setTaxRegion(billingTaxRegion))
-            }
-            updateElement()
-            self.setPaymentOption(paymentOption)
-        }
-    }
-
-    private func billingTaxRegion(
-        for paymentOption: Session.PaymentOptionDisplayData?
-    ) -> Address? {
-        guard session.shouldSendTaxRegion(for: "billing") else {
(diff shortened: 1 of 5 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Internal.swift
```diff
@@ -71,7 +71,14 @@ extension CheckoutController {
...
 
-    func setPaymentOption(_ paymentOption: Session.PaymentOptionDisplayData?) {
+    func setPaymentOption(_ paymentOption: Session.PaymentOptionDisplayData?) async throws {
+        if paymentOption == nil {
+            try await updateBillingTaxRegionIfNecessary(address: nil)
+        }
+        publishPaymentOption(paymentOption)
+    }
+
+    func publishPaymentOption(_ paymentOption: Session.PaymentOptionDisplayData?) {
```
yuki-stripe: Hm not sure I like `publish` as the name for this method, which afaiu "synchronously sets payment option without updating billing tax region".

yuki-stripe: `func dangerouslySetPaymentOptionDirectly`?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Internal.swift
```diff
@@ -71,7 +71,14 @@ extension CheckoutController {
 
     // MARK: - Payment Option
 
-    func setPaymentOption(_ paymentOption: Session.PaymentOptionDisplayData?) {
+    func setPaymentOption(_ paymentOption: Session.PaymentOptionDisplayData?) async throws {
+        if paymentOption == nil {
+            try await updateBillingTaxRegionIfNecessary(address: nil)
+        }
+        publishPaymentOption(paymentOption)
```
yuki-stripe: This is surprising method behavior that it only updates billing when po is nil.

porter-stripe: Yeah, this reads confusingly. It's b/c PE updates the billing directly and doesn't call setPaymentOption to do it. I've moved the tax region update to outside this function now.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Internal.swift
porter-stripe: After a few changes this function doesn't exist anymore, but i renamed the existing b/c I think it's better.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Payment Element/PaymentElement.swift
```diff
@@ -190,16 +190,21 @@ extension PaymentElement {
...
         guard !paymentSheetFlowController.didPresentAndContinue else {
             assertionFailure("Clearing the payment option after presenting PaymentElement is not implemented. File a feature request if you need this.")
             return
         }
+        guard let checkout else {
+            stpAssertionFailure("PaymentElement unexpectedly lost its CheckoutController.")
+            return
+        }
+        try await checkout.updateBillingTaxRegionIfNecessary(address: nil)
+        checkout.dangerouslySetPaymentOptionDirectly(nil)
```
yuki-stripe, line 206: It'd be nice to have a single `setPaymentOption` that does this work once and is reused. I don't suppose that'd be easy to do?

porter-stripe: Yeah, I tried this in an earlier version! It gets a little awkward b/c EPE/PS.FC are the sources of truth for the payment option, but PE is responsible for updating Checkout’s po. By the time PE sees a selection, EPE/PS.FC have already handled the billing update, while clearing is different b/c PE kicks off the update.

A shared `setPaymentOption` would either special-case nil (the surprising behavior you called out above) or maybe something else?

yuki-stripe: I was thinking `setPaymentOption` would unconditionally take care of calling `updateBillingTaxRegionifNecessary` and EPE/PS.FC call `setPaymentOption` rather than `updateBilling` on continue.

I like that there's exactly one way to update the payment option - via `setPaymentOption` - and so you can't accidentally *not* update billing tax region.

porter-stripe: Ah interesting. I feel pretty strongly EPE/PS.FC should not call setPaymentOption on PE. Today EPE/PS.FC have no knowledge of PE, so we'd need to introduce that to them which may be fine, but we also introduce some other interesting stuff, like now the payment option changes while the sheet is open, also maybe this is fine too. 

More broadly, though, I think the ownership graph gets confusing. EPE/PS.FC own the actual selection, PE decides which one is the source of truth. And then PE receives po updates through the EPE delegate or PS.FC publisher. This would add another path where the children write upstream directly. AFAICT this gives us more paths for the same state rather than one clear owner/source of truth. Also payment option updates may also arrive twice in the setPaymentOption and the delegate/publisher, also maybe fine but weird.

yuki-stripe: Discussed potential refactor more offline, if we end up doing it we can make this method call `setPaymentOption` in that future.

yuki-stripe (approved):

PR #7085, opened 2026-09-01
Re-enable LPMs for CheckoutSession

porter-stripe:
## Summary

Adds Checkout Session confirmation to the existing PaymentSheet local payment method confirm-flow matrix and re-records supported payment methods against unified mode. Copies client betas to Checkout Session requests and uses deterministic polling so redirect and asynchronous payment methods can be covered without waiting for live completion.

BLIK remains excluded from the Checkout Session case because unified mode drops `blik_code` before PaymentIntent confirmation. We left the failure details and request ID in the linked bug-bash doc. https://docs.google.com/document/d/1Tv7SJUx1MbWFpgMJ83d23YeyHK1agtjSUhWcAUw7kgQ/edit?tab=t.0

## Motivation

Checkout Session LPM coverage

## Testing

- Re-recorded the local payment method confirm-flow matrix.
- Verified Sequra and Vipps Checkout Sessions reach their expected redirect state and pass in replay.
- BLIK Checkout remains excluded; direct and deferred BLIK flows still run.

## Changelog

N/A


StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetLPMConfirmFlowTests.swift
```diff
@@ -86,6 +86,12 @@ final class PaymentSheetLPMConfirmFlowTests: STPNetworkStubbingTestCase {
         }
     }
 
+    struct CompletedCheckoutSessionPoller: CheckoutSessionPolling {
```
yuki-stripe, line 89: can we see if we don't need it?

yuki-stripe (approved):

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetLPMConfirmFlowTests.swift
porter-stripe: I tried this but removing the fake and re-recording the Checkout Session-only tests gets most of the suite passing, but PayNow and PromptPay stay active long enough to hit the test's 25s timeout before the real poller's 30s timeout. Going to keep the fake one for now.

PR #7086, opened 2026-09-01
CryptoOnramp SDK: Moves Consumer Session Client Secret to HTTP Headers in GET Requests

mliberatore:
## Summary
For `GET /v1/crypto/…` APIs that require a consumer session client secret, this parameter has been moved to HTTP headers.

## Motivation

Slack Link: [Stripe](https://stripe.slack.com/archives/C094P3BRBL1/p1788279018599289) | [Lickability](https://lickability.slack.com/archives/C094P3BRBL1/p1788279018599289)

## Testing

1. Created a new EU user, completing KYC, User Attestation, and tax ID collection to ensure I hit the `GET /v1/crypto/internal/crs_carf_declaration` and `GET /v1/crypto/internal/identifier_requirements` endpoints successfully.
2. ~From an existing account, completed a checkout to ensure calls to `GET /v1/crypto/internal/platform_settings` were successful~. This functionality was removed before opening the PR.

## Changelog

N/A

### DIFF
--- a/StripeCryptoOnramp/StripeCryptoOnramp/Source/Internal/API Bindings/STPAPIClient+CryptoOnramp.swift
+++ b/StripeCryptoOnramp/StripeCryptoOnramp/Source/Internal/API Bindings/STPAPIClient+CryptoOnramp.swift
@@ -127,7 +127,7 @@ extension STPAPIClient {
         let endpoint = "crypto/internal/identifier_requirements"
         return try await get(
             resource: endpoint,
-            parameters: try credentialsParameters(consumerSessionClientSecret: consumerSessionClientSecret)
+            consumerSessionClientSecret: consumerSessionClientSecret
         )
     }
 
@@ -172,7 +172,7 @@ extension STPAPIClient {
         let endpoint = "crypto/internal/crs_carf_declaration"
         return try await get(
             resource: endpoint,
-            parameters: try credentialsParameters(consumerSessionClientSecret: consumerSessionClientSecret)
+            consumerSessionClientSecret: consumerSessionClientSecret
         )
     }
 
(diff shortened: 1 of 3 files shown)


StripeCryptoOnramp/StripeCryptoOnramp/Source/Internal/API Bindings/STPAPIClient+CryptoOnramp.swift
```diff
@@ -419,9 +421,45 @@ private extension STPAPIClient {
             }
         }
     }
+
```
mliberatore, line 416: @jeanregisser this `private` extension method below necessarily duplicates some of the coordination done in `StripeCore`'s `STPAPIClient.request(…)` method called by the other `get(…)` overloads, as there's no `additionalHeaders` parameter in that API. That coordination includes the calls to `configuredRequest`, `stp_addParameters`, `stp_performDataTask` and `STPAPIClient.decodeResponse`.

An alternate approach here would be to add `additionalHeaders` to the callback, URL, and `Promise`-returning `STPAPIClient.get(...)` overloads, then thread these headers through `STPAPIClient.request(...)` into its call to `configuredRequest(for:apiVersionOverride:additionalHeaders:)`. This would expand StripeCore’s API surface for just this use case.

Please let me know which approach (this vs. modifying StripeCore) you prefer.

jeanregisser: Thanks, I'd prefer modifying StripeCore to avoid future drift and have a single code path for requests.
We can merge this one now and submit a follow up PR though.

jeanregisser (approved):
Thanks 🚀

PR #7097, opened 2026-09-02
Fix embedded payment method row layout

yuki-stripe:
## Summary

- Avoid forcing layout while restoring an initially selected embedded payment method row.
- Keep row height constraints idempotent when the row state changes.

## Motivation

Fixes the layout constraint break reported in [RUN_MOBILESDK-5410](https://jira.corp.stripe.com/browse/RUN_MOBILESDK-5410).

## Testing

No new tests.

Verified the existing embedded update flow completes without broken constraint alerts.

https://github.com/user-attachments/assets/b4525c72-d453-4466-8314-8d906c92d1ca

## Changelog

Not user-facing.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButton.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButton.swift
@@ -306,6 +306,9 @@ class RowButton: UIView, EventHandler {
     // MARK: Helper
 
     func makeSameHeightAsOtherRowButtonsIfNecessary() {
+        heightConstraint?.isActive = false
+        heightConstraint = nil
+
         // To make all RowButtons the same height, set our height to the tallest
         // standard variant (a RowButton with text and a plain sublabel).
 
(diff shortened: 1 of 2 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Vertical Main Screen/RowButton.swift
```diff
@@ -306,6 +306,9 @@ class RowButton: UIView, EventHandler {
     // MARK: Helper
 
     func makeSameHeightAsOtherRowButtonsIfNecessary() {
+        heightConstraint?.isActive = false
+        heightConstraint = nil
+
```
yuki-stripe, line 311: change to this method is unrelated to the fix but more correct

gbirch-stripe (commented):
Do you mind attaching a video of the new UX? Is this no-animation visual change visible or jarring to customers?

yuki-stripe:
> Do you mind attaching a video of the new UX? Is this no-animation visual change visible or jarring to customers?

there's no new UX except that the broken constraints are no longer broken. attached a video - currently when you switch modes w/ the card you'll see a bunch of broken constraints.

gbirch-stripe (approved):

PR #7102, opened 2026-09-02
Add CSE height update delegate

porter-stripe:
## Summary
Currency Selector Element can change height when customers show or hide exchange rate details. Adds an EPE-style delegate so UIKit merchants can coordinate surrounding layout animations instead of the element laying out the entire view hierarchy. SwiftUI continues to resize automatically.

## Motivation
Adaptive Pricing

## Testing
- New unit tests
- Existing snapshot tests

## Changelog
N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Currency Selector Element/CurrencySelectorElementUIView.swift
```diff
@@ -22,7 +22,10 @@ public final class CurrencySelectorElementUIView: UIView {
         selectorView?.setEnabled(enabled)
     }
 
-    private weak var delegate: CurrencySelectorElementDelegate?
+    var needsUpdateSuperviewHeight: () -> Void = {}
```
gbirch-stripe: It makes me pretty nervous that everywhere that calls `needsUpdateSuperviewHeight` breaks if for some reason it is not set. Can we enforce it or have some safer way than just removing the old updaters and hoping that all consumers of these components set `needsUpdateSuperviewHeight`?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/ExpandableDetailView.swift
```diff
@@ -181,7 +178,6 @@ final class ExpandableDetailView: UIView {
         detailHeightConstraint?.constant = 0
         expandedContentLabel.setHiddenIfNecessary(true)
         expandedContentLabel.alpha = 0
-        invalidateLayoutUpHierarchy()
```
gbirch-stripe, line 184: Why are we safe to remove this?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/ExpandableDetailView.swift
```diff
@@ -111,20 +113,16 @@ final class ExpandableDetailView: UIView {
...
 
         if isExpanded {
             detailHeightConstraint?.constant = 0
             expandedContentLabel.setHiddenIfNecessary(false)
             expandedContentLabel.alpha = 0
-            invalidateLayoutUpHierarchy()
-            layoutContainer.layoutIfNeeded()
         } else {
             detailHeightConstraint?.constant = expandedContentLabel.bounds.height
-            invalidateLayoutUpHierarchy()
```
gbirch-stripe, line 125: Why are we safe to remove these various `invalidLayoutUpHierarchy`s?

Tests/ReferenceImages_64/StripePaymentSheetTests.CheckoutCurrencySelectorElementSnapshotTests/testDefaultAppearanceWithCompletionHandler__iOS26@3x.png
gbirch-stripe, line 1: Do we have tests that covered what they look like expanded (including sizing)?

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Currency Selector Element/CurrencySelectorElementUIView.swift
porter-stripe: Yeah, good call. Made this a required init arg all the way down.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/ExpandableDetailView.swift
porter-stripe: updateCaption on the CSE now calls invalidateContentSize() after updating the detail view, so this path is covered there.

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Elements/ExpandableDetailView.swift
porter-stripe: The callback now propagates up to the CSE, which invalidates its intrinsic size and notifies SwiftUI. For UIKit, we call the merchant’s delegate inside the animation block so they can lay out the right container.

Tests/ReferenceImages_64/StripePaymentSheetTests.CheckoutCurrencySelectorElementSnapshotTests/testDefaultAppearanceWithCompletionHandler__iOS26@3x.png
porter-stripe: added!

StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Currency Selector Element/CurrencySelectorElement.swift
```diff
@@ -43,10 +43,16 @@ public final class CurrencySelectorElement {
         configuration: Configuration,
         delegate: CurrencySelectorElementCheckoutDelegate
     ) async {
+        // Assign the callback's weak reference after the views finish initializing.
+        weak var element: CurrencySelectorElement?
```
gbirch-stripe, line 47: Pretty weird pattern here. Is there a particular reason for it or can we just put `[weak self]` in the `needsUpdateSuperviewHeight` closure?

porter-stripe: No particular reason. I thought capturing self before initialization would fail, but [weak self] works here. I’ll simplify it.

porter-stripe: Wait actually, I spoke too soon. [weak self] fails to compile with the variable 'self.view' used before being initialized. The local weak reference lets us pass the closure while constructing uiView, then wire it up after both stored views are initialized. Sorry for the back and forth!

gbirch-stripe: Ah makes sense. Weird stuff!

gbirch-stripe (approved):

PR #7103, opened 2026-09-03
Fix resolved Link brand with direct Instant Debits

mats-stripe:
## Summary

The resolved Link brand wasn't being passed into the direct Financial Connections flow when launched directly from MPE. It was working when going through the Link wallet -> add bank, but not direct from the MPE bank tab. This fixes that.

## Motivation

Consistent Link / Onelink branding

## Testing

Before:

https://github.com/user-attachments/assets/7e3b9beb-b980-4825-86d6-147bc70e0f04

After:

https://github.com/user-attachments/assets/800b542a-1584-46a3-a809-cfc85058d77d

## Changelog

N/a


sophiah2-stripe (approved):
🙏

PR #7104, opened 2026-09-03
ECE tax region reset

joyceqin-stripe:
## Summary

When the Apple Pay sheet is canceled without authorizing payment, restore whatever the previous tax region was.
## Motivation

ECE
## Testing

https://github.com/user-attachments/assets/777fb159-0ec2-47aa-9de6-1453ae91c097

https://github.com/user-attachments/assets/2da1a450-795b-46d1-943d-19ec124c3055

## Changelog

N/A


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Internal.swift
```diff
@@ -264,6 +264,9 @@ extension CheckoutController {
                 updatedSessionAPIResponse,
                 shippingAddress: shippingAddress
             )
+            if case .setTaxRegion(let address)? = update {
+                currentTaxRegion = address
```
porter-stripe: what happens if commitSession throws after the API update succeeds? I think we'd leave current tax region at the old value. Should this be moved before commit session?

porter-stripe (approved):

PR #7139, opened 2026-09-10
Consume passive captcha tokens after use

toluo-stripe:
## Summary

Consume passive CAPTCHA tokens when they are retrieved for a request so the same single-use token cannot be returned again.

The initial token is still preloaded, but replacement tokens are created on demand instead of being prefetched automatically.

## Motivation

`PassiveCaptchaChallenge` cached its completed token task until time-based expiration. A second payment method creation or confirmation attempt could therefore reuse a token that had already been sent and invalidated.

https://stripe.slack.com/archives/CFA2HJ99A/p1788976177411529?thread_ts=1788906921.701889&cid=CFA2HJ99A

## Testing

- Added coverage that simultaneous consumers start separate CAPTCHA challenges.
- Added coverage that repeated card Radar option requests each start a fresh CAPTCHA challenge.
- Updated passive CAPTCHA coverage to verify consumed tokens are no longer cached and unused expired tokens are replaced.

## Changelog

N/A


joyceqin-stripe (approved):

toluo-stripe:
cc @yuki-stripe

PR #7143, opened 2026-09-10
Add Checkout payment option billing details type

yuki-stripe:
## Summary
- Add a Checkout-owned billing details model for selected payment options
- Match the Android API nesting and nullable field semantics
- Convert PaymentSheet billing details without exposing PaymentSheet types

## Motivation
Don't use PaymentSheet types in Checkout Elements

## Testing
- Added coverage that selected payment option billing details preserve the customer name, email, and address.
- Added coverage that missing billing addresses are represented as nil.
- Verified session debug descriptions continue masking customer information.

### DIFF
--- a/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Session.swift
+++ b/StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Session.swift
@@ -229,7 +229,7 @@ extension CheckoutController.Session {
         /// A user facing string representing the payment method; e.g. "Apple Pay" or "····4242" for a card
         public let label: String
         /// The billing details associated with the customer's desired payment method
-        public let billingDetails: PaymentSheet.BillingDetails?
+        public let billingDetails: BillingDetails?
         /// A string representation of the customer's desired payment method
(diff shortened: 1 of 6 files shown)


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Session.swift
```diff
@@ -229,13 +229,49 @@ extension CheckoutController.Session {
...
+            public let address: Address?
+
+            /// The customer's email address.
+            public let email: String?
+
+            /// The customer's full name.
+            public let name: String?
+
+            /// The customer's phone number.
+            public let phone: String?
```
gbirch-stripe, line 253: I thought we weren't doing phone?

yuki-stripe: @jaynewstrom-stripe had the same thought. To explain: we aren't doing phone/name/email in BDCC. That's different from here, which represents what PM details we've collected. 

Just because you can't require PM phone collection via BDCC doesn't mean we won't collect phone; we will if the PM requires it.

gbirch-stripe (approved):

PR #7144, opened 2026-09-10
Migrate Checkout Session tests to Mobile Elements playground backend

yuki-stripe:
## Summary

- Migrate the shared Checkout Session test helper from the CI backend to the playground raw API using ui_mode=mobile_elements.
- Keep LPM coverage, disabled customer-update coverage, and the temporary Adaptive Pricing opt-out case on the legacy backend.
- Refresh affected network recordings.

## Testing

No new tests.


StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/STPAPIClient+CheckoutSessionTest.swift
```diff
@@ -149,7 +149,8 @@ final class STPAPIClientCheckoutSessionTest: STPNetworkStubbingTestCase {
     func testInitCheckoutSessionPaymentWithAdaptivePricingDisabled() async throws {
         // Same session config as above (DE location, adaptive pricing active automatically)
         // but client passes adaptivePricingAllowed: false
-        let checkoutSessionResponse = try await STPTestingAPIClient.shared.createCheckoutSession(
+        // TODO: Use Mobile Elements once it honors adaptive_pricing[allowed]=false like Custom Checkout.
```
yuki-stripe, line 152: Polo is fixing this soon

StripePayments/StripePaymentsTestUtils/STPTestingAPIClient+Swift.swift
```diff
@@ -263,4 +343,56 @@ extension STPTestingAPIClient {
...
+
+    private struct TestingBackendError: LocalizedError {
+        let message: String
+        var errorDescription: String? { message }
+    }
+
+    private func playgroundMerchant(for merchantCountry: String?) -> String {
+        let merchant = merchantCountry ?? "us"
+        return merchant.count == 2 ? merchant.uppercased() : merchant
+    }
```
wooj-stripe, line 376: i think this merchant can either be an ID or country, so i think this function would read better as:
```
let playgroundMerchant = merchantCountry ??  "us"
return playgroundMerchant.count == 2 ? playgroundMerchant.uppercased() : playgroundMerchant
```

StripePayments/StripePaymentsTestUtils/STPTestingAPIClient+Swift.swift
```diff
@@ -211,6 +213,87 @@ extension STPTestingAPIClient {
         returnURL: String? = nil,
         useOneTimePrice: Bool = true,
         additionalParameters: [String: Any] = [:]
+    ) async throws -> CreateCheckoutSessionResponse {
+        guard useOneTimePrice else {
+            throw TestingBackendError(message: "Mobile Elements Checkout Sessions require a one-time price item.")
+        }
+
+        let merchant = playgroundMerchant(for: merchantCountry)
```
wooj-stripe: per other comment,  i think this would read better as `let playgroundMerchant = ...`

StripePayments/StripePaymentsTestUtils/STPTestingAPIClient+Swift.swift
```diff
@@ -211,6 +213,87 @@ extension STPTestingAPIClient {
         returnURL: String? = nil,
         useOneTimePrice: Bool = true,
         additionalParameters: [String: Any] = [:]
+    ) async throws -> CreateCheckoutSessionResponse {
+        guard useOneTimePrice else {
+            throw TestingBackendError(message: "Mobile Elements Checkout Sessions require a one-time price item.")
```
wooj-stripe: It's interesting to call this a TestingBackendError,  but i don't feel too strongly about this since it's test code.

StripePayments/StripePaymentsTestUtils/STPTestingAPIClient+Swift.swift
```diff
@@ -211,6 +213,87 @@ extension STPTestingAPIClient {
...
+        if automaticTax {
+            sessionParameters["automatic_tax"] = ["enabled": true]
+        }
+        if let customerEmailLocation {
+            sessionParameters["customer_email"] = "test+location_\(customerEmailLocation)@example.com"
+        }
+        if let customerID {
+            sessionParameters["customer"] = customerID
+        }
+        sessionParameters.merge(additionalParameters) { _, override in override }
```
wooj-stripe: In general, I am not a fan of having overriding/merging in test code -- I won't block on this PR, but feels like we should avoid this if we can.

StripePayments/StripePaymentsTestUtils/STPTestingAPIClient+Swift.swift
```diff
@@ -211,6 +213,87 @@ extension STPTestingAPIClient {
...
+            params: [
+                "merchant": merchant,
+                "stripe_version": Self.checkoutMobileElementsAPISettings,
+                "request_params": sessionParameters,
+            ]
+        )
+        let publishableKeyResponse: PlaygroundPublishableKeyResponse = try await makePlaygroundRequest(
+            endpoint: "publishable_key?merchant=\(merchant)",
+            method: "GET"
+        )
```
wooj-stripe, line 281: I'm surprised that we can't just return the merchant's publishable key in the first request?

StripePayments/StripePaymentsTestUtils/STPTestingAPIClient+Swift.swift
yuki-stripe: Oops, this is silly, we can remove the `useOneTimePrice: Bool = true` parameter

StripePayments/StripePaymentsTestUtils/STPTestingAPIClient+Swift.swift
yuki-stripe: We could, like also return the pk in the CS create request? That seems good.

wooj-stripe (approved):

PR #7145, opened 2026-09-10
Refactor Session local state handling

yuki-stripe:
## Summary
Two ideas here, I don't want to make separate PRs b/c I have a ton already.

1.  Replace `makePublicSession()` with `Session.init(apiResponse:localState:)`, since the API response alone cannot construct a complete session. Initial load passes .empty; refreshes pass the preserved local state.
2. Group local Checkout Session state into `Session.LocalState` and replace the clunky `makeCopyOverriding` with just using LocalState
```
        var localState = session.localState
        localState.shippingAddress = shippingAddress.resolved(currentValue: localState.shippingAddress)
        localState.paymentOption = paymentOption.resolved(currentValue: localState.paymentOption)

        if let apiResponse {
            session = Session(apiResponse: apiResponse, localState: localState)
        } else {
            session.localState = localState
        }
```

## Testing
No new tests.


StripePaymentSheet/StripePaymentSheet/Source/PaymentSheet/Checkout/Checkout+Internal.swift
```diff
@@ -211,24 +211,24 @@ extension CheckoutController {
 
     /// Enqueues a serialized session update.
     ///
-    /// - If `update` is non-nil, the side effect (if any) is applied first, then the
-    ///   API mutation is performed and the session is updated from the response.
-    /// - If `update` is nil, the side effect is applied locally without making a network request.
+    /// - If `update` is non-nil, the API mutation is performed, then the session is updated from
+    ///   the response and the local state mutation is applied.
+    /// - If `update` is nil, the local state mutation is applied without making a network request.
```
gbirch-stripe, line 216: Why are we swapping the order to do the mutation second instead of first?

yuki-stripe: The docstring is wrong :/ line 253 shows we do the update first, then call `commitSession` with the local shipping address.

gbirch-stripe (dismissed):

gbirch-stripe (approved):

PR #7147, opened 2026-09-11
[Identity] Custom Colors for CTA

jkelle-stripe:
## Summary
Adds an internal `primaryButtonStyle` configuration setting to allow Link to edit this style. Previously, a user could edit the `UIWindow.tint` to update this CTA color. Link requires a distinction between this tint color (which is used by native SwiftUI elements and other custom elements of the SDK) and the color used by the CTA (which Link would prefer be black/white depending on dark mode). This is behind a `STP` annotation to signal that it's for internal usage.

## Motivation
Link design would like configuration of the CTAs used in the Identity SDK ([figma](https://www.figma.com/design/UKdEOqZIBMigsA8iw3KZ1C/%E2%9A%93%EF%B8%8F--Project-Iron?node-id=6005-142068&m=dev)).

## Demo

These demos shows the testing done for this future using a non-committed playground toggle to add the black/white button style Link will use. These demos shows that the tint color is still respected while the custom CTA button style is in use.

| Full Demo | Dark/Light Mode |
| --- | --- |
| <video src="https://github.com/user-attachments/assets/d07f4bd0-56d7-464a-ad02-eca45c614bb7" width="300" /> | <video src="https://github.com/user-attachments/assets/da251acd-40cf-493d-8ef4-7cf5f12cebe8" width="300" /> |

### DIFF
--- a/Example/IdentityVerification Example/IdentityVerification Example/PlaygroundViewController.swift
+++ b/Example/IdentityVerification Example/IdentityVerification Example/PlaygroundViewController.swift
@@ -5,7 +5,7 @@
 //  Created by Mel Ludowise on 3/3/21.
 //
 
-import StripeIdentity
+@_spi(STP) import StripeIdentity
 @_spi(STP) import StripeUICore
 import UIKit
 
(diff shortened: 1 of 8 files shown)


luisv-stripe (approved):
LGTM

PR #7152, opened 2026-09-12
Migrate LPM Checkout confirmation tests to mobile_elements

yuki-stripe:
## Summary

Migrate LPM Checkout confirmation tests to the playground backend with `ui_mode = mobile_elements` and re-record their network fixtures.

Stacked on #6762.

## Motivation

Standardize all Checkout Session tests and examples on the playground backend. The LPM confirmation flow tests were the last suite left to migrate; this PR moves them over.

## Testing

- Re-enabled coverage verifies that BLIK Checkout confirmation accepts the entered code.
- Klarna payment and setup forms explicitly verify that the country dropdown is present.
- The full LPM suite passed all 52 tests before the final BLIK/Klarna updates; those updates passed fresh live recordings and replay afterward. Alipay replay also passed after the shared setup-currency adjustment.
- Formatting and lint pass.

## Changelog

None; test-only changes.


StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetLPMConfirmFlowTests.swift
```diff
@@ -1280,15 +1309,23 @@ extension PaymentSheetLPMConfirmFlowTests {
             if shouldTest(.deferredIntent) {
                 intents.append(TestIntent("Deferred PaymentIntent - client side confirmation", makeDeferredIntent(deferredCSC)))
             }
-            // TODO: Re-enable once unified-mode Checkout forwards `blik_code` to PaymentIntent confirmation.
-            if shouldTest(.checkoutSession), paymentMethod != .blik {
-                let checkoutSessionResponse = try await STPTestingAPIClient.shared.createLegacyCheckoutSession(
+            if shouldTest(.checkoutSession) {
+                // CI and playground maintain separate merchant mappings; customers cannot cross accounts.
```
yuki-stripe, line 1313: [AI] The CI and playground backends maintain separate country-to-merchant mappings, and some countries (including the US) resolve to different accounts. Checkout cases therefore use the API client returned by the playground helper and create their customers and saved payment methods on that account; they cannot reuse the CI backend's objects.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetLPMConfirmFlowTests.swift
```diff
@@ -1280,15 +1309,23 @@ extension PaymentSheetLPMConfirmFlowTests {
...
+                        merchantCountry: merchantCountry.rawValue
+                    )
+                } else {
+                    checkoutCustomer = nil
+                }
+                let checkoutSessionResponse = try await STPTestingAPIClient.shared.createCheckoutSession(
                     types: paymentMethodTypes,
                     currency: currency,
                     amount: amount,
                     merchantCountry: merchantCountry.rawValue,
```
yuki-stripe, line 1327: [AI] The suite's Mexico value is `mex`, which is the CI backend's alias. The playground independently maps merchants and expects `MX`, so we translate at the playground boundary.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetLPMConfirmFlowTests.swift
```diff
@@ -1280,15 +1309,23 @@ extension PaymentSheetLPMConfirmFlowTests {
...
+                if customer != nil {
+                    checkoutCustomer = try await STPTestingAPIClient.shared.createCheckoutCustomer(
+                        email: "lpm-confirm@example.com",
+                        merchantCountry: merchantCountry.rawValue
+                    )
+                } else {
+                    checkoutCustomer = nil
+                }
+                let checkoutSessionResponse = try await STPTestingAPIClient.shared.createCheckoutSession(
                     types: paymentMethodTypes,
```
yuki-stripe: [AI] Vipps needs its preview header when the Checkout Session is created, not just on later SDK requests: Checkout retains the creation API version for the underlying PaymentIntent. Other payment methods keep the existing API settings.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetLPMConfirmFlowTests.swift
```diff
@@ -217,7 +219,8 @@ final class PaymentSheetLPMConfirmFlowTests: STPNetworkStubbingTestCase {
     }
 
     func testBLIKConfirmFlows() async throws {
-        try await _testConfirm(intentKinds: [.paymentIntent], currency: "PLN", paymentMethodType: .blik, merchantCountry: .BE,
+        // Use a merchant available in both backends; the playground has no Belgium mapping.
+        try await _testConfirm(intentKinds: [.paymentIntent], currency: "PLN", paymentMethodType: .blik, merchantCountry: .FR,
```
yuki-stripe, line 223: [AI] BLIK Checkout coverage is re-enabled now that the backend forwards `blik_code` during confirmation. The playground has no Belgium mapping, so this uses the French merchant shared by both backends while keeping PLN. Fresh live recording and replay passed.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetLPMConfirmFlowTests.swift
```diff
@@ -587,19 +605,25 @@ final class PaymentSheetLPMConfirmFlowTests: STPNetworkStubbingTestCase {
...
+        // Use the French merchant shared by both backends so all Klarna flows use the same account.
+        // The backends use different US accounts. The playground US account has
+        // enable_klarna_continue_authorization_on_refresh_calls enabled, which makes refreshing
+        // an unfinished Klarna payment cancel it and surface an authentication failure instead of cancellation.
         try await _testConfirm(intentKinds: [.paymentIntent],
-                               currency: "USD",
+                               currency: "EUR",
                                paymentMethodType: .klarna,
-                               merchantCountry: .US,
+                               merchantCountry: .FR,
```
yuki-stripe, line 615: [AI] The two backends use different US accounts. The playground US account has `enable_klarna_continue_authorization_on_refresh_calls` enabled: when this test returns from an unfinished Klarna redirect, the normal refresh cancels the payment and surfaces an authentication failure instead of the expected cancellation. All Klarna flows now use the shared French merchant and EUR (including explicit setup currency), preserving the normal refresh/cancellation path without changing merchant flags or adding a test workaround. The forms also assert that the country field exists, without forcing a country selection.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetLPMConfirmFlowTests.swift
wooj-stripe: What does "both backends" refer to? checkout_client and payment_pages?

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetLPMConfirmFlowTests.swift
```diff
@@ -587,19 +605,25 @@ final class PaymentSheetLPMConfirmFlowTests: STPNetworkStubbingTestCase {
     }
 
     func testKlarnaConfirmFlows() async throws {
+        // Use the French merchant shared by both backends so all Klarna flows use the same account.
+        // The backends use different US accounts. The playground US account has
+        // enable_klarna_continue_authorization_on_refresh_calls enabled, which makes refreshing
+        // an unfinished Klarna payment cancel it and surface an authentication failure instead of cancellation.
```
wooj-stripe: These comments don't seem relevant now that we aren't using the US merchant.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetLPMConfirmFlowTests.swift
```diff
@@ -1289,15 +1318,23 @@ extension PaymentSheetLPMConfirmFlowTests {
...
-            if shouldTest(.checkoutSession), paymentMethod != .blik {
-                let checkoutSessionResponse = try await STPTestingAPIClient.shared.createLegacyCheckoutSession(
+            if shouldTest(.checkoutSession) {
+                // CI and playground maintain separate merchant mappings; customers cannot cross accounts.
+                let checkoutCustomer: String?
+                if customer != nil {
+                    checkoutCustomer = try await STPTestingAPIClient.shared.createCheckoutCustomer(
+                        email: "lpm-confirm@example.com",
+                        merchantCountry: merchantCountry.rawValue
+                    )
```
wooj-stripe: Not sure if i understand this.  if customer. !=nil, then create a checkoutCustomer?  I'd expect that we'd want to correlate the customerId to this new customer?  Also, I would have expected to add a uuid to the email for uniqueness.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetLPMConfirmFlowTests.swift
```diff
@@ -1687,6 +1725,12 @@ extension PaymentSheetLPMConfirmFlowTests {
             for testIntent in intents {
                 let description = testIntent.description
                 let intent = testIntent.intent
+                var configuration = configuration
+                let apiClient = testIntent.checkoutAPIClient ?? apiClient
+                configuration.apiClient = apiClient
+                if testIntent.checkout != nil {
+                    configuration.customer = nil
+                }
```
wooj-stripe: Why are we nil-ing out the customer if testIntent.checkout != nil?

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetLPMConfirmFlowTests.swift
yuki-stripe: Yeah this is just explaining why we switched from BE to FR. It's referring to the playground backend and the CI test backend. I should delete the comment though, it doesn't really make sense outside of this PR's context.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetLPMConfirmFlowTests.swift
yuki-stripe: Ah thanks for spotting this. I've undone the change and instead found a merchant that is the same across both backends (that was the root problem this was trying to solve). 

Do we need a UUID? Emails don't need to be unique.

StripePaymentSheet/StripePaymentSheetTests/PaymentSheet/PaymentSheetLPMConfirmFlowTests.swift
yuki-stripe: Same here, deleted too

wooj-stripe (approved):

PR #7154, opened 2026-09-14
Fix currency selector caption layout in SwiftUI

porter-stripe:
## Summary
The currency selector caption can get clipped in SwiftUI when it wraps onto multiple lines. On iOS 16 and later, size the selector using the width proposed by SwiftUI so the full caption, including Show details, fits as the available width changes.

## Motivation
CheckoutSessions

## Testing
- Added a test that checks the full multiline caption fits at several widths with the largest supported caption font.
- Tests not run (draft PR). Formatting and SwiftLint passed.

## Changelog
[Fixed] Prevent the currency selector caption from being clipped in SwiftUI on iOS 16 and later.


joyceqin-stripe (approved):

porter-stripe:
Closing this since #7102 landed a broader version of the same SwiftUI sizing fix, including height invalidation and updated snapshot coverage.

PR #7157, opened 2026-09-14
Exposes Additional Headers Parameter on STPAPIClient

mliberatore:
## Summary

Exposes `additionalParameters` on the `get`, `post`, and `delete` helper methods. Originally this was done with a helper in Crypto Onramp, but I was asked to modify the `StripeCore` API for this.

Additionally, we need to support this for an upcoming `POST` request, but I’m adding it to `DELETE` for parity as well.

## Motivation
https://github.com/stripe/stripe-ios/pull/7086#discussion_r4006768598

## Testing
1. Ran the new tests.
1. Re-ran the testing steps in #7086.

## Changelog

Updated `CHANGELOG.md`


StripeCore/StripeCore/Source/API Bindings/STPAPIClient.swift
```diff
@@ -558,7 +589,7 @@ extension STPAPIClient {
                 additionalHeaders: [
                     "Content-Length": String(format: "%lu", UInt(formData?.count ?? 0)),
                     "Content-Type": "application/x-www-form-urlencoded",
-                ]
+                ].merging(additionalHeaders) { _, newValue in newValue }
```
gbirch-stripe: Should we make this case insensitive? HTTP headers are

mliberatore: Sure, good suggestion. I’ve added some verbose comments in the code, and an additional test to help ensure the ordering of headers being applied doesn't get accidentally refactored away. 

Note on the change in behavior, that prior to this PR, `configuredRequest` would behave differently than your suggestion, so I’ve adjusted that method in this update as well. Previously, `additionalHeaders` would be applied in such a way that the "winning" header between the defaults and additional header would depend entirely on dictionary iteration order. From `master`, see line `128`. This dictionary could've potentially been modified here to contain multiple keys with different casing, and the following loop order would determine which got applied last:

https://github.com/stripe/stripe-ios/blob/29c2ef15d635d3e246839568e98a2721d756cbe8/StripeCore/StripeCore/Source/API%20Bindings/STPAPIClient.swift#L128-L132

gbirch-stripe (approved):
Looks good, thanks!

PR #7182, opened 2026-09-16
Add email and phone to crypto onramp KycInfo and map them from Apple Pay

xuhui-stripe:
## Summary
- KycInfo.email / KycInfo.phone as defaulted trailing init parameters, documented as prefill-only and not part of the KYC submission payload, with phone explicitly not normalized to E.164.
- KycInfo(payment:) reads emailAddress and phoneNumber?.stringValue from the billing contact, falling back to the shipping contact, trims whitespace, and now returns a value when only contact fields are present.
- Tests: defaults nil, email-only, phone-only, shipping-contact fallback, billing preferred over shipping, display-formatted phone preserved verbatim, whitespace-only dropped; existing name-only/address-only/nil cases unchanged.


StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/KycInfo.swift
```diff
@@ -100,47 +121,32 @@ public struct KycInfo: Equatable {
...
+        self.email = email
+        self.phone = phone
     }
 }
 
 extension KycInfo {
 
     /// Creates a `KycInfo` from Apple Pay billing information.
-    /// Returns `nil` if the `PKPayment` does not contain any usable billing name or address fields.
+    /// Returns `nil` if the `PKPayment` does not contain any usable billing name, address, email, or phone fields.
```
mliberatore: Feel free to reword as you see fit, but I think we should include "shipping" in the docs here.
```suggestion
    /// Returns `nil` if the `PKPayment` does not contain any usable billing or shipping contact name, address, email, or phone fields.
```

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/KycInfo.swift
```diff
@@ -100,47 +121,32 @@ public struct KycInfo: Equatable {
...
-
-            let trimmedFamilyName = familyName.trimmingCharacters(in: .whitespacesAndNewlines)
-            guard !trimmedFamilyName.isEmpty else {
-                return nil
-            }
-
-            return trimmedFamilyName
-        }()
+        let firstName = Self.trimmedNonEmptyValue(billingContact?.name?.givenName)
+        let lastName = Self.trimmedNonEmptyValue(billingContact?.name?.familyName)
```
mliberatore: Since we're now accepting information from `shippingContact`, is there any reason to not fall back to the `shippingContact` for `givenName` and `familyName` as well? It seems like this choice was deliberate, which is ok, so I’m just trying to understand it better.

mliberatore (commented):
Looks good overall, just a docs suggestion and clarifying question.

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/KycInfo.swift
xuhui-stripe: Updated!

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/KycInfo.swift
xuhui-stripe: Yes. Thanks for the suggestion. I updated here as well.

mliberatore (dismissed):

PR #7183, opened 2026-09-16
Allow crypto onramp Apple Pay before Link authentication

xuhui-stripe:
Apple Pay before Link authentication.

- getPlatformSettings takes an optional crypto customer ID and omits crypto_customer_id when absent (ui_mode=headless unchanged), so the platform API client resolves publishable-key-only.
- getPlatformApiClient() no longer requires an ID; with one present the request is identical to today's.
- New setCryptoCustomerId(_:) (used by registerLinkUser, authenticateUserWithToken, authorize) stores the ID and discards a cached pre-auth platform API client; logOut() clears it too.
- createCryptoPaymentToken()'s crypto customer ID gate is untouched.
- Tests: no crypto_customer_id sent when absent, still sent when present, pre-auth resolution succeeds with the platform key, caching plus invalidation after the ID arrives, token creation still fails with no payment method.


mliberatore (approved):
👍

StripeCryptoOnramp/StripeCryptoOnramp/Source/Components/CryptoOnrampCoordinator.swift
```diff
@@ -999,18 +1000,28 @@ private extension CryptoOnrampCoordinator {
...
             return platformApiClient
         }
 
-        guard let cryptoCustomerId = await cryptoCustomerState.getCustomerId() else {
-            throw Error.missingCryptoCustomerID
-        }
-
         // Fetch platform settings and create API client
+        let cryptoCustomerId = await cryptoCustomerState.getCustomerId()
         let platformSettings = try await apiClient.getPlatformSettings(cryptoCustomerId: cryptoCustomerId)
```
yiman-stripe, line 1025: Does EU support Apple Pay today?

AI found an issue: When cryptoCustomerId is nil, this obtains the default US MOR key, which Apple Pay uses to create the PaymentMethod, but later could be consumed under another.

Is this a valid concern?

xuhui-stripe: Yes, this is a valid concern. A missing `cryptoCustomerId` prevents `createCryptoPaymentToken()` only until authentication completes. The integrator can collect Apple Pay first, authenticate the user, and then call `createCryptoPaymentToken()`. The Apple Pay PaymentMethod remains selected across that transition, so it could otherwise have been created with the default key and later used when creating a token for a differently routed customer.

#7184 addresses the pre-auth routing case by forwarding the same merchant-provided country hint to both platform settings and payment-token creation.

xuhui-stripe: Basically I think EU can use apple pay, they can present the apple pay element after this PR. But they can't create a payment token. `createCryptoPaymentToken()` still has a guard to make sure crypto customer id is not nil. 

There could be a case where the MOR they get here is different from the one they get from the crypto customer mapped MOR. I will add a validation to make sure two of them are the same in PR 7184.

PR #7188, opened 2026-09-17
Reuse LinkController from Identity for Networked Identity

fedefrappi:
## Summary
SPI additions to `LinkController` so the Identity SDK can reuse Link login without Link's screens:
- `restoreConsumerSession(consumerSessionClientSecret:consumerPublishableKey:)` to pick up a Link session started in another module (e.g. crypto onramp)
- `startVerification` / `confirmVerification` without presenting UI
- `lookupConsumer` sets `linkAccount` right away, so the next call sees the new account
- `PaymentSheetLinkAccount.consumerPublishableKey` and the Networked Identity sign-up consent action
- `LinkRequestSurface.identity` (value pending backend deployment)

## Motivation
Networked Identity (save and reuse ID documents with Link) in the native Identity SDK. Stacked on `kpa/networked-id`; the Identity side is in the next PR.

## Testing
Added unit tests for session restore and the consumer publishable key.

## Changelog
None, SPI only.


bcapps:
@mats-stripe Hey Mat! We're currently working on Networked Identity as part of our work for the Identity team, and one of the requirements of this feature is to reuse logged-in Link accounts and/or allow logging into a Link account to associate with the Identity flow. 

Since this will seemingly require `LinkController` changes in some form, I asked one of our engineers to draft up a proposal for how it could look to expose what we need for the feature – would you mind taking a look at it as a member of the Link team for some overall feedback? @fedefrappi should be able to answer any more detailed questions or feedback you might have.

PR #7211, opened 2026-09-21
[Identity] Set Custom Height in Primary and Secondary Button Styles

jkelle-stripe:
Committed-By-Agent: goose

## Summary
Updates the primary and secondary button styles to include a height configuration. This is using an internal configuration setting so only Link will set this value.

## Motivation
Link shows buttons with a 52px height.

## Demo

https://github.com/user-attachments/assets/5ee2d20c-95ef-4f91-8f6c-f55ec7f92924


StripeIdentity/StripeIdentity/Source/NativeComponents/Views/IdentityFlowView.swift
```diff
@@ -476,12 +476,33 @@ extension StripeUICore.Button {
...
+
+    private func updateMinimumHeightConstraint(to minimumHeight: CGFloat?) {
+        let identifier = "IdentityFlowView.minimumButtonHeight"
+        constraints.first { $0.identifier == identifier }?.isActive = false
+
+        guard let minimumHeight else {
+            return
+        }
+
+        let constraint = heightAnchor.constraint(greaterThanOrEqualToConstant: minimumHeight)
```
jkelle-stripe, line 502: Using `>=` so that multiline text could in theory grow the button height

